PR #47, opened 2013-03-08
IDEA-99287 ANT toolwindow: prevent duplicate build files

VISTALL:


naXa777:
@eugenezh @ignatov what are we waiting for? :)

please merge it or leave a comment indicating why this can't be done.

eugenezh:
Thanks for pointing to the "duplication" problem. We have fixed the problem but in a different way.
We can't accept the proposed solution because 
1. the method is too low-level. It is used from the initialization code (config loading) as well as from UI actions code and is not supposed to check for duplicates.
2, throwing AntNoFileException for the file that is actually an ant file violates the semantics of this exception and may confuse the corresponding exception-handling code

PR #122, opened 2013-11-08
Possibility opening tasks in browser using Issue URL pattern for generic...

pochemuto:
When using Task plugin configured with generic task server is no way to open current task in browser. Menu item is disabled because Idea doesn't know task url. Assembla doesn't include direct url to task in tickets.xml and if we want to make it we should to get some addition information like space id and tiket number. So, issue url template can help with it: substitute matched groups from ticket pattern and get link to issue.

tt

### DIFF
--- a/plugins/tasks/tasks-core/src/com/intellij/tasks/generic/GenericRepository.java
+++ b/plugins/tasks/tasks-core/src/com/intellij/tasks/generic/GenericRepository.java
@@ -21,7 +21,9 @@ import org.apache.commons.httpclient.methods.PostMethod;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


naXa777:
hey @pochemuto 
1. what's issue number in YouTrack? https://youtrack.jetbrains.com/
2. is this issue still reproducible? there are chances it's fixed somewhere between 2013 and 2019...
    1. if no ->
        * just close this Pull Request.
    2. if yes ->
        * @ghost, @dmitry-avdeev why is it still Open? do you need more time to review?

pochemuto:
@naXa777 It looks like it's possible to provide issue url now. Even if server doesn't provide it I suppose you can build it using string concatination with xpath to issue id. 

image

PR #137, opened 2013-12-05
add ghost mode to tool window properties (IDEABKL-3654)

ol-loginov:
It addresses 7-years-old issue here http://youtrack.jetbrains.com/issue/IDEABKL-3654

This patch adds an option 'Ghost mode' to tool panel. With this option enabled strip button for tool window is hidden from docking toolbar.

Below are two screens about it - before and after checking that option in context menu
step1
step2

Windows is still enabled and can be shown via menu View->Tool Windows or any other way. 
This patch is actually affects only visibility of strip button.

And the state (if changed from visible to ghost only) is saved in workspace.xml


ol-loginov:
Hi guys.

Is there a problem with my pull request? Why it's delayed for years without apply or reject?

Vassiliy-Kudryashov:
I'm here to deal with it as soon as possible.

Vassiliy-Kudryashov:
The patch hasn't been applied "as is" but it was used for implementation. Thank you!

ol-loginov:
Always welcome.
Then this request might be closed, I think.

PR #140, opened 2013-12-26
SSH agent support for git and svn integrations

charles-dyfis-net:
This version allows SSH agent-based authentication to be used for authenticating to SVN or git, as discussed in the ticket http://youtrack.jetbrains.com/issue/IDEA-52034
- Separate out old Trilead SSH2 build213 into private namespace, to avoid conflict with CVS plugin (which depends on a Trilead SSH2 build with patches which were never accepted upstream and do not cleanly merge).
- Import Trilead SSH2 build218 (modern upstream, actively maintained by TMateSoft) for use by non-CVS plugins.
- Import jsch-agentproxy 0.0.6
- Upgrade SVNKit to 1.7.11
- Add user interface to SVN plugin allowing agent-based authentication to be selected.


pchel-:
Feature added in 8d82a44b44082fdfe0631fbdf625d56e4eb8453d.

PR #154, opened 2014-02-23
Fixes path so that local modules get priority

allfro:
By using `sys.path.extend()` the local source and content directories are given lower priority than the global python site-path. This can be problematic if the module being developed is also installed in the site-path.


Elizaveta239:
Hi @allfro! Sorry for a very late response.
It's a default starting command for Python Console and it can be changed in Python Console Settings "Starting Script". I don't think we should change it for all the users.
Thank you very much for your interest!

PR #158, opened 2014-03-04
IDEA-121616 Plugin API improvement: there no ability to detect any settings change in the idea from plugin.

vkravets:
Propose of improvement of Plugin API, to be able to detect any settings 
change in the IDEA Platform product from a plugin.

http://devnet.jetbrains.com/message/5510639

[Review on Reviewable](https://reviewable.io/reviews/jetbrains/intellij-community/158)

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/components/SettingsChangeListener.java
+++ b/platform/platform-impl/src/com/intellij/openapi/components/SettingsChangeListener.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2000-2014 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


ignatov:
@vkravets, thanks for the PR, please rebase the PR and clarify your use case, cc @chashnikov.

chashnikov:
Sorry for the late reply and thank you for your work! 

However the suggested implementation works only if settings are changed via Settings dialog. But some settings may be changed by accessing the corresponding service via API (e.g. when user invokes an intention action) or when a project configuration is reimported from Maven/Gradle or reloaded from the configuration files after user updates the project from VCS. So it won't work reliably.

We have plans to create a unified model to store all settings, and it'll support notification about changes, but until it's done I think it's better to describe which specific setting you want to listen for changes, and we'll be able to provide a special listener for it which will work inside its service so all the cases will be supported.

PR #167, opened 2014-04-09
Give MavenImporters a chance to set user properties.

xian:
Subclasses of `MavenImporter` can override `customizeUserProperties()`.
Improved robustness of maven 2/maven 3 switching in `MiscImportingTest`.

Fixes [IDEA-123761 MavenImporters should be able to set properties for import](http://youtrack.jetbrains.com/issue/IDEA-123761)


vladsoroka:
Hi Christian, I'm sorry for such a long silence. I've looked at the change and found it useful. It's a pity that it was not merged back in 2014 year. I've merged it with the current master branch using the following commit: https://github.com/JetBrains/intellij-community/commit/9fc5d9dedc38c7fe3f78e2d901f374bd19afd5e7
Thanks for the contribution!

PR #196, opened 2014-06-26
IDEA-87163: Make the remove-all-splits action always work

Lenbok:
Currently, the UnsplitAll action is arbitrarily disabled when there are only two windows, which means that people who have this bound to a key find that it doesn't work in all the cases where a split is present. This patch removes that restriction.

I have submitted a contributor agreement and was notified today that it has been accepted.


Lenbok:
This was fixed ages ago.

PR #197, opened 2014-07-01
Resource bundle filtering

derkork:
This is a fix for http://youtrack.jetbrains.com/issue/IDEA-86919  and adds a filter for invalid elements to the built-in resource bundle editor.


ignatov:
@derkork, please rebase the PR, cc @batya239.

derkork:
I don't think it is necessary to do this anymore. This feature has been incorporated into IntelliJ around 2015, so this is obsolete now.  I have marked https://youtrack.jetbrains.com/issue/IDEA-86919 as obsolete now.

PR #200, opened 2014-07-17
IDEA-106381 GUI Designer: JComboBoxes list does not populated under JDK7...

pdradx:
IDEA-106381 GUI Designer: JComboBoxes list does not populated under JDK7 and works fine under JDK6. Removed generics specifications from interface signatures added in JDK1.7 and above.

### DIFF
--- a/plugins/ui-designer/src/com/intellij/uiDesigner/PsiPropertiesProvider.java
+++ b/plugins/ui-designer/src/com/intellij/uiDesigner/PsiPropertiesProvider.java
@@ -17,6 +17,7 @@ package com.intellij.uiDesigner;
 
 import com.intellij.openapi.module.Module;
 import com.intellij.psi.*;
+import com.intellij.psi.impl.source.PsiClassReferenceType;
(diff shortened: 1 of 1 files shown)


plugins/ui-designer/src/com/intellij/uiDesigner/PsiPropertiesProvider.java
```diff
@@ -74,7 +75,12 @@ public HashMap getLwProperties(final String className) {
       }
 
       final PsiType type = getter.getReturnType();
-      final String propertyClassName = type.getCanonicalText();
+      final String propertyClassName;
+      if (type instanceof PsiClassReferenceType) {
+        propertyClassName = ((PsiClassReferenceType)type).rawType().getCanonicalText();
+      } else {
+        propertyClassName = type.getCanonicalText();
+      }
```
ignatov: It's a good candidate for a new method.

ignatov:
Already fixed, @pdradx, thanks for PR.

PR #215, opened 2014-10-16
Hi DPI Improvements

dgodfrey:
These commits improve the layout and size of some of the dialogs on a hidpi screen by doubling the padding in various places and doubling the default size of the TipsPanel.

I have some screen captures of compatibility mode_, before & after, although this just seems to scale them to 100% width :(
\_ IE windows just doubles everything.

I haven't signed the contributor agreement yet, but will do so if this patch is liable to get included in some form or another. I'd like to submit some more along these lines as well, as using IntelliJ in compatibility mode  on a hi-dpi screen is painful.


bulenkov:
I think it make sense to close this pull request. Most of HiDPI issues are covered now by combination of changes in JetBrains Runtime (our JRE) + JBUI class.

PR #228, opened 2014-11-30
optimize plugin version checking when update intellij platform version

shlxue:
Allow update the same build number for a plugin.


ignatov:
@shlxue is it still relevant? If so, please update the PR, cc @alexander-lobas.

PR #234, opened 2015-01-13
Fixes to CoreJavaFileManager

pTalanov:
* Fix for top level class with dollars in name.
* Fix for java file with several top level classes.
* Checking scope.
* Tests.

Needed for Kotlin. See https://youtrack.jetbrains.com/issue/KT-6444


yole:
Merged manually. Thanks!

PR #238, opened 2015-01-29
check myAssertThreading before using assertIsDispatchThread

martinwicke:
The DocumentImpl constructor takes an argument called forUseInNonAWTThread, which is stored in the myAssertThreading field. While this field is checked before some assertions, it is not checked before others. 

This change checks the field before all assertIsDispatchThread calls, enabling users to use more features of DocumentImpl outside the dispatch thread (as indicated by the constructor parameter).

(It is unclear to me why the assertIsDispatchThread assertions are there at all -- the functions don't seem to require it at all. This is the least intrusive change, remove the assertions completely should maybe be considered.)


klikh:

martinwicke:
Just sent the CLA.

klikh:
Received, thanks.

donnerpeter:
Merged manually, thanks!

PR #239, opened 2015-01-29
Add method override autocompletion for Groovy Plugin

Arasthel:
This was created as a separate plugin in https://plugins.jetbrains.com/plugin/7571,
but it fits better inside the official Groovy plugin.

### DIFF
--- a/plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
+++ b/plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
@@ -20,7 +20,6 @@ import com.intellij.codeInsight.generation.OverrideImplementExploreUtil;
 import com.intellij.codeInsight.lookup.LookupElement;
 import com.intellij.codeInsight.lookup.LookupElementBuilder;
 import com.intellij.icons.AllIcons;
-import com.intellij.patterns.PlatformPatterns;
 import com.intellij.psi.*;
 import com.intellij.psi.infos.CandidateInfo;
 import com.intellij.psi.util.PsiFormatUtil;
@@ -31,6 +30,7 @@ import com.intellij.util.ProcessingContext;
 import com.intellij.util.VisibilityUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.plugins.groovy.lang.completion.handlers.GroovyMethodOverrideHandler;
+import org.jetbrains.plugins.groovy.lang.psi.api.statements.typedef.GrTypeDefinitionBody;
 
 import javax.swing.*;
 import java.util.Collection;
@@ -47,21 +47,16 @@ class GrMethodOverrideCompletionProvider extends CompletionProvider<CompletionPa
   protected void addCompletions(@NotNull CompletionParameters parameters, ProcessingContext context, @NotNull CompletionResultSet result) {
     PsiElement position = parameters.getPosition();
 
-    if(psiElement()
-            .inside(PsiClass.class)
-            .accepts(position)) {
+    PsiClass currentClass = PsiTreeUtil.getParentOfType(position, PsiClass.class);
 
-      PsiClass currentClass = PsiTreeUtil.getParentOfType(position, PsiClass.class);
-
-      if (currentClass != null) {
-        addSuperMethods(currentClass, result, false);
-        addSuperMethods(currentClass, result, true);
-      }
+    if (currentClass != null) {
+      addSuperMethods(currentClass, result, false);
+      addSuperMethods(currentClass, result, true);
     }
   }
 
   public static void register(CompletionContributor contributor) {
-    contributor.extend(CompletionType.BASIC, PlatformPatterns.psiElement(PsiElement.class), new GrMethodOverrideCompletionProvider());
+    contributor.extend(CompletionType.BASIC, psiElement().withParent(GrTypeDefinitionBody.class), new GrMethodOverrideCompletionProvider());
   }
 
   private void addSuperMethods(final PsiClass psiClass, CompletionResultSet completionResultSet, boolean implement) {
@@ -89,7 +84,7 @@ class GrMethodOverrideCompletionProvider extends CompletionProvider<CompletionPa
 
     PsiType type = substitutor.substitute(method.getReturnType());
 
-    String parentClassName = currentClass == null ? "" : ((PsiNamedElement) currentClass).getName();
+    String parentClassName = currentClass == null ? "" : currentClass.getName();
 
     String signature = modifiers + (type == null ? "" : type.getPresentableText() + " ") + method.getName();
 
--- a/plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
+++ b/plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2000-2015 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


klikh:

Arasthel:
I got accepted as a contributor this morning. Thank you.

klikh:
Agreement received. Thanks.

maxmedvedev:
Hi Jorge!
This is really cool. I'll accept the pull request after some improvements.

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
```diff
@@ -0,0 +1,106 @@
...
+ * Created by Arasthel on 29/01/15.
+ */
+
+class GrMethodOverrideCompletionProvider extends CompletionProvider<CompletionParameters> {
+
+  @Override
+  protected void addCompletions(@NotNull CompletionParameters parameters, ProcessingContext context, @NotNull CompletionResultSet result) {
+    PsiElement position = parameters.getPosition();
+
+    if(psiElement()
```
maxmedvedev: We don't need to check psiElement().inside(PsiClass.class) since we search for containing class and check if it is not null on the next line.

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
```diff
@@ -0,0 +1,106 @@
...
+
+      if (currentClass != null) {
+        addSuperMethods(currentClass, result, false);
+        addSuperMethods(currentClass, result, true);
+      }
+    }
+  }
+
+  public static void register(CompletionContributor contributor) {
+    contributor.extend(CompletionType.BASIC, PlatformPatterns.psiElement(PsiElement.class), new GrMethodOverrideCompletionProvider());
```
maxmedvedev: The pattern here should be comprehensive pretty much. We expect to see overriding completion variants in the class body only. So you should check something like psiElement().withParent(GrTypeDefinitionBody.class) and so on. You can find more examples in GroovySmartCompletionContributor

maxmedvedev: BTW it should work in the case like this:
```Groovy
class A {
    void foo() {}
}

class B extends A {
    voi<caret> // invoke completion at the <caret> position. void foo(){...} should be suggested
}```

File: plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java [hunk 3]
```diff
@@ -0,0 +1,106 @@
...
+  private LookupElementBuilder createLookUpElement(final PsiMethod method, PsiSubstitutor substitutor, PsiClass currentClass, Icon icon,
+                                                   InsertHandler<LookupElement> insertHandler) {
+    String parameters = PsiFormatUtil.formatMethod(method, substitutor, PsiFormatUtilBase.SHOW_PARAMETERS, PsiFormatUtilBase.SHOW_NAME);
+
+    String visibility = VisibilityUtil.getVisibilityModifier(method.getModifierList());
+    String modifiers = (visibility == PsiModifier.PACKAGE_LOCAL ? "" : visibility + " ");
+
+    PsiType type = substitutor.substitute(method.getReturnType());
+
+    String parentClassName = currentClass == null ? "" : ((PsiNamedElement) currentClass).getName();
```
maxmedvedev: Cast to PsiNamedElement is unnecsesary here

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
```diff
@@ -0,0 +1,106 @@
+/*
```
maxmedvedev: A test would be appreciated =)

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
Arasthel: This was done because otherwise you could get override or implement suggestions while inside a method definition, which I don't think really make sense.

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
Arasthel: Changed the `.inside(PsiClass.class)` to your suggestion. I don't think I understand what you meant on your 2nd message, writing `voi` does trigger the override autocompletion of void methods.

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
maxmedvedev: ```psiElement.inside(PsiClass.class)``` is fully equivalent to ```PsiTreeUtil.getParentOfType(position, PsiClass.class) != null```. Right now completion inside method body suggests overriding variants. 

The correct place to put the pattern is ```register``` method where currently you provide ```PlatformPatterns.psiElement(PsiElement.class)```. See a describing comment there

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
maxmedvedev: This is a screenshot with completion in java. In Groovy it should work the same way. All you need is to provide a correct pattern in contributor.extend()

https://www.dropbox.com/s/as6hfjyuw30qemz/Screen%20Shot%202015-02-03%20at%2013.41.43%20.png?dl=0

plugins/groovy/src/org/jetbrains/plugins/groovy/lang/completion/GrMethodOverrideCompletionProvider.java
Arasthel: I don't really know what to test here, since suggestion candidates are provided by `OverrideImplementExploreUtil.getMethodsToOverrideImplement(psiClass, implement)` and then filtered by IDEA. I could try testing if it should suggest some methods on different places (class definition, body definition, etc.), but to be honest I don't really know how to do that.

maxmedvedev: OK, no problem =)
Actually if you have no time or don't want to fix all that things I wrote about I can accept the pull request and fix them by myself but it will take a while. So I don't guarantee the feature to be included into the nearest release.

Arasthel: I have fixed the other stuff you mentioned, but I honestly don't know what to test here as most of the functionality is delegated to already implemented and tested methods, and I don't know how to test what is left. I can upload what I already fixed.

maxmedvedev:
Don't worry about tests. Not a big deal actually. I'll write one or two examples to guarantee the feature won't be broken in future.

maxmedvedev:
Sorry for the delay. We will accept your commit in a few day. Thanks!

Arasthel:
That's really nice to know! I have a question, though: if I wanted this to be included on the next version of Android Studio should I add another PR to the 135 branch? Thank you :+1:

maxmedvedev:
If you want we can try to cherry-pick this to 135. But are you sure Android Studio uses the 135 branch?

Arasthel:
It is, as far as I know. The about me dialog shows 135 as build number and [this answer](http://stackoverflow.com/a/26262595) in SO seems to support that.

Arasthel:
Hi, is there any way I can help to make this PR merge? It's been almost a month since the last update.

dovchinnikov:
Merged manually with updates. 
@Arasthel actually it was like year ago, sorry for no notification.

PR #240, opened 2015-02-02
Fix loading of C native library on FreeBSD with JNA < 4.1.0.

alexdupre:
For reference: https://github.com/twall/jna/issues/279

### DIFF
--- a/platform/util/src/com/intellij/execution/process/UnixProcessManager.java
+++ b/platform/util/src/com/intellij/execution/process/UnixProcessManager.java
@@ -46,7 +46,13 @@ public class UnixProcessManager {
   static {
     try {
       if (!Platform.isWindows()) {
-        C_LIB = ((CLib)Native.loadLibrary("c", CLib.class));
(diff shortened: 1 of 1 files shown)


klikh:

alexdupre:
CLA submitted.

klikh:
Received, thanks.

alexdupre:
Actually I'm thinking it could make more sense to rebuild the old jna library with this patch: https://people.freebsd.org/~jkim/jna.diff
This would fix not only this issue, but also other libraries (pty4j, jtermios, etc.) that uses jna to load the libc. What's your opinion?

trespasserw:
Sounds reasonable. Just need to check what holds us back from JNA upgrade.

trespasserw:
Status update: we're in the process of migrating to JNA 4.1 - upgrading and checking dependent libraries one-by-one.

trespasserw:
Finished the migration to JNA 4.1 (finally :/). Sorry for it to take so long.

PR #241, opened 2015-02-10
FileSystemUtil: Update JNA code to work on Linux/PPC

zeldin:
The struct stat64 offsets were incorrect for Linux/PPC32.
Add the correct offsets, and use them when OS_ARCH is "ppc".

Also, increase the "ver" parameter of __xstat64 to "1", since the
Linux/PPC64 libc rejects the value "0".

There might be more arches where the default offset table is incorrect,
but I can't test that right now.

I have tested this patch on Linux/PPC with IBM JDK 1.7.0 (both 32 and 64 bit)
and on Linux/x86 with Oracle JDK 1.7.0 (64 bit).

### DIFF
--- a/platform/util/src/com/intellij/openapi/util/io/FileSystemUtil.java
+++ b/platform/util/src/com/intellij/openapi/util/io/FileSystemUtil.java
@@ -397,6 +397,7 @@ public class FileSystemUtil {
     }
 
     private static final int[] LINUX_32 =  {16, 44, 72, 24, 28};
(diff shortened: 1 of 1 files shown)


klikh:

zeldin:
Ok, sent it now.  Thanks.

klikh:
Received. Thanks.

trespasserw:
Pushed to upstream, thanks!

PR #242, opened 2015-02-11
Add support for paramaterizing varargs in PyCharm

kroo:
Previously, `*args` and `**kwargs` were skipped when matching against parameterized types in the type checker.  This prevented pycharm from understanding type hints on functions whose return type are a function of the varardic parameters.

This can be reproduced with the following code:

``` python
def f(*args):
    """
    :type args: tuple[T]
    :rtype: T
    """
    return args[0]

expr = f(1)  # expect expr to be of type "int"
```

### DIFF
--- a/python/src/com/jetbrains/python/psi/types/PyTypeChecker.java
+++ b/python/src/com/jetbrains/python/psi/types/PyTypeChecker.java
@@ -150,7 +150,7 @@ public class PyTypeChecker {
         else if (superTupleType.isHomogeneous() && !subTupleType.isHomogeneous()) {
           final PyType expectedElementType = superTupleType.getElementType(0);
           for (int i = 0; i < subTupleType.getElementCount(); i++) {
-            if (!match(expectedElementType, subTupleType.getElementType(i), context)) {
+            if (!match(expectedElementType, subTupleType.getElementType(i), context, substitutions, recursive)) {
               return false;
             }
           }
@@ -160,7 +160,7 @@ public class PyTypeChecker {
           return false;
         }
         else {
-          return match(superTupleType.getElementType(0), subTupleType.getElementType(0), context);
+          return match(superTupleType.getElementType(0), subTupleType.getElementType(0), context, substitutions, recursive);
         }
       }
       else if (matchClasses(superClass, subClass, context)) {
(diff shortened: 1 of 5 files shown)


klikh:

kroo:
Sent!

kroo:
@klikh, is there anything I can do to help this PR along?

klikh:
I've forwarded your request to the PyCharm developer. He'll look at this one soon. Sorry for delay.

kroo:
@klikh Anything I should do to bump this again?

vlasovskikh:
@kroo Thanks for your PR! Sorry for the late reply. 

We would like the type hints notation for varargs to be compatible with PEP 0484, where you don't have to state that `*args` is a tuple of some type; you just specify the type of the elements in that tuple. For example, we would like this:

```python
def f(*args, **kwargs):
    """
    :type args: T
    :type kwargs: V
    :rtype: (T, V)
    """
    return args[0], kwargs['foo']
```

to be similar to PEP 0484 syntax:

```python
T = typing.TypeVar('T')
V = typing.TypeVar('V')

def f(*args: T, **kwargs: V):
    return args[0], kwargs['foo']
```

python/src/com/jetbrains/python/psi/types/PyTypeChecker.java
```diff
@@ -393,11 +393,26 @@ else if (type instanceof PyCallableType) {
...
       final PyType paramType = context.getType(p);
+
+      PyType argType = context.getType(entry.getKey());
+      if (p.isPositionalContainer()) {
+        final PyClassType tupleClassType = PyBuiltinCache.getInstance(p).getTupleType();
+        if (tupleClassType == null) {
+          return null;
+        }
+        final PyClass tupleClass = tupleClassType.getPyClass();
+        argType = new PyCollectionTypeImpl(tupleClass, false, argType);
```
vlasovskikh: We now have `new PyTupleType(tupleClass, elementTypes, true)` for homogeneous tuples in our type system.

kroo:
Thanks, this makes sense.  I'll give it a go.

sproshev:
Fixes related to PY-19723 are in master.

PR #244, opened 2015-02-16
fsNotifier: Fix glibc version specifier for ppc64

zeldin:
On Linux/PPC64, glibc does not contain `memcpy@GLIBC_2.0`, so changing that to `memcpy@GLIBC_2.3` which does exist.

### DIFF
--- a/native/fsNotifier/linux/inotify.c
+++ b/native/fsNotifier/linux/inotify.c
@@ -29,6 +29,8 @@
 
 #ifdef __amd64__
 __asm__(".symver memcpy,memcpy@GLIBC_2.2.5");
+#elif defined(__powerpc64__)
+__asm__(".symver memcpy,memcpy@GLIBC_2.3");
 #else
 __asm__(".symver memcpy,memcpy@GLIBC_2.0");
 #endif


trespasserw:
What about 32-bit PPC?

zeldin:
PPC32 does have `memcpy@GLIBC_2.0`, so that part is fine.

Incidentally, I just checked AArch64 and it needs `memcpy@GLIBC_2.17`, but maybe that can wait...  :smile:   32-bit ARM needs `memcpy@GLIBC_2.4` (note that e.g. Android doesn't use glibc though, so no versioned variant of the symbol exists).  Those were the Linux arches I have up at the moment, can check MIPS later...

zeldin:
MIPS also supports `memcpy@GLIBC_2.0`, and so does SPARC.

zeldin:
I kind of wonder if it's not possible to use the unversioned symbol on all arches except amd64 though, since that's the only one where I saw more than one version actually being available...

trespasserw:
The symbol is there to make binaries we provide compatible with older versions of glibc. And, right, since we only provide binaries for x86 and amd64, it makes sense to specify the version only for these two.

trespasserw:
Pushed upstream. Thanks!

PR #245, opened 2015-02-18
Allow Git 'Compare with branch' popup to filter based on the branch name

Amwam:
When selecting compare with branch, a list of branches pops up, this list is not searchable/filterable.
This commit allows the list to be searched just by typing.


klikh:
Contribution Agreement was received.

klikh:
Merged. Thanks!
Btw, this fixes https://youtrack.jetbrains.com/issue/IDEA-116852.

Amwam:
Ah awesome thanks! I did try having a look for an existing issue, guess I did look hard enough.

PR #246, opened 2015-02-20
Move "$@@product_uc@@_VM_OPTIONS" to the end of line.

garar:
Hi!

This change will enable JVM variable overriding. Oracle's JVM uses the last variable, which means that if we have the same variable defined twice, only the second value will be used, e.g. if -Xmx is defined twice, once in user's _VM_OPTIONS file and once in idea.vmoptions in bin/idea.vmoptions, user's -Xmx will be used by JVM.

Let me know if you need anything else.

Thanks,
Krystian Kichewko

### DIFF
--- a/bin/scripts/unix/idea.sh
+++ b/bin/scripts/unix/idea.sh
@@ -135,7 +135,7 @@ fi
 
 VM_OPTIONS=""
 VM_OPTIONS_FILES_USED=""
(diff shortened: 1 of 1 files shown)


klikh:

garar:
Hello Kirill!

I've just sent signed Contributor License Agreement to contribute@jetbrains.com. Let me know if you need anything more from me.

Thanks,
Krystian Kichewko

klikh:
Agreement received, thank.

trespasserw:
Kristian, thanks for the suggestion. Unfortunately, I can't accept your patch right away, because  launchers for Windows and OS X should be changed accordingly. Filed [an issue](https://youtrack.jetbrains.com/issue/IDEA-136975) in our tracker for that, please watch it for the progress.

PR #247, opened 2015-02-27
Fixed method clear in TestProjectBuilderLogger

erokhins:


chashnikov:
Why do you need this? In fact myLogLines isn't cleared in 'clear' method intentionally, it allows us to get full log after performing several builds. Though I agree that the name isn't good.

erokhins:
In kotlin we have a test, which rebuilds project many times(over 150) and size of this log is very large(100 mb)

chashnikov:
I've added a separate method 'clearLog' to TestProjectBuilderLogger.

PR #248, opened 2015-03-04
add getContainingIds method to StubIndexImpl

martinwicke:
This pull request concerns [IDEA-137196](https://youtrack.jetbrains.com/issue/IDEA-137196)

I'm wondering whether this is the correct approach for adding the ability to obtain the files touched by a given symbol, in order to iterate over them later, or whether the interface should look different (e.g. wrapping the Collection<Integer> inside a ValueContainer.IntIterator, taking a Processor<Integer> argument, or also restricting the set using a IdFilter). 

I'd be grateful for guidance to make this as painless to integrate as possible.


nicity:
I am concerned with copying logic of working with index just to get ids. Also having Collection<Integer> is quite wasteful in terms of memory. I've attached code to the original ticket https://youtrack.jetbrains.com/issue/IDEA-137196#comment=27-949780

martinwicke:
Yep, that is clearly better, and it will work for me.

PR #249, opened 2015-03-05
fix IDEA-132015

siosio:
Add "assigned me" checkbox.
If this check box is not selected, the all tasks are displayed.


east825:
Hi, @siosio, I'm going to make this feature (filtration of tasks by assignee) consistent across all different connectors and move corresponding checkbox to the "Open Task" popup eventually, but for the time being it's a nice little fix, so I accept your pull request. **Thank you!**

P.S. [I've changed UI in the editor a bit](2af42804760f20367d366feaf81c90b2fa08774b) so it was similar to the one for Trello: checkbox is called "Include issues not assigned to me" and has the opposite effect. Hope you don't mind.

PR #250, opened 2015-03-05
Remove file based storage from project creation options (fixes first part of IDEA-137327)

dlsniper:
Hi,

As per ticket https://youtrack.jetbrains.com/issue/IDEA-137327, this removes the file based storage from new project creations.

Thank you.


chashnikov:
Until [IDEA-80673](https://youtrack.jetbrains.com/issue/IDEA-80673) is fixed it may be convenient to use file-based storage format if you have many small projects which project files aren't shared via VCS. So don't let's remove the option for now.

PR #253, opened 2015-03-14
Added a few additional foldings for Java, some based on Groovy

paplorinc:
Every comment is appreciated, I don't expect this to be merged in as it is currently :)


paplorinc:
folding

klikh:

klikh:
CA was received. Thanks.

JB-Dmitry:
We doubt that the majority of users will like such a change, so we wouldn't like to make this functionality available in platform by default. You can still make it available as a separate plugin.

paplorinc:
If the majority won't like it, it can be disabled by default and we can get feedback on it.
Could you please reopen this?

yole:
In our experience, adding features which are disabled by default is a waste of time, because the number of people who will look in the settings for the feature, understand what it does, and decide to enable it is extremely small. Releasing the functionality as a separate plugin is a much better way to get feedback.

paplorinc:
> adding features which are disabled by default is a waste of time

Maybe, but not adding them at all would definitely be :)
There are already 10 _disabled_ foldings by default in Idea (more than _enabled_ by default).

Please tell me what changes you would like to see, to accept any part of my changes (I've separated  the features in commits, I'm sure, at least my cleanup commit doesn't affect the end users)

JB-Dmitry:
As for cleanup commit - I wouldn't change test data just for the sake of reformatting. And adding `@Language("JAVA")` makes code red in some places (due to `<caret>` markers). The rest of the changes are minor.

PR #255, opened 2015-03-18
IDEA-57472 new tool window mode using frames

krasa:
Maybe it needs a better name, but it is very useful with multiple monitors nevertheless.

https://youtrack.jetbrains.com/issue/IDEA-57472


krasa:
merged as part of b79bf470cf12d378a8d34ef9586297dc6295833c

PR #256, opened 2015-03-19
[IDEA-137300] Added ability to filter psi classes returned by JavaPsiFac...

akosyakov:
It adds a new method PsiElementFinder.getClassesFilter(GlobalSearhcScope). By default it returns null. Methods of JavaPsiFacadeImpl, which are used to look up classes, are changed to apply the filter if it is not null.


akosyakov:
Merged to master: https://github.com/JetBrains/intellij-community/commit/fce192d5f5ecd1e58a4f3f1ffed9375863995a91

PR #260, opened 2015-04-27
Add the ability to get keys and values from KeyFMap

bashor:
Changes in KeyFMap are needed for Kotlin

### DIFF
--- a/java/compiler/openapi/src/com/intellij/openapi/compiler/ExportableUserDataHolderBase.java
+++ b/java/compiler/openapi/src/com/intellij/openapi/compiler/ExportableUserDataHolderBase.java
@@ -17,37 +17,27 @@ package com.intellij.openapi.compiler;
 
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.util.UserDataHolderBase;
+import com.intellij.util.keyFMap.KeyFMap;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.Map;
 
 public class ExportableUserDataHolderBase extends UserDataHolderBase implements ExportableUserDataHolder{
-  private final Set<Key> myKeys = Collections.synchronizedSet(new HashSet<Key>());
-
   @NotNull
   public final Map<Key, Object> exportUserData() {
     final Map<Key, Object> result = new HashMap<Key, Object>();
-    synchronized (myKeys) {
-      for (Key<?> k : myKeys) {
-        final Object data = getUserData(k);
-        if (data != null) {
-          result.put(k, data);
-        }
+
+    KeyFMap map = getUserMap();
+    Key[] keys = map.getKeys();
+
+    for (Key<?> k : keys) {
+      final Object data = map.get(k);
+      if (data != null) {
+        result.put(k, data);
       }
     }
+
     return result;
   }
-
-  @Override
-  public final <T> void putUserData(@NotNull Key<T> key, @Nullable T value) {
-    if (value != null) {
-      myKeys.add(key);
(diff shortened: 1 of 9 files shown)


bashor:
@donnerpeter @chashnikov please review this.

bashor:
Note I didn't test it on CI

donnerpeter:
Why exactly do you need this in Kotlin?

bashor:
> Why exactly do you need this in Kotlin?
We need lightweight container for storing metadata in some places.
We want to use `UserDataHolder(Base)` but sometimes we need merge some containers, so we should know what keys stored inside container... 

TBH we need only `getKeys`, `getValues` added because it's already exist in most implementations and to be symmetrical :)

bashor:
Another possible way to achieve it is declare some utility functions outside, but I afraid that then we can forget to update these functions while adding new implementations.

Thoughts?

bashor:
I thought that maybe will be useful to have `containsKey` too?
What do you think?

platform/util/src/com/intellij/openapi/util/Key.java
```diff
@@ -106,6 +106,17 @@ public void set(@Nullable Map<Key, Object> holder, T value) {
     return (Key<T>)allKeys.get(index);
   }
 
+  @NotNull
+  public static Key[] getKeysByIndices(int[] indexes) {
```
chashnikov: Index of a key is an implementation detail and shouldn't be exposed to open API. It's bad that we already have public 'getKeyByIndex' method, I would prefer to make it package-local. But anyway let's not emphasize this problem by adding implementation-specific utility method to API.

chashnikov:
If you need to merge KeyFMap why not add method 'plus(KeyFMap)' to KeyFMap
instead?

bashor:
>If you need to merge KeyFMap why not add method 'plus(KeyFMap)' to KeyFMap
instead?

Good point, thanks @chashnikov!
What do You think @donnerpeter?

Anyway I think `getKeys` is useful in some cases(e.g. see second commit).

bashor:
I'm sorry, but I jumped to conclusions. We really need `getKeys`.

donnerpeter:
I still don't understand the exact problem you're trying to solve and why just plain old Map isn't enough. KeyFMap is pretty specific and has quite a bad performance when large. If you need immutability, have a look at SmartFMap (and if Kotlin still doesn't have its own immutable collections, that's a shame).

bashor:
I updated the PR

bashor:
> I still don't understand the exact problem you're trying to solve and why just plain old Map isn't enough. 

We want to store metadatas which key type is `Key`.

>KeyFMap is pretty specific and has quite a bad performance when large. 

It's ok for our case -- we don't expected large datas.

bashor:
We (sometimes) prefer to use tested for years things instead of reinvent it :)

donnerpeter:
java.util.HashMap is a very tested thing, for many more years already than KeyFMap. The latter is itself an implementation detail of UserDataHolderBase and might change in future (although not very likely).  It also lacks some functionality that might be important, e.g. equals/hashCode. HashMap is not so type-safe, agreed, but definitely more reliable and proven.

So I'd say I need more arguments that you really-really need KeyFMap and no alternatives will ever suit you :)

bashor:
I updated PR.

bashor:
Please backport changes to 141 branch

donnerpeter:
Pushed to master. Need review for 141. @chashnikov ?

chashnikov:
Looks ok now, though I would generate KEYS and VALUES arrays in KeyFMapTest instead of hardcoding.

bashor:
I created new PR #262 with changes in KeyFMapTest.

Thanks all!

PR #262, opened 2015-04-29
Generate keys and values in each test case instead of use hardcoded constants

bashor:


bashor:
@chashnikov please take a look

chashnikov:
Merged manually.

bashor:
Thanks!

PR #264, opened 2015-05-05
Backport KeyFMap changes

bashor:


bashor:
@chashnikov please take a look

chashnikov:
I'm not sure about backporting changes in interning, @donnerpeter what do you think?

nicity:
I am against, performance implications of first commit isn't yet known.

bashor:
Well, I excluded changes in `UserDataInterner`.

nicity:
What about breaking API ?
Please provide reasons why we should do the backport in 141

bashor:
>What about breaking API ?

Mainly it covers internal things and which (I think) doesn't have many usages.

I don't know your policies about changes like this. 

> Please provide reasons why we should do the backport in 141

I did these changes for Kotlin and I'd like to backport them if it possible.

nicity:
In 141 we are in end of regular updates of IntelliJ Idea 14.1.X and any noncritical thing backported needs reasonable justification

bashor:
Actually this PR didn't rename anything in current public API (`ArrayBackedFMap#getKeys` added and renamed during this PR).
But it added a method to `KeyFMap`, is it OK?

nicity:
Adding method is ok.

bashor:
So, can you merge it?

nicity:
Let's backport If we don't change / decrease existing API

bashor:
Well, could anybody merge it?

//cc @nicity @chashnikov

nicity:
I backported your changes to 141 manually, please review

bashor:
I see, thanks a lot!

PR #265, opened 2015-05-05
IDEA-132005 fix performance problem in ErrorPaneConfigurable.java

xhumanoid:
Concatenating string without StringBuilder, as result if user have 2000+ problem and total size of result html string will be more 50mb then "Project structure" panel freeze all.

I wait 30 minutes but ui keep freeze, after very simple patch all work fine


klikh:

xhumanoid:
Yes. I already sent a letter with CLA today.

klikh:
Agreement received, thank you.

bulenkov:
Thanks for the patch. Rewrote and pushed into master. Alarm should work on pooled thread.

PR #267, opened 2015-05-13
Show line numbers by default

mnemonicflow:
The EditorSettingsExternalizable class is used in the [EditorAppearanceConfigurable ](https://github.com/JetBrains/intellij-community/blob/master/platform/lang-impl/src/com/intellij/application/options/editor/EditorAppearanceConfigurable.java) class to set the selected state of the 'Show line numbers' JCheckBox by reading the value from the isLineNumbersShown() method which returns the default value of the setting.

Why show line numbers by default?
* Because I'm getting tired of enabling 'Show line numbers' every single time I install a JetBrains product on a new machine!
* Because code navigation matters! How do you read and refer to code in a file with a couple hundred/thousand lines of code without line numbers?
* Because debugging maters! How do you debug anything without using line numbers?
* **Because it's the *right* choice for this setting to be enabled by default!**
* Because it's time! You don't want people getting easy points on StackOverflow from asking questions about poor defaults, do you? https://stackoverflow.com/questions/13751/how-can-i-permanently-have-line-numbers-in-intellij  (question asked in 2008, the situation is still the same today...)

->What year is it?<-

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/editor/ex/EditorSettingsExternalizable.java
+++ b/platform/platform-impl/src/com/intellij/openapi/editor/ex/EditorSettingsExternalizable.java
@@ -50,7 +50,7 @@ public class EditorSettingsExternalizable implements PersistentStateComponent<Ed
     public boolean IS_CARET_BLINKING = true;
     public int CARET_BLINKING_PERIOD = 500;
     public boolean IS_RIGHT_MARGIN_SHOWN = true;
-    public boolean ARE_LINE_NUMBERS_SHOWN = false;
+    public boolean ARE_LINE_NUMBERS_SHOWN = true;
     public boolean IS_FOLDING_OUTLINE_SHOWN = true;
 
     public boolean SMART_HOME = true;


yole:
I'll answer all of your points:
- You can use the [Settings Repository plugin](https://plugins.jetbrains.com/plugin/7566?pr=idea) to synchronize your settings between all the machines where you install JetBrains products.
- You read and refer to code using the names of the methods, with features like Goto Declaration, File Structure and Find Usages. Once you have a tool that understands the semantics of your code, line numbers become much less important.
- You debug code by putting breakpoints, examining the values of the variables and fields, and so on. None of this requires referring to line numbers.
- Says who?
- We certainly don't mind people getting points on StackOverflow for any reason.

The year isn't 1975 any more, so it's no longer necessary to refer to line numbers when working with code, when you have tools that actually understand it on a deeper level.

mnemonicflow:
@yole

* Line numbers may not be important in languages that run on the JVM (Java, et. al.) but they are certainly important in native languages (e.g. C++, while doing GDB debugging)
* Line numbers are also helpful when reading a log full of stack traces, you fire up the IDE to check the source code at a certain line number your read in the log message only to find out you don't see line numbers by default in the IDE...
* Says who? I and *many* other developers (see StackOverflow questions and other forums) who disagree with the fact that line numbers are not shown by default.

yole:
* Yes, Settings Repository should work with Community Edition products as well.
* This is a pull request for IntelliJ IDEA, which targets primarily languages that run on the JVM. C++ with GDB debugging is supported in CLion, and it may well be appropriate to turn on line numbers by default in CLion. I have no opinion on that; this is something that you should discuss with CLion's developers.
* IntelliJ IDEA's Analyze Stacktrace feature allows you to paste in a stacktrace from your log and to get clickable links that bring you to the correct line of the code, without ever needing to see or enter line numbers manually.

si-yao:
"IntelliJ IDEA's Analyze Stacktrace feature allows you to paste in a stacktrace from your log and to get clickable links that bring you to the correct line of the code, without ever needing to see or enter line numbers manually."

Copy/paste costs time, opening "Analyze Stacktrace" costs time. More importantly, it is not impossible that co-work brings his laptop and shows the stack trace to me. Do I need to copy stack trace from his machine? I would suggest run a statistics to see how many percentage people are showing line numbers in their Intellij.

PR #268, opened 2015-05-18
Fix for a null-pointer exception while decompiling

jsight:


klikh:

jsight:
I have not and was not aware that a CLA was necessary for something like this. Perhaps this can be treated as a bug report and recoded instead?

trespasserw:
Jesse, a CLA is needed for applying the patch under your name.
Sure I can commit the fix myself if you like.

jsight:
@trespasserw - Ok, that is fine, I don't care about the attribution. :)

trespasserw:
Commited as 6fdcf291aa34b3d6fc5f3514e613d265448d03bd. Thanks!

PR #271, opened 2015-06-04
[IDEA-140807] Added an extension that allows to redirect navigation f…

akosyakov:
…rom a requested editor to another


yole:
I really don't like this API, because can break a lot of features in the product. There is a lot of code that opens a certain file in the editor and then starts modifying editor text, for example, to offer a live template of some kind. That code will randomly corrupt text if the navigation policy replaces the requested file with a different one behind the caller's back.

yole:
Closing as per discussion in https://youtrack.jetbrains.com/issue/IDEA-140807

PR #273, opened 2015-06-08
Fixed issues related to generic type refs from cls

szarnekow:


szarnekow:
see https://youtrack.jetbrains.com/issue/IDEA-141230

trespasserw:
Pushed upstream (with minor reformatting and omitting unneeded test cases). Thanks!

PR #275, opened 2015-06-15
added and marked as deprecated removed constructor ArchiveHandler.Ent…

MichaelNedzelsky:
…ryInfo

This is import in order to support Kotlin plugin


MichaelNedzelsky:
@cdracm
Please take a look

PR #277, opened 2015-06-23
Fix annotation for `Iterable.iterator`

dzharkov:
Add FQ name to return type.
It's necessary because `findExternalAnnotation` uses fully-qualified names when matching annotations.


klikh:

akozlova:
Fixed in 318f98327974fa0d1cf2eca9525f80d2f932d138.

PR #279, opened 2015-06-25
Usage View: Remove usages (like in Eclipse)

allotria:
Added context menu to remove usages from the usage tree instead of just excluding them.

This is a possible solution to https://youtrack.jetbrains.com/issue/IDEA-22996.


klikh:
Contribution Agreement has been received.

cdracm:
integrated, thanks for the patch.

PR #281, opened 2015-07-02
RemoveUsageAction: Now selects a next item from the usage tree…

allotria:
RemoveUsageAction: Now selects a next item from the usage tree after removing one or several items. This makes the function easier to use with the keyboard.

This is an addition to the now merged pull request [#279](https://github.com/JetBrains/intellij-community/pull/279) fixing [IDEA-22996](https://youtrack.jetbrains.com/issue/IDEA-22996).
It's basically the same code as I use in my [Remove Usage Plugin](https://plugins.jetbrains.com/plugin/7833).


klikh:
The pull request has been merged manually by cdr: 6f7ff1646571048ae9e894b536949e19d190d7cc Thanks!

PR #283, opened 2015-07-03
Support navigation to the original element for generated source code

szarnekow:
see discussion in https://youtrack.jetbrains.com/issue/IDEA-140807


platform/analysis-api/src/com/intellij/openapi/roots/GeneratedSourcesFilterEx.java
```diff
@@ -0,0 +1,115 @@
...
+public interface GeneratedSourcesFilterEx {
+
+  /**
+   * Returns all elements that have been processed by a code generator to derive the given element.
+   *
+   * @param element the generated element
+   * @return a list of original elements. An empty result indicates that the element is not considered to be generated by the filter.
+   */
+  @NotNull
+  List<? extends PsiElement> getOriginalElements(@NotNull PsiElement element);
```
cdracm: why not add these (empty implemented) methods to the com.intellij.openapi.roots.GeneratedSourcesFilter instead of creating additional interface?

szarnekow: I wasn't sure if that was an option in terms of backwards compatibility. If you prefer that variant, I'd be happy to adjust the PR.

cdracm: as far as i understand nothing breaks if new empty method added to the abstract class. Please do.

szarnekow:
I adjusted the PR according to your suggestion

szarnekow:
I added an API to optionally ignore generated source information when opening an editor

PR #285, opened 2015-07-08
Structure popup should check if filters are reverted

bjansen:
When I configure a structure filter as a FileStructureFilter so that it's shown in the structure popup as well, matching nodes are removed from the tree although the checkbox is not checked. That's because the algorithm assumes the filters are reverted by default (which is the case in the few implementations I've seen so far).
In my case, the filter is not reverted, so nodes shouldn't be removed from the tree *unless* the checkbox is checked. This PR will apply filters if they are checked and not reverted, or if they are not checked and reverted.


bjansen:
BTW I just signed the CLA and sent it by email.

bjansen:
I created https://youtrack.jetbrains.com/issue/IDEA-142411 for this pull request.

bulenkov:
Thank you for pointing the problem. Pushed to master.

PR #286, opened 2015-07-08
Allow using more than one color in structure view nodes

bjansen:
This patch allows using more than one color in `ItemPresentation`s used in the structure view, for example to match the syntax highlighting in the editor.

### DIFF
--- a/platform/structure-view-api/src/com/intellij/ide/projectView/PresentationData.java
+++ b/platform/structure-view-api/src/com/intellij/ide/projectView/PresentationData.java
@@ -15,6 +15,7 @@
  */
 package com.intellij.ide.projectView;
 
+import com.intellij.ide.util.treeView.MulticoloredItemPresentation;
 import com.intellij.ide.util.treeView.PresentableNodeDescriptor;
 import com.intellij.navigation.ColoredItemPresentation;
 import com.intellij.navigation.ItemPresentation;
@@ -30,6 +31,7 @@ import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
 import java.awt.*;
+import java.util.Collection;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


bjansen:
Are you guys are interested in this PR? If so, I can update it.

SergeyMalenkov:
This idea is good, but...

ItemPresentation is a simple representation, which includes primary text, secondary text and icon only. All colors are retrieved automatically according to selected color scheme and component state.

ColoredItemPresentation adds an ability to set a text color, but it uses a special key to retrive it from selected color scheme.

You suggest to return a ColoredFragment from the Tree API, which allows people to set colors directly. So many people will not take care of supporting different color schemes.

I think that interface should return something like List<Pair<String,TextAttributesKey>>. Also, the documentation should describe what should we do with methods of the super interface. Should we ignore all the methods from ItemPresentation? Should we extend ColoredItemPresentation and use our list as addtitional text fragments to the primary colored fragment?

Also we need to test this functionality in different use cases.

SergeyMalenkov:
Sorry, the following part of the comment was considered as HTML and gone:

I think that interface should return something like List of Pair of String,TextAttributesKey.

bjansen:
Thanks for your feedback. Using `TextAttributesKey` seems like a good idea, I'll try to improve the PR this week. 

I think `getLocationString()` and `getIcon(boolean)` are still useful and should not be ignored, only `getPresentableText()` will not be used because the results of `getColoredFragments()` will be used instead. Since this method is already `@Nullable`, I suppose I can override it in `MulticoloredItemPresentation` to add a Javadoc that clarifies this.

`ColoredItemPresentation` can still be used independently of `MulticoloredItemPresentation` to add an easier way to force a single color.

ignatov:
@bjansen, please rebase the PR, cc @SergeyMalenkov.

bjansen:
Actually I don't have this branch anymore, I will open a separate pull request with my updated changes.

bjansen:
@SergeyMalenkov what about something like this?

```java
/**
 * Extends {@link ItemPresentation} to allow adding text attributes to multiple regions of the text.
 */
public interface MulticoloredItemPresentation extends ItemPresentation {

  /**
   * Returns the text attributes for rendering the item text. Each part of {@link #getPresentableText()}
   * that needs to be displayed differently should have an entry in the returned map.
   *
   * @return a map where the key indicates a region of the presentable text, and the value
   *         the text attributes to apply to that region, or an empty map if default text
   *         attributes should be used everywhere
   */
  Map<TextRange, TextAttributesKey> getTextAttributesKeys();
}
```

`getPresentableText()` would still be used to return the text, and the extra methods returns which parts of the text should be presented differently, and how to present them.

PR #288, opened 2015-07-13
Include more details for FS case setting mismatch

mdaniel:
Per the discussion on YouTrack [IDEA-135546](https://youtrack.jetbrains.com/issue/IDEA-135546), the message was updated to detail which system property one should adjust. For those unfamiliar with the situation, the Confluence page will still explain it. For those who see this a lot, the system property will remind them which property needs adjustment.

If one wishes to put back the contraction (doesn't), then please escape the apostrophe ('') since MessageFormat uses apostrophe as its escape character. This bug is even present in the screenshot on the Confluence page.


trespasserw:
Accepted, thanks!

PR #289, opened 2015-07-14
IDEA-96179 Replace with a camel case string and Preserve Case

nsandalov:
Replace with a camel case string and Preserve Case on converts the string to lower case


nicity:
without test we can not merge the change

PR #296, opened 2015-08-29
Fixed performance of DataNode#hashCode

fkorotkov:
It appeared on a project with 2000+ modules hashCode for DataNodes was super slow. Also added some additional debug logging for better performance analysis.

### DIFF
--- a/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/DataNode.java
+++ b/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/DataNode.java
@@ -279,13 +279,9 @@ public class DataNode<T> implements Serializable, UserDataHolderEx {
 
   @Override
   public int hashCode() {
+    // We can't use myChildren.hashCode() because it iterates whole subtree.
     int result = myKey.hashCode();
     result = 31 * result + getData().hashCode();
-    // We can't use myChildren.hashCode() because it's exponential.
-    // It iterates over children and calls hashCode over and over again.
-    for (DataNode<?> child : myChildren) {
-      result = 31 * result + child.getData().hashCode();
-    }
     return result;
   }
 
--- a/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/DataNode.java
+++ b/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/DataNode.java
@@ -279,9 +279,13 @@ public class DataNode<T> implements Serializable, UserDataHolderEx {
 
   @Override
   public int hashCode() {
-    int result = myChildren.hashCode();
-    result = 31 * result + myKey.hashCode();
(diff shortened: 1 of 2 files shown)


platform/external-system-api/src/com/intellij/openapi/externalSystem/model/DataNode.java
```diff
@@ -279,9 +279,13 @@ private void readObject(ObjectInputStream in)
 
   @Override
   public int hashCode() {
-    int result = myChildren.hashCode();
-    result = 31 * result + myKey.hashCode();
+    int result = myKey.hashCode();
     result = 31 * result + getData().hashCode();
+    // We can't use myChildren.hashCode() because it's exponential.
```
chashnikov: What do you mean by exponential? Do you mean that hashCode may be called many times for single object? For me it looks like the time is linear in size of DataNode (including all its children recursively), though it may be too much indeed.

fkorotkov:
Exponential of a height of a tree. Or just O(amount of nodes in the sub-tree) as you mentioned. It is too much. I saw IJ was recalculating hashCode over and over again in ToolWindowModuleService#processData and just use ExternalSystemApiUtil#groupBy a lot.

Contract of hashCode method is that it should return the same value for equal objects but if it returns the same value it doesn't mean that the objects are equal. Also I don't think it will increase amount of collisions significantly but it affected performance in my testing much.

platform/external-system-api/src/com/intellij/openapi/externalSystem/model/DataNode.java
fkorotkov: Exponential of a height of a tree. Or just O(amount of nodes in the sub-tree) as you mentioned. It is too much. I saw IJ was recalculating hashCode over and over again in ToolWindowModuleService#processData and just use ExternalSystemApiUtil#groupBy a lot.

Contract of hashCode method is that it should return the same value for equal objects but if it returns the same value it doesn't mean that the objects are equal. Also I don't think it will increase amount of collisions significantly but it affected performance in my testing much.

Can change the comment tho.

chashnikov: Yes, the comment is confusing indeed. I think height of a tree is usually 2-3 here, but it may have many nodes.

Why not exclude myChildren from hashcode computation at all? I think nodes with children are project and modules nodes, so it would be enough to compute hashCode of getData(), I doubt that there will be many collisions of ModuleData hashcodes. And this way hashCode for DataNode<ProjectData> won't iterate over all 2000+ module nodes for your project. What do you think?

fkorotkov: Sounds good to me as well.

chashnikov:

PR #297, opened 2015-08-30
Refactored external system services to reduce amount or write actions

fkorotkov:
Description:

It appeared that most of the default services were executing a lot of write actions to commit models. Also services were committing models in write actions even if there were no changes to these models.

This refactoring tries to collect all the modifiable models and perform all commits in just one write action. Also before a commit all models are checked for changes. It affects performance of an incremental refresh drastically for huge projects.

Testing:

All 5 tests for external build system are passing. :-) Also manually tested on a project with 2227 modules. Incremental refresh on dependency/library changes was super fast!

It would be very nice if you guys can run the change on CI.

### DIFF
--- a/platform/external-system-api/src/com/intellij/openapi/externalSystem/service/project/PlatformFacade.java
+++ b/platform/external-system-api/src/com/intellij/openapi/externalSystem/service/project/PlatformFacade.java
@@ -1,6 +1,5 @@
 package com.intellij.openapi.externalSystem.service.project;
 
-import com.intellij.openapi.externalSystem.model.ProjectSystemId;
 import com.intellij.openapi.externalSystem.model.project.*;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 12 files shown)


chashnikov:

I also did some fixes myself. Firstly, I changed order of statements in 'LibraryTableBase.LibraryModel#commit', otherwise 'isChanged' will return 'false' during 'commit' so no events will be fired. Secondly, I introduced method 'ExceptionUtil.rethrowAllAsUnchecked' and used it instead of 'rethrowUnchecked' to ensure that if the code will be changed in the future and start producing checked exceptions they won't be silently ignored.

fkorotkov:
Awesome! Спасибо, Коля!

dancing

fkorotkov:
Does 142.4675.3 include the fix?

fkorotkov:
Nope. It's included in 142.4712.

fkorotkov:
Built IntelliJ from sources. Much faster! But I think there is a room for even more improvements. Will do another round of profiling after my upcoming PTO.

PR #298, opened 2015-08-30
[groovy] format comments properly in Spock 'where' tables

paplorinc:
Fixes https://youtrack.jetbrains.com/issue/IDEA-137145: 'Formatting is broken if there is comments inside Spock Data Table'


plugins/groovy/src/org/jetbrains/plugins/groovy/formatter/blocks/GroovyBlockGenerator.java
```diff
@@ -438,8 +438,7 @@ else if (psi instanceof GrVariableDeclaration) {
 
   private boolean shouldSkip(boolean classLevel, PsiElement psi) {
     if (psi instanceof PsiComment) {
-      PsiElement prev = psi.getPrevSibling();
-      if (prev != null && prev.getNode().getElementType() != GroovyTokenTypes.mNLS || classLevel && !fieldGroupEnded(psi)) {
+      if ((psi.getPrevSibling() != null) || (classLevel && !fieldGroupEnded(psi))) {
```
paplorinc: should skip even if the previous element is a new line

PR #299, opened 2015-08-30
align chained methods with comments properly

paplorinc:
Fixes https://youtrack.jetbrains.com/issue/IDEA-122802: 'Comments between chained calls aren't indented properly' and https://youtrack.jetbrains.com/issue/IDEA-136980: 'Comments at end of the line during chained calls break allignment'

### DIFF
--- a/java/java-tests/testSrc/com/intellij/psi/formatter/java/JavaFormatterAlignmentTest.java
+++ b/java/java-tests/testSrc/com/intellij/psi/formatter/java/JavaFormatterAlignmentTest.java
@@ -51,16 +51,17 @@ public class JavaFormatterAlignmentTest extends AbstractJavaFormatterTest {
 
   public void testChainedMethodWithComments() throws Exception {
     getSettings().ALIGN_MULTILINE_CHAINED_METHODS = true;
-    doMethodTest("AA.bb()\n" +
-                 ".cc() // comment after line\n" +
-                 ".dd()\n " +
-                 "        /* block comment on empty line */\n" +
-                 ".ee();",
-                 "AA.bb()\n" +
-                 "  .cc() // comment after line\n" +
-                 "  .dd()\n " +
-                 "/* block comment on empty line */\n" +
-                 "  .ee();");
+    doMethodTest("AAAAA.b()\n" +
+                 ".c() // comment after line\n" +
+                 ".d()\n" +
+                 "        /* unaligned block comment on separate line */\n" +
+                 ".e();",
+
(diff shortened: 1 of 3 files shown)


java/java-tests/testSrc/com/intellij/psi/formatter/java/JavaFormatterAlignmentTest.java
```diff
@@ -49,6 +49,20 @@ public void testMethodAndChainedField() throws Exception {
     );
   }
 
+  public void testChainedMethodWithComments() throws Exception {
+    getSettings().ALIGN_MULTILINE_CHAINED_METHODS = true;
+    doMethodTest("AA.bb()\n" +
+                 ".cc() // comment after line\n" +
+                 ".dd()\n " +
+                 "        /* block comment on empty line */\n" +
```
lepenkinya: Why you've added this block comment line to test? 
Is it supposed for this comment to be indented with method chain start block?

paplorinc: I agree that it's confusing. It's because the test wraps this in a class and method before formatting, indents all lines, except the comments. So I had to indent that one manually, to start at the same place as the method body would. Will provide you screenshots (and probably push another change that explains this in the comment itself)

paplorinc: formatterbeforeafter

paplorinc: fixed in next commit :)

lepenkinya: Think I get it :)

Block comments were also affecting alignment, so you've added it here. Am I right? If so, I think we could create two tests instead: one for one-line comment and another for block comment. What do you think about it?

paplorinc: We could, but I think conceptually - and even for the fix - they ended up being exactly the same, don't see the added value in separating them.

Let me know if you want me to add it anyway.

(FYI this is related to https://github.com/JetBrains/intellij-community/pull/298)

lepenkinya:
I've merged your request, with minor change: I've separated test, and in the case with block comment used doTextTest method instead to avoid confusion with shifted comment.

PR #300, opened 2015-08-31
Support modifying directory rendering.

tomlu-google:
* Module names shown/not shown
* Source roots shown/not shown

This is used by TreeStructureProviders to suppress the rendering of the
above.

Change-Id: I9c6ccf82b6c85043adebc56c588e714e3e54203f


tomlu-google:
I've updated the commit with a much more modest change. We are now replacing each PsiDirectoryNode with a replacement instance in our TreeStructureProvider, obviating the need for instance variables.

chashnikov:

PR #301, opened 2015-09-02
Canonicalize paths before sending to file watchers

salguarnieri:
Canonicalize paths before sending them to the pluggable
FileWatchers and restore paths to their original potentially
non-canonical form when they are returned from the file
watchers.

This fixes an issue where the fs notifier binary behaved incorrectly when it was presented with symlinks. The fs notifier has some logic to avoid watching certain mount points, but if a user symlinks into a mount, the fs notifier's logic will be bypassed and the mount points will be watched.

This change is done in the FileWatcher rather than the native fs notifier because:
* The filewatcher supports pluggable filewatchers now, so it makes sense to give all of the pluggable file watchers canonical paths rather than making them all canonicalize the paths.
* The implementation in C for the native fs notifier would be similar to what is needed in Java for the FileWatcher and the java implementation is likely cleaner. 

Change-Id: I61fd3ac4e3bfbb8b06c2d92190e6db36e7ff00c1


trespasserw:
`FileWatcher.getManualWatchRoots()` should return unmapped paths. Or its usages should be adopted - which is less desirable.

trespasserw:
Setting roots and processing notifications under the same lock may cause a deadlock between JVM and native watcher processes on Linux - we had such issues in the past.

trespasserw:
There is a problem with `CanonicalFileMapper.getMapping(Collection<String>)`: because on Windows and OS X the native process effectively reports events from entire disc, the method may be quite hot.

trespasserw:
I'd suggest skipping already-canonical paths - ideally, the mapper should introduce little to no overhead when no symlinks are involved.

salguarnieri:
I uploaded a POC for an implementation that uses a tree so almost nothing is done in the case of already canonical paths. There is a problem when there is a symlink watchroot and a canonical watch root that leads to the same file. Currently, it is solved by an extra loop iteration at the end of the method, but I'm not a big fan of it. I'll think more about it tonight. I wanted to get your input on the new implementation though with regards to performance when there are no symlinks present.

salguarnieri:
With regards to windows and OS X, are you saying that when *any* file is changed on disk, the filewatcher gets notified?

salguarnieri:
@trespasserw I was wondering if there were any comments about this approach.

salguarnieri:
@trespasserw Hold off on a review, i have a few things I want to clean up first

salguarnieri:
Alright, it should be good to look at again

trespasserw:
Sorry, had a hard week, will take a look at it tomorrow.

Regarding changes on OS X / Windows - yep, that is true. The current approach is to filter out unwatched changes on the Java side (see NativeFileWatcherImpl.checkWatchable()).

trespasserw:
It just came to me that there already exist a very similar kind of path mapping in NativeFileWatcherImpl.checkWatchable() used to handle drive substitutions on Windows. So, unless having a single mapper across all watchers is strictly needed, I feel it would be much simpler to reuse that code - populating `myMapping` inside `setWatchRoots()` probably is enough. What do you think?

salguarnieri:
Just so I understand, we would have to add the canonicalization code that populates myMapping, because it isn't there right now, correct? I didn't notice this myMapping getting populated during my quick test, but I definitely might have missed something.

trespasserw:
Exactly. Quick check shows it should work - well, at least it passes your tests :)

You never saw `myMapping` populated because it requires Windows host - check out FileWatcherTest.testSubst().

tomlu-google:
Having a single mapper across all watchers isn't strictly needed, but each watcher would have to implement its own. Each file watcher has to know whether a given path belongs to it, and I think it pretty much has to canonicalize to do so. It seems to make sense to have a single one.

Is it a good idea to instead remove myMapping from the native file watcher, if it serves the same purpose?

trespasserw:
You see, it is kind of implementation detail. NativeFileWatcherImpl uses OS APIs and gets events by canonical paths hence the mapping is needed - but it's not necessarily true for other watchers.

tomlu-google:
No, it's not necessarily true for other watchers, but we have three and it's turned out true for all of them. I suspect it would be true for almost any file watcher, but sure, it doesn't *have* to be.

Computation-wise, it seems like we either do it once, or 1 + N times, where N >= 0. If we had to do it ourselves, we'd just stick this watcher implementation into each of our file watchers (which we certainly can - it just seems a nicer interface to the file watchers to know that it's already canonicalized).

trespasserw:
OK, three is convincing enough.
To avoid multiple computations, both mapping and filtering need to be moved out of NativeFileWatcherImpl into FileWatcher then.

salguarnieri:
Ok, that sounds good.

I have a question about the mapping. It would be clean if we could remove that code entirely. Since paths are canonicalized before being sent to the native file watcher, is it still necessary to support the REMAP command in the native file watcher?

trespasserw:
AFAIK File.getCanonicalPath() does not resolve substitutions on Windows, so REMAP remains relevant.

salguarnieri:
Ok, good to know. I'll consolidate the logic and push up a new commit.

salguarnieri:
I pushed a version of the code where the filewatcher does the filtering. The commit comment sums up the changes. One thing I'm a bit unhappy with is how to deal with the complex filtering logic revolving around the "exact" param from the old checkWatchable method. I mostly ignored it and allowed recursive roots to also match their parent.

trespasserw:
That logic is needed to handle notifications on OS X correctly. With the patch applied some tests fail.

salguarnieri:
Thanks for that info.

I locally rebased on the jetbrains master and ran ant test but didn't see any obvious failures. Can you let me know what tests failed and how to run them?

thanks!

trespasserw:
Some of FileWatcherTest tests - 5 or 6; testDirectoryOverlapping certainly was among them. I can post the whole log if you need it.

You need to run the test on OS X to make it fail - because it verifies not only a Java part but native implementations as well, and they all behave differently. Notably, the OS X one produces only directory-level notifications (DIRTY and RECDIRTY) - that's why the "exact" parameter is important.

salguarnieri:
Thanks, I ran those tests in the IDE on linux but saw no errors (for obvious reasons) and ran ant test on my mac, but it looks like that must not run those tests. I'll run the tests from the IDE and that should give me the errors to investigate. thanks!

salguarnieri:
Ok, the tests pass on linux and os x now. I don't have a windows box, so I can't speak to those tests passing or not.

trespasserw:
Checking. Meanwhile, could you please take a look at FileWatcherTest.testIncorrectPath()? It now takes 2 minutes on Linux, seemingly because of the mapper freaking out over the non-existing path.

salguarnieri:
Thanks.

I was able to repro the slowness on OS X, so I'm looking into it there (and will verify any solution on linux). It looks like the problem is actually with the test structure. The test waits (in effect) until some event is sent back from the file watcher impl. Previously, the unwatchable event would cause the waiter to be notified and the test would proceed. With the patch, the invalid path isn't being sent to the file watcher impl, so there will never be a response to notify the waiter. This causes the test to wait until the timeout.

This leads to two questions:
* It looks like the test was relying on a side effect of the native file watcher impl to wake up myWaiter. Do you have a suggestion for a better way to do this? I could add a notifyOnAnyEvent to some spots in native file watcher where we exit early now, but I think that might just hide the root problem.

* Should invalid paths be allowed to be sent to the file watcher impls? My gut says no; that we shouldn't watch directories that don't yet exist. It seems like that could lead to unexpected behavior. But I'm not an expect in all the ways people are using the file watcher.

trespasserw:
People use it in unexpectedly creative ways :) The test was added to make sure native watchers can tolerate invalid paths - there was a bug in Linux implementation which caused it to terminate prematurely.

I rather agree that non-existing paths shouldn't make it to the helper process. The test should be rewritten then to reflect the fact.

There are also some strange failures on Windows, investigating.

salguarnieri:
Rebased and updated file watcher tests for non-existing paths. Commit comment has a description of what I did.

salguarnieri:
Hey, I was wondering if you had updates about the windows test. I can probably get access to a windows VM, but if you happen to already have the failures for me that would be great.

I was also wondering if you had any other comments about the commit.

trespasserw:
Sorry, forgot to post an update. Most of those problems were caused by me running tests as a regular user (creating symlinks on Windows requires elevated privileges). The rest are more complicated and partially caused by substitutions handling in IdeaWin*.dll helper, still checking out.

trespasserw:
I just pushed an alternative implementation based on your patches. It offloads all the mapping and filtering from NativeFileWatcherImpl to FileWatcher, with actual mapping code extracted to a separate class. Breaks the API a bit but makes pluggable watchers simpler to implement.
Please take a look - does it suits your needs?

trespasserw:
Regarding non-existing paths - I would like to preserve the current behavior until I figure out how to deal with substitutions and other kinds of NTFS reparse points.

salguarnieri:
Thanks! It looks good.

My only concern is that the mapping lookup code is linear in the number of paths being watched. What do you think of a new patch on top of your change that replaces the mapping lookup with something that uses a tree or Map so the lookup is on the order of number of path components rather than number of watched roots?

trespasserw:
Sure, CanonicalPathMap was extracted with this in mind.

PR #302, opened 2015-09-18
Fix file length check

salguarnieri:
IJ uses the timestamp and file length to determine if two files
differ. In almost all cases, the timestamp will be sufficient
but there are some important use cases were timestamps are
manipulated inside of archives to ensure the archives only
differ when their actual contents differ.

The operation of reading the currently stored file size could
cause a read of file size from disk. That makes the up-to-date
check broken since it will be comparing the new file size to the
file size just read from disk, which should be the same as the
new file size.

Change-Id: I8d5d0e160c8d7fd4f1847af187e17226aebe7453

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/persistent/PersistentFS.java
+++ b/platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/persistent/PersistentFS.java
@@ -64,6 +64,12 @@ public abstract class PersistentFS extends ManagingFS {
 
   public abstract String getName(int id);
 
+  /**
+   * Get the last length recorded for this file. If the file has changed since the length was recorded, this will return the stale length
+   * value.
+   */
+  public abstract long getLastRecordedLength(@NotNull VirtualFile file);
+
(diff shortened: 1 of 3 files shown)


salguarnieri:
Can you give your experience regarding why hashes/checksums aren't used for this comparison? I suspect that it is because of the performance impact, but it seems much safer than relying on file size.

trespasserw:
Pushed upstream, thanks!

Because of performance, sure - the drop would be dramatic, especially for network mounts where native FS notifications are unavailable.

salguarnieri:
Thanks! Can you point me to the upstream commit?

This is a long shot, but do you know of any extension points or other workarounds, that would allow us to provide a stricter check. We will know that all files we want to check will be on a certain network mount and they will be inside a jar.

salguarnieri:
i see the commit now

trespasserw:
The repository at Github is updated once in an hour AFAIK.

There is no such extension point at the moment; need to be added to the very same place in RefreshWorker where lengths and timestamps are compared.

PR #304, opened 2015-09-28
Always check PATH in Git detector

fkorotkov:
With the change the detector works fine with my custom git location.

### DIFF
--- a/plugins/git4idea/src/git4idea/config/GitExecutableDetector.java
+++ b/plugins/git4idea/src/git4idea/config/GitExecutableDetector.java
@@ -17,13 +17,14 @@ package git4idea.config;
 
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configurations.GeneralCommandLine;
+import com.intellij.execution.configurations.PathEnvironmentVariableUtil;
 import com.intellij.execution.process.CapturingProcessHandler;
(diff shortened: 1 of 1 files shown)


fkorotkov:
FYI @klikh. Would be nice to have it cherry-picked in 15.

klikh:
LGTM. Have you checked `GitExecutableDetectorWindowsTest`?

fkorotkov:
I assumed it's only for windows and I have only Mac.

klikh:
I've checked it myself, it fails because the code is referring to the PathEnvironmentVariableUtil which is no more overridden in tests. I'll prepare the fix and merge the updated change soon.

fkorotkov:
Thanks a ton!

klikh:
Just found another problem: there are two valid executable formats for Git on Windows: git.cmd and git.exe. In this PR only git.exe is considered. Moreover, it is not enough to, for instance, check the PATH for git.exe and then check it for git.cmd, since if both are available in the PATH, the first one should be preferred.

Currently working on it. Luckily, I wrote this test a while ago, otherwise wouldn't capture this.

iasemenov:
I've ported this PR to current master.
See commit 06e4dec6000d3fb7193caa6c597e7107005b0fef
Thank you, @fkorotkov

PR #306, opened 2015-10-01
Improve cross-compilation

Kendos-Kenlen:
Improve make.sh to correctly manage system architecture and cross-compilation.


trespasserw:
Actually, `-D_DEFAULT_SOURCE` should be enough - could you please check on your system?

Kendos-Kenlen:
No problem. I just wasn't sure about it. Search I did wasn't clear about the -D_XOPEN_SOURCE=500.

trespasserw:
Pushed the change upstream. Thanks!

PR #308, opened 2015-10-19
Fixed `BoolUtils#isNegation` to check for *logical* (instead of *bitwise*) negation

paplorinc:
Since there were multiple `BoolUtils` present, the actual fix was their merger, eliminating the faulty method as a side-effect.


plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/codeInspection/utils/BoolUtils.java
```diff
@@ -1,85 +0,0 @@
...
-
-public class BoolUtils {
-
-  public static boolean isNegation(@NotNull GrExpression exp) {
-    if (!(exp instanceof GrUnaryExpression)) {
-      return false;
-    }
-    final GrUnaryExpression prefixExp = (GrUnaryExpression) exp;
-    final IElementType sign = prefixExp.getOperationTokenType();
-    return GroovyTokenTypes.mBNOT.equals(sign);
```
paplorinc: should have been `GroovyTokenTypes.mLNOT`

PR #309, opened 2015-10-20
Put path mapping storage behind an interface

salguarnieri:
Use a map to look up path mapping information. This makes the
lookup on the order of the length of the path (in components)
rather than on the order of number of paths that need mapping.

Added support (and a test) for the situation when a mapped path (symlink)
is then remapped by the native file watcher.


salguarnieri:
I left the original linear list impl in the code so the interface can be examined independently of the map impl. Before the cl is merged, I'd remove the list impl entirely.

trespasserw:
Looks good. But CanonicalPathMapTest does not work :)

salguarnieri:
It passes for me on mac and linux so I'm guessing that it must be failing on my favorite os (windows). I should have a windows laptop waiting for me at the office to debug the test. Is there any abstraction layer in IntelliJ to simulate the windows file system behavior?

trespasserw:
Right, Windows. Um, didn't think it may be platform-dependent, will check it later today.

trespasserw:
There are no such layer - too many small differences. We resorted to having CI agents with all 3 OSes.

trespasserw:
Nothing critical, just file separators. You can go on, I'll make sure the test works before merging.

salguarnieri:
Ah yes, sorry about that.

salguarnieri:
No rush, but I was wondering if you had an estimate on when the merge would happen.

trespasserw:
Oops, sorry :)
I was waiting for you to do final touches (removing a list-based implementation etc).
I'll clean it up and merge it later today.

salguarnieri:
Sorry, I should have clarified sooner. I removed the list based impl and I think I made the test safe for windows.

trespasserw:
Simplified it a bit and pushed upstream. Thanks!

PR #310, opened 2015-10-22
Allow disabling HiDPI scaling with -Dhidpi=false

sabi0:
This is a workaround for the issue described in [IDEA-143854](https://youtrack.jetbrains.com/issue/IDEA-143854).

It allows having larger font size in IDEA ("Appearance > Override default fonts") while keeping the icons at their original size (sharp!).

It does so by short-circuiting calls to `JBUI.setScaleFactor()` in case `-Dhidpi=false` has been defined.
This brings `setScaleFactor()` in sync with `calculateScaleFactor()` behavior.
(And also prevents the splash-screen from lubberly changing the scale midway due to the call from `LafManagerImpl.patchLafFonts()`)

I've been using IDEA with this patch since the "HiDPI build". And haven't noticed any undesired effects.


klikh:

sabi0:
Sent the signed CLA.

klikh:
The CLA was received. Thanks.

PR #313, opened 2015-10-25
Disabled `PointlessArithmetic` inspection when the expression is non-resolvable

paplorinc:
Spock's [Mock interactions](https://spockframework.github.io/spock/docs/1.0/interaction_based_testing.html#_mocking) shouldn't be flagged as `PointlessArithmeticExpression`, e.g.
`0 * mockAmazonCloudWatchClient.putMetricAlarm(_)`


paplorinc:
@dovchinnikov, @maxmedvedev

paplorinc:
@dovchinnikov, @maxmedvedev, @ignatov ?

paplorinc:
... too bad

leonard84:
@dovchinnikov any reason this was not merged? This is a really annoying false positive.

dovchinnikov:
@leonard84 most probably it got lost in notifications, sorry.

I've checked the request, while it's okay as _disabling inspection for non-resolvable_, I think the problem is that such expression are unresolvable in the first place, and to make them resolvable we need bigger changes in the API. I was working on such API recently, but it got delayed due to other unfinished tasks.

paplorinc:
@dovchinnikov, personally I think a temporary working solution that partially fixes things is better than none.
I've committed a few more similar PRs, please check them too

PR #316, opened 2015-10-26
Use commit template if available

rohitkumbhar:
Added a configuration option to use a VCS commit template if its available. This fixes https://youtrack.jetbrains.com/issue/IDEA-66355 for Git

Testing: Local Git repo shows the commit template

### DIFF
--- a/platform/vcs-api/src/com/intellij/openapi/vcs/checkin/CheckinEnvironment.java
+++ b/platform/vcs-api/src/com/intellij/openapi/vcs/checkin/CheckinEnvironment.java
@@ -46,6 +46,9 @@ public interface CheckinEnvironment extends VcsProviderMarker {
   String getDefaultMessageFor(FilePath[] filesToCheckin);
 
   @Nullable
+  Set<String> getCommitTemplatesFor(FilePath[] filesToCheckin);
+
+  @Nullable
   @NonNls
   String getHelpId();
 
--- a/plugins/git4idea/src/git4idea/checkin/GitCheckinEnvironment.java
+++ b/plugins/git4idea/src/git4idea/checkin/GitCheckinEnvironment.java
@@ -108,6 +108,34 @@ public class GitCheckinEnvironment implements CheckinEnvironment {
   }
 
   @Nullable
+  public Set<String> getCommitTemplatesFor(FilePath[] filesToCheckin) {
+    LinkedHashSet<String> templates = ContainerUtil.newLinkedHashSet();
+    for (VirtualFile root : GitUtil.gitRoots(Arrays.asList(filesToCheckin))) {
+      try {
+        final String commitTemplateFile = GitConfigUtil.getValue(myProject, root, "commit.template");
+        if (commitTemplateFile != null && commitTemplateFile.length() > 0) {
+          final String encoding = GitConfigUtil.getCommitEncoding(myProject, root);
+          final String templateContent = FileUtil.loadFile(new File(commitTemplateFile), encoding);
+          if (templateContent.length() > 0) {
+            templates.add(templateContent);
+          }
+        }
+      }
+      catch (VcsException e) {
+        if (log.isDebugEnabled()) {
+          log.debug("VcsException: Unable to load commit template for Git repo: " + root.getPath(), e);
+        }
+      }
+      catch (IOException e) {
+        if (log.isDebugEnabled()) {
+          log.debug("IOException: Unable to load commit template for Git repo: " + root.getPath(), e);
+        }
+      }
+    }
+    return templates;
+  }
+
+  @Nullable
   public String getDefaultMessageFor(FilePath[] filesToCheckin) {
     LinkedHashSet<String> messages = ContainerUtil.newLinkedHashSet();
     for (VirtualFile root : GitUtil.gitRoots(Arrays.asList(filesToCheckin))) {
--- a/platform/vcs-api/src/com/intellij/openapi/vcs/VcsConfiguration.java
+++ b/platform/vcs-api/src/com/intellij/openapi/vcs/VcsConfiguration.java
@@ -124,6 +124,7 @@ public final class VcsConfiguration implements PersistentStateComponent<VcsConfi
 
   public boolean FORCE_NON_EMPTY_COMMENT = false;
   public boolean CLEAR_INITIAL_COMMIT_MESSAGE = false;
+  public boolean USE_COMMIT_TEMPLATE_IF_AVAILABLE = false;
 
   @Property(surroundWithTag = false)
   @AbstractCollection(elementTag = "MESSAGE", elementValueAttribute = "value", surroundWithTag = false)
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/CommitChangeListDialog.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/CommitChangeListDialog.java
@@ -383,6 +383,18 @@ public class CommitChangeListDialog extends DialogWrapper implements CheckinProj
           }
           hasVcsOptions = true;
         }
+
+        if (VcsConfiguration.getInstance(project).USE_COMMIT_TEMPLATE_IF_AVAILABLE) {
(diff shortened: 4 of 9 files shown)


klikh:

rohitkumbhar:
Sent the signed agreement to contribute@jetbrains.com

klikh:
Received, thanks.

klikh:
First of all, thank you for the contribution.

There is a problem with this pull request though. You're requesting commit templates when constructing the commit dialog, i.e. on the EDT. But such methods as `GitConfigUtil.getValue`, which forks a git process, and `FileUtil.loadFile` which reads from disk (less painful but still) can potentially take long time to finish. And they will freeze the UI during that time.

A simple solution could be to perform these actions under a modal progress (which in IntelliJ IDEA doesn't appear if the action takes less than 0.3s), but then there is a risk of blinking modal progresses on slow machines. That would annoy users a lot, especially if they don't have any commit template.

One more idea comes to my mind cache the commit message template in a way similar as we already do for the list of branches (see GitRepositoryImpl.getBranches()): listen to git config and template file changes, load the file in the background when such change happens, use the cached value instantly when commit dialog is shown. However, this needs to be thinked over carefully. 

Your suggestions are welcome.

rohitkumbhar:
Thanks! I had not thought of these conditions. The caching solution seems ok. However, the file doesn't have to change for the template to change. User can update the git global option to point to a different file. Users will have to restart IntelliJ IDEA at that point to pick up the new template.

klikh:
Actually, we can capture even this fact as well. I've made a [request about tracking global git config changes](https://youtrack.jetbrains.com/issue/IDEA-131645) some time ago.

rohitkumbhar:
You are right. Modifying the interface is probably a bad idea for other plugins. I can add an extension point and get the (cached) commit template from the extension in the CommitChangeListDialog class, same place where I call the interface method currently. Would that work?

klikh:
Not exactly. There is some complex logic around choosing initial commit message. In particular, a valuable feature is preserving commit message on cancel (user decides to commit, edits the message, cancels the dialog to make some additional changes, and calls the commit dialog again). See `setComment()` method of `CommitChangeListDialog`. Unfortunately, the whole logic is not entirely clear there due other aspects (such as taking commit message from the changelist, etc.), and we need to decide *when* commit message should be substituted and with *which* value.

There is also a property `VcsConfiguration.getInstance(getProject()).CLEAR_INITIAL_COMMIT_MESSAGE` which corresponds to a user preference (`IDEA Settings | Version Control | Confirmation | Clear initial commit message`). Currently it works incorrectly, because it clears the commit message [even after commit dialog is cancelled](https://youtrack.jetbrains.com/issue/IDEA-120230), but - apart of this bug - it looks like this is the right place to insert loading commit message from the VCS template (with fixing that bug of course). I even think that we may use the same UI property, just rename it (i.e. "clear or take from the VCS template if one exists").

That said, let me take a pause to check the logic of commit message default values, I would also welcome if you investigate this matter as well. Note that there are already many other issues with this logic which you may find amongst the 'related issues' list of [IDEA-66355](https://youtrack.jetbrains.com/issue/IDEA-66355).

neronmoon:
+1

e-cloud:
Any progress?

LeonidMarushevskyi:
+1

plugins/git4idea/src/git4idea/checkin/GitCheckinEnvironment.java
```diff
@@ -107,6 +107,34 @@ public RefreshableOnComponent createAdditionalOptionsPanel(CheckinProjectPanel p
...
+        if (commitTemplateFile != null && commitTemplateFile.length() > 0) {
+          final String encoding = GitConfigUtil.getCommitEncoding(myProject, root);
+          final String templateContent = FileUtil.loadFile(new File(commitTemplateFile), encoding);
+          if (templateContent.length() > 0) {
+            templates.add(templateContent);
+          }
+        }
+      }
+      catch (VcsException e) {
+        if (log.isDebugEnabled()) {
```
GaspardPO: Is the check "isDebugEnabled" already tested in the log.debug() method ?

It seems redundant.

GaspardPO: but it allows not to create the message, and not to call the root.getPath() method.
so a small optimisation.

platform/vcs-api/src/com/intellij/openapi/vcs/checkin/CheckinEnvironment.java
```diff
@@ -45,6 +45,9 @@
   @Nullable
   String getDefaultMessageFor(FilePath[] filesToCheckin);
 
+  @Nullable
+  Set<String> getCommitTemplatesFor(FilePath[] filesToCheckin);
```
GaspardPO: Do we want null objects or Empty Sets ?

harogaston:
+1

pchel-:
One more point here. There could be commented lines (lines starting with `#`) in the commit message template, which will be removed from commit message if using git executable. But currently if such commented lines are present in the "Commit message" field of the commit dialog they are not removed and just become part of the commit message.

So we need to deal with this somehow. Probably not load commit template to the "Commit message" field, but just show it somewhere to be used as a reference. Or add logic to highlight such lines in "Commit message" field and remove them on commit. Etc.

What do you think?

klikh:
>  Lines starting with #, if present in the template, would need to be loaded into the message field and then ignored. Same behaviour as if commiting through git using any text editor.

That's what we can't do. The user interface in the IDE is not the same as a text editor invoked from the command line. Users as used to the fact that everything they enter into the commit message field becomes the commit message. They would be very surprised if quite common texts like "Fixed #23" become ignored and stripped out from the commit message.

harogaston:
> > Lines starting with #, if present in the template, would need to be loaded into the message field and then ignored. Same behaviour as if commiting through git using any text editor.
> 
> That's what we can't do. The user interface in the IDE is not the same as a text editor invoked from the command line. Users as used to the fact that everything they enter into the commit message field becomes the commit message. They would be very surprised if quite common text constructions like "Fixed #23" become ignored and stripped out from the commit message.

Please, don't be so naive. "fixed #23" would never be left out of the commit message, not even in plain git. Only lines starting with #. Users are used to use git, and that is the default behaviour.

klikh:
> "fixed #23" would never be left out of the commit message, not even in plain git. Only lines starting with #.
That's much better, but still will introduce an issue (in rare cases, but still). We have to find a solution anyway.

Possibly, such comment lines could be rendered differently to clearly indicate that they are comments. They should be uneditable, and new lines starting with "#" should not be treated as comments. That could work.

pchel-:
Probably we could try rendering such template comments using `InlayModel` in the commit message editor. Or utilize `Document.createGuardedBlock()`. Need to create prototype to test it.

pchel-:
> Hi guys! Does anyone working on this right now ? I would like to continue the work on this if it's not taken.

Hi! It won't be part of the upcoming 2019.1 release. But we plan to add commit message template support this year (2019.2 or 2019.3).

Pozo:
> > Hi guys! Does anyone working on this right now ? I would like to continue the work on this if it's not taken.
> 
> Hi! It won't be part of the upcoming 2019.1 release. But we plan to add commit message template support this year (2019.2 or 2019.3).

Cool! In this case I'm gonna play around with it :)

harogaston:
Does this has a YouTrack issue we can follow? I would like to also request support for Pull Request message template.

Pozo:
> Does this has a YouTrack issue we can follow? I would like to also request support for Pull Request message template.

This one https://youtrack.jetbrains.com/issue/IDEA-66355

naXa777:
@Pozo are you working on this? what's your estimate?

anstarovoyt:
Thank you for creating this pull request and for your contribution to the IntelliJ platform. We would like to apologize for not acting on it sooner. 

The original issue IDEA-66355 was fixed a couple of weeks ago with a bit different approach in https://github.com/JetBrains/intellij-community/commit/9eb80b7f01465d0b1f681879a7ca0c122cca36e5 so I am closing the PR.

PR #318, opened 2015-10-30
Fixing import statement for ipython 4.0.0

pylover:
Due to the warning from `ipython.utils.traitlets.py`:

```
warn("IPython.utils.traitlets has moved to a top-level traitlets package.")
```


Elizaveta239:
Hi! Sorry for the late response, your fix was included into PyCharm.

PR #324, opened 2015-11-12
Make NativeFileWatcherImpl more generic

salguarnieri:
NativeFileWatcherImpl is nearly generic. It currently hard codes
the property name to look up to find the path to fsnotifier and
the property name to look up to see if the file watcher should
be disabled.

This pushes those values into a provider that subclasses can
provide.


salguarnieri:
We'd like to make use of a customized fs notifier and if we could supply native file watcher impl with a different exec path, we could just extend the class, provide the path, and be done.

I'm open to any solution that would allow us to do that. A similar solution would be to make these getter methods part of the class that subclasses could directly override.

trespasserw:
Pushed upstream. Thanks!

PR #325, opened 2015-11-19
Add support for sending a module name to pydevd

alexsydell:
Previously, the args for a subprocess spawned under debug with a module specified via `-m` would get rewritten incorrectly, resulting in the `-m` flag being split from the module name. This resulted in running a command similar to `/usr/bin/python -m /path/to/pydevd.py --someargs --file [module]`, which doesn't work. This removes the `-m` from that command and adds a `--command` flag passed to pydevd, which results in correctly running as a module instead of a file.

### DIFF
--- a/python/helpers/pydev/pydev_monkey.py
+++ b/python/helpers/pydev/pydev_monkey.py
@@ -51,8 +51,13 @@ def patch_args(args):
             return args
 
         i = 1
+        is_module = False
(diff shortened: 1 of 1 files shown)


alexsydell:
Closing because this no longer works in PyCharm 5.0+ due to a change in pydevd.py.

PR #326, opened 2015-11-23
Allow plugins to provide language-specific implementation of "Introduce Functional Parameter"

asedunov:


asedunov:
Thanks. Could you also look at https://github.com/JetBrains/intellij-community/pull/327?

PR #327, opened 2015-11-23
Do not process non-Java methods in JavaChangeSignatureUsageProcessor

asedunov:
It doesn't process non-Java methods (e.g. Groovy, Scala or Kotlin ones) correctly, so it should let language-specific usage processor do the job


asedunov:
Thank you. Is it possible to have #326/#327 included to the next EAP?

akozlova:
It would be included in 15.1 EAP, yes. New features/api changes are
normally not included in minor releases, so 15.0.2 won't have them or you
may contact Nik and make believe him that these changes are essential for
kotlin,

chashnikov:
Usually it's not a problem to include an API change to a minor release branch provided that it doesn't break API compatibility. However if you plan to backport a change to a release branch it would be better to keep the change is small as possible, in particular, don't convert java code to kotlin just because you can.

akozlova:
I've performed refactoring myself and didn't convert to kotlin. That's
looks safe but I still don't think that the change can't wait for 15.1 EAP

PR #329, opened 2015-12-02
Do allocation-less clear() operations for empty structures.

Steelskin:
In SLRUMap, HashMap and LinkedHashMap, the clear() operation performs
allocations when these structures are already empty. These are called on
worker threads regularly, resulting in a lot of allocations being made
for what is essentially a no-op.


donnerpeter:
Thanks, merged manually

PR #330, opened 2015-12-03
IDEA-91198 Maximize views in floating mode

clumsy:
by making floating mode window a full-fledged 
frame thus making it possible to expand the window 
by double-click and use multiple desktops feature 
on Mac which is especially useful with split views


klikh:

Vassiliy-Kudryashov:
1. com.apple.eawt.FullScreenUtilities is absolutely unknown class under Windows ans Unix.
2. There is 'windowed' mode with required functionality.

PR #331, opened 2015-12-04
Optimize setting watch roots.

tomlu-google:
The file system operations in local file system and canonical path map
can be slow when done in serial for a large number of roots. Run
these in parallel on an executor.

On our benchmarks with about 8000 roots this saves ~700ms, or about 70%,
when setting watch roots.

Change-Id: I6bd8e4fb52b9d1d43714e936f4d3d8a04049d2c1

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/CanonicalPathMap.java
+++ b/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/CanonicalPathMap.java
@@ -20,14 +20,19 @@ import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.util.io.FileSystemUtil;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.util.concurrency.BoundedTaskExecutor;
 import com.intellij.util.containers.ContainerUtil;
 import com.intellij.util.containers.MultiMap;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.ide.PooledThreadExecutor;
 
 import java.io.File;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 
 import static com.intellij.openapi.util.Pair.pair;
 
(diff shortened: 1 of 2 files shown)


tomlu-google:
Let me know if there's a better way to execute a bunch of tasks in parallel than what I added in ParallelUtil. Application#executeOnPooledThread isn't happy to accept 1000 tasks at once without starting to dump threads.

platform/core-api/src/org/jetbrains/concurrency/ParallelUtil.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+import java.util.concurrent.*;
+
+/**
+ * Utilities for executing small jobs in parallel.
+ */
+public class ParallelUtil {
+  private static final Logger LOG = Logger.getInstance(ParallelUtil.class);
+  private static final int ourReasonableThreadPoolSize = Registry.intValue("core.pooled.threads");
+  private static final ThreadPoolExecutor ourExecutorService =
```
nicity: Please check BoundedTaskExecutor, it can work above pooled threads of Application

platform/core-api/core-api.iml
```diff
@@ -22,5 +22,6 @@
     </orderEntry>
     <orderEntry type="library" name="asm" level="project" />
     <orderEntry type="library" name="CGLIB" level="project" />
+    <orderEntry type="library" name="Guava" level="project" />
```
nicity: Could you please avoid introducing new dependencies unless it is absolutely necessary

tomlu-google: I certainly can! But is Guava a heavyweight dependency? I find it's one of those libraries I use everywhere. Then again I work at Google so maybe that's not surprising ;)

Anyway, without Guava I can't use ListenableFutures, which are an arguably much more ergonomic way to use Futures. Eg. Futures.allAsList(futures). But yeah I could avoid it at the cost of reimplementing the bits that I need. Up to you!

nicity: Yes, we have Guava in Intellij but this module is quite low level one and AFAIK used in several projects so we need to check all of them to have Guava.

tomlu-google:
I updated the pull request to use BoundedTaskExecutor and removed the guava dependency. Benchmarks look the same.

nicity:
Thanks! I added couple of comments inline

05 Дек 2015 г. 20:43 пользователь "tomlu-google" <notifications@github.com>
написал:
>
> I updated the pull request to use BoundedTaskExecutor and removed the
guava dependency. Benchmarks look the same.

tomlu-google:
I changed the containers to ContainerUtil and iterate futures in reverse order.

tomlu-google:
Have you had a chance to look at this? Muchos gracias!

trespasserw:
Pushed upstream. Thanks!

PR #333, opened 2015-12-08
Scan files on more threads.

tomlu-google:
On a benchmark project the scanning phase goes from 123 seconds -> 34
seconds with this change.

Change-Id: Ib9daccd1f0188adbb4a9c56e6bf69c2c857c5ebf

### DIFF
--- a/platform/lang-impl/src/com/intellij/openapi/roots/impl/PushedFilePropertiesUpdaterImpl.java
+++ b/platform/lang-impl/src/com/intellij/openapi/roots/impl/PushedFilePropertiesUpdaterImpl.java
@@ -19,6 +19,7 @@
  */
 package com.intellij.openapi.roots.impl;
 
+import com.google.common.collect.Lists;
 import com.intellij.ProjectTopics;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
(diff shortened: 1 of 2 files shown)


platform/lang-impl/src/com/intellij/openapi/roots/impl/PushedFilePropertiesUpdaterImpl.java
```diff
@@ -303,46 +304,50 @@ public boolean processFile(final VirtualFile fileOrDir) {
...
-            @Override
-            public void run() {
-              Runnable runnable;
-              while ((runnable = tasksQueue.poll()) != null) runnable.run();
-            }
-          }, ProgressWrapper.wrap(progress));
-        }
-      });
+    List<Future<?>> results = Lists.newArrayList();
+    int numThreads = Math.min(CacheUpdateRunner.indexingThreadCount() - 1, tasks.size() - 1);
```
nicity: Some comments (for consideration):
- for 2 cpu system (my MacBook Air or our testing servers) there is no concurrency right now
- if we open several projects at once (e.g. 3) we'll create too many workers (it still can happen with previous 2x concurrent code).
- I do not observe significant speed up on my Windows machine with the change. There is one more thing to consider for this place: if we open completely new project (not existing one) the code will do real indexing of files and such indexing is 2x concurrent as well. Adding extra threads for this phase will not increase speed right now at all.  

So I'd suggest to add BoundedExecutor (over PooledThreads in Application) that will remove problems with several projects.

nicity:
Hi,
Thanks for you request!
Some comments were added inline.

nicity:
Thanks! Merged manually.

tomlu-google:
I think we're seeing this drastic difference in performance because we have 12-core machines and it's traversing a networked (but quite fast) file system for a few thousand output jars. We are also seeing similar performance gains in indexing by increasing the index thread count via a system property, so for us it's very nice to have the same characteristics here.

Anyway, thanks so much!

PR #336, opened 2015-12-29
Notify frameworks about absence of library

goodwinnk:
It's easy to skip configuring library during project creation but it can be difficult to configure libraries afterwards.
 It can cause problems for Kotlin (KT-10488), Scala, Groovy

 This callback can be used to give an additional warning similar to "No SDK Specified" (title.no.jdk.specified)


chashnikov:
There is no sense to add additional API for this. FrameworkSupportInModuleConfigurable already has isOnlyLibraryAdded method. If it returns 'true' it makes no sense to select a framework without choosing a proper library. I've pushed a commit which will cause IDEA to always show error message in such cases.

chashnikov:
See https://github.com/JetBrains/intellij-community/commit/aa46c7ac23876b2421d2c2a771de6efddf346b8c

PR #337, opened 2015-12-29
IDEA-143271 - Correctly stub classes containing a dollar in their name.

bjansen:
This PR adds a fix for https://youtrack.jetbrains.com/issue/IDEA-143271.

Classes containing a dollar in their name, but which are not nested classes generated by javac, are correctly stubbed. There is one remaining problem though, and I can't find a good solution to fix it. The class is correctly indexed as `Foo$Bar`, but the `Type`'s canonical text is still `Foo.Bar` because `StubBuildingVisitor` is making an incorrect assumption:

```java
  private static String getTypeText(@NotNull Type type) {
    final String raw = type.getClassName();
    // As the '$' char is a valid java identifier and is actively used by byte code generators, the problem is
    // which occurrences of this char should be replaced and which should not.
    // Heuristic: replace only those $ occurrences that are surrounded non-"$" chars
    //   (most likely generated by javac to separate inner or anonymous class name)
    //   Leading and trailing $ chars should be left unchanged.
    return raw.indexOf('$') >= 0 ? REGEX_PATTERN.matcher(raw).replaceAll("\\.") : raw;
  }
```

### DIFF
--- a/java/java-psi-impl/src/com/intellij/psi/ClassFileViewProvider.java
+++ b/java/java-psi-impl/src/com/intellij/psi/ClassFileViewProvider.java
@@ -25,6 +25,7 @@ import com.intellij.psi.impl.PsiManagerImpl;
 import com.intellij.psi.impl.compiled.ClsFileImpl;
 import com.intellij.psi.impl.file.PsiBinaryFileImpl;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * @author max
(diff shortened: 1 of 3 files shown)


trespasserw:
Unfortunately, we cannot accept this.
I tried this approach before, and it causes a great deal more problem than it solves.
I wish JVMS has never allowed '$' in identifiers :)

bjansen:
Do you plan on fixing the problem in any other way?

trespasserw:
Sure, but it may take some time. Looking now on your sample project.

nicity:
IMO ClsFileImpl.buildFileStub can detect inner class state correctly from content (after calling ClassFileViewProvider.isInnerClass(file))

bjansen:
I'm currently testing the following fix in ClassFileViewProvider:

```java
  protected PsiFile createFile(@NotNull final Project project, @NotNull final VirtualFile vFile, @NotNull final FileType fileType) {
    FileIndexFacade fileIndex = ServiceManager.getService(project, FileIndexFacade.class);
    if (!fileIndex.isInLibraryClasses(vFile) && fileIndex.isInSource(vFile)) {
      return new PsiBinaryFileImpl((PsiManagerImpl)getManager(), this);
    }

    ClsFileImpl clsFile = new ClsFileImpl(this);

    for (PsiClass cls : clsFile.getClasses()) {
      if (cls.getContainingClass() != null) {
        return null; // this is an inner/anonymous class
      }
    }

    return clsFile;
  }
```

i.e. I always instantiate a `ClsFileImpl` to read the bytecode, but I only return it if it doesn't represent an inner class. Then the fix mentioned by @nicity should work for the other usage of `isInnerClass`.

bjansen:
@trespasserw @nicity do you think this fix is better? https://github.com/bjansen/intellij-community/commit/5a97c8026957c1d01da813158b38ea2ea227d179

trespasserw:
@bjansen less effective, and the file without a stub doesn't make much sense anyway. @nicity is right, this should be fixed in `ClassFileViewProvider.isInnerClass()`.

bjansen:
But it's not possible to determine whether it's an inner class without actually reading the class content...

> and the file without a stub doesn't make much sense anyway

Not sure I understand what you mean, I'm not creating files without stubs?

trespasserw:
> But it's not possible to determine whether it's an inner class without actually reading the class content.

That's true. I'm afraid content reading is unavoidable here.

> I'm not creating files without stubs?

You do this by always creating a new ClsFileImpl instance. Invoking `getClasses()` on it will request for a stub for this file - which is absent because 'ClasFileImpl.buildFileStub()` checks, too, for whether a class is inner.

PR #339, opened 2015-12-30
Preprocess options to extract --pythonpath and --settings

hcastilho:
Allows using the --pythonpath and --settings options when running django tests on pycharm

### DIFF
--- a/python/helpers/pycharm/django_test_manage.py
+++ b/python/helpers/pycharm/django_test_manage.py
@@ -4,6 +4,7 @@ import os
 import sys
 
 from django.core.management import ManagementUtility
+from django.core.management.base import CommandError, CommandParser, handle_default_options
 
 from pycharm_run_utils import import_system_module
 from teamcity import teamcity_presence_env_var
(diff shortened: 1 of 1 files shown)


ignatov:
@hcastilho, please rebase the PR, cc @throwable-one, @IlyaKazakevich.

ignatov:
@hcastilho, please rebase the PR, cc @throwable-one, @IlyaKazakevich.

hcastilho:
Done

hcastilho:
Honestly I didn't even remember this PR.

Having slightly different settings for tests is something that I always do. Also I keep settings separate form the code and outside of site-packages. Probably I could have  worked around by setting `PYTHONPATH` and `DJANGO_SETTINGS_MODULE` in the tests configuration. But I have always called the django cli with these options explicitly, so I was surprised when they did not work (I avoid using env variables).

Other than someone else once in a while being also surprised I can see no other argument for including this PR. Which is apparently of very limited use.

PR #342, opened 2016-01-07
LightVirtualFile type fix

hurricup:
LigthVirtualFile contains incorrect file type when created from text but has originalFile


klikh:

hurricup:
Done

klikh:
We didn't receive it yet. Could you please re-send?

hurricup:
Sent again

klikh:
Now I see it. Thanks!

denofevil:
@hurricup, this fix looks ok for me, but probably it would be better to override getFileType in LightVirtualFile to delegate to original file if it's present.
@nicity, what do you think?

hurricup:
That was my first intention, but, my mentor been afraid that such solution may affect platforms or other plugins code. So used the most safe way to my knowlege.

donnerpeter:
Added delegation of getFileType to the original file

PR #344, opened 2016-01-08
Allow the WolfTheProblemSolver to disable the highlighting pass.

jwren:


alexander-doroshko:
@jwren Disabling this highlighting pass breaks functionality for all languages except Dart (where you suggest to call `setEnableHightlightingPass(false)` in a separate pull request).
Imagine that you have Java interface `Foo` with one method and implementing class `FooImpl` (in other file) that doesn't implement this method. Once you open `FooImpl` you get red wave in Project View for it.
Now open `Foo` and remove the method. Current behavior: red wave for `FooImpl` disappears. With your pull request it stays until you open `FooImpl` in editor again.

jwren:
Right.  What if their was one solver per module instead of per project.  It would be a small amount of additional memory, but I think it would solve this issue since there is one language per module.

Unless you say otherwise, I'll take a look at this change, (and make sure to only format the edited parts of the file.)

alexander-doroshko:
> since there is one language per module

That's not true. Even in a simple Dart project you additionally have YAML language. And you may have as many as IDE supports, all in a single module.

PR #346, opened 2016-01-11
Expose UpdateChecker.NOTIFICATIONS as public field for java usages

ilya-g:


klikh:
We've ported the change to 143. Thanks.

PR #355, opened 2016-02-15
Debugger: do not insert cast to runtime type for non java code fragments

NataliaUkhorskaya:


gorrus:
This may affect groovy, jsp, etc...

NataliaUkhorskaya:
@gorrus I've just force pushed the pull-request. Now cast should be inserted for groovy, jps and others, but not for Kotlin (and maybe some others where java-like cast won't be parsed)

gorrus:
expression is cached, so if you switch the language it won't add casts as expected.

anstarovoyt:
Thank you for creating this pull request and for your contribution to the IntelliJ platform. We would like to apologize for not acting on it sooner.

The original problem was resolved other way, so I close the PR.

PR #358, opened 2016-02-19
Show commits in git log history (IDEA-134161)

klassm:
This is a bit of a hack, as I did not want to midfy VcsFileRevision, which is shared for multiple VCS implementations. That's why I just added those git notes to the commit message. 
The null check in GitLogParserTest is to make sure, that the test conforms to the acutal behavior. If not adding the check, the test would get "null" as String, which is wrong. When executing IDEA, GitLogParser returns null for unset git notes.

If the pull request is ok, a backport to IDEA 15 would be great.

Referencing issue: https://youtrack.jetbrains.com/issue/IDEA-134161

Just as a side note: I did already submit the developer agreement.

BTW: How do you handle those code style issues and warnings within your code? In my pull requests, I did not want to resolve anything but the issue itself. However, there are loads of lines that do not conform to your codestyle and huge amounts of warnings, i.e. for unnecessary generics. Is it ok to resolve those in pull requests?

### DIFF
--- a/plugins/git4idea/src/git4idea/history/GitHistoryUtils.java
+++ b/plugins/git4idea/src/git4idea/history/GitHistoryUtils.java
@@ -826,7 +826,7 @@ public class GitHistoryUtils {
     GitSimpleHandler h = new GitSimpleHandler(project, root, GitCommand.LOG);
     GitLogParser.NameStatus status = withChanges ? GitLogParser.NameStatus.STATUS : GitLogParser.NameStatus.NONE;
     GitLogParser.GitLogOption[] options = { HASH, COMMIT_TIME, AUTHOR_NAME, AUTHOR_TIME, AUTHOR_EMAIL, COMMITTER_NAME, COMMITTER_EMAIL,
-      PARENTS, SUBJECT, BODY, RAW_BODY };
+      PARENTS, SUBJECT, BODY, RAW_BODY, NOTES };
     if (withRefs) {
       options = ArrayUtil.append(options, REF_NAMES);
     }
--- a/plugins/git4idea/src/git4idea/history/GitLogParser.java
+++ b/plugins/git4idea/src/git4idea/history/GitLogParser.java
@@ -15,6 +15,7 @@
  */
 package git4idea.history;
 
+import com.google.common.base.Strings;
 import com.intellij.openapi.project.Project;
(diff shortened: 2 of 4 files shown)


klikh:
Thanks for you contribution! However, I don't think we can accept it, because it is really some sort of a hack. I don't think we want to modify the commit message at least for 2 reasons:
* it can be used from somewhere else, and clients can be surprised if they receive some additional unexpected content in the commit message.
* it is better to deal with structured content instead of a string carrying different sorts of information somehow separated. For instance, if we decide to render commit message & notes differently.

The correct fix would be something like the following:
* add some sort of extension point or additional interface method to log providers, which would provide additional information to the log.
* read this EP or call this method from the log table and render this additional content.
* it should be decided, how to render this content and whether we want to customize this rendering process (at once, or keep the possibility to customize this in future).

You can try to follow this way, or you can wait for us to implement this. We'll try to raise the priority of this task.

Note: we are using `VcsLogProvider` interface for the repository log, and we plan to switch to this provider for the file history (which currently uses `HistoryProvider`) in the nearest releases, so it is OK to patch just this one.

Btw, you've got a typo in the commit message of this pull request: "commits" -> "notes".

As per your question about code style changes, I think it is better to avoid such changes even within a separate commit or pull request. The reason is simple: it is easier to make such changes (since many of them are semi- or fully-automated by IDEA), than to review them :)

klassm:
For sure, structured data would be much better. This would also allow a checkbox or something to enable or disable the commit notes.

However, I am still thinking about the implementation of the extension point. The implementing method would have to ask `git log` another time to get that additional data, as there is no such thing as a cache. This somehow feels wrong ... and it is hard to understand for people reading the implementation afterwards.

klikh:
Good point. 
Actually, we plan to have some hash for commit details which can include the notes as well, so I don't really think that additional command will hurt much. However, of course it is better to avoid it when possible.

Maybe this additional data should reside within the `VcsFullCommitDetails` object provided by `VcsLogProvider#readFullDetails`. Unfortunately, type safe implementation would probably force us to generify the `VcsLogProvider` itself.

klassm:
Hm this seems to be a bigger change. I suppose it is better if you discuss the change with your colleagues and implement it later on the way you want it.

PR #360, opened 2016-02-24
Fix for Go to Class window loses focus during typing.

mrawl:
The reason this bug occurs is that when the goto by name dialog dropdown resizes while the mouse is over it. If the window manager uses sloppy focus or focus-follows-mouse, then it will send a focus event to the underlying IDE window since the dialog is not modal. The IDE treats this like any other click in the underlying pane and typically the focus now goes to an underlying editor which is corrupted by further keystrokes, meanwhile, the use thinks she is still typing into the dialog box. 

This could be fixed by:
- modal dialogs (this is really how it should be fixed, but that is a behavioural change)
- implementing a dialog that doesn't have windows shrinking and losing focus
- detecting that the mouse hasn't moved and retaining the focus

This change uses the third method. It retains focus in the goto dialog if the mouse hasn't moved between a dropdown list resize and the focus loss event.

Please note that the check for the mouse not moved could have false positives and could be tightened up by adding a fast timer, or checking that the focus event.getOpposite() instanceof IdeFrame, however in all my testing I didn't find any case where it behaved unexpectedly, that includes testing it when typing into the dialog, testing when alt-tabbing away from the IDE, switching workspaces, etc.

Tested on Linux/Ubuntu with Cinnamon window manager with sloppy focus, focus-follows-mouse, and click to focus. 

Bug: https://youtrack.jetbrains.com/issue/IDEA-112015

### DIFF
--- a/platform/lang-impl/src/com/intellij/ide/util/gotoByName/ChooseByNameBase.java
+++ b/platform/lang-impl/src/com/intellij/ide/util/gotoByName/ChooseByNameBase.java
@@ -56,10 +56,7 @@ import com.intellij.openapi.ui.popup.*;
 import com.intellij.openapi.util.*;
 import com.intellij.openapi.util.registry.Registry;
 import com.intellij.openapi.util.text.StringUtil;
-import com.intellij.openapi.wm.IdeFocusManager;
-import com.intellij.openapi.wm.ToolWindow;
-import com.intellij.openapi.wm.ToolWindowManager;
-import com.intellij.openapi.wm.WindowManager;
+import com.intellij.openapi.wm.*;
 import com.intellij.openapi.wm.ex.WindowManagerEx;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.codeStyle.MinusculeMatcher;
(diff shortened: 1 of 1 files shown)


klikh:

mrawl:
Done.

On Wed, Feb 24, 2016 at 5:12 AM, Kirill Likhodedov <notifications@github.com

klikh:
Received. Thanks.

bulenkov:
Hello, mrawl. Thank you for the pull request. At the moment we're working on a more general fix. However, I've slightly modified and put your fix under a registry key. How to enable the fix? 
Go to "Search Everywhere" or "Go to Action" and type "Registry". Find appropriate action and press Enter. Find "focus.follows.mouse.workarounds" key and check it. Press "Close".

cuongthamhl:
You just saved my day!

stuaxo:
Just tried this and it didn't help on the one I find annoying every day ... if you move the mouse outside the "Find in files" window, it closes as soon as you mouse over another bit of the IDE, this should really only happen when you click, any idea what it is that closes that window?

ristomatti:
@stuaxo That is truly annoying. It doesn't seem to occur with any other dialog. Maybe submit an issue to https://youtrack.jetbrains.com/. This thread is closed and 6 years old, I doubt anyone who could fix the issue will be reading this. :)

PR #361, opened 2016-03-02
New GIT option to disable fast-forward when merging into master from a log window

OndrejSpanel:
This is a PR for git4idea plugin.

Merging with a context menu from a log window is very convenient, however it is not possible to specify Merge options when doing so.

One merge option I am missing very much is "No fast forward".

This PR adds an option to automatically set "No fast forward" for any non-trivial merges into selected branches (typically to "master").

### DIFF
--- a/plugins/git4idea/src/git4idea/GitTaskHandler.java
+++ b/plugins/git4idea/src/git4idea/GitTaskHandler.java
@@ -61,7 +61,7 @@ public class GitTaskHandler extends DvcsTaskHandler<GitRepository> {
 
   @Override
   protected void mergeAndClose(@NotNull String branch, @NotNull List<GitRepository> repositories) {
-    myBrancher.merge(branch, GitBrancher.DeleteOnMergeOption.DELETE, repositories);
+    myBrancher.merge(branch, GitBrancher.DeleteOnMergeOption.DELETE, false, repositories);
   }
 
   @Override
--- a/plugins/git4idea/src/git4idea/branch/GitBranchWorker.java
+++ b/plugins/git4idea/src/git4idea/branch/GitBranchWorker.java
@@ -112,13 +112,13 @@ public final class GitBranchWorker {
   }
 
   public void merge(@NotNull final String branchName, @NotNull final GitBrancher.DeleteOnMergeOption deleteOnMerge,
-                    @NotNull final List<GitRepository> repositories) {
+                    boolean noFastForward, @NotNull final List<GitRepository> repositories) {
     updateInfo(repositories);
     Map<GitRepository, String> revisions = new HashMap<GitRepository, String>();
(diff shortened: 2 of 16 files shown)


klikh:
Thanks a lot for your contribution! Unfortunately, we are not going to merge this pull request, because it introduces a rather strange UI decision which we try to avoid: a separate checkbox in the settings for a single use case affecting only merging from the log.

A correct fix would be to provide all sorts of necessary merge customizations, in particular both `--ff-only` and `--no-ff` options, probably in a dialog appearing on selecting the "Merge" command, but not too distractive (or maybe with a "don't show again" checkbox). There is a request about this task (https://youtrack.jetbrains.com/issue/IDEA-86425), and the UI is one of the reasons why this task hasn't been implemented yet.

Could you please also clarify the need of the list of branches which should never be merged fast-forward, can you give an example when this can be useful? And the same question about "trivial" (single commit) merges.

Please also note that in this PR you are calling Git command from the EDT thread which will likely freeze the UI (`MergeAction#actionPerformed()` calls `TrivialMergeDetector#isTrivial`).

klikh:
Btw, have you submitted a contributor license agreement? If not, please download the PDF from http://www.jetbrains.org/display/IJOS/Contributor+Agreement and then print it out, sign (mention you github account name there), scan and email to contribute@jetbrains.com

OndrejSpanel:
@klikh My CLA sent now - I sent it before, but to busdev@jetbrains.com, as that seemed most appropriate  from addresses I was able to find - the guidelines do not list any mail address, only a fax number. I see now the mail address is listed on http://www.jetbrains.org/display/IJOS/Contribute, but not in the CLA itself, which only mentions a fax numer (see http://www.jetbrains.org/display/IJOS/Contributor+Agreement).

klikh:

OndrejSpanel:
I understand your rejection, I will check if I am able to make it more along the lines of what you write.

As for the motivation: currently I almost never use Merge from log, as it does not allow me to prevent FF. The only situation where FF seems appropriate to me is when merging a single commit. I would like to make it usable for me, as it is much easier to access than the corresponding menu command.

klikh:
> The only situation where FF seems appropriate to me is when merging a single commit.
Can you give a concrete example of this situation? I really don't see much difference in 1 commit compare to 2 or 5 commits :) 

> I would like to make it usable for me
The main problem here is to make it usable for all users. To be honest, sometimes that's the most difficult part of the development process of a desktop app.

OndrejSpanel:
The difference is how the history looks like and how it can be handled in the log viewer. When multiple commits are merged into master, I want to see the commits grouped by the merge (which also makes the merge commit collapsible in the UI).

None of this is applicable when merging one commit only.

PR #362, opened 2016-03-03
Removed Perl from default extensions

hurricup:
So platform could offer Perl5 plugin from the repo

### DIFF
--- a/platform/platform-resources/src/defaultFileTypes.xml
+++ b/platform/platform-resources/src/defaultFileTypes.xml
@@ -88,24 +88,6 @@
       </highlighting>
     </filetype>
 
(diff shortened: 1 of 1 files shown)


nicity:
IMO platform should offer the plugin from repo regardless of presence simple custom file type

hurricup:
Yep, agree, but don't know how to do this :)

hurricup:
Am i getting this right that all file types generated from defaultFileType.xml are AbstractFileType or UserBinaryFileType so i just need to add a checking for it into the PluginAdvertiserEditorNotificationProvider#createNotificationPanel ?

nicity:
Thanks for bringing the issue to our attention. We are working to support this functionality ourselves, it is important part of CE behaviour.

hurricup:
Any news on this feature?

PR #364, opened 2016-03-06
Correctly decompile interface object fields

apixandru:
This commit fixes decompiling interfaces that have object references which fail sometimes. 

for instance 
```
public interface TestInterfaceWithObject {
    BigDecimal BUGS_IN_THE_DECOMPILER = BigDecimal.ZERO;
}
```

decompiles to

```
public interface TestInterfaceWithObject {
    BigDecimal BUGS_IN_THE_DECOMPILER;

    static default {
        BUGS_IN_THE_DECOMPILER = BigDecimal.ZERO;
    }
}
```

which is wrong.

(I submitted a contributor license agreement a while back.)


plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/InitializerProcessor.java
```diff
@@ -139,7 +139,8 @@ private static void extractStaticInitializers(ClassWrapper wrapper, MethodWrappe
             if (fexpr.isStatic() && fexpr.getClassname().equals(cl.qualifiedName) &&
                 cl.hasField(fexpr.getName(), fexpr.getDescriptor().descriptorString)) {
 
-              if (isExprentIndependent(asexpr.getRight(), meth)) {
+              // interfaces fields should always be initialized inline
+              if (isInterface || isExprentIndependent(asexpr.getRight(), meth)) {
```
trespasserw: On the one hand, this is correct. On the other hand, decompiler often deals with a weird bytecode (obfuscated, instrumented, non-Java etc). In your example, `isExprentIndependent()` should have decided that the initializer is independent and could be inlined - so that a "normal" code would decompile into a correct source, and a "weird" code would produce something incorrect but at least understandable.

plugins/java-decompiler/engine/test/org/jetbrains/java/decompiler/SingleClassesTest.java
```diff
@@ -106,7 +107,8 @@ protected void doTest(String testFile, String... companionFiles) {
     File decompiledFile = new File(fixture.getTargetDir(), testName + ".java");
     assertTrue(decompiledFile.isFile());
     File referenceFile = new File(fixture.getTestDataDir(), "results/" + testName + ".dec");
-    assertTrue(referenceFile.isFile());
+    assertTrue("Expecting " + referenceFile.getAbsolutePath() + " to exist", referenceFile.exists());
+    assertTrue("Expecting " + referenceFile.getAbsolutePath() + " to be a file", referenceFile.isFile());
```
trespasserw: `isFile()` implies `exists()`, so I don't feel this change is needed.

apixandru:
I never faced a problem with static initializers in any other context other than interfaces and i really do believe that it's less impacting to just have the interface check.

I created another branch where I modified isExprentIndependent but that will change the current behavior in a lot of classes. Mostly it inlines a lot of field initializations as you would expect. The thing is that the current behavior is not really incorrect for static initializers, they can have it either way, inlined when declaring the field or in a static initializer but interfaces have no choice.

If you still think that I should submit the modified version that inlines fields that "may be independent" please let me know and I will update this pull request.

P.S. I removed the exists() call from SingleClassesTest

trespasserw:
OK, convinced.

PR #366, opened 2016-03-10
[Fernflower] Refactored IIdentifierRenamer, Fixed renaming annotations on classes, Fixed the location of generated comments relative to annotations, Improved the default renaming scheme.

ccarpenter04:
Restructured the IIdentifierRenamer class to be more flexible and allow for easy expansion. Annotations on classes and class members are now also correctly renamed (they weren't before because they're loaded from the constant pool as Utf8 instead of Class or similar). I've also wrote a preliminary version of IIdentifierRenamer#shouldRenamePackage (usage of the method and it's utilities is not implemented). Classes sharing names that are the same except for differing in case are also now renamed so that the files can be properly extracted from the resulting archive. Comments in the generated source code are now placed directly above the class/field/method and below their annotations.

I've previously signed the contributor license agreement.

### DIFF
--- a/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/ClassWriter.java
+++ b/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/ClassWriter.java
@@ -299,6 +299,8 @@ public class ClassWriter {
       appendDeprecation(buffer, indent);
     }
 
+    appendAnnotations(buffer, cl, indent);
+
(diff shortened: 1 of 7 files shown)


ccarpenter04:
Originally this was going to be several individual commits and pull requests, however my fork got corrupted and so after I fixed it, it all had gotten lumped together as a single commit.

ccarpenter04:
I'm going to fix my fork so that I can make pull requests for each of these things individually.

PR #367, opened 2016-03-10
[Fernflower] Fixed annotations not being renamed properly and corrected the position of annotations relative to generated comments

ccarpenter04:
I have previously signed a contributor license agreement.

I fixed a bug that caused annotations to not be properly renamed in the generated source code when the annotation was applied to a class, method, or field.

I also changed it so that generated comments about renamed classes, methods, and fields are placed directly on top of the class, method, or field and below the annotations.

### DIFF
--- a/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/ClassWriter.java
+++ b/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/ClassWriter.java
@@ -299,6 +299,8 @@ public class ClassWriter {
       appendDeprecation(buffer, indent);
     }
 
+    appendAnnotations(buffer, cl, indent);
+
     if (interceptor != null) {
       String oldName = interceptor.getOldName(cl.qualifiedName);
       appendRenameComment(buffer, oldName, MType.CLASS, indent);
@@ -307,8 +309,6 @@ public class ClassWriter {
     if (isSynthetic) {
       appendComment(buffer, "synthetic class", indent);
     }
-
-    appendAnnotations(buffer, cl, indent);
 
     buffer.appendIndent(indent);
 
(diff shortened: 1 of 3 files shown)


trespasserw:
Could you please add a test case to `plugins/java-decompiler/engine/test/org/jetbrains/java/decompiler/SingleClassesTest.java`?

trespasserw:
While comments between an annotation and a declaration are legal, I consider this as a bad style. Please exclude corresponding changes from the PR.

ccarpenter04:
I'll gladly add a test case for it, but are you entirely sure you want me to move the comments back? I understand why one would say it's bad style in most cases, but at least in the case of the renaming comments it definitely added some confusion. For example, if the code previously looked like this:
@one
public class two {}

after being renamed the comments had it looking like this

// renamed from two
@a
public class b {}

which really confused me at first until I realized it was just awkward placement of the note. Thoughts?

trespasserw:
Doesn't look crystal-clear either:
```java
@a
// renamed from two
public class b {}
```

Probably, a better way would be dropping these comments completely and producing a separate log of all renames when requested.

ccarpenter04:
I agree that it's still not perfectly clear but I feel that it's a step in the right direction. But I do like you're idea of producing a separate log instead because currently those comments are being attached to every single renamed class, field, and method. Would you be okay with the comments being moved where I put them until someone (perhaps even myself) can get around to changing the way the name mapping is output?

trespasserw:
Okay.

ccarpenter04:
I went to write a test case but I realized that I accidentally deleted my fork. Would the best approach be for me to close this pull request, re-fork, reapply these changes, add the test case, and make a new pull request referencing this comment?

ccarpenter04:
"a better way would be dropping these comments completely and producing a separate log of all renames when requested."

What type of format would you like for this? A json file called changed_names.json?

trespasserw:
A new pull request would be easier to merge.

Didn't think much about it, but JSON sounds good enough.

trespasserw:
Ouch, commits from another project.
Can't reopen because PR's repository is gone.

PR #368, opened 2016-03-10
[Fernflower] Refactored IIdentifierRenamer to be more flexible.

ccarpenter04:
I went ahead and refactored the IIdentifierRenamer class and added some basic documentation to it. I also separated the default implementation of it into a separate class called DefaultIIdentifierRenamer and left the ConverterHelper class in place to store static methods that are useful and related to the renaming process. Classes with names differing only in case are also now renamed by default since you can't properly extract them on windows without them overwriting each other. In addition, I made it so that getNextClassName has a new argument of int accessFlags which I used to change the default class prefix from "class_" to "interface_", "enum_", and "annotation_" if appropriate. By doing that it's now easier to find what you're looking for at a glance.

### DIFF
--- a/plugins/java-decompiler/engine/readme.txt
+++ b/plugins/java-decompiler/engine/readme.txt
@@ -78,11 +78,13 @@ code leads to a great number of conflicts. Therefore it is advisable to let the 
 ensuring uniqueness of each identifier.
 
 Option 'ren' (i.e. -ren=1) activates renaming functionality. Default renaming strategy goes as follows:
-- rename an element if its name is a reserved word or is shorter than 3 characters
-- new names are built according to a simple pattern: (class|method|field)_<consecutive unique number>  
-You can overwrite this rules by providing your own implementation of the 4 key methods invoked by the decompiler while renaming. Simply 
+- rename a class member element if its name is shorter than 3 characters, is reserved in java as a keyword, or it starts with a digit.
+- rename a class if it meets the criteria for renaming a class member element, its simple name is a reserved word in the windows
+  namespace (case insensitive), or its name is equal to another class's name (case insensitive).
+- new names are built according to a simple pattern: (class|interface|annotation|enum|method|field)_<consecutive unique number>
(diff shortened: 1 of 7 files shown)


nicity:
Hi,
Thank you for your pull requests!
What about new tests and testing your changes?

ccarpenter04:
I tested it in a variety of situations on several large projects to verify it works correctly, however I'm not familiar with how you guys do unit testing because of the nature of the project.

I forgot to mention it before but I've signed the CLA previously and had several similar pull requests accepted. The difficulties that I see with unit testing something like this is that is has to do with renaming and the default renamer has rather defined guidelines that would make unit testing difficult.

ccarpenter04:
I've been working on fixing the renaming system which has long been broken and once the two pulls requests that I've submitted today are accepted there is only one more major bug in the renaming system that I'm aware of and likely a few smaller bugs that I won't be able to identify until these changes are pushed (including the fix for the remaining bug that is well documented on the issue tracker).

nicity:
Tests are needed to avoid regressions in the future

ccarpenter04:
@nicity Is there any documentation on how Fernflowers test system works? I know it doesn't function with standard unit tests. I'll attempt to write some tests for it but due to the nature of renaming and the fact that custom implementations of IIdentifierRenamer can be provided I'm thinking it'll be rather difficult to make reliable tests.

ccarpenter04:
I'm going to close this pull request because I accidentally deleted the branch it was being pulled from and because of that I can't easily add on to this. I'll make a new pull request once I've made some additional tweaks and came up with some reliable tests.

trespasserw:
@ccarpenter04 pretty standard JUnit tests (not many of them, unfortunately). See `plugins/java-decompiler/engine/build.xml` and `plugins/java-decompiler/engine/test/*` for the reference.

ccarpenter04:
@trespasserw Oh okay, now I understand how you guys are handling them. The .dec files confused me until I realized how you guys were generating them to begin with. I'll try to have the unit tests written for this sometime this week.

PR #369, opened 2016-03-12
Fix WM_CLASS x property not being null-terminated

Yannik:
The WM_CLASS must be null-terminated according to the icccm spec 4.1.2.5:

>  The WM_CLASS property (of type STRING without control characters) contains two consecutive null-terminated strings. These specify the Instance and Class names to be used by both the client and the window manager for looking up resources for the application or as identifying information. This property must be present when the window leaves the Withdrawn state and may be changed only while the window is in the Withdrawn state. Window managers may examine the property only when they start up and when the window leaves the Withdrawn state, but there should be no need for a client to change its state dynamically.

https://tronche.com/gui/x/icccm/sec-4.html

This fixes the icon for the jetbrains IDEs appearing twice due to random memory being added to the WM_CLASS because it missed the null-terminator. This happened for example in the gnome launcher.

### DIFF
--- a/platform/platform-impl/src/com/intellij/ui/AppUIUtil.java
+++ b/platform/platform-impl/src/com/intellij/ui/AppUIUtil.java
@@ -108,10 +108,11 @@ public class AppUIUtil {
 
   public static void updateFrameClass() {
     try {
+      String wmClass = getFrameClass() + "\0"; //needs to be null-terminated as defined in the icccm spec 4.1.2.5
       final Toolkit toolkit = Toolkit.getDefaultToolkit();
       final Class<? extends Toolkit> aClass = toolkit.getClass();
       if ("sun.awt.X11.XToolkit".equals(aClass.getName())) {
-        ReflectionUtil.setField(aClass, toolkit, null, "awtAppClassName", getFrameClass());
+        ReflectionUtil.setField(aClass, toolkit, null, "awtAppClassName", wmClass);
       }
     }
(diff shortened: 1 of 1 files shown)


klikh:

Yannik:
@klikh Are you sure it's necessary to sign a 1-2 page legal paper to contribute a 2 line change?
I will gladly state that I waive all my rights to these 2 lines here on github.

trespasserw:
`WM_CLASS` is assembled from an array, and the value of `awtAppClassName` field isn't always the second part of the pair. So, thank you for pointing to a problem, but I cannot accept the fix as-is, sorry. I'll submit a patch for inclusion in our build of JDK, and hopefully upstream to Oracle.

trespasserw:
In fact, the bug is already fixed by Oracle, scheduled for 8u82 (see [JDK-6961123](https://bugs.openjdk.java.net/browse/JDK-6961123) for details).

Yannik:
@trespasserw 

As 8u82 will only be released in July 2016 (see https://bugs.openjdk.java.net/projects/JDK/versions/18319), it would be good to fix this now. 

I've been using the patch I submitted here in production without any issues - can you give clearer references to the issues you are expecting to happen here?

trespasserw:
Right, that's why we'll port the fix to our custom JDK ASAP.

The problem is, at early stages, WM_CLASS would be in a "name\0\0name\0" format. I'm not sure this won't break any other WMs.

PR #373, opened 2016-03-22
replace increment / decrement synthetic access method

apixandru:
Increment / decrement synthetic methods weren't identified and replaced as such. 

For instance
```
public class Counter {
  private int count;

  private class Incrementer {
    void increment() {
      count++;
    }
  }

}
```
Used to decompile to
```
public class Counter {
   private int count;

   // $FF: synthetic method
   static int access$008(Counter x0) {
      return x0.count++;
   }

   private class Incrementer {
      void increment() {
         Counter.access$008(Counter.this);
      }
   }
}
```


gorrus:
manually pushed with minor fixes, thanks!

PR #375, opened 2016-03-24
Make NativeFileWatcherImpl#isRepetition protected.

tomlu-google:
Some file systems aren't 100% accurate in their VFS events,
causing missed events. For instance, consider for a single file:

CHANGE - DELETE - CREATE - CHANGE

If the create is missed, then the file won't be updated because
it looks like a repetition.

This allows implementors for those file systems to override this
and get the CHANGE event.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/NativeFileWatcherImpl.java
+++ b/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/NativeFileWatcherImpl.java
@@ -458,7 +458,7 @@ public class NativeFileWatcherImpl extends PluggableFileWatcher {
     }
   }
 
-  private boolean isRepetition(String path) {
+  protected boolean isRepetition(String path) {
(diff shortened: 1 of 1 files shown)


trespasserw:
VFS refresh in IDEA works in two stages: the file watcher notifications are used to mark parts of VFS tree as dirty, then the marked parts are scanned for actual changes. This way there is no need in duplicate notifications - once the file is marked dirty, it will be scanned. So, I'm not against the change - just wonder, what events are missing.

tomlu-google:
One of our file systems isn't sending the CREATE inotify response, just the
MODIFY. So you can get MODIFY DELETE (CREATE not sent) MODIFY, and the
MODIFY is missed and the file appears not to exist (because it appears to
be a duplicate). Really it's a problem with the particular file system not
being accurate, and not the file watcher, but making this method protected
would allow us to work around the issue.

On Wed, Mar 30, 2016 at 10:48 AM, Roman Shevchenko <notifications@github.com

PR #376, opened 2016-03-29
[Fernflower] Setup increment/decrement test properly

apixandru:
the source code for the test was rewritten but the class wasn't compiled and test result were not mapped, which means that the new method addition wasn't being tested


gorrus:
Thanks, fixed

PR #378, opened 2016-04-04
Fix incorrect manual watch root calculations.

tomlu-google:
Previously, if these were met:

* You have more than one file watcher, and
* One of the file watchers have unwatchable roots, and
* Some other file watcher can watch everything

You would get spurious "Cannot be watched" notifications
and manual watch roots from the system.


trespasserw:
Thanks!

tomlu-google:
No problem. Any chance this could get cherry-picked across to one of your
dot releases? It should be safe, and it's a fairly brutal experience for us
when the entire code base suddenly becomes "manual watch roots" because
it's all on a local drive.

trespasserw:
I'm going to backport it to 145 branch, it will end up in 2016.1.2 eventually. Is this enough?

tomlu-google:
Yes, certainly. We can work around the issue until then. Thank you!

PR #380, opened 2016-04-07
Inplace renaming of elements from different Psi trees 

hurricup:
Current implementation does not allow inplace renaming when related elements are in different subtrees of a MultiPsi file. This should fix this


akozlova:
it was disabled intentionally to avoid performance problems during inplace rename inside jsp. What issue do you fix?

hurricup:
No issues. Just didn't know it was disabled intentionally. Was surprised that inplace not working in different psi trees.

akozlova:
It is tricky to predict all consequences of such change, hopefully we would rewrite this code one day and there won't be such restrictions anymore. Sorry for that

PR #382, opened 2016-04-13
Fix Pydevd breakpoint issues on windows

fitermay:
Fixed a number of issues with debugger path handling on windows.

--Breakpoints weren't being hit inside bdist .egg files on windows
--Breakpoints weren't being in modules inside zip/egg files where the module name has an upper case letter
--Breakpoints not hit when the module is loaded from a short path names on windows (which is actually somewhat common)

Fixes:PY-16767


fitermay:
Typo PyLocalPositionConverter 125:
return file.substring(0, ind + 4) + file.substring(ind + 4); ->
return file.substring(0, ind + 4).toLowerCase()+ file.substring(ind + 4);

python/helpers/pydev/pydevd_file_utils.py
```diff
@@ -73,10 +73,29 @@
...
+if os.name == 'nt':
+    try:
+        import ctypes
+    except ImportError:
+        pass
+    else:
+        def convert_to_long_pathname(filename):
+            buf = ctypes.create_unicode_buffer(260)
+            GetLongPathName = ctypes.windll.kernel32.GetLongPathNameW
+            rv = GetLongPathName(unicode(filename), buf , 260)
```
Elizaveta239: The `unicode()` function doesn't exist in Python 3

python/helpers/pydev/pydevd_file_utils.py
```diff
@@ -73,10 +73,29 @@
 
 normcase = os_normcase # May be rebound on set_ide_os
 
+def convert_to_long_pathname(filename):
```
Elizaveta239: I think it's not good idea to define function `convert_to_long_pathname` twice. Let's rewrite it using inner functions.

Elizaveta239:
@fitermay Thank you for the pull request! Please, fix the problems and we will merge it into the PyCharm.

python/helpers/pydev/pydevd_file_utils.py
fitermay: Hi,

I'm not sure what you mean here. Do you just want me to replace 'def' with assignment? Could you provide an example?

fitermay:
Thanks for the comments. Could you please clarify the inner function comment? I will make the suggested changes ASAP. I also added tests that I will add to the pull request.

python/helpers/pydev/pydevd_file_utils.py
Elizaveta239: I think, we should remove the second `def`, even without inner functions, and check the platform inside the function.
```
def convert_to_long_pathname(filename):
    if os.name != 'nt':
        return filename
    else:
        # the rest of code, without def
```
It's not very important, just minors. I can change it while merging, if you're not sure.

fitermay: I have some concerns about this:
1) If we move the import ctypes inside the function it will cause a deadlock in the debugger. Since it will cause the importer lock to be acquired.  So if another thread is tracing a module level function then it will try to acquires  importer lock, then debugger lock. While our thread acquires debugger lock then importer lock. 

2) I'm trying to be mindful of performance since this function is executed once for every file traced and defining the function twice avoids performing the check every time for windows and ctypes.

fitermay:
Updated with tests and fixed type plus the unicode thing.

python/helpers/pydev/pydevd_file_utils.py
Elizaveta239: Ok, you're right. Let's move ctypes checking out of the function on the top level, save results of the checking to the variable and use it inside function.

PR #383, opened 2016-04-13
Fixed issues with Jump to source/ Jump to Type in python debugger

fitermay:
--Fixed issue with jump to type not working unless type is imported directly in the scope. Usually the type won't be imported in the scope, so it wasn't working in the common case
--Added support for jump to type, jump to source for frames other than the topmost frame
--Fixed issue with jump to source not working if the element is a module name

### DIFF
--- a/python/helpers/pydev/_pydevd_bundle/pydevd_xml.py
+++ b/python/helpers/pydev/_pydevd_bundle/pydevd_xml.py
@@ -176,6 +176,9 @@ def frame_vars_to_xml(frame_f_locals):
     return xml
 
 
+def get_type_qualifier(type):
+    return getattr(type, "__module__", "")
+
 def var_to_xml(val, name, doTrim=True, additionalInXml=''):
     """ single variable or dictionary to xml representation """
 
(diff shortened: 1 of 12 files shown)


klikh:

fitermay:
Sent

klikh:
Received, thanks.

Elizaveta239:
@filtermay Thank you for the pull request! Could you please provide an example which shows  
the functionality added by these changes?

fitermay:
Example:
---module A:
import module_b
foo_object = module_b.make_foo()
print (foo_object.field_1) #put breakpoint on this line

--module B:
class FooClass(object):
    def __init__(self):
        super(FooClass, self).__init__()
        self.field_1 = 42
        self.field_2 = 46

def make_foo():
    return FooClass()

In variables view select foo_object, jump to _type_source
Behavior before: Doesn't work because FooClass is not imported directly (actually most common case)
Behavior with change: Jumps to FooClass declaration

In variables view select foo_object-> field_1, jump to source
Behavior before: Doesn't work because field_1 is inside inner frame
Behavior with change: Jumps to field_1 assignment in the constructor

In variables view select module_b-> FooClass, jump to type source
Behavior before: Goes to the 'type' type which is useless
Behavior with change: Jumps to 'FooClass'

Elizaveta239:
It looks very nice! 
Some Java code style remarks: please, always use braces even in single statement if-blocks and leave the brace at the end of the line.

fitermay:
Sorry, forgot to reformat before push. I assume you can reformat after merging?

Elizaveta239:
Yes, of course.

PR #384, opened 2016-04-15
Minor pycharm fixes

fitermay:
Details:
One issue is refreshing the sdk in details dialog invokes PySdkUpdater, which in turn invalidates the modificator. the dialog tries to then use the invated modificator causing an assertion error and an NPE.

Issue two is framewok updater waits for process output on UI thread causing a major lockup

### DIFF
--- a/python/python-community-configure/src/com/jetbrains/python/configuration/PythonSdkDetailsDialog.java
+++ b/python/python-community-configure/src/com/jetbrains/python/configuration/PythonSdkDetailsDialog.java
@@ -226,6 +226,7 @@ public class PythonSdkDetailsDialog extends DialogWrapper {
       mySdkSettingsWereModified.run();
     }
     for (SdkModificator modificator : myModifiedModificators) {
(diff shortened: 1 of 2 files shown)


fitermay:
@vlasovskikh ping

vlasovskikh:
@fitermay Thanks for your PR. We've already [fixed](https://youtrack.jetbrains.com/issue/PY-19170) the first issue with NPE. I'll get back to this PR in a few days, sorry for the delay.

fitermay:
Hi,
I've seen the fix to the NPE. I'm pretty sure it leaves an issue of not picking up further changes in the dialog after the first Sdk refresh. That is if user refreshes Sdk then makes other changes those changes won't be picked up.

vlasovskikh:
@fitermay Please rebase your PR on top of the master branch and I'll merge it.

fitermay:
@vlasovskikh Done

vlasovskikh:
@fitermay Thank you for the PR.

PR #386, opened 2016-04-21
Fix PY-19266 -- Debugger config not being passed when connecting from console

fitermay:


python/helpers/pydev/pydevd.py
```diff
@@ -1358,6 +1358,33 @@ class SetupHolder:
     setup = None
 
 
+def apply_debugger_options(setup_options):
+    """
+
+    :type setup_options: dict[str, bool]
+    """
+
+    default_options = {'save-signatures': False, 'save-threading': False, 'save-asyncio': False, 'qt-support': False}
```
Elizaveta239: `save-signatures` and `save-threading` are not debugger options, they are used in concurrency visualisation, and can't be used from console.

python/src/com/jetbrains/python/console/PydevConsoleCommunication.java
```diff
@@ -589,13 +587,19 @@ public XSourcePosition getSourcePositionForType(String type) {
...
-  public void connectToDebugger(int localPort) throws Exception {
+  public void connectToDebugger(int localPort, @NotNull Map<String, Boolean> dbgOpts, @NotNull Map<String, String> extraEnvs)
+    throws Exception {
     if (waitingForInput) {
       throw new Exception("Can't connect debugger now, waiting for input");
     }
-    Object result = myClient.execute(CONNECT_TO_DEBUGGER, new Object[]{localPort});
+    /* argument needs to be hashtable type for compatability with the RPC library */
+    Hashtable<String, Object> opts = new Hashtable<>(dbgOpts);
+    opts.put("__extra__envs__", new Hashtable<>(extraEnvs));
```
Elizaveta239: According to other environment variables' names, let's call it in upper case and start its name with "PYDEVD"

python/src/com/jetbrains/python/console/PydevConsoleCommunication.java
```diff
@@ -589,13 +587,19 @@ public XSourcePosition getSourcePositionForType(String type) {
...
    */
-  public void connectToDebugger(int localPort) throws Exception {
+  public void connectToDebugger(int localPort, @NotNull Map<String, Boolean> dbgOpts, @NotNull Map<String, String> extraEnvs)
+    throws Exception {
     if (waitingForInput) {
       throw new Exception("Can't connect debugger now, waiting for input");
     }
-    Object result = myClient.execute(CONNECT_TO_DEBUGGER, new Object[]{localPort});
+    /* argument needs to be hashtable type for compatability with the RPC library */
+    Hashtable<String, Object> opts = new Hashtable<>(dbgOpts);
```
Elizaveta239: Hashtable is obsolete, please, don't use it.

python/src/com/jetbrains/python/debugger/PyDebugRunner.java
```diff
@@ -313,6 +302,23 @@ private void fillDebugParameters(@NotNull Project project,
     }
   }
 
+  public static void configureDebugEnvironment(@NotNull Project project, Map<String, String> environment) {
+    if (PyDebuggerOptionsProvider.getInstance(project).isSupportGeventDebugging()) {
+      environment.put(GEVENT_SUPPORT, "True");
+    }
+
+    PyDebuggerSettings debuggerSettings = PyDebuggerSettings.getInstance();
```
Elizaveta239: Will stepping filters work after connecting to debugger via console? I'm not sure.

python/src/com/jetbrains/python/console/PydevConsoleCommunication.java
fitermay: You have to use it. Otherwise the RPC library wont be able to marshall the type. You need to upgrade the library if you want to use hashmap.

python/src/com/jetbrains/python/debugger/PyDebugRunner.java
fitermay: They should because we set the environment variables before the debugger object is created.

python/src/com/jetbrains/python/console/PydevConsoleCommunication.java
fitermay: sure

Elizaveta239:
Thank you for the pull request! But I'm not sure there is real need for such fix. 
Gevent support will not work in that case. For PyQt support we create an import hook, so it will work if only PyQt is imported after connecting to debugger (also there is a matplotlib import hook in console, they can overlap each other). Does PyQt support really work after your fix?
So the only reason to do that is to collect runtime information. But for collecting it you can just run debugger once in a casual way (via Run configuration) and don't care about it anymore.

python/helpers/pydev/pydevd.py
fitermay: That's fine they are always going to be false when connecting from console. You can move them out of this function if you prefer.

fitermay:
Gevent support wont work but stepping filters, collecting run time type info will work. PyQT support worked for me as well. Ignoring stepping filters when connecting from console is certainly a bug. Users will also expect run time type info to be collected.

fitermay:
I do not understand why you think it is ok to ignore stepping filters when connecting from console. If you were the user what would you expect the behavior to be?

PR #389, opened 2016-04-25
[Fernflower] New Testing Type: stdout comparison tests (including new test cases)

droid666:
For these test you just dump java files into the test folder. They will be compiled, then run (output stored), then the classes are decompiled, compiled and run too (output compared).

So much less maintaining to do then with the other tests (for the others all tests have to be updated on fundamental changes to Fernflower that change decompiled output).

These tests don't require that the decompiler output never changes. They check that the output of the classes when run does not change and is the same as for the original classes. Also we get a failure when the decompiled classes cannot be compiled again.

It also supports optional pattern matching on the created classes.

At the moment all but the example test are deactivate (so skipped), see build.xml begin of test-stdout-prepare:

```
    <!-- here you can skip tests -->
    <patternset id="stdout-tests.exclude.pattern">
      <include name="**/*.java"/>
      
      <!-- general problems -->
      <!-- non-compilable: causes private field/method access -->
      <exclude name="**/PrivateAccess.java"/>
      <!-- non-compilable: assertion code atm can only handle the throw in the if clause, not in the else clause -->
      <exclude name="**/assertions/**/*"/>
      <!-- overloading causes many troubles... -->
      <exclude name="**/overloading/**/*"/>
      <!-- non-compilable: renaming fails -->
      <exclude name="**/renaming/**/*"/>
      
      <!-- old java versions -->
      <!-- non-compilable: (.class field trouble with assert) (remove filter on assertions above too) -->
      <exclude name="**/java14/assertions/**/*"/>
      
    </patternset>
```

### DIFF
--- a/python/educational-core/course-creator/src/com/jetbrains/edu/coursecreator/actions/CCCreateCourseArchive.java
+++ b/python/educational-core/course-creator/src/com/jetbrains/edu/coursecreator/actions/CCCreateCourseArchive.java
@@ -2,7 +2,6 @@ package com.jetbrains.edu.coursecreator.actions;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import com.intellij.icons.AllIcons;
 import com.intellij.ide.projectView.ProjectView;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.CommonDataKeys;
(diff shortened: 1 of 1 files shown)


klikh:

droid666:
Will send it. Have to reopen the pull, messed it up (git is hard :-(... )

fitermay:
@droid666 you know you can always rework the code and force push to the same branch. It will update the pull request

PR #390, opened 2016-04-25
[Fernflower] New tests stdout comparison

droid666:
(resubmitted, had a git accident..., contributor license agreement on the way...)

For these test you just dump java files into the test folder. They will be compiled, then run (output stored), then the classes are decompiled, compiled and run too (output compared).

So much less maintaining to do then with the other tests (for the others all tests have to be updated on fundamental changes to Fernflower that change decompiled output).

These tests don't require that the decompiler output never changes. They check that the output of the classes when run does not change and is the same as for the original classes. Also we get a failure when the decompiled classes cannot be compiled again.

It also supports optional pattern matching on the created classes. Note: the compiler runs with 'renaming' turned on to test that at the same time.

At the moment all but the example test are deactivate (so skipped), see build.xml begin of test-stdout-prepare:

    
    <patternset id="stdout-tests.exclude.pattern">
      <include name="**/*.java"/>

      
      
      <exclude name="**/PrivateAccess.java"/>
      
      <exclude name="**/assertions/**/*"/>
      
      <exclude name="**/overloading/**/*"/>
      
      <exclude name="**/renaming/**/*"/>

      
      
      <exclude name="**/java14/assertions/**/*"/>

    </patternset>

### DIFF
--- a/plugins/java-decompiler/engine/build.xml
+++ b/plugins/java-decompiler/engine/build.xml
@@ -7,6 +7,7 @@
     <property name="dist" value="${basedir}/fernflower.jar"/>
     <property name="test-src" value="${basedir}/test"/>
     <property name="test-out" value="${basedir}/out/test"/>
+    <property name="test-data" value="${basedir}/testData"/>
   </target>
 
   <!-- external dependencies, adjust to your own -->
@@ -45,7 +46,10 @@
     </javac>
   </target>
 
-  <target name="test" depends="init,test-compile">
+
+  <target name="test" depends="test-stdout,test-normal"/>
+
+  <target name="test-normal" depends="init,test-compile">
     <junit printsummary="true" failureproperty="tests.failed" fork="true">
       <classpath path="${test-out}:${out}"/>
       <classpath refid="junit"/>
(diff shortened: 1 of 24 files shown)


droid666:
Note: I think it needs ant 1.9.6, at least I cannot run it on command line with ant 1.9.4.

droid666:
I'm thinking about improving this. But only if you want to merge it.

Have a optional file next to each test with additional decompiler arguments (so that only the renaming tests have renaming turned on, and the overloading tests use the new -osc=1 (if you merge that)).

Also at the moment the test cases are duplicated if they should be compiled with multiple java versions (e.g. in package 'java14' and 'java15'). I could change that. Question is if default is to compile and run with all java versions and build.xml defines exceptions to that? Or the opposite: by default run with current java (1.8 atm), and define in build.xml which should be run with others too?

Also I could maybe remove the optional pattern matching on the decompiled java files. Not sure if this will ever be needed.

trespasserw:
Hi!

Thank you for the contribution, such tests are indeed useful! However, the decompiler now is a part of IDEA codebase, and I'm hesitant in having such tests in it. It would be better to create a separate repository for them. What do you think?

PR #391, opened 2016-04-25
[Fernflower] new option -osc: solves problems with method/field selection

droid666:
New option -osc (OVERLOADING_SAFETY_CASTS), turned off by default.
Fix for IDEA-135385 and others.

Overloaded and hidden fields and methods caused errors. All parameters types and the caller type must match the descriptor exactly, if not a cast is needed. The reason is that the current instance could have been cast to a sub/super class in the original code and we cannot know that. And we don't know if the method/field is overloaded on that sub/super class.

This options adds a lot more casts to the decompiler output, but it is much safer and can solve many cases that fail otherwise.
Test cases included in this pull request: https://github.com/JetBrains/intellij-community/pull/390

### DIFF
--- a/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/ExprProcessor.java
+++ b/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/ExprProcessor.java
@@ -851,12 +851,12 @@ public class ExprProcessor implements CodeConstants {
                                          TextBuffer buffer,
                                          int indent,
                                          boolean castNull,
-                                         boolean castAlways,
+                                         boolean requiresExactTypeMatch,
                                          BytecodeMappingTracer tracer) {
     VarType rightType = exprent.getExprType();
 
     boolean cast =
-      castAlways ||
+      (requiresExactTypeMatch && !leftType.equals(exprent.getExprType())) ||
       (!leftType.isSuperset(rightType) && (rightType.equals(VarType.VARTYPE_OBJECT) || leftType.type != CodeConstants.TYPE_OBJECT)) ||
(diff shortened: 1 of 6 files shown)


plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/ExprProcessor.java
```diff
@@ -851,12 +851,12 @@ public static boolean getCastedExprent(Exprent exprent,
...
                                          int indent,
                                          boolean castNull,
-                                         boolean castAlways,
+                                         boolean requiresExactTypeMatch,
                                          BytecodeMappingTracer tracer) {
     VarType rightType = exprent.getExprType();
 
     boolean cast =
-      castAlways ||
+      (requiresExactTypeMatch && !leftType.equals(exprent.getExprType())) ||
```
trespasserw: First, in many cases the expression is a part of a complex statement (e.g. multi-parameter method invocation). So whether or not the cast is needed should be decided on a call site.

Next, omitting a cast because of a type equality doesn't seem safe. Wouldn't it fail on an example from IDEA-135385?

plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/exps/NewExprent.java
```diff
@@ -205,6 +205,7 @@ public TextBuffer toJava(int indent, BytecodeMappingTracer tracer) {
           }
         }
 
+        boolean forceSameType = DecompilerContext.getOption(IFernflowerPreferences.OVERLOADING_SAFETY_CASTS);
```
trespasserw: For constructor invocations, overloads from a base class are impossible. So it should be possible to adopt the ambiguity check from InvocationExprent.

PR #393, opened 2016-04-26
[Fernflower] Renaming support for .class field access and assertions

droid666:
If you use renaming, .class field access and assertions don't decompile correctly in some cases (output contains synthetics).

Examples:
https://github.com/droid666/intellij-community/tree/398801bec078efc14529078b22e4b109f7efcadf/plugins/java-decompiler/engine/testData/src-stdout/javadefault/renaming

The examples are part of this pull request: https://github.com/JetBrains/intellij-community/pull/390
They are deactivate in the pull by default because they fail without this fix.

Example B needs the first commit of this pull to work (access to .class on renamed class).
Example A also needs the 2nd commit of this pull to work (Assertion location is not recognized as Assertion without adding renaming there).

The question is if this is the correct fix or if these 2 problems mean something does not work as it should?

### DIFF
--- a/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/AssertProcessor.java
+++ b/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/AssertProcessor.java
@@ -105,7 +105,17 @@ public class AssertProcessor {
 
                     ClassNode nd = node;
                     while (nd != null) {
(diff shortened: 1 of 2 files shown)


gorrus:
Is it possible to extract tests specific to this pr and attach here?

trespasserw:
Ouch, commits from another project.

BartvHelvert:
Hi, since there was no activity in this PR for years, I am going to close it. Feel free to open another PR if you still want to contribute this change. I think it needs tests though.

PR #394, opened 2016-04-26
[Fernflower] Support Asserts that have the throw in the else clause instead of the if clause

droid666:
As the title says: the ``throw new AssertionError()`` of a compiled assertion can be in the if or the else clause. Currently Fernflower only deals with the case where it is in the if clause.

Example:
https://github.com/droid666/intellij-community/blob/398801bec078efc14529078b22e4b109f7efcadf/plugins/java-decompiler/engine/testData/src-stdout/javadefault/assertions/Assert1.java

In this example the method ``foo(Object x)`` is the problem, the assertion in it decompiles to: ``if($assertionsDisabled || ...`` with the throw in the else clause.

```
  public static int foo(Object x) {
    assert (x.toString().equals("x") && i < j) || (i > j && y == i);
    return 1;
  }
```

The example is part of this pull request: #390
It is deactivated in that pull by default because it fails without this fix.


gorrus:
merged into community, thanks!

PR #395, opened 2016-04-27
PY-14330: Pandas dataframes viewer in debug or console

fitermay:
PY-14330
Pandas dataframes viewer in debug or console


davidljung:
Awesome! Thanks!

python/pydevSrc/com/jetbrains/python/debugger/pydev/GetDataFrameCommand.java
```diff
@@ -0,0 +1,83 @@
...
+import com.jetbrains.python.debugger.PyDebuggerException;
+
+/**
+ * @author amarch
+ *
+ */
+/*
+  Really copied from GetArrayCommand -- maybe refactor later
+ */
+public class GetDataFrameCommand extends GetFrameCommand {
```
traff: It would be nice to refactor it a bit to avoid copy-paste.

traff:
I got an NPE exception trying to view the data frame having stopped at the last line of the code

```python
import numpy as np
import pandas as pd

s = pd.Series(np.random.randn(5), index=['a', 'b', 'c', 'd', 'e'])

d = {'one' : pd.Series([1., 2., 3.], index=['a', 'b', 'c']),
         'two' : pd.Series([1., 2., 3., 4.], index=['a', 'b', 'c', 'd'])}

df = pd.DataFrame(d)

print(df)

```

Exception:
```
Argument for @NotNull parameter 'chunk' of com/jetbrains/python/debugger/dataframe/DataFrameTable.initUi must not be null
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'chunk' of com/jetbrains/python/debugger/dataframe/DataFrameTable.initUi must not be null
	at com.jetbrains.python.debugger.dataframe.DataFrameTable.initUi(DataFrameTable.java)
	at com.jetbrains.python.debugger.dataframe.DataFrameTable.access$500(DataFrameTable.java:42)
	at com.jetbrains.python.debugger.dataframe.DataFrameTable$6.run(DataFrameTable.java:164)
	at com.intellij.openapi.application.impl.ApplicationImpl$2.run(ApplicationImpl.java:333)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

python/pydevSrc/com/jetbrains/python/debugger/pydev/ProcessDebugger.java
```diff
@@ -89,4 +89,7 @@ void setBreakpointWithFuncName(@NotNull String typeId, @NotNull String file, int
   void addExceptionBreakpoint(ExceptionBreakpointCommandFactory factory);
 
   void removeExceptionBreakpoint(ExceptionBreakpointCommandFactory factory);
+
+  ArrayChunk loadPandasDataFrame(String id, String id1, PyDebugValue var, int offset, int offset1, int rows, int cols, String format)
```
traff: id1 and offset1 aren't good examples of variable naming. But as this method signature is the same as loadArrayItems we could reuse it having the separation only on python side based on the type of the variable.

python/helpers/pydev/_pydevd_bundle/pydevd_process_net_command.py
```diff
@@ -203,6 +203,23 @@ def process_net_command(py_db, cmd_id, seq, text):
                 except:
                     traceback.print_exc()
 
+            elif cmd_id == CMD_GET_DATAFRAME:
```
traff: This repeats CMD_GET_ARRAY, which could be reused instead.

fitermay:
I'll take a look at the exception tonight. To see if I can reproduce. I wasn't sure which way to go on the reuse vs clear naming of functions. Anyways, I will refactor to avoid duplication.

fitermay:
pandas1

I cannot reproduce the error. What version of Python and Pandas are you testing with?

fitermay:
@traff Reworked according to comments. Still cannot reproduce error.Tested with Python 2.7 and Python 3.5, pandas version 0.18.0

fitermay:
@traff Tested down to pandas 0.11.0. Still cannot reproduce.

ophiry:
is there a binary version available (windows)?

fitermay:
@ophiry There will not be a build until this PR hopefully gets accepted by Jetbrains. At that point it should become available in their EAP build.

traff:
@fitermay I can't reproduce the NPE now.

traff:
@fitermay Thanks a lot for the improvements! Still one thing still bothers me in the implemented approach. That is the big portion of duplication of many classes of the data frames support with the existing numpy array support. We plan to make improvements in the array viewer component soon, thus having a code duplication will make the work much harder. The code duplication I'm is basically the classes DataFrameTable and NumpyArrayTable, DataFrameCellRenderer and ArrayTableCellRenderer, PyViewDataFrameAction and PyViewArrayAction, ArrayTableForm and DataFrameTableForm.

fitermay:
I duplicated to reduce the risk of breaking existing code, since there are no tests for the existing code. A good chunk of it is UI code and cannot be automatically tested, at least not with the existing framework.  If you prefer me to merge them I will do it.  I prefer merging them myself but thought you'd feel more comfortable with existing code being unchanged.

traff:
@fitermay Regressions aren't a big problem. If it all well written we a) Test if for regressions b) Include into our EAP for the next version to reveal possible bugs before release

By the way, tests also won't harm. To test things in run-time we have so called Env Tests (look into PythonDebuggerTest for example). They could help to tests models at least.

fitermay:
Ok, I will refactor to remove duplication whenever possible.

fitermay:
@traff Ok, refactored to remove most duplication.

python/pydevSrc/com/jetbrains/python/debugger/pydev/ProtocolParser.java
```diff
@@ -267,27 +268,63 @@ private static PyDebugValue parseValue(final XppReader reader, PyFrameAccessor f
...
 
     Object[][] data = parseArrayValues(reader, frameAccessor);
-    return new ArrayChunk(result.getValue(), result.getSlicePresentation(), result.getRows(), result.getColumns(), result.getMax(),
-                          result.getMin(), result.getFormat(), result.getType(), data);
+    result.setData(data);
+    return result.createArrayChunk();
+
+  }
+
+  private static void pasrseArrayHeaderData(XppReader reader, ArrayChunkBuilder result) throws PyDebuggerException {
```
traff: A tiny typo here.

fitermay: -oops, fixed

python/src/com/jetbrains/python/debugger/dataframe/DataFrameTableModel.java
```diff
@@ -0,0 +1,144 @@
...
+import com.jetbrains.python.debugger.array.TableChunkDatasource;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by Yuli Fiterma on 4/26/2016.
+ */
+public class DataFrameTableModel extends AsyncArrayTableModel {
+  private final Map<Integer, String> rowLabels = Maps.newConcurrentMap();
```
traff: According to the IntelliJ style guide fields should start with 'my' prefix.

traff:
Yuli, thanks a lot for your effort. Now it looks better. Still I find it a bit more duplicated then it could be, for example NumpyArrayTable/Form and DataFrameTable/Form are basically the same. The only major difference is handling of headers, which could be reflected in two different header models instead of having two different approaches of table initialisation.

fitermay:
@traff I handle both column headers and row headers differently.  I handle row headers as simply the first column.  The numpy renderer uses a custom table (JBTableWithRowHeaders) where the column headers are a seperate 1 column table. This is where the main difference come from.  While in theory it is nice to have row header model and table model separate I honestly cannot be sure that the two component approach handles all cases correctly.  If I was a SWING component expert maybe I'd feel more confident.  All I know from general principle is that trying to implement your own components without being a super expert often leads to several iterations of subtle issues.

So what approach would you like me to take here?

fitermay:
@traff Let me elaborate. If I use JBTableWithRowHeaders instead of a regular JBTable , now I have to think:
1) What happens if the theme changes?
2) What happens if I add a row sorter?
3) What happens if I add a row filter?

None of these cases are handled correctly by JBTableWithRowHeaders and these are just the cases I came up with off the top of my head.  So I really do not want to reuse JBTableWithRowHeaders . Maybe change NumpyArrayTable to use regular JBTable and use the approach I used for the pandas table?

python/src/com/jetbrains/python/debugger/dataframe/DataFrameTableModel.java
fitermay: Ok, this is fixed.

fitermay:
@traff Ok, fixed up and moved to the model. This should be ok to merge now

traff:
@fitermay Thanks Yuli, looks good for me. I've put the branch into our repository, it will be merged into master after we test it.

PR #396, opened 2016-04-29
IDEA-118247 fix application of Color Scheme to Terminal.

Elemecca:
Previously the Terminal view only used ANSI colors 1-6 (the non-bright colors except for black and white) from the Color Scheme and inherited colors 0 and 7-15 from the JediTerm defaults. That resulted in strange partial application of the Color Scheme and, frequently, unreadable text. This corrects the palette implementation to pull all 16 ANSI colors from the Color Scheme.

This resolves [IDEA-118247](https://youtrack.jetbrains.com/issue/IDEA-118247).


klikh:
The CLA was received. Thanks.

traff:
Hi Sam,
Thanks a lot for fixing it!

PR #397, opened 2016-04-30
Fix PY-11224 -- Debugger, variable assignment from console inside function fails

fitermay:


python/helpers/pydev/_pydevd_bundle/pydevd_save_locals.py
```diff
@@ -32,32 +22,48 @@ def save_locals(frame):
...
             import __pypy__  # @UnresolvedImport
             save_locals = __pypy__.locals_to_fast
-            save_locals(frame)
-            return
     except:
         pass
+    else:
+        if '__pypy__' in sys.builtin_module_names:
+            def save_locals_pypy_impl(frame):
+                save_locals(frame)
```
Elizaveta239: It's better to use another name for `locals_to_fast`, because we already have `save_locals` function in this file.

python/helpers/pydev/_pydevd_bundle/pydevd_save_locals.py
```diff
@@ -32,32 +22,48 @@ def save_locals(frame):
...
 
     try:
         import ctypes
+        locals_to_fast = ctypes.pythonapi.PyFrame_LocalsToFast
     except:
-        return #Not all Python versions have it
+        pass
+    else:
+        def save_locals_ctypes_impl(frame):
+            locals_to_fast(ctypes.py_object(frame), ctypes.c_int(0))
```
Elizaveta239: Let's use lambdas instead of one-line wrapper functions.

Elizaveta239:
Thank you for the fix, it will be merged later.

python/helpers/pydev/_pydevd_bundle/pydevd_save_locals.py
fitermay: You cannot use a lambda here. It is a statement not an expression.

PR #399, opened 2016-05-15
PY-19541: Fix test runner progress for py.test

kroo:
This change fixes an issue where the PyCharm pytest runner did not report the number of tests to the test runner, resulting in a progress bar stuck at 50%.

https://youtrack.jetbrains.com/issue/PY-19541


throwable-one:
Thank you! Accepted. According to uoyr internal rules we do not merge pull requests here, but pull em in our internal repo and sync here instead. So, your commit should be here in 1 day or so.

PR #404, opened 2016-05-24
PsiPrimitiveType#getText(): avoid call to PsiType#getAnnotations() if redundant

pTalanov:


trespasserw:
Was fixed last year in a very similar manner.

PR #407, opened 2016-05-28
Add additional deserialization note for index value violate

Haehnchen:
I debugged some plugins against this implemention. this gives better clue, what is wrong in detail


klikh:
Merged manually. Thanks!

PR #414, opened 2016-06-17
enable debugging varargs where the first argument resolves to null

apixandru:
Varargs don't get wrapped if the first argument resolves to null (but they do if it is actually null)

image

image

The original author put in extra effort to make sure that this doesn't case doesn't get wrapped into varargs but he did not provide an explanation as to why this is. Also there are currently no tests that cover varargs conversion in ExpressionBuilderImpl (or any other class in the debugger)

Since we don't know what it was protecting against, I suppose that we can remove it until the case that it was trying to avoid comes up and we can see if we can handle it in a different way?

### DIFF
--- a/java/debugger/impl/src/com/intellij/debugger/engine/evaluation/expression/EvaluatorBuilderImpl.java
+++ b/java/debugger/impl/src/com/intellij/debugger/engine/evaluation/expression/EvaluatorBuilderImpl.java
@@ -1416,10 +1416,6 @@ public class EvaluatorBuilderImpl implements EvaluatorBuilder {
                                          final Evaluator[] argumentEvaluators) {
     int lastParam = declaredParams.length - 1;
     if (lastParam >= 0 && declaredParams[lastParam].isVarArgs() && argumentEvaluators.length > lastParam) {
-      // only wrap if the first varargs parameter is null for now
-      if (!TypeConversionUtil.isNullType(actualArgumentExpressions[lastParam].getType())) {
-        return argumentEvaluators;
-      }
       // do not wrap arrays twice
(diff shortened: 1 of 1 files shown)


apixandru:
@gorrus do you remember why the special case for null as the first parameter?

gorrus:
The code fixed the bug https://youtrack.jetbrains.com/issue/IDEA-107925 which is probably actually a jdk bug inside com.sun.tools.jdi.MethodImpl.handleVarArgs which does not do wrapping if the first vararg value is null.
I'll file a jdk bug on this.
Not sure it is a good idea to do the wrapping always, this was a workaround and wrapping inside EvaluatorBuilderImpl is much more dangerous as we do not have values yet and so may do incorrect conversions.

apixandru:
I just realize that just removing those lines will cause it to succeed if we have elements with the same type but not if they are different. (because of the type inference issue that you mentioned)

However, I'm thinking that substituting the expression that evaluates to null to actually null (inside wrap varargs) would probably work because it always works with the current code. Then the type inference will be just as it is today which works because the first argument's type doesn't play a role in determining the varargs type.

What do you think?

gorrus:
I'm not sure.
I've contacted Oracle support.
At the same time maybe a better place for the fix is inside com.intellij.debugger.engine.DebugProcessImpl.InvokeCommand.invokeMethodAndFork?
There is one workaround for varargs case there already (and we have all values there).

gorrus:
I've just fixed this within https://youtrack.jetbrains.com/issue/IDEA-157321, moved all the wrapping code from EvaluatorBuilderImpl into invokeMethodAndFork.
Thanks for mentioning this.

PR #415, opened 2016-06-18
Prevent JNA library conflicts

theroflcoptr:
Adding this option prevents a JNA library conflict with other software when trying to upgrade IntelliJ with the patching mechanism. See [here](https://github.com/java-native-access/jna/issues/384) for more info.


ignatov:
@theroflcoptr, please rebase the PR, cc @chashnikov.

chashnikov:
Sorry for the late reply, we somehow missed this pull request. It looks like we already set `jna.nosys` property inside the IDE (see 83eeacb0e8e) and while installing patches (see cbcfe98e5fffd6bb). Also `IDEA` run configuration is used to run IntelliJ IDEA from sources only, and it isn't possible to update it via patch in that mode.

If you have any problems with JNA library in future, please firstly file an issue in [our issue tracker](https://youtrack.jetbrains.com/issues/IDEA).

PR #416, opened 2016-06-18
Fix PY-19825

fitermay:


sproshev:
Thanks! Your changes has been applied :)

PR #418, opened 2016-06-21
Implemented getHashCode(CharSequence,int,int)

collinsmith:
Implemented getHashCode(CharSequence,int,int) overload which can add nodes using a CharSequence as input.


ignatov:
@donnerpeter please review.

donnerpeter:
What do you need this for?

PR #419, opened 2016-06-23
Fix PY-19826

fitermay:


sproshev:
What do you think about inheriting PyTupleType from PyCollectionType? Maybe it will be more general solution of this issue.

fitermay:
I think it can be done but should be treated as a separate task/issue from this PR due to the risks of such a change.

fitermay:
I  guess what I'm trying to say is that I think it's a good idea but I'm not 100% confident it won't break anything. If you think it won't break things I can do it that way.

fitermay:
@sproshev so which route do you want to go here?

sproshev:
@fitermay Let's try with inheriting `PyTupleType` from `PyCollectionType`.

I think methods below should be updated:

* `PyCollectionTypeImpl#equals` to accept only `PyCollectionTypeImpl` objects
* `PyCallExpressionHelper#getCallTargetReturnType`. I think `PyTupleType` should be processed in the same way as `PyCollectionType` but new type should be constructed with `PyTupleType::new`
* `PyFunctionImpl#getYieldStatementType`. Add all `PyTupleType.getElementTypes` to `types` so they would become a union :)
* `PyNamedParameterImpl#getType` same as for `PyCallExpressionHelper#getCallTargetReturnType`.
* `PyTypeChecker#collectGenerics` swap processing `PyCollectionType` and `PyTupleType` :)
* `PyTypeChecker#match(PyType, PyType, TypeEvalContext, Map<PyGenericType,PyType>, boolean)` swap processing `PyTupleTypes` and `PyCollectionTypes`
* `PyTypeModelBuilder#build(PyType, boolean)` the same, `PyTupleType` should be processed before `PyCollectionType`. 

What do you think? I think it would be great if we succeed :)

sproshev:
Maybe it's better to start from `PyTypeChecker#match(PyType, PyType, TypeEvalContext, Map<PyGenericType,PyType>, boolean)` to make sure that processing become easier

fitermay:
while we're at it I think PyUnionType should be updated to be more
efficient, not rely on the order the types are added and remove the hack
where null inside the set mean "unknown" type.

fitermay:
@sproshev: Ok! Pushed the refactored code to this PR. 
PyNamedParameterImpl#getType and PyCallExpressionHelper#getCallTargetReturnType don't need to be changed , I think, because the tuple stub doesn't have an __init__ method.

fitermay:
@sproshev Bonus changes: Handle the tuple(...) constructor and mixed type lists.
Had to change PyNamedParameterImpl#getType and PyCallExpressionHelper#getCallTargetReturnType after all to support these.

fitermay:
@sproshev ping

python/src/com/jetbrains/python/psi/types/PyTupleType.java
```diff
@@ -124,4 +126,15 @@ public int hashCode() {
...
+  @NotNull
+  @Override
+  public List<PyType> getElementTypes(@NotNull TypeEvalContext context) {
+    if (myHomogeneous) {
+      return Collections.singletonList(myElementTypes[0]);
+    }
+    else {
+      return Collections.singletonList(PyUnionType.union(Arrays.asList(myElementTypes)));
+    }
+  }
```
sproshev: PyTupleType has a bit another meaning of `getElementTypes` method: `getElementTypes().get(i)` should return type of `i`-th element in tuple

python/src/com/jetbrains/python/documentation/PyTypeModelBuilder.java
```diff
@@ -197,7 +197,16 @@ public TypeModel build(@Nullable PyType type,
...
+    if (type instanceof PyTupleType) {
+      final List<TypeModel> elementModels = new ArrayList<TypeModel>();
+      final PyTupleType tupleType = (PyTupleType)type;
+      for (int i = 0; i < (tupleType.isHomogeneous() ? 1 : tupleType.getElementCount()); i++) {
+        final PyType elementType = tupleType.getElementType(i);
+        elementModels.add(build(elementType, true));
+      }
+      result = new TupleType(elementModels, tupleType.isHomogeneous());
+    }
+    else if (type instanceof PyCollectionType) {
```
sproshev: It's better to create test or tests for changes which could change PyCharm behaviour

python/testSrc/com/jetbrains/python/PyTypeTest.java
```diff
@@ -816,6 +816,12 @@ public void testDefaultTupleParameterMember() {
            "  expr, foo = xs\n");
   }
 
+  // PY-19826
+  public void testListFromTupleType() {
+    doTest("List[Union[str, int]]",
+           "t = ('1',2,3); expr = list(t)");
+  }
+
```
sproshev: It's better to put tests in the same commit as code changes

python/psi-api/src/com/jetbrains/python/psi/types/PyClassType.java
```diff
@@ -25,4 +25,11 @@
 public interface PyClassType extends PyClassLikeType, UserDataHolder {
   @NotNull
   PyClass getPyClass();
+
+  /**
+   * @param newClass
+   * @return This class type repalced with another type. Intended to be used for 'upcasting'
+   */
+  @NotNull
+  PyClassType toClass(@NotNull PyClass newClass);
```
sproshev: I think it's more util method than a method which should be in the interface.

python/src/com/jetbrains/python/psi/types/PyTypeChecker.java
```diff
@@ -297,9 +297,8 @@ else if (type instanceof PyUnionType) {
       }
     }
     else if (type instanceof PyTupleType) {
-      //TODO: What about homogeneus tuple types?
       final PyTupleType tuple = (PyTupleType)type;
-      final int n = tuple.getElementCount();
+      final int n = tuple.isHomogeneous() ? 1 : tuple.getElementCount();
```
sproshev: This change isn't related to commit topic so it should be in another one.

python/src/com/jetbrains/python/psi/types/PyTupleType.java
fitermay: I'm pretty sure this was intentional by me. So that PyTupleType would be compatible with lists and collections with one generic parameter.

python/testSrc/com/jetbrains/python/PyTypeTest.java
fitermay: Does this test still pass even after changing the implementation of getElementTypes to return type of i-th element in tuple?

sproshev: Implementation of PyTypeChecker was updated so that case for expected collection and actual tuple was introduced

fitermay: My idea was to treat the tuple as a single generic typed collection when accessed through PyCollectionType API and as multi-generic when accessed through PyTupleType API (unless homogeneous).  Does it even make sense to have PyTupleTyple implement PyCollectionType if you have to special-case it in PyTypeChecker? Maybe just scrap the idea. I'm not too attached to my code...

PR #425, opened 2016-07-10
Fix PY-1177 Control flow analysis doesn't detect while/break

fitermay:


vlasovskikh:
Thanks for your PR! I'll review it later.

We would like to unify our static evaluation code (`PyConstantExpressionEvaluator`, `PyEvaluator`, and others) and make it explicit what get evaluated and how much it affects performance.

fitermay:
Does it mean I need to change something or is it just a general comment?

vlasovskikh:
@fitermay No, your code looks OK at first glance.

fitermay:
ping?

python/src/com/jetbrains/python/codeInsight/controlflow/PyControlFlowBuilder.java
```diff
@@ -738,7 +738,7 @@ public void visitPyAssertStatement(final PyAssertStatement node) {
     super.visitPyAssertStatement(node);
     final PyExpression[] args = node.getArguments();
     // assert False
-    if (args.length >= 1 && PyConstantExpressionEvaluator.evaluate(args[0]) == Boolean.FALSE) {
+    if (args.length >= 1 && PyConstantExpressionEvaluator.evaluateBoolean(args[0], true)) {
```
sproshev: I suppose you forgot about "!" before PyConstantExpressionEvaluator.evaluateBoolean(args[0], true) :)

PR #424, opened 2016-07-05
IDEA-157915: Fix symbolic-ref in HEAD

illabefat:

PR #429, opened 2016-07-21
IDEA-157915: Fix symbolic-ref in HEAD #2

illabefat:


illabefat:
@klikh Here is a fix for problem with packed-refs.
For now there is only a problem with drawing of Branch view (ctrl+V->7)
It correctly says that master is a current branch.
The problem is that you can't switch from sym-link branch to original one: you don't have it 'Local Branches' section.  (See the attached image)
I don't know what is the best solution here. 
And I don't know where to look for the code of this dialog =(

Example:
Branch view after executing `git checkout ilzi`, which points to `master`
screen shot 2016-07-21 at 20 07 05

klikh:
OK, it seems that the problem is a bit more complex than it seemed at first. I think we have to decide on the UI first, and then make an implementation so that it matches the UI we agree on.

Here are my thoughts on this:
0. Let user checkout ilzi branch which is a link to master.
1. The Push dialog should treat ilzi the same as master, in particular, use master's tracked branch as ilzi's tracked branch.
1.1. Same for Update Project action.
2. If I checkout ilzi, I should see "Current branch: ilzi" both in the status bar and in the "Git Branches" popup footer. Here I disagree with the command line prompt that shows master.

What do you think about 2, and what other use cases there are?

klikh:
Hm. Have you rebased the pull request? I don't see neither code comments, nor the pull request discussion. Or are they hidden somewhere inside github UI?

illabefat:
I didn't manage to reopen old PR. Either I am github newbie, or github thinks that reopening merged PR is strange, or both of them. So I created a new PR.

klikh:
Ah, my bad. I thought that you've opened a new PR, but didn't find it because I didn't look in closed PRs. Here it is for reference: https://github.com/JetBrains/intellij-community/pull/424

illabefat:
Regarding usecases, as I understand there are quite few symlink usecases:
- List of branches
- Checkout symlink
- Understand current branch
- _Create, update, delete symlinks_ (out of scope for now) 
All the other time git works with normal physical branches.
I don't think that IDEA should behave in different way from 'official' git client. Official git client shows `master` as a current branch, and I think that we should do the same.
I would think about fixing branch view to understand symlinks in head: i.e. 'Current branch: master(ilzi)', not filtering of master from branch list, etc.
But we can also introduce some cool method `getRealBranchFromHead` and call it from almost everywhere...

klikh:
It seems that to solve these issues, instead of treating symrefs as normal branches and resolving them while reading the repository, we'll have to introduce a special conception for them, e.g. something like `GitSymbolicRef`.

Btw, could you please share your use cases: why do you use links at all, instead of operating the real branches? AFAIK they are not an official feature of Git: back in 2011 Junio [said](http://permalink.gmane.org/gmane.comp.version-control.git/166812) that only the `HEAD` and `refs/remotes/origin/HEAD`s are officially supported links. Has something changed after that.

illabefat:
Well... we have quite a long names of branches. I want to have alias `stable` for stable branch =)
Actually IntelliJ's branch view solves this problem, but I perform `checkout` from terminal, because it usually requires couple more actions.
Not sure if something has changed since Junio's answer, I just wanted to make my work a bit more comfortable.

klikh:
So do you mind trying the approach with separate class for symbolic refs?

> And I don't know where to look for the code of this dialog

It is `GitBranchPopup`. Note that multi-repository case is supported there: branches common for all repositories are displayed at the top-level.

> I don't think that IDEA should behave in different way from 'official' git client. Official git client shows master as a current branch, and I think that we should do the same.

Since we see that there is no official support of symbolic refs, following git client looks less necessary. On the other hand, the approach with braces looks nice. Just in reverse order: "stable (master)". IMHO it would be weird to see "Current branch: master" after checking out "stable", wouldn't it? In addition to that, if we display master (or even "master (stable)"), it would be unclear why there is "master" in Git | Branches popup.

illabefat:
Yes, I'll probably try to do something about this

> Just in reverse order: "stable (master)". IMHO it would be weird to see "Current branch: master" after checking out "stable", wouldn't it?

Well, it a good point =) I've just written first to come in mind.

What is the case of multi-repository? 2 modules from different folders in one project? git repo inside other git repo?

klikh:
> What is the case of multi-repository? 2 modules from different folders in one project? git repo inside other git repo?

Both, and anything else you can imagine. For example, full intellij-community project consists of 3 repositories nested one into another: this one + [/android](https://github.com/JetBrains/android) + [/tools-base](https://github.com/JetBrains/adt-tools-base). We have users with dozens of repositories (usually, some web frameworks with many small repositories for each of web site modules). 

And we even have users with mixed VCS projects having e.g. a Git repository, a Hg repository and an SVN repository in a single project. However, this case is not related to the original issue: here we care only about multiple Git repositories.

Repositories are registered in `Settings | Version Control` by specifying the path and the VCS.

In IDEA some operations are synchronous for all repositories by default (Update Project, Push, Commit), some have no sync support yet (e.g. Git | Pull, Git | Stash). Branches can operate synchronously or individually depending on the value of `Settings | Version Control | Git | Control repositories synchronously`. If this checkbox is enabled, you can e.g. checkout a common branch in all repositories. See [this blog post](https://blog.jetbrains.com/idea/2012/08/git-branches-for-multi-root-projects/) for details.

illabefat:
@klikh 
Here is a draft of fix.
Please, have a look at this.
Current branch is written as 'ilzi(master)', symbolic-refs in list of branches point at actual remote.
screen shot 2016-08-22 at 22 30 00

Push is being performed to actual remote, pull also works.
screen shot 2016-08-22 at 22 32 04

If this is OK, I will continue with some tests for it

klikh:
Looks good. I'd add a space between "ilzy" and "(master)".

Will add some comments to the code. 

And could you please squash 39b8698 and 7966365, to avoid unfinished work in history ;)

klikh:
Please add tests for the following cases:
1. HEAD points to reflink => current branch is reflink, but its hash is correctly resolved.
2. reflink points to another reflink which point to a real ref
3. reflink points to remote branch.
4. cyclic reflinks don't get resolved (probably absent in the list of branches), but don't crash the whole thing :)

You may use either `GitNewRepositoryReader` or `GitRepositoryReader` approach, as you wish. Personally, I find *New* one more clear since it allows to prepare the case manually, and not to look into testData to understand the test case.

plugins/git4idea/src/git4idea/GitReference.java
```diff
@@ -28,7 +28,7 @@
 
   public static final TObjectHashingStrategy<String> BRANCH_NAME_HASHING_STRATEGY = FilePathHashingStrategy.create();
 
-  @NotNull protected final String myName;
+  @NotNull private final String myName;
```
klikh: It is good to encapsulate the name, but I'd prefer it to be made in a separate commit, since it doesn't directly apply to the symbolic-ref feature.

plugins/git4idea/src/git4idea/GitSymbolicRef.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @NotNull private final GitLocalBranch referencedBranch;
+
+  public GitSymbolicRef(@NotNull String name, @NotNull GitLocalBranch branch) {
+    super(name);
+    referencedBranch = branch;
+  }
+
+  @NotNull
+  public GitLocalBranch getReferencedBranch() {
```
klikh: AFAIU it is possible for a reflink to reference a remote branch, isn't it? If it is possible, it should be covered, therefore here the returned class should be `GitBranch` (maybe other changes will be required too).

plugins/git4idea/src/git4idea/GitSymbolicRef.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Nullable
+  @Override
+  public final GitRemoteBranch findTrackedBranch(@NotNull GitRepository repository) {
+    return referencedBranch.findTrackedBranch(repository);
+  }
+
+  @NotNull
+  public String getPrintableName() {
+    return super.getName() + "(" + referencedBranch.getName() + ")";
```
klikh: Let's add a space between `getName()` and the opening parenthesis. 

Btw, why `super.getName()` and not just `getName()`?

plugins/git4idea/src/git4idea/repo/GitRepositoryReader.java
```diff
@@ -85,6 +85,7 @@ GitBranchState readState(@NotNull Collection<GitRemote> remotes) {
     Pair<Map<GitLocalBranch, Hash>, Map<GitRemoteBranch, Hash>> branches = createBranchesFromData(remotes, refs);
     Map<GitLocalBranch, Hash> localBranches = branches.first;
 
+    //HeadInfo headInfo = readHead(refs);
```
klikh: forgot to remove the comment.

plugins/git4idea/src/git4idea/branch/GitBranchUtil.java
```diff
@@ -282,7 +283,12 @@ public static String getDisplayableBranchText(@NotNull GitRepository repository)
     }
 
     GitBranch branch = repository.getCurrentBranch();
-    String branchName = (branch == null ? "" : branch.getName());
+    String branchName;
+    if (branch instanceof GitSymbolicRef) {
+      branchName = (branch == null ? "" : ((GitSymbolicRef)branch).getPrintableName());
```
klikh: Looks reasonable to propagate `getPresentableName()` (which I feel is a better name for the method `getPrintableName()`) to `GitRef`. Default implementation would delegate to `getName()`, while `GitSymbolicRef` would provide an alternative presentation.

And there will be one less instanceof&cast ;)

plugins/git4idea/src/git4idea/ui/branch/GitBranchPopupActions.java
```diff
@@ -27,6 +27,8 @@
 import com.intellij.openapi.util.Condition;
 import com.intellij.util.containers.ContainerUtil;
 import git4idea.GitBranch;
+import git4idea.GitLocalBranch;
```
klikh: These changes can be reverted.

plugins/git4idea/src/git4idea/branch/GitBranchUtil.java
```diff
@@ -75,8 +75,9 @@ private GitBranchUtil() {}
    */
   @Nullable
   public static GitBranchTrackInfo getTrackInfoForBranch(@NotNull GitRepository repository, @NotNull GitLocalBranch branch) {
+    GitRemoteBranch remoteBranch = branch.findTrackedBranch(repository);
```
klikh: This looks a bit weird, since `branch.findTrackedBranch` walks through `getBranchTrackInfos` inside.

Btw, I've roughly checked the usages, and it seems that all of them will be OK to receive a `GitRemoteBranch` object (which luckily has `getRemote()`) instead of the whole `GitTrackInfo`.

I propose to use `branch.findTrackedBranch` in the push dialog and other usages. I propose to check which usages are affected by symrefs and fix them, leaving other as is, and I'll fix them later.

plugins/git4idea/src/git4idea/repo/GitRepositoryReader.java
```diff
@@ -492,10 +516,58 @@ private static boolean duplicateEntry(@NotNull Map<String, Hash> resolved, @NotN
...
+
+    boolean isSymbolicRef() {
+      return symbolicRefName != null;
+    }
+  }
+
+  private static class RefInfo {
+    private final String refName;
+    private final Hash hash;
+    private final boolean isSymbolicRef;
```
klikh: The boolean flag `isSymbolicRef` is not needed: it would be more reliable to declare a method 
```
boolean isSymbolicRef() { 
  return myRefencedRef != null; 
}
```

plugins/git4idea/src/git4idea/repo/GitRepositoryReader.java
```diff
@@ -442,14 +462,18 @@ else if (!resolved.containsKey(link)) {
...
+  }
+
   @NotNull
-  private static Map<String, Hash> getResolvedHashes(@NotNull Map<String, String> data) {
-    Map<String, Hash> resolved = ContainerUtil.newHashMap();
+  private static Map<String, RefInfo> getResolvedHashes(@NotNull Map<String, String> data) {
+    Map<String, RefInfo> resolved = ContainerUtil.newHashMap();
     for (Map.Entry<String, String> entry : data.entrySet()) {
       String refName = entry.getKey();
       Hash hash = parseHash(entry.getValue());
```
klikh: This is not a problem of this PR (since the problem already exists), mentioning here just for reference (mostly for myself): this method is potentially buggy, since it will treat a branch name matching `[a-zA-Z0-9]` as a hash.

plugins/git4idea/src/git4idea/repo/GitRepositoryReader.java
```diff
@@ -253,6 +257,12 @@ else if (refName.startsWith(REFS_REMOTES_PREFIX)) {
     return Pair.create(localBranches, remoteBranches);
   }
 
+  private static GitLocalBranch fromRefInfo(RefInfo refInfo) {
+    return refInfo.isSymbolicRef ?
+           new GitSymbolicRef(refInfo.refName, new GitLocalBranch(refInfo.referencedRef)) :
```
klikh: Here a new `GitLocalBranch` is created instead of reusing existing one. I guess this is not a problem though, since these GitLocalBranch objects are equal, but it would be more clear to reuse the existing branch object, e.g. first by creating objects for non symrefs, and then for symrefs.

aronzhina:
Hey @illabefat, 
It’s been quite a while since this PR was opened 🙂 Just wanted to check if you’re still interested in finishing it, or if we can close it

PR #431, opened 2016-07-22
IDEA-157902 FIX - Duplicated git items in "Find Action"

vvv1559:


klikh:
Thanks for the pull request. However, it looks that there is a better fix for the issue with "Open on GitHub" actions. Instead of adding an unnecessary word to the action title, we merged two actions into a single action: 5c5db8071593186ff51433b4cb7ff2b0d002ba9d

As per "Git" action group, I don't think that naming one of the groups "Git Actions" will solve the problem. It will help with duplication, but won't help with confusion. Unfortunately, currently it is not clear what can be done with these groups. It might help if a hint would be displayed at the right of each group, as it is done for some other actions.

vvv1559:
Kirill, thanks to your response. Merge two actions - best way! About naming and hint - I understood this way and try implement this posibility.

PR #432, opened 2016-07-28
Fix PY-20227,PY-20251,PY-20332,PY-20468

fitermay:


python/src/com/jetbrains/python/debugger/containerview/NumericContainerViewTable.java
```diff
@@ -82,12 +82,15 @@ private void initUi(@NotNull final ArrayChunk chunk, final boolean inPlace) {
       myComponent.getSliceTextField().setText(chunk.getSlicePresentation());
       myComponent.getFormatTextField().setText(chunk.getFormat());
       myDialog.setTitle(getTitlePresentation(chunk.getSlicePresentation()));
+      boolean shouldSetColored = myTableCellRenderer == null || myTableCellRenderer.getColored();
```
traff: I probably miss something, but could you please explain what this condition for?

fitermay: It is to propagate previous colored state. This function is called on initial load , in which case myTableCellRenderer is null and we use the default value or user picked a new slice, in which case we use prev setting. This is is for 20251.

PR #437, opened 2016-08-11
fix debuggability for ipy notebook kernels

complyue:
the ipy nb kernels are subprocesses launched with `-m ipykernel`, while pydev generally has problems in debugging -m startups, the simple modification in this PR has enabled me to debug ipy kernels with PyCharm.

It exploits the fact that `__main__` of ipykernel does not fire relative imports as `__init__` does. It wouldn't work without this fact.

However I'm not sure whether running `__main__` instead of `__init__` for a module will break things in other cases.

### DIFF
--- a/python/helpers/pydev/pydevd.py
+++ b/python/helpers/pydev/pydevd.py
@@ -901,6 +901,10 @@ class PyDB:
                 sys.stderr.write("No module named %s\n" % file)
                 return
             else:
+                if filename.endswith('__init__.py'):
+                    new_target = os.path.join(os.path.dirname(filename), '__main__.py')
+                if os.path.isfile(new_target):
+                    filename = new_target
                 file = filename
 
         if os.path.isdir(file):


klikh:
Is compl.yue@icloud.com your valid e-mail address? I'd like to send you the contributor license agreement to sign, unless you've already done it.

complyue:
Yes, this is my main email address. I haven't signed the LA.

traff:
In the current state the pull request can't be merged as it will break the execution of __init__.py files in general.

Could you please describe what problem you are solving? 
It's better to do that by submitting an issue to our tracker and posting here a link to it:
https://youtrack.jetbrains.com/issues/PY

complyue:
Seems there's already one: https://youtrack.jetbrains.com/issue/PY-14742

complyue:
@traff I'm just curious in what cases should `__init__.py` be executed instead of `__main__.py` for a package when both exist?
I had tried `-m` for `Interpreter options` and package name for `Script` in a run/debug configuration, while it runs okay but debug always fail.

p-i-:
Does this mean there is potential movement on this issue?

If we could only:
- set breakpoints
- single-step through code
- interact with the local Environment from the REPL

... this would at least change the face of scientific computation.

As far as I can see, this is the main reason MATLAB is still able to compete in the marketplace.  Also Julia is coming, but it currently lacks this ability also.

PR #439, opened 2016-08-29
Implement parameter info in Console (PY-9939, PY-18586)

fitermay:


python/helpers/pydev/_pydev_bundle/_pydev_calltip_util.py
```diff
@@ -0,0 +1,160 @@
+'''
+Id:          "$Id$"
+Copyright:   Copyright (c) 2015 Bank of America Merrill Lynch, All Rights Reserved
```
traff: I don't think this copyright is valid as the contribution is covered by CLA that transfers the copyright to JetBrains and assignes the Apache 2.0 license to the code.

traff: Actually our CLA doesn't transfer the copyright it only grants the license, but it was you, Yuli Fitterman, who signed the CLA, not Bank of America, so we assume that you are the copyright owner in this case.

fitermay: Please change the header. I had my IDE at home accidentally set to insert
this.

traff:
I've merged that in.

PR #441, opened 2016-09-02
[PyCharm] Fix python crash in module Qt5Gui.dll

asidonus:
Current skeleton generation cause application crash in module Qt5Gui.dll with exception 0xc0000005 when it called for modules "PyQt5.Qt" and "PyQt5.QtGui".

This patch fix reason of this crash.


klikh:

vlasovskikh:
Closing this one since we haven't got a signed contributor agreement from @asidonus.

PR #442, opened 2016-09-03
IDEA-160742: Ctrl+Shift+F12 doesn't work in Terminal tool window. 

sesm:
Add 'HideAllWindows' to a whitelist of actions for Terminal tool window.


traff:
Merged.

sesm:
Great, thanks!

PR #444, opened 2016-09-12
Better Python Console

fitermay:
Implement proper multi-line console for PyCharm.


fitermay:
This is a work in progress. Submitting a PR now for JB to keep track

Elizaveta239:
Thank you for these changes, the code looks nice. Just a few minors after a short testing:

1. It would be nice if Python Console was available on the status bar on the IDE start and it didn't disappear after pressing the red cross.
2. I've tried to open a few tabs of the Console and than I've closed all of them (with the red cross). During the next PyCharm session all these tabs were opened on the IDE start.
3. Minor: the title "Python Console" is duplicated on the tool window.

fitermay:
1) It should be available on the status bar by default unless you made it disappear the last session.  I can make it stay on the toolbar once you press the red cross. 
2) This is really strange and I cannot reproduce this.  How long did you wait before closing the project? Can you wait 5 seconds and see if the issue persists? What if you save the project before exiting? 

Could you please send me a screencast of how to reproduce these issues? I cannot reproduce them on my machine.

Elizaveta239:
1) Yes, let's make it stay on the toolbar all the time, even after pressing the red cross.
2) I can reproduce it from time to time, not always. I've recorded a screen cast and I've sent it to you by mail.

fitermay:
The last commit should fix PY-13029. There are some other issues I see like connection being closed on UI thread I want to fix here before I go ahead with the actual implementation of the console code...

Elizaveta239:
@fitermay Thank you for your fixes and improvements! I'm not able to reproduce my phantom bug anymore.

A small, but important reminder: if you fix some bug, please, find its id in our tracker and add it to your commit message. If you can't find it, feel free to create it. It will help us to distinguish refactoring commits and bug fixing commits. Also it will make a testing process easier. Thank you!

platform/lang-impl/src/com/intellij/execution/runners/AbstractConsoleRunnerWithHistory.java
```diff
@@ -147,29 +141,7 @@ protected Icon getConsoleIcon() {
   }
 
   protected String constructConsoleTitle(final @NotNull String consoleTitle) {
-    if (shouldAddNumberToTitle()) {
```
Elizaveta239: Why have you move such a small and not important function to a separate class? Are there any reasons for that?

fitermay: 1) Yes,  absolutely. I was planning to rebase and add issue numbers to the commit messages when done
2) I did moved 'constructConsoleTitle' to a seperate class to get rid of the inheritance from 'AbstractConsoleRunnerWithHistory' . It's actually a set of 3 methods "constructConsoleTitle", "shouldAddNumberToTitle",  and "getActiveConsoles".

fitermay:
Ok, this completes what I had planned for this PR. At some point I'd like to implement more enchancements such as a multiplexing console that switches between debug console when process is suspended and regular console when running. These will be submitted as a different PR when I have time to work on them.

Elizaveta239:
Thank you very much for your job! I've found some bugs, please, look at them: PY-20848, PY-20847, PY-20849, comment in PY-20836.
Also I've created a tag "PR 444" for all issues related to your pull request.

fitermay:
PY-20847: As designed. It waits for indent to be 0 to execute. Compare with IPython and IDLE console -- they work the same way
PY-20848,20849,: Fixed 
Comment in PY-20836: Yes, it's a bug. Fixed.

fitermay:
Oops. Never mind. IPython also executes on blank lines.Fixed

traff:
Hi Yuli, this was huge!
Thanks for the contribution.
The changes were tested and seem to be working quite well.

Though I have some questions regarding your design decisions: why did you choose to implement it with Executor and Startup activity instead of normal ToolWindow like it was before?

traff:
I've reverted the implementation from Executors to ToolWindow API as it was before. All other fixes were preserved;

fitermay:
Does rerun action still work properly after the revert? I relied on the reuse content behavior of RunContenrManager to get restart action to work correctly. I'll try to explain more when I have some time today.

fitermay:
Ok, this is why I chose to get rid of the Toolwindow.

From general considerations:
1) Increase code reuse. Gets rid of the code PythonToolWindowConsoleRunner, PythonToolWindowConsoleRunnerFactory, PythonConsoleToolWindow, and most importantly simplifies PyExecuteSelectionAction. In general allows handling toolwindow and run window console to be handled by the same APIs.
2) Makes it easier to test -- since only one code path is behavior is the mostly the same in PyCharm and the Python plugin
3) I consider RunContentManager a more solid foundation since it is used in all products and has been tested thoroughly. 

Specific reasons:
1) The reuse content behavior of RunContentManager makes it simpler to implement rerun correctly.
2) The complexity of micromanaging the toolwindow leads to actual bugs -- off the top of my head NPE PyExecuteSelectionAction line 226. There are other issues I found but can't recall off the top of my head.
3) I plan to implement other actions in the future like "execute module in console", "run tests in console", etc. I know from prior experience that these are easier to implement with only one API to run the console.

python/src/com/jetbrains/python/console/PythonConsoleView.java
```diff
@@ -293,10 +300,78 @@ public void inputRequested() {
...
+      //todo should really be myPromptView.getPromptAttributes() output type
+      //but in that case flushing doesn't get handled correctly. Take a look at it later
+      print(prompt + " ", ConsoleViewContentType.USER_INPUT);
+    }
+
+    highlighter.putUserData(PyConsoleCopyHandler.PROMPT_LENGTH_MARKER, prompt.length() + 1);
+  }
+
+  @NotNull
+  protected String addTextRangeToHistory(@NotNull TextRange textRange, @NotNull EditorEx inputEditor, boolean preserveMarkup) {
```
fitermay: Yes, it is a copy paste with some modification.  I haven't found a way to avoid copying without doing things **worse than copying**.  I consider labyrinthine inheritance hierarchies worse than copying, for example. YYMV.

traff:
Yes, your points are valid and your implementation worked well. The problem is that extending Executor to support console is bit of a API misuse. Toolwindow API was designed exactly for this purpose. So it we think that it doesn't work as it should, we just need to fix it instead making another extension point.

Also there was a tiny problem that the startup activity was launched later then tool window intitializers which resulted into Python Console tool window appear later then other tool windows in the the corresponding panel.  

Anyway now the code look simpler then before there is only one PythonConsoleToolWindow(with it's factory) which reflects PythonConsoleExecutor with a startup activity. 

As for the rerun - it works fine now.

Yes there is now exception which happens sometimes on execute line action due to some async misconfiguration, but I am going to fix it for the next release.

fitermay:
Yes, my main concern not so much duplication but having to work with two different APIs for handling execution in console toolwindow vs run toolwindow. Maybe we can think of a way to converge them and also make the toolwindow API smarter...

PR #447, opened 2016-09-18
IDEA-161410: Quick Documentation: double opening angle brackets are printed as single brackets

stefanneuhaus:
- Repeat replacement on the result string, to cover the case with "overlapping patterns"
- Do not append additional "<" (as was before, but seemed to be ignored by the HTML view pane)


klikh:

stefanneuhaus:
Closing as it has been fixed meanwhile: [IDEA-161410](https://youtrack.jetbrains.com/issue/IDEA-161410)

PR #449, opened 2016-09-24
PY-20771: Fix dataframe rendering issue in frames 100+ columns

fitermay:
PY-20771: Min and max values for columns 100 and above were computed incorrectly sometimes causing exceptions during rendering.
The UI exceptions are also one of the reasons for poor performance


medvector:
Hi Yuli!
Thank you for the fix. I've just merged it to master.

PR #452, opened 2016-10-12
Update IDE setup instructions

lhotari:
- sd-jdi.jar is also required besides tools.jar


chashnikov:
Why do you think that sd-jdi.jar is required? Could you please describe what exact problem is caused by its absence?

lhotari:
I checked again and you are right, it's not needed. It must have been a coincidence that compilation started working after adding `sd-jdi.jar`. The reason I added it was because compilation was failing because not finding some `com.sun.jdi.*` class, but those are already in `tools.jar` so it has nothing to do with the problem I had. I had added `tools.jar`, but I kept getting the error, but I must have done something wrong my self.

PR #453, opened 2016-10-12
Fix memory leak inside the Gradle daemon caused by Idea Gradle plugin

lhotari:
This fixes a memory leak inside the Gradle daemon caused by Idea Gradle plugin.

Here's a description of the memory leak:
The custom classloader created by `ijinit.gradle` was not garbage collected and it was keeping references to a lot of object instances. The instances were held in the static `ExtraModelBuilder.buildersLoader` field. This field referenced an instance of `ExternalProjectBuilderImpl` which has a `cache` field with a lot of instances. 
Each new call using the same Gradle daemon instance would leak more memory.


lhotari:
@vladsoroka Could you take a look at this PR?

lhotari:
I prefixed the commit messages with `Gradle: ` and re-pushed the commits to my PR branch.

vladsoroka:
PR merged, thanks for the contribution Lari!

PR #455, opened 2016-10-21
Bugfix for migration to Stream API for sum() of List<Integer>

FHannes:
This commit is a patch for an incorrect refactoring of a foreach loop over a List<Integer> to a sum() call on a Stream.
When the loop variable's type is a primitive int rather than an Integer, the current implementation did not add a mapToInt() call to obtain a specialized IntStream object first.

Example:
--------
```
List<Integer> values = ...
int sum = 0;
for(int i : values) {
  sum += i;
}
```

The code above would refactor to:
`int sum = values.stream().sum();`

The correct refactoring is:
`int sum = values.stream().mapToInt(i -> i).sum();`


klikh:

FHannes:
I have signed the agreement.

klikh:
Received, thank you.

amaembo:
Thank you for discovering this bug and providing a pull-request. Indeed this case was not covered. Unfortunately the way you fixed this is not correct. First, it breaks primitive arrays iteration. For example:

~~~~java
  public int sum(int[] array) {
    int sum = 0;
    for(int x : array) {
      sum += x;
    }
    return sum;
  }
~~~~

This case should be converted to `Arrays.stream(array).sum()`, but after applying your patch, it adds `mapToInt(x -> x)` step which is incorrect. Also it does not cover more complex problems. For example:

~~~~java
public class Main {
  public boolean check(Integer x) {
    return x % 3 == 0;
  }

  public boolean check(int x) {
    return x % 2 == 0;
  }

  public int sum(List<Integer> list) {
    int sum = 0;
    for(int x : list) {
      if(check(x))
        sum += x;
    }
    return sum;
  }

    System.out.println(new Main().sum(Arrays.asList(1,2,3,4,5,6)));
  }
}
~~~~

Here `check(int)` is called. After applying your fix we have compilable code, but it adds `mapToInt` *after* `filter` which silently changes the code semantics (calls `check(Integer)` now). What we should do is to modify the CollectionStream code generation adding `mapToInt` there. I decided that it would be easier to decline your pull request and fix this by myself in a proper way (see d51f09d0d54dd01f35e2556a3b52afd161c428fb). Nevertheless I really appreciate that you've found a bug in my code. Thank you and looking forward for new pull requests from you!

FHannes:
Alright, that's fine, I'm still in the process of learning how everything in the stream migration system operates. Thank you for the detailed overview!

PR #457, opened 2016-10-26
LightCodeInsightFixtureTestCase: override hook replacing tempDirFixture

yaronyam:
Add an override hook to initialize an instance of `TempDirTestFixture` in Test classes derived from `LightCodeInsightFixtureTestCase`. By default it uses an in-memory implementation - `LightTempDirTestFixtureImpl`. Would be nice to be able to replace it with a persistent filesystem implementation - `TempDirTestFixtureImpl`.

### DIFF
--- a/java/testFramework/src/com/intellij/testFramework/fixtures/LightCodeInsightFixtureTestCase.java
+++ b/java/testFramework/src/com/intellij/testFramework/fixtures/LightCodeInsightFixtureTestCase.java
@@ -108,13 +108,23 @@ public abstract class LightCodeInsightFixtureTestCase extends UsefulTestCase {
     IdeaTestFixtureFactory factory = IdeaTestFixtureFactory.getFixtureFactory();
     TestFixtureBuilder<IdeaProjectTestFixture> fixtureBuilder = factory.createLightFixtureBuilder(getProjectDescriptor());
     final IdeaProjectTestFixture fixture = fixtureBuilder.getFixture();
-    myFixture = JavaTestFixtureFactory.getFixtureFactory().createCodeInsightFixture(fixture, new LightTempDirTestFixtureImpl(true));
+    myFixture = JavaTestFixtureFactory.getFixtureFactory().createCodeInsightFixture(fixture, getTempDirFixture());
 
     myFixture.setUp();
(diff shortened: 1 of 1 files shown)


klikh:
The Contributor License Agreement has been received. Thanks.

donnerpeter:
Why do you want to override the temp dir fixture?

yaronyam:
I would like to have the test case project files stored to disk. It helps in understanding the tested project structure, validating assumptions and problem solving.

donnerpeter:
Have you tried CodeInsightFixtureTestCase? It does exactly that, although it's slower (because of more disk operations).

yaronyam:
I'm using `LightCodeInsightFixtureTestCase` since it supports Java projects, while  `CodeInsightFixtureTestCase` doesn't

donnerpeter:
Indeed. Have you tried `JavaCodeInsightFixtureTestCase`?

yaronyam:
That does the trick. Thanks for the tip Peter
One issue though, according to the [docs](http://www.jetbrains.org/intellij/sdk/docs/basics/testing_plugins/light_and_heavy_tests.html) - the ligth version is more effiecient since projects are reused when possible between tests. I guess this will not happen with `JavaCodeInsightFixtureTestCase`?! is it possible to use `LightCodeInsightFixtureTestCase` with `TempDirTestFixtureImpl` - so files will be persisted to disk while still re-using project files between tests? If so then this PR might be relevant, otherwise it can be closed.

donnerpeter:
No, JavaCodeInsightFixtureTestCase doesn't reuse the project. It'd be more expensive and fragile when working with physical disk.

If you don't absolutely need the disk for tests themselves (this can happen if they run some external process on the project files), I'd recommend to start with JavaCodeInsightFixtureTestCase, do the "understanding the tested project structure, validating assumptions and problem solving" and then switch superclass to use light tests to get all the speed improvements.

yaronyam:
Ok, makes sense. I guess this PR can be closed. Thanks Peter

yaronyam:
I would like to have the test case project files stored to disk. It helps
in understanding the tested project structure, validating assumptions and
problem solving.

PR #459, opened 2016-10-27
IDEA-141243 added support for ant-contrib try/catch-task

zdary:
when exception is thrown in try block it'll get logged as verbosive message


eugenezh:
Merged into master with some changes, thanks!

PR #462, opened 2016-10-28
PY-18144 Support Pycharm import from zip files.

Ahuge:
PY-18144 Update. In response to Ilya Kazakevich's response, I have submitted my patch and created a PR.

We are falling back to importlib to load a module from within a zip file. Imp does not support this.

### DIFF
--- a/python/helpers/pycharm/pycharm_run_utils.py
+++ b/python/helpers/pycharm/pycharm_run_utils.py
@@ -33,7 +33,7 @@ def import_system_module(name):
   try:
     f, filename, desc = imp.find_module(name)
     return imp.load_module('pycharm_' + name, f, filename, desc)
-  except:
+  except ImportError:
     # Hack for python files in a zip file. Imp doesnt work correctly in it.
     import importlib
     mod = importlib.import_module(name)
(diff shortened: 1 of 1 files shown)


klikh:

python/helpers/pycharm/pycharm_run_utils.py
```diff
@@ -30,8 +30,15 @@ def adjust_django_sys_path():
 def import_system_module(name):
   if sys.platform == "cli":    # hack for the ironpython
       return __import__(name)
-  f, filename, desc = imp.find_module(name)
-  return imp.load_module('pycharm_' + name, f, filename, desc)
+  try:
+    f, filename, desc = imp.find_module(name)
+    return imp.load_module('pycharm_' + name, f, filename, desc)
+  except:
```
throwable-one: Is it possible to except certain error?

Ahuge:
Updated with a more specific exception clause.

Should I be logging the exception that I am catching?

Ahuge:
What do I need to do to get this pulled in?

wangjoshuah:
BUMP

PR #465, opened 2016-11-05
 Expose finding class files for alternative bytecode viewers

ingokegel:
The logic of finding the class file that is associated with a PsiElement is non-trivial and can be reused by other bytecode viewers with this simple change.

Alternative bytecode viewers want to depend on the built-in bytecode viewer anyway because of the classSearcher extension point.

### DIFF
--- a/plugins/ByteCodeViewer/src/com/intellij/byteCodeViewer/ByteCodeViewerManager.java
+++ b/plugins/ByteCodeViewer/src/com/intellij/byteCodeViewer/ByteCodeViewerManager.java
@@ -27,8 +27,8 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.CompilerModuleExtension;
 import com.intellij.openapi.roots.ProjectFileIndex;
 import com.intellij.openapi.roots.ProjectRootManager;
-import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.*;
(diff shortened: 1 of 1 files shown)


klikh:

ingokegel:
I just signed the CLA

klikh:

akozlova:
How the alternative code viewers are integrated? What API they use? Thanks

ingokegel:
I'm currently writing a plugin for jclasslib:
https://github.com/ingokegel/jclasslib

The action and the display are completely different from your plugin, but I
still need to search for the class file for the current selection and make
use of the the classSearcher extension point. Right now, I need to
duplicate your code for finding the class file.



2016-11-08 14:56 GMT+01:00 Anna Kozlova <notifications@github.com>:

akozlova:
So you would have strong dependency on a plugin where the only implementation of mentioned extension point is groovy scripts. I'd rather suggest to move the code outside the plugin, based on containing class. You'll be able still to provide optional dependency on plugin for extensions but even without the bytecode viewer your plugin would work for most of cases. Agree?

ingokegel:
Yes, that makes perfect sense and would be much better.

2016-11-08 15:56 GMT+01:00 Anna Kozlova <notifications@github.com>:

akozlova:
I'll close the pull request. Please check my commit 92ad8f3bd4c580cc81184b2a03842a6fbf4eb241, it should be available on public github later today. Thanks

ingokegel:
Thank you very much, this looks good, I'll use it after 2016.3 is released.

akozlova:
That's in master now, 171.+. The change looks simple but still can't be put to the release branch as we are too close to the release date. Sorry

ingokegel:
OK, got it, no problem.

PR #466, opened 2016-11-07
Add a custom python package identifier extension point

brendandouglas:
This allows bypassing the default 'is package' check in PyUtil, giving the option of foregoing __init__.py files prior to python 3.3.

We're writing a plugin allowing python development with IntelliJ at Google. The only blocker is the requirement for __init__.py files through the source tree, which we can't practically enforce.

This patch should have minimal risk or performance cost, and would provide a big benefit to our internal users.

### DIFF
--- a/python/psi-api/src/com/jetbrains/python/psi/PyCustomPackageIdentifier.java
+++ b/python/psi-api/src/com/jetbrains/python/psi/PyCustomPackageIdentifier.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2000-2016 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.jetbrains.python.psi;
+
+import com.intellij.openapi.extensions.ExtensionPointName;
+import com.intellij.psi.PsiDirectory;
+import com.intellij.psi.PsiFile;
+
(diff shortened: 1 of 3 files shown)


brendandouglas:
Hi there, how does this sound? I believe it poses minimal risk, and allows python imports without corresponding init.py files.

It's unlikely to be widely useful, but would help us a lot. Thanks!

tomlu:
Hi guys, Tom here, lead of the Google internal IntelliJ team.

Would it be at all possible to entertain this pull request? While the pull request seems safe and small, we would totally respect a rejection as it's very likely not needed by anyone but us. However we would love a decision or comment from you, as we would in that case at least know and probably move to building a custom plugin ourselves. Thanks!

traff:
Hi guys! Sorry for the delayed response - we were busy with the new 2016.3 release. I can accept the pull request. 
After another collegue reviews it, I'll merge it in.

vlasovskikh:
@brendandouglas I'm generally OK with this change. Could you please add querying your extension point to the other `isPackage(PsiFile)` as well?

brendandouglas:
@vlasovskikh thanks! I'm happy to add another method to the extension point, and query that from `isPackage(PsiFile)`. Could overriding the behaviour of that method have unintended consequences though?

As I understand it, it's not used to determine whether a directory is a python package (at least, not without also checking the containing directory, so it doesn't need to be bypassed for python 3.3+), only whether a file should be treated as an `__init__.py` file. Are there use cases where different files should be treated as `__init__.py` files?

vlasovskikh:
@brendandouglas There are some Python dialects that have different conventions for package files. For example, in Cython a package can be marked by `__init__.pxd`, for Python type hinting stubs it can be `__init__.pyi`, and so on.

If you don't use custom files like these for marking a directory as a package then you would just return `false` in `YourExtension.isPackage(PsiFile)`.

brendandouglas:
@vlasovskikh sounds good -- I've updated the pull request to include that method.

vlasovskikh:
@brendandouglas Thanks! I've rebased your commit onto master, it will appear on GitHub in a few hours.

brendandouglas:
Thanks @vlasovskikh. Is there any chance this could make it into a 2016.3 dot release?

vlasovskikh:
@brendandouglas It will be included into EAP versions of 2017.1.

PR #468, opened 2016-11-22
Fix test that catches AssertionError

dahlstrom-g:
… from its own call to TestCase.fail.


dahlstrom-g:
This follows up commit cd7477b by cdr@jetbrains.com.

akozlova:
I've converted test to junit 4 and use @Test(expected=AssertionError.class) instead.

PR #471, opened 2016-11-22
Make plugins/svn4idea/resources a resources root

dahlstrom-g:
… instead of a sources root, since it contains only PNG icons.

This follows up commit cc7588b1 by Olesya.Smirnova@jetbrains.com.


dahlstrom-g:
@pchel-?

pchel-:
I've applied the fix as part of the commit above.

PR #472, opened 2016-11-23
Fix mismatching varargs/array params in overrides

dahlstrom-g:
Maybe @akozlova would do the honors of reviewing this one? ;-)


akozlova:
I've merged 3 files. ArrayConsumer was not changed as I can imagine external usages with varargs call. Our internal overrides are different - some use array, some varargs and it's impossible to make super class to use varargs (varargs are not allowed in type arguments), so I prefer to leave it as is. If it's a problem, we can suppress it in the code just say how to do that. Thanks for the contribution!

PR #473, opened 2016-11-28
Make platform/dvcs-impl/resources a resources root

dahlstrom-g:
… instead of a sources root, since it contains only icons.


klikh:
Merged manually. Thanks.

PR #476, opened 2016-11-29
Test Comparable IteratorPosition using assertThat

dahlstrom-g:
… which makes failure messages (and maybe assertion code) more readable, and avoids http://errorprone.info/bugpattern/SelfComparison.


cdracm:
fixed

PR #481, opened 2016-12-01
Replace assertTrue(stable.equals(stable))

dahlstrom-g:
… with assertEquals(stable, stable), to avoid inspection EqualsWithItself and http://errorprone.info/bugpattern/SelfEquals.


klikh:
Merged manually. Thanks!

PR #483, opened 2016-12-06
Added additional method to BudledKeymapProvider extension point to be…

da1z:
… able to load keymaps from InputStream

Need this because of if you are developing plugin using gradle-intellij-plugin it wont find keymap by file name

### DIFF
--- a/plugins/git4idea/gen/icons/Git4ideaIcons.java
+++ b/plugins/git4idea/gen/icons/Git4ideaIcons.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2013 JetBrains s.r.o.
+ * Copyright 2000-2016 JetBrains s.r.o.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


yole:
How exactly are you specifying the name? If the keymap .xml file is correctly packaged in the plugin jar and if you specify its name correctly, the existing APIs are sufficient to load it. If not, you need to fix either your build script or your BundledKeymapProvider implementation. There is no need to extend the API.

da1z:
i put my keymap.xml into resources/idea/ and my BundledKeymapProvider looks like

```
List<String> getKeymapFileNames() {
 return Collections.singletonList("keymap.xml");
}
```

and DefaultKeymap.java can not locate my keymap

da1z:
The keymap loader uses the class loader of the URLUtil class, so it never delegates to the plugin classloader. So we either need to return a stream or else be able to specify the classloader to use in order for this extension point to work inside a plugin.

yole:
This is indeed a problem that can be fixed, but it doesn't require adding any new APIs. It's possible to load the keymap .xml file from the classloader of the BundledKeymapProvider implementation class.

da1z:
How? i can't find API for this. please give me a few lines example

da1z:
Sent a new pull request. Please check this out https://github.com/JetBrains/intellij-community/pull/493

PR #484, opened 2016-12-06
Address FieldAccessor.get compilation failure in JDK 9

dahlstrom-g:
The first commit replaces uses of the method, which is all we apparently need at the moment to compile with JDK 9. The second commit removes the method—separate because I'm not sure if it needs to be kept in case of uses outside the project.


dahlstrom-g:
@sevdokimov? @vladsoroka?

PR #486, opened 2016-12-07
make console patcher work

jonnyzzz:
Current implementation of consoleUI for patcher did not check ANY errors and conflicts that may be found during the process. Moreover, it used to skip all error and apply changes as-is

The patch provides a logging and error reporting for that case. 

I failed to find integration tests for console running patcher, so tests were not added to check this case


jonnyzzz:
Only last two commits are actually the fix. The rest are here my mistake

klikh:
@jonnyzzz Please rebase against current origin/master and remove commits that are not needed. Thanks.

jonnyzzz:
This branch was merged by @bulenkov. Gonna close it now

PR #489, opened 2016-12-08
Use existing jsr305 project-level library

dahlstrom-g:
… instead of module-library in gradle module. Also remove identical redundant jsr305-1.3.9.jar and associated library.


dahlstrom-g:
@denis-zhdanov?

PR #492, opened 2016-12-12
Make RunContentManagerImpl.DESCRIPTOR_KEY public

ingokegel:
Rationale: I'm writing a plugin for a profiler integration which defines its own executor. Apart from the live profiling runs that are associated with a console, I want to add other tabs for snapshots and attach sessions to the tool window that have a different layout without any console. `Content` objects in the tool window need a user data value with the above key, otherwise an exception is thrown.

Current workaround: I'm using `Key.findKeyByName("Descriptor") as Key<RunContentDescriptor>`, but that is deprecated.


Vassiliy-Kudryashov:
I've moved DESCRIPTOR_KEY into RunContentDescriptor (master branch, 2017.1 EAP)

ingokegel:
Thank you, that's even better!

PR #495, opened 2016-12-15
i18n contributions for TogglePresentationModeAction, ToggleDistractionFreeModeAction, TabsPlacementAction

smmribeiro:
There're several Actions that are not prepared for localization and/or have their messages hard-coded at declaration.
I've tried to follow how similar classes are implemented to fix some of these cases.
If this pull request is accepted, I'll try to continue to check other cases like these ones.


smmribeiro:
Hi,

As this is my first pull request here, I don't really know how the process is.
Is there anything that I can, or must, do for it to be approved faster?

Basically:

- for class **TabsPlacementAction**, I've removed the hard-coded texts that existed in **PlatformActions.xml** and made the appropriated changes to the class so that text and description to be used came from a resource bundle - in this case from **ActionBundle.properties** (where I've created the new tags).

- For class **ToggleDistractionFreeModeAction** there're strings hard-coded in the class, so they were moved to **ActionBundle.properties**' resource bundle as above. As a minor optimisation to this class, two strings were refactored and created as constants (static final) to avoid multiple instantiations.

- For class **TogglePresentationModeAction**, the change was only to make two strings constant (static final).

To test, you just need to change the tags you want, run IntelliJ and see the new texts/descriptions on the affected options. 

Regards,

Sérgio Ribeiro.

bulenkov:
Hello Sérgio.

Sorry for being silent for so long. There were reasons for that. I'm just back to work and reviewed your patch. Thank you for reporting this problem. It is definitely an issue. However, I know a better solution for the fix and am going to push it soon. Already done it for Tab Placement actions. You can take a look here 8829e112cfd7c3275e05cc64c899c2fa1e0c8eb2 
Main idea is to stay java code as compact as possible. So, I moved everything into the bundle.
I'll let you know when finished.

Thank you very much for collaboration.

Konstantin

smmribeiro:
Hello Konstantin,

First of all, thank you for the update. Secondly, sorry for my delay in replying - the end of sprint at work is always a busy time...

I've already checked your commit and it's much more simple and beautiful than my code :-) I had coded that way because I had found a similar scenario on JetBrains code; I didn't know one could do that for inner classes. Could you point me to where that functionality (the "automatic" finding/calculation of those resource properties) is done?

As I'm working on this at my work (although we're using ICE 2015), I would like to contribute more on i18n questions so that the problems we had on this version won't repeat (or, at least, there're less) when we advance to 2016+. For instance, looking into PlatformActions.xml, we can see several other "text" properties there and that is a problem when one wants to have those actions in another language... Would you be interested in that kind of contribution?

Regards,

Sérgio Ribeiro
Porto - Portugal

smmribeiro:
Hello again,
I saw that classes ToggleDistractionFreeModeAction and TogglePresentationModeAction were also updated. The main difference to my code is the not use of a static final variable for the texts

PR #497, opened 2016-12-21
IDEA-165738 Gradle import is very slow in multi-module projects

ingokegel:
The gradle import resolves source and javadoc JAR files for all artifacts, including project dependency artifacts. These artifacts do not exist in external repositories and so gradle makes many unsuccessful HTTP calls to pom files as well as source and javadoc JARs across all configured repositories every time the project is imported. For large multi-module projects with many external repositories these unsuccessful HTTP calls can take a long time, on the order of many minutes.

This fix limits the resolved artifacts to those from external dependencies. I have checked that all tests in gradle-tooling-extension-impl still pass.

### DIFF
--- a/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/util/DependencyResolverImpl.groovy
+++ b/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/util/DependencyResolverImpl.groovy
@@ -136,7 +136,7 @@ class DependencyResolverImpl implements DependencyResolver {
         //noinspection GroovyAssignabilityCheck
         Set<ComponentArtifactsResult> componentResults = myProject.dependencies.createArtifactResolutionQuery()
           .forComponents(resolvedArtifacts
-                           .findAll{ !isDependencySubstitutionsSupported || it.id.componentIdentifier instanceof ModuleComponentIdentifier }
+                           .findAll { !isDependencySubstitutionsSupported || it.id.componentIdentifier instanceof ModuleComponentIdentifier }
                            .collect { toComponentIdentifier(it.moduleVersion.id) })
           .withArtifacts(jvmLibrary, artifactTypes)
           .execute()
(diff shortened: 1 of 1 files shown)


ingokegel:
With feedback from @oehme (Gradle)

oehme:
I think this change is going to break partial imports, i.e. we have a project with two subprojects Foo and Bar (Foo depends on Bar) and we only import Foo. I think the code that was changed in this commit was making sure that a version of Bar is loaded from the repository instead, so that Foo will still compile. I guess there is no test coverage for this case, so it went undetected. I might be wrong, but it's hard to tell without tests.

I think this can be improved though, using one of the following solutions:

1. Only download jar/sources/javadoc for subprojects that are *not* in the IDE. This should be a quick and easy fix and make imports a lot faster.
2. A more wholistic solution: When a subproject is *not* in the IDE, IDEA should inject a dependency substitution rule into the build to tell Gradle that this project should be fetched from the repository instead. This would make some code in the `DependencyResolverImpl` simpler and make it faster in all cases. It would be a slightly bigger change though.

ingokegel:
> we have a project with two subprojects Foo and Bar (Foo depends on Bar) and we only import Foo

This did not produce a compilable project before this change (I just checked it) and so this is a separate unrelated issue. You could argue that the artifacts might be available in a repository, but in general this is not the case for all modules and as soon as you bump the version number it would not work anymore. As far as I see, you can realistically only exclude projects that no included projects depend on.

ingokegel:
Btw, just to give a sense of scale here:  Before this change, the import time for my project (~80 modules) was 5-10 minutes, after the change it is 20 seconds (with warm daemon)

oehme:
> This did not produce a compilable project before this change

I didn't know that, thanks for checking! In that case the change is good to go imho.

ingokegel:
@vladsoroka Can you have a look at this PR when you have time? It's currently very painful for me to import my larger projects and it would be great if this could go into 2017.1. Thank you!

vladsoroka:
@ingokegel thanks for the contribution, I'll dig into the issue today

ingokegel:
Thank you very much!

praveen12bnitt:
@ingokegel  @vladsoroka  is this fix broken in the recent version of IDEA? Just after this PR got merged, gradle refresh project was super fast. But recently when i imported the same project, it became significantly slower and my guess is, this fix is somehow overridden or a similar fix need to happen somewhere else.  Any guess?

PR #498, opened 2016-12-27
Added building python debugger binaries for 2.6

da1z:
related to this issue https://youtrack.jetbrains.com/issue/PY-19216

### DIFF
--- a/python/build/build-debug-binaries/src/build.gradle
+++ b/python/build/build-debug-binaries/src/build.gradle
@@ -16,6 +16,7 @@ envs {
 
   conda "Miniconda3_64", "Miniconda3-latest", "64"
 
+  condaenv "py26_64", "2.6", "Miniconda3_64", ["cython"]
   condaenv "py27_64", "2.7", "Miniconda3_64", ["cython"]
   condaenv "py34_64", "3.4", "Miniconda3_64", ["cython"]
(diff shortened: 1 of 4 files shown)


fitermay:
@Elizaveta239 I've rebased onto master and cleaned this up. The change will not introduce any additional dependencies on your side. Python 2.6 binaries are built against the same MS SDK as Python 2.7 binaries.

vlasovskikh:
@da1z @fitermay Thanks for your pull request. Unfortunately, we won't include the debugger binaries for Python 2.6 since we consider it obsolete. Python 2.6 reached its official [end of life](https://docs.python.org/devguide/index.html#status-of-python-branches) on 2013-10-29 and in future versions of PyCharm we will drop support for it. If having compiled Cython extensions for Python 2.6 is critical for you, please consider distributing them separately from PyCharm. We could update our language check when looking for Cython extensions so PyCharm will try to load them for 2.6 from `sys.path`.

The related ticket in the issue tracker: https://youtrack.jetbrains.com/issue/PY-19215

Elizaveta239:
@da1z @fitermay Thank you for the pull request!

fitermay:
@Elizaveta239 Thanks for taking this. It's very important for us.

PR #499, opened 2016-12-27
IDEA-165963 Gradle run configurations cannot be profiled

ingokegel:
Provide a way for executors other than the debugging executor to set VM parameters that are patched into the gradle Java task. This is solved with a special key that can be set in the user data of the ExecutionEnvironment.


vladsoroka:
Changes look good, thank you! I've merged it for 2017.2 release. 
However I've also renamed field DEBUGGER_SETUP_KEY and parameter debuggerSetup => JVM_AGENT_SETUP_KEY, jvmAgentSetup.
And added the validation to avoid misusage of the api: all JVM arguments should be started with "-agentlib:", "-agentpath:" or "-javaagent:"

ingokegel:
Thank you! JProfiler 10.0.1 will be able to profile gradle run configurations with 2017.2 builds.

PR #503, opened 2017-01-16
Fix PY-21991

fitermay:
Fix PY-21991


Elizaveta239:
@fitermay Thank you very much! It works fine except cases of executing complete multi-line commands. For example, if user selected such lines:
```
if True:
    a = 1
    b = 2
```
they should be executed without any additional "Enter" pressing.

fitermay:
Good catch, we need to append \n in those cases.

fitermay:
Sorry was very busy. Amended the commit now.

PR #507, opened 2017-01-27
allow replacing items on patch application in ConsoleUI

jonnyzzz:
This fix should help fixing patches application in Toolbox app.

#ALL-1292


trespasserw:
Obsolete (discussed in the issue).

PR #508, opened 2017-02-02
Speedup gradle test output XML processing

jonnyzzz:
As discussed in the https://youtrack.jetbrains.com/issue/IDEA-167449 issue I decided to implement faster XML processing for Gradle plugin that would avoid my most pain point for tests with heavy output flow. 

Before the fix, IDEA UI was too overloaded. According to Activity Monitor IDEA process was eating even more CPU than tests worker process. You may find snapshot in the issue. 

After the fix I see It consumes less CPU wither by IDEA UI responsiveness or by fresh CPU profiling.

Current implementation I use XStreams (it was included in classpath) Hierarchical parser that uses KXML pull parser implementation. The whole processing is done within one class.

Unfortunately, I was not able to implement integration tests to measure performance boost more technically as well as to make sure performance will not be lost again. Also, it looks like there are NO tests for test runner integration at all. Just let me know if you have some correct snippet to implement more test.

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/execution/test/runner/events/XmlXpathHelper.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/execution/test/runner/events/XmlXpathHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2014 JetBrains s.r.o.
+ * Copyright 2000-2017 JetBrains s.r.o.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jetbrains.plugins.gradle.util;
+package org.jetbrains.plugins.gradle.execution.test.runner.events;
 
 import org.jetbrains.annotations.NotNull;
 import org.w3c.dom.Document;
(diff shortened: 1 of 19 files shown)


plugins/gradle/src/org/jetbrains/plugins/gradle/execution/test/runner/events/XmlXpathHelper.java
```diff
@@ -32,139 +29,202 @@
  * @since 2/20/14
  */
 public class XmlXpathHelper {
```
vladsoroka: It's no XpathHelper longer.

plugins/gradle/src/org/jetbrains/plugins/gradle/execution/test/runner/events/XmlXpathHelper.java
```diff
@@ -32,139 +29,202 @@
...
+  @NotNull
+  public String getEventTitle() {
+    return myEventTitle;
   }
 
-  public String getTestEventTest() throws XmlParserException {
-    return queryXml("/ijLog/event/test/event");
+  @NotNull
+  public String getEventOpenSettings() {
+    return myEventOpenSettings;
```
vladsoroka: Should be boolean

.idea/libraries/kxml2.xml
```diff
@@ -4,6 +4,8 @@
       <root url="jar://$PROJECT_DIR$/android/android/lib/kxml2-2.3.0.jar!/" />
     </CLASSES>
     <JAVADOC />
-    <SOURCES />
+    <SOURCES>
+      <root url="jar://$USER_HOME$/.ideaLibSources/kxml2-2.3.0-sources.jar!/" />
```
vladsoroka: it shouldn't be here

.idea/libraries/kxml2.xml
```diff
@@ -4,6 +4,8 @@
       <root url="jar://$PROJECT_DIR$/android/android/lib/kxml2-2.3.0.jar!/" />
```
vladsoroka: It seems the kxm2 library comes from (and used only in) android subsystem. If we want to use it in another subsystems I would move it to community/lib

jonnyzzz:
I have force updated commits. Issues should be resolved. We no longer need KXml2 parser as XStream XPP works faster on my tests, which I have added

vladsoroka:
Merged, thanks for the contribution!

PR #511, opened 2017-02-11
Make MethodData.methodBody() resilient to non–stub-based PsiMethod implementations, as intended

calexHG:
BEFORE: `(method as StubBasedPsiElement<*>?)?.stub != null`
This will throw a cast exception if not stub-based. And if it is stub-based, the `?` in `StubBasedPsiElement<*>?` has the effect of silently needlessly casting the known-to-be-non-null stub-based PSI element into a nullable one.

AFTER: `(method as? StubBasedPsiElement<*>)?.stub != null`
As I believe was probably intended, the parenthesized section in this version will not throw, but instead result in `null` if `method` is not a stub-based PSI element, allowing the surrounding if-else control flow to fallback to using ASTNode-based PSI (instead of throwing an exception).

### DIFF
--- a/java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/inferenceResults.kt
+++ b/java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/inferenceResults.kt
@@ -123,7 +123,7 @@ data class MethodData(
     internal val bodyEnd: Int
 ) {
   fun methodBody(method: PsiMethod): () -> PsiCodeBlock = {
-    if ((method as StubBasedPsiElement<*>?)?.stub != null)
+    if ((method as? StubBasedPsiElement<*>)?.stub != null)
       CachedValuesManager.getCachedValue(method) { CachedValueProvider.Result(getDetachedBody(method), method) }
     else
       method.body!!


donnerpeter:
How can this issue be reproduced? From what I see, methodBody is only invoked after getIndexedData call, which checks if it's PsiMethodImpl (extends StubBasedPsiElementBase).

calexHG:
If so, then at this moment it can not. It's also unlikely that it probably ever will, eh? Sorry if the fix was confused as a report of an exception that has been encountered in the wild. To clarify: it has not. 

I noticed code that wasn't doing what the developer intended, so I fixed it.

Whether it avoids an error today within this code; whether it avoids an error a year from now when the assumptions you and I make today no longer are true; or whether it avoids an error in completely different code because it might have been caused by a language misunderstanding by the developer rather than just a simple typo, and is thus repeated within code that *does* trigger the error; I think is not something to ponder about too much. An error is best fixed before it causes problems, rather than after :) 

P.S.: Thank you for the addition of the *Gist API, Peter. It should be rather helpful.

donnerpeter:
I see your point, thanks! However, I believe that such issues are even better prevented by type system. What do you think of passing PsiMethodImpl everywhere around MethodData instead of PsiMethod?

calexHG:
Honestly? Unless that was intended to be a sarcastic, rhetorical question,
I think you should be made the new vice president of IntelliJ interface
design for that positive suggestion ;)

+1
Definitely, accept only PsiMethodImpl if in truth only PsiMethodImpl can be
handled. (I don't have the code in front of me at the moment)

(reminder to put you at ease: you always have the option to widen the
interface to accept and handle more in the future; it's doing the opposite
will be troublesome..)


On Feb 13, 2017 3:46 AM, "Peter Gromov" <notifications@github.com> wrote:

I see your point, thanks! However, I believe that such issues are even
better prevented by type system. What do you think of passing PsiMethodImpl
everywhere around MethodData instead of PsiMethod?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/JetBrains/intellij-community/pull/511#issuecomment-279325761>,
or mute the thread

donnerpeter:
Done, thanks for inspiration!

PR #513, opened 2017-02-16
Adds new macro expansion $MODULE_CONTENT_ROOT$

jeantil:
This new macro is expanded to the current module's content root when a
the module only has one content root.
I have never seen a case were a module has multiple content roots, and
am therefore unable to decide what to do in these cases.
When a project uses the file based project descriptor (`*.iml`), this
should be strictly equivalent to `$MODULE_DIR$`
When a project uses the directory based project description (`.idea/*`),
this macro points to the content root instead of the parent directory of
the module file.

I am unfamiliar with the codebase and was unable to find a test which
exercises the code path I changed. I tried writing one based on
`PathMacroManagerTest` but was unable to make it work as it seems to
requires a lot of setup which goes beyond the setup already provided. 
Given guidance I am willing to try and add a test for it. Running the IDE 
seemed to resolve it well enough.

Please consider including this since it opens a way to fixing:

https://youtrack.jetbrains.com/issue/IDEA-152227
https://youtrack.jetbrains.com/issue/SCL-9296
https://youtrack.jetbrains.com/issue/IDEA-121819
https://youtrack.jetbrains.com/issue/IDEA-153852
And probably a few more `$MODULE_DIR$` related issues

### DIFF
--- a/jps/model-serialization/src/org/jetbrains/jps/model/serialization/PathMacroUtil.java
+++ b/jps/model-serialization/src/org/jetbrains/jps/model/serialization/PathMacroUtil.java
@@ -35,6 +35,7 @@ import static com.intellij.openapi.util.io.FileUtilRt.toSystemIndependentName;
 public class PathMacroUtil {
   @NonNls public static final String PROJECT_DIR_MACRO_NAME = "PROJECT_DIR";
   @NonNls public static final String MODULE_DIR_MACRO_NAME = "MODULE_DIR";
+  @NonNls public static final String MODULE_CONTENT_ROOT_MACRO_NAME = "MODULE_ROOT";
   @NonNls public static final String DIRECTORY_STORE_NAME = ".idea";
   @NonNls public static final String APPLICATION_HOME_DIR = "APPLICATION_HOME_DIR";
   @NonNls public static final String APPLICATION_CONFIG_DIR = "APPLICATION_CONFIG_DIR";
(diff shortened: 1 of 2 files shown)


klikh:

jeantil:
Hi @klikh, 

I thought I had a contributor license agreement since I contributed to intellij-scala before but it may not be the same. 

I just went through the process and signed the agreement, just to make sure. 

best regards
Jean

klikh:
@jeantil We've received the CLA. Thanks.

chashnikov:
Sorry for the late reply, I was on vacation. Unfortunately the suggested solution won't work.
Firstly, you changed ModulePathMacroManager which works for module-level component, but run configurations are store in project-level component, so your code won't be called for them. Secondly, you changed only getExpandMacroMap, but didn't change getReplacePathMap, so if your code is executed, it'll replace the macro by the full path when the project is opened, but won't replace the full path by the macro when configuration is saved, so you'll end up with absolute paths in iml. Thirdly if we implement replacement in 'getReplacePathMap' it won't work correctly, because the content root are also saved in iml file so its path will be replaced by the macro, and information will be lost.

It looks like that these issues shouldn't be fixed on PathMacroManager's level. We'll think about a proper solution for these problems. Anyway thank you for your attempt to fix this.

jeantil:
As I tried to explain, it opens the way to fixing the issues, it is not a full fix in itself. What I wanted was to be able to :

- introduce the `MODULE_ROOT` expansion key
- populate the ExpandMacroMap of the ModulePathMacroManager with the correct value for the module root module by module

I thought I would then be able to detect the `$MODULE_ROOT$` working path within the test configuration runner and request the `ModulePathMacroManager` to extract the proper path. I was confident the java and scala test configuration runners would already have access to module level information since they are able to use `$MODULE_DIR$` as a working directory  and to refer to the classpath of a given module (usually the module enclosing the test class)

This plan is most certainly idealistic but the intellij codebase is quite a beast :) If you can provide guidance I will definitely have a look into improving this further.

Maybe populating the expandMacro map is useless if the runner has access to the module itself it will be able to retrieve the module root.

PR #514, opened 2017-02-17
Pasting mvn dependency with classifier to gradle

platan:
Gradle supports maven classifiers https://docs.gradle.org/current/userguide/dependency_management.html#sub:classifiers. This change adds support for classifiers in `PasteMvnDependencyPreProcessor`. 

I've signed CLA.


platan:
What do you think about this change? Is there anything missing?

PR #518, opened 2017-02-23
IDEA-168508 make collapse targets configurable

zdary:
IDEA-168508 make collapse targets configurable


klikh:
Has been resolved in c8fbaa5f.

PR #519, opened 2017-02-23
IDEA-105458 XML nodes shouldn't be always treated as leafs

zdary:
IDEA-105458 XML nodes shouldn't be always treated as leafs


SergeyMalenkov:
@zdary Could you please provide a test (or an example XML) to ensure that the fix solves the issue?

PR #520, opened 2017-02-23
IDEA-151925: IntelliJ should be aware that a custom sourceSet contains Test sources for production code

dmgd:
https://youtrack.jetbrains.com/issue/IDEA-151925


plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/builder/ExternalProjectBuilderImpl.groovy
```diff
@@ -471,6 +472,10 @@ class ExternalProjectBuilderImpl implements ModelBuilderService {
     result
   }
 
+  private static boolean isTestSourceSet(Project project, SourceSet sourceSet) {
```
vladsoroka: With the change all IDEA modules related to custom source sets with related `Test` task defined will get test source folders.
It's not configurable behavior and might confuse some users, the source folder scope change will break other dependent IDEA modules.
And I don't think it's a good idea in general to introduce such additional conventions.

The IDEA gradle import should stick to `idea` plugin configuration as much as possible.
Unfortunately, `idea` plugin doesn't support custom source sets but we are working closely with Gradle folks to add the support.
Before the support will be added to `idea` plugin we can use `idea.testSourceDirs+= file('some-extra-test-dir')` API to mark such 'test' IDEA explicitly, e.g.
```
apply plugin: 'java'
sourceSets {
  integrationTest
}
apply plugin: 'idea'
idea {
  module {
    testSourceDirs += project.sourceSets.integrationTest.java.srcDirs
    testSourceDirs += project.sourceSets.integrationTest.resources.srcDirs
  }
}
```
and use the following check `idea.testSourceDirs.containsAll(customSourceSet.allJava.srcDirs)' to resolve 'test' IDEA modules.

vladsoroka:
Thanks for your time working on the PR but I have to reject it for now (see the details in the above comment).
I've added resolving of the 'test' IDEA module for custom source set based on the `idea.module.testSourceDirs` API. The change will be available in 2017.2.3, please check if it will work for you.

PR #521, opened 2017-02-27
Python: support multiple import source directories with the same qualified name

brendandouglas:
This was supported prior to [f0d6fdce](https://github.com/JetBrains/intellij-community/commit/f0d6fdce64b6c011ef876a39a7f9da077b569aae).

To support the Bazel build system, we rely on being able to provide multiple import source candidates for a given directory.

This is required for `from package import symbol` statements, where `PyResolveImportUtil::resolveQualifiedName` is called with only the `from package` context, and doesn't resolve to a unique candidate directory.


brendandouglas:
Hi there! I've added some tests to demonstrate these changes, along with a regression test for PyCustomPackageIdentifer.

vlasovskikh:
The YouTrack issue for this PR is [PY-22522](https://youtrack.jetbrains.com/issue/PY-22522).

brendandouglas:
Thanks, that sounds good. I've updated the PR with your proposed change.

vlasovskikh:
Thank you for your pull request. I've rebased it on top of the current master. The fix will be available in the Python plugin the next week. Follow PY-22522 for updates.

PR #522, opened 2017-02-27
Add a way to skip git pre-commit and pre-push hooks (IDEA-115914)

ajeihala:
Git supports a `pre-commit` and `pre-push` hooks that are executed
after a commit or a push accordingly (https://git-scm.com/docs/githooks).
Sometimes it's needed to skip the hooks and bypass the checks for them.
For that purpose git supports a parameter `no-verify`.
With that commit inside the git commit dialog a checkbox "Skip hook" will be
available if any of the repositories have a `pre-commit` hook installed.
Similar way in the push dialog it will be possible to bypass the
`pre-push` hook.
Options are not visible and not used when there's no such hooks installed.

Currently only local hooks (inside .git/hooks folder) are supported.
It doesn't support a new git configuration variable `core.hooksPath`
that allows customizing where the hook directory is (since git 2.9.0).

### DIFF
--- a/plugins/git4idea/src/git4idea/branch/GitDeleteRemoteBranchOperation.java
+++ b/plugins/git4idea/src/git4idea/branch/GitDeleteRemoteBranchOperation.java
@@ -162,7 +162,7 @@ class GitDeleteRemoteBranchOperation extends GitBranchOperation {
 
   @NotNull
   private GitCommandResult pushDeletion(@NotNull GitRepository repository, @NotNull GitRemote remote, @NotNull String branchName) {
-    return myGit.push(repository, remote, ":" + branchName, false, false, null);
+    return myGit.push(repository, remote, ":" + branchName, false, false, false, null);
   }
 
   @Nullable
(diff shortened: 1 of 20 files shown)


klikh:
Nice work! I don't see any serious issues with this pull request, so almost ready to merge. 

However, we are not sure about the UI of the option: where should it be added and whether its selection should be preserved or not. So please share your use cases in the YouTrack ticket. Thanks.

ajeihala:
@klikh thanks.
I shared my use cases and put them to the comments (https://youtrack.jetbrains.com/issue/IDEA-115914#tab=Comments).
Gave a better thought on the selection preserving and probably it makes more sense not to remember the selection to avoid accidental pushes without hooks verification.

ajeihala:
@klikh I noticed that the branch got conflicts with the upstream. So I repushed the branch again with the next changes:
- Removed the selection persistence for both commit and push dialogs. Now checkboxes are not selected when you open a dialog even if you committed/pushed previously with a "skip hook" option.
- Rebased the branch on top of current upstream master and fixed the conflict.

klikh:
It appears that there are more conflicts after my recent changes in master. Don't worry about them: I've already applied your changes to a private branch, and currently I'm making some small additional changes are passing review. So I'll resolve the conflicts.

ajeihala:
Thank you! I've recently noticed the conflicts and was about to start fixing them. You are right on time :)

klikh:
The fix was merged manually with some cleanups and additional changes discussed in https://youtrack.jetbrains.com/issue/IDEA-115914.

If you're interested, here are the hashes of all related commits: fdf6978..ed116f2

PR #524, opened 2017-03-02
Avoid deregistering roots that will registered again in fsNotifier

vootelerotov:
The current behavior is that once a new set of roots is received,
all the current roots were first deregistered and then
the new ones registered.

In case the majority of the roots belong to the both
sets, this causes two issues:

  - Performance overhead

  - Loss of event if it happened after deregistering
    before registering

To avoid those issues, the set of current and new roots
can be compared and only new roots registered and
only the old roots no longer present deregistered.


trespasserw:
Sorry for the late reply.

Thanks for your contribution!

PR #530, opened 2017-03-26
Avoid splitting inline JavaDoc tags when reformatting.

c2nes:
Fixes [IDEA-122233](https://youtrack.jetbrains.com/issue/IDEA-122233).

There's another small change here with how short lines are handled. The old code would allow the comment line to overrun the right margin if the line would otherwise be less than two-thirds the desired width.

I left this out in the update because I worry that allowing the margin to be overrun in theses cases could cause violations with tools like Checkstyle. I'm happy to re-add this logic if that's preferred.


ignatov:
сс @dyadix

PR #531, opened 2017-03-27
PY-12627: Share and persist 'search for text occurences' option

fitermay:


python/src/com/jetbrains/python/findUsages/PyCommonFindUsageOptions.java
```diff
@@ -0,0 +1,72 @@
...
+import com.intellij.openapi.components.ServiceManager;
+import com.intellij.openapi.components.State;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * @author Yuli Fiterman
+ */
+@State(name = "PyCommonFindUsageOptions")
+public class PyCommonFindUsageOptions implements PersistentStateComponent<PyCommonFindUsageOptions.State> {
```
vlasovskikh: You can extend `FindUsagesOptions` with your `PyCommonFindUsageOptions` (I would prefer to call it `PyFindUsagesOptions`, we have similarly named classes for Java). Then you'll need only `getInstance()`, `getState()`, and `loadState()` here. The state persisting methods could access `isSearchForTextOccurrences` from the base class.

python/src/com/jetbrains/python/findUsages/PyCommonFindUsageOptions.java
```diff
@@ -0,0 +1,72 @@
...
+import com.intellij.openapi.components.PersistentStateComponent;
+import com.intellij.openapi.components.ServiceManager;
+import com.intellij.openapi.components.State;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * @author Yuli Fiterman
+ */
+@State(name = "PyCommonFindUsageOptions")
```
vlasovskikh: Let's store this state inside the existing workspace file which stores local project settings: `@State(name = "PyFindUsagesOptions", storages = @Storage(StoragePathMacros.WORKSPACE_FILE))`.

python/src/com/jetbrains/python/findUsages/PyFindUsagesHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public abstract class PyFindUsagesHandler extends FindUsagesHandler {
+  protected PyFindUsagesHandler(@NotNull PsiElement psiElement) {
+    super(psiElement);
+  }
+
+  @NotNull
+  @Override
+  public FindUsagesOptions getFindUsagesOptions(@Nullable DataContext dataContext) {
+    return PyCommonFindUsageOptions.getOptions(getProject(), dataContext);
```
vlasovskikh: There would be no need in your own `getOptions` if you inherited your options class from `FindUsagesOptions`. It could be just `PyFindUsagesOptions.getInstance(getProject())`.

python/src/com/jetbrains/python/findUsages/PyClassFindUsagesHandler.java
```diff
@@ -28,7 +27,7 @@
 /**
  * @author yole
  */
-public class PyClassFindUsagesHandler extends FindUsagesHandler {
+public class PyClassFindUsagesHandler extends PyFindUsagesHandler {
```
vlasovskikh: Let's add our Python-specific handler as the base class for all Python find usages handlers, including `PyModuleFindUsagesHandler`.

fitermay:
Hi,

I've renamed PyCommonFindUsageOptions and changed persistence to save into the workspace file.  The other suggestion unfortunately do not appear to work.
It's been a while since I've looked at this change but now I recall why I implemented it this way instead of the way you suggested.

1) PyModuleFindUsagesHandler cannot be used with PyFindUsagesHandler because its dialog does not have the 'search for text occurences' option. This causes the option to reset to 'false' no matter what the user set it to last.

2) Extending 'FindUsagesOptions' does not appear to be possible without serializing to 'Element' because 
serialization requires a no-arg constructor.  However, FindUsagesOptions cannot be constructed with no-args ; it needs at least a non-null project

python/src/com/jetbrains/python/findUsages/PyFindUsagesHandler.java
```diff
@@ -0,0 +1,38 @@
...
+import com.intellij.find.findUsages.FindUsagesOptions;
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.psi.PsiElement;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * @author Yuli Fiterman
+ */
+public abstract class PyFindUsagesHandler extends FindUsagesHandler {
```
vlasovskikh: I see your point about `PyModuleFindUsagesHandler`. Let's make `PyTargetExpressionFindUsagesHandler` extend `PyFindUsagesHandler` and add `isSearchForTextOccurrencesAvailable() = true` if necessary.

fitermay:
@vlasovskikh I found a way to work around the issues and do it the way you requested. Also squashed into 1 commit.

python/src/com/jetbrains/python/findUsages/PyFindUsagesHandler.java
```diff
@@ -0,0 +1,43 @@
...
+    super(psiElement);
+  }
+
+  @NotNull
+  @Override
+  public FindUsagesOptions getFindUsagesOptions(@Nullable DataContext dataContext) {
+    PyFindUsagesOptions sharedOpts = PyFindUsagesOptions.getInstance(getProject());
+    final PyFindUsagesOptions opts;
+    opts = !isSearchForTextOccurrencesAvailable(getPsiElement(), false) ? (PyFindUsagesOptions)sharedOpts.clone() : sharedOpts;
+    sharedOpts.searchScope = new FindUsagesOptions(getProject(), dataContext).searchScope;
```
vlasovskikh: @fitermay Why do you set the search scope explicitly here? Why don't you set it for the cloned `opts`?

fitermay: That's a good point. Recalculating scope is not necessary as it is recomputed later anyways. Fixed.

fitermay:
I've made the changes but maybe we can do even better if we can avoid creating stubs for instance attributes entirely by getting rid of:
`  if (functionNode != null && qualifierNode != null) {
      final PsiElement function = functionNode.getPsi();
      if (function instanceof PyFunction && PyNames.NEW.equals(((PyFunction)function).getName())) {
        return true;
      }
      final ASTNode parameterList = functionNode.findChildByType(PyElementTypes.PARAMETER_LIST);
      assert parameterList != null;
      final ASTNode[] children = parameterList.getChildren(PyElementTypes.FORMAL_PARAMETER_SET);
      if (children.length > 0 && children[0].getText().equals(qualifierNode.getText())) {
        return true;
      }
    }`

Inside 'PyTargetExpressionElementType'. However, I'm not sure if this is needed by any custom stub types or not.

fitermay:
Ok, setting it to true fixes the tests.

vlasovskikh:
@fitermay I've just tried searching for the usages of a function. After I turned off "Search for text usages" and restarted PyCharm the text usages setting was turned on back. It turns out that during the refactoring the main functionality got broken. Could you please revisit this PR?

fitermay:
@vlasovskikh Fixed. -- I forgot to set 'isSearchForTextOccurrences' to false by default. It needs to have the same default as the field in the persisted state or else override 'noStateLoaded' since loadState does not get invoked for the default state...

fitermay:
@vlasovskikh ping

vlasovskikh:
@fitermay Sorry, I've missed the notification. I will look into it tomorrow.

vlasovskikh:
@fitermay I've rebased your PR onto master. Thanks!

PR #532, opened 2017-03-27
PY-23150: Improve autocompletion speed for modules with a lot of top level symbols

fitermay:
This should make completions appear instantly even for
modules with a large number of imports. Also includes
a specific speedup for numpy.

### DIFF
--- a/python/src/com/jetbrains/numpy/codeInsight/NumpyModuleMembersProvider.java
+++ b/python/src/com/jetbrains/numpy/codeInsight/NumpyModuleMembersProvider.java
@@ -18,6 +18,7 @@ package com.jetbrains.numpy.codeInsight;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.util.QualifiedName;
 import com.jetbrains.python.codeInsight.PyCustomMember;
+import com.jetbrains.python.codeInsight.PyPsiPath;
 import com.jetbrains.python.psi.PyFile;
 import com.jetbrains.python.psi.PyUtil;
 import com.jetbrains.python.psi.resolve.PyQualifiedNameResolveContext;
@@ -54,15 +55,18 @@ public class NumpyModuleMembersProvider extends PyModuleMembersProvider {
   @Override
   protected Collection<PyCustomMember> getMembersByQName(PyFile module, String qName) {
     if ("numpy".equals(qName)) {
-      final List<PyCustomMember> members = new ArrayList<>();
-      for (String type : NUMERIC_TYPES) {
-        members.add(new PyCustomMember(type, DTYPE, false));
+      PsiElement clazz = new PyPsiPath.ToClassQName(DTYPE).resolve(module);
+      if (clazz != null) {
+        final List<PyCustomMember> members = new ArrayList<>();
+        for (String type : NUMERIC_TYPES) {
+          members.add(new PyCustomMember(type, clazz, DTYPE));
+        }
+        for (String type : PYTHON_TYPES) {
+          members.add(new PyCustomMember(type, clazz, DTYPE));
+        }
+        addTestingModule(module, members);
+        return members;
       }
-      for (String type : PYTHON_TYPES) {
-        members.add(new PyCustomMember(type, DTYPE, false));
-      }
-      addTestingModule(module, members);
-      return members;
     }
     return Collections.emptyList();
   }
--- a/python/src/com/jetbrains/python/psi/resolve/CompletionVariantsProcessor.java
+++ b/python/src/com/jetbrains/python/psi/resolve/CompletionVariantsProcessor.java
@@ -17,13 +17,16 @@ package com.jetbrains.python.psi.resolve;
 
 import com.intellij.codeInsight.lookup.LookupElement;
 import com.intellij.codeInsight.lookup.LookupElementBuilder;
+import com.intellij.codeInsight.lookup.LookupElementDecorator;
+import com.intellij.codeInsight.lookup.LookupElementPresentation;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Condition;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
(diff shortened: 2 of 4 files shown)


east825:
I agree with the most of the changes, though this lazy rendering of `LookupItem`'s type looks a bit doubtful. My original intention was to figure out how to make `findCanonicalImportPath()` itself faster, and, first of all, how come that caching of its result doesn't help up here. I'll look more closely into your changes before the end of the week (on a business trip now).

fitermay:
I know it looks doubtful because of the use of placeholder type text but it is the only way to get real-time performance in cases such as numpy and tensorflow.  This way we look up the type for 10-15 items that are displayed instead of 700+ items on the list.  I checked and the only places that getTypeText is used before are:
1) PythonCompletionWeigher to weight 'dict key' higher, so this change does not affect it. 
2) To estimate the width of the completion list. In this case the completion list may be slightly wider than necessary but from my experiments the difference is very minor. 

Also, the resolves types lazily only if there are more than 50 completions. In other cases we fall back to the old way.


As for findCanonicalImportPath, I have made it faster but it is not possible to make it fast enough for desired real-time performance, especially not on slightly older machines.  The caches do not help because the item being imported is a parameter, so 700 different items necessarily means 700 traversals through the package chain to find the first one from which it can be imported. 

So as I see it the only two possibilities are:
1) Perform findCanonicalImportPath lazily for only displayed items
2) Change what you display as the type text to something that doesn't require expensive lookup

fitermay:
Some more details about why caches don't help:
Each module caches symbols defined at the module level but for symbols imported from other modules ,star imports and from imports, it has to follow the import and do so recursively until the declaring module is found. You can try to flatten this search by caching imported symbols but generating this cache would be very expensive and it would need to be invalidated on every PSI change. So it would likely do more harm than good. 

Alternatively you can use a special cache invalidation strategy for library files by implementing a custom modification tracker and maintaining special caches for libraries but that's a fairly large task that I think goes beyond the scope of this single issue. 

So please consider my proposal of going around the problem entirely and using lazy resolve. It is not as scary as it sounds:)

east825:
I found out that performing the lookup of the topmost package exporting a symbol in the opposite direction in `QualifiedNameFinder#doFindCanonicalImportPath()` doesn't improve the situation. In the extreme case where all packages and modules propagate their symbols one level higher using wildcard imports we are going to scan through all possible combinations of a star import and a symbol's name no matter where we actually start. For instance, in the following example

```
.
├── lib
│   ├── __init__.py
│   ├── pkg0
│   │   ├── __init__.py
│   │   ├── pkg1
│   │   │   ├── __init__.py
│   │   │   ├── mod0.py
│   │   │   └── mod1.py
│   │   └── pkg2
│   │       ├── __init__.py
│   │       ├── mod2.py
│   │       └── mod3.py
│   └── pkg3
│       ├── __init__.py
│       ├── pkg4
│       │   ├── __init__.py
│       │   ├── mod4.py
│       │   └── mod5.py
│       └── pkg5
│           ├── __init__.py
│           ├── mod6.py
│           └── mod7.py
└── main.py  # from lib import <caret> here
```

every module defines a single name that is then imported in `__init__.py` of the corresponding package using "star" import and then propagated further up to `lib/___init__.py`. Despite the fact that there are only eight symbols in total, we do `num_init_py * num_star_imports_per_module * num_symbols = 7 * 2 * 8 = 112` calls to `PyStarImportElementImpl#multiResolveName()` trying to discover canonical names for all of them. This is there the actual (combinatorial) bottleneck is in my opinion.

fitermay:
Yes, but in this case we exit the first time we find the import. The previous algorithm scanned those symbols several times because it never exited.

east825:
I see, but in order to find the import in `lib/__init__.py` we still need to drill down all the nested packages and modules in the hierarchy, and the results of `PyStarImportElementImpl#multiResolveName()` are cached so the previous implementation didn't look for an imported name twice.

fitermay:
Good point. I may be wrong here. I'll need to take a better look at the intermediate caches and run profile without the other changes in place. Can we talk about the lazy type resolve? It is really the change which makes the completions instant but is debatable.

fitermay:
I've tested it with a profiler. Like you said, the improvement is not great because PyStarImportElementImpl#multiResolveName() is cached. However, there is an improvement in the order of around 5% because traversing name definers in submodules is not quite free. You may keep that change or your may remove it if you think it's not worth it.

east825:
To keep things simple we first want to try hiding canonical names in the completion popup altogether using rather the real qualified names of defining modules. An alternative approach would be to somehow pre-populate the cache of resolved symbols for the given name available via each star import element (the one in `PyStarImportElementImpl#multiResolveName()`) while computing completion variants but it's not completely obvious where to do so without affecting the rest of the code insight.

fitermay:
That sounds reasonable -- you went with option #2: changing what you present as types.  I still prefer the lazy resolve approach because it doesn't change functionality. It does so at some potential cost to maintainability, and since you are maintaining it I cannot argue :)  Do you want to do a partial merge of this with just the numpy changes or do you need me to remove the rest of the commits?

fitermay:
Regarding the approach of prepopulating the caches: I think this can only be done with major refactoring or a major hack.

east825:
Yes, I'll integrate your Numpy-related patch alone. No need to modify the PR. It doesn't improve the responsiveness of the completion that much (about 80 milliseconds on my machine), but it doesn't hurt either.  As a nice side-effect it changes the icon of the corresponding entries in the completion popup to a more logical one (they were listed as methods previously) :) 

Overall, there are already tons of various caches in PyCharm that serve as a common source of confusion and cryptic errors. We try not to add additional layers of laziness if it's possible to avoid that. 

Meanwhile, I'll look further into how to reduce the computational cost of `getCanonicalImportPath()`. We rely on it in multiple places besides code completion, so this unfortunate oversight hinders other functionality too (though not that dramatically).

fitermay:
Thanks! Last thing I want to say: I get about 80ms on improvement on my machine as well. I tested the performance with laziness patch already applied so that 80ms made up a significant % of the overall time (IIRC over 20%) for numpy. Once you remove reliance on getCanonicalImportPath for completions I think you will get the same result.

PR #533, opened 2017-03-28
Fixed: Get sdk flavor by sdkpath does not look through flavor from extension points

da1z:
Fixed issue when getting sdk flavor by sdk path was not looking through sdk flavors from PythonFlavorProvider extension point


traff:
Merged

PR #535, opened 2017-03-28
Guess project by file for line separator if document was modified outside of any command

da1z:


da1z:
Hi @traff 
Could you look at this one as well please?

fitermay:
Justification: We had an issue with file seperator setting being sometimes ignored on save and the default fs seperator being used instead. This change fixed it. TBH wasn't able to track down the exact circumstances, just know that this fixes it.

donnerpeter:
Sorry for the delay (I've missed the notification). Merged manually and pushed to master.

PR #536, opened 2017-03-29
Allow jrt file system to be created in CoreApplicationEnvironment

udalov:
This is needed to allow the Kotlin compiler to read the Java 9 runtime
image via a custom VirtualFileSystem implementation similar to
JrtFileSystem/JrtFileSystemImpl


yole:
Merged manually.

PR #537, opened 2017-03-29
Sync idea_default_no_spellchecker.xml

dahlstrom-g:
… with idea_default.xml after divergence since commit a3f7112f4fa.

### DIFF
--- a/.idea/inspectionProfiles/idea_default_no_spellchecker.xml
+++ b/.idea/inspectionProfiles/idea_default_no_spellchecker.xml
@@ -1,11 +1,133 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0" is_locked="false">
     <option name="myName" value="idea.default.no.spellchecker" />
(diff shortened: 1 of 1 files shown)


dahlstrom-g:
@akozlova?

akozlova:
What does 'sync' mean? This profile is used on our internal CI, nothing else. Thanks

dahlstrom-g:
> What does 'sync' mean?

“Sync” means “cause (a set of data or files) to remain identical in more than one location.” In this case, though, the files aren't intended to be completely identical; in commit a3f7112 idea_default_no_spellchecker.xml differs in two small ways: (1) the profile name has “no.spellchecker” appended, and (2) SpellCheckingInspection is disabled. I preserved those differences here.

> This profile is used on our internal CI, nothing else.

As of recently it's used on our internal CI as well; hence the concern about divergence here.

dahlstrom-g:
resolved by commit 872b8dd

PR #540, opened 2017-04-05
IDEA-140862 Changed the "Regex" checkbox ChangeListener to ItemListener

sabi0:
Checkbox `ChangeListener` is also triggered on mouse hover as focus gain / loss state changes. To listen to checkbox 'selected' state changes only you want `ItemListener` instead. The patch does exactly that.

This way the list selection is not reset when mouse is moved over "Regex" checkbox.
I also added a 'g' mnemonic so that regexp mode can be switched with Alt+G hotkey.

### DIFF
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/committed/CommittedChangesPanel.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/committed/CommittedChangesPanel.java
@@ -52,6 +52,8 @@ import javax.swing.*;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 import java.awt.*;
+import java.awt.event.ItemEvent;
+import java.awt.event.ItemListener;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


sabi0:
There are at least 17 other instances of `addChangeListener()` called on a JCheckBox. All of them are used to handle checkboxes selection change. I.e. they are better replaced with `addItemListener()` to avoid receiving unwanted events.
Clicking a checkbox generates at least 5 "change" events:
- armed on
- pressed on
- selected true
- pressed off
- armed off

Should I commit the fix for the other 17 cases to this branch?

klikh:
Yep, it looks reasonable, thanks!

sabi0:
Hi, I don't mean to nag but I'd expect some reaction in a week's time.
Is the patch good enough? Are you going to pull the changes?
(do you actually pull them on GitHub? i.e. will I see the status change from Open to Merged?)

klikh:
Sorry for delay. I've merged the changes manually. Thanks!

GitHub is not the primary repository, so we actually apply the pull request manually and then (within an hour) it is mirrored to GitHub.

platform/vcs-impl/src/com/intellij/openapi/vcs/changes/committed/CommittedChangesPanel.java
```diff
@@ -102,10 +104,13 @@ public CommittedChangesPanel(Project project, final CommittedChangesProvider pro
     toolbarPanel.add(toolBar.getComponent());
     toolbarPanel.add(Box.createHorizontalGlue());
     myRegexCheckbox = new JCheckBox(VcsBundle.message("committed.changes.regex.title"));
+    // FIXME: take mnemonic from VcsBundle too ???
+    myRegexCheckbox.setMnemonic('g');		// same as in Editor > Find toolbar
```
klikh: You can use the "&" symbol to specify the mnemonic char in the bundle, e.g. `Re&gexp`. I've modified your patch like that.

Btw, `setMnemonic(char)` is obsolete, the preferred method is `setMnemonic(int, e.g. VK_G)`

sabi0:
Thanks for the tip!

PR #541, opened 2017-04-07
[Fernflower] Fixed: Use fully qualified static field name in conflict

helgp:
Put 'pkg' in expressions pkg.a.v and pkg.a.f() to refer static members of the class pkg.a when the name 'a' is already used in current class

See the test case pkg/TestClashName for details.


plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/struct/StructClass.java
```diff
@@ -162,6 +164,24 @@ public ConstantPool getPool() {
     return fields;
   }
 
+  // XXX slow, but no lookup table available
+  public boolean containsFieldWithShortName(String shortName){
+    for(StructField f: fields) {
+      if(shortName.equals(f.getName())) {
```
helgp: Could you, please, post a test Java class that fails to decompile this way?

gorrus: This is actually not about anonymous classes, but about switch generated class:
void foo(DayOfWeek d) {
        switch(d) {
            case FRIDAY: return;
        }
    }

helgp: The new test case 'pkg/TestSwitchOnEnum' illustrates the NPE. This was closed with simple NPE guard in the commit  bcc6a3b
Check HeadsUp on commit message.

helgp:
Making all the changes inside the getShortName() was my first attempt to fix the issue. It was found that all other places where the getShortName is invoked do not introduce the observed naming conflict. Without the conflict, the short name can be safely kept at that places. I am preparing a test case to illustrate the statement.

gorrus:
Ok then please fix fields calculation on every request, also in my opinion the code still belongs to the the ImportCollector. Maybe you could introduce getShortNameInClass or something there.

helgp:
1. **Extended test case.** As promised, I have extended TestClashName test case to cover all relevant usages of getShortName() method. Comments in the test case illustrate whether local variables or method may or may not shadow class name in each particular use case. Checking for class name shadowing by local variable/method appears to be required only in FieldExprent, and InvocationExprent use cases. Only these cases use checking for shadowing in the suggested PR. I do not believe that the checking has to be applied to all other use cases, which may happen if the checking logic is unconditionally moved inside the getShortName.getShortName().

2. **Suggested refactoring**. New StructClass.containsFieldWithShortName() is a recursive method in a context of current class. The method traverses up through a class hierarchy when looking for conflicting variable/method name. The method manipulates mainly with locals of the StructClass, namely variables "fields", and "superClass". Moving the method outside of the StructClass context may introduce unclear recursion calls, and direct manipulations on the StructClass variables by a code outside of the StructClass.

Please, revisit your change request considering the above details. I am waiting for your response.

gorrus:
1. Thanks for the test.
2. ImportCollector is created for a class, see ImportCollector(ClassNode root) constructor.
In this single place we could calculate a Set of all fields names and do not do the traverse on every short name request after that.

helgp:
I believe that the today's commits 10b80dd and b3f2e66 should address the change request.

gorrus:
Merged into one commit and pushed into master, thanks!

PR #542, opened 2017-04-08
IDEA-140862 Extend DefaultMutableTreeNode to support restoring tree selection.

sabi0:
DefaultMutableTreeNode does not implement equals() and hashCode().
Because of that it is impossible to re-apply stored selection after re-building tree model.

A new sub-class implements these methods (delegating the calls to "user object").
This fixes the selection reset when changing the filter in "Version Control > Repository" view.


Notoh:
Would not recommend doing the lambda expression instead of just the anonymous interfaces, as adding lambda's would essentially make those classes java 8 dependent which might not be for the best.

klikh:
@Notoh Why is it not for the best? The vcs-impl module is already fully under Java 8.

Notoh:
Lambdas won't add anything other than slight readability (at least in this instance alone) which can sacrifice compatibility if this were to be used with java 7 or earlier.
Because it is java 8 already, at the end of the day its just semantics about practice.

pchel-:
- Save selection. I've applied more common approach for IDEA using `TreeState` class - see https://github.com/JetBrains/intellij-community/commit/c443ab113bcbe1a9f5f4dac5c817d8663126b641.
- Do not re-filter if filter is empty. See https://github.com/JetBrains/intellij-community/commit/36df4c0df9d65bf4e2dc3c43316d0b200b4855b2.

PR #543, opened 2017-04-09
Respect configured up/down user keys in tree views

nd:
Because up/down keys are painful to use on some laptops


klikh:
Please also write your email information to the commit.

nd:
I've updated commit with email & user info

PR #544, opened 2017-04-12
XML element not closed

smmribeiro:
A simple typo: not closing span element.


klikh:
Merged manually. Thanks!

PR #545, opened 2017-04-12
Change return type of PydevConsoleRunnerFactory

da1z:
Changed return type to PydevConsoleRunner to add ability to extend this factory with returning your own PydevConsoleRunner that just implements the interface but not extends PydevConsoleRunnerImpl.

### DIFF
--- a/python/src/com/jetbrains/python/console/PydevConsoleRunnerFactory.java
+++ b/python/src/com/jetbrains/python/console/PydevConsoleRunnerFactory.java
@@ -42,7 +42,7 @@ import java.util.Map;
 public class PydevConsoleRunnerFactory extends PythonConsoleRunnerFactory {
   @Override
   @NotNull
-  public PydevConsoleRunnerImpl createConsoleRunner(@NotNull Project project,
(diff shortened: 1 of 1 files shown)


python/src/com/jetbrains/python/console/PydevConsoleRunnerFactory.java
```diff
@@ -112,8 +112,9 @@ public PydevConsoleRunnerImpl createConsoleRunner(@NotNull Project project,
     putIPythonEnvFlag(project, envs);
 
     Consumer<String> rerunAction = title -> {
-      PydevConsoleRunnerImpl runner = createConsoleRunner(project, module);
-      runner.setConsoleTitle(title);
+      PydevConsoleRunner runner = createConsoleRunner(project, module);
+      if(runner instanceof PydevConsoleRunnerImpl)
+        ((PydevConsoleRunnerImpl)runner).setConsoleTitle(title);
```
traff: It's better to always have curly braces in Java.
Maybe we move setConsoleTitle to the PydevConsoleRunner method?

PR #546, opened 2017-04-12
Speed up PyUnresolvedReference inspection

da1z:
Speed up PyUnresolvedReference exception by caching importable names

### DIFF
--- a/python/src/com/jetbrains/python/codeInsight/imports/PythonImportUtils.java
+++ b/python/src/com/jetbrains/python/codeInsight/imports/PythonImportUtils.java
@@ -23,11 +23,14 @@ import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.registry.Registry;
+import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.*;
 import com.intellij.psi.search.FilenameIndex;
 import com.intellij.psi.search.GlobalSearchScope;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.psi.util.QualifiedName;
+import com.intellij.util.Processor;
+import com.intellij.util.indexing.FileBasedIndex;
 import com.jetbrains.python.PyNames;
 import com.jetbrains.python.codeInsight.PyCodeInsightSettings;
 import com.jetbrains.python.codeInsight.controlflow.ControlFlowCache;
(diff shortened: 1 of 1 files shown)


fitermay:
Justification: We have over 40k __init__ files in our code base. Scanning __init__ files to generate the quick fix takes a prohibitively long time.  Even in medium sized Python projects it is faster to scan for dirs matching package name and then check if it contains __init__ than scan all __init__ files and check that the containing dir matches package name.

Also added cache to further speed it up.

da1z:
@vlasovskikh please take a look at this

fitermay:
@vlasovskikh I agree it is probably fast enough without caching, though I no longer recall the exact numbers. I removed the caching. The reason for the 'private' API usage is to avoid looking up the PsiFile for every file scanned and just resolve the directories when scanning for packages. The public API does not provide scanning only directories.

python/src/com/jetbrains/python/codeInsight/imports/PythonImportUtils.java
```diff
@@ -234,20 +237,31 @@ private static boolean isPossibleModuleReference(PyElement node) {
 
   private static Collection<PsiElement> findImportableModules(PsiFile targetFile, String reftext, Project project, GlobalSearchScope scope) {
     List<PsiElement> result = new ArrayList<>();
-    PsiFile[] files = FilenameIndex.getFilesByName(project, reftext + ".py", scope);
-    for (PsiFile file : files) {
-      if (isImportableModule(targetFile, file)) {
-        result.add(file);
+    //add packages
+    final PsiManager psiMgr = PsiManager.getInstance(project);
+    FileBasedIndex.getInstance().processValues(FilenameIndex.NAME, reftext, null, (file, value) -> {
```
vlasovskikh: `FilenameIndex.NAME` is deprecated. Consider another way of accessing directories. Maybe `processFilesByName(..., includeDirs = true. ...)`?

fitermay: There is no API that achieves the desired effect of avoiding a rather expensive operaton of converting/resolving VirtualFile to PsiFile (psiManager.findFile). Those APIs do not provide for that...

vlasovskikh: `processFilesByName(name, includeDirs = true, ...)` searches only for directories that match `name`, so extra `VirtualFile -> PsiDirectory` happen only for those dirs that match the name but don't contain `__init__.py`. Do you notice any considerable slowdown if you switch from `processValues` to `processFilesByName`?

fitermay: In the use case I'm personally interested in it doesn't make a difference. So for me it would be ok to use 'processFilesByName'.  However,   I've also tested it with mixed Python and non-Python projects with the Python plugin. In the cases where most directories are not Python packages it does make a considerable difference.

fitermay:
I've changed it to 'processFilesByName'. If it causes slowness in mixed projects I guess we can revisit.

vlasovskikh:
@da1z @fitermay Thanks for your PR! I've rebased your changed onto master.

PR #547, opened 2017-04-12
PyConsoleCopyHandler fixes

da1z:
- fixed extra \n appended when copying from console
- fixed columnar selection in console

### DIFF
--- a/python/src/com/jetbrains/python/console/PyConsoleCopyHandler.kt
+++ b/python/src/com/jetbrains/python/console/PyConsoleCopyHandler.kt
@@ -66,7 +66,9 @@ class PyConsoleCopyHandler(val originalHandler: EditorActionHandler) : EditorAct
       val rangeEnd = Math.min(document.getLineEndOffset(i), end)
       if (rangeStart < rangeEnd) {
         sb.append(document.getText(TextRange(rangeStart, rangeEnd)))
-        sb.append("\n")
+        if (rangeEnd < end) {
+          sb.append("\n")
+        }
       }
     }
     if (!sb.isEmpty()) {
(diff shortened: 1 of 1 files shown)


python/src/com/jetbrains/python/console/PyConsoleCopyHandler.kt
```diff
@@ -35,13 +35,13 @@ class PyConsoleCopyHandler(val originalHandler: EditorActionHandler) : EditorAct
...
   override fun doExecute(editor: Editor, caret: Caret?, dataContext: DataContext) {
     if (!RichCopySettings.getInstance().isEnabled) {
-      return originalHandler.execute(editor, null, dataContext);
+      return originalHandler.execute(editor, null, dataContext)
     }
-    if (true != editor.getUserData(ConsoleViewUtil.EDITOR_IS_CONSOLE_HISTORY_VIEW)) {
-      return originalHandler.execute(editor, null, dataContext);
-
+    if (true != editor.getUserData(ConsoleViewUtil.EDITOR_IS_CONSOLE_HISTORY_VIEW)
+        || editor.caretModel.allCarets.size != 1) {
```
traff: Could you please explain what does this caret size comparison do?

python/src/com/jetbrains/python/console/PyConsoleCopyHandler.kt
```diff
@@ -66,7 +66,9 @@ class PyConsoleCopyHandler(val originalHandler: EditorActionHandler) : EditorAct
       val rangeEnd = Math.min(document.getLineEndOffset(i), end)
       if (rangeStart < rangeEnd) {
         sb.append(document.getText(TextRange(rangeStart, rangeEnd)))
-        sb.append("\n")
+        if (rangeEnd < end) {
+          sb.append("\n")
```
traff: If it is an extra \n why not just to trim it and then add \n? I don't quite understand this check.

python/src/com/jetbrains/python/console/PyConsoleCopyHandler.kt
fitermay: This fixes columnar selection. This basically skips this handler when in column selection mode. In row selection mode there is a single caret and in column selection mode there are multiple.

python/src/com/jetbrains/python/console/PyConsoleCopyHandler.kt
fitermay: Yes, trim would accomplish the same thing. The only difference would be if the last selected character is a real '\n' .Not sure if it's actually possible.

traff:
Merged.

PR #551, opened 2017-04-13
Bug Fix : PySkeletonRefresher Cmd line too long

da1z:
Split cmd into batches to avoid command line too long error


python/src/com/jetbrains/python/sdk/skeletons/PySkeletonRefresher.java
```diff
@@ -48,6 +48,7 @@
 import com.jetbrains.python.sdk.InvalidSdkException;
 import com.jetbrains.python.sdk.PySdkUtil;
 import com.jetbrains.python.sdk.PythonSdkType;
+import jdk.nashorn.internal.scripts.JO;
```
traff: This import seems to be unneeded.

da1z: Removed

traff:
Merged

PR #552, opened 2017-04-13
IDEA-171459 fix type cast for int literals used as byte / short

sabi0:
Fixed decompiler to add a type cast for integer values used in `byte` / `short` context.
For example: `Byte.valueOf((byte)123)`

Also removed unnecessary primitive types boxing:
`String.format("%b", new Object[]{Boolean.valueOf(bool)})` =>`String.format("%b", new Object[]{bool})`


ccarpenter04:
Fixing this issue greatly improves the chances of being able to recompile the output of many java applications that I've worked with at various points in the past. Please consider merging this soon.

sabi0:
Sure. I'll check what can be done about the cast for literals that are within respective value ranges.

It's a bit tricky since the compiler behavior is inconsistent for such values.
Both `byte b = 127;` and `return 127;` are ok. But for example `Byte.valueOf(127)` is an error.

gorrus:
I found a note about this here:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.3-500-C

gorrus:
byte[] barr = new byte[]{(byte)0, (byte)0, (byte)0};
same here, casts are not needed

sabi0:
Thanks for the link.

I'll change the code to add the casts for literals only if
a) value is outside respective type range
or
b) value is used as a method / constructor argument

sabi0:
I've realized that literal values will never be "outside respective type range".
E.g. `(byte) 128` will be compiled to `-128`.

gorrus:
merged into master, thanks!

sabi0:
I've just realized that explicit boxing of primitive literals is still present in some cases.
E.g. `Integer.valueOf(1)` in `TestClassLambda.dec`.

Should I create a new PR?

gorrus:
yes, please

PR #553, opened 2017-04-14
SystemHealthMonitor: fix notification race

rsauciuc:
Android Studio defines a custom first-run wizard, which exposes
a race condition in displaying notifications in SystemHealthMonitor.

The root problem is that the appFrameCreated callback happens before
the frame is actually visible, see IdeaApplication where we have:

  lifecyclePublisher.appFrameCreated(args, willOpenProject)
  ... followed by, later
  WelcomeFrame.showNow()

And WelcomeFrame.showNow() first goes through extensions before making
itself available:

  if (ourInstance == null) {
    ...
    for (WelcomeFrameProvider provider : EP.getExtensions()) {
      frame = provider.createFrame();
    }
    ...
    ourInstance = frame;
  }

For any non-trivial extension that displays a dialog and starts
consuming events from the AWT queue, this leads to a race condition:
if SystemHealthMonitor has notifications to show, it may not see
any visible frame and throw an NPE because the welcome frame didn't
get a chance to initialize itself.

We propose to use the appStarting callback, which is guaranteed to
happen later in a separate transaction.

Bug: http://b.android.com/241387

### DIFF
--- a/platform/platform-impl/src/com/intellij/ide/SystemHealthMonitor.java
+++ b/platform/platform-impl/src/com/intellij/ide/SystemHealthMonitor.java
@@ -24,6 +24,7 @@ import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.application.*;
 import com.intellij.openapi.components.ApplicationComponent;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.project.Project;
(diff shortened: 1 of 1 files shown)


ignatov:
@rsauciuc, please rebase the PR, cc @trespasserw.

trespasserw:
The issue should to be solved by ba01935b. Sorry!

PR #554, opened 2017-04-17
Added VC notification

da1z:
We have several actions that auto sync remote sources locally, I am guessing similar to AutoSVN updater. For this we inherit 'AbstractCommonUpdateAction'. Unfortunately we need a little bit more control of the action that 'AbstractCommonUpdateAction' offers, so we added some extensions. The other option would be not to inherit 'AbstractCommonUpdateAction' and instead copy the code. We'd really like to avoid doing this as 'AbstractCommonUpdateAction' implementation handles some very esoteric low level details and we'd like to ensure we receive any updates/bugfixes to it.

CommitChangeListDialog:
We need to be able to pass a parameter to CheckinEnvironment depending on the CommitExecutor. There is currently no way to do this cleanly. Passing the executor(not null) allows CheckinHandler to pass in a parameter to CheckinEnvironment

### DIFF
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/actions/VcsContextWrapper.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/actions/VcsContextWrapper.java
@@ -160,7 +160,7 @@ public class VcsContextWrapper implements VcsContext {
       stream(VcsDataKeys.FILE_PATH_ARRAY.getData(myContext)),
       getSelectedFilesStream().map(VcsUtil::getFilePath),
       stream(getSelectedIOFiles()).map(VcsUtil::getFilePath)
-    );
+    ).distinct();
   }
 
   @Nullable
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/UpdateRequestsQueue.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/UpdateRequestsQueue.java
@@ -204,22 +204,21 @@ public class UpdateRequestsQueue {
       }
     }
 
+    boolean stopped = myStopped;
     synchronized (myLock) {
-      if (! myStopped) {
+      if (!stopped) {
         myWaitingUpdateCompletionQueue.add(data.getCallback());
         schedule();
       }
     }
     // do not run under lock; stopped cannot be switched into not stopped - can check without lock
-    if (myStopped) {
+    if (stopped) {
       LOG.debug("invokeAfterUpdate: stopped, invoke right now for project: " + myProject.getName());
-      SwingUtilities.invokeLater(new Runnable() {
-        public void run() {
-          if (!myProject.isDisposed()) {
-            afterUpdate.run();
-          }
+      ApplicationManager.getApplication().invokeLater(() -> {
+        if (!myProject.isDisposed()) {
+          afterUpdate.run();
         }
-      });
+      }, state != null ? state : ModalityState.any());
       return;
     }
     // invoke progress if needed
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/update/AbstractCommonUpdateAction.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/update/AbstractCommonUpdateAction.java
@@ -301,7 +301,7 @@ public abstract class AbstractCommonUpdateAction extends AbstractVcsAction {
     private LocalHistoryAction myLocalHistoryAction;
 
     public Updater(final Project project, final FilePath[] roots, final Map<AbstractVcs, Collection<FilePath>> vcsToVirtualFiles) {
-      super(project, getTemplatePresentation().getText(), true, VcsConfiguration.getInstance(project).getUpdateOption());
+      super(project, getTemplatePresentation().getText(), true, ActionInfoEx.maybeOverrideBackGroundOption(myActionInfo, VcsConfiguration.getInstance(project).getUpdateOption()));
       myProject = project;
       myProjectLevelVcsManager = ProjectLevelVcsManagerEx.getInstanceEx(project);
       myDirtyScopeManager = VcsDirtyScopeManager.getInstance(myProject);
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/update/ActionInfoEx.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/update/ActionInfoEx.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2000-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.intellij.openapi.vcs.update;
+
+import com.intellij.openapi.progress.PerformInBackgroundOption;
+
+/**
+ * Created by kirylch on 4/12/2017.
+ */
+public interface ActionInfoEx extends ActionInfo {
+  static boolean activeToolWindowOnUpdate(ActionInfo action) {
+    if(action instanceof ActionInfoEx) {
+      return ((ActionInfoEx)action).activeToolWindowOnUpdate();
+    }
+    return true;
+  }
+
+  static PerformInBackgroundOption maybeOverrideBackGroundOption(ActionInfo action, PerformInBackgroundOption option) {
+    if(action instanceof ActionInfoEx) {
+      return ((ActionInfoEx)action).maybeOverrideBackGroundOption(option);
+    }
+    return option;
+  }
+
+  boolean activeToolWindowOnUpdate();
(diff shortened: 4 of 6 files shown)


platform/vcs-impl/src/com/intellij/openapi/vcs/actions/VcsContextWrapper.java
```diff
@@ -160,7 +160,7 @@ public Refreshable getRefreshableDialog() {
       stream(VcsDataKeys.FILE_PATH_ARRAY.getData(myContext)),
       getSelectedFilesStream().map(VcsUtil::getFilePath),
       stream(getSelectedIOFiles()).map(VcsUtil::getFilePath)
-    );
+    ).distinct();
```
klikh: This will make the method actually process all elements, i.e. make it non-lazy. It would eliminate the performance optimization that have been made.
CC @pchel-

platform/vcs-impl/src/com/intellij/openapi/vcs/changes/UpdateRequestsQueue.java
```diff
@@ -204,22 +204,21 @@ public void invokeAfterUpdate(@NotNull Runnable afterUpdate,
...
 
+    boolean stopped = myStopped;
     synchronized (myLock) {
-      if (! myStopped) {
+      if (!stopped) {
         myWaitingUpdateCompletionQueue.add(data.getCallback());
         schedule();
       }
     }
     // do not run under lock; stopped cannot be switched into not stopped - can check without lock
```
klikh: Nice catch, thanks. This comment became not actual (stopped can be switched into not stopped), but the code wasn't fixed.

platform/vcs-impl/src/com/intellij/openapi/vcs/actions/VcsContextWrapper.java
da1z: ok, we can fix duplicates on our side

platform/vcs-impl/src/com/intellij/openapi/vcs/update/ActionInfoEx.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package com.intellij.openapi.vcs.update;
+
+import com.intellij.openapi.progress.PerformInBackgroundOption;
+
+/**
+ * Created by kirylch on 4/12/2017.
+ */
+public interface ActionInfoEx extends ActionInfo {
+  static boolean activeToolWindowOnUpdate(ActionInfo action) {
```
klikh: I think it should be activ**ate**TollWindowOnUpdate, shouldn't it?

Actually, I think that popping the Update Info tab is not a good UI decision anyway. So this better to be fixed unconditionally. But we'll need a link in the notification about successful update to let people focus the toolwindow if they wish. Let me discuss with our UX specialists about that.

platform/vcs-impl/src/com/intellij/openapi/vcs/update/AbstractCommonUpdateAction.java
```diff
@@ -301,7 +301,7 @@ private static boolean supportingVcsesAreEmpty(final ProjectLevelVcsManager vcsM
     private LocalHistoryAction myLocalHistoryAction;
 
     public Updater(final Project project, final FilePath[] roots, final Map<AbstractVcs, Collection<FilePath>> vcsToVirtualFiles) {
-      super(project, getTemplatePresentation().getText(), true, VcsConfiguration.getInstance(project).getUpdateOption());
+      super(project, getTemplatePresentation().getText(), true, ActionInfoEx.maybeOverrideBackGroundOption(myActionInfo, VcsConfiguration.getInstance(project).getUpdateOption()));
```
klikh: Why do you want to override the background option here? Do you want the Update Project operation to be always-modal?

platform/vcs-impl/src/com/intellij/openapi/vcs/actions/VcsContextWrapper.java
fitermay: While 'distinct' does indeed process all elements it is technically not a terminal operation, that is it does not process all elements until a terminal operation is invoked.  However this change is not really critical. If you prefer we can move 'distinct' to 'getSelectedFilePaths' or even remove the change altogether.

platform/vcs-impl/src/com/intellij/openapi/vcs/update/AbstractCommonUpdateAction.java
da1z: We need to be able to start same action in both modes

klikh: Could you please elaborate a bit about the use cases?

da1z: The update task uses 'VcsConfiguration.getInstance(project).getUpdateOption()'. This forces all update actions to share the same option. In our case the option may not be applicable to certain auto update events, some of which must necessarily happen modally and some which should always run in the background and not bother the user.

platform/vcs-impl/src/com/intellij/openapi/vcs/update/ActionInfoEx.java
klikh: I've fixed this in master: e151dea848612cd7e62d602290ac4e3fa1b9ddee
Since now, the Update Info tab never pops up automatically.

platform/vcs-impl/src/com/intellij/openapi/vcs/changes/UpdateRequestsQueue.java
klikh: I've applied this change menually: 134a1db80e0637ee75d60b9003eb6a2f830685db

platform/vcs-impl/src/com/intellij/openapi/vcs/update/AbstractCommonUpdateAction.java
klikh: Could you please shed some light on what update events should happen modally? I'm asking because we were in fact almost ready to remove this setting at all, suspecting (according to our personal feelings and the statistics we get) that nobody updates the project modally.

fitermay: This does not apply to normal VCSes. Even in our VCS implementation most updates can happen in the background.  However our special case we have a concept of layers, similar to ClearCase.  Switching layers may affect the SDK,  console environment, state of several controls, etc.  We'd really rather prefer avoid handling complex in-between states for this operation right now and just perform it modally.

platform/vcs-impl/src/com/intellij/openapi/vcs/update/ActionInfoEx.java
fitermay: We have an auto sync task running in the background when user is idle.  We'd like to avoid spamming tons of sticky notifications if possible. Are these notifications sticky?

platform/vcs-impl/src/com/intellij/openapi/vcs/update/AbstractCommonUpdateAction.java
klikh: Which VCS do you integrate with (if it is public)?

platform/vcs-impl/src/com/intellij/openapi/vcs/update/ActionInfoEx.java
klikh: No, they are non-sticky by default. `VcsNotifier#STANDARD_NOTIFICATION` group is used. 

Note, that there won't be tons of notification even if sticky anyway, because they are merged.

However, receiving even lightweight notifications for a silent action is not a good idea, so it seems that we need some customization here anyway.

platform/vcs-impl/src/com/intellij/openapi/vcs/update/AbstractCommonUpdateAction.java
fitermay: I wish it was a public standard VCS:)
 It's a mixture between CVS and a remote object database. It's a long story, I've written a long email to @traff some time back explaining how it works.

fitermay:
@klikh 
We're going to be withdrawing this request as we decided to implement our own action fully rather than inheriting from AbstractCommonUpdateAction so we can have better control.

There is a fairly serious 'performance leak' in showing 'UpdateInfoTree' when a large number of files is involved.  When expanded each 'FileTreeNode' registers as a root in 'VirtualFilePointerManager'. This causes VirtualFilePointerManager to traverse a large number of paths for every VFS update. I've seen this cause UI freezes of up to 4 minutes with a large change (over 8k files) followed by a large VFS update.

fitermay:
@klikh I'd still like https://github.com/JetBrains/intellij-community/pull/554/commits/8f218e64e1e6770419f76b5e6e78a3a0aad171d8 merged if possible.  It allows to the 'CheckinHandlerFactory' to pass extra parameters to the CheckinEnvironment via 'additionalDataConsumer' depending on the executor chosen.

For example, right now to handle 'CommitAndPush' you rely on an essentially global variable (myCheckinEnvironment.setNextCommitIsPushed(true)). I think it would be cleaner to pass it as a parameter to the CheckinEnvironment.

klikh:
@fitermay 8f218e6 has already been applied as 9e49ea2

fitermay:
Yes, I think it's better to roll our own implementation from scratch rather than inheriting from AbstractCommonUpdateAction. This will avoid complicating your codebase and will let us customize as we need. We should be able to work with changes already merged. Thanks so much for your help.

fitermay:
@da1z Could you please close this PR?

PR #556, opened 2017-04-17
Date: remove usage of deprecated API

smmribeiro:
Replaced Date for Calendar.
Two implementation notes:
- On "dosToJavaTime" method, I clear the millisecond field to maintain what the Date implementation would return.
- On both "javaToDosTime" and "dosToJavaTime", I handle the year "gap" (Calendar uses the real value while Date uses the year minus 1900).


klikh:
Merged manually. Thanks!

PR #557, opened 2017-04-17
Added frame name passing to PyPositionConverter

da1z:
We have our own PyPositionConverter implementation that requires frame name


traff:
Why do you need frame name? Can it be generalized somehow?

fitermay:
We have an in-house graph framework that essentially rewrites the evaluation order of code to be evaluated in excel-like manner. It creates synthetic frames that needs special handling and translation. We can identify those frames based on the name. 

We could generalize (in a sense) by changing the parameter type passed to the converter to 'PyStackFrameInfo', if you prefer that.

traff:
Merged

PR #559, opened 2017-04-17
Made display name mutable for RunContentDescriptor

da1z:


klikh:
As usual ;) , please provide the motivation. The change is perfectly clear, but it would be nice to get a use case description.

Btw, mutability is evil by default. Not critical for this particular class, since it is already quite mutable, but it is a good habit to have a good motivation for each change like that. Thanks.

da1z:
Lets say we have several consoles and when we want to execute console action if no console selected we use ExecutionHelper.selectContentDescriptor that shows popup with available consoles. We want to be able to modify title of a console in this popup. Mutability of the displayName gives as this possibility. This is one of use cases.

fitermay:
The reason we need this change is we have a custom implementation of the Python console with actions such as "Execute Module in Console", "Execute Script in Console" and so on.  Those actions execute the script/module/test in the console without restarting it.  In those cases we change the console title to "Running script x..." ,Running test y.." to indicate to the context to the user.  In any case "Console (1)", "Console (2)"... are not very useful titles. 

As Kiryl explained, unless we change the title in the RunContentDesciptor code that relies on it, like  ExecutionHelper.selectContentDescriptor, would not work correctly.

Vassiliy-Kudryashov:
Unfortunately, RunContentDescriptor.getDisplayName() is required for internal matching, it's like an ID.
So, it's better to create RunContextDescriptor with proper name and then dispose it when you don't need it anymore.

PR #560, opened 2017-04-19
Fix 'with as' statement for union types

da1z:
Fix type inference for with statement when context manager is union type. For example, before the fix PyCharm could not infer the type of f in the following code:
```
class ClassA(object):
    def __enter__(self):
        return self;
    
    def __exit__(self, exc_type, exc_val, exc_tb):
        return 
    
class ClassB(object):
    def __init__(self):
        super(ClassB, self).__init__()
        self.a = None
        
    def fun(self):
        if self.a is not None:
            return self.a
        self.a = ClassA()
        return self.a

foo = ClassB()
with foo.fun() as f:
    f.
```

### DIFF
--- a/python/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
+++ b/python/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
@@ -57,11 +57,13 @@ import com.jetbrains.python.psi.stubs.PyClassStub;
 import com.jetbrains.python.psi.stubs.PyFunctionStub;
 import com.jetbrains.python.psi.stubs.PyTargetExpressionStub;
 import com.jetbrains.python.psi.types.*;
+import one.util.streamex.StreamEx;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static com.jetbrains.python.psi.PyUtil.as;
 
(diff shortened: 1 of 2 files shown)


python/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
```diff
@@ -243,24 +244,37 @@ private static PyType getWithItemVariableType(TypeEvalContext context, PyWithIte
...
-            PyType typeFromProvider = provider.getContextManagerVariableType(cls, expression, context);
-            if (typeFromProvider != null) {
-              return typeFromProvider;
-            }
-          }
-          // Guess the return type of __enter__
-          return PyUnionType.createWeakType(exprType);
+      if (exprType instanceof PyClass) {
+        final PyType enterType = getEnterTypeFromPyClass(context, expression, (PyClassType)exprType);
+        if (enterType != null)
```
vlasovskikh: This check is unnecessary. The function will return `null` anyway.

python/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
```diff
@@ -243,24 +244,37 @@ private static PyType getWithItemVariableType(TypeEvalContext context, PyWithIte
...
+          return enterType;
+      }
+      else if (exprType instanceof PyUnionType) {
+        List<PyType> collect = ((PyUnionType)exprType).getMembers().stream().filter(PyClassType.class::isInstance)
```
vlasovskikh: `StreamEx.of(...).select(PyClassType.class). ...` is shorter than `filter(...isInstance).map(...cast)`.

python/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
```diff
@@ -243,24 +245,35 @@ private static PyType getWithItemVariableType(TypeEvalContext context, PyWithIte
...
-          }
-          // Guess the return type of __enter__
-          return PyUnionType.createWeakType(exprType);
+      if (exprType instanceof PyClass) {
+        return getEnterTypeFromPyClass(context, expression, (PyClassType)exprType);
+      }
+      else if (exprType instanceof PyUnionType) {
+        List<PyType> collect = StreamEx.of(((PyUnionType)exprType).getMembers())
+          .select(PyClassType.class)
+          .map(t -> getEnterTypeFromPyClass(context, expression, t)).collect(Collectors.toList());
```
sproshev: With `StreamEx` library it's possible to use `toList()` instead of `collect(Collectors.toList())`

da1z:
Added all requested changes

da1z:
@vlasovskikh please take a look at this

python/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
```diff
@@ -243,24 +245,36 @@ private static PyType getWithItemVariableType(TypeEvalContext context, PyWithIte
...
-          }
-          for (PyTypeProvider provider : Extensions.getExtensions(PyTypeProvider.EP_NAME)) {
-            PyType typeFromProvider = provider.getContextManagerVariableType(cls, expression, context);
-            if (typeFromProvider != null) {
-              return typeFromProvider;
-            }
-          }
-          // Guess the return type of __enter__
-          return PyUnionType.createWeakType(exprType);
+      if (exprType instanceof PyClass) {
```
vlasovskikh: This change results in several failing tests. I guess you meant `PyClassType` here, not `PyClass`.

python/testSrc/com/jetbrains/python/PyTypeTest.java
```diff
@@ -1702,6 +1702,31 @@ public void testMaxListUnknownElements() {
...
+           "        self.a = None\n" +
+           "\n" +
+           "    def fun(self):\n" +
+           "        if self.a is not None:\n" +
+           "            return self.a\n" +
+           "        self.a = ClassA()\n" +
+           "        return self.a\n" +
+           "\n" +
+           "foo=ClassB()\n" +
+           "with foo.fun() as expr:\n" +
```
vlasovskikh: Please simplify your test by leaving all the non-essential bits. For example, the current version tests not only using expressions of union types in `with`-statements, but type inference for instance attributes assigned in methods (`self.a = ClassA()`), type inference for `None` checks (`if self.a is not None`) and so on.

I would prefer something along these lines:

```python
from typing import Union

class A(object):
    def __enter__(self):
        return self

class B(object):
    def __enter__(self):
        return self

def f(x):
    # type: (Union[A, B]) -> None
    with x as expr:
        pass
```

fitermay:
@vlasovskikh Fixed according to review. Thanks for catching our instanceof check error.

vlasovskikh:
@da1z @fitermay Thanks! I've rebased the PR onto master.

PR #562, opened 2017-04-19
IDEA-171697 Omit explicit arrays for vararg parameters in decompiled code

sabi0:
The patch gets rid of the explicit arrays creation for vararg parameters. So the decompiler output will change from
`String.format("value: %d", new Object[]{this.value});`
to
`String.format("value: %d", this.value);`

"Vararg-ness" is checked for methods known to the decompiler (e.g. in the same class). And also for any external classes available on the classpath (like e.g. `String.format()`).
I.e. for better results external libraries should be added to the decompiler classpath.


gorrus:
Checking the classpath is super cool!
However the whole classpath seems to be too dangerous, maybe add some filters like "sun.*", "java.*"?

sabi0:
Pre-defined filters won't cut it.
Imagine I'm decompiling a class that uses some "vararg-heavy" library. It would be really great if the calls to the library classes were formatted nicely.

A better solution would be to use a separate ClassLoader, independent from the decompiler own classpath.
And when decompiling a class (in IDEA) you'd configure the ClassLoader with the library jar the class belongs to as well as all its dependencies. This will give the best "coverage" for vararg detection.

I don't have much experience with implementing custom isolated ClassLoaders though.

What's the "danger" of checking the whole classpath? Picking the wrong version of a class and mis-detecting the vararg?

gorrus:
There's "-e" option already which allows to specify a library which will not be decompiled but used for reference. This option is not used inside IDEA, but we could fix it.

sabi0:
Looks like exactly what is needed here.
I will check how to access the classes of that library in the decompiler and re-wire classpath checking to use that approach.

gorrus:
the downside is that it needs to read basic info for all classes from that lib :(

sabi0:
So it does run them through a (limited) decompilation?
It could still be faster than using reflection over big classpath. But needs to be checked, of course.

And I hope it supports multiple libs? (via several "-e" on the command line?)

sabi0:
On the other hand `rt.jar` is on the classpath "for free". But decompiler will have to do a lot of work when processing it via "-e".
Do you know by the way if "-e" library is loaded class-by-class on demand?

sabi0:
Unfortunately libraries specified with "-e" option are all scanned during decompiler startup.
So adding something like `-e=rt.jar` will increase decompiler start time significantly.

gorrus:
yes, loading rt.jar is not an option :(
We may try to load class to the custom class loader with the Bootstrap class loader parent. This will allow fast jdk classes load and should not pollute the main class loader.
Something like:
Class cls = new ClassLoader(null) {}.loadClass(targetClass);
What do you think?

sabi0:
And what about 3rd party libraries? Say a class uses SLF4J vararg calls..

By the way, how IDEA calls decompiler? Does it specify the whole jar file?
I.e. when decompiling some class are other classes from the same library "known" to the decompiler?

gorrus:
Right now a file is decompiled on it's own, no libs are provided.
We can (and probably should) introduce a way to provide the class information from the IDEA indexes, but let's not get it from the classpath.

sabi0:
So if I understood it right, for now I should just change the classpath checking part to use bootstrap classloader parent?

gorrus:
Yes, please

gorrus:
fix is merged into master, thanks!

PR #563, opened 2017-04-20
Decompiler: Remove 'Integer.valueOf()' boxing of literals

sabi0:
This is a continuation of #552.

Removed `Integer.valueOf()` boxing of literals whose type is detected as `byte` or `short`.


sabi0:
@gorrus Would you mind checking this?

gorrus:
I'm slightly worried that you do setConstType inside isBoxingCall method, this looks like an invisible side effect. Please find a better place for it.

sabi0:
I removed `setConstType()` from `isBoxingCall()`.
I also fixed another bug: narrowing cast was not applied when target type was `Byte` or `Short`.

On top of that I removed the unboxing calls too. (`intValue()`, etc.)
This should have probably been done in a separate PR, but I couldn't resist.

gorrus:
Thanks, changes merged to master.

PR #564, opened 2017-04-24
Add an extension point to support custom SDKs with future language features

brendandouglas:


brendandouglas:
This allows using a custom python 2.7 SDK which supports function annotations (e.g. via https://github.com/google/pytype/blob/master/pytype/patches/python_2_7_type_annotations.diff), and otherwise has no effect for anyone not implementing this extension point.

vlasovskikh:
@brendandouglas PyCharm already has an API for customized dialects of Python. You can register your own subclasses of `PythonParserDefinition`, `PyParser`, `FileTypeFactory`.

See `PyFunctionTypeAnnotation*` classes that introduce a dialect for PEP 484 type hints in comments for function types. I believe `FunctionParsing` is flexible enough then, so you can override the necessary pieces to introduce a Python 2 dialect with function annotations.

brendandouglas:
Thanks for the pointer @vlasovskikh, I wasn't aware of the DependentLanguage API. I'll give that a try instead.

PR #565, opened 2017-04-24
i18n for Change Type Signature and Thread Dump actions

smmribeiro:
These two actions had their texts hard-coded at declaration.
For the Change Type Signature action, it was only necessary to delete the attribute from XML (the bundle had the associated text and description); for the Thread Dump action, I had to manually create the entries.

### DIFF
--- a/resources/src/idea/JavaActions.xml
+++ b/resources/src/idea/JavaActions.xml
@@ -58,10 +58,7 @@
     </group>
 
     <group id="RefactoringMenu1">
-      <action id="ChangeTypeSignature"
-              class="com.intellij.refactoring.typeMigration.actions.ChangeTypeSignatureAction"
-              text="T_ype Migration..."
-              description="Change type of the return type of the method, field, parameter, variable or class type arguments and correct all references"/>
+      <action id="ChangeTypeSignature" class="com.intellij.refactoring.typeMigration.actions.ChangeTypeSignatureAction"/>
       <action id="MakeStatic" class="com.intellij.refactoring.actions.MakeStaticAction"/>
       <action id="ConvertToInstanceMethod" class="com.intellij.refactoring.actions.ConvertToInstanceMethodAction"/>
       <add-to-group group-id="RefactoringMenu" anchor="after" relative-to-action="ChangeSignature"/>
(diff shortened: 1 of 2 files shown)


resources/src/idea/JavaActions.xml
```diff
@@ -58,10 +58,7 @@
     </group>
 
     <group id="RefactoringMenu1">
-      <action id="ChangeTypeSignature"
-              class="com.intellij.refactoring.typeMigration.actions.ChangeTypeSignatureAction"
-              text="T_ype Migration..."
-              description="Change type of the return type of the method, field, parameter, variable or class type arguments and correct all references"/>
+      <action id="ChangeTypeSignature" class="com.intellij.refactoring.typeMigration.actions.ChangeTypeSignatureAction"/>
```
klikh: This change makes sense: no code duplication.

resources/src/idea/JavaActions.xml
```diff
@@ -127,7 +124,7 @@
       <add-to-group group-id="XDebugger.Frames.Tree.Popup"/>
       <add-to-group group-id="XDebugger.Actions"/>
     </action>
-    <action id="DumpThreads" class="com.intellij.debugger.actions.ThreadDumpAction" text="Get thread dump" icon="AllIcons.Actions.Dump">
+    <action id="DumpThreads" class="com.intellij.debugger.actions.ThreadDumpAction" icon="AllIcons.Actions.Dump">
```
klikh: Not sure about this one. Why did you choose this single action (among several other ones which have the "text" attribute as well) to move to the bundle? Are you planning to reuse the text from a plugin?

smmribeiro: Hi,

I'd like to see every string of IntelliJ available for translation in resource bundles as it's important for our project; as so, I (we) sometimes identify these kind of things.
However we're not yet using the most recent code so there's no possibility of contributing our changes directly; but, as we explore the new code, we can start to contribute (but as my last PR shown, we need to know the code better (and how it behaves) so that everything is compact and no duplication of code/functionality exists).
For the DumpThreads action, it's something I consider kind of annoying on IntelliJ: when I hover a menu entry, an icon or anything else that will trigger an action, I like to see the description on the Status Bar, however there're several that don't have description and what is shown it the last message sent to the Status Bar.
You can see this behaviour in the following pic:
withandwithoutdescription

As for being that specific action was pure coincidence: I stumbled upon it and made the change :-)

smmribeiro: Hi there,

Will you be using anything else from this PR or can I close it?

Regards,

Sérgio Ribeiro
Porto - Portugal

smmribeiro:
Closing.

PR #566, opened 2017-04-28
IDEA-73669 Allow branch to have longer name in status bar

zoonman:
This change allows to have a much longer branch name in status bar.
This is not perfect but working solution.


klikh:
The issue with this fix is that it doesn't take the IDE window width into account. So if the window is narrow, and the branch name is very long, there is very little space left for the status bar messages and background tasks.

zoonman:
Kirill, previous implementation did not take this into account either. 
Look at my status bar:
status bar
If you give me some recommendations how can I figure out building IDEA on Mac, I can spend some time on weekend to extract max length of branch name into settings.

klikh:
You're right, it doesn't. Merged manually, thank ;)

PR #567, opened 2017-05-02
Add missing package to pydev

DrAA:
Remote debugging with IntelliJ Ultimate fails unless the _pydevd_frame_eval package is included in the pydev helper.

### DIFF
--- a/python/helpers/pydev/setup.py
+++ b/python/helpers/pydev/setup.py
@@ -86,6 +86,7 @@ args = dict(
         '_pydev_runfiles',
         '_pydevd_bundle',
         'pydev_ipython',
+        '_pydevd_frame_eval',
 
         # 'pydev_sitecustomize', -- Not actually a package (not added)
 


Elizaveta239:
Hi! Unfortunately, we can't reproduce the problem. Do you mean debugging with Remote debug server? Do you use `.egg` files provided with the installation? 
In fact, this setup.py is used only for `pydevd` package (and PyDev debugger). For PyCharm and Python plugin we build it separately.

anders-peltarion:
Hi Elizaveta. Yes, the problem occurs when using pydev for remote debugging. More specifically, it is used for being able to debug code running in a docker container. For that purpose, the setup.py is used to build the pydevd-1.0.0-cp35-cp35m-linux_x86_64.whl file, which is installed in the docker image. Without this dependency, it fails with '_pydevd_frame_eval' as a missing module. Hope that gave some more context to be able to reproduce the problem, let me know if you want more info.

Elizaveta239:
Hi @anders-peltarion! Sorry for the late response.
Thank you very much for your patch, but this bug has already been solved.

PR #568, opened 2017-05-02
Additional line number pattern matching for the Open File Dialog

jwren:
Additional line number pattern matching for the Open File Dialog (TreeFileChooserDialog), `#` and `?l=` are now matched covering web source code viewers such as

Chromium: https://cs.chromium.org/chromium/src/apps/app_restore_service_factory.cc?l=15

and

Android: https://android.googlesource.com/device/sample/+/master/apps/LeanbackCustomizer/src/com/google/android/leanbacklauncher/partnercustomizer/PartnerReceiver.java#100

Some more context/ additional improvement on: https://youtrack.jetbrains.com/issue/IDEA-162418


donnerpeter:
Merged manually, thanks!

PR #569, opened 2017-05-04
IDEA-145520 - Don't ask which changelist to make active if there is only one

sabi0:
The confirmation will not appear if only one changelist remains and all the deleted changelists are empty.
The remaining changelist will then be made active silently.

If at least one deleted changelist is not empty the confirmation will be shown to give user a chance to cancel the deletion.

Also 'RemoveChangeList' action text is updated based on the selection size.
("Delete Changelists" looks a bit confusing when only one changelist is selected)


NadyaZabrodina:
Merged with a few small changes. 05d14ef. Thanks!

PR #570, opened 2017-05-12
Remove usages and deprecate some obsolete StringUtil methods

sabi0:
Following the discussion here
https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000240144
I removed usages and deprecated some obsolete / duplicating methods in `com.intellij.openapi.util.text.StringUtil`.

Namely:
`replaceChar(String,char,char)` was replaced with `String.replace(char,char)`
`charsEqual(char,char,boolean)` was replaced with `charsMatch(char,char,boolean)`
`charsEqualIgnoreCase(char,char)` was replaced with `charsMatch(char,char,boolean)`

Also usages of `indexOf(CharSequence,char)` applied to `String` were replaced with `txt.indexOf(char)`.


amaembo:
Pushed long ago (see bb8ba29 )

PR #571, opened 2017-05-15
IDEA-172832. Support long module names in Find in Path.

kunickiaj:
Sets the preferredWidth to the length of the longest module name in the selected project.


Vassiliy-Kudryashov:
The issue was fixed as https://youtrack.jetbrains.com/issue/IDEA-172116

PR #575, opened 2017-05-23
+ added Git rebase current branch action directly from VCS log

dmitry-zhuravlev:
Added Git rebase current branch action directly from VCS log.
It is more convinient to do rebase directly from VCS log because user can choose some commit (point in history) and initiate rebase to it.
This is a common use case when an user do a lot of "reword" and "edit" rebases.

### DIFF
--- a/plugins/git4idea/src/META-INF/plugin.xml
+++ b/plugins/git4idea/src/META-INF/plugin.xml
@@ -109,6 +109,7 @@
     <action class="git4idea.actions.GitCreateTagAction" id="Git.CreateNewTag" text="New Tag..."
             description="Create new tag pointing to this commit"/>
     <action id="Git.Reset.In.Log" class="git4idea.reset.GitResetAction" text="Reset Current Branch to Here..." />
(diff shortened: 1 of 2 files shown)


dmitry-zhuravlev:
untitled

klikh:
This action should be enabled only for unpushed commits (more precisely, for commits not pushed to a protected branch), like it is already done for `GitUncommitAction`.

I'm just about to push new "Reword" action, implementation of which generalizes this "enabled only for unpushed commits" stuff. I will take a look at the possibility to integrate this general interactive rebase action after that.

dmitry-zhuravlev:
@klikh Thanks for review!

>more precisely, for commits not pushed to a protected branch

Indeed that would be good to restrict. 
Also found the following small typo to notice:
 `ParametersListUtil.COLON_LINE_PARSER` in fact it should be named `SEMICOLON_LINE_PARSER` since `;` used like a delimiter. The same applied for `ParametersListUtil.COLON_LINE_JOINER`

klikh:
I've added the action in 197d30515085819bb440283740b77b9e935e2697 with some simplifications to the pull request, since we already have the `GitCommitEditingAction` which encapsulates the logic of preventing modification of pushed commits.

Thanks for the contribution!

PR #576, opened 2017-05-23
Several fixes in UAST value analysis including

mglukhikh:
* https://youtrack.jetbrains.com/issue/IDEA-171712
* Correct value caching / cache dropping
* More correct evaluation for local variables


yole:
Merged manually.

PR #577, opened 2017-05-26
Added support for anchors and aliases (YAML)

semonte:
This commit adds navigation support for YAML anchors and aliases.

For example, a user is able to navigate from "*example" to "&example"
in the following YAML by placing cursor at "*example":

```
example: &example
  foo: bar
test: *example
```

More about YAML anchors at http://www.yaml.org/spec/1.2/spec.html

I discovered the need for this feature from https://github.com/zalando/intellij-swagger. Navigating YAML anchors help users to write Swagger specs which tend to be big.


valich:
Thanks! I'll review the PR soon, just leaving a an issue on YT here: https://youtrack.jetbrains.com/issue/RUBY-12492

plugins/yaml/src/org/jetbrains/yaml/references/YAMLAliasReference.java
```diff
@@ -0,0 +1,55 @@
...
+  private static final String ANCHOR_PREFIX = "&";
+
+  public YAMLAliasReference(@NotNull PsiElement element) {
+    super(element);
+  }
+
+  @Nullable
+  @Override
+  public PsiElement resolve() {
+    Collection<YAMLScalar> values = PsiTreeUtil.findChildrenOfType(myElement.getContainingFile(), YAMLScalar.class);
```
valich: This approach works correctly, however, from my point of view, it solves only the resolution task properly while some further features based on it will work slow, basically O(n^2) in some cases where n ~ file size.

I think what we need to add here is proper lexing of anchors and refs (aliases) like tags are being lexed right now (see `org/jetbrains/yaml/lexer/yaml.flex:161` and usages); it will eliminate the need to instantiate text of the scalars in the whole document.

What do you think?
If you don't like the idea of digging into lexer and parser, I think I can help with that :)

semonte: Yes, I also had similar thoughts after seeing the YT ticket; this needs to be solved more comprehensively. I do like the idea of trying out lexer and parser, I never had a real need until now :) Let me check them out and write up a proposal how this could be done. This might take few weeks though as I have my daily work as well!

semonte: @valich I played around a bit with JFlex, what a great tool! I added YAML anchor support in YAML flexer by modifying the `KEY_block` regex to this

```
KEY_block = {NS_PLAIN_ONE_LINE_block} {WHITE_SPACE_CHAR}* ":" (\ &[^\[\]\{\}\n\t\r\ ]*)?
```
Sample:
```
node1:
    node11: &node11
        node111: value111
        node112: value112
    node12: *node11
```

image

The anchor is part of the `YAMLKeyValue`'s key. We could then add methods such as `isAnchorKey()` and `getAnchorName()` to `YAMLKeyValue`. 

Should we create a new class for the anchors and have a method like this (?): 

```
@Nullable
  public PsiElement getAnchor() {
    final PsiElement result = findChildByType(YAMLTokenTypes.ANCHOR);
    if (result != null) {
      return result;
    }
    return null;
  }

Or is it enough to have the anchor information in the key as String?

valich: @semonte Your solution does not cover all cases of anchors usage. According, to the spec:
* anchors may be present not only after colon and not only in KVs;
* anchors may be located with tag, but _after_ it.

In my opinion, it makes sense to add the mirror rules (like tag ones) for anchor both in lexer and parser.

I am leaving for a vacation for a couple of weeks now so I am sorry for the long delay in advance :)

semonte: I am also on vacation for the next few weeks so no progress from side either :)

semonte:
I updated the PR files. I modified the lexer to detect anchor tokens. Following anchors are supported:

image

I introduced a new interface which YAML elements can implement if they support anchors. `YAMLKeyValue` and `YAMLSequenceItem` extend this interface. 

Feedback is appreciated!

valich:
@semonte Sorry, it looks I completely lost this PR :(( Anyway, one of my coworkers has been introducing massive improvements to YAML subsystem lately, and the anchors/aliases [are supported thoroughly](https://youtrack.jetbrains.com/issue/RUBY-12492#focus=streamItem-27-2885974-0-0), including refactorings on them.
I'll close this PR now.

PR #578, opened 2017-05-27
Simplify logging output options to three, matching the options availa…

arvindanekal:
Signed-off-by: Arvind Ranganathan <arvindanekal@gmail.com>

This fix constrains the output options to just three (debug, info and warn) matching the options available in apache maven (--debug, --quiet and for no option, everything). This avoids ambiguity and makes the UI simpler. Existing tests work fine and no new tests were needed.

This credits for the idea of the fix goes to Timothy Basanov (https://github.com/timothybasanov)

### DIFF
--- a/plugins/maven/src/main/java/org/jetbrains/idea/maven/execution/MavenExternalParameters.java
+++ b/plugins/maven/src/main/java/org/jetbrains/idea/maven/execution/MavenExternalParameters.java
@@ -527,6 +527,11 @@ public class MavenExternalParameters {
     if (coreSettings.getOutputLevel() == MavenExecutionOptions.LoggingLevel.DEBUG) {
       cmdList.add("--debug");
     }
+    if (coreSettings.getOutputLevel() == MavenExecutionOptions.LoggingLevel.WARN) {
+      cmdList.add("--quiet");
+    }
+
+
     if (coreSettings.isNonRecursive()) {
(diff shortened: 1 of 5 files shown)


plugins/maven/src/main/java/org/jetbrains/idea/maven/execution/MavenExternalParameters.java
```diff
@@ -527,6 +527,11 @@ private static void encodeCoreAndRunnerSettings(MavenGeneralSettings coreSetting
     if (coreSettings.getOutputLevel() == MavenExecutionOptions.LoggingLevel.DEBUG) {
       cmdList.add("--debug");
```
timothybasanov: Previous implementation seem to add exceptions as well (`-X` and `-e`).
Add it to debug level or create a separate trace level with all the output?
Or may be exception only on debug level and `--debug` on trace level?

arvindanekal: Have added a new level for error output (ERROR). The four logging levels are:

1. All (INFO, default)
2. Warnings and errors (WARN, -q, --quiet) 
3. Debug output (DEBUG, -X, --debug)
4. Error output (ERROR, -e, --errors)

ignatov:
@vladsoroka please review.

PR #579, opened 2017-05-30
IDEA-173612 added ant log colors configuration

zdary:
added

- ant log colors configuration and ant log icons, 
- PRIORITY_BRIEF renamed to PRIORITY_INFO to allign naming with ant library
- allowed -debug option for verbose action

I haven't found build/scripts/icons.gant file so that change was done manually.

/**
 * NOTE THIS FILE IS AUTO-GENERATED
 * DO NOT EDIT IT BY HAND, run build/scripts/icons.gant instead
 */


eugenezh:
Thanks! Merged manually with some corrections regarding artwork

PR #581, opened 2017-06-03
Add Nullable for YAMLKeyValue::getParentMapping to sync it with YAMLKeyValueImpl

Haehnchen:
npe issue because of not seeing nullable return value: https://github.com/Haehnchen/idea-php-symfony2-plugin/issues/942


ignatov:
@Haehnchen is it still valid?

ignatov:
cc @valich

valich:
This _is_ still actual. I've reviewed some of our usages of this method and found some newly appeared warnings. I've fixed them and also added "pure" contract to simplify checks/assertions.

@Haehnchen Thank you very much and sorry for the delay.

valich:
Merged.

PR #582, opened 2017-06-05
README.md: escape * and +

bmintz:
The globbing characters are treated as bullet points by GFM.

### DIFF
--- a/plugins/java-decompiler/engine/README.md
+++ b/plugins/java-decompiler/engine/README.md
@@ -14,8 +14,8 @@ Fernflower is licenced under the Apache Licence Version 2.0. It can be found her
 
 java -jar fernflower.jar [-<option>=<value>]* [<source>]+ <destination>
 
-* means 0 or more times
-+ means 1 or more times
+\* means 0 or more times\
+\+ means 1 or more times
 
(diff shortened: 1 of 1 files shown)


plugins/java-decompiler/engine/README.md
```diff
@@ -14,8 +14,8 @@ Fernflower is licenced under the Apache Licence Version 2.0. It can be found her
 
 java -jar fernflower.jar [-<option>=<value>]* [<source>]+ <destination>
 
-* means 0 or more times
-+ means 1 or more times
+\* means 0 or more times\
```
trespasserw: Thanks.
What is the trailing slash for?

bmintz: Now that I look at the rich diff, I see it's not necessary.
I thought that newlines were ignored at the end of lines, except when there's two of them (to create a new paragraph).
The \ was to force a \<br\> element but it seems that's inserted anyway.

trespasserw: No, you're right. Github makes this into two lines, but some other formatters merge them into one. There are other issues in the file - I'll clean 'em up after merging the PR. Thanks!

PR #583, opened 2017-06-06
Ignore idle threads of the common pool

dahlstrom-g:
When running tests, the leak checker was catching
ForkJoinPool.commonPool() idle threads, and was
failing.


cdracm:
pushed

PR #588, opened 2017-06-09
Avoid logging passwords in GradleExecutionHelper

jimgoog:
The GradleExecutionHelper was previously logging android keystore passwords.  This change will obfuscate the values of a parameter ending in '.password' within the `idea.log` file.


jimgoog:
Thanks @vladsoroka!

PR #590, opened 2017-06-11
UI: allow customizing the TabsUtil constants through PropertiesComponent

mallowigi:
It would be useful for plugin makers to be able to customize the tab vertical padding, border and underline height using the PropertiesComponent instead of hard-coded fixed values.

### DIFF
--- a/platform/platform-api/src/com/intellij/ui/tabs/TabsUtil.java
+++ b/platform/platform-api/src/com/intellij/ui/tabs/TabsUtil.java
@@ -15,16 +15,18 @@
  */
 package com.intellij.ui.tabs;
 
+import com.intellij.ide.util.PropertiesComponent;
+
 import javax.swing.*;
 
 /**
  * @author pegov
  */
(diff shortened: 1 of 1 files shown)


Vassiliy-Kudryashov:
I don't understand your case and the problem to be solved. It's so easy to spoil UI... 
As for IconProvider, it should return 16x16 icon (or 32x32 in case of HiDPI) otherwise we'll got a zoo from different providers in tabs and trees. Small icons should be wrapped (for example try IconUtil.IconSizeWrapper) to get proper size.

mallowigi:
This commit wasn't about the icon provider. The reference was from another case that apparently has been solved.

This change about tabs is making the vertical padding customizable by plugin makers, like myself. This way during component initialization I can simply invoke PropertiesComponent.setValue() and allow bigger tabs. This would be useful for implementing custom LAFs as I do.

mallowigi:
@bulenkov maybe you can help

PR #591, opened 2017-06-11
[Fernflower] autoboxing primitives should not always be ignored

apixandru:
This commit puts more restrictions on when we can safely ignore the boxing primitives.
The initial restrictions were a bit too loose. https://github.com/JetBrains/intellij-community/commit/cdecb09b4a5bdb443d2d7d49698348535ea06b78


apixandru:
I resolved the merge conflicts.

@gorrus did you get a chance to have a look at this pull request?

gorrus:
Not yet, but it is in the list, sorry for the delay.

plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/ExprProcessor.java
```diff
@@ -540,6 +540,7 @@ else if (cn instanceof LinkConstant) {
             }
 
             InvocationExprent exprinv = new InvocationExprent(instr.opcode, invoke_constant, bootstrap_arguments, stack, bytecode_offsets);
+            markPreviousResultUsedIfBoxingExpression(exprlist, exprinv);
```
gorrus: this looks like an overkill, isn't it better to add something like
`if (instance instanceof InvocationExprent) {
            ((InvocationExprent)instance).markUsingBoxingResult();
          }`
right before `instance.toJava...` inside the InvocationExprent.toJava?

apixandru:
Yes, it does feel like overkill, the new tests pass with your suggested improvement. 

The reason why I had made it so specific was to make sure that this doesn't potentially break any other scenario that I can't think of right now so I made it as specific as I could get.

I removed the highly specific boxing restrictions using another commit rather than amending the previous one just in case it turns out that this is too broad and someone in the future will go through the commit log to see these original restrictions. If you prefer that I squish these commits into one please let me know and I will.

gorrus:
Hi, thanks for changes, please make it in one commit, thanks!

apixandru:
done

gorrus:
Thanks! Manually merged into master.

PR #592, opened 2017-06-15
git annotate show author time instead of committer time

seansonggi:
default git blame command seems to display author time. why does intellij idea git annotate display committer time?


klikh:
Thank you for the pull request. It addresses https://youtrack.jetbrains.com/issue/IDEA-85484. 
Unfortunately simply changing commit time to author time is not correct, because the time is used for revision coloring, and author time can't be used for this. So, display time and sorting time should be different ones in this case.
We'll try to address the issue soon.

PR #594, opened 2017-06-19
Corrected typo

SainathB:
Corrected typo in the constructor.


sproshev:
Our skeletons are no longer supported, we're moving to [typeshed](https://github.com/python/typeshed). Stubs for `datetime` module are enabled in PyCharm.

PR #595, opened 2017-06-21
Fix hierarchy export to file indentation

gharrma:
If a hierarchy node had a null descriptor, indentation
was still appended to the string builder, leading to
malformed indentation later on.

Originally a fix for a bug reported to Android Studio, now
submitting upstream. (For questions, email gharrma@google.com)


akozlova:
merged manually. thanks!

PR #596, opened 2017-06-23
Fix for performance problem IDEA-174594

mglukhikh:
First commit is mandatory and two others are just helpful


yole:
Merged manually.

PR #598, opened 2017-07-01
PY-24987/PY-22496 (probably others): Search class attributes in implicit ref resolve and 'goto symbol'

fitermay:
https://youtrack.jetbrains.com/issue/PY-24987


python/src/com/jetbrains/python/PyGotoSymbolContributor.java
```diff
@@ -45,6 +42,7 @@
     symbols.addAll(PyModuleNameIndex.getAllKeys(project));
     symbols.addAll(StubIndex.getInstance().getAllKeys(PyFunctionNameIndex.KEY, project));
     symbols.addAll(StubIndex.getInstance().getAllKeys(PyVariableNameIndex.KEY, project));
+    symbols.addAll(StubIndex.getInstance().getAllKeys(PyClassAttributesIndex.KEY, project));
```
vlasovskikh: We don't suggest navigating to instance attributes here. I would either suggest both instance and class-level attributes or neither of these.

fitermay: I'm sort of scared of add instance attributes here for performance reasons in large projects. Can we put a limit on the max results we return?

vlasovskikh: `getAllKeys(PyClassAttributesIndex.KEY, project)` already fetches the names of all the attributes of all classes, including instance attributes and methods. Only those elements that match the pattern in `getItemsByName()` are used for creating a navigation item. So in the worst case we'll create navigation items for all the instance attributes that contain a single letter.

Instance attributes are cached in a class in `myInstanceAttributes` so it's not that time-consuming to find an attribute by its name.

python/src/com/jetbrains/python/psi/impl/references/PyQualifiedReference.java
```diff
@@ -158,6 +159,11 @@ private void addImplicitResolveResults(String referencedName, ResolveResultList
       }
     }
 
+    PyClassAttributesIndex.findAttributes(referencedName, project, scope).forEach(classAttribute -> {
```
vlasovskikh: If we add instance attributes here alongside with class-level attributse, we could drop `PyInstanceAttributesIndex` completely.

python/src/com/jetbrains/python/psi/stubs/PyClassAttributesIndex.java
```diff
@@ -30,6 +33,20 @@
     return StubIndex.getElements(KEY, name, project, GlobalSearchScope.allScope(project), PyClass.class);
   }
 
+  public static Collection<PyTargetExpression> findAttributes(@NotNull String name, @NotNull Project project, GlobalSearchScope scope) {
```
vlasovskikh: Since this index also contains methods, it would be a good idea to name this method `findClassAttributes`.

fitermay:
Fixed according to review

python/src/com/jetbrains/python/psi/stubs/PyClassAttributesIndex.java
```diff
@@ -30,6 +34,36 @@
     return StubIndex.getElements(KEY, name, project, GlobalSearchScope.allScope(project), PyClass.class);
   }
 
+  public static Collection<PyTargetExpression> findClassAtrributes(@NotNull String name,
```
vlasovskikh: There is no need in distinguishing class and instance attributes since for all usages we are interested in both. It won't be hard to split getting class / instance attributes into two methods if we ever need it, but for now I would prefer a simpler version with just one method.

python/src/com/jetbrains/python/psi/impl/references/PyQualifiedReference.java
```diff
@@ -158,16 +158,13 @@ private void addImplicitResolveResults(String referencedName, ResolveResultList
...
-    for (Object attribute : attributes) {
-      if (!(attribute instanceof PyTargetExpression)) {
-        FileBasedIndex.getInstance().scheduleRebuild(StubUpdatingIndex.INDEX_ID,
-                                                     new Throwable(
-                                                       "found non-target expression object " + attribute + " in target expression list"));
-        break;
-      }
-      ret.add(new ImplicitResolveResult((PyTargetExpression)attribute, getImplicitResultRate((PyTargetExpression)attribute, imports)));
-    }
+    PyClassAttributesIndex.findClassAtrributes(referencedName, project, scope).forEach(classAttr -> {
```
vlasovskikh: Let's drop `PyInstanceAttributeIndex` from the codebase since it's not used any longer.

fitermay:
I've made the changes but maybe we can do even better if we can avoid creating stubs for instance attributes entirely by getting rid of:

```
 if (functionNode != null && qualifierNode != null) {
      final PsiElement function = functionNode.getPsi();
      if (function instanceof PyFunction && PyNames.NEW.equals(((PyFunction)function).getName())) {
        return true;
      }
      final ASTNode parameterList = functionNode.findChildByType(PyElementTypes.PARAMETER_LIST);
      assert parameterList != null;
      final ASTNode[] children = parameterList.getChildren(PyElementTypes.FORMAL_PARAMETER_SET);
      if (children.length > 0 && children[0].getText().equals(qualifierNode.getText())) {
        return true;
      }
    }
```

Inside 'PyTargetExpressionElementType'. However, I'm not sure if this is needed by any custom stub types or not.

vlasovskikh:
@fitermay The stubs for instance attributes are used heavily. For example, look into `PyClassImpl.findInstanceAttribute()`: it comes down to `PyClassImpl.getTargetExpressions(PyFunction)` that finds the target expression children of the function stub.

vlasovskikh:
@fitermay Thanks for your PR! I've cherry-picked it into master. It will go through QA and is likely to appear in the next major version after 2017.2.

PR #600, opened 2017-07-07
IDEA-175089 Move updateHistoryEntry call

dahlstrom-g:
… from TextEditorComponent.dispose to EditorWindow.clear.

Calling this method updates the last used editor to the current editor.
In TextEditorComponent.dispose, the editor may already be removed from
an EditorWithProviderComposite, causing updateHistoryEntry to record the
wrong editor as most recent for the file.


dahlstrom-g:
resolved by commits 99b53ca0 and 050ffff

PR #601, opened 2017-07-09
IDEA-154288 - Gradle import: Support jdkName property in IdeaModule

ingokegel:
In

https://github.com/gradle/gradle/commit/8e02bc268f6478773ca6daf49bea9aed14c5d1eb

support for this property was added to the Gradle tooling API, now
it can be used for the Gradle import in IDEA in order to correctly
select the JDK that was configured by the user in the gradle project.

This is important for projects that have modules with different
language levels.


vladsoroka:
Thank you for the contribution! I'll merge it soon.

ingokegel:
Thank you! I'm looking forward to using this in IJ.

PR #603, opened 2017-07-19
Enhance Java builder completion for superclasses and generics

kiiadi:
Adding the ability for builder completion to understand when the `build` method is defined in an interface, and also enabling it to handle generic types. Tested on the following sample code: 

A base-builder defined in an interface: 

```java 
public interface BaseBuilder<T> { 
    T build(); 
} 
``` 

An implementation as follows: 

```java 
public class MyObject { 

    static Builder builder() { 
        return newDefaultBuilder(); 
    } 

    public interface Builder extends BaseBuilder<MyObject> { 
    } 

    static class DefaultBuilder implements Builder { 

        @Override 
        public MyObject build() { 
            return new MyObject(); 
        } 
    } 
} 
``` 

The builders in V2 of the AWS Java SDK are structured in this way, see [interface](https://github.com/aws/aws-sdk-java-v2/blob/shorea-netty-default-thread-factory/utils/src/main/java/software/amazon/awssdk/utils/builder/SdkBuilder.java) and [example implementation](https://github.com/aws/aws-sdk-java-v2/blob/shorea-netty-default-thread-factory/core/src/main/java/software/amazon/awssdk/client/builder/ClientHttpConfiguration.java#L93).

### DIFF
--- a/java/java-impl/src/com/intellij/psi/filters/getters/BuilderCompletion.kt
+++ b/java/java-impl/src/com/intellij/psi/filters/getters/BuilderCompletion.kt
@@ -44,10 +44,11 @@ internal class BuilderCompletion(private val expectedType: PsiClassType, private
   }
 
   private fun methodsReturning(containingClass: PsiClass, returnedClass: PsiClass, isStatic: Boolean): Collection<List<PsiMethod>> {
-    return containingClass.methods
-      .filter { it.hasModifierProperty(PsiModifier.STATIC) == isStatic &&
-                returnedClass == PsiUtil.resolveClassInClassTypeOnly(it.returnType) &&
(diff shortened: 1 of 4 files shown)


kiiadi:
I'll add tests shortly - wanted to get this out to get some initial feedback.

kiiadi:
Now with tests!

donnerpeter:
Thanks! But why do you use substitutionMap, doesn't PsiSubstitutor#substitute solve the problem?

kiiadi:
@donnerpeter no good reason, didn't realise the substitute method did what I wanted. I've amended the code.

donnerpeter:
Thanks, merged manually!

PR #604, opened 2017-07-26
Improve diagnostic when failing to delete files in tests

dahlstrom-g:
* Before this change, a failure to delete a directory containing
  many files would be reported as a single error, for example:

java.io.IOException: Can't delete xxx
       (stack trace)

* After this change, a failure to delete a directory reports
  as many errors as there were files that could not be deleted,
  with the original IOException (and message) containing a more
  descriptive error message, for example:

------------------------------
[0]: [myDirectory]\myfile.txt: The process cannot access the file because it is being used by another process.
[1]: [myDirectory]
------------------------------

CompositeException (2 nested):
------------------------------
[0]: java.nio.file.FileSystemException: [myDirectory]\myfile.txt: The process cannot access the file because it is being used by another process.

       (stack trace)

[1]: java.nio.file.DirectoryNotEmptyException: [myDirectory]
       (stack trace)

------------------------------


donnerpeter:
Thanks, merged manually!

PR #605, opened 2017-07-27
Fix WideSelectionListUI for *_WRAP orientation

vcaen:
The WideSelectionListUI that fixes the bug
https://youtrack.jetbrains.com/issue/IDEA-160777
was causing another bug on Linux and Window where the background were
drawn over component on the left of the one selected in VERTICAL_WRAP
or HORIZONTAL_WRAP.

This change just skip the fix if the layout orientaion is not VERTICAL

Bug: https://youtrack.jetbrains.com/issue/IDEA-176759

Test: This fix a bug inside the paint() method and can't be
automatically tested.

### DIFF
--- a/platform/platform-api/src/com/intellij/ui/components/WideSelectionListUI.java
+++ b/platform/platform-api/src/com/intellij/ui/components/WideSelectionListUI.java
@@ -39,6 +39,13 @@ final class WideSelectionListUI extends BasicListUI {
                            ListModel model,
                            ListSelectionModel selectionModel,
                            int leadSelectionIndex) {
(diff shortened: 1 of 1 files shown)


SergeyMalenkov:
Thank you for the fix, but it can be improved:
we should not repaint a whole line if layout orientation is not vertical.
http://upsource.jetbrains.com/idea-ce/review/IC-CR-78

vcaen:
Nice! Thanks! 
I wanted my edit to have the smallest impact possible to avoid breaking anything considering the wide usages of JBList.

PR #608, opened 2017-08-03
PY-25491: Fix PyImportElement.getAsName incorrectly returning null

fitermay:


fitermay:
@vlasovskikh A more correct fix would be to fix `shouldCreateStub` in `PyTargetExpressionElementType` but it would require forcing everyone updating to reindex...

fitermay:
@vlasovskikh Decided to fix it properly by adding the stubs. Also added test

vlasovskikh:
@fitermay Thanks for your investigation and your pull request! It helped me a lot with debugging this problem.

We don't store target expressions for all import statements, not only for function-local imports. We actually store "as" names in the stubs for `PyImportElement` objects. The problem was that after partially switching to green stubs we now got the situation where the `getStub() == null` check doesn't guarantee that `getPsiOrChild*()` methods will use the AST representation. I've fixed this problem by switching to AST-based search for this case since most of the methods already use the stub-aware `getAsName()` instead of `getAsNameElement()` that now becomes AST-based.

My fix will be in master in a few hours.

fitermay:
@vlasovskikh Hi, I started with the same fix as you're proposing but I believe it is better to generate the stub.

-- "We don't store target expressions for all import statements, not only for function-local imports. We actually store "as" names in the stubs "

> I'm not sure this is correct.  If you follow the logic of 'shouldCreateStub', it will create a stub for all target expressions that have no parent function and no qualifier. You can also verify this using a debugger. So for consistency sake it is better to generate a stub for all as imports IMO.

--The problem was that after partially switching to green stubs we now got the situation where the getStub() == null

> Yes, I realize the problem manifested due to partial switch to green stubs.  That broke 'getAsName' in these cases. However, 'getAsNameElement' was always broken for inner imports, although it didn't have much consequences.

vlasovskikh:
@fitermay Thanks for your comment. You're right, PyCharm currently creates stubs for `PyTargetExpression` elements inside `PyImportElement` elements at the file level. But this is unintentional. We don't use these target expressions because the "as" names are already stored in the stubs for `PyImportElement`. Having stubs within functions is meaningless (at least that this pre-green stubs stage) since these names aren't exported from the file they are defined in, so the only way to access them is to have this file as AST. We have stubs for `PyTargetExpressions` inside functions only for instance and class attributes.

A more appropriate thing to do would be to remove `PyTargetExpression` stubs inside import elements from the file level. It won't affect performance since code inspections only access "as" names via `PyImportElement.getAsName()` or they already have the AST for the current file being analysed so there are no stubs anyway.

I'm not removing these file-level stubs inside import elements now because we don't want the minor updates to cause the re-indexing of the whole project.

fitermay:
@vlasovskikh 
Hi,
You're right , 'getAsNameElement()' inside a function does not appear to be ever accessed from a non-ast element because it is not an exported symbol. However, 'getAsNamedElement' in general is accessed from a non-ast context.  Through iterateNames, getVisibleElement and so on. So I believe 6291058 will likely cause performance issues.

Example stack:
{code}
	  at com.jetbrains.python.psi.impl.PyImportElementImpl.getAsNameElement(PyImportElementImpl.java:77)
	  at com.jetbrains.python.psi.impl.PyImportElementImpl.getVisibleName(PyImportElementImpl.java:113)
	  at com.jetbrains.python.psi.resolve.PyResolveProcessor.execute(PyResolveProcessor.java:79)
	  at com.jetbrains.python.psi.impl.PyFileImpl$ExportedNameCache.multiResolve(PyFileImpl.java:161)
	  at com.jetbrains.python.psi.impl.PyFileImpl$ExportedNameCache.access$300(PyFileImpl.java:74)
	  at com.jetbrains.python.psi.impl.PyFileImpl.lambda$multiResolveName$0(PyFileImpl.java:420)
	  at com.jetbrains.python.psi.impl.PyFileImpl$$Lambda$1231.1674518324.compute(Unknown Source:-1)
	  at com.intellij.openapi.util.RecursionManager$2.doPreventingRecursion(RecursionManager.java:113)
	  at com.intellij.openapi.util.RecursionManager.doPreventingRecursion(RecursionManager.java:68)
	  at com.jetbrains.python.psi.impl.PyFileImpl.multiResolveName(PyFileImpl.java:419)
	  at com.jetbrains.python.pyi.PyiFile.multiResolveName(PyiFile.kt:36)
	  at com.jetbrains.python.psi.impl.PyFileImpl.multiResolveName(PyFileImpl.java:413)
	  at com.jetbrains.python.psi.impl.PyFileImpl.getElementNamed(PyFileImpl.java:452)
	  at com.jetbrains.python.psi.impl.PyBuiltinCache.getByName(PyBuiltinCache.java:255)
	  at com.jetbrains.python.inspections.PyShadowingBuiltinsInspection$Visitor.processElement
{code}

vlasovskikh:
@fitermay I'm not sure that I got you. `PyImportElementImpl` defines only `getAsNameElement()` and it's public, there is no `getAsNamedElement()`. The `getPresentation()` method may be invoked quite often in "Navigate to ..." completion lists, and we do have some performance problems for `PyClass` by the way. But import elements are never shown in these "Navigate to ..." actions.

For clarity `getPresentation()` and `iterateNames()` should be rewritten to use `getAsName()` or `getVisibleName()` (I guess the latter might be more appropriate). But it's a minor issue, since it doesn't really affect performance.

fitermay:
@vlasovskikh I meant 'getAsNameElement' . I agree it's not a problem for `getPresentation` because it's not invoked. `iterateNames`  actually causes very many unnecessary stub to ast conversions. You can check this is true for by testing with any file that has star imports.

fitermay:
@vlasovskikh I can send you a PR:

- Changing iterateNames to use iterateNames and getVisibleName
- I'll also take a look at PyClass

I'm dealing with super large projects , so performance is important

vlasovskikh:
@fitermay Yes, `iterarateNames()` cannot be rewritten without changing the `PyImportedNameDefiner` interface, but again it isn't actually invoked from a non-AST context. `getVisibleName()` is still a good candidate for fixing.

Improving the performance of `PyClassImpl.getPresentation()` requires adding the text of the superclass list into the class stub. We already have some plans for it in the context of [PY-25655](https://youtrack.jetbrains.com/issue/PY-25655), so I'm not sure it's worth looking into right now.

fitermay:
@vlasovskikh The problem is that iterarateNames is invoked in non-ast context by `PyUnresolvedReferencesInspection` in `collectUnusedImportElements` . It causes stubbing of many files when star imports are present.

vlasovskikh:
@fitermay Indeed.

We can change it to:

```java
@Override
@Nullable
public String getName() {
  return getVisibleName();
}

@NotNull
public Iterable<PyElement> iterateNames() {
  return Collections.singletonList(this);
}
```

since the semantics of `iterateNames()` is not that well defined and this implementation will be enough.

PR #609, opened 2017-08-10
IDEA-146215 SILENT Install should use admin parameters, if installed …

James-Dengel:
This will install the application silently with admin if the user has run the installer as an admin else it will do what it did previously and install as a normal user.


klikh:

James-Dengel:
I am in the process of doing that.

ignatov:
@James-Dengel Already fixed, thanks for PR.

PR #610, opened 2017-08-11
UAST: Fix receiverType for JavaUCallExpression

4u7:
Return correct receiver class for a method call with implicit receiver instead of super class, when method defined in one of super classes

#IDEA-177417 Fixed


yole:
Merged manually.

PR #611, opened 2017-08-14
Move 'dropSelf' to PyFunctionType

fitermay:
@sproshev  In our framework we have decorators that add certain attributes to the functions that they decorate. To handle these decorators my plugin returns a PyFunctionType from 'provider.getCallableType' that handles these extensions.  I do not want to return a simple 'PyCallableType' as it isn't handled as well as 'PyFunctionType' by the rest of PyCharm code. 

Anyways, these used to work in 2016.3. In 2017 the change '804354727e46f3c1f02b23faa264f6b1a80c4837' broke it because 'dropSelfForInstanceMethod' drops the concrete implementation of PyFunctionType always replacing it with 'PyFunctionTypeImpl'

This change delegates 'dropSelf' to PyFunctionType so that the concrete class can be preserved.

### DIFF
--- a/python/psi-api/src/com/jetbrains/python/psi/types/PyFunctionType.java
+++ b/python/psi-api/src/com/jetbrains/python/psi/types/PyFunctionType.java
@@ -31,5 +31,6 @@ public interface PyFunctionType extends PyCallableType {
   @NotNull
   PyCallable getCallable();
 
+  @NotNull
   PyFunctionType dropSelf(@NotNull TypeEvalContext context);
 }
--- a/python/psi-api/src/com/jetbrains/python/psi/types/PyFunctionType.java
+++ b/python/psi-api/src/com/jetbrains/python/psi/types/PyFunctionType.java
@@ -30,4 +30,6 @@ public interface PyFunctionType extends PyCallableType {
    */
   @NotNull
   PyCallable getCallable();
+
+  PyFunctionType dropSelf(@NotNull TypeEvalContext context);
 }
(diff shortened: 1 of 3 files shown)


python/src/com/jetbrains/python/psi/types/PyFunctionTypeImpl.java
```diff
@@ -171,4 +171,13 @@ public void assertValid(String message) {
   public PyCallable getCallable() {
     return myCallable;
   }
+
+  public PyFunctionType dropSelf(@NotNull TypeEvalContext context) {
```
sproshev: Let's mark it as returning `@NotNull`

python/psi-api/src/com/jetbrains/python/psi/types/PyFunctionType.java
```diff
@@ -30,4 +30,6 @@
    */
   @NotNull
   PyCallable getCallable();
+
+  PyFunctionType dropSelf(@NotNull TypeEvalContext context);
```
sproshev: Let's mark it as returning `@NotNull`

python/src/com/jetbrains/python/psi/impl/PyReferenceExpressionImpl.java
```diff
@@ -485,11 +485,7 @@ private static PyType dropSelfForInstanceMethod(@Nullable PyType type,
       final PyFunctionType functionType = (PyFunctionType)type;
 
       if (qualifier != null && PyCallExpressionHelper.isQualifiedByInstance(functionType.getCallable(), qualifier, context)) {
-        final List<PyCallableParameter> parameters = functionType.getParameters(context);
-
-        if (!ContainerUtil.isEmpty(parameters) && parameters.get(0).isSelf()) {
-          return new PyFunctionTypeImpl(functionType.getCallable(), ContainerUtil.subList(parameters, 1));
-        }
+         return ((PyFunctionType)type).dropSelf(context);
```
sproshev: Please use `functionType` defined above

fitermay:
@sproshev applied changes and rebased onto latest master

sproshev:
LGTM. Let's wait for approvement from @vlasovskikh

vlasovskikh:
@fitermay Thanks! I've cherry-picked the commits from this PR. Could you please consider introducing a `PyCallableMembersProvider` for your case?

fitermay:
@vlasovskikh That sounds like a good idea. I will check and see if it will work

PR #612, opened 2017-08-14
Create extension point for pregenerated skeletons

fitermay:
@vlasovskikh This patch allows plugins to provide their own pregenerated skeletons. In my case this is needed because in our environment every release contains a different packaged SDK.  Generating these skeletons takes a very long time because the SDKs contain a lot of binary packages and because in windows spawning processes is a very slow operation.

### DIFF
--- a/python/src/META-INF/python-core-common.xml
+++ b/python/src/META-INF/python-core-common.xml
@@ -671,6 +671,9 @@
     <extensionPoint qualifiedName="Pythonid.debugSessionFactory" interface="com.jetbrains.python.debugger.PyDebugSessionFactory"/>
     <extensionPoint qualifiedName="Pythonid.customPackageIdentifier" interface="com.jetbrains.python.psi.PyCustomPackageIdentifier"/>
     <extensionPoint qualifiedName="Pythonid.pythonSdkComparator" interface="com.jetbrains.python.sdk.PySdkComparator" />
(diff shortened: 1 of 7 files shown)


fitermay:
@vlasovskikh @traff Rebased against latest master and resolved with @traff's changes

fitermay:
@vlasovskikh Sorry, I haven't explained properly why I've done it this way. 

 We have a very large number of libraries bundled with the SDK. They do not change most of the time but on occasion they do. When they do change this API will allow detecting that and actually running a skeleton generator for that specific library instead of providing a pregenerated one.

Also, when running as a plugin there is no guarantee that the generator version of the pregenerated skeletons will match that if the platform. For those cases I will be to return a skeleton with a fake generator version. It's a bit of a hack but I think it will work well enough.

In short this API offers more flexibility of how the skeletons are provided than simply returning an archive.

vlasovskikh:
@fitermay Thanks for your explanation! We've slightly updated our pre-built skeletons handling in order to use the skeletons for a major Python even when minor versions don't match. Could you please rebase your  changes and then I'll merge your commits?

fitermay:
@vlasovskikh Thanks. Rebased on master and incorporated recent changes

vlasovskikh:
@fitermay Thanks! I've merged it into master.

PR #615, opened 2017-08-23
Add name ID for usage tree component

afwang:
UI tests benefit from having an identifier set on
the view components, since programming UI tests
to locate a UI component is much easier when using
an identifier.


cdracm:
pushed

PR #616, opened 2017-08-25
PY-25794: Prevent unnecessary stub to AST switches in Python inspections

fitermay:


vlasovskikh:
Thanks for your pull request! I've added stub->AST assertions for the project files and the Python SDK modules to the majority of inspections in a2033f8042c1161f4af989f2bdd13dce127b0a79.

I would like to fix the problem with the `PyImportElementImpl` in the minor updates for 2017.2 as a part of [PY-25491](https://youtrack.jetbrains.com/issue/PY-25491), so I would like to put your in a separate commit that will go into a minor update.

@fitermay If you don't mind, I will extract your fixes for `PyImportElementImpl` and `PyProtectedMemberInspection` into separate commits and create unit tests for them.

And as we've discussed in [PY-22215](https://youtrack.jetbrains.com/issue/PY-22215), @sproshev will put a simpler fix into master instead of your fix based on the original element visitor.

fitermay:
@vlasovskikh Yes, it makes sense to release any changes that involve reindexing separately. The changes are already in separate commits. You can just cherry pick them out.

fitermay:
@vlasovskikh Also, I didn't include tests in cases except for the stubs because I assumed the existing tests already cover testing the functionality.  The assertions that you added should also take care of checking for unstubbing. I'm curious what tests you have in mind.

vlasovskikh:
@fitermay After I added stub->AST check in the commit I've mentioned several tests started to fail, but non of them failed due to the unstubbing problems `PyProtectedMemberInspection` and `PyImportElementImpl`. I've added a test for your fix for `PyProtectedMemberInspection` and now I'm thinking of a test for `PyImportElementImpl`. Do you have any code examples in mind?

vlasovskikh:
My tests: 5e20034a0afd3528e2b6555f8fda7b92ccd65b68 and 5d6c451033cde78e628347b87e877c3b4f03c8e7.

PR #617, opened 2017-08-26
PY-13273: Fix args in decorators assigned to incorrect scope

fitermay:


python/src/com/jetbrains/python/codeInsight/dataflow/scope/ScopeUtil.java
```diff
@@ -176,4 +176,24 @@ public static ScopeOwner getDeclarationScopeOwner(PsiElement anchor, String name
     }
     return result;
   }
+
+  /*  This function compensates for the fact that scope children do not always match the AST children. For example, the decorator list is the
+      AST child of a function but is not in function scope. So AST children that are not scope children must be excluded.
+   */
+  public static void visitChildrenInScope(ScopeOwner owner, PyElementVisitor visitor) {
+    PyClass pyClass = PyUtil.as(owner, PyClass.class);
+    if (pyClass != null) {
```
vlasovskikh: `PyClass` and `PyFunction` inheritors could contain children other than `PyParameterList` and `PyStatementList`. It's better to exclude the decorator list here rather than include only specific nodes.

fitermay: @vlasovskikh Yes, it looks like 06af0d23c6ce4d does the same thing. Though, I believe that a function annotation should also be resolved in the outer scope.

vlasovskikh: @fitermay I believe it's already the case. Do you have any code examples that are not analyzed properly?

fitermay: @vlasovskikh All I'm trying to say is that annotations on functions should be skipped just like the decorator list because it doesn't belong in the function scope. I would need to comb through a lot of code to determine what, if anything, it breaks. However, it still makes no sense to add things to the function CFG/scope that don't belong there.

vlasovskikh: @fitermay I've found an example when it matters, but the example is quite obscure, see https://youtrack.jetbrains.com/issue/PY-26018

PR #618, opened 2017-08-26
Git4idea: pushurl without capital u

almegren:
In .git/config the remote parameter 'pushurl' is not written in camel case (although other git parameters are) and therefore the pushurl parameter is never loaded properly. As it defaults to the url value and the tests didn't cover this it has not yet been detected. The tests had faults in the test data and also used the equals() method in GitRemote which only compares the name so this was not covered by the tests.

A fix in the gerrit plugin,
https://github.com/uwolfer/gerrit-intellij-plugin/commit/d464a7aab689fccfd83654cdfab68048b59bcb9f ,
is dependent on this change to work properly.

### DIFF
--- a/plugins/git4idea/src/git4idea/repo/GitConfig.java
+++ b/plugins/git4idea/src/git4idea/repo/GitConfig.java
@@ -376,7 +376,7 @@ public class GitConfig {
 
     @NotNull
     private Collection<String> getPushUrls() {
-      return nonNullCollection(myRemoteBean.getPushUrl());
+      return nonNullCollection(myRemoteBean.getPushurl());
     }
 
     @NotNull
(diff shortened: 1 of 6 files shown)


klikh:
In fact, Git config variables are case insensitive, so this pull request unfortunately doesn't fix the real problem for which we have [an issue](https://youtrack.jetbrains.com/issue/IDEA-81364).

In the manpages, `pushurl` is written in lowercase indeed, so it might be considered as a better default than `pushUrl` until the case sensitivity issue is solved. But on the other hand, some people may already depend on the "pushUrl", and we don't want to break any functionality for them.

The correct fix to the issue should be making Git config variables case insensitive. It requires either patching `init4j` library which we use for parsing `.git/config`, or rewriting the whole parsing code to avoid using the library at all. We'd welcome a contribution for this issue.

almegren:
Thank you for quick feedback. I'll look into the existing issue.

klikh:
Just in case if you decide to try patching ini4j, here are the sources of our fork of the library: https://github.com/JetBrains/ini4j

PR #619, opened 2017-08-29
override SmartList.equals() for better performance

stsypanov:
Default implementation of AbstractList.equals() creates two Iterators and then compare their items one by one. For array-base implementations (ArrayList/SmartList) it's possible to enhance comparison by
1) comparing sizes
2) if lists have the same size then we can use get(i) instead of iterators


donnerpeter:
Why, is there any specific performance problem you're trying to address with this change?

stsypanov:
I've found 174 usages of List.equals() in the code of Idea CE where concrete instance of List is ArrayList. Changing ArrayList into SmartList probably gives some performance improvement supposing this patch is applied.

stsypanov:
Is it possible to exclude other commits from this pull-request or should I create new one?

donnerpeter:
Merged

PR #620, opened 2017-08-29
Ini4j replaced by custom parser for .git/config.

almegren:
Looking at [IDEA-81364](https://youtrack.jetbrains.com/issue/IDEA-81364) and the possibility to patch ini4j to allow case insensitivity in parameter names in .git/config I found that ini4j was a very generic library. That could of course be very nice but it seems very hard to patch and change. I guess the age of IDEA-81364 testifies to this. Since the format of .git/config is pretty straight forward and the current Intellij parsing of it only covers some parts it seems to me that it would be simpler way to replace it. I've here tried to replace it with a small custom parser. It doesn't cover everything, for example it doesn't handle quoted special characters etc, but it passes all the old test and one I added.

A review would be welcome and I could use some ideas of what tests to add to avoid unwanted consequences. I've made sure to only change the inner workings of GitConfig and therefore there are some ugly solution with private static inner classes etc. Could possibly be refactored to a better structure.

### DIFF
--- a/plugins/git4idea/src/git4idea/repo/GitConfig.java
+++ b/plugins/git4idea/src/git4idea/repo/GitConfig.java
@@ -15,27 +15,27 @@
  */
 package git4idea.repo;
 
-  import com.intellij.openapi.diagnostic.Logger;
-import com.intellij.openapi.util.Pair;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.util.ArrayUtil;
 import com.intellij.util.containers.ContainerUtil;
 import git4idea.GitLocalBranch;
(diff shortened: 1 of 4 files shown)


klikh:
We found out that `ini4j.Config` has `setLowerCaseOption`, so the fix became quite simple: 0b8b39ebc613004f05a4efe980d78d834ee14c0a

But thanks a lot for the attempt and for raising attention to the issue!

almegren:
Yes, I found that option too but didn't get it to work. I'll be happy to test your fix, however, I couldn't find your commit. Is it pushed?

klikh:
At the time when I wrote the comment, the commit was pushed to the original repository, but not mirrored to github yet. It should be available now.

PR #622, opened 2017-08-30
Add copy URL to clipboard for Gist creation dialog in GitHub plugin

h1nk:
Adds an additional check box in the Create Gist dialog to copy the URL to clipboard.
image

### DIFF
--- a/plugins/github/src/org/jetbrains/plugins/github/GithubCreateGistAction.java
+++ b/plugins/github/src/org/jetbrains/plugins/github/GithubCreateGistAction.java
@@ -24,6 +24,7 @@ import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.fileEditor.FileDocumentManager;
 import com.intellij.openapi.fileTypes.FileTypeManager;
+import com.intellij.openapi.ide.CopyPasteManager;
 import com.intellij.openapi.progress.ProgressIndicator;
 import com.intellij.openapi.progress.Task;
 import com.intellij.openapi.project.DumbAwareAction;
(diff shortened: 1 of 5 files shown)


plugins/github/src/org/jetbrains/plugins/github/GithubCreateGistAction.java
```diff
@@ -129,6 +132,12 @@ public void onSuccess() {
         if (dialog.isOpenInBrowser()) {
           BrowserUtil.browse(url.get());
         }
+        if (dialog.isCopyURLToClipboard()) {
+          StringSelection stringSelection = new StringSelection(url.get());
+
+          Clipboard clipboard = Toolkit.getDefaultToolkit().getSystemClipboard();
```
AMPivovarov: Please, use `CopyPasteManager.getInstance().setContents`.

AMPivovarov:
Hi,
Thanks for the submission.

How do you think, wouldn't it be better to replace these checkboxes with notification actions?
image

H1nk:
Ah, I knew there was a clipboard manager, just couldn't find it, thanks!

As for the notification actions suggestion, I do agree that it works a lot cleaner than adding an additional check box to the Gist creation dialog box, but I do personally like having it persistently remember the action. Seeing that the Open in browser action has both a check box and notification action would it be ok to do the same with the copy to clipboard action and have both a check box and notification action?

yzozulya:
Hi!
If you want this option to be persistently remembered maybe it is better to move the setting to VCS -> GitHub settings page?

h1nk:
It's kind of something I want to be like the open in browser action. If I selected it last time I want it to remember, and if I didn't last time I don't want it to the next time. It's not something I really want hidden in a settings page, I'd rather have it out front when I'm about to upload.

ignatov:
@h1nk please rebase the PR, cc @AMPivovarov.

iasemenov:
Required functionality was implemented based on this pull request and merged.
See commit `abc2fe4934f705cf0fb89ccec2c858a113c5dfd0`.
Thank you, @h1nk.

PR #623, opened 2017-08-31
Run all nested tests in delegate to gradle mode

semoro:


vladsoroka:
As discussed, this is not correct fix.

PR #625, opened 2017-09-11
Add elements into collection directly

stsypanov:
Direct adding of elements into collection by the means of `ContainerUtil.addAll` and `new SmartList(T...)` allows to reduce amount of garbage instances produced by `Arrays.asList`

See [https://youtrack.jetbrains.com/issue/IDEA-178648](https://youtrack.jetbrains.com/issue/IDEA-178648)


amaembo:
Merged, thanks.

PR #626, opened 2017-09-11
PY-8933: Fix try with import error false positives

fitermay:


vlasovskikh:
@fitermay Thanks for your PR! I've merged it with some changes. There is `ScopeUtil.getDeclarationScopeOwner` that is similar to your check for a name being defined in an outer scope. This function handles names defined via star-imports and similar wildcard name definers.

vlasovskikh:
FYI here is the commit with my fixes 47679f824c2947eb893506205e9cbd79ea1778e6.

fitermay:
@vlasovskikh 
    boolean nextIsTerminal =
 -          instruction.allSucc().stream().anyMatch(success -> success.getElement() == null && success.allSucc().isEmpty());

Was an optimization that prevents crawling up the AST for every instruction.  It's pretty expensive to do so for every instruction.

vlasovskikh:
@fitermay Could you elaborate please? We traverse the control flow graph starting from the `except` class reference until we are no longer a child of `except`. If we add the `nextIsTerminal` check then we will traverse all vertices inside `except` but the terminal ones. It's still pretty much all the vertices inside `except`, isn't it? Moreover, `nextIsTerminal` effectively traverses terminal vertices as well.

fitermay:
@vlasovskikh It's optimized for worst case. Let d be the depth of the AST and n be the number of instructions. With nextIsTerminal we only check isAncestor for terminal nodes, which is effectively constant , making thr algorithm worst case O(n+d) when we check every node being an ancestor then it is worst case O(n*d)

vlasovskikh:
@fitermay I agree with your analysis, but I have to mention that d and n are the depth and the control flow graph size inside `except ImportError` + the edges that start inside `except` and lead outside (we stop the traversal of the path that leads outside of `except` by returning `false`). So `controlFlowAlwaysTerminatesInsideGuard` traverses only a small part of the control flow graph inside `except`. Whereas `nextIsTerminal` forces us to traverse the whole control flow graph up until we reach the terminal vertex on all the paths, doesn't it?

vlasovskikh:
@fitermay And my conclusion is that it's better to have O(n*d) where n and d refer to the internals of `except ImportError` than to have `O(n+d)` where n and d refer to the whole contents of a control flow graph that contains `except ImportError`.

fitermay:
@vlasovskikh I agree that it is a different n in the two cases, I should have mentioned that.  My motivation for traversing the larger n is that I am assuming that traversing the CFG is very cheap as it is just indexed array access whereas performing instanceof checks up the AST is more expensive.

vlasovskikh:
@fitermay I'll stick with a simpler version for now since I believe that on average it's faster. If we ever encounter a real-world performance problem in this piece of code, I will be open to change it. Thanks for your feedback!

PR #627, opened 2017-09-12
Do not force reindex on content files with compatible language levels

fitermay:
We switch between 2.6 and 2.7 frequently. Currently this forces a reindex of the 
entire project (100k+ files). I believe this reindexing is not necessary for Python 
Language levels which are AST compatible.

### DIFF
--- a/python/src/com/jetbrains/python/psi/impl/PythonLanguageLevelPusher.java
+++ b/python/src/com/jetbrains/python/psi/impl/PythonLanguageLevelPusher.java
@@ -15,6 +15,7 @@
  */
 package com.jetbrains.python.psi.impl;
 
+import com.google.common.collect.Maps;
 import com.intellij.ProjectTopics;
 import com.intellij.diagnostic.PerformanceWatcher;
 import com.intellij.facet.Facet;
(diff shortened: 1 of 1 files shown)


python/src/com/jetbrains/python/psi/impl/PythonLanguageLevelPusher.java
```diff
@@ -65,11 +67,16 @@
...
+import static com.jetbrains.python.psi.LanguageLevel.*;
+
 /**
  * @author yole
  */
 public class PythonLanguageLevelPusher implements FilePropertyPusher<LanguageLevel> {
   public static final Key<LanguageLevel> PYTHON_LANGUAGE_LEVEL = Key.create("PYTHON_LANGUAGE_LEVEL");
+  public static final Set<Pair<Integer, Integer>> COMPATIBLE_LEVELS = Sets.newHashSet(
+    Pair.create(PYTHON26.ordinal(), PYTHON27.ordinal()),
+    Pair.create(PYTHON33.ordinal(), PYTHON34.ordinal()));
```
sproshev: Consider using `EnumMap`

python/src/com/jetbrains/python/psi/impl/PythonLanguageLevelPusher.java
```diff
@@ -166,11 +173,22 @@ public boolean acceptsDirectory(@NotNull VirtualFile file, @NotNull Project proj
 
   private static final FileAttribute PERSISTENCE = new FileAttribute("python_language_level_persistence", 2, true);
 
+  private static boolean areLanguageLevelsCompatible(int oldLevelOrdinal, int newLevelOrdinal) {
```
sproshev: I think it's more type safe to work with `Enum`, maybe `Nullable`

python/src/com/jetbrains/python/psi/impl/PythonLanguageLevelPusher.java
```diff
@@ -166,11 +173,22 @@ public boolean acceptsDirectory(@NotNull VirtualFile file, @NotNull Project proj
...
+    return COMPATIBLE_LEVELS.contains(Pair.create(minLevel, maxLevel));
+  }
+
   public void persistAttribute(@NotNull Project project, @NotNull VirtualFile fileOrDir, @NotNull LanguageLevel level) throws IOException {
     final DataInputStream iStream = PERSISTENCE.readAttribute(fileOrDir);
+    int oldLevelOrdinal = -1;
     if (iStream != null) {
       try {
-        final int oldLevelOrdinal = DataInputOutputUtil.readINT(iStream);
+        oldLevelOrdinal = DataInputOutputUtil.readINT(iStream);
```
sproshev: Let's convert int value to enum, or null if there is no enum value with such ordinal

python/src/com/jetbrains/python/psi/impl/PythonLanguageLevelPusher.java
fitermay: It's really would be more of a multimap since one language level can be compatible with several others. Unless I'm misunderstanding something?

sproshev: The type could be `EnumMap<LanguageLevel, EnumSet<LanguageLevel>>`

fitermay:
@sproshev I've made the suggested changes and rebased onto latest master. An EnumMap actually suffices because currently no level is compatible with more than one other.

traff:
Merged

PR #628, opened 2017-09-14
PY-26081: Make 'getProcessOutput' respect cancellation by ProgressIndicator

fitermay:


east825:
Thanks for the PR! I indeed didn't realize that Quick Documentation is actually opened under a special write action sensitive progress indicator coming from `ProgressIndicatorUtils#runInReadActionWithWriteActionPriority()`.

PR #630, opened 2017-09-23
IDEA-179472: Fix maven module not retrieving "release" version from config

theminecoder:


vladsoroka:
Thanks for the contribution, I'm sorry for the late answer.
However similar change was already added for another ticket, you can track it at https://youtrack.jetbrains.com/issue/IDEA-173143

PR #631, opened 2017-09-24
fix TODO in ChangesListView.java

stsypanov:
I found TODO in ChangesListView and fixed it. At runtime changelist view seems to be working as before changes were introduced.


AMPivovarov:
Thanks, merged.

AMPivovarov:
This cleanup introduces an issue with `WideSelectionTreeUI`. 
https://youtrack.jetbrains.com/v2/issue/IDEA-179663

Probably, this should be fixed in `DnDAwareTree`, but doing so would require checking for regressions all around IDE and under different OS and LaF.
So this pull request is reverted for now.

PR #632, opened 2017-09-27
PY-26239: Fix find class usages performance

fitermay:
Instead of searching for "__init__" text. Search for references named the same as the class but with __init__ method target


vlasovskikh:

PR #634, opened 2017-10-01
generate hashCode: prefer HashCodeBuilder with default constructor

bedla:
Hi,
hashCode generator uses template with builder class with two parameters constructors. There is default constructor with same constants as in current template. It should be handy not to duplicate this constants and use default constructor instead. Second point of view could be that when you have hard-coded constants many of static code analysis tools marks them as magic numbers.
Is it this change ok for you?
Thx,
Ivos


bedla:
YouTrack issue is https://youtrack.jetbrains.com/issue/IDEA-179868

ignatov:
Already fixed, thanks for PR.

bedla:
Sorry, but I have checked current IntelliJ IDEA 2018.3 EAP (Community Edition) and that constants are still there. Do i have to change some settings in IDEA?

PR #635, opened 2017-10-04
Extends IdeaModuleBase to support jars

dahlstrom-g:
IdeaModuleBase already extends a file set that supports zips or directories as inputs. This changes it so if the output of the model is a .jar, it will use the zip code path. It's unfortunate that the ant base class has two variables and expects only one of them to be set.

### DIFF
--- a/jps/antLayout/src/jetbrains/antlayout/datatypes/IdeaModuleBase.java
+++ b/jps/antLayout/src/jetbrains/antlayout/datatypes/IdeaModuleBase.java
@@ -26,6 +26,7 @@ import java.io.File;
  */
 public abstract class IdeaModuleBase extends LayoutFileSet {
   private String name;
+  private File output;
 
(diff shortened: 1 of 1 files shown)


chashnikov:
Sorry for the late reply, we somehow missed this pull request.
Why do you need this? `IdeaModuleBase` describes location of module classes, and it's used to pack them into JARs to build IDE distribution (see `org.jetbrains.intellij.build.impl.DistributionJARsBuilder`). If it refers to a JAR file it means that that JAR file will be packed into another JAR file in IDE distribution and this won't work. 

Also this change will cause problems if there is an existing module which output directory name ends with `.jar` (e.g. because the module name ends with `.jar`).

If you want to build a distribution of an IDE by taking compiled classes from JARs, this should be done at a different level, see `org.jetbrains.intellij.build.impl.CompilationTasksImpl#compileModules` which already does something similar.

PR #639, opened 2017-10-07
[Fernflower] prevent null to be cast as short/byte

apixandru:
Prevent casting null as primitive.

Instead of 
```
withInteger((Integer)null);
withShort((short)null);
withByte((byte)null);
new TestPrimitiveNarrowing((short)null);
```
We will have
```
withInteger((Integer)null);
withShort((Short)null);
withByte((Byte)null);
new PrimitiveOverloading((Short)null);
```


gorrus:
merged into master

PR #640, opened 2017-10-08
Don't display parameter names for functional interfaces

xfix:
Displaying `predicate.test(t: this)` is hardly useful.


ignatov:
@xfix please add a test for the fix.

ignatov:
@JB-Dmitry reassing to Roman Ivanov if needed.

PR #637, opened 2017-10-06
Fix for IDEA-180133: allow hg4idea to set hg executable per-project

mkillianey:
Details at https://youtrack.jetbrains.com/issue/IDEA-180133

PR #641, opened 2017-10-08
IDEA-144811

mkillianey:
Fix for issue https://youtrack.jetbrains.com/issue/IDEA-144811

There are currently application-wide settings for: (1) the path to the git executable and (2) the choice of ssh executable.  This change honors those existing settings, but allows individual projects to choose to override them.

I believe this also fixes/addresses these related issues:
https://youtrack.jetbrains.com/issue/IDEA-120335
https://youtrack.jetbrains.com/issue/IDEA-175515

Without regressing on this closed issue:
https://youtrack.jetbrains.com/issue/IDEA-66127

### DIFF
--- a/plugins/git4idea/src/git4idea/GitVcs.java
+++ b/plugins/git4idea/src/git4idea/GitVcs.java
@@ -104,7 +104,7 @@ public class GitVcs extends AbstractVcs<CommittedChangeList> {
   private final GitHistoryProvider myHistoryProvider;
   @NotNull private final Git myGit;
   private final ProjectLevelVcsManager myVcsManager;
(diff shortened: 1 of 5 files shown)


mkillianey:
Note:  I have signed the contributors agreement (see https://github.com/JetBrains/intellij-community/pull/637 )

Naid405:
Thank you for your help, @mkillianey 
Unfortunately there's a problem with the approach you suggested - not many users know about "default project" setup, so setting up application-wide executable can become a pain in the neck.
We have come up with a slightly different approach that allows to "override" current application-wide setting via a checkbox and will implement this shortly.

PR #644, opened 2017-10-13
IDEA-180548 Introduce PlatformUtils.isAndroidStudio method

sprigogin:


chashnikov:
We don't recommend to use PlatformUtils anymore, see [the issue](https://youtrack.jetbrains.com/issue/IDEA-180548) for details.

PR #645, opened 2017-10-13
IDEA-180447 Reduce UI clutter by hiding Synchronize action from menu and toolbar when it is unlikely to be used

sprigogin:


trespasserw:
Thank you for contribution, but we don't think it's a good idea.
More details in [the issue](https://youtrack.jetbrains.com/issue/IDEA-180447).

PR #646, opened 2017-10-13
IDEA-180614: Rename Synchronize action to Synchronize with File System

sprigogin:


ignatov:
cc @bulenkov

yole:
The action is now called "Reload from disk".

chashnikov:
We now have some problems with publishing new commits to GitHub repository, so the commit with the change isn't available here yet. I hope the problems will be fixed soon.

PR #647, opened 2017-10-14
Uses native temp file creation

dahlstrom-g:
This should fix the issue that prevents us from creating patches
in parallel. The old way when two patchers run at the same time
will likely return the same file for both to use.

### DIFF
--- a/updater/src/com/intellij/updater/Utils.java
+++ b/updater/src/com/intellij/updater/Utils.java
@@ -38,14 +38,12 @@ public class Utils {
     return fileName.endsWith(".zip") || fileName.endsWith(".jar");
   }
 
-  private static File findUniqueName(String path) {
-    int index = 0;
-    File myTempFile;
-    do {
-      myTempFile = new File(path + ".tmp." + index++);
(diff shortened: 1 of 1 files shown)


trespasserw:
AFAIU, the core problem is that two patchers may end up sharing the same temp. directory, right?

I'm ok with the 'getTempDir()' part of the patch, but would like to keep the old scheme inside the temp. directory - it makes update logs more readable.

estebandlc:
That is correct, when running patchers in parallel they hit the same directory and things go very wrong :)

trespasserw:
OK, I pushed a [minimal fix](https://github.com/JetBrains/intellij-community/commit/f8d33e720cb696170b53fe64be88606941c8fd60) that only changes the directory creation. If you don't see any problem in it, I will backport it to branch 173.

estebandlc:
Thanks! I believe the fix will be fine, as the directories will be different. I'd still caution the use of findUniqueName as is inherently not safe.

trespasserw:
There was only one usage left, so I just inlined the method.

PR #648, opened 2017-10-15
Duplicate entries in RESERVED_WINDOWS_NAMESPACE and absense of `con`

Maccimo:
RESERVED_WINDOWS_NAMESPACE contains duplicate entries of `aux` and doesn't contain `con` which is reserved name too.


trespasserw:
Thanks!

PR #651, opened 2017-10-20
Take moved lines into account

stephanenicolas:
This option is useful after a global formatting of a project. At least in Java, as imports can be easily moved by a formatter. (Using the spotless plugin + google java format as an example).


klikh:
Please take a look at the discussion in https://youtrack.jetbrains.com/issue/IDEA-133270
In short, the issue can't be fixed just by adding the "-M" option to git blame: it has a performance impact, and annotations are missing full commit comments. Performance issue might be worked around if we make the parameter optional, the other issue requires some work to be done (and also affects the performance).

stephanenicolas:
Thx for the link. Sorry for the simplistic approach.

S.

Le 21 oct. 2017 08:48, "Kirill Likhodedov" <notifications@github.com> a
écrit :

PR #652, opened 2017-10-22
make EmptyList compatible with java.utilCollections.EmptyList

stsypanov:
Current implementation of ContainerUtilRt.EmptyList invokes methods of java.util.AbstractList. As a result
- EmptyList.listIterator allocates new instance of ListIterator
- EmptyList.equals allocates two instances of ListIterator (for both compared lists)

These behaviour can be avoided by this patch.


donnerpeter:
Thanks, merged manually!

PR #654, opened 2017-10-25
Fix NPE when decompiling constructor

Maccimo:
Fix NPE when decompiling constructor with call to the absent constructor of the same class.


trespasserw:
@gorrus we need an option which would allow decompiling incorrect code (off by default, on in IdeaDecompiler).

gorrus:
@Maccimo please describe when do you need it. Why constructors are better than any other methods in this case?

Maccimo:
@gorrus The question is not clear. Need _what_?

As long as I can see only call to the absent constructor of the class from one of methods of class itself (not only `<init>`)  will cause decompilation problem.

Concerning suggestion of @trespasserw I personally think it would be handy to show disassembly in such cases.

gorrus:
@Maccimo I will clarify: in what cases do you decompile classes with missing constructor calls? Is it possible that other methods are missing but being called?

Maccimo:
This particular bug was revealed while debugging code generation tool.
Obviously, it is possible that any method or field may be missed while it is not causing class loading problems.

I saw other places in decompiler where result returned by `getMethodWrapper()` is dereferenced without null check, but not investigated that places yet.

gorrus:
manually merged, thanks for the contribution!

PR #656, opened 2017-10-31
Improve the doc-comment on PsiNameIdentifierOwner

klazuka:
While implementing reference resolve and rename in a custom language
plugin, I ran into a subtle bug caused by failing to override
`PsiElement#getTextOffset()`. I eventually figured it out by tracing
the code into `TargetElementUtil` and finding a useful comment about
the importance of overriding the text offset on named elements.


donnerpeter:
Merged manually, thanks!

PR #659, opened 2017-11-08
Splice side effect from String concatenation

stsypanov:
According to https://bugs.openjdk.java.net/browse/JDK-8043677 
```java
new StringBuilder().append(counter++).toString()
```
is slower than 
```java
int cnt = counter++; 
new StringBuilder().append(cnt).toString();
```
I've created a feature request https://youtrack.jetbrains.com/issue/IDEA-181103 (wtih measurement results) to have IDEA detecting such cases automatically and replaced them in the code of CE.


klikh:
Closing since it was [decided](https://youtrack.jetbrains.com/issue/IDEA-181103) not to implement this kind of inspections. 

But thanks for your contribution at any case!

PR #661, opened 2017-11-13
Making pydev compatible with Python 2.6

da1z:
Changed format string to keep pydev compatible with python 2.6


fitermay:
@traff @Elizaveta239 One character change needed to make console work against 2.6:)

Elizaveta239:
Thank you! Your fix was included into PyCharm 2017.3

PR #662, opened 2017-11-14
PY-25655: Store super classes text in stub

fitermay:
@vlasovskikh Wanted to get this in because it involves stub changes so it's only possible to get it in before a major release. Implementation is slightly strange to preserve compatibility.

### DIFF
--- a/python/psi-api/src/com/jetbrains/python/psi/stubs/PyClassStub.java
+++ b/python/psi-api/src/com/jetbrains/python/psi/stubs/PyClassStub.java
@@ -43,6 +43,7 @@ public interface PyClassStub extends NamedStub<PyClass> {
   @NotNull
   List<String> getSubscriptedSuperClasses();
 
+
   @Nullable
   QualifiedName getMetaClass();
 
@@ -51,4 +52,9 @@ public interface PyClassStub extends NamedStub<PyClass> {
 
   @Nullable
   String getDocString();
+
+  @Nullable // Made nullable for compatibility
+  default List<String> getSuperClassesText() {
+    return null;
+  }
 }
--- a/python/psi-api/src/com/jetbrains/python/psi/stubs/PyClassStub.java
+++ b/python/psi-api/src/com/jetbrains/python/psi/stubs/PyClassStub.java
@@ -25,6 +25,7 @@ import com.jetbrains.python.psi.PyClass;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -53,8 +54,8 @@ public interface PyClassStub extends NamedStub<PyClass> {
   @Nullable
   String getDocString();
 
-  @Nullable // Made nullable for compatibility
+  @NotNull
   default List<String> getSuperClassesText() {
-    return null;
+    return Collections.emptyList();
   }
 }
--- a/python/src/com/jetbrains/python/psi/impl/PyClassImpl.java
+++ b/python/src/com/jetbrains/python/psi/impl/PyClassImpl.java
@@ -29,8 +29,8 @@ import com.intellij.psi.search.SearchScope;
 import com.intellij.psi.stubs.IStubElementType;
 import com.intellij.psi.stubs.StubElement;
 import com.intellij.psi.tree.TokenSet;
-import com.intellij.psi.util.*;
 import com.intellij.psi.util.CachedValueProvider.Result;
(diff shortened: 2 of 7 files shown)


python/psi-api/src/com/jetbrains/python/psi/stubs/PyClassStub.java
```diff
@@ -51,4 +52,9 @@
 
   @Nullable
   String getDocString();
+
+  @Nullable // Made nullable for compatibility
+  default List<String> getSuperClassesText() {
```
east825: I doubt about the nullable return value here. Does it really make sense considering that you add an explicit assert on it afterwards in `PyClassElementType#serialize()`?

python/src/com/jetbrains/python/psi/impl/stubs/PyClassStubImpl.java
```diff
@@ -48,17 +48,34 @@
...
+                         @NotNull List<String> subscriptedSuperClassesText,
                          @Nullable QualifiedName metaClass,
                          @Nullable List<String> slots,
                          @Nullable String docString,
                          @NotNull IStubElementType stubElementType) {
+    this(name, parentStub, superClasses, subscriptedSuperClassesText, null, metaClass, slots, docString, stubElementType);
+  }
+
+  public PyClassStubImpl(
+    @Nullable String name,
```
east825: Could you please change the formatting so that the indent of the parameter list doesn't clash with that of the function body (unfortunately, there is no unambigous indent guidelines for Java as in PEP 8).

east825:
Our initial intend was to get rid of the list of base classes in the class presentation altogether (since no other language support seems to do that) rather than store additional information in stubs. But after giving it some thought we decided that in may be beneficial in a longer term in other places in code insight (generics, `six.with_metaclass`, special "keyword" base classes in `typing`, etc.).

python/psi-api/src/com/jetbrains/python/psi/stubs/PyClassStub.java
fitermay: @east825 I saw that @sproshev was adding compatablity code to other stubs. My deduction is the reason for the reason for this is that the stub classes are being created by some plugins for their own uses, but these stubs are not being serialized.

python/src/com/jetbrains/python/psi/impl/stubs/PyClassStubImpl.java
fitermay: @east825 That’s true but there are IDEA formatter settings for the intellij-community project. Did I forget to format when checking in or is that formatting option not supported?

fitermay:
Regarding your last comment. Storing the text will not be very useful for code analysis IMO because class names can be aliased in multiple ways. On the other hand, speaking as a user, I personally find this information very useful when performing ‘find class’.

fitermay:
@east825 After thinking about it  -- you're right. There is no point of keeping the method nullable for compatibility since it's a new method. Requested changes applied.

fitermay:
@east825 ping

python/src/com/jetbrains/python/psi/impl/PyClassImpl.java
```diff
@@ -356,20 +356,27 @@ public ItemPresentation getPresentation() {
...
           result.append(")");
         }
         return result.toString();
       }
     };
   }
 
+  private List<String> getSuperClassesText() {
+    PyClassStub stub = getGreenStub();
+    if (stub == null || stub.getSuperClassesText() == null) {
```
east825: `stub.getSuperClassesText() == null` is now redundant

python/src/com/jetbrains/python/psi/impl/stubs/PyClassStubImpl.java
```diff
@@ -99,6 +116,12 @@ public String getDocString() {
     return myDocString;
   }
 
+  @NotNull
+  @Override
+  public List<String> getSuperClassesText() {
+    return mySuperClassesText;
```
east825: Nullability warning here.

east825:
Sorry, I was on Christmas/New Year holidays that are somewhat lengthy in Russia. I left a few comments about code style and inspection warnins, but apparently forgot to submit them :sweat_smile: (not that familiar with reviewing on GitHub) and already fixed everything myself. Do you mind me squashing your two commits into one?

fitermay:
@east825 Sure, squash away. By the way, the formatting issue is what the IDEA precommit formatter did to the code. Please direct complaints about it to JetBrains 😈

PR #663, opened 2017-11-14
PY-15360 Error on trailing comma after star args for Py<3.5

fitermay:


python/src/com/jetbrains/python/validation/CompatibilityVisitor.java
```diff
@@ -704,6 +705,19 @@ else if (seenKeywordContainer) {
         }
       }
     }
+
+    /* check for trailing comma */
+    PyExpression lastArg = ContainerUtil.getLastItem(Arrays.asList(callExpression.getArguments()));
```
sproshev: `com.intellij.util.ArrayUtil#getLastElement`

PR #664, opened 2017-11-14
PY-27004: Use multiResolve to check if Python refs match

fitermay:


vlasovskikh:

PR #665, opened 2017-11-15
support atomic save

fitermay:
@trespasserw @nicity 

Please accept this PR. It prevents many heisenbugs results from the file being temporarily deleted when saving. For example Git rescan sometimes picking up the deleted file. It is also safer and faster than the alternate save. 

Had to do some copy-paste reuse because the module with SafeOutputStream must support JDK 1.6 and this needs 1.7+.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/AtomicSafeFileOutputStream.java
+++ b/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/AtomicSafeFileOutputStream.java
@@ -1,39 +1,28 @@
-/*
- * Copyright 2000-2016 JetBrains s.r.o.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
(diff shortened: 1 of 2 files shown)


trespasserw:
Just for the record: first proposed in https://youtrack.jetbrains.com/issue/IDEA-155415.

Need some time. I understand the patch, but there are other complaints on class' behavior, and fixing them may eliminate the need for the move.

fitermay:
@trespasserw Thanks for the prompt response. Let's discuss this. I'm having trouble imagining any proper implementation of save that doesn't end up calling Files.move or otherwise an OS-specific API. What other complaints are there about the class' behavior?

trespasserw:
The main complaint is that the class deletes original files and replaces them with new ones. This may cause permissions and ownership loss.

fitermay:
@trespasserw What is your plan for implementing save without replacing the file?
On Windows you can use the ReplaceFile OS call that atomically replaces the file while keeping th original file's indentity, including ACLs.

On the other hand on UNIX based systems there is no such call so you would have to write directly to the file. This would make the write non atomic, even with a file lock because file locks on UNIX systems are advisory,not mandatory,

trespasserw:
In 2019.3, we have finally rewritten the class logic and the problem is no longer actual (there is no "move" anymore).

Sorry for this taking so long.

PR #666, opened 2017-11-15
Honor .mailmap

albfan:
Using `%aN` instead of `%an` git log respects the mappings in .mailmap file

More details on https://git-scm.com/docs/git-log

### DIFF
--- a/plugins/git4idea/src/git4idea/history/GitLogParser.java
+++ b/plugins/git4idea/src/git4idea/history/GitLogParser.java
@@ -228,7 +228,7 @@ public class GitLogParser {
    * These are the pieces of information about a commit which we want to get from 'git log'.
    */
   enum GitLogOption {
-    HASH("H"), TREE("T"), COMMIT_TIME("ct"), AUTHOR_NAME("an"), AUTHOR_TIME("at"), AUTHOR_EMAIL("ae"), COMMITTER_NAME("cn"),
+    HASH("H"), TREE("T"), COMMIT_TIME("ct"), AUTHOR_NAME("aN"), AUTHOR_TIME("at"), AUTHOR_EMAIL("ae"), COMMITTER_NAME("cn"),
     COMMITTER_EMAIL("ce"), SUBJECT("s"), BODY("b"), PARENTS("P"), REF_NAMES("d"), SHORT_REF_LOG_SELECTOR("gd"),
     RAW_BODY("B");
 


albfan:
Related to https://youtrack.jetbrains.com/issue/IDEA-160677 and https://stackoverflow.com/a/39301372/848072

juliabeliaeva:
There is a problem with this change: it can break filtering in log. User filters use indexes to filter, and indexes done prior to this change would contain old user names. We need either to rebuild indexes on every .mailmap change (which we do not want to do), or to read .mailmap and use this information to query indexes by both new and old names.

albfan:
Yes I guess that adding `%an` and `%aN`, both can be parsed. Then you can use `%an` for filter if you want. But when you use .mailmap to remap same person, like:

```
user <newemail> <oldemail>
user <newemail> <oldemail2>
```

what you want is to filter by mapped name (I guess)

`.mailmap` is not a file with much traffic. Once mapped it stay the same most of the time

I didn't recompile all Intellij to try this change, if you need me to test anything, just tell me

albfan:
@juliabeliaeva lastly I managed to build Intellij (for other PR) I can try or implement anything if you need it.

All my repos comes with an ugly corporate username and .mailmap is a real improvement to review changes and triage bugs

arranbartish:
@juliabeliaeva : I second @albfan. Is there something I can do to help this PR get merged?

albfan:
I'm avaliable to any further implementation.

Manually invalidate cache in case you add `.mailmap` to an existing project already imported in Intellij seems a reasonable workaround

juliabeliaeva:
In order to correctly implement this feature, one needs to read mailmap and use it to convert names when displaying / convert back when filtering.

albfan:
I think it can use a flag, so if cache was not made with mailmap, do one step conversion. Maybe that flag should be a sha of .mailmap, so that conversion should be done everytime .mailmap changes.

Sounds it correct @juliabeliaeva?

juliabeliaeva:
@albfan yes, we can detect when index was made without a .mailmap or with a different .mailmap, but to do something with it we still need to read .mailmap and convert names back and forth (and reversing names from index made with different mailmap could be awkward). And since we are reading .mailmap anyway, we could just do it all the time, keeping original names in the index as it is done now.

juliabeliaeva:
To summarize: presented implementation would not work well with the log indexes. Names conversion should be done when displaying commit details or filtering, this will allow to keep original names in the index. So, closing this request for now until an improved implementation is provided.

PR #669, opened 2017-11-17
PY-20075: Fix false positive unused imports

fitermay:
When importing multiple modules from the same package


fitermay:
@sproshev ping!

fitermay:
@sproshev @vlasovskikh I've looked over the code again and I think we can get rid of
PyImportedModule  and PyImportedModuleType.

 What do you think?

fitermay:
@sproshev @vlasovskikh 

Also found issues https://youtrack.jetbrains.com/issue/PY-28393 and https://youtrack.jetbrains.com/issue/PY-28392 while digging through this code. I think the confusion with PyImportedModuleType/PyModuleType is largely to blame for PY-28393

python/src/com/jetbrains/python/psi/types/PyImportedModuleType.java
```diff
@@ -43,20 +37,33 @@ public PyImportedModuleType(@NotNull PyImportedModule importedModule) {
     final PsiElement resolved = myImportedModule.resolve();
     if (resolved != null) {
       final PsiFile containingFile = location != null ? location.getContainingFile() : null;
-      List<PsiElement> elements = Collections.singletonList(ResolveImportUtil.resolveChild(resolved, name, containingFile, false, true,
-                                                                                           false));
+      PsiElement resolvedChild = ResolveImportUtil.resolveChild(resolved, name, containingFile, false, true,
+                                                                false);
```
sproshev: I'd use `resolveChildren` instead.

python/src/com/jetbrains/python/psi/types/PyImportedModuleType.java
```diff
@@ -43,20 +37,33 @@ public PyImportedModuleType(@NotNull PyImportedModule importedModule) {
...
+        if (isPackage) {
+          return res;
         }
       }
-      return ResolveImportUtil.rateResults(elements);
+      if (resolvedChild != null) {
+        ResolveResultList l = new ResolveResultList();
+        l.add(new ImportedResolveResult(resolvedChild, RatedResolveResult.RATE_NORMAL, importElement));
+        return l;
+      }
```
sproshev: I think it's OK to return `Collections.singletonList(...)` here

python/src/com/jetbrains/python/psi/types/PyImportedModuleType.java
```diff
@@ -43,20 +37,33 @@ public PyImportedModuleType(@NotNull PyImportedModule importedModule) {
...
+          PyModuleType.processImportedSubmodules((PsiFileSystemItem)resolved, location, (nameDefiner, importedSubmodule) -> {
+            if (importedSubmodule.equals(resolvedChild)) {
+              res.add(new ImportedResolveResult(resolvedChild, RatedResolveResult.RATE_NORMAL, nameDefiner));
+              return false;
+            }
+            return true;
+          });
+
+        if (isPackage) {
+          return res;
```
sproshev: I'm not sure what is the key change, `isPackage` flag? What is the smallest part of code to comment or revert to make the test failed?

python/src/com/jetbrains/python/psi/types/PyModuleType.java
```diff
@@ -368,21 +368,35 @@ private static CompletionVariantsProcessor createCompletionVariantsProcessor(Psi
...
+      else if (importedSubmodule instanceof PsiNamedElement) {
+        lookupElements.add(LookupElementBuilder.createWithIcon((PsiNamedElement)importedSubmodule));
+      }
+      return true;
+    });
+    return lookupElements;
   }
 
-  @Nullable
+  @Deprecated
```
sproshev: See how we deprecate methods in PyCallExpression, there are notes about version to remove and alternative way to achieve the result.

python/src/com/jetbrains/python/psi/types/PyImportedModuleType.java
fitermay: This is just refactoring. Changing returning a list or null to a processor. This is because it was unclear what was going on PyImportedModuleType.resolveMember. It was looking for a single element by filtering a list against a set containing a single item.  Using a processor makes it clear we're looking for a single element and is also more efficient since we  can stop once we find it. 

The only key change is wrapping the return values of PyImportedModuleType.resolveMember as `ImportedResolveResult`

fitermay: The isPackage flag was because the return value of the function was treated differently if it returned emptyList and null.  I think I can make it cleaner by making it return `void` and adding the package check seperately

python/src/com/jetbrains/python/psi/types/PyModuleType.java
fitermay: Which version should I put down?

python/src/com/jetbrains/python/psi/types/PyImportedModuleType.java
fitermay: Yes, that might fix https://youtrack.jetbrains.com/issue/PY-28393

python/src/com/jetbrains/python/psi/types/PyModuleType.java
sproshev: 2018.3

python/src/com/jetbrains/python/psi/types/PyImportedModuleType.java
sproshev: Thank you, it's clear to me now

vlasovskikh:
@fitermay I like the approach you've taken in #728. May I close this one in favor of #728?

vlasovskikh:
@fitermay Shall we close it?

PR #670, opened 2017-11-21
Enumerate module roots directly in ModuleHighlightUtil.getModuleDescriptor

udalov:
This is faster than going through FilenameIndex which would call
GlobalSearchScope.accept for all "module-info.java" files in the
project, and there can be many of those


trespasserw:
That's not quite correct. These files may live anywhere in module sources, not just under a source root.

Would it help you if I add caching to the method?

PR #671, opened 2017-11-24
FernFlower: Rename classes whose filenames will exceed 255 characters

thenickdude:
An obfuscated application I'm reversing has a class called "OoOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO" in it (256 characters long). This decompiles fine into a JAR, but if I want to unpack the JAR in order to refactor code and/or recompile it, I quickly run in to filesystem filename length limits ([almost all filesystems](https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits) top out at 255 characters).

This patch renames classes that would exceed 255 characters in length once ".class" is added to the name.


trespasserw:
Thanks!

PR #672, opened 2017-11-29
Annotate AnActionEvent parameter in AnAction.update as @NotNull

Flamenco:
I can't find any implementation that actually checks to see if this parameter is NULL. If it was ever NULL I think bad things would happen across the board.

Also, my Kotlin stubs always mark it as AnActionEvent? - changing it to AnActionEvent every time is a bit tedious.

### DIFF
--- a/platform/editor-ui-api/src/com/intellij/openapi/actionSystem/AnAction.java
+++ b/platform/editor-ui-api/src/com/intellij/openapi/actionSystem/AnAction.java
@@ -225,7 +225,7 @@ public abstract class AnAction implements PossiblyDumbAware {
    *
    * @param e Carries information on the invocation place and data available
    */
-  public void update(AnActionEvent e) {
+  public void update(@NotNull AnActionEvent e) {
   }
 
   /**


platform/editor-ui-api/src/com/intellij/openapi/actionSystem/AnAction.java
```diff
@@ -225,7 +225,7 @@ public boolean displayTextInToolbar() {
    *
    * @param e Carries information on the invocation place and data available
    */
-  public void update(AnActionEvent e) {
+  public void update(@NotNull AnActionEvent e) {
```
ignatov: Will have lots of yellow code in all AnAction's inheritors afterward.

Flamenco: I would prefer rock-solid non-null contract over yellow code every time.  Also, when you remove your 'yellow code', you will get more 'yellow code' suggestions to remove all of your redundant null checks;  I'll gladly take those too.

ignatov: @bulenkov, @chashnikov, @cdracm what do you think?

cdracm: i'm all for it

chashnikov: Of course, we must change all the inheritors to avoid yellow code in our sources. Since this will modify many existing files it's better to postpone this change at least for two weeks until massive cherry-picks to 182 branch is over to avoid merge conflicts.

But before we do that, we firstly need to fix all cases in IntelliJ Community and Ultimate where `null` is passed to this method (like [this one](https://github.com/JetBrains/intellij-community/blob/5dccf937e429c03be07dfb6ebf8355d1c9e6f8a7/platform/lang-impl/src/com/intellij/util/ui/tree/PerFileConfigurableBase.java#L572)). Otherwise we can get exceptions because we have 'Add runtime assertions for not-null-annotated elements' option enabled in IDEA project.

Also it makes sense to annotate the parameter of `actionPerformed` with `@NotNull` as well (for consistency and because there are cases when `update` is called from `actionPerformed`).

So I suggest to firstly find and fix all places where `null` is passed to implementations of this method, and then change the signature and the implementations. Since this requires changes not only in intellij-community, but also in IntelliJ Ultimate sources, I doubt that this can be properly done by an external developer.

ignatov: @cdracm could you please find all such potential places and update them before the massive changes?

Flamenco:
I would like to suggest finding other places in the openapi package that should also have this annotation and refactoring them as well (not just the callers and implementers of this function).
Since moving over to Kotlin, I can't stand letting NULLs in when they don't need to be...

DmPanov:
Closed by mistake, sorry

Flamenco:
I was so excited.  😭

chashnikov:
Actually we added `@NotNull` annotation to `update` method 2 years ago, see b32cced7a477.

PR #674, opened 2017-12-04
Fix an ildefined comparator in AbstractTreeUi.

yuriy-solodkyy:
Hi,

this patch is to fix an ildefined comparator in AbstractTreeUi which causes a lot of crashes related to https://youtrack.jetbrains.com/issue/IDEA-181904. (#IDEA-181904) However, please note that the patch does not fix the problem entirely. I suspect it is not the root cause but just something triggered by too many null descriptors or similar. 
Attn:  @vladsoroka


klikh:

yuriy-solodkyy:
Hi Kirill, I'm a Google employee so I believe it should be covered by the
agreement between companies.

klikh:
Ah, it was not clear from your profile. You don't have to sign a separate CLA indeed.

vladsoroka:
Thanks for the contribution, Yuriy! I've merged the PR using the following commit https://github.com/JetBrains/intellij-community/commit/adc31c97a04402f5dcb4af70764c620e856b2fee

PR #675, opened 2017-12-05
Fix sporadic exceptions in AbstractTreeUi.

yuriy-solodkyy:
AbstractTreeUi has been sporadically failing in different places and
most often at

    assert descriptor != null;

in sortChildren().

The other manifistation of the problem was "null" nodes sometimes
appearing in the tree (nodes that visually appear with text "null").

The root cause of the problem is getNodeForElement() which disposes
any node that is not in the tree. However, because of async nature of
queued updates a node which is being dosposed can be already scheduled
for addition to the tree.

Here is the original comment from the commit from 2008 thad added
diposing:

IDEADEV-29853 - invalid nodes, found by custom builders are disposed

This CL replaces the getNodeForElement() with the equiavalent code that
does not dispose any nodes.
Attn: @vladsoroka 

Test: "null" node does not appear visually. makeSimpleApplication() UI
      test passes.
Bug: IDEA-181904
Change-Id: I837f2f4d3ab7a235d8b9ec6cb1afb215b4c230e4

### DIFF
--- a/platform/platform-api/src/com/intellij/ide/util/treeView/AbstractTreeUi.java
+++ b/platform/platform-api/src/com/intellij/ide/util/treeView/AbstractTreeUi.java
@@ -1008,7 +1008,7 @@ public class AbstractTreeUi {
       DefaultMutableTreeNode nodeToUpdate = null;
       boolean updateElementStructure = updateStructure;
       for (Object element = fromElement; element != null; element = getTreeStructure().getParentElement(element)) {
-        final DefaultMutableTreeNode node = getNodeForElement(element, false);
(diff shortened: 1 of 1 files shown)


yuriy-solodkyy:
The PR basically removes the fix introduced at https://github.com/JetBrains/intellij-community/commit/768178ea9e23d3efdf5c063d5200bb436370129c
for the queued updates case.

vladsoroka:
Looks good, thank you! I've megred it to master branch (https://github.com/JetBrains/intellij-community/commit/267d7b48414b28a0b144b0effe43e815b5473180) and when related tests will remain green the change will be cherry-picked  to 173.

PR #677, opened 2017-12-06
Remove unnecessary throws declarations

dahlstrom-g:


platform/testFramework/src/com/intellij/testFramework/LeakHunter.java
```diff
@@ -37,12 +37,12 @@
 
 public class LeakHunter {
   @TestOnly
-  public static void checkProjectLeak() throws Exception {
+  public static void checkProjectLeak() {
```
dahlstrom-g: This method appears to be unused. Can we just remove it?

klikh:
I've pushed the fix. Thanks.

This method is not unused: it is used in `LeakCheck` and in `ProjectLeakTest`.

PR #678, opened 2017-12-07
IDEA-183411 Release lock while updating UI in AbstractTreeBuilder

yuriy-solodkyy:
It seems that the  purpose of Abstract TreeBuilder async updates is to allow
background submition of updates while the UI is being updated in EDT.

All the fields are guarded by the lock (sync(this)) apart from the myTreeBuilder
which refers to the UI and the UI should only be update from the EDT. However,
performUpdate() keeps the lock while updating the UI which effectively prevents other
threads to submit new updates. Under the high load of updates it results in other
threads being blocks and thread pools spawning many new threwads.

This change releases makes the performUpdate() hold the lock only while accessing the
fields guarded by the lock.  Also, it removes unnecessary null checks on final
always-not-null myTreeBuilder to avoid confusion.

Bug: IDEA-183411
Change-Id: I4516f38375e8ba194383ad6060acd4f05225751b

### DIFF
--- a/platform/platform-api/src/com/intellij/ide/util/treeView/AbstractTreeUpdater.java
+++ b/platform/platform-api/src/com/intellij/ide/util/treeView/AbstractTreeUpdater.java
@@ -241,18 +241,22 @@ public class AbstractTreeUpdater implements Disposable, Activatable {
     myTreeBuilder.updateSubtree(node);
   }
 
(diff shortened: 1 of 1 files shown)


yuriy-solodkyy:
Attn: @vladsoroka

PR #680, opened 2017-12-10
Git integration check-ignore command

Dmitry-Kucher:
Added `check-ignore` command for git integration to allow check if separate pathname/filename under `.gitignore`. To see specification for command click on link [check-ignore](https://git-scm.com/docs/git-check-ignore)


Dmitry-Kucher:
@klikh `git ls-files --ignored` shows list of all **ignored** paths and you need to check all list to find out if your _pathname_ under `.gitignore`. `git check-ignore pathname` checks only selected pathname without additional actions

klikh:
Nope, you can still give a path to `ls-files` as well. For example:
```
# git ls-files --ignored --other --exclude-standard -- .idea/shelf/Default.xml
.idea/shelf/Default.xml
```

Dmitry-Kucher:
@klikh Thank you so much! `ls-files` is much more powerful than I imagined before.

I'll close MR.

PR #683, opened 2017-12-13
Attach LeakHunter output to the exception message

dahlstrom-g:
… for better diagnostic output in test results.


dahlstrom-g:
@cdracm?

cdracm:
pushed, thanks

PR #685, opened 2017-12-15
Fix bug in resizing of DialogWrapper with large error label

jgaillard85:
If the error label in the DialogWrapper is displayed on multiple lines,
any update of the error message that keeps it on several lines will
cause the dialog window to shrink.

### DIFF
--- a/platform/platform-api/src/com/intellij/openapi/ui/DialogWrapper.java
+++ b/platform/platform-api/src/com/intellij/openapi/ui/DialogWrapper.java
@@ -236,7 +236,7 @@ public abstract class DialogWrapper {
           if (!myResizeInProgress) {
             myActualSize = myPeer.getSize();
             if (myErrorText != null && myErrorText.isVisible()) {
-              myActualSize.height -= myErrorText.getMinimumHeight();
+              myActualSize.height -= myErrorText.getMinimumSize().height;
             }
           }
         }
@@ -1253,7 +1253,7 @@ public abstract class DialogWrapper {
         if (height != myHeight) {
           myHeight = height;
           myResizeInProgress = true;
-          myErrorText.setMinimumHeight(height);
+          myErrorText.setMinimumSize(new Dimension(0, height));
           JRootPane root = myPeer.getRootPane();
           if (root != null) {
             root.validate();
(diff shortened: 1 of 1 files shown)


jgaillard85:
CLA signed

dahlstrom-g:
No individual CLA should apply to this contribution, since it's work for hire at Google and covered by the agreement between Google and JetBrains.

@jgaillard85, you should join the [Google org](https://github.com/google).

platform/platform-api/src/com/intellij/openapi/ui/DialogWrapper.java
```diff
@@ -1253,7 +1253,7 @@ public void componentResized(ComponentEvent event) {
         if (height != myHeight) {
           myHeight = height;
           myResizeInProgress = true;
-          myErrorText.setMinimumSize(new Dimension(0, height));
+          myErrorText.setMinimumHeight(height);
```
kkmic: This change is redundant. It technically changes nothing.

platform/platform-api/src/com/intellij/openapi/ui/DialogWrapper.java
```diff
@@ -236,7 +236,7 @@ public void componentResized(ComponentEvent e) {
           if (!myResizeInProgress) {
             myActualSize = myPeer.getSize();
             if (myErrorText != null && myErrorText.isVisible()) {
-              myActualSize.height -= myErrorText.myLabel.getHeight();
+              myActualSize.height -= myErrorText.getMinimumHeight();
```
kkmic: I'd put it myErrorText.getMinimumSize().height without adding new methods.

jgaillard85: Updated accordingly

platform/platform-api/src/com/intellij/openapi/ui/DialogWrapper.java
jgaillard85: Removeed

dahlstrom-g:
@kkmlc I'm not seeing this change in JetBrains:master where the update above shows it merged. Do you know why?

kkmic:
'Cause I did it a wrong way. Let me do it the right way. And it'll be there. Sorry, doing it the first time.

dahlstrom-g:
I see it commit d796888 on master now. Thanks!

PR #686, opened 2017-12-18
[IDEA-183953] Add test for unicode command

FTOH:
https://youtrack.jetbrains.com/issue/IDEA-183953


ignatov:
Already fixed, thanks for the RP. See https://youtrack.jetbrains.com/issue/IDEA-183953

PR #689, opened 2017-12-19
Scale should matter only if the type is NOT SVG

kylekatarnls:
The following project use this loader with SVG format: https://github.com/davidsommer/IconViewer/blob/master/src/ch/dasoft/iconviewer/ImageIconProvider.java

All icons are well displayed except SVG.

My inspection lead me to this ImageLoader class and I try to see what's wrong with SVG.

### DIFF
--- a/platform/util/src/com/intellij/util/ImageLoader.java
+++ b/platform/util/src/com/intellij/util/ImageLoader.java
@@ -107,7 +107,7 @@ public class ImageLoader implements Serializable {
       }
       if (stream == null) {
         if (useCache) {
-          cacheKey = path + (type == Type.SVG ? "_@" + scale + "x" : "");
+          cacheKey = path + (type != Type.SVG ? "_@" + scale + "x" : "");
           Image image = ourCache.get(cacheKey);
           if (image != null) return image;
         }


forantar:
In IDEA SVG icons are used in two ways:
(1) as a predefined icon
(2) as an image in the SVG viewer

In (1) case, an SVG file is loaded and is rendered into a buffered image in some scale. The scale can vary depending on the IDE settings and multi-monitor environment. It's assumes the SVG file is not changed in runtime, so the out buffered image is cached, per scale. The code:

cacheKey = path + (type == Type.SVG ? "_@" + scale + "x" : "");

produces keys as follows:

/path/to/my/icon_@2x
/path/to/my/icon_@2.5x

So that when the same SVG icon is requested in scale 2, 2.5 (etc.) the proper cached buffered image is used. (Actually, the same logic is applicable to PNG icons. We could have cached them per scale)

The (2) case differs in that the SVG file, being viewed, is eligible for editing. For that case no cache is used, and the SVG is loaded from disk on every request.

As to your code. I can see the only call used for icon loading: ImageLoader.loadFromStream. But it actually loads no svg icons, unlike ImageLoader.loadFromUrl. So, I don't clearly understand your problem. Am I missing something?

kylekatarnls:
I try to get a SVG preview using ImageLoader.loadFromStream, if it's not possible, that's probably my first problem. The davidsommer/IconViewer plugin allow to preview images from the explorer icons. I tried to add SVG support (wished to see SVGs icons as their content), but using the same principle as for PNG/JPG,etc. it just display nothing.

This is what I just did: https://github.com/davidsommer/IconViewer/commit/dae5bccc54aa05b590d381a83ecfd5a3595a19cf

And where the SVG is beeing loaded:
https://github.com/davidsommer/IconViewer/blob/master/src/ch/dasoft/iconviewer/ImageIconProvider.java#L30

Would you know why we cannot get a SVG file this way?

forantar:
As I wrote above, ImageLoader.loadFromStream(..) doesn't allow to load SVG images, just because JDK itself doesn't support SVG (IDEA uses Batik to deal with SVG). I'd suggest that you change it to this call: ImageLoader.loadFromUrl(containingFile.getVirtualFile().getUrl()) which is SVG-aware.

stachenov:
This PR is almost 10 years old now, and the code around that subsystem was heavily modified since then. If there's still an issue with SVG scaling, please submit a YouTrack issue, I'll have a look.

PR #693, opened 2017-12-21
IDEABKL-6449 Enable ISO-8601 date format via system property

simon04:
Setting the system property `date.iso8601` to `true` (`java -Ddate.iso8601=true`), enables the ISO-8601 format in the `DateFormatUtil`.

This addresses https://youtrack.jetbrains.com/issue/IDEABKL-6449.

I'm not sure whether and how to test this contribution: programmatically setting the system property would have to be done prior to the first access of `DateFormatUtil` (in all variations of running the tests).

### DIFF
--- a/platform/util/src/com/intellij/util/text/DateFormatUtil.java
+++ b/platform/util/src/com/intellij/util/text/DateFormatUtil.java
@@ -311,7 +311,15 @@ public class DateFormatUtil {
   private static SyncDateFormat[] getDateTimeFormats() {
     DateFormat[] formats = null;
     try {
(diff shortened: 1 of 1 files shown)


ignatov:
@simon04, please update the PR due to conflicts, cc @klikh.

simon04:
Done. Please take less than 7 months for reviewing / commenting this PR! (This is not motivating at all for spare-time contributions to open-source projects!)

klikh:
Finally, the issue has been fixed in 2019.3. It was decided to implement it not as a JVM property, but provide a UI in the Settings.

PR #696, opened 2017-12-23
Add git4idea dialog handlers to extension points.

syuchan1005:


klikh:
So why do you need these extension points?

syuchan1005:
I'm developing [EmojiPrefix](https://github.com/syuchan1005/EmojiPrefix) Plugin. but When I attach emoji to commit message, I couldn't put JPanel in dialogs.
There is a way to hook Toolbar action on my plugin, but I don't think that is a good way.

klikh:
Have you tried to use the `CommitMessage` class or `CommitMessageI` interface available from the `VcsDataKeys.COMMIT_MESSAGE_CONTROL` DataKey.

syuchan1005:
VcsDataKeys.COMMIT_MESSAGE_CONTROL need DataContext or DataProvider. but It doesn't gave from Git4Idea.

klikh:
Sorry for a delay. So what's the problem of getting the `COMMIT_MESSAGE_CONTROL` from the `DataContext`? Could you please point me to the exact place in the code where you'd like to modify the commit message?

syuchan1005:
I want to add JPanel to those dialogs, and i want to change commit message when those dialogs are close.

pchel-:
It doesn't seem reasonable to customize all these dialogs separately, but it makes sense to allow plugins to provide custom commit message components. So I've created corresponding tasks - [IDEA-193251](https://youtrack.jetbrains.com/issue/IDEA-193251), [IDEA-193252](https://youtrack.jetbrains.com/issue/IDEA-193252).

As for the solution you could use right now. You could try overriding corresponding git actions, like
```
      <action id="Git.Tag" class="<custom git tag action>" text="Tag..." overrides="true" />
```
In your actions you could get access to the corresponding dialogs and update commit message components as you need.

PR #697, opened 2017-12-23
[PY-27710] Fix for erroneous "Too few arguments for format string"

andreisilviudragnea:


andreisilviudragnea:
I will take a look at it over the weekend. Thanks for pointing it out!

andreisilviudragnea:
I have updated my fix. It was more complicated than I expected, because I had to make sure I did not break anything else while trying to fix this.

ignatov:
Has been already fixed, thanks for PR. See https://youtrack.jetbrains.com/issue/PY-27710

PR #700, opened 2017-12-30
Get rid of unnecessary array wrapping

stsypanov:
See [https://youtrack.jetbrains.com/issue/IDEA-184240](https://youtrack.jetbrains.com/issue/IDEA-184240) for details


amaembo:
Applied, thanks.

PR #707, opened 2018-01-11
Fixes Task issue status being saved in a corrupt fashion for JIRA

flaw600:
Especially when connected to a JIRA server, there are times where the status will not update correctly upon closing the issue. After analysis between Opem amd Close Task Dialog (because the status would successfully update on opening the task, I noticed that the Task is being saved at different levels of inheritance, potentially corrupting the state that the task is in and so checked for that and cast it so


flaw600:
Re-opening after signing CLA

ignatov:
cc @dmitry-avdeev

dmitry-avdeev:
I don't see how casting can chage state saving behaviour.

PR #708, opened 2018-01-11
Fix incorrect closing streams in sanselan IOUtils

dplusic:


klikh:
This class is deprecated and is not intended to be used. It was not removed only to keep compatibility with plugins which have not yet been updated.

dplusic:
To keep compatibility, it must work properly.
The method, however, has bug that cannot copy streams of which size less than the size of a buffer.
So, not updated plugins won't work properly, and I'm using one of them.

klikh:
Nope, "keeping compatibility" means "not breaking the stuff that used to work".

Maybe you could send a pull request to the plugin, to help the author in migrating to new API?

dplusic:
The bug will "break the stuff that used to work".
This class is just "added" at 2017.3 with removing `org.apache.sanselan` package.
([change notes](https://github.com/JetBrains/intellij-sdk-docs/blob/master/reference_guide/api_changes_list.md#changes-in-intellij-platform-20173), [commit log](https://github.com/JetBrains/intellij-community/commit/e791557ca1489b02d178aa68960d645ab501e674))
I cannot find the original source of your customized `org.apache.sanselan` package, but it doesn't have the same bug. (Please refer this [source](https://github.com/apache/sanselan/blob/trunk/src/main/java/org/apache/commons/imaging/util/IoUtils.java#L172))

klikh:
@gregsh Please have a look.

gregsh:
@dplusic Thanks for the fix! Will see it backported.

PR #709, opened 2018-01-14
Improve malformed identifiers handling in `-ren=1` mode

Maccimo:
This patch propose following improvements to the default renaming functionality:

1. All identifiers should be compliant to the JLS9 section 3.8 and SHOULD NOT contains so-called "ignorable" characters.
Ignorable characters are silently removed by javac during the compilation and thus may appear only in specially crafted obfuscated classes.
For more information about "ignorable" characters see <a href="https://bugs.openjdk.java.net/browse/JDK-7144981">JDK-7144981</a>.

2. Add dedicated unit tests for `org.jetbrains.java.decompiler.modules.renamer.ConverterHelper` class.


trespasserw:
@Maccimo thank you.

Could you please update the PR so that it doesn't change formatting in `toBeRenamed()` method? And I'm not sure the new method deserves such a thorough test :)

Maccimo:
@trespasserw I've updated PR, please review.
Tests attempt to cover original `toBeRenamed()` behaviour as well as changes made by this PR.
IMHO better to have test coverage than not to have.

PR #712, opened 2018-01-20
prevent django_manage.py from tracking changes in node_modules

stebunovd:
The patch prevents `manage.py` console in PyCharm from being painfully slow when there is a `node_modules` somewhere in the project tree.

Timings from my current project, running `migrate` command from `manage.py` console in PyCharm:
- with the patch: 7 sec
- without the patch: 5 min 19 sec

Developers are not supposed to manually change the contents of `node_modules`, so I guess it's fine to exclude it from tracking. I also think it is ok to just hardcode `node_modules` in the method code, because it is hardcoded in Node and [developers cannot rename it](https://stackoverflow.com/a/21818724).

### DIFF
--- a/python/helpers/pycharm/_jb_utils.py
+++ b/python/helpers/pycharm/_jb_utils.py
@@ -28,7 +28,7 @@ class FileChangesTracker(object):
     def _get_changes_from(folder, patterns):
         result = {}
         for tmp_folder, sub_dirs, files in os.walk(folder):
-            sub_dirs[:] = [s for s in sub_dirs if not s.startswith(".")]
+            sub_dirs[:] = [s for s in sub_dirs if not s.startswith(".") and s != 'node_modules']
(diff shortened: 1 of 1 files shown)


throwable-one:
Hello.
Since 2018.1 PyCharm does not track all folders, but filters them by pattern instead: https://youtrack.jetbrains.com/issue/PY-27260
Please check if this issue is still actual in 2018.1

stebunovd:
Hi there,

PyCharm 2018.1 is indeed faster, it demonstrates 3-4x speed improvement without the patch. However, it is still slow. Here are measurements from my project with PyCharm 2018.1, running `migrate` without any active migrations to apply:

* with the patch: 5 sec
* without the patch: 1 min 13 sec - 1 min 35 sec (best and worst results)

Of course, 1+ minute is much better than 5+ minutes, but the delay is still very noticeable and annoying.

PR #714, opened 2018-01-25
Allow to bypass OOCB for lazy elements in non-Java languages

goodwinnk:
As JavaCodeBlockModificationListener works for all PsiClassOwner files,
it may unexpectedly cause OOCB for lazy elements in languages, that
don't expect it. Allow to bypass this by implementing
PsiModifiableCodeBlock interface.

 #IDEA-185462 Fixed


donnerpeter:
Merged, thanks!

PR #716, opened 2018-01-27
More detailed tests for UnnecessaryUnicodeEscapeInspection

rillig:
I had been confused when I saw the `do while` loop in the code since it
looked like an obvious endless loop. Therefore I wrote some tests to
make sure this code is covered (which it already was before).

Since extracting the escape sequences from the source code is
independent of reporting the errors, I extracted that code to its own
method, which made it easily testable on its own.

### DIFF
--- a/plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/internationalization/UnnecessaryUnicodeEscapeInspection.java
+++ b/plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/internationalization/UnnecessaryUnicodeEscapeInspection.java
@@ -98,6 +98,68 @@ public class UnnecessaryUnicodeEscapeInspection extends BaseInspection {
     return new UnnecessaryUnicodeEscapeVisitor();
   }
 
+  // @VisibleForTesting
+  static void forEachUnicodeSequence(String text, Charset charset, Callback callback) {
+    final CharsetEncoder encoder = charset.newEncoder().onUnmappableCharacter(CodingErrorAction.REPORT);
+    final CharBuffer charBuffer = CharBuffer.allocate(1);
+    final ByteBuffer byteBuffer = ByteBuffer.allocate(10);
+    final int length = text.length();
+    for (int i = 0; i < length; i++) {
+      final char c = text.charAt(i);
+      if (c != '\\') {
+        continue;
+      }
+      boolean isEscape = true;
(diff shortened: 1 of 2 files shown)


plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/internationalization/UnnecessaryUnicodeEscapeInspection.java
```diff
@@ -98,6 +98,68 @@ public BaseInspectionVisitor buildVisitor() {
     return new UnnecessaryUnicodeEscapeVisitor();
   }
 
+  // @VisibleForTesting
```
GaspardPO: Did you forget some commented annotation ?

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/internationalization/UnnecessaryUnicodeEscapeInspection.java
```diff
@@ -113,69 +175,24 @@ public void visitFile(PsiFile file) {
...
+          final RangeMarker rangeMarker = document.createRangeMarker(start, end);
+          registerErrorAtOffset(file, start, end - start, Character.valueOf(codeUnit), rangeMarker);
+          return true;
         }
-      }
+      });
     }
   }
+
+  // @VisibleForTesting
```
GaspardPO: Did you forget some commented annotation?

rillig: No, I didn't. It looks like an annotation, but since IDEA doesn't have this annotation imported by default, I cannot remove the comment. Still, I find it nice to have as kind of documentation.

ignatov:
cc @BasLeijdekkers

BasLeijdekkers:
Hi @rillig,
Sorry I have to decline this pull request. In my opinion this change makes the code harder to read than it already was. The increased tests I like, but I would prefer it if the existing test was just expanded instead. I hope you don't mind (I like your other pull requests:-)

rillig:
Follow-ups:
* https://github.com/JetBrains/intellij-community/pull/923
* https://github.com/JetBrains/intellij-community/pull/924

PR #717, opened 2018-01-27
Improve description of "Type can be weakened" inspection

rillig:
Bold font should be used to draw attention. Type names and method names
are not the important eye catchers here, it's the names of the
configuration knobs.


ignatov:
@BasLeijdekkers, please take a look.

BasLeijdekkers:
Applied manually to Master with some minor changes, thanks for your contribution! I like how the unclear "use the second checkbox to..." was improved.

PR #722, opened 2018-01-31
python: Remove import that's not needed and absent

rohanjoyce:
The import causes builds to fail (it's missing), but it seems to be unnecessary.


sproshev:
Fixed in 51f834d377ec60c2fa6418e93928f60f1c7a5cd1

PR #730, opened 2018-02-11
Improve operations for document listeners

ninja-inc:
**Purpose**

1. I would like to add a listener to `DocumentImpl.myDocumentListeners` safety to avoid an error message if it exists.
2. I would like to remove a listener from `DocumentImpl.myDocumentListeners` safety to avoid an error message if it does not exist.

Currently access to `myDocumentListeners` is prohibited so I implemented 2 public methods in `DocumentImpl`.

Thanks.

### DIFF
--- a/platform/core-api/src/com/intellij/openapi/editor/Document.java
+++ b/platform/core-api/src/com/intellij/openapi/editor/Document.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2017 JetBrains s.r.o.
+ * Copyright 2000-2018 JetBrains s.r.o.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


donnerpeter:
Why do you need this? Note that you can track yourself whether you've added your listener or not. Double addition/removal is probably a bug somewhere in the client, and the new API would obscure that bug.

ninja-inc:
@donnerpeter 
Thank you for comments!

>Why do you need this? Note that you can track yourself whether you've added your listener or not. 

My plugin will add `documentListener` and remove it every time several specific events happen. Of course I can manage these things by myself but it makes my code complicated. To simplify it, I wanted to access `myDocumentListeners` to verify whether it is already added or removed but I made new public methods instead.

>and the new API would obscure that bug.

Hmm, I thought if there are no special reasons then developers can choose to use existing methods so that they can detect double addition/removal.

donnerpeter:
Unfortunately people too often invoke a method with convenient-looking name without giving much thought :(

ninja-inc:
I understood.
Then how about making `public boolean existDocumentListener(@NotNull DocumentListener listener)`?
(1) If people invoke this method without giving much thought but I believe it will not obscure bugs.
(2) People can easily implement simple codes to verify whether a listener has already added or not.

But if you still think this method produces bad result or it is not needed in core API then I will close PR, if not then I will update codes. (If you have better method name, plz let me know)

ninja-inc:
@donnerpeter 
OK, thank you for the comment. Let me close this PR.

PR #731, opened 2018-02-11
173

baocatsamac:
I need IntelliJ IDEA source code for the purpose of developing a new plugin. Thanks!


alexander-doroshko:
@baocatsamac you do not need to create malformed pull requests to develop plugins for IntelliJ IDEA.
Start reading docs from this page: https://www.jetbrains.org/intellij/sdk/docs/welcome.html

PR #733, opened 2018-02-13
PY-28461: Fix completions for implicitly imported sub-packages from inside a package

fitermay:
@vlasovskikh This fixes PY-28461. I didn't want to add this to previous PR since it was already too large. 

btw do you know the purpose of the check `!inSameFile(location, importedModule)`  in PyModuleType? It looks suspicious to me.


vlasovskikh:
@fitermay `!inSameFile(location, importedModule)` was necessary for `PyMultiFileResolveTest#testRelativeFromSubmodule` where we had to add the original import for the imported module located in anther file. Actually it's OK to remove this check and always add the original import, there just may be duplicates. Making `importElements` a `Set<PyImportElement>` will likely overcome this duplication.

python/src/com/jetbrains/python/psi/types/PyModuleType.java
```diff
@@ -132,6 +109,103 @@ else if (location != null) {
     return null;
   }
 
+  private static void processImplicitPackageMembers(@NotNull PsiFileSystemItem anchor,
```
vlasovskikh: This method has become quite long. Could we split it into parts? Some parts that I see are:

* Collecting the list of elements that may introduce implicit names
* Searching for implicit names in the collected imports
* Searching for implicit names in the importable qualified name of the location module

python/src/com/jetbrains/python/psi/types/PyModuleType.java
```diff
@@ -395,51 +443,26 @@ private static CompletionVariantsProcessor createCompletionVariantsProcessor(Psi
...
-                ResolveImportUtil.resolveChildren(pyPackage, directChild, location.getContainingFile(), true, true, false, false);
-              result.addAll(ResolveResultList.getElements(results));
-            }
-          }
-        }
-      }
-    }
-    return result;
+    final List<PsiElement> elements = new ArrayList<>();
+    processImplicitPackageMembers(pyPackage, location, null, any -> true, results -> {
```
vlasovskikh: Thanks for unifying it!

python/src/com/jetbrains/python/psi/types/PyModuleType.java
```diff
@@ -132,6 +109,103 @@ else if (location != null) {
...
+            if (!resultProcessor.process(ResolveResultList.asImportedResults(results, packageImportElement))) {
+              return;
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private static String findFirstComponentAfterPrefix(QualifiedName qualifiedName, QualifiedName prefix) {
```
vlasovskikh: Could you please add appropriate nullity annotations?

python/src/com/jetbrains/python/psi/types/PyModuleType.java
```diff
@@ -132,6 +109,103 @@ else if (location != null) {
...
+    if (anchor.getVirtualFile() == null) {
+      return;
+    }
+    final PsiElement footHold = location != null ? location.getContainingFile() : module;
+
+    if (footHold == null) {
+      return;
+    }
+
+    final List<QualifiedName> myQnames = QualifiedNameFinder.findImportableQNames(footHold, anchor.getVirtualFile());
```
vlasovskikh: Our spellchecker thinks that `Qnames` is a typo. Could you rename it to `myQNames` or `myQualifiedNames` here and `locationQnames` below as well? Then the module will become green.

fitermay:
@vlasovskikh I've refactored it as you suggested , I think.  Some operations ended up being repeated twice, but I think it's probably better than passing a confusing number of parameters.

fitermay:
@vlasovskikh I've noticed a couple of issue that may be related to my previous set of changes. I've added fixes to them in this PR.

1)  Imported sub packages don't resolve in scratch files
2) Navigating to a package navigates to the directory instead of the package element (`__init__.py`)

vlasovskikh:

PR #735, opened 2018-02-15
py/pydev Add real directory of target to syspath

rohanjoyce:
See PY-27886


Elizaveta239:
Thank you very much! Your fix was merged.

PR #739, opened 2018-02-19
PY-28562: Add implicit `object` super when class has new style metaclass

fitermay:


vlasovskikh:
@fitermay Thanks for your pull request! I've merged it into master.

PR #740, opened 2018-02-20
JPS: add extension to provide additional constant affection resolvers

alexey-tsvetkov:
JVM languages such as Kotlin can inline constant values at use-site.
When a constant's value is changed, all use-sites must be recompiled.
Since values are inlined, there are no references in bytecode
so Java builder does not know when a Java constant is referenced in Kotlin.
Kotlin builder knows where each FQ-name is used, but does not know
when a Java constant is changed.

To fix KT-16091, Kotlin plugin needs to provide additional ConstantAffectionResolver,
so it would be notified when a constant value is changed.

In this commit ConstantSearchProvider class is added.
Additional instances of ConstantAffectionResolver provided by ConstantSearchProvider
implementations are used only when "main" ConstantAffectionResolver
(that searches using Intellij) is used (otherwise Java builder forces rebuild,
so it does not make sense).

// cc @chashnikov

### DIFF
--- a/jps/jps-builders/src/org/jetbrains/jps/builders/java/ConstantSearchProvider.java
+++ b/jps/jps-builders/src/org/jetbrains/jps/builders/java/ConstantSearchProvider.java
@@ -1,0 +1,17 @@
+// Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package org.jetbrains.jps.builders.java;
+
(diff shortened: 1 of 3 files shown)


eugenezh:
I have replaced it with the composite implementation of Callbacks.ConstantAffectionResolver which is used as a ConstantSearch service, if there are some custom extensions. So so code in Mappings remains the same

eugenezh:
Merged into master with changes, thanks!

PR #741, opened 2018-02-23
Synchronize access to mutable fields

dahlstrom-g:
… mainly to stop an NPE when closeInput dereferences myTask.

### DIFF
--- a/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/execution/ExternalSystemProcessHandler.java
+++ b/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/execution/ExternalSystemProcessHandler.java
@@ -69,7 +69,7 @@ public class ExternalSystemProcessHandler extends BuildProcessHandler implements
   }
 
   @Override
-  protected void destroyProcessImpl() {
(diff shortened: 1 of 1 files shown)


dahlstrom-g:
@vladsoroka?

platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/execution/ExternalSystemProcessHandler.java
```diff
@@ -69,7 +69,7 @@ public void notifyTextAvailable(@NotNull final String text, @NotNull final Key o
   }
 
   @Override
-  protected void destroyProcessImpl() {
+  protected synchronized void destroyProcessImpl() {
```
dahlstrom-g: Instead of the synchronized keywords, Radomir Malaczek suggests a local variable (Task task = myTask) here and in closeInput, in case you prefer that approach.

vladsoroka:
Please find the fix at https://github.com/JetBrains/intellij-community/commit/d6120cc97e657de5712ec807e6070ee945487497

PR #742, opened 2018-02-23
PY-28629: Fix resolving skeletons for implicit relative imports

fitermay:
@vlasovskikh 
I would really like to get this fixed before 2018 release as it is causing a lot of problems in Python 2.7 when dealing with older libraries. 

Tests pending... Is there a fixture I can use to setup the test skeletons?


fitermay:
@vlasovskikh added test

vlasovskikh:
@fitermay. Yes, the `WithBinaryModules` is a good place for testing skeletons. I've rebased your PR onto master with a couple of fixes. Thanks!

PR #744, opened 2018-02-24
Allow "Switch after create" when creating branch from repository location in SVN plugin

PLkolek:
The common scenario where I work is to create a branch from repository location (i.e. trunk) in one commit and then in a separate one commit the changes. Unfortunately the option to automatically switch after create is not available when branching from repository location, which forces us to switch manually, which can be easily forgotten and leads to accidental commits to wrong branch. This pull requests makes this checkbox available for both options.


pchel-:
Consider the following case:
- Some file is opened in the editor
- VCS -> Subversion -> Branch or Tag

Note that working copy location and repository location will be set according to the file in the editor.

My concern is that the user could set necessary repository location and "copy to" parameters - so the correct branch will be created. But "switch after create" option will switch only that one file (opened in the editor). And this could be confusing.

Probably we should disable this checkbox for some cases, or detect working copy path to switch (based on repository location) and specify it explicitly in the textbox text, or something else. What do you think?

PLkolek:
I'm sorry to react so late, but as I got assigned to another team at work and stopped using SVN, I had little motivation to pursue it futher. (Un)fortunately I am back to SVN, so I found some time to work on it.

I added a text field which informs the user which working copy location will be switched when selecting the checkbox. When choosing repository location as a source, it displays the file which was chosen when opening the "Create branch or tag" popup. I hope the typical case for this functionality is to branch & switch whole project (when it's clear what will happen) and in the other cases the provided information should alleviate the possible confusion. What do you think? ;)

vinterdo:
Any progress with this pull request? My team has same problem.

sjwilczynski:
My team is also interested. Is it going to be included?

pchel-:
I've made some changes (just specify path to switch in the checkbox label) and merged your pull request - see 38a852203808d08376cefedbe625960caa391444.
Thanks!

PR #747, opened 2018-02-28
Add versionDetailsProvider extension point to the platform.

adehtiarov:
This extension point allows the extensions to supply
application-specific version details in addition to the platform,
system and JDK information always populated by the platform by
default.

For example, Android Studio would provide additional details like
Gradle and Android Gradle Plugin versions, as well as versions
of some Android SDK components.

This is going to make it easier for users to collect version
information when they file bugs, which apart from the general
UX improvement would also reduce turnaround time as it'll be
less likely that a necessary detail might need to be requested
from the user additionally.

This CL also makes the relevant changes to SendFeedbackAction
and AboutPopup, which are the two sources of the version
information used by people when filing bugs.

### DIFF
--- a/platform/platform-impl/src/com/intellij/ide/actions/AboutPopup.java
+++ b/platform/platform-impl/src/com/intellij/ide/actions/AboutPopup.java
@@ -3,16 +3,21 @@ package com.intellij.ide.actions;
 
 import com.intellij.icons.AllIcons;
 import com.intellij.ide.BrowserUtil;
+import com.intellij.ide.DataManager;
 import com.intellij.ide.IdeBundle;
 import com.intellij.ide.ui.UISettings;
 import com.intellij.openapi.Disposable;
+import com.intellij.openapi.actionSystem.CommonDataKeys;
 import com.intellij.openapi.actionSystem.CustomShortcutSet;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationNamesInfo;
+import com.intellij.openapi.application.VersionDetailsProvider;
 import com.intellij.openapi.application.ex.ApplicationInfoEx;
 import com.intellij.openapi.application.impl.ApplicationInfoImpl;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.ide.CopyPasteManager;
 import com.intellij.openapi.project.DumbAwareAction;
+import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.GraphicsConfig;
 import com.intellij.openapi.ui.popup.JBPopup;
 import com.intellij.openapi.ui.popup.JBPopupFactory;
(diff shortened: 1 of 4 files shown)


adehtiarov:
@bulenkov @develar Hi folks, can you have a look? We'd like to get a feeling of how we should proceed with this for Android Studio: either merging the EP into platform & cherry-picking as per this PR, or limiting this to AS only (will not be accessible from platform). Thanks!

platform/platform-impl/src/com/intellij/ide/actions/AboutPopup.java
```diff
@@ -410,7 +419,27 @@ protected JBPoint getCopyIconCoord() {
...
+      for (VersionDetailsProvider versionDetailsProvider : VersionDetailsProvider.EP_NAME.getExtensions()) {
+        try {
+          Map<String, String> versionDetails = versionDetailsProvider.getVersionDetails(project);
+          for (Map.Entry<String, String> entry : versionDetails.entrySet()) {
+            sb.append(String.format(VERSION_KEY_VALUE_FORMAT, entry.getKey(), entry.getValue()));
+            sb.append(VERSION_KEY_VALUE_PAIRS_SEPARATOR);
+          }
+        }
+        catch (Throwable e) {
+          LOG.info("Exception while calling one of the version details providers", e);
```
bulenkov: Looks useless because contains no information about provider. Why info?

platform/platform-impl/src/com/intellij/ide/actions/AboutPopup.java
```diff
@@ -410,7 +419,27 @@ protected JBPoint getCopyIconCoord() {
     }
 
     public String getText() {
-      return myInfo.toString() + SystemInfo.getOsNameAndVersion();
+      Project project = CommonDataKeys.PROJECT.getData(DataManager.getInstance().getDataContextFromFocus().getResultSync());
```
bulenkov: We don't think that mixing project specific information with Application info is a good idea. Also, it could break user's privacy. However, EP for providing some project independent information like OS version, etc is a viable idea. I'll take a look what we can improve here

platform/platform-impl/src/com/intellij/ide/actions/SendFeedbackAction.java
```diff
@@ -43,11 +62,11 @@ public static void launchBrowser(@Nullable Project project) {
       .replace("$BUILD", eap ? appInfo.getBuild().asStringWithoutProductCode() : appInfo.getBuild().asString())
       .replace("$TIMEZONE", System.getProperty("user.timezone"))
       .replace("$EVAL", isEvaluationLicense() ? "true" : "false")
-      .replace("$DESCR", getDescription());
+      .replace("$DESCR", getDescription(project));
     BrowserUtil.browse(urlTemplate, project);
   }
 
-  public static String getDescription() {
```
bulenkov: It was a public static method with possible usages from plugins. It's better to wrap it with delegation to `getDescription(null);` instead of removing

batya239:
Hi, @adehtiarov we close this PR for now since it was no activity for 7 years. Feel free to re-open it once you'll address highlighted problems above. If you need some help with that please also ask us.

PR #751, opened 2018-03-08
Leak hunter sometimes uses stale roots

radomirm-g:
LeakHunter.processLeaks() gets a list of root objects, typically a result
of LeakHunter.allRoots() call. As a part of leak search, LeakHunter will
flush all outstanding tasks in AWT EventQueue, which should also include
all tasks schedules with LaterInvocator.

allRoots() call gets all LaterInvocator tasks and uses them as roots for
leaks search. This list will become stale (i.e. will include runnables that
were already executed and removed from the LaterInvocator queue) after
processLeaks() flushes all events. To fix this the list of roots should be
computed after all event flushing is completed.

Fix: processLeaks() now takes a supplier for roots collection, so the set
of root objects can be computed after all events are flushed.

This fix has already been applied in Android Studio source tree as some 
test runs intermittently reported failures because of these timing issues.

### DIFF
--- a/platform/testFramework/src/com/intellij/testFramework/LeakHunter.java
+++ b/platform/testFramework/src/com/intellij/testFramework/LeakHunter.java
@@ -36,6 +36,7 @@ import org.jetbrains.annotations.TestOnly;
 
 import javax.swing.*;
 import java.util.*;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


cdracm:
pushed, thanks

PR #757, opened 2018-03-13
Add support for nullability annotations in androidx package

tnorbye:
This is similar to https://github.com/JetBrains/kotlin/pull/1559


chashnikov:
Merged manually to 'master' branch so it'll be included into 2018.2 release. Do you need to have this in 2018.1 updates?

tnorbye:
It would be ideal to get it into 2018.1 but it's not strictly necessary. We're going to start shipping libraries with these annotations in them pretty soon, but they won't be stable right away, so this is most important for people who are early adopter types, and they're probably more likely to be using 2018.2 EAPs? (I think for the corresponding Kotlin compiler change they're putting it into 1.2.40 and maybe 1.2.31 if there's a 1.2.31 for other bug fixes.)

chashnikov:
We don't start 2018.2 EAP soon so I've backported the change to 181 branch. It'll be included into 2018.1.1  bugfix update.

PR #758, opened 2018-03-14
fix NPE: PsiFile.getContainingDirectory can return null

fitermay:
PyUtil.turnInitIntoDir can return null for not null values. In my case this problem caused the reference highlighter to throw an exception in the Diff Viewer.


sproshev:
Merged, a4cb153596a00c6064d69811059ed9096d032fc2

PR #759, opened 2018-03-14
Disable rerunImportAction when it is performed on refreshProject

srmurguia:
There is a "Refresh Project" button on the Build view, when this button
is pressed, the rerunImportAction action inside
ExternalSystemUtil#refreshProject is performed, but if you double click
this button, this action is performed twice.

This button is not disabled until the update method is called and inside it,
processHadles.isProcessTerminated() returns false. This causes a small
time window in wich this button can be cliicked multiple times.

This change disables this button when this action is performed instead
of waiting for update and isProcessTerminated to catch up.


ignatov:
cc @vladsoroka

vladsoroka:
Thanks for the contribution! I've merged the change using the following commit https://github.com/JetBrains/intellij-community/commit/90c2bb3183e861256565c78c94a0a4382e219e54

PR #760, opened 2018-03-14
Fix ThreeComponentsSplitter focus cycle

PureSin:
Allow tab to navigate focus between the 3 panels of
ThreeComponentsSplitter by changing the traversal policy and removing
ThreeComponentsSplitter as the focus cycle root. This allows for users
of the component the control of the focus cycle instead.

Also allows for StripeButton to be focused.

JetBrains Contributor License Agreement signed under kelvinhanma@google.com

### DIFF
--- a/platform/platform-api/src/com/intellij/openapi/ui/ThreeComponentsSplitter.java
+++ b/platform/platform-api/src/com/intellij/openapi/ui/ThreeComponentsSplitter.java
@@ -72,75 +72,7 @@ public class ThreeComponentsSplitter extends JPanel implements Disposable {
 
   private boolean myShowDividerControls;
   private int myDividerZone;
-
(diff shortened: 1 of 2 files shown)


PureSin:
@denis-fokin would you take a look based on your most recent edit on ThreeComponentSplitter. This PR helps fix some a11y issues in Android Studio. Thanks

denis-fokin:
@PureSin The idea of the traversal policy is to find the best child in a split component. As far as I remember LayoutFocusTraversalPolicy does not always satisfy this approach. The problem is in the split component structure. Usually, they are composite components and layout does not suggest the best component to focus.

We have a ScreenReader class which is supposed to enable accessibility features. Could you surround your code by ScreenReader methods?

PR #761, opened 2018-03-15
Symmetric start and end points for init ProgressIndicator

SokolovMS:
Fixed with changing total width value.

Before (Now):
Before
After:
After


bulenkov:
The proposed change will break desing guidelines for Splash screens. Start and end points are correct in current implementation

PR #762, opened 2018-03-16
Replace echo with printf to interpret \n as newline

dahlstrom-g:
see https://wiki.ubuntu.com/DashAsBinSh#echo


dahlstrom-g:
@trespasserw?

trespasserw:
Yep, thanks. But then it needs a trailing newline.

dahlstrom-g:
done (in amended commit)

PR #763, opened 2018-03-19
terminal: quote shell variable when testing

rohanjoyce:
In the case LOGIN_SHELL is undefined, the test becomes:

if [ -n ]; then
...
fi

which will execute the block inside the if, when it shouldn't.


traff:
@rohanjoyce Thanks! I've merged the fix in.

PR #765, opened 2018-03-20
Add support for Geb content definition based methods

erdi:
This PR adds support for Geb's content definition based methods (as documented in [this chapter](http://gebish.org/manual/current/#content-dsl) in `TemplatedPageWithDiv` example and implemented [here](https://github.com/geb/geb/blob/f1294846366e7c28f7d8d04ba14c95319cd2adad/module/geb-core/src/main/groovy/geb/content/PageContentSupport.groovy#L31)) to complement the current support of content definition based fields. 

Do not hesitate to let me know if this PR needs more work before it can be merged. I'll happily apply any necessary changes to it.

Please note that I have signed the CLA.

Given that this contribution is related to Groovy, I believe that the best person to review it is @dovchinnikov.

### DIFF
--- a/plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/geb/GebContentDeclarationSearcher.java
+++ b/plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/geb/GebContentDeclarationSearcher.java
@@ -17,10 +17,7 @@ package org.jetbrains.plugins.groovy.geb;
 
 import com.intellij.pom.PomDeclarationSearcher;
 import com.intellij.pom.PomTarget;
-import com.intellij.psi.PsiClass;
-import com.intellij.psi.PsiElement;
-import com.intellij.psi.PsiField;
-import com.intellij.psi.PsiModifier;
+import com.intellij.psi.*;
 import com.intellij.psi.util.InheritanceUtil;
 import com.intellij.util.Consumer;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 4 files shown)


dovchinnikov:
Hi @erdi! Thank you for the pull request.

Note that both `PsiField` and `PsiMethod` implement `com.intellij.psi.PsiTarget`. Could you please try to use it instead of introducing new `PsiElementPomTarget` class?

erdi:
Using `PsiElement` as the value type for the map returned from `GebUtil.getContentElements()` was what I tried initially but obviously it does not implement `PomTarget`. I completely missed `PsiTarget` though, I will look into updating the PR to use `PsiTarget` ASAP, it should simplify it massively.

erdi:
I've just force pushed the latest changes. In the end I went with using `PsiMember` as the value type for the map returned from `GebUtil.getContentElements()` because `PsiTarget` does not implement `PsiElement`.

erdi:
Thanks for accepting the PR. 

Am I correct in thinking that it will be part of the 2018.2 release?

dovchinnikov:
@erdi yes, it's planned for 2018.2, but I think it's safe to backport it to 2018.1.2 or 2018.1.3.
Would you like me to do it?

erdi:
I don't have a specific need for it to be backported and do not want to increase your workload unnecessarily given that I might be sending additional PRs in that area your way. I only asked to have an idea which release it would be part of to tweet about it from Geb's account when it ships.

dovchinnikov:
@erdi that's ok, it shouldn't take much time, and it won't require any PRs from your side. You may safely tweet about 2018.2 and then it will be just a nice bonus if we decide to backport it to 2018.1.X.

PR #767, opened 2018-03-22
PY-29092: Merge PYTHONPATH instead of overriding in console debugger

fitermay:


Elizaveta239:
Merged in PyCharm 2018.2

PR #770, opened 2018-03-24
use equals for module name comparison

guoci:


ignatov:
@traff, please reassign the PR to the right person if needed.

PR #775, opened 2018-04-02
map scipy/numpy objects to webpage name

guoci:
map objects in numpy and scipy to the corresponding html page


python/src/com/jetbrains/python/documentation/PythonDocumentationMap.java
```diff
@@ -29,6 +37,28 @@ public static PythonDocumentationMap getInstance() {
     return ServiceManager.getService(PythonDocumentationMap.class);
   }
 
+  private static class ScipyExternalDocumentationMap {
+    private static final Map<String, String> nameToWebpageName = ((Supplier<Map<String, String>>)() -> {
+      final ImmutableMap.Builder<String, String> b = ImmutableMap.builder();
+      try (final BufferedReader inputStream = new BufferedReader(
+        new InputStreamReader(
+          ScipyExternalDocumentationMap.class
+            .getResourceAsStream("/com/jetbrains/python/documentation/scipyNameMapping.tsv"),
```
traff: How did you create this file? It looks like it is possible to reduce it to several regexp, which would be easier to maintain.

guoci: I created that with a Python script. I don't think it is easy to simplify it to regexp, but I can release my script.

guoci: Link to my script for generating the mapping to the external documentation:
https://gist.github.com/guoci/9421d87a717168c79ee0403f0b213ae9

traff: Thanks!

traff:
@guoci Thanks for the pull-request. I refactored it a bit and merged in.

PR #776, opened 2018-04-05
PY-29338: Sort out foreign relative imports from python ones

fitermay:
@sproshev @vlasovskikh The problem is resolving relative imports, foreign imports and relative skeletons are all tangled up in the same codepath. The returned foreign imports were not ordered properly relative to the rest of the results causing the selected result to be a foreign import (PsiPackage) when a python one was available. This fix is far from perfect (and also didn't have time to write test) but  it should fix the issue for now without breaking anything else.  It will also effect caches in cases when a package can be imported as a relative import or absolute under exactly the same name. I don't expect it to meaningfully affect performance though.

### DIFF
--- a/python/src/com/jetbrains/python/psi/resolve/PyResolveImportUtil.kt
+++ b/python/src/com/jetbrains/python/psi/resolve/PyResolveImportUtil.kt
@@ -28,10 +28,8 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.projectRoots.Sdk
 import com.intellij.openapi.roots.FileIndexFacade
 import com.intellij.openapi.vfs.VirtualFile
-import com.intellij.psi.PsiDirectory
-import com.intellij.psi.PsiElement
-import com.intellij.psi.PsiFileSystemItem
-import com.intellij.psi.PsiManager
+import com.intellij.psi.*
+import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.psi.util.QualifiedName
 import com.jetbrains.python.codeInsight.typing.PyTypeShed
 import com.jetbrains.python.codeInsight.userSkeletons.PyUserSkeletonsUtil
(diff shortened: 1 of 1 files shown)


fitermay:
@sproshev @vlasovskikh I was considering taking a shot at cleaning up the import logic for a while now but there are some questions I'm still thinking about. I'd like to know your team's thoughts about how to resolve those issues. Was going to write an email to @vlasovskikh but may as well post it here.


1. The configuration model does not correctly capture the runtime model. That is ,the runtime can be configured in more ways than can be captured by the configuration. For example, we can run with content roots, source roots, neither or both. This is not expressed in the configuration so we must assume that we are including all these roots during static analysis. This essentially forces us to guess the qualified names of modules and packages. findShortestImportableQ

vlasovskikh:
@fitermay Answering to your points:

1. The user is expected to mark their source roots properly to explain their run-time `sys.path` model to the code insight engine of PyCharm. Usually it's enough to mark source roots in the correct order to achieve the necessary ordering. If you have any specific examples where we can improve `findShortestImportableQName` I'll be interested in looking into them.

2. `findCanonicalImportPath` is needed for things covered by `PyCanonicalPathProvider` like `PyStdlibCanonicalPathProvider`. For example, `os.path.join` is defined in `posixpath.join`, but we shouldn't use this name in code insight and prefer `os.path.join` instead.

3. Currently we use only `RATE_HIGH`, `RATE_NORMAL`, and `RATE_LOW`, which was enough for a while and maybe it's still enough. As for foreign resolve results, I was thinking about introducing "overriding" foreign import providers that take precedence over normal resolve results.

fitermay:
@vlasovskikh 
1) I'll address this in more detail later. For now I just want to make a few simple points:
         

- Why even allow the user to choose a run configuration that doesn't match the module config? Why not always run with source roots and content roots?
- Excluding the content root from the python path is not even expressible in the module configuration
- I expect that resolving qualified names precisely will become more of an issue as namespace packages become more popular. For instance, when resolving import spanning namespace packages if the qualified name is guessed wrong then those imports cannot be resolved correctly

2) I understand that findCanonicalImportPath normalizes reexported names. There are two steps for this process.  The first step is `smart` and expensive re-normalization --  it looks for the topmost package from which the name can be imported.  The second step is straightforward iteration through `PyCanonicalPathProvider`s. What I'm wondering is whether the first step is strictly necessary here.

3) All I'm saying is that combining those result lists is a very frequent source of bugs.  I think this is so because a result's rating does not convey enough information about the kind of result it is.

vlasovskikh:
@fitermay There are users who want to be able to run their scripts with `sys.path` left intact. For example, one may have some scripts unrelated to the rest of the project that assume that they are run outside of the context of the project:

```
project1/
  scripts/
    script1.py
  lib1/  # <- this should not be available in script1.py
    __init__.py
    module1.py
```

but I'm sure there are other examples when users want full control on their `sys.path`.

fitermay:
That’s a good example. So if they need to run that way and our code
analysis assumes differently won’t our static analysis be wrong?

vlasovskikh:
It might be wrong, but the ability to run stuff the way the users expect is more important for them. When users report about their problems with false warnings, we advise them to make their static configuration reflect their dynamic `sys.path`, but it's not always easy, sometimes even not possible.

vlasovskikh:
@fitermay Thanks for your quick solution to this problem! Unfortunately, the solution proposed in this pull request isn't clean enough and it will complicate future refactorings of imports. You're right about the fact that the code tangles together foreign imports, relative stubs, and binary skeletons. I had to revert the fix for [PY-28629](https://youtrack.jetbrains.com/issue/PY-28629) that resulted in this regression since the issue was affecting too many users.

It would be great to come up with the refactoring of the resolving imports that also will make it possible to address the reverted issue with the skeletons in a cleaner way.

vlasovskikh:
@fitermay Yes, sure.

PR #778, opened 2018-04-11
IDEA-189919 provide a setting for property comments not convert to ascii

xinkunZ:
* Provide a feature like eclipse properties editor plugin：
pic

* default behavior is convert comments, if you don't want to convert comments to ascii,  a string below need add to the $IDEA_INSTALLATION/bin/idea.properties
`idea.native2ascii.skip.comments=true`
* then, you need to add a string to tell IDEA what line sperator in use.  by default, it is "\r\n",you can change to "\n" or "\r"
 `idea.native2ascii.line.separator=\r\n`


ignatov:
cc @batya239

PR #779, opened 2018-04-11
Fix the type reported in completion dialog for field based Geb page content

erdi:
As previously mentioned to @dovchinnikov, this is another PR around support for Geb.

I've noticed that in recent versions of IntelliJ the return type for field based Geb page content always comes up on the completion dialog as `Object` even though the right type seems to be known and used when completing symbols on the field itself. 

This fixes the above issue.


erdi:
Is there anything I can do to help in this getting merged?

dovchinnikov:
Sorry for long response. All is fine, I've just had not enough time to actually merge the request. 
I'll do it before the next EAP.

PR #780, opened 2018-04-12
terminal: disable bracketed paste (IDEA-177768)

rohanjoyce:


ignatov:
cc @traff

traff:
@rohanjoyce Thanks for the PR, but the better solution would be to support bracketed paste.

PR #782, opened 2018-04-22
IDEA-190710 Add Gitlab Time Tracking integration

llamositopia:
This allows for GitLab Time Tracking through the Tasks plugin. GitLab already supports this feature, but never got support in the IDE.


llamositopia:
GitLab v11 has removed the API v3 and this PR is no longer valid.

PR #784, opened 2018-05-03
Fix L&F change in ProjectView panel

dahlstrom-g:
Switching L&F didn't use to update the ProjectView panes except the one
that was displayed at the time.

http://b.android.com/208136


dahlstrom-g:
@SergeyMalenkov? @xtmq?

bulenkov:
Thanks! Merged 210f54d14e0b99fd0fc4f08b66925b2379250198

dahlstrom-g:
@gregsh Do you prefer to revert commit 210f54d then?
@jgaillard85 What's your understanding of the situation?

bulenkov:
@gregsh Thanks! I started Android Studio from sources from master and can't reproduce the problem with or without the fix. So, we probably need to revert it. 

P.S.
The only thing I noticed is wrong foreground for Design Editor
image

@dahlstrom-g Please, use JBColor with function for such cases 
image

PR #786, opened 2018-05-06
PY-26863 Support 3dsmaxpy.exe

cmcpasserby:
https://youtrack.jetbrains.com/issue/PY-26863

Simply support to allow PyCharm to recognize 3ds Max's `3dsmaxpy.exe`, this approach is based on the PythonSdk flavor approach that was used to make PyCharm work with Maya's `mayapy`.

### DIFF
--- a/python/src/META-INF/python-core-common.xml
+++ b/python/src/META-INF/python-core-common.xml
@@ -774,6 +774,7 @@
                    implementationClass="com.jetbrains.python.codeInsight.functionTypeComments.PyFunctionTypeAnnotationVisitorFilter"/>
 
     <pythonFlavorProvider implementation="com.jetbrains.python.sdk.flavors.MayaFlavorProvider"/>
+    <pythonFlavorProvider implementation="com.jetbrains.python.sdk.flavors.MaxFlavorProvider"/>
     <pyPregeneratedSkeletonsProvider id="default" implementation="com.jetbrains.python.sdk.skeletons.DefaultPregeneratedSkeletonsProvider"/>
   </extensions>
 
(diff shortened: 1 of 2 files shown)


vlasovskikh:
@cmcpasserby Thanks for your pull request! I've noticed that there are not that many votes for this issue. Have you considered making it a separate plugin like [the one that supports Maya](https://plugins.jetbrains.com/plugin/8218-mayacharm)? I guess there will be other feature requests specific to 3ds Max that the users might want to see in the future. These improvements will be easier to implement and maintain as a separate plugin.

cmcpasserby:
yeah, i can easily do that. I made it a PR since PyCharm supports the mayapy interpreter name out of the box already.

vlasovskikh:
@cmcpasserby The reason PyCharm supports `mayapy` inside its code base is that historically the Maya integration was a part of PyCharm. When a third-party plugin with extra features for Maya had appeared, we decided to drop built-in support for Maya, leaving just the ability to specify it as an interpreter for backwards compatibility.

vlasovskikh:
@cmcpasserby I've just realized that you're the author of MayaCharm, sorry for explaining obvious facts to you. Are you OK with having 3ds Max support as a third-party plugin? I'll close this pull request then.

cmcpasserby:
ok, yeah might as well close it then and i will submit it as a plugin to the repo later this week then.

@vlasovskikh so currently my MayaCharm plugin does not add a sdk flavour for mayapy since it was already a supported interpreter. Should i add it as a SDK flavour to the plugin as well now if you say the built in support will be dropped?

vlasovskikh:
@cmcpasserby With the SDK flavor for Maya we just wanted not to break backwards compatibility for those who had been using our Maya integration without your plugin. For 3ds Max it's better to start off with a separate plugin. If it get popular and if the license will permit it, we might consider bundling it.

Thanks for your MayaCharm plugin and the future plugin for 3ds Max!

PR #787, opened 2018-05-07
Fix ThreeComponentsSplitter focus cycle

PureSin:
Allow tab to navigate focus between the 3 panels of
ThreeComponentsSplitter by changing the traversal policy and removing
ThreeComponentsSplitter as the focus cycle root. This allows for users
of the component the control of the focus cycle instead.

Also allows for StripeButton to be focused.

@denis-fokin will use this PR to follow up on your comment: https://github.com/JetBrains/intellij-community/pull/760#issuecomment-387036701


PureSin:
So you're ok with this change if I surround the new code with ` if (ScreenReader.isActive()) {` ? 

```
 if (ScreenReader.isActive()) {
  new code 
} else {
  old code
} 
```

?

ignatov:
cc @denis-fokin

alexander-lobas:
Sorry, but the fix it outdated. We've rewritten this place several times.

PR #789, opened 2018-05-08
[PsiTreeUtil] adding findChildrenByFilter() method

ArtsiomCh:
That will give flexibility of filtering children not just by  Type (Class) but aslo by any conditions.

Also existing bottom most `findChildrenOfAnyType()` could be rewritten to something like that:
```Java
  @SafeVarargs
  @NotNull
  public static <T extends PsiElement> Collection<T> findChildrenOfAnyType(@Nullable PsiElement element,
                                                                           boolean strict,
                                                                           @NotNull Class<? extends T>... classes) {
    return findChildrenByFilter(element, filteredElement -> 
         !(strict && filteredElement == element) && (instanceOf(filteredElement, classes)));
  }
```
PS Same kind of methods could be done for `findChildOfType()` and others...

### DIFF
--- a/platform/core-api/src/com/intellij/psi/util/PsiTreeUtil.java
+++ b/platform/core-api/src/com/intellij/psi/util/PsiTreeUtil.java
@@ -336,6 +336,24 @@ public class PsiTreeUtil {
   }
 
   /**
+   * Recursive (depth first) search for all elements filtered by given {@code filter}.
+   *
+   * @param element a PSI element to start search from.
+   * @param filter element to search for.
+   * @param <T>     type to cast found elements to.
+   * @return {@code Collection<T>} of all found elements, or empty {@code List<T>} if nothing found.
+   */
+  @NotNull
+  private static <T extends PsiElement> Collection<T> findChildrenByFilter(@Nullable final PsiElement element,
+                                                                           @NotNull final PsiElementFilter filter) {
+    if (element == null) return ContainerUtil.emptyList();
+    PsiElementProcessor.CollectFilteredElements<T> processor =
+            new PsiElementProcessor.CollectFilteredElements<T>(filter);
+    PsiTreeUtil.processElements(element, processor);
+    return processor.getCollection();
+  }
+
+  /**
    * Non-recursive search for element of type T amongst given {@code element} children.
    *
    * @param element a PSI element to start search from.


donnerpeter:
Why do you need this?

ArtsiomCh:
Let's say we need to find all elements by Type with specific Text in a file.
Presently, we have to fetch all elements by Type and then for-each them by Text:
```java
for (ElementType element : findChildrenOfType( file, ElementType.class)) {
   if (  element.getText().equals("-some text-")  ) {
     result.add(element);
   }
```
With `n` elements in the file and `m` elements of ElementType, that will take O(n*m). Also, that will envolve allocating space for the collection of all `m` elements of ElementType while we might need only few of them at the end. We might hope for some smart optimisations inside JVM, but I would not relay on that...

With suggested method all checks will happened once per each element withing the file, so it will take O(n) and we'll have to allocate the collection with only searched elements.

Another solution would be copy-past and modify `findChildrenOfType()` code to `findChildrenOfTypeWithText()` or even provide some kind of `comparator` and use it inside `PsiElementProcessor.execute()`. But we already have `CollectFilteredElements` subclass with `PsiElementFilter` filter that will do all needed works for us.

PS Full "production" code of usage example is here: [CMakePSITreeSearch.java](https://github.com/ArtsiomCh/CMake/blob/dev/src/com/cmakeplugin/utils/CMakePSITreeSearch.java)

donnerpeter:
Why do you need it in `PsiTreeUtil`? I'd prefer not to add API methods unused in our codebase, just for one plugin.

BTW note that filtering by `getText` is extremely inefficient, especially when invoked for all children recursively in large composite elements.

There's also quite handy `SynTraverser.psiTraverser(root)` API which gives you an `Iterable` of all its children recursively, which you can filter as you like.

ArtsiomCh:
Actually,  I don't need it in `PsiTreeUtil`. I already implemented it in my package. I thought it might be useful for others to have universal search method instead of multiples `findChildrenOf[Any]Type()` and `findChildOf[Any]Type()` for each specific case, all of which at the bottom comes to the same filtering inside `PsiElementProcessor.execute()` method. Its mostly been just suggestion for next refactoring round of `PsiTreeUtil`. Hence, we can close this PR.

As for `getText`, unfortunatelly it's the only way I can see to find element with variable declaration in CMake psi after filtering for `element instanceof CMakeLiteral`.

Thank you for `SynTraverser.psiTraverser(root)`! Will explore it and check if it speeds up search more then using `CollectFilteredElements`.

donnerpeter:
`SynTraverser` is unlikely to speed things up, but might deliver comparable performance.

I'd also suggest to try `PsiElement#textMatches`, it's more efficient than querying `getText` of everything

ArtsiomCh:
Just checked `textMatches()` and didn't find any visible speed up for my case.
Also looking into `textMatches()` declaration at `PsiElement` interface:
```java
  //Q: get rid of these methods?

  /**
   * Checks if the text of this PSI element is equal to the specified character sequence.
   *
   * @param text the character sequence to compare with.
   * @return true if the text is equal, false otherwise.
   */
  @Contract(pure=true)
  boolean textMatches(@NotNull @NonNls CharSequence text);
```
And used implementation at `PsiElementBase` class: 
```java
  //Q: get rid of these methods?
  @Override
  public boolean textMatches(@NotNull CharSequence text) {
    return Comparing.equal(getText(), text, true);
  }
```
I can see two issue:
1. Does `//Q: get rid of these methods?` means it will be deprecated soon?
2. `textMatches()` still using `getText()` inside: 
Comparing.equal(**getText()**, text, true); 
So, do you mean `Comparing.equal()` should be faster then `String.equal()` ? Quickly looking through the code of both of them I don't see any significant difference except possible slowdown at `anObject instanceof String`...

donnerpeter:
Sorry, I mistakenly thought that `ASTDelegatePsiElement#textMatches` delegates to `TreeElement#textMatches`, which is indeed faster because it doesn't need to traverse all text if the beginning doesn't match. But, surprisingly, it doesn't. I'll try to correct that.

The `//Q` comment is outdated, I don't think `textMatches` will be deprecated soon.

ArtsiomCh:
Actually, `Comparing.equal()` is roughly 20% faster then `String.equal()`...
Very naive test:
```java

public class Main {

    runEmpty();
    runString();
    runComparing();
  }

  private static void runEmpty () {
    long t= System.currentTimeMillis();
    long end = t+1000;
    long countStr = 0;
    while(System.currentTimeMillis() < end) {
      countStr++;
    }
    System.out.printf("%20s %10d\n", "Empty test:", countStr);
  }

  private static void runString () {
    long t= System.currentTimeMillis();
    long end = t+1000;
    long countStr = 0;
    while(System.currentTimeMillis() < end) {
      Long.toString(countStr).equals( "aaa");
      countStr++;
    }
    System.out.printf("%20s %10d\n", "String.equal():", countStr);
  }

  private static void runComparing () {
    long t= System.currentTimeMillis();
    long end = t+1000;
    long countStr = 0;
    while(System.currentTimeMillis() < end) {
      Comparing.equal(Long.toString(countStr), "aaa", true);
      countStr++;
    }
    System.out.printf("%20s %10d\n", "Comparing.equal():", countStr);
  }
}
```
Results:
```
         Empty test:   21844574
     String.equal():    9300748
  Comparing.equal():   11779188

Process finished with exit code 0
```

ArtsiomCh:
`Comparing.equal()` and `String.equal()` also stop to traverse at first unmatched char. So `TreeElement#textMatches` (implementation at `LeafElement#leafTextMatches`, right?) will not give any advantages over them. 

PS Can't stop to play around. :) In the performance test above changing "comparing against" string to:
`Long.toString(countStr)` - same string test;
```
         Empty test:   21463406
     String.equal():    7149935
  Comparing.equal():    8053432
```
`Long.toString(countStr-1)` - almost same string (9/10 cases), last char differ;
```
         Empty test:   22143264
     String.equal():    6801718
  Comparing.equal():    7689098
```
`Long.toString(countStr>>1)` - same lenght (9/10 cases) but differents chars.
```
         Empty test:   21650424
     String.equal():    7098570
  Comparing.equal():    7950961
```
Still getting `Comparing.equal()` faster then `String.equal()` on 15-25%. The only explanation I have is slowdown on typecasting `anObject` to `String` at `String.equal()`.

donnerpeter:
The most interesting implementation of `textMatches` is in `CompositeElement`, it doesn't traverse whole subtree if it's not needed.

If you're interested in comparing performance of such small methods, I'd recommend using JMH.

PR #790, opened 2018-05-13
[PY-27594] disable debug when Python gets input from stdin

gaborbernat:
See https://youtrack.jetbrains.com/issue/PY-27594 for more details. Resolves https://github.com/fabioz/PyDev.Debugger/issues/103. 

Note this is a workaround, as it will have the effect of not breaking things (as it does now). However, it will not allow users to debug library calls from within the stdin content. 

If we could interpret the whole input, write it to a temp file, that would allow the user to debug. However, I think that will require a significant refactor of this monkey patching.


gaborbernat:
Any updates on this? Thanks!

Elizaveta239:
@gaborbernat Thank you! Merged.

gaborbernat:
Thanks!

PR #791, opened 2018-05-22
Rename library to javax-inject

dahlstrom-g:
(cherry picked from commit 7b13127)

### DIFF
--- a/.idea/libraries/javax_inject.xml
+++ b/.idea/libraries/javax_inject.xml
@@ -1,5 +1,5 @@
 <component name="libraryTable">
-  <library name="javax.inject:javax.inject:1" type="repository">
+  <library name="javax-inject" type="repository">
     <properties include-transitive-deps="false" maven-id="javax.inject:javax.inject:1" />
     <CLASSES>
       <root url="jar://$MAVEN_REPOSITORY$/javax/inject/javax.inject/1/javax.inject-1.jar!/" />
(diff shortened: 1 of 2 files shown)


.idea/libraries/javax_inject.xml
```diff
@@ -1,5 +1,5 @@
 <component name="libraryTable">
-  <library name="javax.inject:javax.inject:1" type="repository">
```
dmitry-avdeev: Why do you need this? This is a standard way to name a maven-based library.

dahlstrom-g: 1. For consistency with the overwhelming majority of library names in the project.
2. For readability, since the current name repeats the same word twice.
3. Eliminating the version number from a library names lets you update the library without changing all the orderEntry dependencies in .iml files, which often reach from intellij-community to android, and so on.

dmitry-avdeev: I would consult @chashnikov first.

dahlstrom-g: Sounds good. Thanks!

chashnikov: I agree with Dana. I also would add two more reasons:
3a. When you update name of a project library because its version is changed it leads to renaming the library configuration xml file in .idea/libraries and therefore complicates analyzing git history for that library.
3b. When you update name of a project library because its version is changed it may break other projects which are built on top of IntelliJ-community and reuse its project configuration (e.g. IntelliJ IDEA Ultimate project).
So I'll merge this pull request.

chashnikov:
Merged to 'master' manually, thank you!

PR #792, opened 2018-05-22
build scripts: allow products to define special icons for EAP builds

dahlstrom-g:
… following commit 6e3a084 for Mac icns files only.


dahlstrom-g:
@chashnikov?

chashnikov:
Note that it isn't enough to copy proper icons to the product distributions, the IDE also need to use them at runtime. This is now fixed in 183.* builds, see [IDEA-192585](https://youtrack.jetbrains.com/issue/IDEA-192585) for details.

PR #793, opened 2018-05-23
In my opinion, there aren't many people that committed message same as previous

first087:


klikh:
The commit message is preselected, so if you don’t need it, just start typing, and it will be replaced with yours.

Seeing the previous message, or even reusing it at least partially could be useful. With the default changed, people who find it useful will not have a message out of the box, and will need to go enable it first, while those who do not need it just type their own.

If you don’t need this in any of your projects, configure the option in Default settings, so all your new project have this option enabled.

first087:
@klikh My team dislike this default option because sometime forget to change commit message.

first087:
Hi @klikh 

I think this problem is due to user error because position of commit message box is bad UI/UX.

When I open the commit dialog, I'll select the change file (top position). and review changes (bottom position), then click the commit button.

If I do not forget to change the text. When I click on the text box. Selected text will be discarded. The message must be re-selected. It's not easy.

If the commit message box move to near commit button, it will reduce the problem.

Hope you will review this case again.

klikh:
Usually the commit message is considered as a very important part of the commit, so it is strange to forget to change it. Nevertheless, we are currently discussing a new layout of the commit dialog, so it may be changed in one of future releases. 

Meanwhile your team members can change that option manually. If you create new projects frequently, you can even change the default value of option for new projects (`File | Other Settings | Preferences for New Project...`).

PR #794, opened 2018-06-01
Pass caught exception when generating ExternalSystemException

srmurguia:
This will cause the message to not be empty when getMessage(rootCause)
returns an empty string.


vladsoroka:
Thanks for the contribution, I've merged the change using the following commit https://github.com/JetBrains/intellij-community/commit/557554107f69cab9850c61ecee9899cfc44c09a6

PR #795, opened 2018-06-01
IDEA-193153 Allow BaseRefactoringProcessor subclasses to veto RefactoringHelpers and to provide custom UsageSearcher

diegopl:
In Android Studio we've found two cases where we could use some additional extensibility for BaseRefactoringProcessor:

- In one of them, it would be good to allow subclasses to selectively disable RefactoringHelpers. For example, OptimizeImports seems to sometimes undo the refactoring changes since it can not find the new classes yet (the project has not been synched).

- The other is allowing the subclasses to provide their own UsageSearcher. In the case of Android Studio it allows us to run refreshElements outside of a read action. That way, we can run parts of refreshElements on the UI thread and allows us to create migration packages and classes.

### DIFF
--- a/platform/lang-impl/src/com/intellij/refactoring/BaseRefactoringProcessor.java
+++ b/platform/lang-impl/src/com/intellij/refactoring/BaseRefactoringProcessor.java
@@ -256,17 +256,17 @@ public abstract class BaseRefactoringProcessor implements Runnable {
     }
   }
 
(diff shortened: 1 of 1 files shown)


ignatov:
Closing due to the pending request https://youtrack.jetbrains.com/issue/IDEA-193153.
Please reopen if needed.

PR #796, opened 2018-06-01
Add @Nullable annotation to DomElement.getXmlTag()

vcaen:
This was misleading consumer of xmlTag causing NPE in Android Studio.
Also changed some koltin code that was expecting a non-nullable type.


vcaen:
If we mark this method as nullable, implementation class can still mark it as not-null I believe.
I really think it would help to have the annotation, whether it is then handled with an exception or a null check. At least there wouldn't be any bad surprise.

donnerpeter:
Reasonable. Thanks for the suggestion, I performed the change myself in master.

PR #797, opened 2018-06-02
Add an extension point for checking plugin versions during plugin updates

juan-chen:
The main purpose of this PR is to allow Android Studio to decide which Kotlin plugin to advertise to users.

A following PR to Kotlin will change the Kotlin plugin update flow to use this extension point.

### DIFF
--- a/platform/platform-api/src/com/intellij/ide/plugins/PluginVersionChecker.java
+++ b/platform/platform-api/src/com/intellij/ide/plugins/PluginVersionChecker.java
@@ -1,0 +1,28 @@
+// Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package com.intellij.ide.plugins;
+
+import com.intellij.openapi.extensions.ExtensionPointName;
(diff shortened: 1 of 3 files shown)


platform/platform-api/src/com/intellij/ide/plugins/PluginVersionChecker.java
```diff
@@ -0,0 +1,28 @@
...
+
+public interface PluginVersionChecker {
+
+  ExtensionPointName<PluginVersionChecker> EP_NAME = new ExtensionPointName<>("com.intellij.pluginVersionChecker");
+
+  /** Returns a priority for the checker, which is used to pick a checker if there are more than one. */
+  int getPriority();
+
+  /** Returns a pair of:
+   *   the latest Kotlin plugin version for the current build number and channel;
```
Alefas: If the only purpose is Kotlin, it's a bad idea to have it in IntelliJ platform. Alternatively, this code should do something more than just an interface in the platform and it should be possible to reuse it for other plugins as well.

juan-chen:
This PR is no longer needed. Closing

PR #798, opened 2018-06-02
Fix an error when the HEAD is detached

ahmadshahwan:
The error occurred when there is no current branch (e.g. a commit is checked out).

Fixed by checking if the parameter is null before passing it to `getBranchPresentation()` method.

### How to reproduce error
* Checkout a previous commit that is not on top of a branch.
* Try to change the branch from the bottom right branches list.


klikh:
Thanks for the contribution! But this has already been fixed in 0e2571c4d200deb31584fa6a1b77235b0f6aef9c

Unfortunately, it seems that this repository is not being updated for 3 days, that's why you didn't see the commit. We'll check ASAP.

PR #800, opened 2018-06-04
IDEA-177512 added ExternalSystem TOPIC to notify when import has fini…

zdary:
added ExternalSystem TOPIC to notify when import has finished

### DIFF
--- a/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/project/ExternalSystemListener.java
+++ b/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/project/ExternalSystemListener.java
@@ -1,0 +1,15 @@
+// Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package com.intellij.openapi.externalSystem.service.project;
+
+import com.intellij.openapi.externalSystem.model.project.ProjectData;
+import com.intellij.openapi.module.Module;
+import com.intellij.util.messages.Topic;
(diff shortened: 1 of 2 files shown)


vladsoroka:
Hi Martin, sorry for the late reply on your issue.
Have you seen `ProjectDataImportListener#onImportFinished` api?
Could you check if it applicable for you?

zdary:
Hi @vladsoroka 
no, I haven't noticed that api. Although it doesn't give me the new module, it seems to be enough to resolve my problem. Thank you very much.

Martin

vladsoroka:
Also you can look on the `com.intellij.openapi.externalSystem.service.project.manage.ProjectDataService` extension point and extended import lifecycle methods in `AbstractProjectDataService`.
It should be enough to get all imported data and/or customize it.

If you need module information you can extends `AbstractProjectDataService` with the `ProjectKeys.MODULE` and use `IdeModifiableModelsProvider#findIdeModule(ModuleData module)` method:
```
@NotNull
  @Override
  public Key<ModuleData> getTargetDataKey() {
    return ProjectKeys.MODULE;
  }

@Override
  public void onSuccessImport(@NotNull Collection<DataNode<ModuleData>> imported,
                              @Nullable ProjectData projectData,
                              @NotNull Project project,
                              @NotNull IdeModelsProvider modelsProvider) {
   List<Module> importedModules = imported.stream()
                                          .map(node -> modelsProvider.findIdeModule(node.getData()))
                                          .collect(Collectors.toList());  }

```
This will work for initial import and all subsequent re-imports.

PR #803, opened 2018-06-05
Eliminating tail recursion in JBDiff#split

dahlstrom-g:
Previously, on large-ish files, JBDiff#split would sometimes cause
StackOverflowExceptions due to the use of tail recursion. Changed to
use stacks to maintain state during iteration, and eliminated the
recursion.
Also updating variable names to be more readable.

(cherry picked from commit 59aee5d)


trespasserw:
Thanks!

PR #805, opened 2018-06-07
Delete to the first newline after indexToDeleteTo

juancnuno:
Instead of the first one before. Doing that can still leave partial log
messages which break Android Studio.


juancnuno:
Hi @akozlova,

Remember me? I think we're gonna have to resize to the first newline *after* indexToDeleteTo instead of before because we're still getting uncaught java.time.format.DateTimeFormatter exceptions from partial dates left in Android Studio logcat messages. I added unit tests this time and LogConsoleBaseTest.resizeBufferIndexToDeleteToIsInFirstMessage will fail if you resize to the newline before.

Happy to chat about this, thanks!

akozlova:
Thanks for the test! I've merged manually with small re-refactoring to avoid unnecessary changes (so the place where the change was made is evident)

PR #808, opened 2018-06-11
Make EditorWindow.INITIAL_INDEX_KEY public for plugins

krasa:
I don't want tabs jumping around when switching between files, so I need to clear it from userData before opening files.

For: https://github.com/krasa/EditorGroups


JB-Dmitry:
I've made the change in master.

PR #809, opened 2018-06-14
IDEA-189671: Refresh Editor Focus after closing split

lumie1337:


klikh:

lumie1337:
I have signed the contributor license agreement now.

Vassiliy-Kudryashov:
Thank you, but the fix you suggested is a bit tricky, I've fixed the problem in more obvious way (I believe).

lumie1337:
the reason i fixed it this way is because there are quite a few edge cases which just get fixed by "resyncing" the state. it would obviously be better to make sure that things are always kept in sync, but that would involve a much bigger change.

ignatov:
@lumie1337 thanks for the fix, the original issue has been fixed in master.

PR #810, opened 2018-06-15
Have common Icons for nodes in build view

srmurguia:
When there are multiple builds on the build view, each one of them get
its Icon from AbstractViewManager#getIcon, but the icons used are not
consistent with the ones used in ExecutionNode.

This change causes getIcon to use a new function (getEventResultIcon)
that shares the same icons as ExecutionNode#getCurrentIcon. Also, the
icons are defined using constants to cause any change on them to be used
in both functions.


vladsoroka:
Thank you for the contribution! I've merged the change using the following commit https://github.com/JetBrains/intellij-community/commit/23b13bab5ac4f94f060c6bec8016ff317ce7fac8

PR #812, opened 2018-06-16
Create project level notification configuration implementation

brian-mcnamara:
Hook up non managed pom notification configuration to project level

### DIFF
--- a/platform/platform-impl/src/com/intellij/notification/impl/AbstractNotificationsConfigurationImpl.java
+++ b/platform/platform-impl/src/com/intellij/notification/impl/AbstractNotificationsConfigurationImpl.java
@@ -1,0 +1,122 @@
+// Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package com.intellij.notification.impl;
+
+import com.intellij.notification.NotificationDisplayType;
+import com.intellij.notification.NotificationGroup;
+import com.intellij.notification.NotificationsConfiguration;
+import com.intellij.openapi.components.PersistentStateComponent;
+import com.intellij.openapi.components.State;
+import com.intellij.openapi.components.Storage;
+import com.intellij.openapi.diagnostic.Logger;
+import gnu.trove.THashMap;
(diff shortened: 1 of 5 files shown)


brian-mcnamara:
Our environment at Salesforce requires this change. We have a specialized Maven project that does not use the IntelliJ maven plugin. We have our own code generator to generate the IntelliJ project structure so we do not need to import the project. We would like to disable to import notification at the project level to prevent developers from accidentally importing the project.

Happy to talk more about our use case more outside of github comments.

ignatov:
@chashnikov, please reassign the PR if needed.

chashnikov:
Sorry for the late reply, we somehow missed your pull request.

I have some questions about your implementation, but I firstly want to understand your use case. How does your generator work? Do you generate .iml file and .idea directory in it? Why did you decide to implement your own generator instead of creating a plugin for IntelliJ which extends functionality of IntelliJ Maven plugin to properly support your projects?

brian-mcnamara:
Hey @chashnikov ,

No worries for the delay. 
I agree, a plugin would be best, and that is what I have in mind down the line. The reason I went with a maven plugin to generate the .iml and modules.xml files (everything in .idea is checked into source control except for module.xml and .name) is there is a bunch of code already implemented by our build team that handles setting up the project structure for developers. Creating a maven plugin inherits all of this logic and allows the plugin to use mostly native maven plugin APIs. This was the easiest path forward for me.

To be honest, there is currently no official team working on IntelliJ support here, I have been spending my free time to improve the functionality to support IntelliJ on our build. We currently have roughly 70% Eclipse users (Which is maintained by a team) and roughly 30% IntelliJ users (Its been slowly going up as ease of development has been improved).

Like I mentioned, a plugin is ultimately the best solution. The team working on Eclipse integration has spent a lot of time improving the M2E plugin to work with our maven environment. This work would be more or less the same work I would need to do for the IntelliJ maven plugin. Its just a matter of me finding free time to sink into this project.

Like I mentioned earlier, if you are interested in more details of our maven environment, shoot me an email at brian.mcnamara@salesforce.com

Thanks for following up!

PR #813, opened 2018-06-19
IDEA-192953 Fix replace switch with if on primitive wrapper uses equals

goto1134:


ignatov:
cc @BasLeijdekkers

BasLeijdekkers:
Thanks for you contribution! Does this fix a YouTrack issue?

goto1134:
Yes, it does. 
There are tests for replacement with Character and Integer classes.

BasLeijdekkers:
Sorry, I missed the YouTrack ID here. Applied manually to Master, thanks!

PR #814, opened 2018-06-21
Allow alternative resolve in CombinedTestLocator.

juan-chen:
Android Studio runs gradle builds in dumb mode. This change allows
Android Studio to resolve test case locations even in dumb mode, if
alternative resolve is enabled.


akozlova:
CombinedTestLocator is dumb aware, if you would mark android's test locator as dumb aware, then test location should be resolved in dumb mode as well. (How you would do that, with help of slow alternative resolve or based on some fast checks - that's up to you). 
Would this work for you?

juan-chen:
Thanks for the suggestion Anna.

In Android Studio, CombinedTestLocator wraps JavaTestLocator, not an Android-specific test locator. JavaTestLocator is not dumb aware, and thus the getLocation method (without my proposed change) returns an empty list.

akozlova:
Do you use `com.android.tools.idea.testartifacts.junit.AndroidJUnitConsoleProperties`? Or are the pure junit tests without anything added by android?

akozlova:
I guess the pull request is outdated now as I don't get any answer since Jun

PR #815, opened 2018-06-25
Refactor ExportAction

ekudel:
- Split ExportAction into 2 actions and add them into "InspectionToolWindow.ExportPopup" action group to allow plugins to add more actions there.
- Add "InspectionToolWindow.Toolbar" action group to allow plugins to add actions into toolbar of "Inspection Results".

### DIFF
--- a/platform/lang-impl/src/com/intellij/codeInspection/ex/GlobalInspectionContextImpl.java
+++ b/platform/lang-impl/src/com/intellij/codeInspection/ex/GlobalInspectionContextImpl.java
@@ -20,7 +20,7 @@ import com.intellij.codeInspection.ui.DefaultInspectionToolPresentation;
 import com.intellij.codeInspection.ui.InspectionResultsView;
 import com.intellij.codeInspection.ui.InspectionToolPresentation;
 import com.intellij.codeInspection.ui.InspectionTreeState;
-import com.intellij.codeInspection.ui.actions.ExportHTMLAction;
+import com.intellij.codeInspection.ui.actions.ExportToXMLAction;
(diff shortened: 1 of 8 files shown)


klikh:

ekudel:
@klikh Ok, I've done it.

zdary:
@ekudel there are simple merge conflicts now

klikh:
@ekudel Please rebase the PR on top of current master.

@akozlova @batya239 please take a look at the PR.

ekudel:
@klikh Done

batya239:
@ekudel Now it's possible to add any action to "InspectionToolWindow.ExportPopup". I don't think it should be a kind of public API. I believe export html/xml actions should be unified and single abstract class should be extracted to export results to arbitrary given format

ekudel:
@batya239 Well, now the only common part of ExportToHTMLAction and ExportToXMLAction is showing "Export" dialog. However this dialog is actually HTML-specific (for example it contains "open in browser" checkbox), and for our custom action we don't need it, so I hardly understand what that abstract class should contain. By the way we want to provide a custom action for exporting to Hybris-compatible XLSX format. However there could be other ways to export inspection results to XLSX provided by the platform or another plugin, so one-to-one relationship between formats and handlers wouldn't seem to work here.

akozlova:
If needed, I'd prefer the approach which is done for exporting tests
image

yopox:
`InspectionResultsExportActionProvider` was added in 2022, it makes it possible to add new export actions for inspection results. This review is now obsolete.

PR #816, opened 2018-06-26
Resolve dependency sources in all repositories.

safetytrick:
As reported here: https://discuss.gradle.org/t/idea-plugin-isnt-downloading-sources-when-mavenlocal-is-used/2160 library sources are not downloaded for certain gradle configurations. 

Download source fails when the first maven repository configured doesn't already contain the artifact requested, a common case of this issue is use of mavenLocal which likely does not already have sources present.


ignatov:
cc @vladsoroka

vladsoroka:
Thanks for the contribution! However the change doesn't work for me locally. Also the check for empty files of the configuration will not work if there are transitive dependencies.
I've added the fix of the issue using the following commit https://github.com/JetBrains/intellij-community/commit/e7d2931e401dd35e1e030dfffcbbf679c8b42960
Please, let us know if it will not work for you.

PR #819, opened 2018-06-30
Print useful debugging info when compilation fails

dahlstrom-g:
@chashnikov?

### DIFF
--- a/platform/build-scripts/groovy/org/jetbrains/intellij/build/impl/JpsCompilationRunner.groovy
+++ b/platform/build-scripts/groovy/org/jetbrains/intellij/build/impl/JpsCompilationRunner.groovy
@@ -194,7 +194,7 @@ class JpsCompilationRunner {
         Standalone.runBuild(loader, compilationData.dataStorageRoot, messageHandler, scopes, false)
       }
       catch (Throwable e) {
-        context.messages.error("Compilation failed unexpectedly", e)
+        e.printStackTrace();
       }
     }
     if (!messageHandler.errorMessagesByCompiler.isEmpty()) {


chashnikov:
Firstly if compilation fails because of errors in source code, `Standalone.runBuild` won't throw an exception, the errors will be reported to `org.jetbrains.intellij.build.impl.JpsCompilationRunner.AntMessageHandler` and printed via `org.jetbrains.intellij.build.BuildMessages::compilationErrors`.
Secondly, even if `Standalone.runBuild` fails with exception (which means that it was an internal error in the compiler implementation), `context.messages.error` will throw `BuildException` which will stop the build, and its cause will be printed to the output.

So why do you need this?

dahlstrom-g:
Without this change I get inscrutable failures:
```
     [java] Compilation
     [java] 
     [java] BUILD FAILED
     [java] build/gant.xml:50: Compilation failed with exception: Compilation: Compilation failed unexpectedly
```

This change makes the output more informative:
```
     [java] Compilation
     [java]      [gant] java.lang.NoClassDefFoundError: Could not initialize class org.jetbrains.jps.incremental.BuilderRegistry$Holder
     [java]      [gant]         at org.jetbrains.jps.incremental.BuilderRegistry.getInstance(BuilderRegistry.java:42)
     [java]      [gant]         at org.jetbrains.jps.cmdline.BuildRunner.runBuild(BuildRunner.java:133)
     [java]      [gant]         at org.jetbrains.jps.build.Standalone.runBuild(Standalone.java:206)
     [java] […]
     [java] BUILD FAILED
     [java] build/gant.xml:50: Build provided modules list: Cannot run application starter [listBundledPlugins, out/studio/artifacts/AI-builtinModules.json], 'java-runtime' module isn't compiled (out/studio/classes/production/intellij.java.rt doesn't exist)
```

chashnikov:
Ok, I see, Ant just catches `BuildException` in `org.apache.tools.ant.Main#startAnt` method and don't print its stracktrace by default. I'll think how to fix this properly.

chashnikov:
I've [implemented this](https://github.com/JetBrains/intellij-community/commit/a11ef0903e54a1206f5dfe23834405f0eaba565c) in a little different way, by printing stacktrace via `BuildMessageLogger`. This will work better then build is running under TeamCity. Also it makes sense to stop build by throwing exception, not just print stacktrace.

dahlstrom-g:
Thank you!

PR #821, opened 2018-07-04
Fix for tabs alignment

c5inco:
Open Youtrack issue: https://youtrack.jetbrains.com/issue/IDEA-195031


ignatov:
@Vassiliy-Kudryashov please take a look.

ignatov:
Already fixed by @kkmic.

PR #823, opened 2018-07-08
Add "delete unversioned" action shortcut in commit changes dialog

SomeoneToIgnore:
This PR fixes the https://youtrack.jetbrains.com/issue/IDEA-160857 issue.

### DIFF
--- a/platform/platform-api/src/com/intellij/openapi/actionSystem/IdeActions.java
+++ b/platform/platform-api/src/com/intellij/openapi/actionSystem/IdeActions.java
@@ -312,6 +312,7 @@ public interface IdeActions {
 
   String CONSOLE_CLEAR_ALL = "ConsoleView.ClearAll";
   String MOVE_TO_ANOTHER_CHANGE_LIST = "ChangesView.Move";
+  String DELETE_UNVERSIONED = "ChangesView.DeleteUnversioned";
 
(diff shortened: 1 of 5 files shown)


pchel-:
Actually, I think, it's not quite OK to add "DELETE" shortcut to the "ChangesView.DeleteUnversioned" action, because it is enabled "almost everywhere" and will highly intersect with common "$Delete" action (though the implementation is quite the same).

I'll make changes to use "$Delete" action instead of "ChangesView.DeleteUnversioned" (and so delete shortcut will also be available).

SomeoneToIgnore:
If I've understood you correctly, I need to edit `VcsActions.xml` only.

If I do that, then I the delete icon disappears from the context menu.
Will it be ok?
no_icon

Otherwise it works on both Linux and macOs, thanks for the tip.

pchel-:
I've pushed several changes related to the topic - see 023f55d7abeada9e76528b5e3758779e1eed2840, 9bae28b30f6fe09dca3a1130de3f0971ec3b5bd5, 7627ae197bd128ef2255dd6a750903c9c586d4a4, c4f3bd954fe857701d299fa2029b1bbff7ed75e3.

Thanks!

SomeoneToIgnore:
Great, thanks a lot for the fix.

PR #824, opened 2018-07-09
Allow override implementation handlers to customize result

dahlstrom-g:
The OverrideImplementsAnnotationHandler extension point
allows plugins to specify names of annotations that
should be copied into overriding methods.

This CL adds an additional optional method on the interface:
"cleanup", which runs afterwards. This allows plugins to
do further post processing on the annotations. In particular,
this will let plugins *rename* annotations (which we're
using to for example rename the new @RecentlyNullable
and @RecentlyNonNull annotations that are specially recognized
by the Kotlin compiler as nullable annotations with
@Migrate, which means violations should be treated as only
warnings). However, the cleanup step can do other tasks too,
such as additing additional annotations not present in the
overridden method, or changing attribute values, etc.

### DIFF
--- a/java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementsAnnotationsHandler.java
+++ b/java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementsAnnotationsHandler.java
@@ -44,6 +44,10 @@ public interface OverrideImplementsAnnotationsHandler {
     return ArrayUtil.EMPTY_STRING_ARRAY;
   }
 
+  /** Perform post processing on the annotations, such as deleting or renaming or otherwise updating annotations in the override */
+  default void cleanup(PsiModifierListOwner source, @Nullable PsiElement targetClass, PsiModifierListOwner target) {
+  }
+
   static void repeatAnnotationsFromSource(PsiModifierListOwner source, @Nullable PsiElement targetClass, PsiModifierListOwner target) {
     Module module = ModuleUtilCore.findModuleForPsiElement(targetClass != null ? targetClass : target);
     GlobalSearchScope moduleScope = module != null ? GlobalSearchScope.moduleWithDependenciesAndLibrariesScope(module) : null;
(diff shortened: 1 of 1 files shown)


dahlstrom-g:
@cdracm?

ignatov:
cc @akozlova

akozlova:
What do you think about extracting `transferToTarget` method. Should work well for your case and it's more clear what to expect in the implementations
Index: java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementsAnnotationsHandler.java
IDEA additional info:
<+>UTF-8
--- java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementsAnnotationsHandler.java	(date 1532345703000)
+++ java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementsAnnotationsHandler.java	(date 1532346400156)
@@ -56,14 +56,17 @@
 
         int flags = CHECK_EXTERNAL | CHECK_TYPE;
         if (AnnotationUtil.isAnnotated(source, annotation, flags) && !AnnotationUtil.isAnnotated(target, annotation, flags)) {
-          PsiModifierList modifierList = target.getModifierList();
-          assert modifierList != null : target;
-          PsiAnnotation srcAnnotation = AnnotationUtil.findAnnotation(source, annotation);
-          PsiNameValuePair[] valuePairs = srcAnnotation != null ? srcAnnotation.getParameterList().getAttributes() 
-                                                                : PsiNameValuePair.EMPTY_ARRAY;
-          AddAnnotationPsiFix.addPhysicalAnnotation(annotation, valuePairs, modifierList);
-        }
-      }
-    }
+          each.transferToTarget(annotation, source, target);
+        }
+      }
+    }
+  }
+
+  default void transferToTarget(String annotation, PsiModifierListOwner source, PsiModifierListOwner target) {
+    PsiModifierList modifierList = target.getModifierList();
+    assert modifierList != null : target;
+    PsiAnnotation srcAnnotation = AnnotationUtil.findAnnotation(source, annotation);
+    PsiNameValuePair[] valuePairs = srcAnnotation != null ? srcAnnotation.getParameterList().getAttributes() : PsiNameValuePair.EMPTY_ARRAY;
+    AddAnnotationPsiFix.addPhysicalAnnotation(annotation, valuePairs, modifierList);
   }
 }
\ No newline at end of file

tnorbye:
Sorry, I missed this earlier. Yes, that LGTM!

PR #825, opened 2018-07-09
Add RecentlyNullable and RecentlyNonNull to nullable list

dahlstrom-g:
These two new magic names are also known to the
Kotlin compiler as nullness annotations representing
recently migrated nulls, which will be treated as
warnings rather than errors when violated.


dahlstrom-g:
@peter?

donnerpeter:
Cherry-picked manually, thanks!

PR #828, opened 2018-07-14
Add a Way To Specify Class Names to Not Import Inner Classes For

alonalbert:
Like Builder, Factory etc

https://youtrack.jetbrains.com/issue/IDEA-135092

Submitted as @google.com


akozlova:
Could you please fix UI issues:
1. If 'insert imports on inner classes' option is off, the panel with specific inner classes should be disabled. 
2. When I add an inner class to the list, I get the Go To Class chooser. This makes me think that only that class (by QName) would be ignored, though it's not the case. I guess, simple text field is enough.
Thanks

alonalbert:
>
>
>    1. If 'insert imports on inner classes' option is off, the panel with
>    specific inner classes should be disabled.
>
> This is trivial. Will fix when I get a chance.


>
>    1. When I add an inner class to the list, I get the Go To Class
>    chooser. This makes me think that only that class (by QName) would be
>    ignored, though it's not the case. I guess, simple text field is enough.
>    Thanks
>
> Hmm, I really like that I can use the picker. At first, I intended to
simple text field as you suggest but this is so much nicer. How about I
make the text label above it have a verbose explanation to what it actually
being filtered.
Something like:
"Don't insert imports for inner classes if their names are listed bellow.
Note that while the picker will show actual inner classes, only the class
names are considered for this setting."

I think IntelliJ users are clever enough to follow this explanation. It
would be a shame to not show the picker.

akozlova:
What is the picker for? People don't read the labels when they see the UI which they are used to (and the picker is used normally to choose a fully qualified class name).

alonalbert:
Is there a picker that handles plain text entries or will I have to write
one?

akozlova:
I'd suggest to stay with simple text field. Or please explain what you like in picker.

alonalbert:
Done.
I replaced the Class Picker table with a text only table.

akozlova:
1) Unfortunately the panel is still enabled when I just open the panel.
2) On '+' I get a <enter class name> entry which I have to delete explicitly. It's not common to insert explicit place holders in IDEA. Empty item should be added instead and when user doesn't fill it out, then it should be just ignored on save

akozlova:
Moreover, it's probably better to have one text field "exclude inner classes matching *Builder", semicolon separated, like for file types:
image
What do you think?

alonalbert:
> Unfortunately the panel is still enabled when I just open the panel.

Fix coming
> On '+' I get a entry which I have to delete explicitly

I was working off an example I found in Run Configurations/Applet/Applet Parameters that inserts a dummy value. 
[AppletConfigurable](https://github.com/JetBrains/intellij-community/blob/2cb686f8103dec27da3ba63181f5c2178240b4d2/java/execution/impl/src/com/intellij/execution/applet/AppletConfigurable.java#L160)

I later found another example that does it with a blank value and I already updated the pull request. I still need to handle ignore-empty-values-on-save though.

> Moreover, it's probably better to have one text field "

I was trying very hard to avoid this pattern. I don't like it very much. It requires the user to know what the delimiter is and makes it harder to remove entries. More chance for user error.
Personally, I like my original solution with a class picker best. It gives he user a good feeling that he didn't misspell the class name. It also sometimes highlights classes with similar names for example, while entering Builder, I might also decide to add BuilderFactory etc'
But since you feel strongly against that, I think a table is better than a text field.

akozlova:
I don't think that separator is a problem, see
image
Still now it's not evident that the table contains short names.

alonalbert:
> I don't think that separator is a problem, see

Yes, I know there's a bunch of places that use this pattern. I still don't like it. There's other places that could use a text field but use a table instead:

file-types

I really prefer a table.

> Still now it's not evident that the table contains short names.

Not sure what you mean. Is the "Do not import inner classes for" title not clear enough? What would you suggest I call it?

akozlova:
"Exclude inner classes by short name". I'd also align the table to the right so it's clear that it depends on the checkbox above

alonalbert:
done & done

akozlova:
Thanks. Merged manually with some small cosmetics changes (in code)

alonalbert:
Cool, thanks for the review.

plumpy:
Thanks so much for doing this. This is a highly requested feature inside Google, it comes up on internal mailing lists every few months :)

PR #829, opened 2018-07-14
Auto-hide editor scrollbars

nvartolomei:
Since `editor.xcode.like.scrollbar` seems broken, I decided to make few changes to make it at least somewhat usable.

Demo https://youtu.be/ro2A7EzjrB0 with `editor.xcode.like.scrollbar` enabled after change.

cc @ignatov (based on git blame)

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/editor/impl/EditorMarkupModelImpl.java
+++ b/platform/platform-impl/src/com/intellij/openapi/editor/impl/EditorMarkupModelImpl.java
@@ -17,7 +17,6 @@ import com.intellij.openapi.editor.actionSystem.DocCommandGroupId;
 import com.intellij.openapi.editor.colors.EditorFontType;
 import com.intellij.openapi.editor.ex.*;
 import com.intellij.openapi.editor.ex.util.EditorUIUtil;
-import com.intellij.openapi.editor.ex.util.EditorUtil;
 import com.intellij.openapi.editor.markup.ErrorStripeRenderer;
 import com.intellij.openapi.editor.markup.RangeHighlighter;
 import com.intellij.openapi.fileEditor.impl.EditorWindowHolder;
@@ -490,11 +489,6 @@ public class EditorMarkupModelImpl extends MarkupModelImpl implements EditorMark
     }
 
     @Override
-    protected boolean isMacScrollbarHiddenAndXcodeLikeScrollbar() {
-      return super.isMacScrollbarHiddenAndXcodeLikeScrollbar() && EditorUtil.isRealFileEditor(myEditor);
-    }
-
-    @Override
     public boolean alwaysShowTrack() {
       if (scrollbar.getOrientation() == Adjustable.VERTICAL) return !transparent();
       return super.alwaysShowTrack();
@@ -581,8 +575,7 @@ public class EditorMarkupModelImpl extends MarkupModelImpl implements EditorMark
 
     @Override
     protected boolean alwaysPaintThumb() {
-      if (scrollbar.getOrientation() == Adjustable.VERTICAL) return !(xcodeLikeScrollbar() && EditorUtil.isRealFileEditor(myEditor));
-      return super.alwaysPaintThumb();
+      return true;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


ignatov:
@nvartolomei thanks, will check it out.

SergeyMalenkov:
Thank you for the fix. I'm testing it locally, but I do not want to push it into master as is.
First, I think many people will ask to return old behaviour and we should provide such a way.
Second, I need to discuss suggested solution with our designer who is on vacation now.
Stay tuned!

nvartolomei:
@SergeyMalenkov old behaviour is? For me `editor.xcode.like.scrollbar` turns scrollbar into a black rectangle, is this intended?

screen shot 2018-07-16 at 22 47 23

To me that looked like a bug/abandoned feature, so I came up with this patch which hides editor scrollbar handle (thumb) only when this registry entry is set to true. When `editor.xcode.like.scrollbar` is false, this patch should not change anything, if it does than it is my fault.

ignatov:
@SergeyMalenkov I agree with @nvartolomei, the option is experimental and we should fix the original behavior, there's nothing to discuss with UI team.

platform/platform-impl/src/com/intellij/openapi/editor/impl/EditorMarkupModelImpl.java
```diff
@@ -581,8 +575,7 @@ protected boolean isDark() {
 
     @Override
     protected boolean alwaysPaintThumb() {
-      if (scrollbar.getOrientation() == Adjustable.VERTICAL) return !(xcodeLikeScrollbar() && EditorUtil.isRealFileEditor(myEditor));
-      return super.alwaysPaintThumb();
+      return true;
```
SergeyMalenkov: It is suspicious. Why it returns true, that mean always paint thumb?

SergeyMalenkov:
@nvartolomei I mean 2 scenarios for scroll thumb in editor:
1. it should be always visible (current behaviour)
2. it should be hidden after awhile (editor.xcode.like.scrollbar=true)
Suggested patch removes first behaviour and adds the second one.
It does not provide an ability to switch to the needed behaviour.
But it should fix painting when editor.xcode.like.scrollbar=true.

@ignatov We should discuss which behaviour should be used by default, at least.
And I think we are going to leave the current behaviour, because no one use editor.xcode.like.scrollbar.

nvartolomei:
@SergeyMalenkov I think you misinterpreted the code a bit. Re-compile with this patch applied and test both cases (`editor.xcode.like.scrollbar=true/false`) and you'll see that default behaviour is not changed, only behaviour under `editor.xcode.like.scrollbar=true` is changed.

platform/platform-impl/src/com/intellij/openapi/editor/impl/EditorMarkupModelImpl.java
nvartolomei: This method is inside `MyErrorPanel`, to have event listeners bound, this method should return true.

From what I understand it is only used in code editor and only in vertical context, so `scrollbar.getOrientation() == Adjustable.VERTICAL` always evaluates to true, so it always returns result of `return !(xcodeLikeScrollbar() && EditorUtil.isRealFileEditor(myEditor));`.

Under default behaviour `xcodeLikeScrollbar()` evaluates to false, condition is short-circuited and negated, so it returns true.

Under `editor.xcode.like.scrollbar=true` this would return false and hide `MyErrorPanel`, disabling mouse event handlers when thumb is hidden by logic from the parent class that implements OS X scrollbar hiding behaviour.

In the case of this patch, I decided to always render `MyErrorPanel` thumb (which is actually invisible) and have mouse event handlers active, you can see this in action in the demo video linked in PR description where I hover cursor on error stripes.

tl;dr: alwaysPaintThumb returns whether or not to render invisible thumb for error stripes, which under all conditions should be true

Is there a case when this is not desired?

SergeyMalenkov:
Sorry. Seems that patch was applied incorrectly first time. Or full rebuild helped...
I'll test a bit and then push these changes into master.

PR #832, opened 2018-07-18
Switch JavaHelp license type to match bundled version

dahlstrom-g:
License text is from:
https://github.com/javaee/javahelp/commit/3ca862d

We "Mavenized" this library in the recent merge, and we're now
using javax.help:javahelp:jar:2.0.02.


dahlstrom-g:
@yole?

PR #833, opened 2018-07-19
Shorten command line option "Classpath file": Support for Java 9 and …

ksaua:
…above

The classloader heirarchy has changed with Java 9. With the
modularization the bootstrap classloader has been split into two. One
for the core java classes, and a platform classloader which contains
miscellaneous classes.

The classloader created in the "Classpath file" uses a null parent to
signify that the parent should be the bootstrap classloader. This no
longer works with Java 9 and above. The correct way is to just get the
parent of the class itself. When running in Java 8 (and below) this will
be null which is the bootstrap classloader. When running Java 9 (and
above) this will be the platform classloader.

### DIFF
--- a/java/java-runtime/src/com/intellij/rt/execution/CommandLineWrapper.java
+++ b/java/java-runtime/src/com/intellij/rt/execution/CommandLineWrapper.java
@@ -184,7 +184,10 @@ public class CommandLineWrapper {
     }
 
     String mainClassName = args[startArgsIdx - 1];
-    ClassLoader loader = new URLClassLoader((URL[])classpathUrls.toArray(new URL[0]), null);
+    // Java 8 and below will get a null, which means the parent will be the bootstrap classloader
+    // Java 9 and above will get the platform classloader as defined in the new classloader hierarchy.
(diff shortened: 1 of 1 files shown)


ksaua:
This code will throw ClassNotFoundException when run with the "Classpath file" option because the java.sql.SQLException class is in Java 9 not found in the bootstrap classloader, but rather in the platform classloader

    public class Main {
      public static void main(String[] args) throws ClassNotFoundException {
        System.out.println(Class.forName("java.sql.SQLException"));
      }
    }

akozlova:
There is dedicated way to shorten classpath for java 9 and above. I doubt that our custom workarounds should be adjusted for the new java versions. Is there anything why you prefer special classpath to the argument file? I'd rather hide other options for java 9 and above.

ksaua:
If the option can be hidden for java 9 and above that's fine. I figured: If the option was available then it should at least work.

akozlova:
Done as part of https://youtrack.jetbrains.com/issue/IDEA-195937. Thanks!

PR #836, opened 2018-07-24
[vcs-log] fix typo in helper method

Walingar:
This typo produces NPE


klikh:
Nice catch, thanks. However, it appears that this method is unused, at least in our sources and the sources of known plugins. Do you need this method?

Walingar:
Yes, I really need!

AMPivovarov:
Thanks, pushed.

PR #837, opened 2018-07-25
Uniformize all color scheme hexadecimal colors

TwinProduction:
With the `colorSchemes` folder left untouched for the past 6 months and my obsession with consistency, I could no longer deal with the mix of uppercase and lowercase hexadecimal color values.

Thus, I've replaced all lowercase hexadecimal colors to uppercase.

Since the color schemes are very unlikely to be modified anyways, I figured it might as well be properly formatted.

### DIFF
--- a/colorSchemes/src/colorSchemes/WarmNeon.xml
+++ b/colorSchemes/src/colorSchemes/WarmNeon.xml
@@ -1,176 +1,176 @@
 <scheme name="WarmNeon" version="142" parent_scheme="Darcula">
   <colors>
     <option name="ADDED_LINES_COLOR" value="395439" />
-    <option name="ANNOTATIONS_COLOR" value="8178af" />
-    <option name="CARET_COLOR" value="ff00" />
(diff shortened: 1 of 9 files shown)


denofevil:

denofevil:
@dyadix I'd suppose that we need to update https://github.com/JetBrains/colorSchemeTool/blob/master/colorSchemeTool.py#L42 accordingly

TwinProduction:
@denofevil I already signed the contributor agreement, and good point. I'll do that when I get home.

TwinProduction:
@dyadix @denofevil done. 

Is there anything else that should be updated?

dyadix:
@denofevil I agree, we should update the conversion tool instead and load the updated schemes.

TwinProduction:
@dyadix See https://github.com/JetBrains/colorSchemeTool/pull/22, it does just that.

denofevil:
Merged as 8c12470, thank you!

PR #838, opened 2018-07-26
Move GradleInspectionBundle.properties into resource directory

afwang:
This resource file was causing an error for Android Studio tests due to
the unconventional nature of the directory structure. The resource file
should be found underneath a directory tree that appears as
org/jetbrains/plugins/gradle/codeInspection.

Additionally, resource files should be kept separate from source files.


afwang:
Actually, disregard this. I think if we're going to be moving a resource file outside of the source directory without the package prefix, every file in this module should sit in a directory structure without the package prefix. This change should be considered incomplete.

PR #839, opened 2018-07-27
Change resolve class scope to include libraries

brian-mcnamara:
Wanted to open a conversation about this. The build at salesforce uses libraries to cut down on the amount of compiling that is done on a developers machine. What this means is a developer could have references of classes in their workspace that are part of libraries. 

This change is to allow us to search for implementors and usages inside of these libraries. 

My concern, something like this is probably only found in very small number of projects such as ours. I do worry that this adds unneeded performance hits for the 99% so wanted to get opinions on whether this should be gated somehow.

This is also a work in progress. I have some comments in the code as to whether I implemented it correctly. I also have not done much testing (its getting late) but I will get around to it.

### DIFF
--- a/platform/indexing-impl/src/com/intellij/openapi/module/impl/scopes/ModuleWithDependentsAndLibrariesScope.java
+++ b/platform/indexing-impl/src/com/intellij/openapi/module/impl/scopes/ModuleWithDependentsAndLibrariesScope.java
@@ -3,6 +3,7 @@ package com.intellij.openapi.module.impl.scopes;
 
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.roots.TestSourcesFilter;
+import com.intellij.openapi.roots.impl.DirectoryInfo;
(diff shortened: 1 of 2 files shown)


brian-mcnamara:
@chashnikov No worries, like I mentioned in my OP, I was hoping to use this PR as a means to open a conversation (I prefer to talk around code than over bug reports). I realize our build is unique where the dependency graph is turned upside down. I was hoping we could talk about it over the PR but if you prefer to take it to youtrack I can open an item. My hope was we could have this feature behind a setting or some sort of logic branch to make sure we could benefit from this implementation but not impose the potential performance gap on others. 

Anyway, I did this work as a fun side project. I had fun learning some code and navigating the IntelliJ sources. Again, if you prefer the conversation to be taken to youtrack, Ill open a request there.

brian-mcnamara:
Actually I managed to stick this in a plugin. Should work for us.

PR #840, opened 2018-07-27
Prevent exception wrapping to UndeclaredThrowableException

dmitry-zhuravlev:
If the corresponding external system api implementation throw some exception (e.g. ExternalSystemException) this exception will be wrapped to UndeclaredThrowableException. This happens because method method.invoke throw InvocationTargetException which is checked exception and such exception not a part of the corresponding proxied interface signature (e.g. RemoteExternalSystemProjectResolver.resolveProjectInfo method)


vladsoroka:
Thanks for the contribution! The change has been merged using the following commit https://github.com/JetBrains/intellij-community/commit/090de8bcd1684d1408f29cea5a1b2f5b54b47420

PR #841, opened 2018-07-27
Use conventional directory structure for intellij.gradle.java

afwang:
This module followed a non-conventional directory structure.  Java
packages usually correspond with a directory. This module used a package
prefix which caused problems in downstream projects.

This change moves all source files to follow a more conventional
directory structure, and removes the usage of the package prefix from
the module. Moreover, the GradleInspectionBundle.properties file is
moved to the resources directory.


afwang:
Sure! The Android Studio project was unable to run some of its UI tests with the Bazel build system. The custom rule that was building IML modules handled package prefixes for source files, but prefixes were not applied to resource files. For example, the GradleInspectionBundle.properties file was packaged as codeInspection/GradleInspectionBundle.properties rather than org/jetbrains/plugins/gradle/GradleInspectionBundle.properties.

afwang:
@vladsoroka any input?

afwang:
@klikh Polite ping, just to make sure that this PR doesn't end up getting ignored.

vladsoroka:
@afwang 
> The custom rule that was building IML modules handled package prefixes for source files, but prefixes were not applied to resource files. 

So, the only GradleInspectionBundle.properties file movement is needed for you.
I've moved the file using the following commit https://github.com/JetBrains/intellij-community/commit/c51505f88a02d868b410aba718de3cdc64a74f77

afwang:
Thanks!

PR #842, opened 2018-07-27
Fix format-string mismatch

dahlstrom-g:
… introduced by commit 673eddd.

See http://errorprone.info/bugpattern/FormatString#missingformatargument.


dahlstrom-g:
@klikh?

klikh:
TBH, I don't see a serious issue here, since extra arguments are simply ignored and in this case it is obvious that they are not missing by accident, but I've applied the patch since it improves readability a bit. Thanks!

PR #843, opened 2018-07-28
fix extra semicolon in decompiled empty blocks IDEA-196314

temp1011:
This should fix the mentioned issue (and also other cases of empty blocks that I didn't consider). I have changed the tests so that they all pass, if there are problems that the tests do not cover I am happy to add more.


gorrus:
There's a semicolon param in the method specifically for this case, why don't change it to false?

temp1011:
I initially just changed `jmpWrapper` as it fixed all cases and I couldn't find anything that used it to provide a semicolon correctly. However, some cases with if statements aren't covered by it and it's not a great fix. Hopefully the output formatting is OK.

gorrus:
Merged into one commit and integrated, thanks for contribution

PR #844, opened 2018-07-31
Optimize plugin version checking when update intellij platform version

shlxue:
Merge remote-tracking branch 'JetBrains/master', and update #228.

Checking plugin version by plugin version and since build

@alexeypegov @ignatov


shlxue:
Sorry, I removed it.
Now: the plugin version is found when idea platform upgraded, but downgrade don't.

@klikh

PR #845, opened 2018-08-02
Decompilation of inverted floating point comparisons (IDEA-196302)

malte0811:
[Bug report](https://youtrack.jetbrains.net/issue/IDEA-196302)
This is a copy of my original PR (as mentioned in the issue), see https://github.com/MinecraftForge/ForgeFlower/pull/11.

`a<b` and `!(a>=b)` have different results if either `a` or `b` is NaN/"Not a number". In that case the first expression will be false, the second will be true. The fix is fairly straightforward, I replace `a<b` with `!(a>=b)` (and the same for other comparisons) when the statement should be true for NaN and prevent simplification of negated floating point comparisons. The test class contains tests for non-inverted and inverted floating point comparisons as well as non-inverted and inverted integer comparisons (the last case can safely be simplified).


LexManos:
Throwing my support for this, for what it is worth.
I've spent the last several days verifying the changes and how they effect the resulting code.
It all seems to be consistent with the functionality of the pre-decompiled code.

klikh:

malte0811:
I did sign the CLA.

klikh:
For some reason, I can't find the CLA in our records. Could you please sign or resend it once again?

malte0811:
I signed it again, but I guess I'm doing something wrong if it didn't work the first time. The instructions aren't entirely clear (and I haven't used DocuSign before): Do I need to manually send the document after signing it on the DocuSign website or does DocuSign do that automatically?

klikh:
Now we've received the CLA, everything is fine now. Thanks.

ccarpenter04:
I'd be interested in seeing this edge case resolved.

trespasserw:
Sorry for the late reply.

I just merged the changes. Thanks!

PR #849, opened 2018-09-01
IDEA-166142 IDEA-198152 find and return also jar:/// files…

zdary:
find and return also jar:/// files if they are inside contentRootOf


SergeyMalenkov:
We found another way to fix the issue.
Thank you a lot for your help!

PR #851, opened 2018-09-02
build scripts: increase Xmx to fix OOM

guoci:
For PyCharm build, similar to https://github.com/JetBrains/intellij-community/commit/4d31c825d7b49157d3b8acd60ab4d4b1d31c7ee6.


batya239:
Hi! Sorry for the very late response. We already reworked our build scripts and, unfortunately, given PR is outdated. I close it for now since we don't use ant anymore. Have a good day!

PR #854, opened 2018-09-11
[Decompiler] Fix short name imports shadowed by super inner classes (IDEA-196315)

JDLogic:
Currently, decompiling the modified test class will result in the short names `Builder` and `Entry` incorrectly being used as both these imports will be shadowed. This PR adds a check to the import collector to see if a short name will be shadowed a super inner class and adjusts the import accordingly.


JDLogic:
I just thought of a way to slightly improve this. I'll update the PR this weekend when I have time.

JDLogic:
Cleaned up this PR. Should be much better.

plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/main/collectors/ImportCollector.java
```diff
@@ -158,4 +161,24 @@ public int writeImports(TextBuffer buffer) {
       .map(ent -> ent.getValue() + "." + ent.getKey())
       .collect(Collectors.toList());
   }
+
+  private boolean isSuperInnerClass(StructClass cl, String className) {
```
gorrus: I'm a little bit worried about performance here, wouldn't it be better to populate all classes names set in the ImportCollector constructor like we have already for fields (see org.jetbrains.java.decompiler.main.collectors.ImportCollector#setFieldNames)?

JDLogic: I just got a chance to look this over, and you are correct. I'll update this PR soon.

JDLogic:
Updated with the requested changes made.

gorrus:
merged into master, thanks for contribution!

PR #856, opened 2018-09-12
Unintended autocomplete when typing B or P in Kotlin

goodwinnk:
Fix for https://youtrack.jetbrains.com/issue/KT-8474.

### DIFF
--- a/resources/src/liveTemplates/surround.xml
+++ b/resources/src/liveTemplates/surround.xml
@@ -14,6 +14,7 @@
       <option name="GROOVY" value="false" />
       <option name="GROOVY_STATEMENT" value="false" />
       <option name="JSON" value="false" />
+      <option name="KOTLIN" value="false" />
       <option name="OTHER" value="true" />
     </context>
   </template>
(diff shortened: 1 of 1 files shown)


donnerpeter:
I'd say:
1. The platform shouldn't have Kotlin reference here (or should have references to all other possible languages for symmetry, which is infeasible)
2. Surround (and maybe others) templates shouldn't be shown in autopopup completion list and selected by keys they're not assigned to. @zolotov ?

goodwinnk:
1. I completely agree there shouldn't be references to particular languages but think that under current architecture this fix is a valid cure for a long-standing problem.
2. Also true, the whole feature looks questionable to me.

zolotov:
2. not sure about this. Surround templates are not any different to regular templates. The fact that they have a `$SELECTION$` variable is just a nice bonus.

I'd say we need to rethink `OTHER` context type. Why `Kotlin` context matches `Other` context type?

zolotov:
There was a suggestion by Peter do not show templates in auto popup if there is nothing but them. I'm still not sure about it but I'll try to implement it

PR #859, opened 2018-09-14
IDEA-198933: Fix painting of splash screen progress bar

rpaquay:
* On Windows HiDPI mode (JRE managed) with a fractional scaling
  factor (e.g. 1.5), the progress tail shows as garbled pixels instead
  of a semi transparent icon.

* There were 2 HiDPI specific issue when loading the tail icon

  * The icon should allow fractional scaling, otherwise it is just scaled
    too much (scaling of 2.0 instead of 1.5 for a 1.5 monitor scaling)

  * The icon image should be wrapped into a JBImageIcon, so that paintIcon
    method ends up calling into a HiDPI aware paint method.


forantar:
@rpaquay Thank you, pushed it: 2494a7d

PR #860, opened 2018-09-17
Add the ability to highlight java primitives separately from keywords.

mfarrugi:
Add the ability to highlight java primitives separately from keywords. This leaves the default highlighting unchanged (ie. the same as keywords).

image

The README.md doesn't cover running tests, so lmk if there's anything else I should do.

(CLA should be good from intellij-rust contributions)

### DIFF
--- a/java/java-analysis-impl/src/com/intellij/ide/highlighter/JavaFileHighlighter.java
+++ b/java/java-analysis-impl/src/com/intellij/ide/highlighter/JavaFileHighlighter.java
@@ -29,6 +29,7 @@ public class JavaFileHighlighter extends SyntaxHighlighterBase {
     ourMap2 = new HashMap<>();
 
     fillMap(ourMap1, ElementType.KEYWORD_BIT_SET, JavaHighlightingColors.KEYWORD);
+    fillMap(ourMap1, ElementType.PRIMITIVE_TYPE_BIT_SET, JavaHighlightingColors.PRIMITIVE);
     fillMap(ourMap1, ElementType.LITERAL_BIT_SET, JavaHighlightingColors.KEYWORD);
     fillMap(ourMap1, ElementType.OPERATION_BIT_SET, JavaHighlightingColors.OPERATION_SIGN);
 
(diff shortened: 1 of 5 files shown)


mfarrugi:
@dyadix does this look it will be merged eventually?

ignatov:
@akozlova please assist.

akozlova:
Why do you want primitives to be highlighted differently from other keywords? The place under "classes and interfaces" is unexpected and I doubt that the highlighting would be used by many users. 

Moreover, how would you highlight `var`? I guess, it would be inconsistent as in some cases it must be "primitive" and in others "non-primitive".

mfarrugi:
Personally I wanted `int` and other primitive types to be the same color as classes in my color scheme. I don't know where else in the menus it would go, and I'm not attached to this particular placement.

Primitives are the keywords that are actual type names, whereas `var` is a non-type keyword. If we really wanted, `var` and any/every other (type of) keyword could be separately colorable as well (eg. visibility modifiers).

akozlova:
I believe that `var` is a type-keyword in java 10+, https://openjdk.java.net/jeps/286

Frankly saying I don't want to add a feature for the sake of adding a feature. I think, it's a good candidate to be implemented as a plugin, e.g. see how soft keywords are processed `com.intellij.codeInsight.daemon.impl.JavaSoftKeywordHighlightingPassFactory`

mfarrugi:
The implementation slots in very cleanly in a few lines of configuration here, I imagine as a plugin it takes substantially more code and maintenance to implement. I view this in the same vein as being able to highlight class/anonymous class/interface all separately if one chooses, just a little bit more customization.

I don't mean anything pedantic by type-keyword, just that `int` and `long` are named types while `var` is not.

akozlova:
I would be happy to accept the PR with the new API which would make writing the plugin to highlight different keywords easy. 
I feel uncomfortable here because the next PR would ask to highlight `boolean` differently from `int` and `long` and then another one to separate `float` and `double`. This makes the settings to mess and complicates the configuration a lot. Thus I think it's better to achieve through the plugin. With good API I can imagine that the plugin could consist mostly from `plugin.xml` with registered keywords.

mfarrugi:
I'm not interested in making a plugin to avoid making a 7 line change.

PR #865, opened 2018-09-19
IDEA-197106 Fix race condition when opening a project

dahlstrom-g:
When a project is opened, the thread that is performing the openProject
method call (task within ProjectManagerImpl:loadAndOpenProject())
is being run in parallel to ToolWindowManagerImpl:updateToolWindowHeader,
which results in a modification to the myIdToInfo THashMap.

This simultaneous access and modification results in a
ConcurrentModificationException being thrown.


dahlstrom-g:
Unfortunately commit 1d40e12 failed for us; see comments there.

PR #866, opened 2018-09-20
Catch ClassCastException on GroupByTypeComparator

srmurguia:
When comparing ProjectViewNodes using type, the comparator calls to
getTypeSortKey, this function returns an instance of Comparable, but the
only warranty is that we can compare elements of the same type. In most
of the cases the value is a String but there are a couple of places that
it is not, possibly causing ClassCastException when comparing an String
to a different type. This is the cause of
[https://issuetracker.google.com/issues/113159120](https://issuetracker.google.com/issues/113159120).

This change ignores ClassCastException when comparing keys while sorting
by type.


SergeyMalenkov:
Thank you for your fix. I'll integrate it soon.

srmurguia:
Merged with commit a740f0d4

PR #867, opened 2018-09-20
Fix bug in ScrollColorProducer.

jlauridsen:
The ScrollBarProducer has code to produce a color based on the
view that is hosting the scrollbar.
If the hosting view has not changed its background color
instance then the background color of the scrollbar will remain
the same.

This works fine unless the hosting view uses a JBColor for its
background, in which case the color never changes after a theme
change.

Bug: IDEA-197981
Change-Id: I6d9c3482ae9712cc3add6f3c9465a0edd2dc6cd5


jlauridsen:
Any chance this will be accepted?
We would like to cherrypick this into Android Studio.

bulenkov:
@jlauridsen I don't think so. Problem is not in ScrollColorProducer, but in ColorUtil which doesn't work correctly with JBColor. I have fixed it here 36e49caaee003cc1c08bcb79b460f6a258dfbd07 Could you please check if this solves your problems in Android Studio?

jlauridsen:
This issue has been fixed in a different way, closing the pull request.

PR #873, opened 2018-09-26
Add an explicit import for com.intellij.openapi.module.Module

dahlstrom-g:
to disambiguate with java.lang.Module under JDK 9.

This is a work-around for:
https://youtrack.jetbrains.net/issue/KT-27141


cdracm:
pushed, thanks

PR #876, opened 2018-09-28
IDEA-199694 Preserve line breaks in XML when reformatting.

jlauridsen:
The line breaks of Xml comments at the end of an XML file were not preserved when the file was reformatted.


klikh:
It seems that #875 is a subset of this and thus can be closed?

jlauridsen:
Yes, thank you (sorry for the delay).

PR #877, opened 2018-09-29
[Fernflower] add test case for bad variable initialization

apixandru:
Add a test case for bad initialization of variables. The test is @Ignored so that it won't fail the build but it will be there to know that there is a case that has to be fixed. Also uncommented the tests that are currently failing and marking them as @Ignored so it will be obvious when running the test suite that there are test cases that have to be fixed.


apixandru:
I have also attempted to fix the issue by scanning all statements in the block to try and override the contant exprent type.

gorrus:
I'm observing strange effect with this:
javax.swing.text.html.AccessibleHTML.TableElementInfo.TableCellElementInfo#invalidate
has become:
```
protected void invalidate(boolean var1) {
            super.invalidate(var1);
            this.getParent().invalidate((boolean)1);
         }
```
com.sun.xml.internal.messaging.saaj.packaging.mime.internet.AsciiOutputStream#longLine
initialization is now:
`private boolean longLine = 0;`
and so on.
It seems like this requires more work.

BartvHelvert:
Hey, since there was no activity for years, I am going to close this PR.

PR #878, opened 2018-09-30
Detect extra syspath correctly

cboehme:
Code completion for GTK and other PyGObject based libraries is broken because the stub generation does not work. The reason for this is that the gi-repository directory which contains the typelibs is not detected by PyCharm due to a bug in the extra_syspath.py helper script. This script fails because it assumes that module.\_\_path__ is a string while it actually is a list. This causes the call of p.split() to fail and the script to terminate seemingly successfully (as AttributeErrors are caught and ignored) but without returning an extra syspath directory.

This commit fixes this problem by extracting the first element of the module.\_\_path__ list. After applying this fix stub generation finds the typelibs in the gi-repository and code completion for GTK is working again.

Fixes PY-29817: No code completion for GTK

### DIFF
--- a/python/helpers/extra_syspath.py
+++ b/python/helpers/extra_syspath.py
@@ -5,10 +5,10 @@ path = qualified_name.split(".")
 try:
   module =  __import__(qualified_name, globals(), locals(), [path[-1]])
   try:
-    p = module.__path__
+    p = module.__path__[0]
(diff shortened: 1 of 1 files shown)


kmare:
The patch works great for me. Please include it in the next PyCharm release.

PR #879, opened 2018-09-30
[Fernflower] Improve output for float and double literals

Pokechu22:
This PR does a few things regarding float and double literals:

* Adds support for `Float.MIN_NORMAL` and `Double.MIN_NORMAL`, added in Java 1.6
* Fixed invalid code being generated for `Float.NaN`, `Float.POSITIVE_INFINITY`, and `Float.NEGATIVE_INFINITY` when literals-as-is is enabled (`-lit=1`), which all used `0.0` for the denominator causing the expression to be a double instead of a float.  `0.0F` is now used, and `0.0D` is used for the double versions.
* Negative `MAX_VALUE` and similar are decompiled now, so `-Double.MAX_VALUE` is produced instead of `-1.7976931348623157e+308`.
* `Math.E` is now regenerated, instead of producing `2.7182818284590452354`
* `Math.PI` is now regenerated, along with various common factors of it.
    * The float version is recorded as `(float)Math.PI` for `3.1415927F`.  This is likely to occur if a program defines its own constant that is a float version of pi and uses that.
* Cases where a float literal was used to assign a double are now properly detected, so that `(double)0.01F` is output instead of `0.009999999776482582D`.  This is only applied in cases where the float value is shorter than the double value, so `1.0D` is not changed to `(double)1.0F`.
    * This also applies to the other constants, so `(double)((float)Math.PI / 4F)` is produced instead of `0.7853981852531433D`.  This is a bit ugly due to the double casting, but it's the easiest way to make sure it's not ambiguous.

With regards to possible future work on this, there are two things I can think of:
* `PI_DOUBLES` and `PI_FLOATS` are a bit awkward, and I think a more general way of handling fractions would work better than them; however I don't know of any easy method for doing this.
* The casting from a float literal to a double is a bit of a mess, and it would be nice if it could be avoided if it was known that it would be ambiguous.  However I don't think there's an easy way of doing this at this time.

Previously submitted to a fork at MinecraftForge/ForgeFlower#24.  I meant to upstream this a while back but I was busy and then forgot about it.  This does have a few differences from that: I did a bigger diff here; I have unit tests; and also negative `MAX_VALUE`/`MIN_VALUE`/etc are supported (which was done later in MinecraftForge/ForgeFlower#37).


plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/exps/ConstExprent.java
```diff
@@ -31,6 +31,67 @@
...
+    PI_DOUBLES.put(-PI_D / 10D, new String[] { "(-", " / 10D)" });
+    PI_FLOATS.put(PI_F / 10F, new String[] { "(", " / 10F)" });
+    PI_FLOATS.put(-PI_F / 10F, new String[] { "(-", " / 10F)" });
+
+    // Radian/degree conversions
+    PI_DOUBLES.put(PI_D / 180D, new String[] { "(", " / 180D)" });
+    PI_DOUBLES.put(180D / PI_D, new String[] { "(180D / ", ")" });
+    PI_FLOATS.put(PI_F / 180F, new String[] { "(", " / 180F)" });
+    PI_FLOATS.put(180F / PI_F, new String[] { "(180F / ", ")" });
+  }
```
trespasserw, line 94: I don't think deciphering these expressions significantly improves readability of a decompiled code, especially since it leaves out some constants anyway.

plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/exps/ConstExprent.java
```diff
@@ -234,6 +241,138 @@ else if (constType.equals(VarType.VARTYPE_CLASS)) {
...
+      float nearestFloatVal = (float)doubleVal;
+      if (doubleVal == (double)nearestFloatVal) {
+        // Value can be represented precisely as both a float and a double.
+        // Now check if the string representation as a float is nicer/shorter.
+        // If they're the same, there's no point in the cast and such (e.g. don't decompile 1.0D as (double)1.0F).
+        if (Float.toString(nearestFloatVal).length() < Double.toString(doubleVal).length()) {
+          // Include a cast to prevent using the wrong method call in ambiguous cases.
+          return createFloat(literal, nearestFloatVal, tracer).prepend("(double)");
+        }
+      }
```
trespasserw, line 294: Not sure about this. Upscaling floats to doubles doesn't seem too common, and additional casts may harm readability more than shorter literals are able to improve. Personally, I'd prefer to see an exact double value instead of a float approximation - as an additional reminder that IEEE 754 is tricky :)

plugins/java-decompiler/engine/test/org/jetbrains/java/decompiler/SingleClassesTest.java
```diff
@@ -127,6 +127,10 @@ public void tearDown() {
   @Test public void testSuspendLambda() { doTest("pkg/TestSuspendLambdaKt"); }
 
   private void doTest(String testFile, String... companionFiles) {
+    doTest(this.fixture, testFile, companionFiles);
+  }
+
+  static void doTest(DecompilerTestFixture fixture, String testFile, String... companionFiles) {
```
trespasserw, line 133: Meh. Would be better to move the method into `DecompilerTestFixture`.

plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/exps/ConstExprent.java
Pokechu22: The main reason I do this is because it happens in a surprisingly large number of places in the Minecraft codebase.  I don't know *why* Mojang has so many, but they do, and it makes things a lot more readable for e.g. `9.8F` and `0.1F`.  (See diff linked in the downstream PR).

plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/exps/ConstExprent.java
Pokechu22: I'd want to keep at least `PI / 2`, `PI * 2`, and the radian-degree ones as they're pretty helpful for me and come up in trig-related code (for instance in MC `57.2957763671875D` is in a number of places which is really hard to recognize, but happens to be `(double)(180F / (float)Math.PI))`).

It might be possible to try and identify any rational number (or rational number multiplied by pi) using [Farey sequences](https://en.wikipedia.org/wiki/Farey_sequence#Applications), though I don't know if that's worthwhile.

trespasserw: I see. Still, I do believe it would be more accurate to leave values as-is (except exact `Math.PI` and `Math.E` values), and provide possible interpretation in a comment block. The same applies for float approximations.

PR #881, opened 2018-10-04
Added an AccessibleName to Parameters table when refactoring parameters of a method. IDEA-199219

molaachir:
Added an AccessibleName to the Parameters table when refactoring - adding/changing parameters in a method. This is related to issue IDEA-199219, telling screen reader users the usage of the table that currently has focus. Note that this pull request does not fix the entire problem

### DIFF
--- a/.idea/libraries/pty4j.xml
+++ b/.idea/libraries/pty4j.xml
@@ -4,11 +4,31 @@
     <CLASSES>
       <root url="jar://$MAVEN_REPOSITORY$/org/jetbrains/pty4j/pty4j/0.8.4/pty4j-0.8.4.jar!/" />
       <root url="jar://$MAVEN_REPOSITORY$/org/jetbrains/pty4j/purejavacomm/0.0.11.1/purejavacomm-0.0.11.1.jar!/" />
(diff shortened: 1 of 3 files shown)


klikh:
Thanks for your contribution. It seems that you've modified ` .idea/libraries/pty4j.xml`. It doesn't seem relevant to the main issue, is it?

molaachir:
@klikh no, it is not relevant at all. I did not notice this, it is not immediately obvious for someone using screen readers. By the way, I fixed the entire problem now, should I expand this PR, or wait untill something has been done with this change and create a new one?

klikh:
I think the best would be to put the whole fix here (maybe in different commits, if they are atomic), and remove the pty4j.xml file. You may rebase this PR or open a new one.

molaachir:
Done, please let me know if the pty file is still there, as I have a hard time using the local hstory feature with a screen reader. This fix was suggested by Anton Tarasov, I tested it and it works in my specific case, using jaws

klikh:
Well, the file is still there: https://github.com/JetBrains/intellij-community/pull/881/files
But I guess @forantar can remove it himself after reviewing the fix. Anton, could you please take a look?

forantar:
@molaachir I've pushed the change: 2d988829c09af4bafa487f35bcffee7b5c9cec19. Thank you for the contribution.

PR #882, opened 2018-10-05
Refactoring - constant described several times

BorzdeG:
The text variable is used in several places - it replaces the use of a shared constant.


klikh:

klikh:
And could you also please provide a better commit message.

I'd also define the constant in the PluginManagerCore, rather than in the JBProtocolHandler, since the first one is more generic.

klikh:
Merged. Thank you for the contribution!

PR #883, opened 2018-10-07
add community discord

MattIPv4:


klikh:
Thank you for your interest, but we don't have plans to refer to Discord service at the moment.

MattIPv4:
This is with regards to https://youtrack.jetbrains.com/issue/IDEA-199404 which is currently unresolved.

PR #884, opened 2018-10-10
[Fernflower] Cleanup of collections-related classes

Pokechu22:
I'm not 100% sure how useful this actually is, but this PR cleans up some of the different collections classes Fernflower has.  Originally prompted by me seeing `VBStyleCollection` and running away in fear.

This does include some various renames.  I don't think that the new names are perfect, though.  I've changed `VBStyleCollection` to `KeyedList` and `FastSparseSet`/`FastFixedSet` to `UniversedSet`.  I didn't rename `SFormsFastMapDirect` since I don't fully understand the more general context of its use.

The code should behave exactly the same both before and after this change; all tests remain the same as before.


trespasserw:
I don't think this change is needed. Adding an explanatory comment to the original collections or replacing their usage with standard ones would be better.

Pokechu22:
Alright.  I don't have any plans to do further work related from this, but if anything in this PR is salvageable feel free to re-use it.

PR #885, opened 2018-10-10
IDEA-199921 Revert commit d8423f0 after commit b6a57a9

dahlstrom-g:
@develar?

### DIFF
--- a/platform/platform-impl/src/com/intellij/util/SingleAlarm.java
+++ b/platform/platform-impl/src/com/intellij/util/SingleAlarm.java
@@ -47,7 +47,7 @@ public final class SingleAlarm extends Alarm {
   }
 
   public void request(boolean forceRun) {
-    if (isEmpty() && !isDisposed()) {
+    if (isEmpty()) {
       addRequest(forceRun ? 0 : delay);
     }
   }


platform/platform-impl/src/com/intellij/openapi/wm/impl/ToolWindowManagerImpl.java
```diff
@@ -152,7 +152,7 @@ public void projectOpened(@NotNull Project project) {
         PropertyChangeListener focusListener = it -> myUpdateHeadersAlarm.request();
         KeyboardFocusManager keyboardFocusManager = KeyboardFocusManager.getCurrentKeyboardFocusManager();
         keyboardFocusManager.addPropertyChangeListener("focusOwner", focusListener);
-        Disposer.register(ToolWindowManagerImpl.this, () -> keyboardFocusManager.removePropertyChangeListener(focusListener));
+        Disposer.register(ToolWindowManagerImpl.this, () -> keyboardFocusManager.removePropertyChangeListener("focusOwner", focusListener));
```
dahlstrom-g: It looks like this was done by @fastasturtle in commit b6a57a9.

fastasturtle:
Oh,  sorry, I wasn't aware of this activity. Yes, I'd say calling `request` on a disposed alarm is a pre-condition violation (and "assertion failed" is an expected result of it), and `removePropertyChangeListener` fix should be enough.

Let's wait for @develar to merge the revert back, though, as he owns the code.

dahlstrom-g:
@develar?

develar:
Does this extra check leads to any runtime issues? Or because this check hides  real issues like "and "assertion failed" is an expected result of it)"?

dahlstrom-g:
I'm not aware of any problems encountered yet but I get the impression it's not what should be done in principle.

PR #886, opened 2018-10-12
Group External Modules in UI

fkorotkov:
Related to IDEA-198453

New `Group Modules` action looks like this in action:

intellij-grouping

### DIFF
--- a/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/project/ModuleData.java
+++ b/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/project/ModuleData.java
@@ -226,9 +226,7 @@ public class ModuleData extends AbstractNamedData implements Named, ExternalConf
 
   @Nullable
   public String getIdeGrouping() {
-    String parentGrouping = getIdeParentGrouping();
-    if (parentGrouping == null) return StringUtil.nullize(getInternalName());
-    return StringUtil.join(parentGrouping, ".", getExternalName());
+    return StringUtil.nullize(getInternalName());
   }
 
   @Nullable
--- a/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/project/ModuleData.java
+++ b/platform/external-system-api/src/com/intellij/openapi/externalSystem/model/project/ModuleData.java
@@ -226,10 +226,9 @@ public class ModuleData extends AbstractNamedData implements Named, ExternalConf
 
   @Nullable
   public String getIdeGrouping() {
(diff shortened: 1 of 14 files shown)


fkorotkov:
@vladsoroka looking a little bit more into how `BaseGradleProjectResolverExtension#createModule` works it seems that any `ModuleData` will always be a direct child of a `ProjectData` which will make the structure flatten. I was wondering if it makes sense to add some structure there instead of duplicating this logic here and potentially in `ExternalProjectDataSelectorDialog` (since it will be great to group modules there too to load/unload groups of modules). What do you think?

fkorotkov:
I've also added grouping to select projects to import in the last commit

image

vladsoroka:
There are several issues with the change (the `gradle-4.10.2/samples/java/multiproject` used for testing):

- `buildSrc` module is not shown at the `ExternalProjectDataSelectorDialog` 
- the `Web artifact` of the `services:webservice` project is not shown at the `ExternalProjectDataSelectorDialog`
- If the _File | Settings | ... | Gradle | Group modules_: **_Using qualified names_** settings is enabled some nested modules are not shown at the `Gradle tool window` and at the `ExternalProjectDataSelectorDialog` 
- because of issues above the summary about all and selected modules is incorrect
- If the _File | Settings | ... | Gradle | **_Create separate module per source set_** settings is enabled source set modules are on the same level that the main module `ExternalProjectDataSelectorDialog` 
screenshot from 2018-11-07 09-33-44

vladsoroka:
Thanks for the contribution, I've merged the changes using the following commits:
https://github.com/JetBrains/intellij-community/commit/9b48f9a2d9952bc5b83214e3d580bb4cf143e80b, https://github.com/JetBrains/intellij-community/commit/5874b0d25bc6a104fd2f53dbe1515ebba8a1de20, https://github.com/JetBrains/intellij-community/commit/882a673f3953a89a3128da193fd66b51900e7313, https://github.com/JetBrains/intellij-community/commit/8facb70f7f54f97ff5717f5f96c0b8766f92902f, https://github.com/JetBrains/intellij-community/commit/96407557687cdd6bf3f5430e03e0f81a2c591e0d, https://github.com/JetBrains/intellij-community/commit/93b30010bc762196ea3bbf5cf6bd2ca05d2b6e23

However I've had to add some additional fixes to the change:
https://github.com/JetBrains/intellij-community/commit/274c22df148aa46345fe9905a4a3179b0f91e084, https://github.com/JetBrains/intellij-community/commit/272411123d2525cc88bbc14a44adfce47fc9023f

PR #887, opened 2018-10-13
merge duplicated ArrayUtil.indexOf into single method

stsypanov:
There are some duplicates related to indexOf/lastIndexOf methods. I've merged them.


amaembo:
Please leave an explicit `indexOf` delegate in ArrayUtil class: this is part of a public API, so it would be good to keep it there. Relying on `ArrayUtil extends ArrayUtilRt` looks fragile to me.

stsypanov:
Sure, I'll revert it.  Btw, is it reasonable to use `Arrays::asList` for `indexOf`/`find`? This will bring additional allocation, but from another point of view reduce amount of compiled code.

amaembo:
I think, for such low-level methods specialized implementation is better than delecation to Arrays.asList

PR #889, opened 2018-10-14
IDEA-200509 - More suggested types of constructor method references

bedla:
Hi,
pls take a look at ticket https://youtrack.jetbrains.com/issue/IDEA-200509 ... what do you think about changes?
Thx,
Ivos

### DIFF
--- a/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
+++ b/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
@@ -23,13 +23,13 @@ import com.intellij.psi.search.searches.AnnotatedMembersSearch;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.psi.util.PsiUtil;
 import com.intellij.psi.util.TypeConversionUtil;
-import com.intellij.util.NullableFunction;
 import com.intellij.util.Processor;
 import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.*;
+import java.util.function.Function;
 
 public class FunctionalInterfaceSuggester {
   public static final String[] FUNCTIONAL_INTERFACES = {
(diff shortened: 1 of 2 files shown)


java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -139,27 +139,29 @@
...
     final Set<PsiType> types = new HashSet<>();
     final Processor<PsiMember> consumer = member -> {
-      if (member instanceof PsiClass && Java15APIUsageInspection.getLastIncompatibleLanguageLevel(member, PsiUtil.getLanguageLevel(element)) == null) {
+      if (member instanceof PsiClass &&
+          Java15APIUsageInspection.getLastIncompatibleLanguageLevel(member, PsiUtil.getLanguageLevel(element)) == null) {
         if (!JavaPsiFacade.getInstance(project).getResolveHelper().isAccessible(member, element, null)) {
           return true;
         }
-        ContainerUtil.addIfNotNull(types, acceptanceChecker.fun((PsiClass)member));
+        ContainerUtil.addAll(types, acceptanceChecker.apply((PsiClass)member));
```
bedla: I have changes suggester signature to return list of types, because I am suggesting every suitable ctor when expression is constructor method reference.

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -182,77 +184,149 @@
...
+      if (type != null) {
+        types.add(type);
       }
     }
-    return null;
+    return types;
+  }
+
+  @Nullable
+  private static PsiType composeLambdaExpressionType(PsiClass interface2Consider,
```
bedla: At this context I was trying to somehow find return type of lambda-expression using com.intellij.psi.LambdaUtil#getReturnExpressions but was not able to find easy solution how to detect return type from list of PsiExpressions.

akozlova:
Thanks for the PR! I've fixed a bug which lead to incorrectly processing of constructors and as you've suggested started to process all available overloads for methods and constructors. In order to ensure that I don't break your idea, I copied the test data from this PR though the code is now completely different so it's committed separately as 40a9a6fc2d6e34a5face6aa9a4379adbf931e91b, will be available on github shortly

bedla:
cool, it was fun to help :) ivos

bedla:
ahh i see... i like that overloads change and usage of `getElementFactory().createExpressionFromText` (much better than mine `Main` class "hack").

akozlova:
The test data is always very helpful, many thanks for it. 

The `Main` class is totally ok if there are more or less complex situations (like with `foo::foo` actually) but when the expressions are nearly context independent, I think that explicit creation looks better.

Sorry, that I commit from my name, it would probably better to get your commit and then refactor but it looked too complicated from the first glance.

bedla:
no problem i am signed for this PR :) 

i did not noticed that overloads and have refactored it without that knowledge, thats why it looked complicated - to return it "back".

PR #893, opened 2018-10-19
IDEA-200871 Tool Window should NOT transfer focus unconditionally.

jlauridsen:
Only transfer focus if this tool window has the focus in the first
place.

Change-Id: Icad107068eb5f2de46de218c87f8de471367e320

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/wm/impl/ToolWindowsPane.java
+++ b/platform/platform-impl/src/com/intellij/openapi/wm/impl/ToolWindowsPane.java
@@ -14,6 +14,7 @@ import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.util.registry.Registry;
 import com.intellij.openapi.util.registry.RegistryValue;
 import com.intellij.openapi.util.registry.RegistryValueListener;
+import com.intellij.openapi.wm.IdeFocusManager;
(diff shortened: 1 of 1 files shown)


platform/platform-impl/src/com/intellij/openapi/wm/impl/ToolWindowsPane.java
```diff
@@ -1127,14 +1127,29 @@ public final void run() {
...
+  private void transferExistingFocus(@Nullable String id) {
+    if (id == null) {
+      return;
+    }
+    ToolWindow toolWindow = myManager.getToolWindow(id);
+    if (toolWindow == null) {
+      return;
+    }
+    Component toolComponent = toolWindow.getComponent();
+    Component owner = KeyboardFocusManager.getCurrentKeyboardFocusManager().getFocusOwner();
```
yole: I think this should use `IdeFocusManager.getFocusOwner()` instead. Can you please make this change and check that the code still works as you need?

jlauridsen: Yes, using the IdeFocusManager works fine.
I updated the pull request.

jlauridsen:
This was fixed. Thank you.

PR #895, opened 2018-10-23
[platform] Support Python 2.6 in python launcher

d10n:
sys.version_info does not return a named tuple until Python 2.7


trespasserw:
Thanks!
BTW, do people really still use Python 2.6?

d10n:
Yes, through a series of unfortunate events I'm doing some development in CentOS 6 and Python 2.6 is the latest Python version in the CentOS 6 repos. Thanks for merging!

PR #896, opened 2018-10-24
[gradle] allow to customize resolving modules per source set

fkorotkov:
It will be nice to specify the default for this setting. On a big projects it does make sense to not have 2 modules per project. 😅


vladsoroka:
Hi, the gradle test source set and other custom source set might have dependencies which can not be correctly expressed using he single IDEA module. And the only option for such projects is to use separate module per sources set. However indeed, for gradle projects that have more simple configuration the single IDE module can be more convenient.

Anyway now you can choose which mode you'd want to use on the initial import and also you can change the option further in the Gradle IDEA settings.

Also we have the gradle plugin which is intended to bring the IDE specific configuration from the build, so it will be shared with other project's users.

You can vote for the related issue at [https://github.com/JetBrains/gradle-idea-ext-plugin/issues/22](https://github.com/JetBrains/gradle-idea-ext-plugin/issues/22)

PR #897, opened 2018-10-25
Guard against NPE from early publication in NonClasspathClassFinder

dahlstrom-g:
As revealed in
https://youtrack.jetbrains.com/issue/IDEA-200807, the parent class of
GradleClassFinder - NonClasspathClassFinder - will publish references to
the object under construction in the NonClasspathClassFinder
constructor. This early publication can allow other threads to call
methods on the object under construction before the fields of the object
under construction are initialized. In this particular case, this led to
a NullPointerException (NPE) in the GradleClassFinder class.

A better solution would have been to remove the registration of the
listeners out of the constructors completely.  However, the extension
system only allows us to instantiate objects of extensions by calling
their constructors or by specifying the values for the fields of a bean
class. As stated in IDEA-200807, there does not seem to be another spot
for registering listeners other than in the constructor of the
NonClasspathClassFinder class.

This workaround marks the field as volatile and includes a null check to
guard against an NPE. This ensures the threads running
GradleClassFinder.clearCache will encounter one of three cases:

1. The thread reads a null value for the myCaches variable. In this
case, the object has not completed construction yet. In this case, the
thread will exit from the method without running any method on the
object referenced by myCaches. No NPE will be generated. When the
myCaches variable finally gets initialized in the constructor, the map
will be empty. An empty map is the state the map should be in after the
clearCache method is called, so there is no need to actually call
clear() on myCaches in this case.

2. The thread reads a non-null value for myCaches, and the object has
not completed construction. In this case, however, the myCaches map
should still be empty, so calling clear() on the variable will not
change the map.

3. The thread reads a non-null value for myCaches, and the
object has completed construction. In this case, calling clearCache()
will empty the map, as expected in the case where clearCache() is called
in a more typical context rather than in cases 1 or 2.

Additional considerations we should beware of due to early publication
problems are whether or not GradleClassFinder.getCache can be called
before the object has completed construction. If getCache() is called
before construction is complete, then there may be another NPE generated
in getCache. However, there is currently no known path to call
getCache() before the object completes construction.

We should also be aware of the myBuildClasspathManager instance
variable, as this variable is published prematurely together with the
object under construction. If another thread were to read the
myBuildClasspathManager instance variable before the object completes
construction, it might be possible for the thread to read an
uninitialized value for the myBuildClasspathManager variable. However,
there is currently no known path to read this variable before the object
completes construction.

### DIFF
--- a/plugins/gradle/java/src/config/GradleClassFinder.java
+++ b/plugins/gradle/java/src/config/GradleClassFinder.java
@@ -43,7 +43,16 @@ import java.util.Map;
 public class GradleClassFinder extends NonClasspathClassFinder {
 
   @NotNull private final GradleBuildClasspathManager myBuildClasspathManager;
-  private final Map<String, PackageDirectoryCache> myCaches;
+
+  // Bug: IDEA-200807
+  // The parent class can publish a reference to this object before the constructor has returned.
+  // Thus, it's possible that not all fields of this object are initialized by the time they
+  // are accessed in clearCache(). Workaround is to mark this field as volatile and do a
+  // null check in clearCache().
+  //
+  // It is unknown if different threads can read stale or uninitialized values for the
+  // myBuildClasspathManager instance variable at this moment.
+  private volatile Map<String, PackageDirectoryCache> myCaches;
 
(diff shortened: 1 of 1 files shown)


dahlstrom-g:
Attn: @donnerpeter

PR #898, opened 2018-10-25
Fixed an issue related to accessibility of changes tree

molaachir:
Changes tree that shows up when reverting issues and commiting was not accessible to screen readers. File names where not read out, and checked/unchecked states of the files (signaling inclusion of files) where not announced. Fix tested with NVDA and Jaws. NVDA announces a bit too much information for my taste, but that is probably due to settings. Related YouTrack issue: https://youtrack.jetbrains.com/issue/IDEA-201167*

### DIFF
--- a/platform/util/src/com/intellij/util/ui/ThreeStateCheckBox.java
+++ b/platform/util/src/com/intellij/util/ui/ThreeStateCheckBox.java
@@ -185,7 +185,7 @@ public class ThreeStateCheckBox extends JCheckBox {
    * Emulate accessibility behavior of tri-state checkboxes, as tri-state checkboxes
    * are not part of the JAB specification.
    */
-  protected class AccessibleThreeStateCheckBox extends AccessibleJCheckBox {
+  public class AccessibleThreeStateCheckBox extends AccessibleJCheckBox {
     @Override
     public AccessibleRole getAccessibleRole() {
       if (myThirdStateEnabled) {
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java
@@ -41,6 +41,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.accessibility.AccessibleContext;
+import javax.accessibility.AccessibleRole;
 import javax.swing.*;
 import javax.swing.event.TreeSelectionEvent;
 import javax.swing.event.TreeSelectionListener;
@@ -597,12 +598,32 @@ public abstract class ChangesTree extends Tree implements DataProvider {
       return myTextRenderer.getToolTipText();
     }
 
+    @Override
     public AccessibleContext getAccessibleContext() {
       if(accessibleContext == null) {
-        myCheckBox.getAccessibleContext().setAccessibleName(myTextRenderer.getAccessibleContext().getAccessibleName());
-        return myCheckBox.getAccessibleContext();
+        return new AccessibleChangesTreeRenderer();
       }
       return accessibleContext;
+    }
+
+    // this is, in effect, a ThreeStateCheckboxRenderer, since another renderer is returned when the checkbox is not enabled
+    // Therefore, it has to behave like a checkbox, in almost every way
+    protected class AccessibleChangesTreeRenderer extends AccessibleJTree{
+
+      @Override
+      public String getAccessibleName() {
+        // AccessibleThreeStateCheckbox returns standard accessible name, plus checked, unchecked, partially checked information
+        // therefore, all we need to do is add the file name as the standard accessibleName, and let the AccessibleThreeStateCheckBox do the rest
+        myCheckBox.getAccessibleContext().setAccessibleName(myTextRenderer.getAccessibleContext().getAccessibleName());
+        return myCheckBox.getAccessibleContext().getAccessibleName();
+      }
+
+      @Override
+      public AccessibleRole getAccessibleRole() {
+        // Because of a problem with NVDA, we have to make this an AccessibleLabel
+        // if we don't do this, NVDA will read out the entire tree path, causing confusion
+        return AccessibleRole.LABEL;
+      }
     }
   }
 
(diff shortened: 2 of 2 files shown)


forantar:
@molaachir Thank you for the patch, let me please test it.

forantar:
@molaachir Well, it seems we can just use 'myTextRenderer' which is fully capable of the text representation of the cell. Please, try the simplified patch below:

Index: platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java
IDEA additional info:
<+>UTF-8
--- platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java	(revision 224d9ced7023142eb28eba9d1def3011792a0075)
+++ platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java	(date 1540559541948)
@@ -40,6 +40,7 @@
 import org.jetbrains.annotations.NotNull;
 
+import javax.accessibility.AccessibleContext;
 import javax.swing.*;
 import javax.swing.event.TreeSelectionEvent;
 import javax.swing.event.TreeSelectionListener;
@@ -595,6 +596,14 @@
     public String getToolTipText() {
       return myTextRenderer.getToolTipText();
     }
+
+    @Override
+    public AccessibleContext getAccessibleContext() {
+      if (accessibleContext == null) {
+        return accessibleContext = myTextRenderer.getAccessibleContext();
+      }
+      return accessibleContext;
+    }
   }

molaachir:
@forantar this does work for the file names, but it does not announce the state of the checkbox, omitting crucial information - at least when using Jaws. This is also the reason i chose to use the checkbox and add the file name to its name instead of using myTextRenderer.

forantar:
@molaachir I've got your point. Yes, you're right the checkbox state should be included. However, with NVDA I observe some odd behavior. When I select an item in the tree NVDA reads all the path of the item starting from the root. Jaws doesn't do that. As I could quickly identify, this is specific to AccessibleRole.TOGGLE_BUTTON role the ThreeStateCheckBox reports. Same is for AccessibleRole.CHECK_BOX, but when I for experiment change it to, say, AccessibleRole.LABEL then NVDA reads only the selected item. So, my question: is the default NVDA behavior ok from your point of view? If not, should we wrap AccessibleContext here with something more reasonable?

molaachir:
@forantar I found that too. I thought it was specific to my settings, but if you have the same problem, I think we should fix it. it confused me the first time, because it is hard to immediately understand what NVDA is doing there. You need to wait in order to know where you are in the tree, it is an unecessary addition, you don't need the path. I'll resolve this, unless you have the changes ready already? Would you like me to squash my changes? or are two commits ok??

forantar:
@molaachir No, I don't have fix for that problem. So, I'd appreciate if you investigate it please. Also, as it makes your current fix hard to use, I think we're better to find the solution first and only then commit.

molaachir:
@forantar: done. I cannot rebase at the moment, since vim is hard to use with my screen reader and the IntelliJ rebase dialogs need some work to become usable. another problem is that I saw 2 files were committed, but I could not read which files, untill this pr goes through. Please test this, and let me know if it works, and let's see which other file got sucked in to the commit.

forantar:
@molaachir In your last fix I see that AccessibleChangesTreeRenderer extends AccessibleJTree which doesn't seem right to me. Was it a typo? Anyway, I suppose we want to extend the original AccessibleThreeStateCheckBox, but it is an inner class in ThreeStateCheckBox and we can't do that. Why don't we use our standard approach in this case, namely AccessibleContextDelegate? Please, see the patch below. I also assign the newly created context to the 'accessibleContext' field to match the common pattern used in such case. The only thing I'm not sure about is that we override the role of the checkbox's accessible context to label. Probably this can hide some accessible functionality of a checkbox from accessibility tools. But anyway it provides some basic stuff to a screen reader, so I won't worry about it until we find a real problem with it.

```
Index: platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java
IDEA additional info:
<+>UTF-8
--- platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java	(revision 2494a7d763f498e51dd61829403579402d5e3ba6)
+++ platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ui/ChangesTree.java	(date 1540910433661)
@@ -30,6 +30,7 @@
 import com.intellij.util.containers.ContainerUtil;
 import com.intellij.util.ui.ThreeStateCheckBox;
 import com.intellij.util.ui.UIUtil;
+import com.intellij.util.ui.accessibility.AccessibleContextDelegate;
 import com.intellij.util.ui.tree.TreeUtil;
 import com.intellij.util.ui.tree.WideSelectionTreeUI;
 import com.intellij.vcsUtil.VcsUtil;
@@ -40,6 +41,8 @@
 import org.jetbrains.annotations.NotNull;
 
+import javax.accessibility.AccessibleContext;
+import javax.accessibility.AccessibleRole;
 import javax.swing.*;
 import javax.swing.event.TreeSelectionEvent;
 import javax.swing.event.TreeSelectionListener;
@@ -595,6 +598,32 @@
     public String getToolTipText() {
       return myTextRenderer.getToolTipText();
     }
+
+    @Override
+    public AccessibleContext getAccessibleContext() {
+      if (accessibleContext == null) {
+        return accessibleContext = new AccessibleContextDelegate(myCheckBox.getAccessibleContext()) {
+          @Override
+          protected Container getDelegateParent() {
+            return getParent();
+          }
+
+          @Override
+          public String getAccessibleName() {
+            myCheckBox.getAccessibleContext().setAccessibleName(myTextRenderer.getAccessibleContext().getAccessibleName());
+            return myCheckBox.getAccessibleContext().getAccessibleName();
+          }
+
+          @Override
+          public AccessibleRole getAccessibleRole() {
+            // Because of a problem with NVDA we have to make this a LABEL,
+            // or otherwise NVDA will read out the entire tree path, causing confusion.
+            return AccessibleRole.LABEL;
+          }
+        };
+      }
+      return accessibleContext;
+    }
   }
```

molaachir:
@forantar, I have only just started to contribute, so I guess this was a beginners mistake from me :)

I think your patch makes way more sense, is there a way to directly apply it, or do you want me to test it first? I guess this works the same way as the table fix I added a couple of weeks ago, so I don't see any reason why it wouldn't work.

forantar:
@molaachir No problem, thank you for the contribution! I've just pushed the fix.

molaachir:
@forantar cool, thanks, that'll help a lot

PR #900, opened 2018-10-25
Fix JSON index performance

yuriy-solodkyy:
The existing implementation of the JsonSchemaFileValuesIndex relies on psi trees to extract just two top level properties $id and $schema. On projects having multiple relatively large JSON files it often takes substantial time. Many larger JSON files come in JARs for some popular third party libraries. 

This PR replaces the implementation of map() function with a Lexer based extraction. 

Apart from the submitted unit tests, it has been tested that the editor properly recognises the JSON schema referred from a JSON file.


klikh:

json/src/com/jetbrains/jsonSchema/impl/JsonSchemaFileValuesIndex.java
```diff
@@ -44,15 +48,60 @@
...
+              case "$id":
+                captureValueIfString(lexer, map, JsonCachedValues.ID_CACHE_KEY);
+                break;
+              case "\"$schema\"":
+              case "$schema":
+                captureValueIfString(lexer, map, JsonCachedValues.URL_CACHE_KEY);
+                break;
+            }
+          }
+          lexer.advance();
```
zhuravlikjb: Looks like we can break the iteration if we've already found both properties. In most real-life cases, these properties are located on top, so skipping the remaining part should help a lot.

zhuravlikjb:
Thanks a lot for the improvement! There's one place that can be improved further. If you agree with that and can fix it, I'll wait before merging. If not, then I'll merge as is, because this PR already significantly optimizes the performance.

zhuravlikjb:
Thank you for the fast update!

json/src/com/jetbrains/jsonSchema/impl/JsonSchemaFileValuesIndex.java
yuriy-solodkyy: Thank you! Done. It won't help in the case when none is present, though.
I've also allowed the keys to be single-quoted for compatibility with json5.

zhuravlikjb: Thanks. Yes, that's true, there's nothing to improve when none of the properties is there.
Merged the PR, thanks for your work.

yuriy-solodkyy:
@zhuravlikjb  I don't see the commits in master (anywhere in github repo). Did it break any tests ?

zhuravlikjb:
@yuriy-solodkyy Sorry, I accepted the PR here, but not to our main repository, and so it was synced back without the commits. Fixed, now the commits will appear here in about an hour. Thanks a lot for reminding!

Also will cherry-pick the fix to the release branch soon.

PR #901, opened 2018-10-26
[gradle] don't list too many projects in a dialog

fkorotkov:
For a big project it just doesn't make sense to list more than 5 projects. The list will be just too big for a screen. Instead, let's just mention how many additional modules are affected.

to: @vladsoroka


vladsoroka:
Merged, thanks!

PR #902, opened 2018-10-27
Fix StringSearcher.scan bug

takc923:
`new StringSearcher(pattern, any, false).scan(text)` starts scan from index `text.length() - myPatternLength + 1` in `text`.
So it throws StringIndexOutOfBoundsException if myPatternLength == 1.


cdracm:
thanks, pushed

PR #903, opened 2018-10-27
IDEA-201317: Add Cucumber4 Formatter

james-bjss:
### Background:
Add the ability to see tests executing in realtime as per current Cucumber3 Behaviour. Currently IntelliJ only displays test results at the end of the complete test run.

See details tracked here: https://youtrack.jetbrains.com/issue/IDEA-201317

### Summary:
Added a new Cucumber formatter module that implements the ```ConcurrentEventListener``` interface
* New formatter simply delegates to the existing Cucumber3 formatter.

**NB**. No attempt has been made to make this threadsafe as the current IDE plugin for cucumber does not pass the new --threads option and therefore cucumber defaults to a single thread. If this functionality is required then further changes will need to be made.

Also note the cucumber plugin will need to be updated to pass the new Formatter package name.

### DIFF
--- a/.idea/libraries/cucumber_core_4_0_1.xml
+++ b/.idea/libraries/cucumber_core_4_0_1.xml
@@ -1,0 +1,16 @@
+<component name="libraryTable">
+  <library name="cucumber-core:4.0.1" type="repository">
+    <properties maven-id="io.cucumber:cucumber-core:4.0.1" />
+    <CLASSES>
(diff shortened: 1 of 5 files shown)


james-bjss:
@avokin Hi Andrey, I have made the changes as per our discussion. I will raise the changes for `org.jetbrains.plugins.cucumber.java.run.CucumberJavaRunConfigurationProducer` separately.

avokin:
Merged

PR #904, opened 2018-10-28
[IDEA-201282] Ignore setting Maven Resolver's LoggerFactory for Maven…

slachiewicz:
… 3.6.0

From Maven 3.6.0 and Maven Resolver 1.3.0 internal logging has been migrated to the SLF4J.


plugins/maven/maven3-server-impl/src/org/jetbrains/idea/maven/server/Maven3ServerEmbedderImpl.java
```diff
@@ -1236,12 +1236,20 @@ private Artifact resolve(@NotNull final Artifact artifact, @NotNull final List<A
       final org.eclipse.aether.impl.ArtifactResolver artifactResolver = getComponent(org.eclipse.aether.impl.ArtifactResolver.class);
       final MyLoggerFactory loggerFactory = new MyLoggerFactory();
       if (artifactResolver instanceof DefaultArtifactResolver) {
-        ((DefaultArtifactResolver)artifactResolver).setLoggerFactory(loggerFactory);
+        try {
+          ((DefaultArtifactResolver)artifactResolver).setLoggerFactory(loggerFactory);
+        } catch (NoSuchMethodError ignore) {
+          // ignore for maven 3.6.0 as maven-resolver 1.3.0 migrated to slf4j
```
amakeev: Could you please make sure the problem is not suppressed, but instead reported in the IDE? (you can use `loggerFactory` instance to report error back to the IDE process).

slachiewicz:
Code modified but I don't know where I can find this message logged. In all code in this DefaultArtifactResolver class there is only one debug line and here we catch specific NoSuchMethodError exception.

rosti-il:
Take a look at my pull request with a better fix: https://github.com/JetBrains/intellij-community/pull/999

PR #905, opened 2018-10-30
Added test state to AccessibleName of test results entry

molaachir:
This PR adds the state information, visually represented by icons, to the AccessibleName of the entries in the test results tree structure. It partly solves: https://youtrack.jetbrains.com/issue/IDEA-193979

It does not solve the problem of the "Hide successful tests" button not being found by tabbing through the tool window, but the most pressing problem - the test results not being immediately clear, is gone.

### DIFF
--- a/platform/platform-resources-en/src/messages/SMTestsRunnerBundle.properties
+++ b/platform/platform-resources-en/src/messages/SMTestsRunnerBundle.properties
@@ -18,6 +18,14 @@ sm.test.runner.ui.tests.tree.presentation.labels.no.tests.were.found.with.errors
 sm.test.runner.ui.tests.tree.presentation.labels.all.tests.passed=All Tests Passed
 sm.test.runner.ui.tests.tree.presentation.labels.all.but.ignored.passed=All Tests Passed (except ignored)
 sm.test.runner.ui.tests.tree.presentation.labels.test.noname=<no name>
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.passed=passed
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.ignored=ignored
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.failed=failed
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.terminated=terminated
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.tests.running=running
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.not.run=Not run
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.error=error
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.contains.ignored.tests=Contains ignored tests
 
 sm.test.runner.ui.tabs.statistics.columns.test.title=Test
 sm.test.runner.ui.tabs.statistics.columns.test.total.title=Total:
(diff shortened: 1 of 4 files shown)


molaachir:
@forantar Just a little reminder that this exists :)

molaachir:
@forantar could you please have a look? This is a crucial addition, I keep running into tests - and i cannot see if they succeeded. Thanks.

molaachir:
@forantar, a reminder that this exists. I could really use its addition, or a similar fix. I would save me a considerable amount of time.

platform/platform-resources-en/src/messages/SMTestsRunnerBundle.properties
```diff
@@ -18,6 +18,14 @@ sm.test.runner.ui.tests.tree.presentation.labels.no.tests.were.found.with.errors
 sm.test.runner.ui.tests.tree.presentation.labels.all.tests.passed=All Tests Passed
 sm.test.runner.ui.tests.tree.presentation.labels.all.but.ignored.passed=All Tests Passed (except ignored)
 sm.test.runner.ui.tests.tree.presentation.labels.test.noname=<no name>
+sm.test.runner.ui.tests.tree.accessible.presentation.labels.test.passed=passed
```
akozlova: `Magnitute` already contains title, why do you need to repeat it?

molaachir: This is specifically meant to capture the meaning of the left hand side icons in the test result tree. This means that the message spoken by the screen reader has to be short and to the point. It serves a very specific purpose, thus it should have its own message. That way, if other messages change for any reason, the message spoken by the screen reader does not change unintentionally. This is something that wouldn't be easily found, because so few people use a screen reader.

akozlova: It's code duplication and it would lead to the common duplication problem: when the renderer would be intentionally changed, these messages won't be changed automatically so the reader would be out-of-date.

molaachir: I think we misunderstand each other. This is not code duplication, because it serves a different purpose than the titles you see in the rest of the file, even though the words you see are mostly the same. This is meant to quickly convey the meaning of the icons on the left hand side of the test names, so the green OK or the orange icon when a test failed or was ignored. Without these, you do not know what the state of a specific test is. You only know that some test has failed, or was ignored. Good luck finding it. If the renderer changes, you would only have to change these if test outcomes change. Again, it is not the renderer that matters here, it is the test outcomes. You need the outcomes of specific tests, without that you are stuck performing every test on it's own, just to make sure that you find any errors. It is a huge time waster. I want the info from the icons, no titles, I want quick, concise information. Which is why I almost exclusively use one word.

molaachir:
@forantar Does it make sense for me to fix the conflicts here? I'd really like this to be merged, as it would save me, and anyone else with a screen reader enormous amounts of time. i keep running into situations where I want to no the result of a single test run within a group of tests. But I'll only fix the conflicts if there is a chance of this being merged.

lukaszgo1:
@forantar Just one more user  who would benefit from this being reviewed and merged.

thgcode:
+1, I also would like to see it merged.

dmitrii-drobotov:
This was implemented in the scope of issue [IJPL-129540](https://youtrack.jetbrains.com/issue/IJPL-129540) and is available from the 2023.2 IDE version. The test result is now added after the name of the test in the tree.

PR #906, opened 2018-10-30
Suspend indexing while refreshing project structure

fkorotkov:
In my testing it actually improved the initial importing time from roughly 14 minutes to just 3 and a half minutes.

Also included pausing indexing while choosing projects to import. It's very handy with `external.system.show.selective.import.dialog`  to avoid unnecessary indexing.


vladsoroka:
Merged, thank you!

PR #907, opened 2018-10-31
Fix race conditions in DumbServiceImpl and in ProgressSuspender.

yuriy-solodkyy:
The PR fixes a race condition related to the they way DumbServiceImpl
and ProgressSuspender interoperate.

The race condition occurs when a batchChangeStarted() notification
is delivered after the last dumb mode task has made its last check-
cancelled call (if it does any at all), but before the dumb mode
background process has ended.

If the notification arrives between these two events, the following
sequence of event happens:
1) the current ProgressSuspender still gets suspended but it does
   not suspend any tasks and
2) the dumb mode background process exists and the ProgressSuspender
   remains registered as a check cancelled hook.
3) A subsequent dumb mode starts and registers its own progress
   suspender.
*) At this point two progress suspenders are registered as a check-
   cancelled hook.
3) ** Any dumb mode task calling check-cancelled gets frozen by the
   previous progress suspender.

This PR fixes the problem by:
a) preemptively setting myCurrentsuspender to null at the earliest
   possible moment.
b) since a) only reduces the probability of the race condition, the
   close() method is added to ProgressSuspender which corretly cleans
   up the suspender and unregisters it from the list of check-cancelled
   hooks, if it happens to remain registered.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/project/DumbServiceImpl.java
+++ b/platform/platform-impl/src/com/intellij/openapi/project/DumbServiceImpl.java
@@ -122,7 +122,9 @@ public class DumbServiceImpl extends DumbService implements Disposable, Modifica
     ApplicationManager.getApplication().assertIsDispatchThread();
     myUpdatesQueue.clear();
     myQueuedEquivalences.clear();
-    myRunWhenSmartQueue.clear();
+    synchronized (myRunWhenSmartQueue) {
+      myRunWhenSmartQueue.clear();
+    }
     for (DumbModeTask task : new ArrayList<>(myProgresses.keySet())) {
       cancelTask(task);
       Disposer.dispose(task);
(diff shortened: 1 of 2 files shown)


platform/platform-impl/src/com/intellij/openapi/project/DumbServiceImpl.java
```diff
@@ -326,6 +328,10 @@ private void enterDumbMode(@Nullable TransactionId contextTransaction, @NotNull
 
   private void queueUpdateFinished() {
     if (myState.compareAndSet(State.RUNNING_DUMB_TASKS, State.WAITING_FOR_FINISH)) {
+      // There is no tasks to suspend with the current suspender. If the execution reverts to the dumb mode, a new suspender will be
+      // created. The suspender might have already got suspended between the point of the last check cancelled call and this point, however.
+      // It if has happened it will be cleaned up when the suspender is closed on the background process thread.
```
donnerpeter: small typo

yuriy-solodkyy: I updated the comment to make more clear, but I'm not sure what you refer to.

donnerpeter: "it if" should probably be "if it"

yuriy-solodkyy: sure, updated

donnerpeter:
Merged manually, thanks!

PR #908, opened 2018-11-02
Add option 'ignore != 0' to inspection 'Conditional expression with negated condition

rillig:


BasLeijdekkers:
Hi @rillig, is there a YouTrack request open for this? If not, could you create one?

rillig:
@BasLeijdekkers I just created https://youtrack.jetbrains.com/issue/IDEA-202323 for it.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/InspectionGadgetsBundle.properties
```diff
@@ -1241,7 +1241,8 @@ string.equals.empty.string.option.do.not.add.null.check=Do not report when null-
 tail.recursion.replace.quickfix=Replace tail recursion with iteration
 if.statement.with.too.many.branches.max.option=Maximum number of branches:
 if.statement.with.too.many.branches.problem.descriptor=<code>#ref</code> has too many branches ({0}) #loc
-negated.conditional.ignore.option=Ignore '!= null' comparisons
+negated.conditional.ignore.null.option=Ignore '!= null' comparisons
```
BasLeijdekkers: Could you just reuse the two identical "negated.if.else..." properties below, and remove the two "negated.conditional..." properties?

plugins/InspectionGadgets/test/com/siyeh/igtest/controlflow/negated_conditional_expression/NegatedConditionalExpression.java
```diff
@@ -2,13 +2,21 @@
...
-
-    boolean x(NegatedConditionalExpression myClass) {
-      System.out.println(!(myClass.myField2 != null));
-      System.out.println(myField2 != null ? !myField2.equals(myClass.myField2) : myClass.myField2 != null);
-      final boolean b = myField1 != null ;
-      return <warning descr="Negating conditional expression">!</warning>(myField2 != null ? !myField2.equals(myClass.myField2) : myClass.myField2 != null);
-    }
-}
\ No newline at end of file
+  String myField1 = null;
```
BasLeijdekkers: This is the data file of the test of a different inspection. NegatedConditionalInspection does not yet have a test. Could you create one and undo this change?

rillig:
@BasLeijdekkers The next try is ready. Is there a better way for testing different options of one inspection? I'd rather have liked a method `doTest(inspection)` instead, but currently both calls to `getInspection` are during the setup phase of the test, and I also didn't find a hook for modifying the inspection before actually applying it.

plugins/InspectionGadgets/testsrc/com/siyeh/ig/controlflow/NegatedConditionalInspectionTest.java
```diff
@@ -0,0 +1,50 @@
...
+  public void testNegatedConditionalNullAndZeroAllowed() {
+    doTest();
+  }
+
+  public void testNegatedConditionalNullAndZeroDisallowed() {
+    doTest();
+  }
+
+  public void testNegatedConditionalNullAllowed() {
+    doTest();
```
BasLeijdekkers: What you can do is call `myFixture.enableInspections(inspection);` before `doTest()` instead. Then you don't need `getInspection()`, but unfortunately it is `abstract` so a dummy implementation is needed. Do you want to make this change?

rillig: Thanks for the hint. The test code looks much better now.

I wonder why `enableInspections` has a comment "Should be called BEFORE setUp" when it works well to add an inspection afterwards and it even overrides the previous dummy inspection from `getInspection`.

BasLeijdekkers: That's weird, because it is actually not possible to call the method before setUp() is called. I'll try to get this fixed.

Also I though of a cleaner way to handle configuration of the inspection under test. Check the manual merge when it becomes available on GitHub.

BasLeijdekkers:
Thanks for your efforts, merged manually as d001216c69f39d03b425d9bcfb337e2e349038d5
For next time: please squash all commits into one, this makes merging into our project easier (unfortunately we can't use GitHub's automatic pull request merging).
Thanks again.

plugins/InspectionGadgets/testsrc/com/siyeh/ig/controlflow/NegatedConditionalInspectionTest.java
rillig: 👍 I don't think there's any way to make the test even clearer now.

PR #910, opened 2018-11-03
Don't suggest using streams if minSdkVersion < 24

dahlstrom-g:
This CL makes the Guava inspection check the JavaFeature
support lookup instead of being hardcoded to Java 8.

This is the same logical change that was made in commit
f995b33d86b6673a111de944aba77bce88d8c775 to various
other inspections.

Bug: https://issuetracker.google.com/118596848


akozlova:
Thanks. Merged manually

PR #911, opened 2018-11-03
removed deprecated option --recompile-scripts

dahlstrom-g:
Gradle 5.0 has deprecated --recompile-scripts
see https://github.com/gradle/gradle/issues/1425

Gradle considers that it is useless to have such flags as caching
should always be correct.


vladsoroka:
Thanks, it seems the flag was removed today by @nskvortsov, see https://github.com/JetBrains/intellij-community/commit/94009317a192c265ff232a3a4eed6ca96fde9027

PR #913, opened 2018-11-08
Fixed assigning single elements to varargs in named form.

juan-chen:
Kotlin 1.3.0 reports such cases as errors.


nskvortsov:
Thanks. Merged manually

PR #914, opened 2018-11-09
Show modules in bold on Project View

srmurguia:
When a GradleModuleDirectoryNode is shown the name of the module is
appended in boldeface, unless it is the same as the path, but since
shouldShowModuleName is overloaded to return false, this causes that the
module's directory is not shown in bold face.

This change causes the name of the module to appear in bold face when
the module name is not appended. This fixes this bug:

https://issuetracker.google.com/issues/77235561


vladsoroka:
Thanks for the contribution, however the change causes unexpected presentation of the Gradle source set modules on Project View:
Screenshot from 2019-11-25 10-46-11

I've fixed it using the following commit https://github.com/JetBrains/intellij-community/commit/f74f118ec7cd8e1da534ac5be325a377aaa8bd77

Please, let me know if it doesn't work for you.

PR #915, opened 2018-11-12
JUnit5 suggest, make reference and validate @EnumSource…

bedla:
… values from enum constants

for details see https://youtrack.jetbrains.com/issue/IDEA-202111


plugins/junit/src/com/intellij/execution/junit/codeInsight/references/JUnitReferenceContributor.java
```diff
@@ -34,6 +34,12 @@
     return PlatformPatterns.psiElement(PsiLiteral.class).and(new FilterPattern(new TestAnnotationFilter(annotation, paramName)));
   }
 
+  private static PsiElementPattern.Capture<PsiLiteral> getEnumSourceNamesPattern() {
+    return PlatformPatterns.psiElement(PsiLiteral.class)
+      .and(new FilterPattern(new TestAnnotationFilter(JUnitCommonClassNames.ORG_JUNIT_JUPITER_PARAMS_ENUM_SOURCE, "names")))
```
bedla: that EnumSource annotation is marked as experimental, but imo widly used and I tried to write code to do nothing when api changes (i hope :]) in the future

plugins/junit/src/com/intellij/execution/junit/codeInsight/enumsource/EnumSourceProblems.kt
```diff
@@ -0,0 +1,27 @@
...
+  if (supportedMode && !enumConstants.contains(name.value)) {
+    holder.registerProblem(name, "@EnumSource 'names' contains invalid value. See ${enumClass.name}'s constants.")
+  }
+}
+
+private fun enumConstants(psiClass: PsiClass): Set<String> {
+  return psiClass.fields.asSequence()
+    .filterIsInstance(PsiEnumConstant::class.java)
+    .map { it.name }
+    .toSortedSet()
```
bedla: forgot to remove sorted because first i had those enums in error msg, than changed my mind to not to have theme in error message.

akozlova:
Ivo, I think that the request is feasible. I don't like that the utility functions are available inside the module which means e.g. that you publish very specific `com.intellij.execution.junit.codeInsight.enumsource.EnumSourceUtilsKt#getParentOfType`. Moreover I am not sure that `stopAt` arguments are required: they are supposed to stop the traversal but looks like all usages are called inside annotation already and you never got those stop elements.
So my suggestion is to merge everything about problems into inspection (otherwise package `enumsource` is separated from the inspection though I hardly can imagine how it can be used outside of it) and have another one class with reference and then make all utilities private or inline them.

It's also not common to have "see smth" messages. I'd rather have "Can't resolve enum constant reference"

I can perform the suggested changes myself. What do you think?

bedla:
Anna, thx for review, I am going to fix all the issues. :)

bedla:
i have fixed all issues and amended commit. i am not sure i really understand "make all utilities private or inline them" (i did it) because it leads to duplicate code, or am i missing something? :]

akozlova:
`package com.intellij.execution.junit.codeInsight.references` still exists?

bedla:
yep it does... so you mean to move all files from .references one level up?

bedla:
Maybe I am on right track. I have refactored code a bit and hide elements in EnumSourceNamesInsight.kt file. I have moved everything from .reference pkg one level up to not call from child pkg to parent pkg - this is imo codesmell. Are you ok with it? :] Thx, Ivos

akozlova:
I think, I am ok with current structure. Let me check the details

akozlova:
Why `com.intellij.execution.junit.codeInsight.EnumSourceNamesInsight#checkEnumSourceNames` is separated from `com.intellij.psi.JavaElementVisitor#checkEnumSource`? there would be small duplicates but you won't need passing functions around, etc. I am not sure that I may insist to change kotlin code as mostly I do java but this just looks overcomplicated to me...

bedla:
I have refactored it a bit / added some comments. 

"checkEnumSource" calls two checks, so I separated "Names" check to special method. Currently I removed that recursion for "value" annotation value - it is not necessary because @EnumSource has only one "Class<?>" value.

I have created new code in Kotlin because I thought that Kotlin was primary created to have better language (than Java) for creating new features in IDEA :]

I am open for any suggestion... I can change what ever you want, If you have feeling from code that it is overcomplicated.

In general code for Filter, Reference or Inspection works as follows: First it detects "mode" default value from @EnumSource class definition (there are some versions of JUnit5 that does not specify default value for "mode"). Than it finds "mode" value from PsiAnnotation (INCLUDE or EXCLUDE means that string value is interpreted as enum constant by JUnit5, other values are interpreted as regural expression). And finally it compares "names" values to constant-fields detected on selected Enum<?>.

akozlova:
I slightly changed the code and support additional warnings on duplicated literals, also duplicated constants won't be suggested now. Merged manually. Thanks

akozlova:
Don't get me wrong, I don't like code duplicates but when it just one method call with arguments than I **personally** prefer to make the structure easier and duplicate the arguments rather than create 2 additional classes.

bedla:
I see :-) anyways it was fun investigating. Looking forward to finding some other thing i can help with PR.

akozlova:
please, submit youtrack issue and ask if PR is feasible first.

PR #917, opened 2018-11-14
Fix Shift+Tab shortcut in RegExp form (IDEA-202189)

rillig:
An AnAction must not be reused for multiple shortcuts, or all but the
last shortcut will be ignored.


BasLeijdekkers:
Merged manually with some changes as d6e2c54b1c032366a26fbf85e8e7668c072cdc37

PR #918, opened 2018-11-15
In the RegExp form, check for HIT_END

rillig:
If the regular expression does not match the current sample text, but
a longer input might match, display a hint to the user. This is helpful
in deciding whether the regular expression is completely wrong or the
sample text is just unfinished.


BasLeijdekkers:
Merged manually with some changes 227fbd2f73fcabc4d6823ed972928534cb382fff

Nice change, thanks!

BasLeijdekkers:
No more nested conditional please😉

PR #919, opened 2018-11-15
Prevent ProgressSuspender from crashing in isInImpatientReader mode.

yuriy-solodkyy:
If ProgressSuspender has installed its check cancelled hook it should
not crash on checking isReadAccessAllowed (especially when the current
thread is not under the suspended ProgressIndicator).

It happens when JobLauncher.invokeConcurrentlyUnderProgress runs a job
in isInImpatientReader mode.

Change-Id: Ib6fec3045626fd09d8ec1e38e121419b416b6d27


donnerpeter:
Merged manually, but then another change occurred and fixed `isReadAccessAllowed` throwing `PCE`, so now I'm considering reverting this change.

PR #921, opened 2018-11-16
Differentiate syntax highlighting between control and non-control keywords

raymondwdamon:
This is an idea from the TextMate grammars, in which different syntax highlighting can be customized for keywords which affect control flow and and keywords which do not.


raymondwdamon:
@klikh Yes, I submitted the CLA.

raymondwdamon:
@klikh @akozlova Pinging since it's been a month. Is there anything I can do to address concerns about this PR?

akozlova:
We are hesitating if the idea of the PR is feasible for the IDEA. Could you please explain what problem this separation solves? E.g. I see that you put `new` to control keywords though super and this are not, java `default` keyword is a control keyword though it can be used for default methods in interfaces and there it should not be a control keyword.

raymondwdamon:
The particular problem this solves is that I use different text editors and like to keep my syntax highlighting similar between them. After using TextMate grammar-based editors, I find the distinction between control and non-control keywords helpful. This PR does not change the default colors, so there is no effect for users unless they explicitly specify in settings.

You make a good point about the default keyword. I updated this PR so that:

- `default` is a non-control keyword in the lexer stage
- `default` is a control keyword after parsing
- `super` and `this` are control keywords after parsing

I do need assistance in one area, as I am not deeply familiar with the syntax highlighting mechanism. That is, there is a bug currently in the PR: e.g., if keyword font attribute has Bold selected and control keyword does not, then `default`, `super`, and `this` will show bold when highlighted as control keywords after parsing. Would you have suggestion(s) how to fix this?

akozlova:
> I find the distinction between control and non-control keywords helpful
How exactly do you use this information? This is a very important point before proceeding further, please don't skip the question.

With the PR there are even more issues than that: you choose to support java and python but IDEA supports much more languages and in order to provide a feature, it's common to support the majority of languages. PR breaks the api compatibility with some plugins, this means that the api must be adjusted. 
BTW when something is disabled by default, in many cases means that nearly nobody would use it and integrating new code will probably introduce new issues, so I strongly disagree with the idea that it won't disturb anyone and thus should be accepted.

PR #922, opened 2018-11-19
IDEA-201670 added error popup while creation setter for final field

iselimov:


akozlova:
Thanks for the PR. I slightly modified the changes and pushed to master, see 11ff07b6c5a442514ff90c4693f1181d64d90b26.

PR #924, opened 2018-11-20
Make code in UnnecessaryUnicodeEscapeInspection shorter

rillig:
The keyword "while" at the beginning of a line usually starts a while
loop instead of ending a do-while loop. To avoid further confusion,
replace the do-while loop with a typical while-loop.


BasLeijdekkers:
Merged manually as e94a1968274906a9ead110af9d7ed9b4d05a5684, thanks!

PR #994, opened 2018-11-28
Add OpenJ9 support to Windows Launchers.

allanxp4:
This is a proposed fix for https://youtrack.jetbrains.com/issue/IDEA-202981 and makes the Windows launcher comptatible with OpenJ9. The only changes are path checks for the OpenJ9 jvm.dll locations.

However, I tested it by manually copying the .exe generated by the sln, so some help to understand how the executables are generated for each version of IntelliJ would be appreciated.


koppor:
Seems to be closed, because the functionality is offered by the [Choose Runtime plugin](https://plugins.jetbrains.com/plugin/12836-choose-runtime).

(Just noted down for the googlers out there)

PR #995, opened 2018-11-28
[external-system] make IDE grouping by external project path

fkorotkov:
In the IDE grouping we want to pursue the same structure as we have projects structured on a disk. In this situation it makes sense to switch to grouping by an external project path. With the current implementation if, for example, a Gradle sub-project in a `projects/foo` folder overrides `group` in `build.gradle` like this:

```groovy
group "com.example.foo"
```

Then the current grouping will put `:project:foo` under `com.example.foo` node which is not what we want.


nskvortsov:
merged manually as 56bfaf36

PR #1000, opened 2018-12-04
Fix missing type parameter

vcaen:
FilteringListModel can have type parameter added to more of its methods
to avoid compilation error when the delegated list is not casted.


klikh:
Merged the commit. Thanks!

PR #1001, opened 2018-12-04
Move Arrays::equals to the end of equals()

stsypanov:
Small performance improvement, see new inspection request https://youtrack.jetbrains.com/issue/IDEA-170178


klikh:
Merged manually. Thanks.

PR #1002, opened 2018-12-05
Pycharm, add warning for shadowing system modules

Snyssfx:
Add warning when user's modules shadow modules of standard libraries names.
For this:
* Add python-stdlib-list helper that produce lists of standard libraries
* Add warning and warningMessage for more info about shadowing user's modules


larroy:
...

vladimir-koshelev:
Instead of introducing an inspection, we should address this issue on the debugger side.

PR #1003, opened 2018-12-05
IDEA-203651: Improve check that source root type is for tests

ausatiy:


jps/model-api/src/org/jetbrains/jps/model/java/JavaModuleSourceRootTypes.java
```diff
@@ -31,6 +31,6 @@
   public static final Set<? extends JpsModuleSourceRootType<?>> TESTS = ContainerUtilRt.newHashSet(JavaSourceRootType.TEST_SOURCE, JavaResourceRootType.TEST_RESOURCE);
 
   public static boolean isTestSourceOrResource(@Nullable JpsModuleSourceRootType<?> type) {
-    return JavaSourceRootType.TEST_SOURCE.equals(type) || JavaResourceRootType.TEST_RESOURCE.equals(type);
+    return type.isForTests();
```
chashnikov: This method is located in JavaModuleSourceRootTypes class so users won't expect that it'll return `true` for non-Java test source roots as well. It'll be better to use `isForTests` directly in the places where this method is currently called.

jps/model-api/src/org/jetbrains/jps/model/java/JavaResourceRootType.java
```diff
@@ -24,10 +24,17 @@
...
                                                                                          JpsModuleSourceRootType<JavaResourceRootProperties> {
-  public static final JavaResourceRootType RESOURCE = new JavaResourceRootType();
-  public static final JavaResourceRootType TEST_RESOURCE = new JavaResourceRootType();
+  public static final JavaResourceRootType RESOURCE = new JavaResourceRootType(false);
+  public static final JavaResourceRootType TEST_RESOURCE = new JavaResourceRootType(true);
 
-  private JavaResourceRootType() {
+  private final boolean isForTests;
+  private JavaResourceRootType(boolean isForTests) {
+    this.isForTests = isForTests;
```
chashnikov: We use `my` prefix for instance fields in Java classes to avoid such `this.` qualifiers.

jps/model-api/src/org/jetbrains/jps/model/java/JavaModuleSourceRootTypes.java
chashnikov: Also this code may fail with NullPointerException, and IDEA should show you a warning about that.

jps/model-api/src/org/jetbrains/jps/model/module/JpsModuleSourceRootType.java
```diff
@@ -28,4 +28,9 @@
  * @author nik
  */
 public interface JpsModuleSourceRootType<P extends JpsElement> extends JpsElementType<P>, JpsElementTypeWithDefaultProperties<P> {
+
+  /**
+   * Returns true if SourceRoot represents test resources, false otherwise
+   */
+  boolean isForTests();
```
chashnikov: This will break backward compatibility with plugins implementing this interface. `default` method should be used.

chashnikov:
I've fixed these problems and committed the changes myself.

chashnikov:
Implemented as [2852d1e53d1d050a](https://github.com/JetBrains/intellij-community/commit/2852d1e53d1d050ad285f85afade104ab15fcf29) and [f59206f1bc21b91](https://github.com/JetBrains/intellij-community/commit/f59206f1bc21b9179c047373afab2d0071c24d54).

PR #1005, opened 2018-12-07
[IDEA-201343] Add standard Maven properties to pom.xml of just created Maven based module

rosti-il:
Properties that are added are:
1. project.build.sourceEncoding - always "UTF-8"
2. maven.compiler.source - according to the module SDK
3. maven.compiler.target - according to the module SDK

This should resolve the IDEA-201343 improvement ticket submitted by me.


nskvortsov:
Thank you for the proposal. The original issue is fixed and verified, so I assume this PR is obsolete. Feel free to reopen if this feature is still relevant.

PR #1007, opened 2018-12-09
IDEA-203912: assertTrue and assertFalse with a negative expression can be simplified

michaelsiepmann:
This is the code for [IDEA-203912](https://youtrack.jetbrains.com/issue/IDEA-203912).

It marks the code, when using assertTrue or assertFalse with a negated expression. This assertion can be changed to the opposite method and without the negation.


klikh:
Could you please rebase the pull request so that it doesn't contain these "Merge pull request" commits? Thanks!

PR #1008, opened 2018-12-10
[IDEA-123039] Use non-recursive Xalan str:tokenize() extension function instead of the recursive XSLT 1.0 hack

rosti-il:
This fixes the IDEA-123039 bug.

For more information about Xalan EXSLT extensions go to:
https://xalan.apache.org/xalan-j/extensionslib.html#exslt

### DIFF
--- a/platform/testRunner/src/com/intellij/execution/testframework/export/intellij-export.xsl
+++ b/platform/testRunner/src/com/intellij/execution/testframework/export/intellij-export.xsl
@@ -1,5 +1,7 @@
 <?xml version="1.0"?>
-<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
+<xsl:stylesheet version="1.0"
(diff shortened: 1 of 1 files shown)


smoczyna:
that's lovely but it would be even better if you'd put add some guidelines how to do this.

rosti-il:
@smoczyna, how to do what? I don't really understand your statement.
I've provided the fix and now @akozlova from JetBrains should review and merge it.

smoczyna:
How to "use" xalan instead of other one of course. I have no clue how to do
that.

rosti-il:
@smoczyna, take a look at the changes of this pull request and stop trolling. PR contains changes in the code, not only a title that you don't understand.

smoczyna:
Look you rosti or perhaps rusty dickhead. I've been directed here by JetBrain support to get the solution of my problem which is exactly the same. However there is not any "solution" here I can use straight away, thus my comment.

akozlova:
Sorry for late response. PR is merged manually, the fix should appear in the second 2019.1 EAP

PR #1010, opened 2018-12-11
[external-system] faster module select

fkorotkov:
There is no need to explicitly set selected state for nodes. It will be taken care of by `CheckboxTree`.

I've tested it on a multi-project. Just selected/deselected node.

I found it while profiling a ~20 seconds UI freeze while operating with select project dialog.

to: @vladsoroka @nskvortsov


nskvortsov:
this code not only sets selected states for nodes, but also sets `ignored` states for all children of a data node recursively.
While other tree nodes will be taken care of by `CheckboxTree`, there are some data nodes (not used as tree nodes, but still presented in the graph) that should be ignored too

nskvortsov:

PR #1011, opened 2018-12-12
[external-system] don't always validate recursively

fkorotkov:
`checkIsSerializable` is simply trying to write the object to noop object writer. Let's first see if we can serialize at all and if not try to find a particular node and fix it.

This fixes a nasty ~20 seconds UI freeze on updating project data.

to: @vladsoroka @nskvortsov


nskvortsov:
This patch only checks if `ProjectData` stored in root `DataNode` can be serialized and does no assertions on any other nodes. Actual serialization is not performed by `externalProjectStructure.checkIsSerializable();`

nskvortsov:

PR #1013, opened 2018-12-15
Display multiline replacements in the preview (IDEA-204269)

rillig:
Fixes IDEA-108476.
Fixes IDEA-108477.


Vassiliy-Kudryashov:
@rillig your patch was adapted and pushed in master (4f9c8770), thank you!

PR #1014, opened 2018-12-16
Fix IDEA-141538: XML validator does not use a relative path to the XSD

rosti-il:
Fixes https://youtrack.jetbrains.com/issue/IDEA-141538
and probably https://youtrack.jetbrains.com/issue/IDEA-146915

### DIFF
--- a/xml/xml-psi-impl/src/com/intellij/xml/actions/validate/ValidateXmlActionHandler.java
+++ b/xml/xml-psi-impl/src/com/intellij/xml/actions/validate/ValidateXmlActionHandler.java
@@ -35,13 +35,8 @@ import org.apache.xerces.jaxp.SAXParserFactoryImpl;
 import org.apache.xerces.util.SecurityManager;
 import org.apache.xerces.util.XMLGrammarPoolImpl;
 import org.apache.xerces.xni.grammars.XMLGrammarPool;
-import org.jetbrains.annotations.NonNls;
 import org.jetbrains.annotations.Nullable;
-import org.xml.sax.InputSource;
-import org.xml.sax.SAXException;
-import org.xml.sax.SAXNotRecognizedException;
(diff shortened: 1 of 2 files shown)


rosti-il:
The original code before this fix had registered two instances of `ResourceResolver` three times: 1) `myXmlResourceResolver`; 2) `DefaultHandler` (that also implements `ResourceResolver`); 3) `myXmlResourceResolver` again by a dirty trick with `startDocument()`. My fix not only fixes the bug but also makes the code clean and only one instance `ResourceResolver` is created and register in the `myParser` only once.

rosti-il:
I accidentally closed this PR in posting my the previous comment. Reopened it now.

rosti-il:
When you debug this code you probably wants to debug the Xerces code as well. To do so with a debugger you need to make and use your own build of Xerces with `debug` property turned `on` because the official release contains JAR without any debug information including the line number information.

For discussion about why they did so go to http://markmail.org/thread/qkbaup5dzjrx5ffj

dmitry-avdeev:
Thanks @rosti-il, unfortunately this cannot be accepted as-is because it breaks some tests from XmlHighlightingTest (mainly XInclude support). Based on your main idea of providing systemId to the original InputSource I've mastered a minimal patch to be backported to 2018.3 bugfix updates. Thanks again for great job and feel free to submit corrected patch with other improvements, we'll take it into our master.

rosti-il:
@dmitry-avdeev ok. Could you advise me what is the right way to run all tests before commit? It's not clearly explained in the `README.md`.

dmitry-avdeev:
Running all tests from intellij.xml.tests would be enough.

rosti-il:
@dmitry-avdeev please take a look at my new PR that continues this one:
https://github.com/JetBrains/intellij-community/pull/1021

PR #1015, opened 2018-12-18
IDEA-204310: Decompiler adds redundant imports for lambdas

md-5:
Please see https://youtrack.jetbrains.com/issue/IDEA-204310


gorrus:
merged, thanks for contribution!

PR #1016, opened 2018-12-18
IDEA-204408 - Stop ignoring exceptions that occur during included bui…

Slvin:
…ld configuration

### DIFF
--- a/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/init/init.gradle
+++ b/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/init/init.gradle
@@ -27,14 +27,10 @@ class JetGradlePlugin implements Plugin<Gradle> {
     else {
       gradle.rootProject { project ->
         registries.add(((ProjectInternal)project).services.get(ToolingModelBuilderRegistry.class))
-        try {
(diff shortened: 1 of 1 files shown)


Slvin:
I can't see a reason as to why we would want to ignore all exceptions coming from here and I imagine it would be safe to remove. Is there tests that cover this area? I couldn't seem to find any tests that cover this specifically, everything in intellij.gradle.toolingExtension.tests seems unchanged when removing this.

nskvortsov:
Composite project import is covered by `org.jetbrains.plugins.gradle.importing.GradleCompositeImportingTest` 
Also, while having error message is indeed required, the import should continue to work on other included builds. 

What we really need, is to create a stub module for failed included build, with content root pointing to included project's root to allow a user to see files in project structure tree, to navigate to them and to fix the content, so that project can eventually be imported.
At the moment, if the included build is outside root project, files will be missing.

Slvin:
That makes sense, thanks for the explanation. In Android Studio (and Android projects in general) this causes very odd error messages which completely hide the issue. For example if a buildscript dependency fails to resolve the error we surface is "Plugin with id 'com.android.application' not found." which is really not helpful. I'm guessing this is due to later configuration outside the try { } catch { } (most probably triggered by the Android Gradle Plugin) failing.

nskvortsov:

PR #1017, opened 2018-12-19
Fix IDEA-202846: Use the Document and not the root Element to run XPath

rosti-il:
The IDEA-202846 is actually a regression introduced in the a992508f728fdefee4de267d229edf05dc300695 commit. `Element`, even root element, is not always a good chose to run xpath on it (for example if your xpath is absolute, i.e. starts with `/`) and this is why the old and correct xpath-es stopped to work properly.

This PR makes xpath-es in the `com.intellij.jarFinder.MavenCentralSourceSearcher` and in the `com.intellij.jarFinder.SonatypeSourceSearcher` classes run properly again.

Please review, merge and consider to undepricate the `JDOMUtil.loadDocument()` methods.


develar:
Thanks, I decided to fix the root cause and fix XPath expression instead of using Document again.

PR #1019, opened 2018-12-22
1st stage of fixing IDEA-204628

rosti-il:
IntelliJ IDEA can be build by other IntelliJ IDEA but not yet by the 'ant' command.
I can't figure out how to add the lib/tools.jar of JDK into the compile time classpath.
Ant calls Gradle and in Gradle you can call 'org.gradle.internal.jvm.Jvm.current().toolsJar'
to get full path into the lib/tools.jar library, but I didn't find the right place to put it in.
Standard examples of using it, found in Google, don't work here.

Hope someone will help me to figure out how to add lib/tools.jar of JDK into the compile time classpath and I will do it in the second commit in this PR.


rosti-il:
Link to the bug ticket: https://youtrack.jetbrains.com/issue/IDEA-204628

PR #1020, opened 2018-12-22
IDEA-202032 replace '/' with '.' when typing class name in the Create New Class dialog

rosti-il:
Link to the ticket: https://youtrack.jetbrains.com/issue/IDEA-202032


donnerpeter:
Thank you for the pull request! Sorry, but I can't accept it. First, the issue in question isn't going to be fixed until we see more convincing use cases. Second, even if it is, this change introduces too ad hoc API to already overdesigned code, and this way of solving the problem doesn't look right to me.

PR #1022, opened 2019-01-03
IDEA-204927 Add divider resize listener to Splitter.

jlauridsen:
Add a listener for user generated resize events of the first and
end components in a ThreeComponentsSplitter.
Hookup these resize events onto mouse release events for the
dividers.


jlauridsen:
This was fixed. Thank you.

PR #1024, opened 2019-01-05
IDEA-204998: Use decompile classes in preference to library classes

md-5:
Fixes https://youtrack.jetbrains.com/issue/IDEA-204998


trespasserw:
Merged. Sorry for this taking so long.

md-5:
Thanks @trespasserw

PR #1026, opened 2019-01-09
Fix NPE inside `ExceptionRangeCFG::toString()` for `finally` exception range.

Maccimo:
This PR fix `NullPointerException` inside `ExceptionRangeCFG::toString()` when `exceptionTypes == null`.
For ranges representing `try ... finally` block `null` is a valid value.


Maccimo:
@trespasserw Is there any reasons this PR is not merged yet?

trespasserw:
@Maccimo only lack of time. I'm terribly sorry for this taking so long.

I've omitted the test, because the code being changed is not a part of the core logic.

PR #1027, opened 2019-01-10
Fixed #IDEA-205163 performance leak

victor-cr:


victor-cr:
Yes I did

cdracm:
fixed the other (less scary) way for now

PR #1029, opened 2019-01-11
Fix a bug in JBLoadingPanel in setBounds method.

yuuki3655:
When JBLoadingPanel#setBounds is called it only updates its boundaries and its underlying LoadingDecorator panel doesn't get updated resulting an unexpected layouts.
This patch fixes it by overriding setBounds method and propagating the update to LoadingDecorator properly.

https://youtrack.jetbrains.net/issue/IJSDK-545

### DIFF
--- a/platform/platform-api/src/com/intellij/ui/components/JBLoadingPanel.java
+++ b/platform/platform-api/src/com/intellij/ui/components/JBLoadingPanel.java
@@ -122,4 +122,10 @@ public class JBLoadingPanel extends JPanel {
   public Dimension getPreferredSize() {
     return getContentPanel().getPreferredSize();
   }
+
+  @Override
+  public void setBounds(int x, int y, int width, int height) {
+    super.setBounds(x, y, width, height);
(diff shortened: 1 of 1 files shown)


Vassiliy-Kudryashov:
I commented in https://youtrack.jetbrains.net/issue/IDEA-205343:
_Cannot see the problem, setBounds() in JBLoadingPanel causes revalidation and finally calls LoadingDecorator.MyLayeredPane#doLayout
Can you give an example of noticeable "unexpected layout"?_

yuuki3655:
Sorry for leaving this PR open for a long time. I agreed with you. Calling #setBound should cause revalidation. I hit this issue when I work on ChooseAndroidProjectStep in Android plugin ([code](https://github.com/JetBrains/android/blob/master/android/src/com/android/tools/idea/npw/project/ChooseAndroidProjectStep.java#L119)). If I removed that override, I saw no loading spinner displayed on the screen. I was very new to this codebase when I was open this bug and now I think the issue was that I used BorderLayout there. Let me close this PR and make a fix in Android plugin side. Thank you!

PR #1030, opened 2019-01-13
Don't modify TestPlan

marcphilipp:
Although `TestPlan` exposes an `add(TestIdentifier)` method, it was only
ever intended for internal use by JUnit's `DefaultLauncher` which
already adds dynamic test identifiers to the `TestPlan`. The JUnit team
is going to make the `add()` throw an exception on modifications from
listeners. Thus, this commit removes the unnecessary call.

@akozlova Please have a look and let me know what you think.

### DIFF
--- a/plugins/junit5_rt/src/com/intellij/junit5/JUnit5TestExecutionListener.java
+++ b/plugins/junit5_rt/src/com/intellij/junit5/JUnit5TestExecutionListener.java
@@ -165,11 +165,6 @@ public class JUnit5TestExecutionListener implements TestExecutionListener {
   }
 
   @Override
-  public void dynamicTestRegistered(TestIdentifier testIdentifier) {
-    myTestPlan.add(testIdentifier);
-  }
-
-  @Override
   public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
     final TestExecutionResult.Status status = testExecutionResult.getStatus();
     final Throwable throwableOptional = testExecutionResult.getThrowable().orElse(null);


akozlova:
So it was added after an issue in junit 5 tracker, see https://youtrack.jetbrains.com/issue/IDEA-181642 (unfortunately I can't find corresponding junit issue). This means that for older versions of junit, IDEA would  stop to work. On the other side, we would need to publish 4 releases for older IDEA versions, 2017.3-2018.3 to support new junit 5 versions so people would be able to use it. And don't forget that people would need to download those versions, update to them, etc. This means that everyone would spend a lot of time on that. Are you sure that it's a good idea to change junit 5 this way?

marcphilipp:
Thanks for finding that issue link! 👍 

TBH I'm a bit confused. The `TestIdentifier` has always been added to the `TestPlan` by the JUnit Platform (at least since JUnit 5.0.0-ALPHA) so I'm wondering what was actually fixed by adding it again in IDEA's listener. Do you remember how you reproduced the issue?

For now, we'll emit a warning and ignore any attempt to modify the `TestPlan`. However, if there really was an issue that was fixed by calling `add()` (again), that would break it so we should really try to get to the bottom of this issue.

akozlova:
I don't remember but I think it was something about restarting parameterized tests or similar

marcphilipp:
I investigated a bit further and found out that up until 5377617c9f14e9948dbeef7f4585f190702c5f6a IDEA used the `TestPlan` returned by `discover()` to look up elements instead of the one passed to `TestExecutionListener.testPlanExecutionStarted(TestPlan)`. Thus, it was necessary to `add()` it to the `TestPlan` but now it isn't anymore.

We'll deprecate this behavior (calling `TestPlan.add()`) in JUnit but won't break it until 2018.3 is no longer supported.

@akozlova I think we should still merge this PR so IDEA no longer relies on `TestPlan.add()`.

akozlova:
Please submit the contribution agreement https://www.jetbrains.com/agreements/cla/.

marcphilipp:
Done.

akozlova:
Thanks. Merged manually

nhojpatrick:
What release should it be fixed in?
As getting the error with IntelliJ Ultimate 2018.3.5.

The code which is throwing it is;
@org.junit.jupiter.api.TestFactory
public java.util.stream.Stream<org.junit.jupiter.api.DynamicTest> tests() {
    return xxx.stream()
            .map(p -> DynamicTest.dynamicText(p, () -> {
                    yyy
            }));
}

sproshev:
@nhojpatrick in 2019.1, [EAP](https://www.jetbrains.com/idea/nextversion/) builds are already available

PR #1031, opened 2019-01-17
IDEA-205629 Shift text range when getStartTagNameElement is null

dahlstrom-g:
Apply necessary range shift both to startToken and original tag to avoid
“Argument rangeInElement … endOffset must not exceed descriptor text range”

### DIFF
--- a/xml/dom-openapi/src/com/intellij/util/xml/DomUtil.java
+++ b/xml/dom-openapi/src/com/intellij/util/xml/DomUtil.java
@@ -479,9 +479,9 @@ public class DomUtil {
   }
 
   public static Pair<TextRange, PsiElement> getProblemRange(final XmlTag tag) {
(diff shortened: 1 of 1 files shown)


davidherman-goog:
Just FYI we're starting to dig into this again. I updated the YouTrack issue with more details: https://youtrack.jetbrains.com/issue/IDEA-205629

Thanks!
David

davidherman-goog:
This was marked as fixed in YouTrack (https://youtrack.jetbrains.com/issue/IDEA-205629). It may be worth closing this issue as well. (I'm still having trouble finding the github commit that resolves the issue though)

dahlstrom-g:
@dmitry-avdeev Is IDEA-205629 fixed? What commit(s) fixed it?

dmitry-avdeev:
@dahlstrom-g It's fixed by https://github.com/JetBrains/android/commit/3a148a4c64478d4b6b529d0dd4a17430d06893b0

PR #1033, opened 2019-01-19
IDEA-205705 Include file name in tab title for pom.xml and gradle build scripts (like plugin.xml)

alshain:
Fixes IDEA-205705 Access Gradle and Maven files easier via Recent files by including file name in title

https://youtrack.jetbrains.net/issue/IDEA-205705

Same format used as in org.jetbrains.idea.devkit.dom.ide.PluginDescriptorEditorTabTitleProvider.getEditorTabTitle


alshain:
Looks like it's been incorporated/superseded with commit 123242c4

PR #1034, opened 2019-01-20
IDEA-189773 Action to copy about info to clipboard

alshain:
https://youtrack.jetbrains.net/issue/IDEA-189773


bulenkov:
Declined. See https://youtrack.jetbrains.net/issue/IDEA-189773

PR #1035, opened 2019-01-20
Allow more variants in Go to Line/Column dialog (IDEA-205713)

rillig:


JB-Dmitry:
In this variant, newly added abilities are absolutely non-discoverable. We try to avoid such implementations. Something in the UI should suggest there's an option to address lines in a 'relative' manner. Maybe the feature can be restricted to only '+/-number' case, then the its description can probably be added to the displayed dialog.

rillig:
Agreed. I didn't miss this advanced relative navigation often, and documenting it exhaustively would probably be too much effort.

PR #1036, opened 2019-01-22
Remove dead code from DebugProcessImpl

amaurymedeiros:
We don't need to check if we're in unit test mode, as we're going to return
executionResult anyway. There used to be a block of code (now commented) to
be executed in case we were not testing but it has not been changed in the
past several years, so it's unlikely it will ever be uncommented. This
commit removes this commented block as well.


gorrus:
pushed, thanks!

PR #1038, opened 2019-01-23
Re-enable canonicalization cache

diegopl:
This CL still keeps the canonicalization prefix cache disabled since
that is known to have consistency issues.
The canonicalization cache helps to address performance issues on the
Windows platform where the File.getCanonicalPath is significantly slower
than in other platforms.

As per my conversation with @trespasserw, this was disabled back in Apple JDK6 since it didn't handle correctly symlinks and kept re-indexing parts of the JDK. I've tested this locally and it seems to work correctly now.

### DIFF
--- a/platform/build-scripts/groovy/org/jetbrains/intellij/build/impl/TestingTasksImpl.groovy
+++ b/platform/build-scripts/groovy/org/jetbrains/intellij/build/impl/TestingTasksImpl.groovy
@@ -291,7 +291,7 @@ class TestingTasksImpl extends TestingTasks {
       jvmArgs.addAll([
         "-Xmx750m",
         "-Xms750m",
-        "-Dsun.io.useCanonCaches=false"
+        "-Dsun.io.useCanonPrefixCache=false"
       ])
     }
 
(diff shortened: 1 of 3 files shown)


trespasserw:
Sorry, not yet :( Thank you for pinging, will try to look at it tomorrow.

yole:
How exactly would I test this change?

trespasserw:
@yole look for VFS events for .jar files under the project JDK root (IIRC they were happening on a project open).

diegopl:
@yole Did you have chance to look into this?

Thanks!

trespasserw:
I’m terribly sorry for this to take so long. Started looking.

trespasserw:
It seems to me that the problem was caused by a combination of two factors: how IDEA's main class loader work and how Java SDK class paths are organized. Since IDEA no longer runs on Java 6 and class paths organization was changed since then, the problem is no longer actual.

In my tests, I didn't see a difference between `sun.io.useCanonPrefixCache=false` and `sun.io.useCanonPrefixCache=true` - so maybe drop the option completely and go with the default?

diegopl:
Hey @trespasserw, thanks for looking into it. The only reason why I used useCanonPrefixCache=false is that it is known to have consistency issues and that's partly the reason why it will be removed in the future.
An example is https://bugs.java.com/bugdatabase/view_bug.do?bug_id=7066948

I wanted to stay on the safe side.

trespasserw:
Okay, makes sense.
Merged, should be available on GH within an hour.
Thanks!

trespasserw:
One problem, though. I just noticed that the cache is [disabled by default](https://bugs.openjdk.java.net/browse/JDK-8207005) in Java 12.

Are there any particular places in the code where a slowness of `File.getCanonicalPath` is particularly noticeable?

trespasserw:
Had to revert the change - the cache is still too unreliable and now the effect is observable in IDEA tests. Thought  about having different settings for Windows and Unix, but then decided against (first, it would complicate the things, second, the platform is going to migrate to NIO2 API anyway).

PR #660, opened 2017-11-10
[IDEA-169111] load fish functions on init

vcfvct:
More detail on this issue: 
https://youtrack.jetbrains.com/issue/IDEA-169111

PR #1048, opened 2019-01-26
Add support for Fish completions in terminal

nazar-pc:
Just like functions from #660, `.config/fish` can contain completions (I have them for `docker-compose` for instance).
This trivial change adds support for them into terminal plugin.


segrey:
@nazar-pc Thanks for the PR. The fix is done in d5cd457615639f982adc95658774bdc931f2a143. ETA: 2023.1

nazar-pc:
Better late then never :slightly_smiling_face:

PR #1049, opened 2019-01-29
Make Gradle's ProjectImportAction more flexible

Slvin:
This change gives more flexibility providers of GradleProjectResolverExtensions to inject logic when requesting models from Gradle during the ProjectImportAction. We allow extensions to provide a ProjectImportExtraModelProvider instead of a set list of classes, this provider has two methods which are both run within the build action. Each method provides a consumer-like interface for the provider to register the models that they are interested in so they are visible to the resolver context during setup.
</p>
The motivation behind this is part of an ongoing effort to unify the Sync infrastructure in Android Studio with the External System pipeline in Intellij. In Android Studio we only request models for the specified variant that is needed by the IDE, in order to pass this information to Gradle we need to be able to do two things:
<ul>
<li>Use the new parameterized model builder API to tell the model builder not to build variant information</li>
<li>The ability to alter model requests based on information contained in previously obtained models</li>
</ul>
</p>
This change attempts to provide both of these things to GradleProjectResolverExtensions.

I've ran any tests that I could that look like they might be affected. E.g in intellij.gradle.tests the results are consistent with or without the change (though both end up with ~100 failures). Any instruction on how best to test the new functionality would be appreciated.

Thanks!

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/AbstractProjectResolverExtension.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/AbstractProjectResolverExtension.java
@@ -31,6 +31,7 @@ import org.gradle.tooling.model.idea.IdeaProject;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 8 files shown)


Slvin:
Thanks for the quick review!

PR #1050, opened 2019-01-30
Rename Co_verage to Coverage

emirb:
Right-click on a folder shows "Run with Coverage" option:
image
and is mentioned on documentation page: https://www.jetbrains.com/help/idea/running-test-with-coverage.html

However, right click on a file gives an option that looks like a typo:
image


AMPivovarov:
The issue here is that "_" character is used to set action mnemonics.
When file name is added to the action text, "_" from it was used as mnemonic marker instead (see "configtest.go instead of "config_test.go" on your screenshot).

So the proper fix, probably, should be to fix escaping when creating this action.
See https://github.com/JetBrains/intellij-community/blob/master/plugins/coverage-common/src/com/intellij/coverage/CoverageExecutor.java#L29

AMPivovarov:
I've pushed fix into master.

emirb:
Thanks @AMPivovarov.

PR #1053, opened 2019-02-06
IJSDK-563 Fix incorrect rendering of copyable labels

dahlstrom-g:
By default JBLabels will autowrap its text contents when the container
is not wide enough to render the whole text on a single line.  In
copyable mode however, the JBLabel will render "..." if  the container
isn't wide enough when autowrapping is set to false.  This lead to error
messages in the Edit Configuration window (a very narrow window) to
render as "..." with no obvious clue as to what's wrong.

This commit fixes the bug in JBLabel to respond properly to autowrapping
and adds autowrapping to the JBLabel in Edit Configuration window.


Vassiliy-Kudryashov:
Thank you!

PR #1054, opened 2019-02-08
IDEA-206928: Show "plugin.xml (pluginName)" tab title irrespective of project type

alshain:
Implements https://youtrack.jetbrains.net/issue/IDEA-206928


YannCebron:
Thanks!

PR #1055, opened 2019-02-08
IDEA-206930 Don't complain when renaming branch to itself

alshain:
Implements https://youtrack.jetbrains.net/issue/IDEA-206930


aronzhina:
Hi @alshain, apologies for the long delay in getting back to you on this PR, and thank you for your contribution! We’ve recently changed the behavior in this area, so could you please check if the problem is still relevant in the latest IDE build?

PR #1056, opened 2019-02-08
IDEA-201941 Add countdown to "resolve changelist conflict" dialog to prevent accidental confirmations via the keyboard

alshain:
Fixes https://youtrack.jetbrains.net/issue/IDEA-201941


klikh:
Thanks for the idea and for the contribution. Unfortunately, have to decline the PR for the reasons mentioned in the mentioned YouTrack issue.

PR #1060, opened 2019-02-08
IDEA-201693 New SE: popup position should be relative to IDE, not to screen

alshain:
Fixes https://youtrack.jetbrains.net/issue/IDEA-201693


lppedd:
This should be merged. On 32" inch is a pain having the SE popup not inside the IDE window.

PR #1062, opened 2019-02-11
IDEA-207106 Log deprecation notices for uses of registerFileType() instead of FileTypeFactory

alshain:
Implements https://youtrack.jetbrains.net/issue/IDEA-207106


chashnikov:

PR #1063, opened 2019-02-13
Use String::join instead of hand-written joining pattern

stsypanov:
subj

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/wm/impl/status/TextPanel.java
+++ b/platform/platform-impl/src/com/intellij/openapi/wm/impl/status/TextPanel.java
@@ -107,15 +107,7 @@ public class TextPanel extends JComponent implements Accessible {
 
     final String[] lines = UIUtil.splitText(text, fontMetrics, widthLimit, ' ');
 
-    final StringBuilder result = new StringBuilder();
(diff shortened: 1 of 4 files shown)


platform/platform-impl/src/com/intellij/openapi/wm/impl/status/TextPanel.java
```diff
@@ -107,15 +107,7 @@ private static String splitText(final JLabel label, final String text, final int
...
-    final StringBuilder result = new StringBuilder();
-    for (int i = 0; i < lines.length; i++) {
-      final String line = lines[i];
-      if (i > 0) {
-        result.append('\n');
-      }
-      result.append(line);
-    }
-    return result.toString();
+    return String.join("\n", lines);
```
amaembo: Note that this private method was completely unused for a long time. I removed it in the subsequent commit.

amaembo:
Merged, thanks.

platform/platform-impl/src/com/intellij/openapi/wm/impl/status/TextPanel.java
stsypanov: Right, it was highlighted as unused, however I thought it's too risky ro remove it

PR #1064, opened 2019-02-15
IDEA-207347 Allow a JTextComponent to avoid a swing undo manager.

jlauridsen:
The base class: UndoRedoAction calculates the undo manager based
on the class of the component. If the component is a JTextComponent
(that is not a main text editor) the undo manager will use the
Swing undo manager.
There are cases were this is unwanted.

This CL adds a ClientProperty on the JTextComponent such that the
Swing undo manager will not be used.

Change-Id: I3e8d7a6681e51d30bc9f6052b3e74d7467fcca1c


jlauridsen:
I created a new pull request for this bug with a different implementation.

Vassiliy-Kudryashov:
The patch was applied with minor improvements. Thank you!

PR #1066, opened 2019-02-17
Maven: Import folders from multiple executions

FloGa:
With maven-build-helper, one can define multiple executions that can
each add source or resource folders to the project. With the current
implementation, IntelliJ only considers the very first execution that
matches the given maven goal, like add-sources. However, it is perfectly
fine with Maven to define multiple executions with the same goal. This
also happens automatically, for example when multiple Maven profiles are
defined and active, that each call add-source of the maven-build-helper.

With this commit, I changed the folder importing strategy, so that not
only the first matching execution is considered, but all executions that
call the appropriate Maven goal.

I also added a test case that fails on current master, but passes with
this commit. It simulates a project structure with multiple additional
sources and resources folders, with each of them added in different
executions of the maven-build-helper. This can be a common scenario if
one extensively uses Maven profiles to trigger specific actions.

### DIFF
--- a/plugins/maven/src/main/java/org/jetbrains/idea/maven/importing/MavenFoldersImporter.java
+++ b/plugins/maven/src/main/java/org/jetbrains/idea/maven/importing/MavenFoldersImporter.java
@@ -38,6 +38,7 @@ import com.intellij.util.containers.MultiMap;
 import com.intellij.util.containers.NotNullList;
 import org.jdom.Element;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.idea.maven.model.MavenPlugin;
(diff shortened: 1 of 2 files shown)


FloGa:
Did anyone already have a chance to take a look at this? The company I work for would very much appreciate to see this patch integrated into IntelliJ Ultimate. We rely heavily on conditionally generated source folders which have to be manually marked by the developers because of this issue.

vladsoroka:
Thanks for the contribution Florian! 
I've merged the PR using the following commit https://github.com/JetBrains/intellij-community/commit/3f555f9623579759ca0247b38227e964dbbfd413
After internal review it can be added for 2019.1.x.

PR #1067, opened 2019-02-21
IDEA-207657 Fix StringIndexOutOfBoundsException in DocumentWindowImpl#replaceString

dahlstrom-g:


dmitry-avdeev:
IMHO it would be better to fix the intention, not to suppress it.

kainan1112:
Thanks for reviewing, I'd like to add a little more comment about this issue. 
Cause of the issue:
1. CachedIntentions add IntentionActionWithTextCaching for both the xml and injected DataBinding file.
2. ShowIntentionActionsHandler prefers to use injected file (see ShowIntentionActionsHandler::chooseFileForAction)
3. The intention registered in XmlErrorQuickFixProvider replace the element with offset from the xml file and cause the error.

By fix the intention, do you mean:
1. disable the invoke and implement the fix for injected file separately
        @Override
        public void invoke(@NotNull Project project, Editor editor, PsiFile file) {
          if (element.getLanguage().equals(file.getLanguage())) {
            final int textOffset = element.getTextOffset();
            editor.getDocument().replaceString(textOffset, textOffset + 1, AMP_ENTITY);
          }
        }
2. Adjust the offset of the element with its language 

Or what's your proposal for fixing the bug?

dmitry-avdeev:
I've created a proper fix for this: https://github.com/JetBrains/intellij-community/commit/d881375cffe6251d22746c0141f159380478ab2f

kainan1112:
Thanks

PR #1068, opened 2019-02-23
Changes the behavior of the package creation dialog

kirpichik:
Creating multiple packages in the current IDEA version required specific efforts.

To create packages, for example, `com.example.project.sample` and `com.example.project.test`, there are several options:

1. Create `com.example.project`, create a temporary file in it, create `* .sample` and `* .test`, delete a temporary file
2. Create packages from `src` completely prescribing the name
3. On the `Project` tab, turn off the `Compact Middle Package` and create a package where necessary

As a correction of this situation, I suggest that when creating a package in the `initialText` of dialog, set the name of the package on which the action of creating a new package was performed.
For example:

To implement this functionality, I had to split the validator of the entered text in the old code of the `CreateDirectoryOrPackageAction` action, which was previously used simultaneously to create a directory and to create a package. Saving the `CreateDirectoryOrPackageAction` class allowed maintaining compatibility with all dependent code, and `CreateDirectoryOrPackageHandler` is still used to create the directory. Actions are delegated to it by `CreateDirectoryHandler`.

I think it’s worth thinking about split actions handler itself into two kinds, but this can disrupt the work of a lot of old code. And it may also be worth adding drop-down hints when entering the names of existing packages.


klikh:

kirpichik:
Ok, I submitted the contributor license agreement.

donnerpeter:
Thanks! It looks like this addresses the first part of https://youtrack.jetbrains.com/issue/IDEA-182029.

donnerpeter:
Merged into master with minor changes

donnerpeter:
People complain: https://youtrack.jetbrains.com/issue/IDEA-209397

PR #1069, opened 2019-02-27
IDEA-137065: Support escape sequences for 8-bit and 24-bit colors in console window

mchernyavsky:
Currently IntelliJ Platform supports only 16 ANSI colors (standard colors and high intensity colors). The old ColoredOutputTypeRegistry instance simply ignores 8-bit and 24-bit ANSI color escapes. In this PR ColoredOutputTypeRegistry replaces unsupported 8-bit and 24-bit colors with nearest (in Euclidean space) supported ANSI colors. Note that user can configure color mapping in editor settings (`Preferences | Editor | Console Scheme | Console Colors | ANSI Colors`). In addition, some themes also set the colors. So, this solution gives us interoperability with existing themes.

Here is the simple Python script that uses all the 8-bit and 24-bit escape sequences: https://gist.github.com/mgedmin/2762225
The output of this script is shown below:

* Embedded local terminal:
Light Dark

* Console window without this fix (just ignores the 8-bit and 24-bit escape sequences):
Light Dark

* Console window with this fix (uses the nearest supported color):
Light Dark

Also you can check the original PR from Rust plugin for more information - https://github.com/intellij-rust/intellij-rust/pull/3312.

### DIFF
--- a/platform/platform-api/src/com/intellij/execution/process/ColoredOutputTypeRegistry.java
+++ b/platform/platform-api/src/com/intellij/execution/process/ColoredOutputTypeRegistry.java
@@ -30,7 +30,8 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.awt.*;
-import java.util.Map;
+import java.util.*;
+import java.util.stream.IntStream;
 
 /**
  * @author yole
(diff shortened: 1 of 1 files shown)


piotr-dobrogost:
*Currently IntelliJ Platform supports only 16 ANSI colors* – do you know where does this restriction come from?

mchernyavsky:
> _Currently IntelliJ Platform supports only 16 ANSI colors_ – do you know where does this restriction come from?

@piotr-dobrogost No, but I think the reason is that we want to give the user the ability to customize the console colors, but now it's impossible to do, because new colors will break the existing color schemes.

piotr-dobrogost:
Well, I don't see any conflict in supporting any number of colors and allowing to customize the chosen 16 of them.

mchernyavsky:
@piotr-dobrogost For example, a program may explicitly use white (255;255;255) for text. In this case, the user of the console with a white background can change the text color to black. But if the program uses an “almost” white color (e.g. 254;254;254), the user will not be able to configure the color. It looks like a problem.

Grief:
@mchernyavsky @piotr-dobrogost I don't think that colors from 256-palette or especially true colors must be customizable. While this is possible, I don't think that it can be useful in any imaginable scenario. The corner case of black on black or white on white can be work-arounded by the addition of the slider in the settings which will specify the minimal HSV distance between foreground and background colors and in the case when the actual distance is less than specified, colors are forcibly moved apart from each other until the distance is equal to the specified in the settings. 

However, this is not required as well: almost any terminal application, like konsole or gnome shell simply displays the color specified by the application and the problem of matching background and foreground treats as application's problem, not terminal's problem.

Grief:
Also, personally I'd prefer usage of the exact color, not the nearest one, since otherwise any gradient would become invisible to the user.

mchernyavsky:
Ok, I'm closing this pull request, since it's [already fixed](https://youtrack.jetbrains.com/issue/IDEA-137065#focus=streamItem-27-3320694.0-0).

PR #1073, opened 2019-03-01
Check that build view is not null before sending events

srmurguia:
There is a NPE when multiple projects are open in Android Studio, the
reason is that if there are multiple build views, the map that holds
them can be updated by other threads. Every time this map::get method is
called in MultipleBuildsView.java the code checks that the value is not
null, except for a case.

This change also makes sure that the view is not null in the missing
part.


dahlstrom-g:
@vladsoroka Can you attend to this?

PR #1074, opened 2019-03-05
Decrement the current index after removing first element from buffer

AmrAfifiy:
There is an issue in Android Studio where the build output parsers are
not receiving all errors. The reason to that is when we try to read the
next line while the buffer is full, the first element is removed and the
current index will be pointing to the next element so this line will be
skipped in the next call of readLine.

This change decrements the current index after the first element is
removed from the buffer so it will point to the same element before
the removal.

### DIFF
--- a/xml/dom-tests/tests/com/intellij/util/xml/stubs/DomStubBuilderTest.java
+++ b/xml/dom-tests/tests/com/intellij/util/xml/stubs/DomStubBuilderTest.java
@@ -107,6 +107,11 @@ public class DomStubBuilderTest extends DomStubTest {
   }
 
   public void testInclusion() {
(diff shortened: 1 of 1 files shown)


vladsoroka:
Thanks for the contribution, I've added some tests for the reader and it finds additional issues with or w/o the change. You can find the following commits fixing found issues:
https://github.com/JetBrains/intellij-community/commit/e8c6f169afc99ae33f860f5ce792379c1dc03be9
https://github.com/JetBrains/intellij-community/commit/fbef901fa5e92867747dfc9a4c570f8d43f7a2b3

Please, verify if it fixes issue with your `BuildOutputParser`s and let me know if it doesn't.

PR #1075, opened 2019-03-05
IDEA-208450: Fix a memory leak

sprigogin:
Objects that inherit from RootModelComponentBase get registered in CompositeDisposable of RootModelImpl. These objects are often disposed much sooner than RootModelImpl with methods like RootModelImpl.removeOrderEntry or RootModelImpl.removeContentEntry which explicitly call dispose on these objects.

This change removes objects extending RootModelComponentBase from the CompositeDisposable of RootModelImpl.


chashnikov:

PR #1076, opened 2019-03-06
IDEA-208453: Fix a memory leak in GitXmlRpcHandlerService

sprigogin:
GitXmlRpcHandlerService.registerHandler was registering anonymous Disposable objects with the Disposer. Since the parentDisposable passed to the registerHandler method is always a Project, these anonymous Disposables accumulated in the Disposer's tree for the life time of the project creating an memory leak.

This change replaces anonymous Disposables with instances of the HandlerWrapper class that are disposed either by the unregisterHandler method or together with the project, whichever comes first.


AMPivovarov:
The disposable is no longer created after 509fead23a6650c3481b2ca987e96880e9a19024.
This, probably, makes this fix obsolete?

sprigogin:
Obsolete indeed. Sorry for the noise.

AMPivovarov:
Thanks for the report anyway!

PR #1077, opened 2019-03-06
Add missing break statement (see IDEA-208467)

stsypanov:
This raise-a-flag pattern is very wide-spread and often implemented without break statement after flag is raised.
In case `if`-statement doesn't have any side-effects all iterations executed after the flag is raised are useless, as flag's value won't change.
The statement or iteration however are often time-consuming which impacts performance.


amaembo:
Merged, thanks.

PR #1078, opened 2019-03-07
decompiler: Recognize enum as keyword on Java 5

TheMrMilchmann:
The current code recognizes `enum` as keyword on versions **greater than 5**. However, enums and the keyword have been introduced in Java 5. This PR simply changes the check to check whether a given version is **greater than or equal to 5**.


trespasserw:
Thanks!

PR #1079, opened 2019-03-08
IDEA-208610: Fix memory leaks in MouseDragHelper.attach and MouseDrag…

sprigogin:
Fix memory leaks in MouseDragHelper.attach and MouseDragHelper.start.

Make the detach method dispose the object added to the Disposer's tree by the attach method.

Make the stop method dispose the object added to the Disposer's tree by the start method.


Vassiliy-Kudryashov:
I believe it's better to avoid Disposable at all: add and remove listener in showNotify() and hideNotify() (revision 3dd717a4)

PR #1081, opened 2019-03-08
Ensure GradleProjectResolver uses extensions extra model provider

Slvin:
I missed this when submitting the last pull request, apologies. This ensures that the providers are taken correctly from the GradleProjectResolverExtensions.

This change should not affect the behavior since currently no resolver extensions override this method, the default implementation in AbsractProjectResolverExtension uses the ClassSetProjectImportExtraModelProvider which is exactly what this is currently doing.


vladsoroka:
Merged using the following commit https://github.com/JetBrains/intellij-community/commit/9d24211d3a3ce3f67cf8afb92f6c2ee2ac273919. 
Thanks for the contribution!

PR #1086, opened 2019-03-12
ActionButton component should be focusable.

vcaen:
Keyboard navigation can be needed not only for user using a screen
reader but also for user with difficulties using a mouse.

Enabling focus on ActionButton allows a better focus traversal policy.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/actionSystem/impl/ActionButton.java
+++ b/platform/platform-impl/src/com/intellij/openapi/actionSystem/impl/ActionButton.java
@@ -61,8 +61,7 @@ public class ActionButton extends JComponent implements ActionButtonComponent, A
     myAction = action;
     myPresentation = presentation;
     myPlace = place;
-    // Button should be focusable if screen reader is active
-    setFocusable(ScreenReader.isActive());
+    setFocusable(true);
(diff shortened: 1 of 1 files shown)


vcaen:
@rpaquay You are the one adding the screen reader condition. What do you think about this change?

rpaquay:
I think it is ok. AFAIK, this is the behavior I have seen in most app I have tried (e.g. Visual Studio).

The reason I made this conditional in the first place was to avoid friction when submitting PRs, i.e. I did not want to enter into a long argument about whether it is the right thing to do or not for all users. I was focused on improving keyboard navigation for a11y users.

FYI, there are many other conditional code paths like this.

vcaen:
Well I'm definitely not an a11y expert but I agree that this is a common behavior in most app.
Unless the guideline is that all actions should be accessible via the Search Action Popup.

vcaen:
@Vassiliy-Kudryashov Do you have any update on the PR?

Vassiliy-Kudryashov:
This simple change won't help in full so it cannot be 'just applied'. The problem  "how to provide global  focus traversal policy" is not solved yet, unfortunately focus may stuck here and there when you jump with <tab> key. Thus, some buttons stay unreachable even if they're focusable and shortcuts are really helpful when you're fond of keyboard.

PR #1087, opened 2019-03-12
IDEA-207347 Allow a JTextComponent to avoid a swing undo manager.

jlauridsen:
The base class: UndoRedoAction calculates the undo manager based
on the class of the component. If the component is a JTextComponent
(that is not a main text editor) the undo manager will use the
Swing undo manager.

In the property editor in Android Studio this is unwanted.
Here we want to be able to undo events from the IJ undo manager
when a JTextComponent has focus and has no pending swing undo
operations.

This CL adds a ClientProperty on the JTextComponent such that the
Swing undo manager will only be used if there are pending undo
operations. If not the Intellij undo manager is allowed to undo
or redo the latest recorded undo/redo event.

Change-Id: I3e8d7a6681e51d30bc9f6052b3e74d7467fcca1c


jlauridsen:
This was fixed. Thank you.

PR #1088, opened 2019-03-13
Fix a memory leak in InternalDecorator (IDEA-208847)

sprigogin:
The leak is prevented by not adding a lambda object as a child of project to the Disposer's tree and calling removeAncestorListener from the InternalDecorator.dispose method instead.


Vassiliy-Kudryashov:
Disposer.register(myHeader /* instead of myProject */, () -> removeAncestorListener(ancestorListener)); solves the issue

PR #1089, opened 2019-03-13
Surface error nodes on top of build steps

vcaen:
Hi @vladsoroka, 

This is the PR for the work we discussed on the BOW.
Instead of attaching nodes to the root, I created an intermediate build step node. If an error group with a failure is created, it will be attached to the root, before the build steps.

Feel free to ping me if you have another idea. 
Thanks,
Vadim


vcaen:
I'm having one issue though, which is that I wanted the first error lead node to be expanded but it does not seem to work.

PR #1090, opened 2019-03-13
IDEA-208913 : better @AutoValue support in nullability inspection

willemv:
IDEA-208913 : use annotations from ConcurrencyAnnotationsManager iso hardcoded string, and add @AutoValue to the list of known immutable annotations

### DIFF
--- a/java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/value/DfaExpressionFactory.java
+++ b/java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/value/DfaExpressionFactory.java
@@ -16,6 +16,7 @@
 package com.intellij.codeInspection.dataFlow.value;
 
 import com.intellij.codeInsight.AnnotationUtil;
+import com.intellij.codeInsight.ConcurrencyAnnotationsManager;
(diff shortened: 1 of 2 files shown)


java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/value/DfaExpressionFactory.java
```diff
@@ -242,6 +242,16 @@ public static VariableDescriptor getAccessedVariableOrGetter(final PsiElement ta
     return null;
   }
 
+  private static boolean isClassAnnotatedImmutable(PsiMethod method) {
+    List<String> annotations = ConcurrencyAnnotationsManager.getInstance(method.getProject()).getImmutableAnnotations();
+    for (String annotation : annotations) {
+      if (AnnotationUtil.findAnnotation(method.getContainingClass(), annotation) != null) {
```
donnerpeter: There's findAnnotation overload that takes a list

donnerpeter:
Could you please add some tests demonstrating how things should work with AutoValue?

donnerpeter:
Merged manually into master after minor corrections and adding a test. Thanks!

PR #1092, opened 2019-03-16
Enabling cross plugin communication.

cyclic-reference:
### Change Summary
Added a new `Topic` instance and interface which is a `EventListener` subclass. 

### Reason for Change
I maintain a couple of plugins and would like to enable communications between the two products. So I can implement new and cool features! 

Currently, plugins can only can publish and receive messages from a topic whose instance they can reference. In order for plugins to have access to a shared instance, it needs to come from the intellij SDK. Since all JetBrains products contain the `com.intellij.modules.platform`, I figured that this would be the best place to put this.

**Thanks for your time!**

### DIFF
--- a/platform/util/src/com/intellij/util/PluginCommunicationsListener.java
+++ b/platform/util/src/com/intellij/util/PluginCommunicationsListener.java
@@ -1,0 +1,17 @@
+// Copyright 2000-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package com.intellij.util;
+
+import com.intellij.util.messages.Topic;
+
+import java.util.EventListener;
+
+/**
+ * Enables cross plugin communications.
+ * Plugins may publish and subscribe to the topic instance provided with this interface.
+ *
+ */
+public interface PluginCommunicationsListener extends EventListener {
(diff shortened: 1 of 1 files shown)


chashnikov:
This API looks inconvenient for me: it's stringly typed and you need to encode message parameters in a single string. I think it would be better to use simple method calls instead: if you want to send a message from plugin A to plugin B, you may declare a application/project-level [service](https://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_services.html) in plugin B, define a method in it, add a dependency on plugin B to plugin A and call that method from plugin A. This way you may use parameters of any types and the java compiler checks that you pass all required parameters with proper types to the called method.

If you want to make plugin A work without plugin B installed, you may extract the code which sends the message to a separate service/extension and declare it in a separate configuration file included as  [an optional dependency](https://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_dependencies.html#optional-plugin-dependencies).

chashnikov:
If you have use cases there the way suggested by me doesn't work properly, let's discuss them.

cyclic-reference:
That should work. However, I wanted to avoid having to bring in outside dependencies for plugins to work with each other. That way I have to maintain less libraries. If having plugins share a common jar and use the service class supplied in the jar facilitate cross plugin communication works, then that is what I will do. Thanks for your time!

PR #1093, opened 2019-03-20
IDEA-208067: Fix memory leak via InlayModelImpl$AfterLineEndElementTree

npaige:
AfterLineEndElementTrees register themselves as document listeners in
their constructor, but the disposal call in InlayModelImpl which removes
this listener was omitted.


JB-Dmitry:
Patch has been pushed to master. It will also be merged into 191 branch.

JB-Dmitry:
Sorry forgot to set 'Author' field, so the commit doesn't have your name in it :(
(We apply pull requests manually in our repository).

PR #1094, opened 2019-03-20
Reduce memory used for comparing branches (IDEA-209343)

sprigogin:
Memory consumption is reduced by not creating duplicate String objects. In my testing it reduced amount of memory occupied by the String objects used by the branch comparison by 11x. Opening the comparison dialog became faster by 13%, probably due to reduced memory pressure.

Deduplication of strings is done using a new GitStringInterner class that contains a global string cache. The cache is cleared at the end of the GitBranchWorker.loadCommitsToCompare method. It would be nice to use a local cache and pass it explicitly to all GitLogParser instances created by the branch comparison code, but that would require significant changes to existing APIs.


juliabeliaeva:
Closing due to reasons described in the corresponding issue: https://youtrack.jetbrains.com/issue/IDEA-209343#focus=streamItem-27-3351040.0-0.

PR #1096, opened 2019-03-22
Update inotify.c

lots0logs:
Fix bug that caused crash when encountering a hardlink or bind mounted file.

### DIFF
--- a/native/fsNotifier/linux/inotify.c
+++ b/native/fsNotifier/linux/inotify.c
@@ -155,7 +155,6 @@ static int add_watch(int path_len, watch_node* parent) {
       // This is probably a hard link to another file we are watching. Having identical inode is expected.
       char buf1[PATH_MAX], buf2[PATH_MAX];
       const char* normalized1 = realpath(node->path, buf1);
-      const char* normalized2 = realpath(path_buf, buf2);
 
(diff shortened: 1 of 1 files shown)


trespasserw:
Bind mounts and hard links make the same physical bytes to appear via several FS paths. With this fix, `fsnotifier` would report only one of them as changed, and this may cause hard-to-debug sync/overwrite issues.

Sorry, can't accept.

lots0logs:
@trespasserw I figured what I did was not complete. The code needs to ignore cases where a duplicate inode id is returned from inotify because that is not an error. It does that when there are hard linked or bind mounted files. This behavior is in the inotify documentation btw. Are you going to fix it properly?

trespasserw:
The change could be a-okay in the context of the `fsnotifier` binary, but the latter doesn't work in isolation. For an IDE, bind-mounted or hard-linked files 'A' and 'B' are two independent FS entries. When only one of them is reported as changed, it may lead to an outdated content in the editor and eventual data loss.

One day I'm going to fix this of course, but it's not as simple as it may seem.

JamesXNelson:
Any idea what day that might be?  I have tooling that can mount all my transient build folders to tmpfs so they live in RAM, but with IJ unable to handle bind mounts, I can't readily use it w/out a million tmpfs mounts. :-/

PR #1097, opened 2019-03-22
Do not throw exception when unable to get constant pool

kunshenks:
Bug: https://issuetracker.google.com/issues/123863053
It is becuase ART(Android) JVM doesn't support the 'getConstantPool' capability.

### DIFF
--- a/java/debugger/impl/src/com/intellij/debugger/jdi/MethodBytecodeUtil.java
+++ b/java/debugger/impl/src/com/intellij/debugger/jdi/MethodBytecodeUtil.java
@@ -292,6 +292,11 @@ public class MethodBytecodeUtil {
       return locations;
     }
 
+    // Android Studio: 'getConstantPool' is not supported on android. (b/123863053)
+    if (!method.declaringType().virtualMachine().canGetConstantPool()) {
+      return locations;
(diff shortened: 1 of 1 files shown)


gorrus:
manually merged, thanks for your contribution!

java/debugger/impl/src/com/intellij/debugger/jdi/MethodBytecodeUtil.java
```diff
@@ -292,6 +292,11 @@ public void visitMethodInsn(int opcode, String owner, String name, String desc,
       return locations;
     }
 
+    // Android Studio: 'getConstantPool' is not supported on android. (b/123863053)
```
dahlstrom-g: There shouldn't be an “Android Studio” comment here.

dahlstrom-g: Also “b/123863053” is a Google-internal URL that shouldn't be here.

gorrus: Thanks, I removed the comment

PR #1098, opened 2019-03-25
IDEA-168007 Make icon filters aware of CachedImageIcon

jgaillard85:
Use IconLoader.filterIcon when filtering icons without providing a
Graphics object, as that still allows the filterd icons to be HiDPI
aware.


forantar:
Thank you, applied! fea6f64

PR #1099, opened 2019-03-25
IDEA-188964 prefer an older but compatible plugin

rsauciuc:
We expand on the fix for IDEA-188964 from commit 11b64c4617, by also
taking into account the possibility that the first plugin version
(loaded from the user's config) is newer but incompatible, while the
version bundled with the platform is compatible.

This is a recurring problem for Studio and the Kotlin plugin, whenever
we update the platform: http://issuetracker.google.com/129260858


rsauciuc:
@trespasserw

trespasserw:
Tried to collapse both cases into a single condition. Please take a look and let me know if I did miss something.

Thanks!

rsauciuc:
Looks good, thank you for the quick turnaround!

PR #1101, opened 2019-03-26
IDEA-205103 Sort maven archetype versions

timja:
Use maven version sorting instead of lexicographical order

Closes [IDEA-205103](https://youtrack.jetbrains.com/issue/IDEA-205103)


timja:
🤷

PR #1102, opened 2019-03-26
Remove option USE_PROJECT_PROFILE

dahlstrom-g:
IntelliJ IDEA 2019.1 deletes it automatically.


develar:
It is known bug in 191 and older branches and it is fixed in 192. I decided that better to not cherry-pick fix to 191 because actually it is not regression. (fixes are not safe).

PR #1104, opened 2019-03-28
Use StringJoiner instead of hand-written joining pattern

stsypanov:
сабж


amaembo:
I think using Stream API for these cases would produce more readable code. Will convert to streams instead.

PR #1106, opened 2019-03-29
Fix bug in BuildOutputInstantReaderImpl

tgeng:
The bug is that the buffer size stops at 50 though myCurrentIndex keeps
going up. So after finishing reading 50 lines, myCurrentIndex would
always go over the buffer size, in which case getCurrentLine() would
always return null.

https://youtrack.jetbrains.com/issue/IDEA-209816


vladsoroka:
Thanks for the contribution, I've added some tests for the reader and it finds some additional issues with or w/o the change. You can find the following commits fixing found issues:
https://github.com/JetBrains/intellij-community/commit/e8c6f169afc99ae33f860f5ce792379c1dc03be9
https://github.com/JetBrains/intellij-community/commit/fbef901fa5e92867747dfc9a4c570f8d43f7a2b3

Please, verify if it fixes issue with your `BuildOutputParser`-s and let me know if it doesn't.

PR #1108, opened 2019-03-30
Fix a memory leak in DocumentImpl (IDEA-210051)

sprigogin:
Keep track of created DocumentListenerDisposable and dispose them in the removeDocumentListener method.


donnerpeter:
Closed until there are many further similar leaks, see the discussion in https://youtrack.jetbrains.com/issue/IDEA-210051

PR #1109, opened 2019-04-02
Fix Gradle Daemons window with Gradle 5.3

chyingo:
The constructor of DaemonParameters class in Gradle tooling API
is updated since Gradle 5.3. It now takes FileCollectionFactory
as required parameter.
This change calls the constructor by reflection if current Gradle
version is newer than 5.3.

Change-Id: I21eeb8f1e95267892ab378a98ca4517eb3d50e07

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/internal/daemon/DaemonAction.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/internal/daemon/DaemonAction.java
@@ -3,15 +3,19 @@
  */
 package org.jetbrains.plugins.gradle.internal.daemon;
 
+import java.io.File;
+import java.lang.reflect.InvocationTargetException;
+import org.gradle.api.internal.file.DefaultFileCollectionFactory;
+import org.gradle.api.internal.file.FileCollectionFactory;
+import org.gradle.api.logging.Logging;
 import org.gradle.initialization.BuildLayoutParameters;
 import org.gradle.internal.logging.events.OutputEvent;
 import org.gradle.internal.logging.events.OutputEventListener;
 import org.gradle.internal.service.ServiceRegistry;
 import org.gradle.launcher.daemon.client.DaemonClientFactory;
 import org.gradle.launcher.daemon.configuration.DaemonParameters;
+import org.gradle.util.GradleVersion;
 import org.jetbrains.annotations.NotNull;
-
-import java.io.File;
 
 /**
  * @author Vladislav.Soroka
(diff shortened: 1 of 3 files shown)


plugins/gradle/src/org/jetbrains/plugins/gradle/internal/daemon/DaemonAction.java
```diff
@@ -28,7 +32,22 @@ protected ServiceRegistry getDaemonServices(DaemonClientFactory daemonClientFact
...
+      daemonParameters = new DaemonParameters(layout);
+    }
+    else {
+      try {
+        //noinspection JavaReflectionMemberAccess
+        daemonParameters = DaemonParameters.class.getConstructor(BuildLayoutParameters.class, FileCollectionFactory.class)
+          .newInstance(layout, new DefaultFileCollectionFactory());
+      }
+      catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+        Logging.getLogger(this.getClass())
```
chyingo: @vladsoroka I am not sure how to handle the exceptions here, do we want to re-throw it? Or make this method return nullable ServiceRegistry?

vladsoroka: Let's re-throw exceptions, it will be handled at https://github.com/JetBrains/intellij-community/blob/191/plugins/gradle/src/org/jetbrains/plugins/gradle/internal/daemon/GradleDaemonServices.groovy#L162-L170

plugins/gradle/src/org/jetbrains/plugins/gradle/internal/daemon/DaemonAction.java
```diff
@@ -28,7 +32,22 @@ protected ServiceRegistry getDaemonServices(DaemonClientFactory daemonClientFact
     if (myServiceDirectoryPath != null && !myServiceDirectoryPath.isEmpty()) {
       layout.setGradleUserHomeDir(new File(myServiceDirectoryPath));
     }
-    DaemonParameters daemonParameters = new DaemonParameters(layout);
+    DaemonParameters daemonParameters = null;
+    boolean isGradle5Dot3OrNewer = GradleVersion.current().compareTo(GradleVersion.version("5.3")) >= 0;
```
vladsoroka: minor: better to use base version for comparison to handle RC versions also.
`GradleVersion.current().getBaseVersion().compareTo(GradleVersion.version("5.3")) >= 0;`

chyingo: Thanks for catching it. Done.

plugins/gradle/src/org/jetbrains/plugins/gradle/internal/daemon/DaemonAction.java
chyingo: Thanks. Done.

vladsoroka:
Thanks for the contribution!
I've merged the change using the following commit https://github.com/JetBrains/intellij-community/commit/77ee05b9af3a44366d3754adc63f0dd4b9f00d52

Also I've fixed the same issue when `Stop Selected` action is used, see the commit at https://github.com/JetBrains/intellij-community/commit/55a0d663b6cbbf711c381c5296021b8555090c16

PR #1111, opened 2019-04-03
IDEA-210360 Make arrows in the navigation bar scale aware

jgaillard85:
This ensures that the navigation bar arrows are not blurry on Windows
with HiDPI screens.


forantar:
Thank you, I've slightly modified your fix to use ScaleContext.Cache and pushed it.

I had to make two commits, because in the first I forgot to mention you as the author (sorry).
fb739440f79d79ba82ed606f05befc064f77fd2d
9302d44332d1205f8f377de2caac8bb52d08f09f - this is authored by you with the link to pull-request

PR #1112, opened 2019-04-06
Remove listener before calling deactivateIfNeeded (IDEA-210564)

sprigogin:
Remove listener from myMouseListeners before calling deactivateIfNeeded (IDEA-210564)


bulenkov:
Already merged

PR #1114, opened 2019-04-10
Add missing quotes in NSIS configuration

jnthn:
Without this, a manufacturer with a space in the name causes `UnStrStr`
to be invoked with bogus extra arguments.


vorlovjb:
Fixed. Thank you.

PR #1115, opened 2019-04-11
Unstage file for deletion if it's being reused

dahlstrom-g:
Currently in SchemeManagerImpl it is possible for a scheme's containing
file to be deleted AFTER another scheme of the same containing file name
starts using the file.

SchemeManagerImpl stages files to be deleted but doesn't actually do so
until save() is called. While a file is staged for deletion, another
scheme can be "soft" saved with saveScheme(), which saves the scheme in
memory. saveScheme() does contain code to unstage the scheme's
containing file for deletion, but there is an edge case where this won't
happen due to an early exit that assumes no operation is required to
"soft" save the scheme if the new scheme is identical to the old one.
This results in a zero net-change in schemes, and a file staged for
unintentional deletion.

This moves the "unstage file for deletion" code before the exit so
it's always checked, which fixes the edge case scenario.

### DIFF
--- a/platform/configuration-store-impl/src/schemeManager/SchemeManagerImpl.kt
+++ b/platform/configuration-store-impl/src/schemeManager/SchemeManagerImpl.kt
@@ -364,6 +364,10 @@ class SchemeManagerImpl<T : Any, MUTABLE_SCHEME : T>(val fileSpec: String,
       fileNameWithoutExtension = nameGenerator.generateUniqueName(schemeNameToFileName(processor.getSchemeKey(scheme)))
     }
 
+    val fileName = fileNameWithoutExtension!! + schemeExtension
+    // file will be overwritten, so, we don't need to delete it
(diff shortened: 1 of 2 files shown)


develar:
Thanks a lot. 

From what I see, your change is safe, it is quite ok to call it a little bit earlier.

> During this process, old scheme files are scheduled for deletion.

It is not true anymore. Fixed already in master.

develar:
Your test is passed for me without changes. But anyway I found your change useful, so, merged, thanks again for test and comprehensive description.

develar:
@dahlstrom-g Cherry-picked to 191.

develar:
If you do this investigation because of https://youtrack.jetbrains.com/issue/IDEA-210208 (and related issues) — your fix and my latest fixes don't fix the problem. Right now I can reproduce the issue in our IDEA Ultimate project. I am sure that I will find cause soon.

PR #1116, opened 2019-04-11
IDEA-210834 fix gradle properties not being read on distribution initialization

monosoul:
This fixes an issue when Gradle distribution can't be downloaded if remote repo requires an authentication.

The suggested solution duplicates the `GradleWrapperMain` behavior: https://github.com/gradle/gradle/blob/ae6478d485a613b6b4c1f9761882a956adb9156c/subprojects/wrapper/src/main/java/org/gradle/wrapper/GradleWrapperMain.java#L70


monosoul:
I'll reopen it to the master branch

PR #1117, opened 2019-04-11
IDEA-210834 fix gradle properties not being read on distribution initialization

monosoul:
This fixes an issue when Gradle distribution can't be downloaded if remote repo requires an authentication.

The suggested solution duplicates the `GradleWrapperMain` behavior: https://github.com/gradle/gradle/blob/ae6478d485a613b6b4c1f9761882a956adb9156c/subprojects/wrapper/src/main/java/org/gradle/wrapper/GradleWrapperMain.java#L70

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleExecutionHelper.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleExecutionHelper.java
@@ -428,7 +428,7 @@ public class GradleExecutionHelper {
           break;
         case WRAPPED:
           if (settings.getWrapperPropertyFile() != null) {
-            DistributionFactoryExt.setWrappedDistribution(connector, settings.getWrapperPropertyFile(), serviceDirectory);
+            DistributionFactoryExt.setWrappedDistribution(connector, settings.getWrapperPropertyFile(), serviceDirectory, projectDir);
           }
           break;
       }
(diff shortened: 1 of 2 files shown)


vladsoroka:
Thanks for the contribution Andrei!
I've merged the change using the following commit https://github.com/JetBrains/intellij-community/commit/d015d357d293418dcc4c9d591b2ec24f50210b59
Also I've added the fix to avoid breaking of the 3d party IntelliJ plugins https://github.com/JetBrains/intellij-community/commit/c7dfc2164894751a06915359e5a33bae12d75241
And the change to avoid pollution of the IDE system properties https://github.com/JetBrains/intellij-community/commit/b0e6e149bbb2fb50af2b43fae6d0280ef9e39ad0
Please, let me know if you encounter further issues with those changes.

leemeador:
I couldn't figure out what your change had to do with the issue that the gradle wrapper couldn't be downloaded when the source of the download required authentication (as in a corporate repo proxy of the gradle main repository). So this comment is to add this info for future searchers.

You can put these two lines in $HOME/.gradle/gradle.properties:

systemProp.gradle.wrapperUser=userid
systemProp.gradle.wrapperPassword=password or token

This issue was about IntelliJ not using the contents of the gradle.properties file in this location and the fix was to use it. Note that the gradle home folder can be changed. Get a hint of all the possibilities here https://docs.gradle.org/5.4.1/userguide/build_environment.html

PR #1119, opened 2019-04-11
Added comparing description in MessageEvent.equals

AmrAfifiy:
In Android Studio we have problem where multiple error messages are
skipped because they have the same message, although they differ in
their descriptions.

This changes adds comparing the message description to the equals
function.


vladsoroka:
Thanks for the contribution, merged using the following commit https://github.com/JetBrains/intellij-community/commit/d005fdf5681e9494c2e40b0b62f2a71f9431ba2e

PR #1121, opened 2019-04-12
 Add zip store properties from Gradle wrapper

srmurguia:
GradleUtil#getWrapperConfiguration reads properties from the gradle
wrapper into a WrapperConfiguration, but it is missing zip store
properties. Currently, instances of WrapperConfiguration generated by
this function will have the default values regardless of what the values
are in the wrapper properties.


vladsoroka:
Thanks for the contribution, merged using the following commit https://github.com/JetBrains/intellij-community/commit/b454662394acf948b3b52abd914e70ca37949536

PR #1128, opened 2019-05-01
Extend the Gradle project resolver to allow the following customizations

Slvin:
We allow resolver extensions to initialize task running, this allows any model builders
to set tasks that should be run by Gradle during its invocation. If the tasks are not
changed by the model builders then this will cause the default task to be run. This is
currently used in Android IDEA Plugin to allow the Android Gradle Plugin to schedule
source generation tasks to be run after the build action.

We also add a callback to the resolver (buildFinished) that is called from the Gradle
connection thread that is called once the whole Gradle operation (including the running
of tasks) is finished.


nskvortsov:
merged manually

PR #1131, opened 2019-05-04
fsNotifier: fix build on musl

Cogitri:


klikh:

Cogitri:
Just signed it, thanks for the heads up.

trespasserw:
Hm, maybe it's time to drop the limit completely? It defaults to 2.14 - which doesn't look too strict now. What do you think?

Cogitri:
Yup, AFAICS about every distro has at least 2.14 in its repos: https://repology.org/project/glibc/versions

trespasserw:
OK, just dropped the symbols. Thanks!

Cogitri:
Thank you :)

Maybe it'd be possible to distribute musl versions of the FsNotifier with IntelliJ products, so it works out of the box (or to just statically built it against musl)?

trespasserw:
Then it won't work on systems without `musl` :(

Linking statically is a variant I'd very match like to avoid - the IDEs are used on a pretty broad spectrum  of kernel versions, and AFAIK `inotify` ABI has changed a couple of times in the past.

Cogitri:
> Then it won't work on systems without musl :(

Ah, meant to make fsnotifier*-musl variants

> Linking statically is a variant I'd very match like to avoid - the IDEs are used on a pretty broad spectrum of kernel versions, and AFAIK inotify ABI has changed a couple of times in the past.

Ah, I'm not very well versed with C (mostly Rust, thanks for the awesome tooling you provide :), can we just link libc statically and leave the rest be? Musl is pretty good at being linked statically (at least in Rust it's used by default for static linking)

trespasserw:
Well, I'm not sure whether or when we will bundle `-musl` variants, but a [feature request](https://youtrack.jetbrains.com) certainly won't harm.

`libc` is exactly the thing which shields the binary from kernel ABI changes - that's why it shouldn't be linked statically. There are no other dependencies.

Cogitri:
> libc is exactly the thing which shields the binary from kernel ABI changes - that's why it shouldn't be linked statically. There are no other dependencies.

Derp, sorry.


Will open a feature request, thanks :)

PR #1132, opened 2019-05-06
IDEA-201196 Change signature: Moving a new element downwards doesn't transfer focus correctly

alshain:
Fixes https://youtrack.jetbrains.com/issue/IDEA-201196


yole:
The bug was fixed with a different change by someone on our team. Thanks for your contribution anyway!

PR #1133, opened 2019-05-09
Removed const and goto in Java Flex lexer

SplotyCode:
In this PR I removed the const and goto keywords that were in the Java Flex implementation.
I know that this will not improve performance a lot but it is unnecessary and it also gets highlighted when you write it (not a big surprise).


klikh:

BasLeijdekkers:
Even though `const` and `goto` are not used in Java, they are still keywords and can't be used as identifiers. See https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.9
This also means that they should not be removed from the lexer.

PR #1135, opened 2019-05-10
IDEA-212479 Prioritize overrides and getter/setter generation when sensible in auto complete

alshain:
See https://youtrack.jetbrains.com/issue/IDEA-212479

### DIFF
--- a/java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
+++ b/java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
@@ -7,6 +7,7 @@ import com.intellij.codeInsight.lookup.LookupElement;
 import com.intellij.codeInsight.lookup.LookupElementBuilder;
 import com.intellij.codeInspection.ex.GlobalInspectionContextBase;
 import com.intellij.icons.AllIcons;
+import com.intellij.openapi.actionSystem.EmptyAction;
 import com.intellij.openapi.application.AccessToken;
 import com.intellij.openapi.command.CommandProcessor;
 import com.intellij.openapi.util.Iconable;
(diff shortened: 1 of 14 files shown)


java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
```diff
@@ -45,38 +46,84 @@ public static void fillCompletionVariants(CompletionParameters parameters, Compl
...
+
+      boolean hasOverrideAnnotation = false;
       if (modifierList != null) {
         String fileText = position.getContainingFile().getText();
+        String prefix = fileText.substring(modifierList.getTextRange().getStartOffset(), parameters.getOffset());
+        PsiAnnotation[] annotations = PsiTreeUtil.getChildrenOfType(modifierList, PsiAnnotation.class);
+        if (annotations != null) {
+          for (PsiAnnotation annotation : annotations) {
+            if (Override.class.getName().equals(annotation.getQualifiedName())) {
+              hasOverrideAnnotation = true;
```
donnerpeter: PsiModifierList.findAnnotation?

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
```diff
@@ -45,38 +46,84 @@ public static void fillCompletionVariants(CompletionParameters parameters, Compl
...
     if (psiElement(PsiIdentifier.class).withParents(PsiJavaCodeReferenceElement.class, PsiTypeElement.class, PsiClass.class).
       andNot(JavaKeywordCompletion.AFTER_DOT).accepts(position)) {
+      int priority = -1;
+      boolean followedByWhitespace = psiElement()
+        .withSuperParent(
+          2, // Go to PsiTypeElement
+          psiElement().beforeSiblingSkipping(
+            psiElement(PsiErrorElement.class), // If no annotations are present, the Psi will contain a "identifier expected" node right after
+            psiElement(PsiWhiteSpace.class)
+              //.withText(StandardPatterns.string().matches("(?s) *\n *\n.*")) // Require line to be followed by empty line
```
donnerpeter: Why is this commented out?

java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementUtil.java
```diff
@@ -75,14 +75,33 @@ private OverrideImplementUtil() { }
...
+  }
+
+  /**
+   * generate methods (with bodies) corresponding to given method declaration
+   *  there are maybe two method implementations for one declaration
+   * (e.g. EJB' create() -> ejbCreate(), ejbPostCreate() )
+   * @param aClass context for method implementations
+   * @param method method to override or implement
+   * @param toCopyJavaDoc true if copy JavaDoc from method declaration
+   * @param forceOverrideAnnotation use code style if false
```
donnerpeter: What happens in the language levels that don't require or prohibit override annotations?

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
```diff
@@ -45,38 +46,84 @@ public static void fillCompletionVariants(CompletionParameters parameters, Compl
     PsiElement position = parameters.getPosition();
     if (psiElement(PsiIdentifier.class).withParents(PsiJavaCodeReferenceElement.class, PsiTypeElement.class, PsiClass.class).
       andNot(JavaKeywordCompletion.AFTER_DOT).accepts(position)) {
+      int priority = -1;
+      boolean followedByWhitespace = psiElement()
```
donnerpeter: I suspect that written as plain old `if`, that might be more debuggable.

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
```diff
@@ -198,15 +264,18 @@ private static LookupElement createGenerateMethodElement(PsiMethod prototype,
                                               p -> getShortParameterName(substitutor, p) + " " + p.getName(),
                                               ", ") + ")";
 
-    String overrideSignature = " @Override " + signature; // leading space to make it a middle match, under all annotation suggestions
-    LookupElementBuilder element = LookupElementBuilder.create(prototype, signature).withLookupString(methodName).
-      withLookupString(signature).withLookupString(overrideSignature).withInsertHandler(insertHandler).
+    String depriorizationPrefix = (deprioritize) ? " " : "";
```
donnerpeter: Why deprioritize? Why in so strange way? Making it a middle match seems a very indirect way to change sorting. Especially given that now and then people come who want some middle matches above some start matches.

alshain: @donnerpeter I completely agree that it's a very indirect way. I reused the trick you introduced here: https://github.com/JetBrains/intellij-community/commit/ca3481dbd3f476fb03927b4a621685498d5c82d8#diff-909ebf4cda26a1ba075b0879f61dcdedR158

I wanted to preserve the old behavior that seemed to ensure that starting to type `@O` will always prefer annotation completions, e.g. maybe I wanted to type `@Overloaded` and the user shouldn't be shown suggestions for 20 different possible overrides at the top until they've typed `@Overl`

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
alshain: Cool, I didn't know that existed. I take it `PsiModifierList.hasAnnotation` is an even better choice here?

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
alshain: It's an idea I experimented I thought you might consider. It's a heuristic trying to differentiate between a user adding a "missing" `@Override` annotation to an existing method vs trying to write a new method/generate an override.

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
alshain: You mean instead of using Psi pattern matching, write out the checks using "manual" Psi traversals and checks?

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
donnerpeter: Ouch. You're right. Apparently I had no better ideas :(

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
donnerpeter: I see, thanks! Might be worthy

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
donnerpeter: True!

java/java-impl/src/com/intellij/codeInsight/completion/JavaGenerateMemberCompletionContributor.java
donnerpeter: Yes. Some `PsiTreeUtil` methods might be handy here, or maybe a `JBIterable.generate(e, PsiElement::getPrevSibling)`

java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementUtil.java
alshain: Currently, the caller would need to ensure that they only pass true if the language level supports override annotations. The calling code I wrote sets this flag to true if the user themselves has already typed out a `@java.lang.Override` annotation. The idea is that if the user writes `@Override` but the code style is set not to use `@Override`, the user still gets what they typed. If the language level doesn't support `@Override`, they probably wouldn't type it either (and or be able to access those completions)  I'm not sure what you mean by "if the language level doesn't require override annotations", isn't that always the case on Java >= 1.5?

donnerpeter: I meant Java <= 1.4. Having such a method in API could lead to misuse and bugs in such projects. Maybe the annotation should be added manually (if missing) in the caller instead, without any new API?

alshain: That makes sense. I wouldn't know how to best go about that, though. What about using `PsiUtil.getLanguageLevel(aClass)` to check whether `@Override` is supported and silently ignoring `forceOverrideAnnotation` in case it isn't?

alshain:
Would the comments be lost if I pushed my rebased branch here? If so, I can transfer my additional work back to the original branch and push them here so we can continue with the current comments.

I have added a few test cases, let me know if that's the direction you intended for the tests.
In the meantime, you can see them here:
https://github.com/JetBrains/intellij-community/compare/master...alshain:IDEA-212479-preview#diff-8c7153275601179f7e67ee8f6fd33fefL883

java/java-impl/src/com/intellij/codeInsight/generation/OverrideImplementUtil.java
donnerpeter: That'd make a confusing API. I'd try avoiding adding new overloads here at all, of possible.

donnerpeter:
> Would the comments be lost if I pushed my rebased branch here?

Who knows... But probably they won't if you just add new commits here without rebasing.

> I have added a few test cases, let me know if that's the direction you intended for the tests.

The link above gives me "There isn’t anything to compare." that I don't know how to resolve. However I've managed to find tests in your fork: https://github.com/alshain/intellij-community/commit/792600c3031677b9ab9a19a9f50382113bcda5e7 . The direction looks right, except that I wouldn't go into leading space assertions and leave things as they are right now: the change is complicated enough already.

alshain:
Thanks. About the leading spave assertions: do you mean I should instead include the leading spaces *in the test case expectations*?

Because currently I make the test framework pretend that there are never any leading spaces.

donnerpeter:
I think that the test framework should reflect the reality, not pretend it's something else. The test assertion could be rewritten in some way (e.g. using `contains`) to avoid mentioning leading space explicitly.

alshain:
@donnerpeter  I see. I prefer the tests to model my intentions, that's why I had removed them. I have now re-added the leading spaces in the tests and transferred my changes back to this branch.

alshain:
@donnerpeter Anything you need from me for this PR?

donnerpeter:
There's still the confusing new API in OverrideImplementUtil, TreeElementPattern and a couple more unresolved comments

PR #1136, opened 2019-05-13
IDEA-212508: Reduce memory footprint of MultipleBuildsView.myBuildsMap

sprigogin:
The change is based on the observation that build events with java.lang.Object IDs never have child events.


vladsoroka:
Thanks for the contribution, Sergey!

The change fixes some issue with the memory footprint of MultipleBuildsView.myBuildsMap, however it looks more like a workaround for the bad events routing implemented for parallel builds.
I'd not add a new method to BuildEvent interface as a workaround. 
Please, find the following change at https://github.com/JetBrains/intellij-community/commit/3a8e9d0b10fcf61060e5817c6551259cdf2c2e0a
It should do the same as your change but w/o changing of the BuildEvent interface.

nskvortsov:

PR #1138, opened 2019-05-15
ComboboxUI: Use user defined background if available

vcaen:
If a background color has been defined on a combobox, use it instead of
the default on.

This allows usage of the Combobox on Component with the non-default
background color


vcaen:
Hi @gregsh, in Android Studio, we have some panel which are not using the default background color for JPanel. The current implementation of ComboboxUI ignores any custom background set on the Combobox causing color bleeding around the combobox. This PR fixes the issue by using the custom background color if defined.

gregsh:
@vcaen Can you please share screenshots of that panel with
both Darcula LaF and MacIntelliJ (Default on macOS)?

vcaen:
@gregsh Here they are:

## Mac Dark
Mac

## Linux Dark
Linux

## Linux Light
Linux2

gregsh:
Thanks, I've pushed Darcula and MacIntelliJ changes to master, 
and discuss WinIntelliJComboBoxUI which also needs update.

PR #1139, opened 2019-05-15
IDEA-212623 Folding comments in Java out of FileHeader/Class blocks

ArtsiomCh:
see https://youtrack.jetbrains.com/issue/IDEA-212623


artfly:
Merged into master, thanks!

PR #1126, opened 2019-04-20
make getExecutable() generate a platform-specific filename

jglathe:
Hi there,

I'm trying to get the PyCharm IDE running faster on a Jetson Nano (linux-aarch64). Compiling the fsnotifier source with the local GCC is straight-forward enough (integrating the change into the build process is not yet), however NativeFileWatcherImpl.getExecutable() will error out with unsupported platform if it's not one of the specifically encoded platforms. To make this a little more hackable, I have changed the code to give a platform-specific executable name to look for instead. This way, providing fsnotifier-aarch64 on the Jetson Nano should do the trick without further need to change the PyCharm package.
There is room for improvement for sure, like ensuring that you get to know the filename that is wanted, however this is a start.

with best regards

Jens Glathe

PR #1140, opened 2019-05-19
native fsNotifier: try to use a rule-based filename as alternative for ex. aarch64

jglathe:
These patches add a platform-specific fsnotifier for linux-aarch64. It also extends NativeFileWatcherImpl.getExecutable() to generate a platform-specific filename to look for if the hardcoded choices are not met. 
Follow-up on #1126 , please see for more details on the whereabouts.
Tried and tested on a Jetson nano, with IDEA and PyCharm 191.7479.


franckrasolo:
FYI https://www.linuxondex.com is another use case where this would help.

trespasserw:
Sorry for the late reply.

At the moment, JB does not want to support other architectures besides x86/x86-64. Starting from the next 2019.3 build, IDEs will provide a user with instructions on building and installing the `fsnotifier` binary on their system.

jglathe:
If this means that the local-arch fsnotifier will be loaded, good. Otherwise, how does this work?

trespasserw:
With some help from a user, yes. Here are the instructions: https://confluence.jetbrains.com/display/IDEADEV/Compiling+File+Watcher.

jglathe:
Great, thanks!

jpalus:
Note that IJ 2019.3 with native fsnotifier compiled still logs exception in other part:
```
2019-11-28 14:12:24,641 [  17880]   WARN - openapi.util.io.FileSystemUtil - Failed to load filesystem access layer: Linux, 13.0.1 
java.lang.IllegalStateException: Unsupported OS/arch: Linux/aarch64
	at com.intellij.openapi.util.io.FileSystemUtil$JnaUnixMediatorImpl.<init>(FileSystemUtil.java:290)
	at com.intellij.openapi.util.io.FileSystemUtil.getMediator(FileSystemUtil.java:55)
	at com.intellij.openapi.util.io.FileSystemUtil.<clinit>(FileSystemUtil.java:46)
	at com.intellij.openapi.vfs.impl.local.LocalFileSystemBase.getAttributes(LocalFileSystemBase.java:723)
```

trespasserw:
^ Just a warning; one can get rid of it by adding `idea.io.use.nio2 = true` to *Help | Edit Custom Properties*.

jglathe:
That's great, should be added to https://confluence.jetbrains.com/display/IDEADEV/Compiling+File+Watcher. No idea if it increases the speed, when I understood the code correctly nio2 is already the fallback when no libc stat is available.

trespasserw:
No, it shouldn't - first, it is not related in any way to `fsnotifier`; second, we're going to use NIO2 by default (probably in the next version).

PR #1141, opened 2019-05-20
Fix the variable name to not conflict with the implicit setter

thagikura:
Otherwise there is a following error caused by the methods having
the same signature.

Platform declaration clash: The following declarations have the same JVM signature (setPluginDescriptor(Lcom/intellij/openapi/extensions/PluginDescriptor;)V):
private final fun <set-pluginDescriptor>(<set-?>: PluginDescriptor?): Unit defined in com.intellij.openapi.util.registry.RegistryKeyBean
@Transient public open fun setPluginDescriptor(pluginDescriptor: PluginDescriptor?): Unit defined in com.intellij.openapi.util.registry.RegistryKeyBean


thagikura:
Any update on this issue?

PR #1142, opened 2019-05-20
Update sphinx_runner to use Sphinx >= 2

d9pouces:
sphinx_runner does not work with Sphinx >= 2.0 since it checks for sphinx >= 1.7, < 2 instead of checking for sphinx >= 1.7.


ktisha:
Merged manually. Thank you for the pull request!

PR #1144, opened 2019-05-23
IDEA-182029: Improve New Package dialog usability

kirpichik:
Implements the requested functionality: https://youtrack.jetbrains.com/issue/IDEA-182029

The new package creation dialog for a single package looks like:

Снимок экрана 2019-05-22 в 19 52 24

When you press the "+" button or a combination of Shift + Enter, a new input field is added:

Снимок экрана 2019-05-22 в 19 44 21

Also includes: https://github.com/JetBrains/intellij-community/pull/1068

The dialog expands on its own when the package name reaches the edge of the input field. Also added an empty space in front of the right edge, so that the name does not hide behind the right edge. It was requested here: https://youtrack.jetbrains.com/issue/IDEA-209397


kirpichik:
Just want to ask, how should I notify the user about the presence of a hot-key Shift + Enter, which adds a new input field?

I thought about adding it to the “+” button in the tooltip, but it would create the illusion that this combination is available from any part of the dialog, whereas it is available only when the window is focused on the input field.

alshain:
Cool! For consistency, you could add the "generate/add new" shortcut that works basically anywhere there is an "add" bttton.

How does this a interact with multiple source/test roots?

kirpichik:
> For consistency, you could add the "generate/add new" shortcut that works basically anywhere there is an "add" bttton.

Did not quite understand what shortcut you mean?

> How does this a interact with multiple source/test roots?

This dialog allows you to create packages at the same time only within the same root. It seems to me that the possibility of creating a few roots too much complicates the dialogue.

alshain:
I meant Alt-Insert/Command-N:

See com.intellij.openapi.actionSystem.CommonShortcuts#getNew

BartvHelvert:
The use case for this feature seems quite rare, right now we are not ready to support it.
If you want to implement it anyway, probably it could be done through a plugin without official support.

PR #1145, opened 2019-05-26
Skip/ignore disposed modules in the project view.

yuriy-solodkyy:
The project view is updated asynchronously via a MergeQueue and
invokeLater calls. Therefore its nodes may for short periods of time
refer to disposed modules. If it happens that such a node is accessed
before it is updated it typically results in a crash.

This change fixes several most common cases where this type of crash can
happen. The fix is similar to how disposed modules are already handled
in the same classes.

Bug: 133469297
Test: n/a (manually by increasing MergeQueue update time to 10 seconds)
Change-Id: I0b1d247990b8d491c744249a02d2ef12897a1179


yuriy-solodkyy:
Could someone take a look at this?

SergeyMalenkov:
Thank you for the suggested fix

PR #1147, opened 2019-05-31
Use getArrowButtonPreferredSize(comboBox) instead

juancnuno:
Of arrowButton.getPreferredSize(). There are times when getSizeWithButton
gets called with a null arrowButton.


bulenkov:
Thanks! Merged!

PR #1148, opened 2019-06-02
Remove interactive option for shells

aleksator:
Fixes IDEA-215394 (https://youtrack.jetbrains.com/issue/IDEA-215394)
Tested with bash and fish, removal of the option produces no visible effects for them. For Ion, it allows it to start.


aleksator:
An issue has been fixed differently with 7ac33657ebf70f6eefc8d39c132ee1b295d0678e, so I'm closing this PR.

PR #1149, opened 2019-06-04
Support contributed references for the ShVariable PSI element

jansorg:
This PR adds support for contributed references to the `ShVariable` implementation. Variables can be resolved to definitions and this adds the possibility to let extensions contribute references.

I'm following the style which is used by other elements of the `sh.bnf` grammar.


jansorg:
@ignatov This is a new take to allow 3rd-party plugins to provide references for a `ShVariable`. It now uses `ShSupport` to get the references. 
The default value is the same as before, i.e. no references. 

It would be great to allow other plugins to provide references for variables which are not supported by Sh. 

Please let me know if this isn't what you'd like to merge.

PR #1151, opened 2019-06-05
Missing remote repository for org.jetbrains:markdown:0.1.31

alshain:
I have updated the remote jar repositories to contain the repository for org.jetbrains:markdown:0.1.31

Otherwise, the build fails if the library is not yet in the local maven repository with this error:

14:45	Repository library synchronization: No files were downloaded for org.jetbrains:markdown:0.1.31


chashnikov:
Merged to 'master', thank you for the fix!

PR #1154, opened 2019-06-10
Use the total column width instead of the component width

juancnuno:
Otherwise the last column looks wider than it actually is when a table's
auto resize mode is set to AUTO_RESIZE_OFF


sokolov-m:
Hi! We require our users to submit contributor agreements for any contributions. So could you please visit https://www.jetbrains.com/agreements/cla/  and sign CLA to make me able to apply your patch.

klikh:
@sokolov-m We've got a corporate agreement with Google, there is no need to sign an individual one.

sokolov-m:
Sorry, my bad. Going to apply patch

PR #1155, opened 2019-06-10
Skip Kotlin classes in NotNullInstrumentingBuilder

AlexeyTsvetkov:
Kotlin classes already contain null checks for parameters, so they should not be added by Intellij nullability instrumenter (see KT-13563 & KT-31879).

Instrumentation support was added in 1.3.40-eap-95, but it was disabled by default.
To enable it, check `kotlin.jps.instrument.bytecode` in registry and rebuild.

### DIFF
--- a/jps/jps-builders/src/org/jetbrains/jps/incremental/instrumentation/NotNullInstrumentingBuilder.java
+++ b/jps/jps-builders/src/org/jetbrains/jps/incremental/instrumentation/NotNullInstrumentingBuilder.java
@@ -17,9 +17,7 @@ import org.jetbrains.jps.incremental.CompiledClass;
 import org.jetbrains.jps.incremental.messages.BuildMessage;
 import org.jetbrains.jps.incremental.messages.CompilerMessage;
 import org.jetbrains.jps.model.java.JpsJavaExtensionService;
-import org.jetbrains.org.objectweb.asm.ClassReader;
-import org.jetbrains.org.objectweb.asm.ClassWriter;
-import org.jetbrains.org.objectweb.asm.Opcodes;
+import org.jetbrains.org.objectweb.asm.*;
 
 import java.io.File;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


jps/jps-builders/src/org/jetbrains/jps/incremental/instrumentation/NotNullInstrumentingBuilder.java
```diff
@@ -52,8 +50,10 @@ protected boolean isEnabled(CompileContext context, ModuleChunk chunk) {
   }
 
   @Override
-  protected boolean canInstrument(CompiledClass compiledClass, int classFileVersion) {
-    return (classFileVersion & 0xFFFF) >= Opcodes.V1_5 && !"module-info".equals(compiledClass.getClassName());
+  protected boolean canInstrument(CompiledClass compiledClass, ClassReader reader, int classFileVersion) {
+    return (classFileVersion & 0xFFFF) >= Opcodes.V1_5
+           && !"module-info".equals(compiledClass.getClassName())
+           && !IsKotlinClassCheckingVisitor.isKotlinClass(reader);
```
chashnikov: This code will parse each class file to check presence of Kotlin annotation, and this looks inefficient. Also since this code is included in Java plugin these checks will be performed even if Kotlin plugin is disabled. And anyway it looks untidy when Java plugin includes Kotlin-specific code.

I think it would be better to do something in Kotlin plugin code which will turn off not-null instrumentation for a particular class file. E.g. you may put a flag to CompiledClass instance (it implements UserDataHolder) and check that flag in `NotNullInstrumentingBuilder`. This will be much more efficient and you'll get rid of Kotlin-specific code in Java plugin. What do you think?

chashnikov: @altsvetkov ?

nikitabobko:
Close the PR since there is no recent activity and both mentioned issues ([KTIJ-17093](https://youtrack.jetbrains.com/issue/KTIJ-17093), [KTIJ-17267](https://youtrack.jetbrains.com/issue/KTIJ-17267)) are marked as resolved

chashnikov:
@nikitabobko for the record: [KTIJ-17093](https://youtrack.jetbrains.com/issue/KTIJ-17093) is marked as fixed, but actually the implemented behavior is hidden by a system property which is false by default, and [KTIJ-17267](https://youtrack.jetbrains.com/issue/KTIJ-17267) which was about enabling that property by default was closed as 'shelved'.

chashnikov:
An alternative implementation [was merged](https://github.com/JetBrains/intellij-community/commit/09fb01a1d1d8c8ebed4cba5bfd816262f6b0d6f2).

PR #1157, opened 2019-06-11
Use StringJoiner instead of hand-written joining pattern

stsypanov:
Some minor code simplifications.

### DIFF
--- a/plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/lang/documentation/GroovyPresentationUtil.java
+++ b/plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/lang/documentation/GroovyPresentationUtil.java
@@ -94,12 +94,11 @@ public class GroovyPresentationUtil {
       if (!structural.isEmpty()) {
         builder.append(".");
         String[] array = ArrayUtilRt.toStringArray(structural);
-        if (array.length > 1) builder.append("[");
-        for (int i = 0; i < array.length; i++) {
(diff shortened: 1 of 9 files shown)


plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/lang/documentation/GroovyPresentationUtil.java
```diff
@@ -94,12 +94,11 @@ else if (typePresentation == TypePresentation.CANONICAL) {
       if (!structural.isEmpty()) {
         builder.append(".");
         String[] array = ArrayUtilRt.toStringArray(structural);
-        if (array.length > 1) builder.append("[");
-        for (int i = 0; i < array.length; i++) {
-          if (i > 0) builder.append(", ");
-          builder.append(array[i]);
+        if (array.length > 1) {
```
amaembo: This change is incorrect: originally if array.length == 1, then the array[0] is still added (without brackets).

amaembo:
I applied the changes except the incorrect one. Also in couple of places I used Stream API instead. Thanks.

plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/lang/documentation/GroovyPresentationUtil.java
stsypanov: Indeed, you are right!

PR #1159, opened 2019-06-19
Rid usage of StringBuilder by chaining appends

stsypanov:


amaembo:
DocumentationComponent.java was updated after your patch and your patch cannot be applied there. Other two files are applied. We have temporary problems with public repo synchronization, so it may take a while till commit appears on GitHub.

In general I don't think that we need more of such patches. They add some constant overhead to apply them, review, check whether the result is compilable, etc., and the outcome (in terms of code readability or performance) is really low.

PR #1160, opened 2019-06-21
Change UI designer form compiler modality to any

CMoH:
I do not understand why the form compilation task referred to in this commit is requiring the application to have a non-modal state, as it seems to me this process can fulfill its job even if there is a modal dialog present. This PR changes the requirement to any modality state.

I hope I understood the API correctly, and please check that there are no modality requirements for the form compile task.


yole:
What problem is this change intended to solve?

CMoH:
I have to say, almost 2 years after, I vaguely remember I had to copy-paste the entire class instead of subclassing because of this unconfigurable constant, but the actual situation has long been erased from memory.

yole:
I assume this is no longer relevant. Please let me know if it is.

PR #1161, opened 2019-06-26
Optimize async stack frame loading

gharrma:
For each async stack frame we have to find a com.sun.jdi.Location
corresponding to a given line number. To do this we called
ReferenceType.locationsOfLine()---but that calls locationsOfLine() on
all methods in the class. Since we already know the name of the method
we're looking for, it's faster to first filter by method name, then
call locationsOfLine() directly on the matching methods.

This optimization is less important when running on a local JVM,
but very useful for remote debugging sessions. For example,
when debugging on an Android device, this change reduces the time spent
in findLocation() by 10x when loading 15 async stack frames, and the
frames show up in the UI significantly faster.

Before: 2767 ms spent in findLocation()
After: 247 ms spent in findLocation()

### DIFF
--- a/java/debugger/impl/src/com/intellij/debugger/engine/AsyncStacksUtils.java
+++ b/java/debugger/impl/src/com/intellij/debugger/engine/AsyncStacksUtils.java
@@ -222,15 +222,15 @@ public class AsyncStacksUtils {
 
   private static Location findLocation(DebugProcessImpl debugProcess, ReferenceType type, String methodName, int line) {
     if (type != null && line >= 0) {
-      try {
-        Location location = type.locationsOfLine(DebugProcess.JAVA_STRATUM, null, line).stream()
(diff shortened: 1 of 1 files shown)


gharrma:
Actually, now that I think about it: is there any downside to just always returning a GeneratedLocation instance instead of trying to look up an existing Location by line number?

I.e., can we just remove the whole block that starts with "if (type != null && line >= 0) {" ?

gorrus:
thanks for the fix, always returning GeneratedLocation may work, though I'm slightly worried about the method field inside, current implementation of GeneratedLocation constructor takes the first matched by name. If we choose to always return GeneratedLocation, we should probably find the method more correctly matching the line number again and so gathering line number information, so it seems that there will be no real gain here.

gorrus:
merged

PR #1167, opened 2019-07-09
Expose TextEditorWithPreview actions to subclasses

amaurymedeiros:
Subclasses of TextEditorWithPreview might want to customize the
presentation of ToggleActions used to change the editor mode. That can
be achieved by exposing the actions creation to subclasses through
protected methods. Subclasses, in turn, can create their own actions,
using TextEditorWithPreview's as delegates.

An advantage of this approach is also providing subclasses the ability
to create listeners for actions selection, selecting an action
programatically, etc.


bulenkov:
Thanks. Pushed to master with some customizations 0895f02cef91b8e6f7c6958658b53fd0e9728f2f Will be automatically synced with GitHub later.

dahlstrom-g:
> Thanks. Pushed to master with some customizations. Will be automatically synced with GitHub later.

I gather that's commit ec7128a. Can this be included in IDEA 2019.2?

PR #1168, opened 2019-07-09
Fix unstable order of dependencies and modules listed in iml files after import from Gradle

ausatiy:
What actually done:
1. add OrderAware for dependencies with PROVIDED scope
2. library order entries with the same order are sorted alphabetically (case when a dependency is resolved in file collection)
3. OrderEntries without order are added in the beginning of the order list (these entries are actually added by extension plugins)


nskvortsov:
merged manually

PR #1174, opened 2019-07-16
Throw IAE instead of AE from isExcluded

dahlstrom-g:
Fixes testIllegalArgumentInIsExcludedMethod after commit 1dc7f6b.


dahlstrom-g:
@cdracm

cdracm:
pushed, thanks

PR #1175, opened 2019-07-17
Register FileEditorManagerImpl for disposal

dahlstrom-g:
… with its parent project, since commit 84880bc registers its children
to it instead of the project.


dahlstrom-g:
@develar

develar:
Why it is required? FileEditorManagerImpl it is project component, so, automatically disposed once parent container (project in this case) is disposed. So, no need to register it explicitly to be disposed.

dahlstrom-g:
Without this tests fail `Caused by: java.lang.AssertionError: Found leaked class com.intellij.openapi.project.impl.ProjectImpl […] via 'com.intellij.openapi.fileEditor.impl.FileEditorManagerImpl.myProject`

dahlstrom-g:
The failing test code has been removed.

PR #1176, opened 2019-07-19
Add LibraryLicense for module-level library Joni

dahlstrom-g:
… which was added by commit 1093cb7.


dahlstrom-g:
@denofevil

denofevil:
Thank you for noticing, fixed via ddb08ac14d0493e7861535bbd6cf5495c6ad134f along with jcodings licence

PR #1178, opened 2019-07-19
JBR-1337: add logs on JavaFX init errors, prevent stuck IDE

ahus1:
This PR is a proof-of-concept that:
* adds log messages once the JavaFX initialization fails (like this is the case for example in GlobalMenuLinux)
* adds a crude, yet effective way to detect if an exception in a previous run of PlatformImpl.startup() lead to a stuck state in the JavaFX initialization

Please advise if a different kind of notification for the user would be appropriate.

I was unable to build the complete IDE locally, but I tested the code as part of the AsciiDoc plugin including the "stuck" detection.


forantar:
Thank you for the contribution. However, the original cause of the deadlock was fixed right in JavaFX we ship with our JBR 11. Please find it here: https://github.com/JetBrains/jdk8u_jfx/commit/0fcd66f384b2dd543990d8f2cae1fc4813e0819b

PR #1180, opened 2019-07-27
Minor improvements in editorconfig language support

kirillgla:
'No matching files' inspection used to fail with assertion error when header was incomplete, like in the newly added test


kirillgla:
'No matching files' inspection used to fail with assertion error when header was incomplete, like in newly added test

kashmervil:
Pushed to master

PR #1181, opened 2019-07-30
Fix AndroidXmlFormat issue when creating new layout

aquagray:
We're seeing an issue where Default XML styling takes higher priority
over Android XML styling when creating Android XML Layout.

Newly created layout files do not have directory set yet when format is
applied. However, almost everywhere, directory info is used to determine
the format. Which causes all android formatting to be ignored for newly
created files, and default formats were being applied.

This change sets the directory info before applying style formats.

Related bugs :
https://buganizer.corp.google.com/issues/135967844
https://youtrack.jetbrains.com/issue/IDEA-219512


klikh:
As mentioned in https://youtrack.jetbrains.com/issue/IDEA-219512#focus=streamItem-27-3705110.0-0 this issue breaks some tests. Could you please fix the issue and update the PR?

PR #1183, opened 2019-08-03
IDEA-22407: Comply with XDG Base Directory Specification on Linux

cxrvh:
Implement platformPath for the XDG Base Directories Specification and
split the IDEA configuration into the corresponding XDG_CONFIG_HOME,
XDG_DATA_HOME & XDG_CACHE_HOME directories.

In case no old configuration directory under XDG_CONFIG_HOME exists
then import the last old configuration from the home directory. But
move the plugins directory into the corresponding directory under
XDG_DATA_HOME.

Parameters are renamed to indicate the usage on multiple systems.

### DIFF
--- a/platform/configuration-store-impl/testSrc/ConfigImportHelperTest.kt
+++ b/platform/configuration-store-impl/testSrc/ConfigImportHelperTest.kt
@@ -55,7 +55,7 @@ class ConfigImportHelperTest {
   }
 
   @Test
-  fun `find recent config directory on macOS`() {
+  fun `find recent config directory with base directories`() {
     doTest(true)
   }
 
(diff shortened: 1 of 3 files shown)


klikh:

cxrvh:
I submitted the CLA.

trespasserw:
Thanks!

I have a couple of remarks to changes in the import code, though, and it would be nice to add some tests. Would you like to improve the patch yourself or rather want me fix it?

cxrvh:
Since it's only a small patch, it'll quicker if you just fix it.

trespasserw:
Status update: we've planned the change for IDEA 2020.1, along with a similar one for Windows.

PR #1184, opened 2019-08-05
cache JAXBContext to avoid overhead on parsing every SVN operation result

deathy:
PR for https://youtrack.jetbrains.net/issue/IDEA-219881


amaembo:
The patch is applied, thanks.

PR #1185, opened 2019-08-06
IDEA-220057: NPE in findWindowForBalloon()

gharrma:


gharrma:
I'm not 100% certain what to do in the case that `wrapper == null`. For now I just break out of the loop and return `frame`.

gharrma:
It looks like this was fixed in https://github.com/JetBrains/intellij-community/commit/fff0a775aeaf36b948ab8bacaee430ff691ab53a

This PR can probably be closed.

PR #1187, opened 2019-08-12
[Gradle] Avoid task configuration in tooling model builders.

chyingo:
When building Gradle tooling models, there are a few usages of
TaskCollection.all, which triggers creation and configuration of
all registered tasks.
To avoid this expensive operation, replace the usage with lighter
API - TaskCollection.withType, which only configures the affected
tasks.

Besides, introduce project property idea.do.not.resolve.task.list.
When defined, the task list in ExternalProject will not be populated.
Because this operation requires configuration for all tasks and there
is no light-weight alternatives.
This property provides a way to disable task configuration from idea.


nskvortsov:
Thank you for the request.

What is the intended usage of `idea.do.not.resolve.task.list` ? Without tasks information from Gradle multiple issues may rise (still to be checked with the patch): missing autocomplete in different places, problems runnig unit tests in delegated mode, maybe other...

vladsoroka:
> I am not aware of all of the side effects this may cause. Is it possible to use alternative APIs to only get certain tasks?

We can introduce some kind of `filters` to exclude certain tasks. Is it possible to distinguish such "useless" for IDE and IDE user android tasks? Maybe by task type or task naming pattern?

chyingo:
I simplified this change to only replace TaskCollection.all with TaskCollection.withTypes.
Will address the task filtering problem separately.

vladsoroka:
Thanks, Ying! 
merged

PR #1188, opened 2019-08-14
IDEA-219943 language counters corner case

gharrma:
I have thought of a corner case for the language-specific PSI modification trackers.

Say I get a tracker with
```
exampleTracker = PsiModificationTrackerImpl.forLanguages(l -> l.is(JavaLanguage.INSTANCE) || l.is(KotlinLanguage.INSTANCE))
```

Now suppose the project has just been opened and the `myLanguageTrackers` is empty. Then `exampleTracker.getModificationCount()` will be 0.

Now, if the user deletes a Java file, "all" language counters will be incremented. Since `myLanguageTrackers` is still empty, though, no counters will be incremented. And so `exampleTracker.getModificationCount()` will still be 0, even though a Java file changed.

I guess the root cause is that `forLanguages(Condition)` allows a modification tracker to depend on languages that aren't in the `myLanguageTrackers` map yet. One possible fix is to have a separate "`unattributedLanguage`" modification tracker whose modification count is included with the modification trackers returned from `forLanguage()` and `forLanguages()`.

### DIFF
--- a/platform/core-impl/src/com/intellij/psi/impl/PsiModificationTrackerImpl.java
+++ b/platform/core-impl/src/com/intellij/psi/impl/PsiModificationTrackerImpl.java
@@ -32,8 +32,10 @@ public class PsiModificationTrackerImpl implements PsiModificationTracker, PsiTr
 
   private final SimpleModificationTracker myModificationCount = new SimpleModificationTracker();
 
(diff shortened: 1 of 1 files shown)


gregsh:
Thanks, merged.

PR #1193, opened 2019-09-01
Make descriptor loading lazy in editorconfig plugin

kirillgla:
A descriptor is an internal structure used by the plugin. It is not required until an editorconfig file is opened, so it can be initialized lazily. This should slightly speed up startup


kashmervil:
Cleanup is okay, but absolutely not necessary. As for improving the startup time, services are lazily instantiated **by default**, meaning it will be created on the first `EditorConfigOptionDescriptorManager.Companion#getInstance` call, creating an additional provider won't help there in any possible way

PR #1195, opened 2019-09-07
Restore module intellij.jsp.base to java-api.jar

dahlstrom-g:
This was removed by commit 4438460 but runtime dependencies on it remain
for example in module intellij.jsp where JspFile extends BaseJspFile.

### DIFF
--- a/platform/build-scripts/groovy/org/jetbrains/intellij/build/JavaPluginLayout.groovy
+++ b/platform/build-scripts/groovy/org/jetbrains/intellij/build/JavaPluginLayout.groovy
@@ -37,6 +37,7 @@ class JavaPluginLayout {
         "intellij.java.indexing",
         "intellij.java.psi",
         "intellij.java",
+        "intellij.jsp.base",
         "intellij.jsp",
         "intellij.platform.uast"
       ].each {


dahlstrom-g:
@chashnikov

chashnikov:
`intellij.jsp.base` shouldn't be included to Java plugin layout, because it's used in JSP plugin which may be installed in non-java IDEs (e.g. WebStorm). So we need to restore it in the platform layout for Java IDEs instead. Actually it was already done for IDEA Community by 3aac73a795. I'll improve that fix by including the module to `BaseIdeaProperties#JAVA_IDE_API_MODULES` instead, so it'll be available in all Java-based IDEs including Android Studio.

chashnikov:
> I'll improve that fix by including the module to BaseIdeaProperties#JAVA_IDE_API_MODULES instead, so it'll be available in all Java-based IDEs including Android Studio.

Done: f854c6b090867f7, so this change isn't needed anymore.

chashnikov:
BTW it would be simpler to avoid such problems in future if you move AndroidStudioProperties.groovy from your fork of intellij-community repository (tools/idea) to Android plugin repository (tools/adt/idea).

PR #1197, opened 2019-09-14
[Sh] Change visibility of properties of ShRunConfiguration

jansorg:
This pull request makes the properties of `ShRunConfiguration` public to allow 3rd-party plugins to use the run configuration.
This would allow 3rd-party plugins to add debug support, for example. Properties of ShRunConfiguration should be accessible to allow reuse.

It's not strictly needed to have public setter-methods. But public getter-methods and package-level setter-methods seems a bit awkward. I'm happy to change this if you want me to, though.


ignatov:
Is it the only way how you may add debug support? I'll be happy to review the way how you're planning to use information from these fields.

PR #1199, opened 2019-09-16
Open a project by launcher.py when other window is already open

nwtgck:
Hi developers. Thank you very much for the wonderful project! 

The `launcher.py` opens a project properly when no IntelliJ window is running. However, when some IntelliJ windows are running, the `launcher.py` doesn't open properly in macOS. This PR fixes the problem.

I provide the following two demo video about before-PR and after-PR.

## Before PR
idea-open-before-pr mp4 opt

## After PR

idea-open-after-pr mp4 opt

### DIFF
--- a/platform/platform-resources/src/launcher.py
+++ b/platform/platform-resources/src/launcher.py
@@ -4,6 +4,7 @@
 import os
 import socket
 import struct
+import subprocess
 import sys
 
 # see com.intellij.idea.SocketLock for the server side of this interface
@@ -124,7 +125,7 @@ def start_new_instance(args):
     if sys.platform == 'darwin':
         if len(args) > 0:
             args.insert(0, '--args')
-        os.execvp('/usr/bin/open', ['-a', RUN_PATH] + args)
+        subprocess.call(['/usr/bin/open', '-na', RUN_PATH] + args)
     else:
         bin_file = os.path.split(RUN_PATH)[1]
         os.execv(RUN_PATH, [bin_file] + args)


trespasserw:
Sorry for the late reply.

Unfortunately, from the animated GIFs it is not clear at all what's going wrong :( Could you please describe the problem in plain words?

As for the fix - it basically adds `-n` to the list of `open` options, right? The problem is that this code shouldn't be called when there is another instance running - it is supposed to be activated through `try_activate_instance()` call above.

nwtgck:
@trespasserw Thanks for the reply.

Sorry for the GIFs. 
I try to write in plain words. In the "Before PR", I had opened project A already, and then, I tried to open project B by `idea .` command. However, the `idea .` command just focuses on the existing project A window, not opening project B.

In "After PR",  I had opened project A already, and open another project B by `idea .` command. The command opens new window and shows project B.

trespasserw:
I see, thanks, makes sense then.

Replacing `os.exec*()` with `susbsystem.call()` is not quite accurate, though - the former preempts the current process instead of forking a new one and is therefore somewhat faster.

nwtgck:
I also think `os.exec*()` is better. But unfortunately when I used `os. execvp()`, it could not open a new window and just forced on the existing window.

But now, I found the way to use `os.execvp()`. I'll update the commit.

trespasserw:
Found a reason - the `execv()` call was missing the `argv[0]` parameter. The complete fix will be included in 2019.3.3.

Thanks!

PR #1200, opened 2019-09-17
Fix path bug ('/' instead of ':' on osx for example)

virusdave:
Simple bug, but critical.  IJ shouldn't be looking on the path to find binaries
that are relative, such as `./myBinary`.  `pathSeparatorChar` actually is the 
character used to separate elements in the PATH environment variable, not
the character used to separate directories in the filesystem.

https://docs.oracle.com/javase/7/docs/api/java/io/File.html#pathSeparatorChar
vs
https://docs.oracle.com/javase/7/docs/api/java/io/File.html#separatorChar

### DIFF
--- a/platform/platform-api/src/com/intellij/execution/configurations/GeneralCommandLine.java
+++ b/platform/platform-api/src/com/intellij/execution/configurations/GeneralCommandLine.java
@@ -393,7 +393,7 @@ public class GeneralCommandLine implements UserDataHolder {
     }
 
     String exePath = myExePath;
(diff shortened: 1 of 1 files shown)


klikh:

virusdave:
@klikh Yup, done.

virusdave:
@klikh @trespasserw Any thoughts on merging this?  It's pretty simple fix, but damn is the underlying bug super annoying and frankly pretty dangerous (can lead to execution of unexpected programs by IJ pretty easily).

trespasserw:
Thanks for the fix! I've merged it already, should appear in GH in an hour (we don't host sources here, it is used as a mirror).

PR #1202, opened 2019-09-22
PY-25689 image Sphinx directive doesn't work in docstring documentation

Eitol:
Fix the problem: the prefix "file: //" was not added to local image references.

image

ref: https://youtrack.jetbrains.com/issue/PY-25689

contributor license agreement -> ready

### DIFF
--- a/platform/lang-impl/src/com/intellij/codeInsight/documentation/DocumentationComponent.java
+++ b/platform/lang-impl/src/com/intellij/codeInsight/documentation/DocumentationComponent.java
@@ -6,6 +6,7 @@ import com.intellij.codeInsight.CodeInsightBundle;
 import com.intellij.codeInsight.hint.HintManagerImpl;
 import com.intellij.codeInsight.lookup.LookupEx;
 import com.intellij.codeInsight.lookup.LookupManager;
+import com.intellij.configurationStore.JbXmlOutputter;
 import com.intellij.icons.AllIcons;
 import com.intellij.ide.DataManager;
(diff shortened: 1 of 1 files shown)


Eitol:
hi @ignatov Is there any way to move this forward?

ignatov:
@denofevil and @JB-Dmitry, please look on this.

platform/lang-impl/src/com/intellij/codeInsight/documentation/DocumentationComponent.java
```diff
@@ -776,14 +779,73 @@ private void setDataInternal(@Nullable SmartPsiElementPointer<PsiElement> elemen
...
+   *              then it returns the value passed by parameter
+   */
+  private String fixImageReferences(String text) {
+    if (text == null || text.isEmpty() || !BasicHTML.isHTMLString(text)) {
+      return text;
+    }
+    text = text.replace("<br>", "<br/>");
+    String basePath = this.myManager.myProject.getBasePath();
+    try {
+      org.jdom.Element dom = JDOMUtil.load(text);
```
denofevil: Sorry, this approach won't work: `text`  is an HTML document with all funny things like unclosed tags and unquoted attribute values, which makes it unreadable by XML parsers.
Also, I don't think that it's a proper place to fix Python specific problem, maybe `com.jetbrains.python.documentation.PythonDocumentationProvider` would be a better place.

denofevil: @vlasovskikh

PR #1204, opened 2019-09-23
Make default Layout of TextEditorWithPreview configurable

amaurymedeiros:
If the Layout can't be retrieved from PropertiesComponent, we were
defaulting to SHOW_EDITOR_AND_PREVIEW, but for some subclasses of
TextEditorWithPreview other modes might make more sense.


bulenkov:
Thank you. Merged.

PR #1205, opened 2019-09-24
Add unit tests for unindent with selection

Lenbok:
This adds two unit tests that reproduces IDEA-223397.

* The first test is using the normal selection model. Test currently passes.
* The second test repeats the equivalent edit operation using sticky selection model. Test currently fails.

(Sorry I don't know my way around the editor enough to attempt an actual fix for the bug)


JB-Dmitry:
Cannot push a test which fails (without a fix for the problem).

Lenbok:
@JB-Dmitry Note that I didn't expect you to directly merge it with a failing test. This is a demonstration that the bug really exists and is reproducible, to serve as a starting point for someone knowledgeable to address the issue. Is there a way to annotate the test as being expected to fail until the issue is addressed? (equivalent to Python's http://docs.python.org/dev/library/unittest.html#unittest.expectedFailure)

PR #1206, opened 2019-09-26
Add invokeLater helper

Qumeric:
Hi,

I am Android Studio developer and I have found myself using ApplicationManager.getApplication().invokeLater quite often, so I think it deserves a helper function.


klikh:
There is `runInEdt` defined in the same file. It calls `invokeLater` if it is being invoked not on the EDT. Are you sure you need to `invokeLater` on the EDT so frequently?

Qumeric:
Oh, okay, this is probably what I need. Maybe we should rename (or do something similar) it then because current name is not very discoverable. Not sure, but I think that nobody in the Studio team is aware of this helper.

yole:
Thanks! I've refactored the code a bit so there is now an `invokeLater` helper.

PR #1207, opened 2019-09-29
[IDEA-218836] Added SortLinesActions

fluidsonic:
New `EditorSortLinesActionGroup` + children to sort selected lines ([IDEA-218836](https://youtrack.jetbrains.com/issue/IDEA-218836)).

Eight child actions:
- Alphabetically
- Alphabetically (case-sensitive)
- By Length
- Naturally
plus a *descending* variant for each.

### Other changes

- Added relevant `IdeActions`
- Added default shortcut `control shift S` for *Sort Lines* similar to `control shift J` for *Join Lines*
- Added Sublime-specific shortcuts
- Added tests for new actions
- Muted `ActionsWithoutUpdateMethodTest` for `EditorSortLinesActionGroup`. The group is enabled/disabled by making all children invisible in case the action cannot be used which in turn automatically disables the *Sort Lines* group in the menu completely.

#### Change in *Editor Actions* actions tree

I also had to make a little change to `ActionsTreeUtil.createEditorActionsGroup()` so that *Editor Actions* in the *Keymap* settings pane will also group certain actions. Otherwise the *Sort Line* child actions (Alphabetically, By Length, …) would be listed in there without context which is confusing. The group is also intended to have a shortcut on its own which allows selecting the sort order on-the-fly through a popup.

Note that only groups with a non-empty template text will be listed. So far that's only *Sort Lines* (not accounting for plugins here). Action groups without text fall back to the standard behavior of being flattened into *Editor Actions* directly. Thus IDE and plugin developers can opt for grouping on a per-case basis.

In the worst case some plugin developers have already defined a text for a group they add to *Editor Actions* which will cause their actions to be grouped.

It may even make sense to deliberately group actions in *Editor Actions* in the future to make it easier to navigate as it's getting rather long.


JB-Dmitry:
We've just implemented IDEA-218836 ourselves.

PR #1208, opened 2019-10-03
visually separate DUMMY_IDENTIFIER text from the main text

ArtsiomCh:
For the debuging pupose:
image
image

PS Also more `zzz` might need to be added at the end. ;)


donnerpeter:
Unfortunately, this change breaks completion in about 10 languages we support, and an unknown amount of third-party plugins. Some languages treat underscore differently from uppercase letters, so parse trees there could change.

ArtsiomCh:
Yeah, you are right. Already seen "IntellijIdeaRulezzz" string hardcoded inside CMake plugin bandled with Clion...

PR #1210, opened 2019-10-07
Add linux-ppc64le check-condition to support fsnotifier-ppc64le binary usage

rubin55:
fsnotifier (which lives under `native/fsNotifier/linux` for Linux platforms) can be compiled on various linux platforms. 

For Linux based 64-bit platforms, NativeFileWatcherImpl.java only knows about `x86_64` and `arm`. This tiny pull-request adds `ppc64le` to the mix. 

This will give people on Little-Endian Power systems the ability to run IntelliJ and its deratives with a fast filesystem changes notifier.


trespasserw:
Sorry for the late reply.

At the moment, JB does not want to support other architectures besides x86/x86-64. Starting from the next 2019.3 build, IDEs will provide a user with instructions on building and installing the `fsnotifier` binary on their system.

rubin55:
You’re supporting arm already..

trespasserw:
Not anymore.

PR #1215, opened 2019-10-17
Another custom Logger.getInstance() inspection

SamB:
I noticed some invocations like:

    Logger.getInstance("#fqn.of.Class");

And, since the actual FQNs are invariably longer than in my example, I naturally began to wonder if any of these
invocations was misspelled or out-of-date.

So here's an inspection with a quick-fix, which I made by copying the one that rewrites:

    Logger.getInstance("#" + $c$.class.getName())

to:

    Logger.getInstance($c$.class)

and then changing the search template appropriately.

(FYI, I did this directly in the XML, because the inspection settings page just felt too ... cramped and modal.)

I was pleasantly surprised that I didn't need to do anything to turn the FQN text into an expression.

P.S. There are invocations like this in, e.g., Kotlin that would need to be handled separately.

Signed-off-by: Samuel Bronson <naesten@gmail.com>


SamB:
Hmm, I seem to have made a mistake in naming the inspection ... I'll push a fixed version soonish ...

SamB:
Well, I was all ready to submit a version with an HTML-safe title, but then I decided I might as well also send you a preview of what happens when you actually apply the quick fix, and it turned out that there were several instances where it generated invalid code (because the FQN from the string did not refer to a class in the compilation classpath).

So I'm postponing this until I find a satisfactory way of checking whether an FQN resolves to a class.
(Which might involve also figuring out how to get Groovy injection working on these `script` attributes.)

donnerpeter:
Yes, checking that the reference resolves is a must. And maybe reporting only when the class is the current one.

PR #1217, opened 2019-10-21
Debugger: Provide MethodsTracker from JavaExecutionStack to PositionManagerEx t…

Pr0p1k:
In JavaExecutionStack.createFrame it tries to create a custom frame (e.g KotlinStackFrame), but PositionManager doesn't obtain MethodsTracker and there's no sane way to get it, so some functionality like recursion depth in frames view doesn't work.


gorrus:
implemented differently

PR #1218, opened 2019-10-24
Enable brace highlighting for read-only viewers

sxhya:
The current implementation of brace highlighter is written in such a way that brace highlighting is disabled for editors in read-only mode (editor.isViewer()==true). There is no obvious reason to do that, so this one-liner pull-request enables brace highlighting in this situation. Brace Highlighting tests in platform.lang.test pass normally after this change.

### DIFF
--- a/platform/lang-impl/src/com/intellij/codeInsight/highlighting/BraceHighlightingHandler.java
+++ b/platform/lang-impl/src/com/intellij/codeInsight/highlighting/BraceHighlightingHandler.java
@@ -111,7 +111,7 @@ public class BraceHighlightingHandler {
 
   private static boolean isValidEditor(@NotNull Editor editor) {
     Project editorProject = editor.getProject();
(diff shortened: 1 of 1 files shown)


sxhya:
Do I need one if I am already a JB employee?

klikh:
No, of course not.

JB-Dmitry:
@chashnikov You've disabled brace highlighting for viewers long time ago (in commit 4887978 with message "FBQ-11657 fixed"). Any chance you remember why it was done?

chashnikov:
@JB-Dmitry This was done 15 years ago, unfortunately I don't remember why I did that. The issue is from JIRA project for JetBrains Fabrique product which was never released and this JIRA instance isn't available anymore. So I think it's ok to revert my change.

JB-Dmitry:
OK, pushed the fix.

PR #1220, opened 2019-10-25
IDEA-225641: Use SofterReferences instead of SoftReferences

sprigogin:
Use SofterReferences instead of SoftReferences in ResolveCache


sprigogin:
Superseded by https://github.com/JetBrains/intellij-community/pull/1222.

PR #1221, opened 2019-10-25
IDEA-225647 "View Mode" menu item text changes

justcla:
Now using proper strings from resource bundles so the mnemonics work.


Vassiliy-Kudryashov:
Thank you, the issue was fixed in a similar way.

PR #1223, opened 2019-10-29
Null check is added for PopupChooserBuilder.myItemChosenRunnable:crea…

cyberspaceru:
Null check is added for PopupChooserBuilder.myItemChosenRunnable:createPopup

When I try to create JBPopup like:
```java
  JBPopupFactory.getInstance()
                .createPopupChooserBuilder(table)
                .setCloseOnEnter(false)
                .createPopup()
                .showInScreenCoordinates(component.getAnchor(), point);
```
I have to invoke setItemChoosenCallback with a stub to prevent NPE:
```java
  JBPopupFactory.getInstance()
                .createPopupChooserBuilder(table)
                .setItemChoosenCallback(point::toString) // Stub
                .setCloseOnEnter(false)
                .createPopup()
                .showInScreenCoordinates(component.getAnchor(), point);
```


yole:
Since the PR was submitted the null check was added a few lines above, so the PR is no longer relevant. Thanks for your contribution anyway.

PR #1224, opened 2019-10-29
Fix bug: environment variables of before run task can not be saved to…

caiyunlei:
Fix bug: environment variables of before run task can not be saved to xml config file.
image


nskvortsov:
Thank you!
Meged as 230b52b3894c9b0b5eeabcd27507e44da52d430b

PR #1225, opened 2019-10-29
IDEA-225826 Update font of ActionButtonWithText

jgaillard85:
when the overall font is changed.


Vassiliy-Kudryashov:
Thank you! The patch (after some tuning) was applied to master, it's going to be cherry-picked to 193 too, just after code-review.

PR #1226, opened 2019-10-29
Remove final keyword from IdeTooltipManager

dahlstrom-g:
This was added incidentally by commit 977a1d1
breaking tests that mock the class.


dahlstrom-g:
@develar?

dahlstrom-g:
@develar?

develar:
Mockito is able to mock final classes, please see files likes `org.mockito.plugins.MockMaker` (there are two in android repo).

dahlstrom-g:
We use standard Mockito. I appreciate the pointer that we could use extensions providing alternative mock implementations, but I'd rather not ask folks on the Android Studio team to rework many of our existing tests when we could with much less effort simply not add final modifiers that weren't there when the tests were written. Frankly there are many more significant issues demanding attention. Can we agree on that and move forward with this and #1228?

PR #1227, opened 2019-10-30
IDEA-225846: Cache GitRepositoryImpl instances to save memory

sprigogin:


klikh:
Thanks, but this fix is incorrect for the reasons described in the ticket.

PR #1231, opened 2019-11-04
Fix PY-38835

Akababa:
Fix for https://youtrack.jetbrains.com/issue/PY-38835
Not tested


vladimir-koshelev:
The issue is obsolete.

east825:
Specifically, the problem seems to have been fixed with another PR #2408 for PY-36368.

PR #1233, opened 2019-11-05
fix potential NullPointerException in Graphics2DLog

rendner:
This PR contains the following changes:

-  fix a potential `NullPointerException` in `drawImage(Image img, int x, int y, ImageObserver observer)`
- log the right method name
- remove `String.format` if not required


klikh:
Merged manually. Thanks for the contribution.

PR #1234, opened 2019-11-06
Add SpeedSearch in GitHub PullRequest List

alex-yh99:
Currently PRList search queries through GitHub API, which might be slow for unstable network connection to GitHub.
Also keep consistency with other listview components which support SpeedSearch.


plugins/github/src/org/jetbrains/plugins/github/pullrequest/GHPRComponentFactory.kt
```diff
@@ -242,6 +242,8 @@ internal class GHPRComponentFactory(private val project: Project) {
       installSelectionSaver(it, listSelectionHolder)
       val shortcuts = CompositeShortcutSet(CommonShortcuts.ENTER, CommonShortcuts.DOUBLE_CLICK_1)
       EmptyAction.registerWithShortcutSet("Github.PullRequest.Timeline.Show", shortcuts, it)
+      val search = ListSpeedSearch(it) { item -> item.title }
+      search.comparator = SpeedSearchComparator(false)
```
iasemenov: You don't need to set this comparator, since it is already the default.

alex-yh99: Thanks

PR #1235, opened 2019-11-06
Let TypeMigrationProcessor handle primitive/non-primitive conversion

reda-alaoui:
Let be:
```java
public Integer foo() {
  return bar();
}

public int bar() {
  return 0;
}
```

When I ask TypeMigrationProcessor to migrate method `bar` return type from int to long, I end up with:
```java
public long foo() {
  return bar();
}

public long bar() {
  return 0;
}
```

`foo` return type is migrated from `Integer` to `long`. This is incorrect and will cause issues among which null pointer exceptions.

The current PR let TypeMigrationProcessor maintain the primitive/non-primitive aspect of type during the migration. e.g. Integer -> long migration will be turned to Integer -> Long before executing it. Although I think it should be the default behaviour, the new behaviour is introduced as an opt-in.


reda-alaoui:
Tests added

reda-alaoui:
Hello?

PR #1236, opened 2019-11-06
IDEA-226374 Constant String field evaluation shows PsiType prefix

bedla:
Hi, for details see https://youtrack.jetbrains.com/issue/IDEA-226374 .
Thx,
Ivos


akozlova:
Merged manually with small fixes

PR #1239, opened 2019-11-08
TypeMigrationStatementProcessor should process non array PsiNewExpression the same way it processes PsiMethodCallExpression

reda-alaoui:
Let be:
```java
class Foo {
  Foo(int i){

  }
}

class Bar {

  public Foo foo(){
    int i = 0;
    return new Foo(i);
  }
}
```

When triggering TypeMigration on `int i = 0` to migrate `int` to `long`, it ends up with the following result:
```java
class Foo {
  Foo(int i){

  }
}

class Bar {

  public Foo foo(){
    long i = 0;
    return new Foo(i); // Compilation error, cannot convert long to int
  }
}
```

`new Foo(i);` does not compile anymore.

This PR fixes it by processing non  array PsiNewExpression the same way as PsiMethodCallExpression.


reda-alaoui:
Tests added

reda-alaoui:
Hello?

PR #1241, opened 2019-11-08
Expose Invoker#useReadAction to StructureTreeModel

abrooksv:
We load data from the network in the background using a `StructureTreeModel`, this exposes the `Invoker`'s `useReadAction` so that the read lock is not captured during the background actions of the Tree / AsyncTreeModel


abrooksv:
@SergeyMalenkov Thoughts on this change?

SergeyMalenkov:
We've added another constructor that allows to specify Invoker instead of boolean value:
StructureTreeModel((structure, comparator, Invoker.forBackgroundThreadWithoutReadAction(parent), parent);

abrooksv:
Thank you Sergey. Your approach looks much cleaner

PR #1242, opened 2019-11-08
IDEA-226566: Use caching for large images

sprigogin:
Caching of large images is important to prevent memory waste due to having multiple copies of the same image.

### DIFF
--- a/platform/util/ui/src/com/intellij/ui/icons/ImageDescriptor.java
+++ b/platform/util/ui/src/com/intellij/ui/icons/ImageDescriptor.java
@@ -26,6 +26,8 @@ import java.util.concurrent.ConcurrentMap;
 @ApiStatus.Internal
 public final class ImageDescriptor {
   private static final ConcurrentMap<String, Pair<Image, ImageLoader.Dimension2DDouble>> ourCache = ContainerUtil.createConcurrentSoftValueMap();
+  private static final ConcurrentMap<String, Image> ourLargeImageCache = ContainerUtil.createConcurrentWeakValueMap();
+  private static final ConcurrentMap<Image, ImageLoader.Dimension2DDouble> ourLargeImageDimensionMap = ContainerUtil.createConcurrentWeakMap();
 
   final @NotNull String path;
   public final double scale; // initial scale factor
@@ -62,6 +64,8 @@ public final class ImageDescriptor {
 
   public static void clearCache() {
     ourCache.clear();
+    ourLargeImageCache.clear();
+    ourLargeImageDimensionMap.clear();
     ImageLoader.clearCache();
   }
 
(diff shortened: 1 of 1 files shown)


platform/util/ui/src/com/intellij/ui/icons/ImageDescriptor.java
```diff
@@ -26,6 +26,8 @@
 @ApiStatus.Internal
 public final class ImageDescriptor {
   private static final ConcurrentMap<String, Pair<Image, ImageLoader.Dimension2DDouble>> ourCache = ContainerUtil.createConcurrentSoftValueMap();
+  private static final ConcurrentMap<String, Image> ourLargeImageCache = ContainerUtil.createConcurrentWeakValueMap();
+  private static final ConcurrentMap<Image, ImageLoader.Dimension2DDouble> ourLargeImageDimensionMap = ContainerUtil.createConcurrentWeakMap();
```
jonnyzzz: That (second) map contains small objects on weak references. It is a possible case where these small objects are not collected along with the bigger values from the first map. Also, the more weak references are created the harder time would a GC have to deal with that.

It could probably be better to use a small immutable value object to keep both an image and it's dimensions on a hard reference and use the only one map with weak values instead

jonnyzzz:
We also have a code to cache loaded images in the disk, see the `com.intellij.util.SVGLoaderCache` class for more details. I'm not sure it's in the scope, but probably it would make sense to use something similar as well. It is assumed in the implementation that the cache is only used to the images from IDE resources

platform/util/ui/src/com/intellij/ui/icons/ImageDescriptor.java
sprigogin: Not sure I understand. The map(s) have to support lookup by a string key. This seems to imply that at least one map has to be keyed by string and contain strong key references. In order for this map not to leak memory, it has to have weak reference values. Weak reference values have to point to objects that are referenced elsewhere while in use. Neither Pair<Image, ImageLoader.Dimension2DDouble>, nor ImageLoader.Dimension2DDouble satisfy this condition. Do you mind elaborating on your proposal?

sprigogin:
> We also have a code to cache loaded images in the disk, see the `com.intellij.util.SVGLoaderCache` class for more details. I'm not sure it's in the scope, but probably it would make sense to use something similar as well. It is assumed in the implementation that the cache is only used to the images from IDE resources

SVGLoaderCache doesn't seem to be relevant to IDEA-226566 since background image may not be an SVG.

platform/util/ui/src/com/intellij/ui/icons/ImageDescriptor.java
jonnyzzz: Sorry for the delay in answering. 

I was about to suggest the following:
```
ConcurrentMap<String, Pair<Image, ImageLoader.Dimension2DDouble>> map = ContainerUtil.createConcurrentWeakValueMap();
```
Now I see it will unlikely work because a `Pair<>` object from map's value is only referenced from the weak value map and thus has a good chance to be collected before an Image object is collected itself. Your solution is fine

jonnyzzz:
I've merged the commit. Thank you, Sergey. It will show up in the `master` soon

PR #1243, opened 2019-11-11
IDEA-226683 - "Fix doc comment" action removes valid type param tags from JavaDoc

rendner:
This PR fixes IDEA-226683.


amaembo:
Merged.

amaembo:
Due to an unfortunate mistake I misattributed [several my commits](https://github.com/JetBrains/intellij-community/commits?author=rendner&since=2019-10-31&until=2019-11-19) to you. I'm sorry for this and hopefully, you are not very disappointed by this fact. Rewriting published history is not very cheap, so if you don't mind let's keep them as is.

rendner:
I don't have a problem with that.

PR #1244, opened 2019-11-12
Hides device selection from the configuration menu

dunno99:
Android Studio externally manages execution targets, so we hide all
execution target information from the configuration combo box.

We do this by adding support for externally managed targets, and filter
out UI elements that match externally managed targets.


dunno99:
@bulenkov

bulenkov:
Thank you. Merged with minor changes.

PR #1247, opened 2019-11-15
IDEA-227006 Associate stdout/stderr JUnit5 ReportEntry/reporting entries with proper test in IntelliJ as regular stdout/stderr

alshain:
See https://youtrack.jetbrains.com/issue/IDEA-227006


akozlova:
Thanks for the PR. I don't think that changing the order of entries in the entry is a good idea, so I just put everything to sout where it was before but ensure the test association. The fix would be available later today

PR #1249, opened 2019-11-16
Bad conversion of Ellipsis method parameter receiving a migrated array

reda-alaoui:
Let be

```java
public class Foo {
  public void bar(){
    int[] array = new int[]{0,0};
    baz(array);
    baz(0, 0);
  }

  public void baz(int... values){}
}
```

When asking type migration of array from `int[]` to `long[]`, the conversion fails.


reda-alaoui:
Tests added

reda-alaoui:
Hello?

PR #1251, opened 2019-11-16
Do not try to migrate string binary expressions

reda-alaoui:
Let be

```java
public class Foo {
    public void bar() {
        int number = 5;
        String query = "";
        query += number;
    }
}
```

When asking type migration of variable number from `int` to `long`, the migration fails because it fails to convert variable `query` from `String` to `long`.

This PR makes type migration consider those kind of assignment as migration compatible, i.e. not requiring any kind of conversion.


reda-alaoui:
Tests added

reda-alaoui:
Hello?

PR #1256, opened 2019-11-19
TypeMigration converts some method arguments without checking compliance with the parameter declaration type

reda-alaoui:
Let be:
```java
class Test {
    public void foo() {
        Integer.toString(migrationMethod());
    }

    private int migrationMethod() {
        return 1;
    }
}
```

When asking to migrate `migrationMethod` from `int` to `long` I obtain the following result:
```java
class Test {
    public void foo() {
        Integer.toString(migrationMethod()); // Does not compile
    }

    private long migrationMethod() {
        return 1;
    }
}
```
` Integer.toString(migrationMethod())` does not compile anymore.

This PR disable covariantPosition (if necessary) and adds a `OverloadedMethodsConversionRule` allowing to keep this kind of conversion working when a equivalent method accepting the new parameter type exists.


reda-alaoui:
Hello?

PR #1257, opened 2019-11-20
enhancement: javadoc inspection "unexpected param tag order"

rendner:
This PR adds a new javadoc inspection.

From the inspection description:

> Reports Javadoc @param tags if they have an unexpected order.
> The quickfix will put the @param tags in parameter-declaration order.
> Nonexistent @param tags will be created and invalid @param tags will be removed.

The main reason for creating this inspection was, that javadoc lists parameters in the order they are documented inside the javadoc comment.
However the IntelliJ "Quick Documentation" displays the parameters in the parameter-declaration order. Therefore the problem is only present/visible in the generated javadocs.


rendner:
The implementation doesn't handle [all possible cases](https://youtrack.jetbrains.com/issue/IDEA-228317#focus=streamItem-27-3856647.0-0).

PR #1258, opened 2019-11-21
When the migrated method parameter is bound to a super class type parameter, migrate the class type parameter

reda-alaoui:


reda-alaoui:
Hello?

reda-alaoui:
So much for that

PR #1259, opened 2019-11-22
Improve Method's param count evaluation

stsypanov:
I've investigated usage of `java.lang.reflect.Method.getParameterTypes();` and found that it's frequently called from parts of the application responsible for startup, indexing and PSI:

Each call to `Method.getParameterTypes()` clones underlying array, and in many cases the copy is used only to get it's length i.e. number of method params. So I propose to use `Method.getParameterCount()` which returns number of method's arguments without cloning of underlying array.

In some cases it also allows to make a decision about constructor's param count and if it's 0 then we can call `Constructor.newInstance()` without using vararg.

### DIFF
--- a/java/java-psi-api/src/com/intellij/codeInsight/AnnotationInvocationHandler.java
+++ b/java/java-psi-api/src/com/intellij/codeInsight/AnnotationInvocationHandler.java
@@ -41,8 +41,7 @@ class AnnotationInvocationHandler implements InvocationHandler {
 
   @Override
   public Object invoke(Object proxy, Method method, Object[] args) {
-    Class<?>[] paramTypes = method.getParameterTypes();
-    assert paramTypes.length == 0: Arrays.toString(paramTypes);
+    assert method.getParameterCount() == 0: Arrays.toString(method.getParameterTypes());
 
     String member = method.getName();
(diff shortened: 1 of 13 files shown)


stsypanov:
Any updates on this?

donnerpeter:
This fails quite a few tests, e.g. org.jetbrains.idea.devkit.actions.service.ServiceCreatorTest

stsypanov:
@donnerpeter hello, before pushing I've run ant's test config. Was it the wrong way?

donnerpeter:
I don't know what that does, sorry :( But now you know the name of a failing test. Does it run locally for you?

stsypanov:
@donnerpeter how do you run the tests?

I've tried to run mentioned `ServiceCreatorTest` and it failed, turned out the reason is a Kotlin related bug
https://youtrack.jetbrains.com/issue/KT-35627
https://youtrack.jetbrains.com/issue/KT-17433

stsypanov:
I've reverted breaking changes.

stsypanov:
@donnerpeter I've rerun all tests, now everything is OK.

donnerpeter:
Merged manually, thanks!

PR #1260, opened 2019-11-22
Fix Built Output Tool Window performance issues on projects with thousands of tasks

yuriy-solodkyy:
This PR address several problems in the build output tool window which significantly slow down the UI and overall build time especially for Android projects with many modules (above 200). This is extremely noticeable when all the Gradle tasks are up-to-date. 

(1) The dedicated thread updating the tree structure does not  keep up with the rate of events coming from Gradle because of acquiring the read lock for each event (and allocating a progress indicator. The  later problem is not addressed by this PR). 

(2) Running TreeUtil.expand() over the entire tree on each event (this operation allocates a lot of sync operations and becomes slower as the tree grows)

(3) CachingSimpleNode is "simple" and resets its cache of nodes on each change. It results in large arrays being re-allocated and a filter being applied to thousand of nodes on each event. 

See the commit descriptions for more specific details. 

@nskvortsov

### DIFF
--- a/platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
+++ b/platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
@@ -53,6 +53,7 @@ import com.intellij.ui.treeStructure.Tree;
 import com.intellij.util.EditSourceOnDoubleClickHandler;
 import com.intellij.util.EditSourceOnEnterKeyHandler;
 import com.intellij.util.containers.ContainerUtil;
+import com.intellij.util.containers.SmartHashSet;
 import com.intellij.util.text.DateFormatUtil;
 import com.intellij.util.ui.UIUtil;
 import com.intellij.util.ui.tree.TreeUtil;
@@ -116,8 +117,9 @@ public class BuildTreeConsoleView implements ConsoleView, DataProvider, BuildCon
     myFocusFirstError = !(buildDescriptor instanceof DefaultBuildDescriptor) ||
                         ((DefaultBuildDescriptor)buildDescriptor).isActivateToolWindowWhenFailed();
 
-    myRootNode = new ExecutionNode(myProject, null, true);
-    myBuildProgressRootNode = new ExecutionNode(myProject, myRootNode, true);
+    myRootNode = new ExecutionNode(myProject, null, true, this::isCorrectThread);
+    myBuildProgressRootNode = new ExecutionNode(myProject, myRootNode, true, this::isCorrectThread);
+    myRootNode.setFilter(getFilter());
     myRootNode.add(myBuildProgressRootNode);
 
     AbstractTreeStructure treeStructure = new MyTreeStructure();
@@ -138,8 +140,14 @@ public class BuildTreeConsoleView implements ConsoleView, DataProvider, BuildCon
     myThreeComponentsSplitter.setSecondComponent(myConsoleViewHandler.getComponent());
     myPanel.add(myThreeComponentsSplitter, BorderLayout.CENTER);
     BuildTreeFilters.install(this);
-    myRootNode.setFilter(getFilter());
     myOccurrenceNavigatorSupport = new ProblemOccurrenceNavigatorSupport(myTree);
+  }
+
+  private boolean isCorrectThread() {
+    if (myTreeModel != null) {
+      return myTreeModel.getInvoker().isValidThread();
+    }
+    return true;
   }
 
   private void installContextMenu(@NotNull StartBuildEvent startBuildEvent) {
@@ -172,10 +180,12 @@ public class BuildTreeConsoleView implements ConsoleView, DataProvider, BuildCon
 
   @Override
   public void clear() {
-    getRootElement().removeChildren();
-    nodesMap.clear();
-    myConsoleViewHandler.clear();
-    myTreeModel.invalidate();
+    myTreeModel.getInvoker().runOrInvokeLater(() -> {
+      getRootElement().removeChildren();
+      nodesMap.clear();
+      myConsoleViewHandler.clear();
+    });
+    scheduleUpdate(getRootElement(), true);
   }
 
   @Override
@@ -211,8 +221,10 @@ public class BuildTreeConsoleView implements ConsoleView, DataProvider, BuildCon
 
   private void updateFilter() {
     ExecutionNode rootElement = getRootElement();
-    rootElement.setFilter(getFilter());
-    scheduleUpdate(rootElement);
+    myTreeModel.getInvoker().runOrInvokeLater(() -> {
+      rootElement.setFilter(getFilter());
+      scheduleUpdate(rootElement, true);
+    });
   }
 
   private ExecutionNode getRootElement() {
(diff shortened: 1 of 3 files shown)


plugins/gradle/src/org/jetbrains/plugins/gradle/settings/GradleExtensionsSettings.java
```diff
@@ -116,28 +116,28 @@ public void add(@NotNull String rootPath, @NotNull Map<String, GradleExtensions>
...
-            }
-          }
-          if (task.getGroup() != null) {
-            description.append("<i>Task group: ").append(task.getGroup()).append("<i>");
-          }
-
-          gradleTask.description = description.toString();
-          extensionsData.tasksMap.put(gradleTask.name, gradleTask);
-        }
+        //for (ExternalTask task : gradleExtensions.getTasks()) {
```
nskvortsov: is this commenting out code intentional, or added to pull request by mistake?

yuriy-solodkyy:
This is unrelated. I forgot to undo the change when moving the patches
between repos. It only helps me it testing as it is a different performance
bottleneck on a large projects. I'll reupload later today when get back to
my desk.

nskvortsov:
Can we also split this pull request into different parts?

Let's review changes in `Subscribe for fewer Gradle events`  in a separate pull request, as they are somewhat separate.
And we can concentrate on Tree related features here

yuriy-solodkyy:
> Can we also split this pull request into different parts?
> 
> Let's review changes in `Subscribe for fewer Gradle events` in a separate pull request, as they are somewhat separate.
I've split the PR and rebased this one on top of the new master. I'll upload the Gradle events part and update this thread with the link.
> And we can concentrate on Tree related features here

yuriy-solodkyy:
PR with Gradle related changes: https://github.com/JetBrains/intellij-community/pull/1296

platform/lang-impl/src/com/intellij/build/ExecutionNode.java
```diff
@@ -215,16 +238,21 @@ public void setFilter(@Nullable Predicate<ExecutionNode> filter) {
...
-    if (myVisible != visible) {
-      myVisible = visible;
-      SimpleNode parent = getParent();
-      if (parent instanceof CachingSimpleNode) {
-        ((CachingSimpleNode)parent).cleanUpCache();
-      }
+  private void maybeReapplyFilter() {
+    if (myVisibleChildrenList != null) {
+      myVisibleChildrenList.clear();
+      myVisibleChildrenList.addAll(myChildrenList.stream().filter(it -> nodeIsVisible(it)).collect(Collectors.toList()));
```
nskvortsov: Using `forEachOrdered` may save you building a large intermediate list (e.g., for thousands of nodes):
`myChildrenList.stream().filter(it -> nodeIsVisible(it)).forEachOrdered(visibleChildrenList::add)`

platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
```diff
@@ -232,12 +244,17 @@ public void print(@NotNull String text, @NotNull ConsoleViewContentType contentT
   private ExecutionNode getOrMaybeCreateParentNode(@NotNull BuildEvent event) {
     ExecutionNode parentNode = event.getParentId() == null ? null : nodesMap.get(event.getParentId());
     if (event instanceof MessageEvent) {
-      parentNode = createMessageParentNodes((MessageEvent)event, parentNode);
+      ExecutionNode node = parentNode;
```
nskvortsov: Why do we need this variable here?

yuriy-solodkyy: It used not to be ExecutionNode originally. I'll remove it.

platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
```diff
@@ -495,30 +523,25 @@ private Runnable addChildFailureNode(@NotNull ExecutionNode parentNode, @NotNull
...
-            if (executionChild.isRunning()) {
-              finishChildren(executionChild, result);
-              executionChild.setResult(result);
-            }
-          });
-        }
+  private static void finishChildren(@NotNull SmartHashSet<ExecutionNode> structureChanged,
+                                     @NotNull ExecutionNode node,
+                                     @NotNull EventResult result) {
+    for (ExecutionNode child : node.getChildList()) {
```
yuriy-solodkyy: This needs to be copied before being iterated over since `child.setResult(result)` may update the list. We learned this from canary builds. I'll update the PR a little later.

nskvortsov:
We have reviewed the code, and it's good to go.
Let's clean up existing issues and I will merge the pull request.

platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
nskvortsov: ok )

platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
nskvortsov: thanks you

platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
yuriy-solodkyy: updated

platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
yuriy-solodkyy: removed

plugins/gradle/src/org/jetbrains/plugins/gradle/settings/GradleExtensionsSettings.java
yuriy-solodkyy: This change has been moved to a separate PR and reverted there.

platform/lang-impl/src/com/intellij/build/ExecutionNode.java
yuriy-solodkyy: Done

platform/lang-impl/src/com/intellij/build/BuildTreeConsoleView.java
yuriy-solodkyy: done

yuriy-solodkyy:
I believe I addressed all your comments. Let me know if I missed anything. 
Thank you!

nskvortsov:
Thank you, will send it to our CI pipeline now

nskvortsov:
Please check `BuildTreeConsoleViewTest` - looks like some description for empty nodes has changed

yuriy-solodkyy:
There are in fact two unrelated failures in the test execution results:
1. The root node toString() returns "null" instead of "" - fixing. 
2. Auto-expansion: I changed it from 2 levels to just 1 level (visible root(s)). The reasons are: (a) in Gradle we don't have anything under tasks now (if we can still have warnings/errors the first one will expand anyway) and (b) the existing code does a lot of work to maintain auto-expansion and while my version does less it is still expensive. 
3. However, I'm looking at whether auto-expand can further be simplified. I'll get back to you soon. I need to test it.

yuriy-solodkyy:
1. I restored the old way to decide whether a node needs to be auto-expanded and fixed "null" problem. The test is green for me.  I was wrong, I don't think it affects performance now. Re-uploaded.

2. I cannot simplify it. I missed some details.

nskvortsov:
Merged manually, thank you!

PR #1261, opened 2019-11-22
Pass along existing progress indicator if one exists.

brian-mcnamara:
Running a plugin in headless mode, this is overriding my progress bar I have configured to pipe the progress to the terminal. Can't think of a reason we cant use the existing one if it exists.

### DIFF
--- a/platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
+++ b/platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
@@ -859,7 +859,7 @@ public class ProgressIndicatorTest extends LightPlatformTestCase {
 
   public void testProgressIndicatorNotOverriddenIn_InRunProcessWithProgressSynchronously_duringWriteAction() {
     final String progressIndicatorText = "Progress indicator text";
-    ProgressIndicator progressIndicator = new ProgressIndicatorStub() {
+    ProgressIndicator progressIndicator = new ProgressIndicatorBase() {
       String text;
 
       @Override
@@ -870,11 +870,6 @@ public class ProgressIndicatorTest extends LightPlatformTestCase {
       @Override
       public String getText() {
         return text;
-      }
-
-      @Override
-      public void processFinish() {
-        //Prevent throwing runtimeException.
       }
     };
     ApplicationManager.getApplication().runWriteAction(() -> {
--- a/platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
+++ b/platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
@@ -856,4 +856,23 @@ public class ProgressIndicatorTest extends LightPlatformTestCase {
     catch (ProcessCanceledException ignored) {
     }
   }
+
+  public void testProgressIndicatorNotOverriddenIn_InRunProcessWithProgressSynchronously_duringWriteAction() {
+    ProgressIndicator progressIndicator = new ProgressIndicatorStub() {
+      @Override
+      public void processFinish() {
+
+      }
+    };
+    ApplicationManager.getApplication().runWriteAction(() -> {
+      ProgressManager.getInstance().executeProcessUnderProgress(() -> {
+        ApplicationManagerEx.getApplicationEx().runProcessWithProgressSynchronously(() -> {
+          assertFalse(progressIndicator.isCanceled());
+          ProgressManager.getInstance().getProgressIndicator().cancel();
+        }, "NotUsed", true, null);
+      }, progressIndicator);
+    });
+
+    assertTrue(progressIndicator.isCanceled());
+  }
 }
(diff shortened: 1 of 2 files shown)


donnerpeter:
Could you please add a test, probably, to `ProgressIndicatorTest`?

brian-mcnamara:
Thanks for taking a look Peter. I have added a test for the write action case, however (please correct me if there is a way), I could not find a way to test the headless && !unitTest case (the case I will use this for).

platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
```diff
@@ -856,4 +856,23 @@ private static void checkIndicatorNotThrowInThisOldStaleDisgustingNonCancelableS
     catch (ProcessCanceledException ignored) {
     }
   }
+
+  public void testProgressIndicatorNotOverriddenIn_InRunProcessWithProgressSynchronously_duringWriteAction() {
+    ProgressIndicator progressIndicator = new ProgressIndicatorStub() {
+      @Override
+      public void processFinish() {
```
donnerpeter: What's this processFinish for?

platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
```diff
@@ -856,4 +856,23 @@ private static void checkIndicatorNotThrowInThisOldStaleDisgustingNonCancelableS
...
+
+      }
+    };
+    ApplicationManager.getApplication().runWriteAction(() -> {
+      ProgressManager.getInstance().executeProcessUnderProgress(() -> {
+        ApplicationManagerEx.getApplicationEx().runProcessWithProgressSynchronously(() -> {
+          ProgressIndicator progressManagerIndicator = ProgressManager.getInstance().getProgressIndicator();
+          assertEquals(progressIndicator, progressManagerIndicator);
```
donnerpeter: I had a more behavioral test in mind, i.e. you set text on one indicator and check that the other has it as well. Then there would be other ways of fulfilling this requirement, if this one is found to be problematic for any reason.

platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
brian-mcnamara: Need to override it otherwise com/intellij/openapi/progress/impl/CoreProgressManager.java:169 throws a runtime exception.

platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
brian-mcnamara: Sure, I had a more behavioral test in my last commit, but thought it may not be easy to understand the test is ultimately ensuring the progress indicator is the one being sent into the executeProcessUnderProgress. Let me go back and rework it a bit.

platform/platform-tests/testSrc/com/intellij/openapi/progress/impl/ProgressIndicatorTest.java
donnerpeter: You could use a more normal indicator like ProgressIndicatorBase instead of this test-only one.

donnerpeter:
Great, thanks! Do you need this in 2019.3?

donnerpeter:
Interestingly, this leads to a failure in a couple tests in Ultimate. So far I haven't investigated why.

Assertion failed
	at com.intellij.testFramework.LoggedErrorProcessor.processError(LoggedErrorProcessor.java:56)
	at com.intellij.testFramework.TestLogger.error(TestLogger.java:40)
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:166)
	at com.intellij.openapi.diagnostic.Logger.assertTrue(Logger.java:180)
	at com.intellij.openapi.diagnostic.Logger.assertTrue(Logger.java:189)
	at com.intellij.openapi.progress.util.ProgressWindow.start(ProgressWindow.java:113)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:154)

brian-mcnamara:
> Great, thanks! Do you need this in 2019.3?

I'm in no rush, especially with the failures you mentioned.

> Interestingly, this leads to a failure in a couple tests in Ultimate. So far I haven't investigated why.

Weird, I did check the other tests and didn't see any new failures. I assume ultimate tests are internal only. Let me know if there is anything I can do to help.

Thanks again for your help looking at this.

donnerpeter:
So I've finally digged a bit deeper, and it turns out there's code that calls `progressIndicator.stop` prematurely, and after this change it does so on the outer indicator, and after that the next `runProcessWithProgressSynchronously` with the same outer indicator will try to `start` it again and on some indicators (e.g. `ProgressWindow`) that'll fail. Calling `stop` prematurely is most likely a bug, but there's an unknown amount of code doing that, and now they'd start getting quite cryptic errors. I don't have specific ideas on addressing that yet, just keeping you posted.

brian-mcnamara:
Thanks for the update! Was looking forward to hearing what you uncovered.

I'd be curious what the plugin is trying to do by calling stop on the EmptyProgressBar (My guess would be some kind of message passing). One way around would just be to have the plugin set the progress bar to an Empty progress bar. However I am curious how many plugins would run into this. My assumption was only a handful would be running headless and be using the empty progress bar.

Anyway thanks for looking into this more. Let me know if you have any ideas, Ill keep thinking on it, but I worry there is not much that can be done unless we add another gating mechanism (not saying that is the right thing todo).

donnerpeter:
It seems that an easiest way around is to still create a new indicator but propagate its text/fraction changes to the outer indicator. I can do this, but only if it's still needed: see, in master this code is no longer executed under headless mode, only from write action. Do you still need this functionality?

brian-mcnamara:
Oh nice! Removing headless instances from that branch will solve this issue for me. Thanks for bringing that to my attention. Closing this out since master has those changes.

PR #1262, opened 2019-11-22
Make getters public again, like their setters

dahlstrom-g:
Commit 3cf5f1d made getters package-private but left setters public
which triggers KT-20602 (“Cannot set synthetic property”).


dahlstrom-g:
@cdracm?

cdracm:
pushed

PR #1263, opened 2019-11-26
Remove no longer used getDetectedFileTypes

tgodzik:
Since https://github.com/JetBrains/intellij-community/commit/e9cd494b63ed53dc6cb4e9a81140431a49cd469c#diff-a2590eb4630da07c27f3010a7688c35b `getDetectedFileTypes` method in `FileTypeDetector`
is no longer being used. 

Despite that fact people might still try to implement it for their own FileTypeDetectors, which will be a waste of time and might lead to potential confusion when changes to the method are not reflected in how the plugins function. I think it's better to be completely explicit about the method removal.

This is just a proposal, so let me know what you think! This will cause compilation errors in plugins implementing the method, but I view it as better than just ignoring the method altogether.

This came up when investigating an issue in the pants intellij plugin: https://github.com/pantsbuild/intellij-pants-plugin/issues/431

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/fileTypes/impl/HashBangFileTypeDetector.kt
+++ b/platform/platform-impl/src/com/intellij/openapi/fileTypes/impl/HashBangFileTypeDetector.kt
@@ -31,4 +31,5 @@ open class HashBangFileTypeDetector @JvmOverloads constructor(
   }
 
   override fun getVersion() = _version
+
 }


klikh:

tgodzik:

Hi! I did submit it, but let me know if I mixed something up.

klikh:
Yep, I've found it now. Thanks.

donnerpeter:
Thanks for spotting the unused code that I missed! The fate of that change isn't fully decided yet, so I'd wait a bit with removing API that plugins implement and introducing source-level incompatibilities.

tgodzik:
> Thanks for spotting the unused code that I missed! The fate of that change isn't fully decided yet, so I'd wait a bit with removing API that plugins implement and introducing source-level incompatibilities.

Sure, I was just thinking that if we indeed go this way and that method is not used it would better to let know plugin authors by having it break existing plugins or at least deprecating the method.

tgodzik:
@donnerpeter I case you decide to stay with `getDetectedFileTypes` what would be a suggested way to implement `FileType` for files without extension for example Pants build tool? Pants define targets by creating `BUILD` files which contain python syntax for defining the target.

tgodzik:
@donnerpeter I see that the method is deprecated now and the code is remove. Thanks for the help!

PR #1264, opened 2019-11-26
IDEA-227649 LocalCanBeFinal: false negative in lambda body

t-kameyama:
https://youtrack.jetbrains.com/issue/IDEA-227649


akozlova:
Merged manually, thanks!

PR #1265, opened 2019-11-29
IDEA-227734: report "return null" statements inside non-null methods

rendner:
https://youtrack.jetbrains.com/issue/IDEA-227734
The `ReturnNullInspection` is enhanced to report `return null` statements inside non-null methods.


amaembo:
Sorry for the long delay. We will implement a proper alternative solution for the original problem, so pattern-based report will be unnecessary.

PR #1266, opened 2019-11-30
fix NPE in MacIntelliJButtonUI::getDarculaButtonSize

rendner:
`MacIntelliJButtonUI` overwrites `getDarculaButtonSize` and doesn't handle the the case that the dimension returned by `BasicGraphicsUtils.getPreferredButtonSize` is null
when "myButton.getComponentCount() > 0". More about this undocumented behavior can be found here: https://bugs.openjdk.java.net/browse/JDK-4694008.

The same fix was added for `DarculaButtonUI::getDarculaButtonSize` (https://github.com/JetBrains/intellij-community/commit/9372853553274c3c2d9431382b66403c337d1faf).

I couldn't test this fix because I don't have a mac available.
I found one report about this NPE here: https://github.com/exbin/bined-intellij-plugin/issues/22


AMPivovarov:
Sorry for the late reply,

Should be fixed in https://github.com/JetBrains/intellij-community/commit/9372853553274c3c2d9431382b66403c337d1faf

PR #1268, opened 2019-12-05
[Gradle] Always create tooling model copy for GradleExtensions.

chyingo:
With custom serialization, multiple projects can share the same instance
of tooling model after serialization, because the serialization mechanism
uses hashcode as identifier to avoid re-creating objects. While two projects
can have the same hashcode if their fields are identical.

This will cause problems if the shared tooling model is updated by one project,
and then all projects sharing that instance will get unintended change.

For example, with GradleExtensions model, each project adds their
own task list to the same instance of GradleExtensions.

Fix the problem by always making a copy of the de-serialized model.


nskvortsov:

PR #1269, opened 2019-12-05
IDEA-146200: fix for Unused Property inspection in gradle.properties files

62mkv:
This PR solves only one (but quite frequent) problem: false Unused Properties inspections in `gradle.properties` files, when the property is actually _used_ in `build.gradle` (or any other `.gradle`) file. It also fixes "Find Usages" behavior for such property definitions.

It *DOES NOT* solve: 
- Go to Declaration (from usages)
- Potentially over-defined symbols (say, when property `propA` is defined in `gradle.properties`, and then variable `propA` is also defined in a `.gradle` file.
- Nor does it resolve false "usages detection" if a symbol with the same name is defined in a Java code in a context that might imply property substitution (this is due to the nature of Properties plugin, it seems)

### DIFF
--- a/plugins/gradle/java/intellij.gradle.java.iml
+++ b/plugins/gradle/java/intellij.gradle.java.iml
@@ -35,5 +35,6 @@
     <orderEntry type="module" module-name="intellij.maven" />
     <orderEntry type="module" module-name="intellij.platform.objectSerializer.annotations" />
     <orderEntry type="module" module-name="intellij.platform.jps.model.impl" />
+    <orderEntry type="module" module-name="intellij.properties.psi" />
   </component>
 </module>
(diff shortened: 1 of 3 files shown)


klikh:

62mkv:
@klikh now I have (submitted a CLA)

nskvortsov:
Merged manually, thank you!

62mkv:
thanks @nskvortsov ! where can I see the changes in master branch? I can't figure it out yet.. :(

BasLeijdekkers:
You can find the changes here:
https://github.com/JetBrains/intellij-community/commit/b69a466631b01c408897e1b9f4159f6632470a0b
Because we don't use GitHub as our main repo, changes can take some time to show up.

PR #1270, opened 2019-12-06
IDEA-217968: Force DirectoryBasedStorage VirtualFile to be a directory

amaurymedeiros:
That will prevent an IOException such as '[...].idea/codeStyles' is not
a directory in VFS.


amaurymedeiros:
That fixes the Android Studio issue described in **https://youtrack.jetbrains.com/issue/IDEA-217968#focus=streamItem-27-3614125.0-0**

PR #1272, opened 2019-12-12
Fix CustomizedActionGroup refresh

nazmulidris:
Currently, if the underlying ActionGroup that is wrapped by a
CustomizedActionGroup changes after the CustomizedActionGroup
is created, this change is not recognized and the
CustomizedActionGroup will display stale Actions in the
toolbar.

The fix adds property change firings to DefaultActionGroup on
Action add/remove/reset/replacement/etc..., with the
CustomizedActionGroup acting as the property change listener.
Once the property change fires, the CustomizedActionGroup
refreshes its children by querying the DefaultActionGroup
again.

### DIFF
--- a/platform/platform-api/src/com/intellij/openapi/actionSystem/DefaultActionGroup.java
+++ b/platform/platform-api/src/com/intellij/openapi/actionSystem/DefaultActionGroup.java
@@ -30,6 +30,7 @@ import java.util.*;
  */
 public class DefaultActionGroup extends ActionGroup {
   private static final Logger LOG = Logger.getInstance(DefaultActionGroup.class);
+  public static final String PROP_CHILDREN_MODIFICATION_COUNT = "DefaultActionGroup.childrenModificationCount";
   private static final String CANT_ADD_ITSELF = "Cannot add a group to itself: ";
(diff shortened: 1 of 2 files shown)


dunno99:
The CL is just one way of going about fixing the issue. Basically, CustomizedActionGroup never gets notified that the underlying DefaultActionGroup has changed. I suspect prior to 2019.3 (your org's recent code optimization/cleanup effort), CustomizedActionGroup always gets initialized after Executors are created and registerd in the Run/Debug DefaultActionGroup. I'm guessing that because the recent release (some CL in that time frame) flips the order, it's causing this issue.

The PR CL basically just sends notifications to the CustomizedActionGroup if the underlying DefaultActionGroup changes. I didn't want to use the Schema change mechanism, since I believe that should be reserved for users modifying the customization. But if all know of a better way (perhaps changing/reverting the ordering or initialization), then go with that instead.

Vassiliy-Kudryashov:
This fix was merged to master branch with some changes: now DefaultActionGroup has own modificationTimeStamp and getter for it (as well as CustomActionsSchema has its own stamp) and we check these modifications in the same way. Thank you!

PR #1274, opened 2019-12-13
PY-14631 Use IPython config file in Python Console

msoxzw:
@Elizaveta239 
I seem to solve the [problem](https://youtrack.jetbrains.com/issue/PY-14631)
Could you help me examine if this PR indeed makes PyCharm load IPython config flie and possibly break other things?

### DIFF
--- a/python/helpers/pydev/_pydev_bundle/pydev_ipython_console_011.py
+++ b/python/helpers/pydev/_pydev_bundle/pydev_ipython_console_011.py
@@ -27,6 +27,7 @@ from IPython.core.usage import default_banner_parts
 from IPython.utils.strdispatch import StrDispatch
 import IPython.core.release as IPythonRelease
 from IPython.terminal.interactiveshell import TerminalInteractiveShell
+from IPython.terminal.ipapp import load_default_config
 try:
     from traitlets import CBool, Unicode
 except ImportError:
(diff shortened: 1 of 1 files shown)


Elizaveta239:
Thank you @msoxzw! I'll pass a build to our QAs for testing!

Elizaveta239:
Hi @msoxzw! Could you please elaborate a bit, which config parameters became supported after the fix? It looks like, for example, `c.InteractiveShellApp.exec_lines` isn't supported yet.
Which of them did you use for testing?

msoxzw:
The IPython integrated into PyCharm inherits the `TerminalInteractiveShell`:
https://github.com/JetBrains/intellij-community/blob/29b41da382e03cc65fd8b5b038f113b796a22c16/python/helpers/pydev/_pydev_bundle/pydev_ipython_console_011.py#L117
It is akin to [Embedding IPython](https://github.com/ipython/ipython/blob/master/IPython/terminal/embed.py), so the configuration which applies to Embedding IPython will also apply to this PR. Simply put, this config containing App keyword may be ignored just like Embedding IPython.
There is a method to set all the configuration: [Running IPython from Python](https://ipython.readthedocs.io/en/stable/config/intro.html#running-ipython-from-python), but it will cause severe issues.
By the way, the only option I use is [`c.HistoryManager.hist_file = ':memory:'`](https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-HistoryManager.hist_file). You could try the options without 'App' word, e.g. [`c.InteractiveShell.banner1`](https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-InteractiveShell.banner1), [`c.TerminalInteractiveShell.banner1`](https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-TerminalInteractiveShell.banner1)

msoxzw:
Launch a normal IPython:
```
import IPython
IPython.start_ipython()
```
Launch a embedded IPython:
```
import IPython
IPython.embed()
```

Elizaveta239:
Ok, I got it! If it fixes at least some use cases, it's already better than the previous version.
I've created a separate issue for mentioned use cases, you can follow it on YouTrack: https://youtrack.jetbrains.com/issue/PY-39795

vladimir-koshelev:
The issue has been fixed

PR #1275, opened 2019-12-17
Corrected code, when NIO methods silently failed for non-existing files

Vest:
Added additional checks to some classes/methods, where the file existance was checked with NIO to Path.toFile().exists().
This approach doesn't throw a silent exception and it is a little bit faster.

### DIFF
--- a/platform/bootstrap/src/com/intellij/ide/BootstrapClassLoaderUtil.java
+++ b/platform/bootstrap/src/com/intellij/ide/BootstrapClassLoaderUtil.java
@@ -13,6 +13,7 @@ import org.jetbrains.annotations.Nullable;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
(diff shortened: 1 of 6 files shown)


Vest:
Hello dear Developers,

I was analyzing the IntelliJ startup process with the debugger and I found out that there are several places (more than I corrected), where the used files might not exist. E.g. when you call the method Files.exists or Files.isRegularFile, it might throw an exception quietly.

Of course you don't see it in the debugger, if you do not have the option "catch handled exceptions". I analyzed your code and corrected it a little, so it will be exception free.

I did also a performance measurement with JMH, and if you are interested with the results, please check [my gist](https://gist.github.com/Vest/a43b0efa0f81accbffbbf695deae2654).

I decided to stop with the correction at some point, because I don't know if you appreciate micro-optimizations. If you like my commit, I will correct some other places, where (for example) try-with-resources can be used. Without redundant exceptions, I hope the startup process of IntelliJ will be a little bit faster.

p.s. I have signed your DocuSign document regarding the contribution.

Thanks,
Vest

Vest:
Half a year ago I posted this PR. And it seems it has conflicts now. Should I close and delete it, because you are probably busy?

PR #1276, opened 2019-12-17
Idea 229202

rendner:
Possible solution for https://youtrack.jetbrains.com/issue/IDEA-229202.


rendner:
The issue is already marked as fixed.

PR #1279, opened 2019-12-27
Update the MarkdownTest to include `<br>` outside of code blocks.

jwren:


jwren:
@alexander-doroshko

alexander-doroshko:
@jwren, I've restored `empty line -> <p>paragraph</p>` handling in dart docs, see https://github.com/JetBrains/intellij-plugins/commit/4b180675d93c2fe209d735f912cd9a5642440dd1.

Maybe, we don't need this change in `MarkdownUtil`?

jwren:
Yes, this looks sufficient, I experimented with the additional `split` parameters, I thought that I had added them at one point.

I thought that I was getting an actual empty line, but with and without the `<br>` in change here, I am now seeing the same representation in the markdown hover.

PR #1280, opened 2019-12-27
IDEA-229943: Enhance type inference for extracted variables

neshkeev:
`Intellij IDEA` used to traverse through all known `@FunctionalInterface`s and checks if their signatures match the examined method reference when extracting a local variable. It used to skip `@FunctionalInterface`s which have generic return types due to inexhaustibly cases in `TypeConversionUtil#isAssignable()`.

This patch adds a new case to check if the examined types are type parameters.

Signed-off-by: Nikita Eshkeev <kastolom@gmail.com>

### DIFF
--- a/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
+++ b/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
@@ -248,10 +248,10 @@ public class FunctionalInterfaceSuggester {
 
                 for (int i = 0; i < targetMethodParameters.length; i++) {
                   left[i + offset] = parameters[i + offset].getType();
-                  right[i + offset] = targetMethodParameters[i].getType();
+                  right[i + offset] = result.getSubstitutor().substitute(targetMethodParameters[i].getType());
                 }
 
-                left[parameters.length] = method.isConstructor() ? qualifierType : method.getReturnType();
+                left[parameters.length] = method.isConstructor() ? qualifierType : result.getSubstitutor().substitute(method.getReturnType());
                 right[parameters.length] = returnType;
 
                 final PsiSubstitutor substitutor = PsiResolveHelper.SERVICE.getInstance(project)
--- a/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
+++ b/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
@@ -248,14 +248,14 @@ public class FunctionalInterfaceSuggester {
 
                 for (int i = 0; i < targetMethodParameters.length; i++) {
                   left[i + offset] = parameters[i + offset].getType();
-                  right[i + offset] = result.getSubstitutor().substitute(targetMethodParameters[i].getType());
+                  right[i + offset] = targetMethodParameters[i].getType();
                 }
 
                 left[parameters.length] = method.isConstructor() ? qualifierType : result.getSubstitutor().substitute(method.getReturnType());
                 right[parameters.length] = returnType;
 
                 final PsiSubstitutor substitutor = PsiResolveHelper.SERVICE.getInstance(project)
-                  .inferTypeArguments(interface2Consider.getTypeParameters(), left, right, PsiUtil.getLanguageLevel(expression));
+                  .inferTypeArguments(interface2Consider.getTypeParameters(), left, right, result.getSubstitutor(), PsiUtil.getLanguageLevel(expression));
 
                 PsiType type = JavaPsiFacade.getElementFactory(project).createType(interface2Consider, substitutor);
 
--- a/java/java-psi-api/src/com/intellij/psi/util/TypeConversionUtil.java
+++ b/java/java-psi-api/src/com/intellij/psi/util/TypeConversionUtil.java
@@ -853,6 +853,12 @@ public class TypeConversionUtil {
              && rText.endsWith(lText)
              && rText.charAt(rText.length() - lText.length() - 1) == '.';
     }
+
+    final PsiClass leftClass = PsiUtil.resolveClassInType(left);
+    if (leftClass instanceof PsiTypeParameter) {
+      return true;
+    }
+
     return isClassAssignable(leftResult, rightResult, allowUncheckedConversion, left.getResolveScope(), capture);
   }
 
--- a/java/java-psi-api/src/com/intellij/psi/util/TypeConversionUtil.java
+++ b/java/java-psi-api/src/com/intellij/psi/util/TypeConversionUtil.java
@@ -853,12 +853,6 @@ public class TypeConversionUtil {
              && rText.endsWith(lText)
              && rText.charAt(rText.length() - lText.length() - 1) == '.';
     }
-
-    final PsiClass leftClass = PsiUtil.resolveClassInType(left);
-    if (leftClass instanceof PsiTypeParameter) {
-      return true;
-    }
-
     return isClassAssignable(leftResult, rightResult, allowUncheckedConversion, left.getResolveScope(), capture);
   }
 
--- a/java/java-psi-impl/src/com/intellij/psi/impl/source/tree/java/MethodReferenceResolver.java
+++ b/java/java-psi-impl/src/com/intellij/psi/impl/source/tree/java/MethodReferenceResolver.java
@@ -81,19 +81,9 @@ public class MethodReferenceResolver implements ResolveCache.PolyVariantContextR
                                                               final boolean staticProblem,
                                                               final boolean accessible,
                                                               final boolean varargs) {
-              PsiSubstitutor subst = substitutor;
               final PsiExpressionList argumentList = getArgumentList();
               final PsiType[] typeParameters = reference.getTypeParameters();
-
-              if (typeParameters.length > 0) {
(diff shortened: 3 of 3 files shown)


java/java-psi-api/src/com/intellij/psi/util/TypeConversionUtil.java
```diff
@@ -853,6 +853,12 @@ private static boolean isAssignable(@NotNull PsiType left,
              && rText.endsWith(lText)
              && rText.charAt(rText.length() - lText.length() - 1) == '.';
     }
+
+    final PsiClass leftClass = PsiUtil.resolveClassInType(left);
+    if (leftClass instanceof PsiTypeParameter) {
```
neshkeev: Currently I assume that I can assign any type to a type parameter. This solution allows `Intellij IDEA` add to the list of inferred types `Function`, `BiFunction`, (any `@FunctionalInterface` with generic return type)

This idea might not be appropriate, so I require guidance from the `jetbrains` team.

neshkeev:
I opened the draft PR to share my progress on the issue. Any feedback will be valuable.

java/java-psi-api/src/com/intellij/psi/util/TypeConversionUtil.java
akozlova: I am afraid, the approach is wrong: at very least type parameters can have bounds and such change would mean lots of bad code green/good code red problems. I suggest not to change this utility, cause changes here would influence the whole java type system

neshkeev: Thank you for your response, I do realize that this change is wrong, but it covers the case with generic types and I just need to figure out how to handle this case. Furthermore after this change IDEA starts suggesting more `@FunctionalInterface` types when a local variable is extracted (I described it in the issue). I assume there has to be a method that handles unification of generic types, it would be great if the `JetBrains` team guides me.
I would really like to complete the PR, because this feature is needed in my `Intellij IDEA` plugin.

akozlova:
You need to support better inference for the special case (`com.intellij.codeInsight.FunctionalInterfaceSuggester#suggestFunctionalInterfaces(com.intellij.psi.PsiMethod, boolean)`). I'd say that good test would help a faster fix

neshkeev:
@akozlova , here is a primitive solution just to share my progress. Please review it and help me make it better.

**There is a known issue (feature?)**: for the `Demo::<String, Integer>` method reference for the following class:
```java
public class Demo {
    public static<TO> TO process(int from) { return null; }
    public static<FROM, TO> TO process(FROM from1, FROM from2) { return null; }
    public static<TO> TO process() { return null; }
}
```
the list of suggested types are:
ctor
Even though there is only one method has two type parameters, which is convertible to `BiFunction`, the list of suggested types also contains `Function`.

If I change the method reference to `Demo::<String>process` I get the following list of types (both `Function` and `BiFunction` are present):
ctor

java/java-psi-impl/src/com/intellij/psi/impl/source/tree/java/MethodReferenceResolver.java
```diff
@@ -81,9 +81,19 @@ protected MethodCandidateInfo createCandidateInfo(@NotNull final PsiMethod metho
...
+              PsiSubstitutor subst = substitutor;
               final PsiExpressionList argumentList = getArgumentList();
               final PsiType[] typeParameters = reference.getTypeParameters();
-              return new MethodCandidateInfo(method, substitutor, !accessible, staticProblem, argumentList, myCurrentFileContext,
+
+              if (typeParameters.length > 0) {
+                final PsiTypeParameter[] methodTypeParameters = method.getTypeParameters();
+                final int until = Math.min(methodTypeParameters.length, typeParameters.length);
+                for (int i = 0; i < until; i++) {
+                  subst = subst.put(methodTypeParameters[i], typeParameters[i]);
```
akozlova: type arguments are processed in `getSubstitutor()` method, initial substitutor corresponds to the substitutor which is defined by the call site, like we call `List#add` on a list of `String`s, then call site substitutor should map `E (of java.util.List) -> String`.

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -248,10 +248,10 @@ else if (expression instanceof PsiMethodReferenceExpression) {
 
                 for (int i = 0; i < targetMethodParameters.length; i++) {
                   left[i + offset] = parameters[i + offset].getType();
-                  right[i + offset] = targetMethodParameters[i].getType();
+                  right[i + offset] = result.getSubstitutor().substitute(targetMethodParameters[i].getType());
```
akozlova: looks wrong to infer substitutor to suggest type which would be used to calculate the substitutor. 

Do you want to make use of type arguments provided by the reference? This is missed indeed. It's possible to create such substitutor right here (instead of MethodReferenceResolver) and pass it as `partialSubstitutor` to another overload of `inferTypeArguments` below. 

I suggest that you write tests for the changes, just add corresponding tests to `com.intellij.java.refactoring.FunctionalInterfaceSuggesterTest`

java/java-psi-impl/src/com/intellij/psi/impl/source/tree/java/MethodReferenceResolver.java
neshkeev: I removed this code since it is no longer needed

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
neshkeev: 1. I refrain from writing tests, because I don't know what the resulting list is going to be like yet since there are lots of `@FunctionalInterface`s and there are lots of combinations
2. I removed this substitution and added `result.getSubstitutor()` as `partialSubstitutor` to `com.intellij.psi.PsiResolveHelper#inferTypeArguments(...)` and seems to be working with a known issue (I will describe it below)

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -251,11 +251,11 @@ else if (expression instanceof PsiMethodReferenceExpression) {
                   right[i + offset] = targetMethodParameters[i].getType();
                 }
 
-                left[parameters.length] = method.isConstructor() ? qualifierType : method.getReturnType();
+                left[parameters.length] = method.isConstructor() ? qualifierType : result.getSubstitutor().substitute(method.getReturnType());
```
neshkeev: Is this correct? I noticed that if I change the code from [`InferenceSession`](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/source/resolve/graphInference/InferenceSession.java#L90)
```java
      final PsiType rightType = mySiteSubstitutor.substitute(rightTypes[i]);
      PsiType t = substituteWithInferenceVariables(leftTypes[i]);
      PsiType s = substituteWithInferenceVariables(rightType);
```
to
```java
      final PsiType rightType = mySiteSubstitutor.substitute(rightTypes[i]);
      final PsiType leftType = mySiteSubstitutor.substitute(leftTypes[i]);
      PsiType t = substituteWithInferenceVariables(leftType);
      PsiType s = substituteWithInferenceVariables(rightType);
```
I can revert this change here, will it be correct?

neshkeev:
The current solution contains the following known issue: **`Intellij IDEA` does not highlight errors**

**IMPORTANT**: It might be related to a different bug in `Intellij IDEA` itself, because the same code does not get highlighted for my current instance of `Intellij IDEA`

**Given:** Consider the following class (there are two overloaded method with the same name, but different number of arguments and type parameters):
```java
class Demo {
    public <TO> TO process(int from) { return null; }
    public <FROM, TO> TO process(FROM from1, FROM from2) { return null; }

        Demo demo = new Demo();
        Function<Integer, String> func = demo::<String, Integer>process;
    }
}
```
`Intellij IDEA` does not highlight the following line as a error:
```java
Function<Integer, String> func = demo::<String, Integer>process;
```
ctor

If the methods are static:
```java
class Demo {
    public static<TO> TO process(int from) { return null; }
    public static<FROM, TO> TO process(FROM from1, FROM from2) { return null; }

        Function<Integer, String> func = Demo::<String, Integer>process;
    }
}
```
`Intellij IDEA` highlight the line as a error:
ctor

akozlova:
I've pushed a fix for the initial problem (where explicit type arguments were ignored). Let's see if it would help with the rest. Thanks for the raising this up

neshkeev:
@akozlova , I really wanted to fix it myself, I have been working on it for more than a week and now it seems to be wasted.

akozlova:
Sorry, I just put in code my suggestion.
>looks wrong to infer substitutor to suggest type which would be used to calculate the substitutor.

>Do you want to make use of type arguments provided by the reference? This is missed indeed. It's possible to create such substitutor right here (instead of MethodReferenceResolver) and pass it as partialSubstitutor to another overload of inferTypeArguments below.

neshkeev:
@akozlova, the important part is that the best java IDE gets even better. I am glad this issue resolved quickly, I just hopped I would receive guidance in resolving the issue myself since it is not an urgent one.

PR #1281, opened 2019-12-31
Faster DebuggerUtilsImpl.readBytesArray

soywiz:
By doing an intellij extension for a custom Image library, I wanted to read the image payload.
Since I was not aware this was already done I did my own approach by serializing and deserializing to Base64, then simplified it to use the just `String` class + `getBytes` method.

Then noticed that byte reading was already implemented here. But looking at the implementation noticed that it was using the jdi `ArrayReference.getValues` that is creating a boxed item for every single element in the array, by requiring one instance per element it requires at least 8x the memory and potentially adds a lot of GC overhead.

This version serializes the byte[] into a String, then loads that string into the current process, and decodes it there. Since newer java versions can use byte[] for storing non-unicode strings, potentially could not use additional memory.

### DIFF
--- a/java/debugger/impl/src/com/intellij/debugger/ui/tree/render/ByteArrayAsStringRenderer.java
+++ b/java/debugger/impl/src/com/intellij/debugger/ui/tree/render/ByteArrayAsStringRenderer.java
@@ -24,7 +24,7 @@ final class ByteArrayAsStringRenderer extends CompoundReferenceRenderer {
           Value value = descriptor.getValue();
           if (value instanceof ArrayReference) {
             // TODO: read charset from the target vm
-            byte[] bytes = DebuggerUtilsImpl.readBytesArray(value);
+            byte[] bytes = DebuggerUtilsImpl.fastReadBytesArray(value, evaluationContext.getSuspendContext().getThread());
             if (bytes != null) {
               return new String(bytes, StandardCharsets.UTF_8);
             }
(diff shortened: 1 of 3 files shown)


ybedrov:
Hello Carlos,

soywiz:
Done

gorrus:
Thanks for the idea! For this particurar case I'd rather modify ImageSerializer methods to return Strings right away.

gorrus:
I filed and fixed https://youtrack.jetbrains.com/issue/IDEA-230367

java/debugger/impl/src/com/intellij/debugger/ui/tree/render/ByteArrayAsStringRenderer.java
```diff
@@ -24,7 +24,7 @@ public String calcLabel(ValueDescriptor descriptor, EvaluationContext evaluation
           Value value = descriptor.getValue();
           if (value instanceof ArrayReference) {
             // TODO: read charset from the target vm
-            byte[] bytes = DebuggerUtilsImpl.readBytesArray(value);
+            byte[] bytes = DebuggerUtilsImpl.fastReadBytesArray(value, evaluationContext.getSuspendContext().getThread());
```
gorrus: does not make much sence here as evaluation is not possible, so you won't be able to invoke String constructor

soywiz:
I guess you fixed this here, right? Since the class is internal, I guess that won't break anything. 
https://github.com/JetBrains/intellij-community/commit/fee7d3a3e0c050d0a590390b83dbe3ca0509455f

Glad the idea fixed an already opened issue :)

I'm closing the issue since I think that would do and I can always put this code in my own plugin. Feel free to reopen if it is not the case.

gorrus:
If you use this approach in your plugin, better use `com.intellij.debugger.engine.DebugProcess.newInstance` instead of direct invocation through the ClassType. This helps to avoid many issues with possible simultaneous invocations.

soywiz:
@gorrus Thanks for the info! Makes sense. I'll check that method :) 👍

PR #1284, opened 2020-01-07
IDEA-224062 Fixed unexpected paste behavior of Environment Variable table cell

targeter21:
The current behavior of pasting in the cell appends the copy buffer to the end of the cell contents instead of overwriting a highlighted section and was introduced in IDEA-210252. This PR resolves IDEA-224062 by making the cell behave like a normal text field. You can insert text at any point as well as remove a highlighted section with the copy buffer as expected.


targeter21:
Hi! Has there been any movement on this? This issue is still present in the current version of Intellij.

Vassiliy-Kudryashov:
Thank you for reminder, butactual fix is much easier as javax.swing.text.JTextComponent#paste does the job.

PR #1286, opened 2020-01-08
[IDEA-230303] Enhance type inference for constructor references

neshkeev:
This patch adds a test case to exhibit the problem from [the issue](https://youtrack.jetbrains.com/issue/IDEA-230303)
This is just a draft PR where I would like to share my progress

### DIFF
--- a/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
+++ b/java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
@@ -26,6 +26,7 @@ import com.intellij.psi.util.TypeConversionUtil;
 import com.intellij.util.Processor;
 import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.*;
 import java.util.function.Function;
@@ -230,23 +231,16 @@ public class FunctionalInterfaceSuggester {
         }
         else if (expression instanceof PsiMethodReferenceExpression) {
           List<PsiType> types = new ArrayList<>();
-          PsiType[] typeArguments = ((PsiMethodReferenceExpression)expression).getTypeParameters();
-          for (JavaResolveResult result : ((PsiMethodReferenceExpression)expression).multiResolve(true)) {
+          final PsiMethodReferenceExpression referenceExpression = (PsiMethodReferenceExpression) expression;
+          for (JavaResolveResult result : referenceExpression.multiResolve(true)) {
             final PsiElement element = result.getElement();
+            if (element == null) continue;
+
             if (element instanceof PsiMethod) {
               PsiMethod method = (PsiMethod)element;
-              int offset = hasOffset((PsiMethodReferenceExpression)expression, method) ? 1 : 0;
+              int offset = hasOffset(referenceExpression, method) ? 1 : 0;
               final PsiParameter[] targetMethodParameters = method.getParameterList().getParameters();
               if (targetMethodParameters.length + offset == parameters.length) {
-                PsiSubstitutor partialSubstitutor = PsiSubstitutor.EMPTY;
-                if (typeArguments.length > 0) {
-                  PsiTypeParameter[] typeParameters = ((PsiMethod)element).getTypeParameters();
-                  if (typeParameters.length == typeArguments.length) {
-                    Map<PsiTypeParameter, PsiType> substitutorMap = ContainerUtil.newHashMap(Arrays.asList(typeParameters), 
-                                                                                             Arrays.asList(typeArguments));
-                    partialSubstitutor = PsiSubstitutor.createSubstitutor(substitutorMap);
-                  }
-                }
 
                 final PsiType[] left = new PsiType[parameters.length + 1];
                 final PsiType[] right = new PsiType[parameters.length + 1];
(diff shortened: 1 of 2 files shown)


java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -230,24 +229,16 @@
...
             final PsiElement element = result.getElement();
+            if (element == null) continue;
+
             if (element instanceof PsiMethod) {
               PsiMethod method = (PsiMethod)element;
-              int offset = hasOffset((PsiMethodReferenceExpression)expression, method) ? 1 : 0;
+              int offset = hasOffset(referenceExpression, method) ? 1 : 0;
               final PsiParameter[] targetMethodParameters = method.getParameterList().getParameters();
               if (targetMethodParameters.length + offset == parameters.length) {
-                PsiSubstitutor partialSubstitutor = PsiSubstitutor.EMPTY;
```
neshkeev: I removed this part because I noticed that `result.getSubstitotor()` already contains the proper substitution map.

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -260,14 +251,16 @@ else if (expression instanceof PsiMethodReferenceExpression) {
 
                 for (int i = 0; i < targetMethodParameters.length; i++) {
                   left[i + offset] = parameters[i + offset].getType();
-                  right[i + offset] = partialSubstitutor.substitute(targetMethodParameters[i].getType());
```
neshkeev: This substitution is handled inside of `com.intellij.psi.PsiResolveHelper#inferTypeArguments()`

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -276,6 +269,20 @@ else if (expression instanceof PsiMethodReferenceExpression) {
                 }
               }
             }
+            else if (maybeDefaultConstructor(referenceExpression, parameters, element)) {
```
neshkeev: This code is meant to show the idea: since `MethodReferenceResolver#resolve` encapsulates the default constructor into [`com.intellij.psi.infos.ClassCandidateInfo`](https://github.com/neshkeev/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/source/tree/java/MethodReferenceResolver.java#L65) it cannot be handled uniformly, so this if-else-branch fixes the problem, but I require guidance from the **jetbrains** team to make it beter

neshkeev: This code is meant to show the idea: since [`MethodReferenceResolver#resolve`](https://github.com/neshkeev/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/source/tree/java/MethodReferenceResolver.java#L65) encapsulates the default constructor into `com.intellij.psi.infos.ClassCandidateInfo` it cannot be handled uniformly here, so this if-else-branch fixes the problem.
I require guidance from the **jetbrains** team to make it beter

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -260,14 +251,16 @@ else if (expression instanceof PsiMethodReferenceExpression) {
 
                 for (int i = 0; i < targetMethodParameters.length; i++) {
                   left[i + offset] = parameters[i + offset].getType();
-                  right[i + offset] = partialSubstitutor.substitute(targetMethodParameters[i].getType());
+                  right[i + offset] = targetMethodParameters[i].getType();
                 }
 
+                final PsiSubstitutor partialSubstitutor = result.getSubstitutor();
```
akozlova: you should not call `result.getSubstitutor()` here, because it relies on the context which is absent here. I've already told you that before.

neshkeev: Thank you for your comment, I didn't see your remark about the substitutor, I just went through [the previous PR](https://github.com/JetBrains/intellij-community/pull/1280) and did not find it there. I guess the GitHub interface is overcomplicated. I will bring the generation of the substitutor back.

neshkeev: @akozlova , I restored the in-place `PsiSubstitutor` generation for method references with type arguments, please review

neshkeev:
@akozlova, please review the PR

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -260,22 +253,28 @@ else if (expression instanceof PsiMethodReferenceExpression) {
 
                 for (int i = 0; i < targetMethodParameters.length; i++) {
                   left[i + offset] = parameters[i + offset].getType();
-                  right[i + offset] = partialSubstitutor.substitute(targetMethodParameters[i].getType());
+                  right[i + offset] = targetMethodParameters[i].getType();
                 }
 
+                final PsiSubstitutor partialSubstitutor = getPartialSubstitutor(referenceExpression, method);
+
                 left[parameters.length] = method.isConstructor() ? qualifierType : partialSubstitutor.substitute(method.getReturnType());
```
akozlova: partial substitution here is applied to the left type and when you pass partial substitutor to helper it would be applied to the `returnType` as well. Though I don't have a sample where it would lead to problems, I'd rather have previous behaviour as I don't see why it should be changed at the first place

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
```diff
@@ -260,22 +253,28 @@ else if (expression instanceof PsiMethodReferenceExpression) {
...
                 right[parameters.length] = returnType;
 
-                final PsiSubstitutor substitutor = PsiResolveHelper.SERVICE.getInstance(project)
-                  .inferTypeArguments(interface2Consider.getTypeParameters(), left, right, PsiUtil.getLanguageLevel(expression));
-
-                PsiType type = JavaPsiFacade.getElementFactory(project).createType(interface2Consider, substitutor);
+                final Optional<PsiType> type = getAcceptableType(interface2Consider, expression, left, right, partialSubstitutor);
```
akozlova: return an optional to add if present in both branches to the same collection, looks like an overcomplicating to me, just pass the collection and fill it when applicable

neshkeev: Thank you for your review.
I looked inside of `Optional#ifPresent` and it does the same thing: it checks if the value is not null and is so the `Consumer` is invoked. It seemed too me as a good idea to let the standard library do the job. But you might have different standards and conventions in your team, so I will get rid of `Optional` and bring back the original code.

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
neshkeev: Thank you for the review.
Are you talking about [this code](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/source/resolve/graphInference/InferenceSession.java#L90)?

akozlova: yes

neshkeev: Ok, I don't fully understand it, so I will trust your judgement. I will bring back the substitution of `right` back here and remove the `partialSubstitutor` from the argument list of `PsiResolveHelper#inferTypeArguments`

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
neshkeev: I removed `Optional` here

java/java-impl/src/com/intellij/codeInsight/FunctionalInterfaceSuggester.java
neshkeev: I removed `partialSubstitutor` from the argument list

neshkeev:
@akozlova, I amended the PR according to your comments, please review

akozlova:
I think it's in a good shape now, thanks! To be able to merge the PR I would need a CLA from you: https://www.jetbrains.com/agreements/cla.

neshkeev:
@akozlova I have squashed all the commits into a single commit, and now the PR is ready to be reviewed for the final time and merged if everything is ok.
I have signed the CLA

neshkeev:
@akozlova , please review the PR and merge it if possible

akozlova:
merged manually. thanks

PR #1287, opened 2020-01-09
Make map (Collection) always return a mutable list

Qumeric:
There is a fairly new linter warning and refactoring proposing to replace `collection.stream().map(...).collect(Collectors.toList())` to `ContainerUtil.map(collection, ...)`. 

This changes behaviour -> dangerous!

The problem is that `ContainerUtil.map` is supposed to return a read-only map, while `.collect(Collectors.toList())` returns a mutable map. 
However, actually `ContainerUtil.map` returns a mutable map in all cases, except when the given collection is empty. It makes situation even worse!

We should change either map's or linter's behaviour.

### DIFF
--- a/platform/util/src/com/intellij/util/containers/ContainerUtil.java
+++ b/platform/util/src/com/intellij/util/containers/ContainerUtil.java
@@ -2097,12 +2097,11 @@ public class ContainerUtil extends ContainerUtilRt {
   /**
    * @param collection an input collection to process
    * @param mapping a side-effect free function which transforms iterable elements
-   * @return read-only list consisting of the elements from the input collection converted by mapping
+   * @return a list consisting of the elements from the input collection converted by mapping
    */
   @NotNull
   @Contract(pure=true)
   public static <T,V> List<V> map(@NotNull Collection<? extends T> collection, @NotNull Function<? super T, ? extends V> mapping) {
-    if (collection.isEmpty()) return emptyList();
     List<V> list = new ArrayList<>(collection.size());
     for (final T t : collection) {
       list.add(mapping.fun(t));


Qumeric:
I suppose it is not enough to change only one function, but this commit/changelist mostly points to the problem.

amaembo:
I don't think we should do this. `Collectors.toList()` doesn't specify that the list will be mutable, so if you relied on this, it's a mistake in the first place (you should use `Collectors.toCollection(ArrayList::new)`). Making `ContainerUtil.map` returning a mutable list always sounds like a bad practice. It's a pity that it doesn't return an immutable list always, so I would instead wrap its result into `Collections.unmodifiableList`.

@cdracm what do you think?

Qumeric:
Hmm, okay, I thought that `Collectors.toList()` always returns a mutable list for some reason.

Maybe we should make map() and other methods always return immutable list then?

I wish Java had mutable and immutable versions of collections like Kotlin...

cdracm:
I also think that ContainerUtil.map() shouldn't return a mutable list, for performance reasons (to avoid allocating empty list and to avoid storing it in memory) and for code clarity reasons (what should happen when you'd remove element from the map() result? Should the original collection be modified too?)
It's not clear what "linter warning" we are talking about. If it's the "Settings|Editor|Inspections|Structural search|Can simplify to ContainerUtil.map" internal rule, then you can disable this altogether or in some scope to make it e.g. "suggest everywhere except in Android sources"

amaembo:
Returning always immutable list from ContainerUtil.map is a good idea, though it's somewhat risky. It's possible that some code already relies on the mutability of non-empty list returned from this method, so extensive testing would be necessary. In any case, this particular PR should be declined.

PR #1288, opened 2020-01-12
[IDEA-230486] Constructor references with type parameters are not recognized as PsiMethodReferenceExpression

neshkeev:
**`StatementParser`** used to ignore type parameters to constructor references and it prevents the tokens of such statements from being grouped under the **`PsiMethodReferenceExpression`** node in the PSI tree.

This patch adds a new check in **`StatementParser#parseStatement`** if the examined line is a constructor reference with type parameters.

Here is [the link to the issue](https://youtrack.jetbrains.com/issue/IDEA-230486)

### DIFF
--- a/java/java-psi-impl/src/com/intellij/lang/java/parser/StatementParser.java
+++ b/java/java-psi-impl/src/com/intellij/lang/java/parser/StatementParser.java
@@ -160,16 +160,31 @@ public class StatementParser {
       skipQualifiedName(builder);
       IElementType suspectedLT = builder.getTokenType(), next = builder.lookAhead(1);
       refPos.rollbackTo();
+
       if (suspectedLT == JavaTokenType.LT || suspectedLT == JavaTokenType.DOT && next == JavaTokenType.AT) {
+        final boolean constructorRef;
+
         PsiBuilder.Marker declStatement = builder.mark();
         PsiBuilder.Marker decl = myParser.getDeclarationParser().parse(builder, DeclarationParser.Context.CODE_BLOCK);
+
         if (decl == null) {
           PsiBuilder.Marker marker = myParser.getReferenceParser().parseType(builder, 0);
-          error(builder, JavaErrorBundle.message("expected.identifier"));
-          if (marker == null) builder.advanceLexer();
+          // if the type declaration ends with "::" then it is a method reference to a constructor
+          constructorRef = builder.getTokenType() == JavaTokenType.DOUBLE_COLON;
+          if (!constructorRef) {
+            error(builder, JavaErrorBundle.message("expected.identifier"));
+            if (marker == null) builder.advanceLexer();
+          }
         }
-        done(declStatement, JavaElementType.DECLARATION_STATEMENT);
-        return declStatement;
+        else {
+          constructorRef = false;
+        }
+
+        if (!constructorRef) {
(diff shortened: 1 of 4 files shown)


neshkeev:
I plan to add tests, I just need to figure out where to put them

java/java-psi-impl/src/com/intellij/lang/java/parser/StatementParser.java
```diff
@@ -160,7 +160,14 @@ else if (tokenType == JavaTokenType.IDENTIFIER || tokenType == JavaTokenType.AT)
       skipQualifiedName(builder);
       IElementType suspectedLT = builder.getTokenType(), next = builder.lookAhead(1);
       refPos.rollbackTo();
-      if (suspectedLT == JavaTokenType.LT || suspectedLT == JavaTokenType.DOT && next == JavaTokenType.AT) {
+
+      // skip the the type with possible type parameters to see if it is followed by "::"
+      myParser.getReferenceParser().parseType(builder, 0);
+      final boolean constructorRef = builder.getTokenType() == JavaTokenType.DOUBLE_COLON;
```
trespasserw: Statistically, a variable declaration is much more likely here, so it makes sense to check for double colon _after_ that attempt is failed, at line 168 (in original code).

java/java-psi-impl/src/com/intellij/lang/java/parser/StatementParser.java
```diff
@@ -160,7 +160,14 @@ else if (tokenType == JavaTokenType.IDENTIFIER || tokenType == JavaTokenType.AT)
...
+
+      refPos.rollbackTo();
```
trespasserw: After the first rollback (at line 162), the marker becomes invalid; you need to mark the position again.

neshkeev: I moved the check if the examined line is a method reference to a constructor inside of the `if`-statement below. This marker is no longer required

java/java-psi-impl/src/com/intellij/lang/java/parser/StatementParser.java
neshkeev: @trespasserw, Thank you for the review.
I moved the check if the examined line is a method reference to a constructor to the `if`-statement below as you suggested

neshkeev:
@trespasserw I amended the PR according to your comments, please review.

trespasserw:
That's better, thanks! The only thing I'd change is any mention of "constructor" to more generic "method".

neshkeev:
The difference in constructor references from method references in the shape: `

neshkeev:
As I see it method references differ from constructor references in the location of type parameters: `Foo::<Integer, String>methodName` vs `Foo<Integer, String>::new`

trespasserw:
Oh, sure, sorry.

neshkeev:
@trespasserw , I will squash the commits into a single one so it can be merged if you don't have any more objections

trespasserw:
👍

neshkeev:
@trespasserw I have squashed all the commits and prepared the PR to be merged, please view it and merge if possible

trespasserw:
Merged as 707933f. Thanks!

PR #1289, opened 2020-01-14
Do not override Scope Provider return type

brian-mcnamara:
We have a very unique dependency graph and need to override the default behavior of moduleWithDependentsScope and moduleTestsWithDependentsScope. Since ModuleWithDependentsTestScope is package protected, this change removes the need to override the return type and instead create a new helper method.

### DIFF
--- a/platform/indexing-impl/src/com/intellij/openapi/module/impl/scopes/ModuleScopeProviderImpl.java
+++ b/platform/indexing-impl/src/com/intellij/openapi/module/impl/scopes/ModuleScopeProviderImpl.java
@@ -44,6 +44,15 @@ public class ModuleScopeProviderImpl implements ModuleScopeProvider {
     return scope;
   }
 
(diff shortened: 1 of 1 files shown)


chashnikov:
What's different in your dependency graph? How do you change computation of these scopes?

brian-mcnamara:
Our dependency graph has references from source modules in libraries. So we need to have a scope to search dependents as well as select libraries. 

The plan is to create two new scopes for this use case and use the same logic from ModuleWithDependentsScope but also check for library code references. 

Taking a closer look this morning there need to be additional changes, I must have misread how the ModuleScopeProvider was set in ModuleImpl and this will not work out of the box. Let me poke around more tonight. 

Ideally I would like to use the ModuleScopeProvider to override the default scope.

brian-mcnamara:
Going to close this off, after more digging this will not be a good solution.

PR #1291, opened 2020-01-21
Fix default layout property being unset incorrectly

amaurymedeiros:
In TextEditorWithPreview, there is a property responsible for storing
the last used layout corresponding to that editor. We get this property
when creating new files to make sure they'll be open in the last used
layout. We're supposed to set it every time we select a layout other than
the default one (in this case, we reset it). However, we were hardcoding
Layout.SHOW_EDITOR_AND_PREVIEW to reset the property, while we're supposed
to be using the myDefaultLayout to avoid a misbehavior when opening new
files.


bulenkov:
Thank you, Amaury. Merged.

PR #1293, opened 2020-01-24
IDEA-231405 Fix `getGradleVersion` for wrapper distribution type

fedochet:
- The previous implementation was not using found gradle wrapper location to detect gradle version correctly; it used `last.used.gradle.home` property instead
- This caused plugins that used this function (like Kotlin Plugin) to incorrectly detect gradle version, which led to incorrect behavior (for example, generating `build.gradle.kts` with old API usages)
- ^IDEA-231405 Fixed


fedochet:
Related issue on YouTrack: https://youtrack.jetbrains.com/issue/KT-33401

PR #1296, opened 2020-01-29
Subscribe for fewer Gradle events

yuriy-solodkyy:
Subscribe for fewer Gradle events (~8x fewer for Android projects)

1. Replace missing events with synthetic events reconstructed from
    those we still subscribe to.

2. Replace partially implemented build phase detection with a simple
    based on text matching mechanism to update the text of the root node.

3. Do not unnecessary update the root node if nothing has changed.


vladsoroka:
Yuriy, thanks for the patch! This change also solves the problem when no progress information is shown during the download (e.g. dependencies download),
I've merged the PR using the following commit https://github.com/JetBrains/intellij-community/commit/9484352e989291da26fdeaf4925d9fb04a379624

I've found 1 regression with the change - it's displaying progress information during the Gradle wrapper distribution download.
image

I've added the fix, see the commit at https://github.com/JetBrains/intellij-community/commit/9abe34afa218ce0ae5e4539924fc1222f335a205

yuriy-solodkyy:
Thank you!

PR #1297, opened 2020-02-06
Pass current presentation mode status

srmurguia:
... when exiting distraction free mode.

https://youtrack.jetbrains.com/issue/IDEA-232296


srmurguia:
Hello!

 Do you have any comment on this pull request? Merging it would also solve a similar issue for Android Studio.

Thanks!

PR #1298, opened 2020-02-08
[IDEA-232518] The balloon panel of a breakpoint is not FocusCycleRoot

neshkeev:
The panel of Balloon used to define a FocusTraversalPolicy which was
supposed to define a custom focus traversal policy, but it did not do
the job very well: the TAB key was prevented from moving the focus to
elements of the Balloon panel.

This patch makes the balloon panel a FocusCycleRoot which makes it more
comfortable to use the keyboard and removes the custom focus traversal
policy since it is not needed anymore.

Here is the link to [the issue](https://youtrack.jetbrains.com/issue/IDEA-232518)

@amaembo , we discussed it, please take a look if possible
@denis-fokin , I refactored a bit your changes from [your earlier commit](https://github.com/JetBrains/intellij-community/commit/5ad3469deb14835c061bc38cf69c6ccf7da190ea#diff-af8399b3067ea60af13798ec90ed5c6b), please take a look


neshkeev:
I plan to add tests, I need to figure out how to do it for this case

neshkeev:
@denis-fokin , I hope you find time to take a look at the PR some day

Vassiliy-Kudryashov:
The issue https://youtrack.jetbrains.com/issue/IDEA-232518 was fixed in master (eda82b72)

PR #1302, opened 2020-02-13
Shell: Implement missing functionality to allow resolving and usage highlighting

jansorg:
This extends the grammar definition of `SFunctionDefinition`, `ShAssignmentCommand` and `ShAssignmentExpression` to implement `PsiNamedIdentifierOwner`. This is needed to allow implementations of reference resolving for Shell PSI elements, e.g. by 3rd-party plugins.

`getName()`, etc. are needed for highlighting references and the target element in the editor.

`setName` delegates to an ElementManipulator, the default implementation is unavailable (as before). The implementor of reference resolving is also able to implement rename by providing an element manipulator.

`ShLiteralExpression` now supports references in identifiers and `PsiNamedIdentifierOwner` for the `ShAssignmentExpression`. Shell arithmetic expressions contain variable definitions and references. This wasn't possible to resolve so far.

- I tried to follow the existing coding style and also applied the shared code style formatting
- I generated the ShPsi sources from the sh.bnf grammar, only needed changes are included
- Tests of the Shell plugin are passing

Please let me know if changes are needed.

### DIFF
--- a/plugins/sh/gen/com/intellij/sh/psi/ShAssignmentCommand.java
+++ b/plugins/sh/gen/com/intellij/sh/psi/ShAssignmentCommand.java
@@ -4,8 +4,9 @@ package com.intellij.sh.psi;
 import java.util.List;
 import org.jetbrains.annotations.*;
 import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiNameIdentifierOwner;
 
-public interface ShAssignmentCommand extends ShCommand {
+public interface ShAssignmentCommand extends ShCommand, PsiNameIdentifierOwner {
 
   @Nullable
   ShArrayExpression getArrayExpression();
@@ -18,5 +19,14 @@ public interface ShAssignmentCommand extends ShCommand {
 
   @Nullable
   PsiElement getPlusAssign();
+
+  int getTextOffset();
+
+  @Nullable
+  String getName();
+
+  PsiElement setName(String name);
+
+  PsiElement getNameIdentifier();
 
 }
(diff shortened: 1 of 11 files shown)


mmazurkevich:
Hello @jansorg. First of all, thank you for your contribution. Two days ago we have merged the implementation of the function resolve. I would like you to familiarize yourself with it. I also want to know in which points you want to extend `sh` plugin? E.g do you want your own function resolve?  And how you want to extend?
Regards your implementation, now it conflicts with our changes in the plugin. From my point of view, it's better to extend a separate class instead of creating a lot of methods in `ShPsiImplUtil`. Now the file looks weird: it has two getName, setName methods, it's not convenient to distinguish them. You can do the class extension for this task and move your methods there.
```
do_block ::= do  block_compound_list done {
  extends="com.intellij.sh.psi.impl.ShLazyDoBlockImpl"
}
```

jansorg:
Hi @mmazurkevich , thank you for your feedback.
I was planning to add support for sourced files and more advanced shell project setups, e.g. sets of shared or common files. The best way to achieve this would probably be to able to override the list of function references, e.g. via ˋShSupportˋ.  The method, which creates the array of the function ref + contributed references could just be moved to ShSupport for this, I think.

If that sounds fine to you, then I‘d update this pr with that and adapt to the style you suggested. 

The other changes are necessary to resolve variable references, once that‘s available. 

Regarding function resolve: I‘ll check out the implementation more closely with the next eap. Source file support will need matching, expanded usage and reference scopes. I‘d like to use the existing extension points for that and would check if that is used already.

mmazurkevich:
Hello @jansorg.  We have merged your changes. You can check the commit https://github.com/JetBrains/intellij-community/commit/aa530aeef6601dc2549aca95827879268f0c1d0f If there are no objections I think we can close review

jansorg:
@mmazurkevich Thank you for committing the changes and cleaning up parts of my PR. The commit seems fine to me. I'll test it and the function resolving soon -- next eap at the latest.

PR #1303, opened 2020-02-13
Shell: Don't show intention "Rename all occurrences" if the rename was turned off

jansorg:
A plugin may override the ShSupport service to disable the `Rename all occurrences` feature. The intention `Rename all occurrences` was still showing up, even if the rename was turned off. 
This PR fixes this issue.


jansorg:
@mmazurkevich It would be great if this could be merged. I wasn't sure if you knew about this PR.

PR #1304, opened 2020-02-17
Shell: support variable declarations outside of an assignment command

jansorg:
Fixes https://youtrack.jetbrains.com/issue/IDEA-231792

Many Bash commands declare a variable, but don't assign a value. These elements are plain `ShLiteral` PsiElements.
This pull requests adds support for `getName`, `setName`, and `getNameIdentifier` to allow proper variable resolve, highlighting and rename in 3rd-party plugins. Without this, variable references resolving to a definition of such a `ShLiteral` don't highlight properly and can't be renamed.

This PR also moves the instantiation of a ShFunctionReference into ShSupport to allow plugins to provide more sophisticated function resolving, e.g. to support include files. This is related to the handling of ShLiteral , but could be split into a seperate PR.

This PR isn't restricting this to a subset of commands, because it's easy to overlook a command which declares variables and this may also be different with older versions of Bash. This allows plugins to implement this without restrictions up-front.

Some examples, there are more:
```bash
# declares two vars without value, one with a value
export var1 var2 var3=name
# declares two read-only variables
declare -r var1 var2
# declares two read-only variables
readonly var1 var2
# declares two local variables (can only be used in functions)
local localVar1 localVar2

# declares loopVar
for loopVar; do ... ; done
```

Possible improvements in the Sh plugin, if requested:
- Check that a ShLiteral isn't the the name of a function definition. At the moment it's assumed that the 3rd-party plugin takes care of this.
- Move `getReferences` to the ShLiteralMixin, I didn't want to change the existing code

### DIFF
--- a/plugins/sh/gen/com/intellij/sh/psi/ShLiteral.java
+++ b/plugins/sh/gen/com/intellij/sh/psi/ShLiteral.java
@@ -4,9 +4,10 @@ package com.intellij.sh.psi;
 import java.util.List;
 import org.jetbrains.annotations.*;
 import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiNameIdentifierOwner;
 import com.intellij.psi.PsiReference;
 
-public interface ShLiteral extends ShSimpleCommandElement {
+public interface ShLiteral extends ShSimpleCommandElement, PsiNameIdentifierOwner {
 
   @Nullable
   PsiElement getWord();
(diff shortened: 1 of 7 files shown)


mmazurkevich:
Hi Joachim, Fix is already in the master branch https://github.com/JetBrains/intellij-community/commit/9e10cedd18871a56946d698669a5b02ef2c9a9d1 . Thanks for your contribution.  I think  PR and the related ticket can be closed.

jansorg:
@mmazurkevich thank you for merging this! Closing this PR now, but I don't have permissions to close the related ticket on YouTrack. You'll have to close it, I guess.

PR #1306, opened 2020-02-19
Use a Supplier to get EditorComposite's focusComponent

amaurymedeiros:
EditorComposite was setting the preferred focus component at construction
time by passing FileEditor#getPreferredFocusedComponent to the instance
of MyComponent it creates. Later on, when we request focus on the editor,
we will delegate to that component passed when creating MyComponent.

This would work fine if the preferred focus component never changes, but
that's not the case for some FileEditor implementations. For example, in
TextEditorWithPreview we return a different preferred focus component
depending on the current editor mode. In order to get the correct focus
component, we need to call FileEditor#getPreferredFocusedComponent again
whenever requesting focus on the EditorComposite.


amaurymedeiros:
It seems this was merged as https://github.com/JetBrains/intellij-community/commit/346c00ab4ed64e1eda3bc71e81440d32bb6c4237

Thanks!

PR #1307, opened 2020-02-21
[Gradle] Avoid ConcurrentModificationException

chyingo:
Use iterator to access task collection instead of iterating the collection directly, the later would
cause ConcurrentModificationException if the items are changed during iteration.


ybedrov:
Hello,

nskvortsov:
"For-each" loop is exactly the same as an explicit iterator definition that you suggest. This patch would change nothing.

PR #1308, opened 2020-02-23
[12 voted][IDEA-169034] Recognize and hyperlink project file path

blindpirate:
It's very common that language tools (e.g. lint tools) print
source file path on console, but IDEA can't recognize and
hyperlink them even when they're in the same project. This
commit makes it much more convenient when console outputs
source file paths by recognizing and hyperlinking them.


blindpirate:
This is how it looks like: 

image

Maven output is also enhanced:

image

I was intended to write some unit tests, can anyone please give me some guidance?

Note that if one line contains more than one source file path, it will only be hyperlinked once.

blindpirate:
CLA signed.

platform/lang-api/src/com/intellij/execution/filters/LineColumnNumberParser.kt
```diff
@@ -0,0 +1,53 @@
...
+        LineColumn.of(toLineColumnNumber(it.group(1)), toLineColumnNumber(it.group(2)))
+      }
+    }
+  },
+  COLON_COLON_COLON {
+    // Example: /tmp/my-project/src/main/kotlin/MyKotlin.kt:1:10:
+    // ktlint & checkstyle
+    // https://github.com/checkstyle/checkstyle/blob/a73ff0890c27cbe8affc1a55f2710a231687bd85/src/main/java/com/puppycrawl/tools/checkstyle/AuditEventDefaultFormatter.java#L43
+    // https://github.com/pinterest/ktlint/blob/4f764cd60b50c13403571e48f784b04bae4e63d7/ktlint-reporter-plain/src/main/kotlin/com/github/shyiko/ktlint/reporter/plain/PlainReporter.kt#L12
+    private val pattern = ":(\\d+):(\\d+):".toPattern()
```
Huangxuny1: `:(\d+):(\d*:?)`   may more compatible 

in [IDEA-169034] given example :
```
[ERROR] /Users/athos/Downloads/checkstyle-nolink/src/main/java/Example.java:9: Line matches the illegal pattern 'System\.(out|err).*?$'. [RegexpSinglelineJava]
```
the line number is `:line:`

my opinion :)

Huangxuny1: `:(\d+):(\d*):?`may more compatible
by given example in [IDEA-169034] 
it's output 
```
[ERROR] /Users/athos/Downloads/checkstyle-nolink/src/main/java/Example.java:9: Line matches the illegal pattern 'System\.(out|err).*?$'. [RegexpSinglelineJava]
``` 
pattern is `:line:`

so in `:(\d+):(\d*):?` , it may can compatible with two conditions

just my view :)

blindpirate: Thanks. According to https://github.com/apache/maven-checkstyle-plugin/blob/17c66b8824eef1c13837676716cf305740a830f0/src/main/java/org/apache/maven/plugins/checkstyle/CheckstyleViolationCheckMojo.java#L648 the column number is optional. I updated the regular expression and tested with following cases:

```
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:3:8: Unused import - java.util.concurrent.lock.Lock. [UnusedImports]
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:3: Unused import - java.util.concurrent.lock.Lock. [UnusedImports]
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:[3,8] Unused import - java.util.concurrent.lock.Lock. [UnusedImports]
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:[3] Unused import - java.util.concurrent.lock.Lock. [UnusedImports]
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:3:8:
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:3:
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:[3,8]
[ERROR] /Users/zhb/Projects/tmp/idea-hyperlink/src/main/java/MyClass.java:[3]
```

java/execution/impl/src/com/intellij/execution/filters/ProjectAbsoluteFileLocationFilter.kt
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @author Bo Zhang
+ */
+class ProjectAbsoluteFileLocationFilter(private val myProject: Project) : Filter {
+  companion object {
+    val FILE_ENDING_EXTENSION_PATTERN = "((\\.[\\w-]+)+)[^\\w-.]*".toPattern()
+  }
+
+  override fun applyFilter(line: String, entireLength: Int): Filter.Result? {
+    return myProject.basePath?.takeIf {
```
rendner, line 17: I expect that `myProject.basePath` will always return the same, so it would be good to evaluate this only once.

nskvortsov:
Thank you for the pull request.
Unfortunately, there is no plan to add global links highlighting to every console as the fix.
Also, please take a look at `com.intellij.execution.filters.RegexpFilter` class and it's usages - it already does a lot of things similar to your code. Specifically for Maven, it was adapted here: `org.jetbrains.idea.maven.project.MavenConsoleImpl#CONSOLE_FILTER_REGEXP`

PR #1310, opened 2020-02-28
Shell: id attributes to allow ordering

jansorg:
Adds id attribute values to most extensions registered by the Shell plugin.
This allows a 3rd-party plugin, e.g. BashSupport Pro, to apply an order to own extensions where needed.

- Synced to latest master
- Only attribute values were added (the GitHub diff isn't very good here)
- Tests of Shell are passing

cc @mmazurkevich


mmazurkevich:
The changes were merged to the master and will be available in the next EAP

PR #1312, opened 2020-03-03
Use System.lineSeparator() instead of System.getProperty("line.separator")

stsypanov:
According to JavaDoc of `System.lineSeparator()` it always returns the initial value of `System.getProperty("line.separator")`. Unlike the latter `System.lineSeparator()` returns the value of field while `System.getProperty("line.separator")` gets its value from `Properties` or `CHM` (depending on Java version).
If we take this benchmark
```java
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@BenchmarkMode(value = Mode.AverageTime)
@Fork(jvmArgsAppend = {"-Xms2g", "-Xmx2g"})
public class LineSeparatorBenchmark {

  @Benchmark
  public String getProperty() {
    return System.getProperty("line.separator");
  }

  @Benchmark
  public String lineSeparator() {
    return System.lineSeparator();
  }

}
```
then without contention it gives
```
Benchmark                                                 Mode  Cnt   Score    Error   Units
LineSeparatorBenchmark.getProperty                        avgt    5   8.246 ±  2.154   ns/op
LineSeparatorBenchmark.lineSeparator                      avgt    5   2.508 ±  1.143   ns/op
```
With contention `getProperty` becomes slower:
```
2 threads

Benchmark                                                 Mode  Cnt   Score    Error   Units
LineSeparatorBenchmark.getProperty                        avgt    5   8.890 ±  2.049   ns/op
LineSeparatorBenchmark.lineSeparator                      avgt    5   2.663 ±  0.326   ns/op

4 threads

Benchmark                                                 Mode  Cnt   Score    Error   Units
LineSeparatorBenchmark.getProperty                        avgt    5  11.539 ±  7.983   ns/op
LineSeparatorBenchmark.lineSeparator                      avgt    5   2.758 ±  0.198   ns/op

6 threads

Benchmark                                                 Mode  Cnt   Score    Error   Units
LineSeparatorBenchmark.getProperty                        avgt    5  16.271 ±  2.127   ns/op
LineSeparatorBenchmark.lineSeparator                      avgt    5   3.094 ±  1.001   ns/op
```
Itself
```java
System.getProperty("line.separator");
```
is not very widely used, but it is delegated from `SystemProperties.getLineSeparator()` so I think this change might be useful especially for the cases of concurrent use.

### DIFF
--- a/platform/platform-impl/src/com/intellij/ui/messages/SheetController.java
+++ b/platform/platform-impl/src/com/intellij/ui/messages/SheetController.java
@@ -62,7 +62,7 @@ public final class SheetController implements Disposable {
 
   private static final int GAP_BETWEEN_BUTTONS = 5;
 
-  @NonNls private static final String SPACE_OR_LINE_SEPARATOR_PATTERN = "([\\s" + System.getProperty("line.separator") + "]|(<br\\s*/?>))+";
+  @NonNls private static final String SPACE_OR_LINE_SEPARATOR_PATTERN = "([\\s" + System.lineSeparator() + "]|(<br\\s*/?>))+";
 
   // SHEET
   public int SHEET_WIDTH = 400;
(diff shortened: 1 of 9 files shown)


amaembo:
Please resolve the conflicts.

stsypanov:
@amaembo done!

amaembo:
Merged, thanks: 95fdba6befa29e5724d30764d68bf54f38fa7678

PR #1313, opened 2020-03-05
Fix support for CREATE_EMPTY_DIRECTORIES

yuriy-solodkyy:
When a source folder already exists in a content entry but not in
the file system, a duplicate source root is created.

This does not seem to happen in the way IDEA uses
CREATE_EMPTY_DIRECTORIES. We find it very useful in Android Studio
tests, but it produces duplicate source folders if importData
runs for an existing but empty project.

This change simply makes sure that if a directory should be created it is created 
before we test for its existence and not between the first test in findSourceFolder 
and the second test in addSourceFolder.

@vladsoroka


vladsoroka:
Thanks Yuriy! Merged.

PR #1314, opened 2020-03-07
[JUnit5] Added calculating LocationHint attribute via ClasspathResour…

sshuvalov:
…ceSource type.

In the process of creating a custom plugin became necessary to obtain the status of a test running on classpath source. See TestFailedLineManager


akozlova:
It's strange that this kind of location won't be processed by IDEA itself. Is it possible to use "file" protocol for that?

sshuvalov:
> It's strange that this kind of location won't be processed by IDEA itself. Is it possible to use "file" protocol for that?

In my case yes, it possible, but I was suprised that IDEA did not give me the opportunity to do whis with ClasspathResource.

sshuvalov:
Up. Same problem with original Cucumber framework - https://github.com/cucumber/cucumber-jvm/blob/main/junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureOrigin.java#L36

akozlova:
I am sorry, I still don't understand what issue this PR fixes. You changed the location on the tests side, IJ knows nothing about this protocol so IDE would ignore the location instead of navigation, correct? What advantages of a new location url do I miss then?

PR #1315, opened 2020-03-08
Change "ToggleBookmarkAction" to respect selected text

CeH9:
When add bookmark via "Ctrl + mouse1" it uses selected text as bookmark's name. 
At the same time if you toggle bookmark with "ToggleBookmarkAction" (default hotkey 
is "F11") it doesn't respect selected text. This patch fixes this issue.

image


Vassiliy-Kudryashov:
@CeH9 thank you! The request was merged to master (7d695949).

PR #1316, opened 2020-03-18
PY-41143: correct format for pydev_log call

loveatar:
Hi @avli 🙏 You forgot this occurrence while fixing similar cases. See 53bd8c07e5259407bb4029b73955c25522b4b520


avli:
Hi @loveatar,

The issue is already fixed (see: https://youtrack.jetbrains.com/issue/PY-41143) and the fix will be available in the next 2020.1 EAP.

Thank you for the PR anyway it's really appreciated 👍

PR #1317, opened 2020-03-19
Support RGBA hex color values in XML spell checker

gharrma:
Previously, only RGB color values were supported.


akozlova:
Sorry for the late response.

Could you please provide a sample rgba which cause spellchecker to warn? Thanks

gharrma:
From https://issuetracker.google.com/126612669:
> This line:
>        `<item android:offset="1" android:color="#00FFFFFF"/>`
>
> Produces this lint spelling error:
> Typo: In word 'FFFFFF'

As mentioned in the bug, this is something we can fix on the Android Studio side with our `AndroidXmlSpellcheckingStrategy`, but I thought it made sense to allow RGBA values upstream as well.

akozlova:
Thanks. Merged manually and added a test data

PR #1267, opened 2019-12-03
Remove final modifiers added by commit eb5cf4a

dahlstrom-g:
… which break tests that mock these classes. See related commit d8d2676.

PR #1318, opened 2020-03-20
Remove final keyword from ToolWindowHeadlessManagerImpl

dahlstrom-g:
This was added incidentally by commit 983b3e7
breaking test doubles for the class.


dahlstrom-g:
@develar

See related #1226 and #1267.

dahlstrom-g:
@develar?

develar:
Why do you need to mock ToolWindowHeadlessManagerImpl? Maybe it possible to modify class itself to avoid mocking? Or even simply fix some platform code to avoid usage of tool window manager?

dahlstrom-g:
I'm not the author and don't have answers to your questions, but you can refer to [commit 86f530b](https://android.googlesource.com/platform/tools/adt/idea/+/86f530b) and send a fix there instead if you prefer. To me, as in previous instances, it seems simpler to avoid adding the final modifier.

PR #1319, opened 2020-03-22
typo: String @NotNull [] -> @NotNull String[]

ArtsiomCh:


ArtsiomCh:
Found the same at `PsiFile` as well:
`  PsiFile @NotNull [] getPsiRoots();`
Maybe I didn't understand the intent of such syntax?... 
Please give me a qlue if such usage of `@NotNull` annotation have sence? Thank you in advance.

ArtsiomCh:
Got it. For those who has similar question:
https://stackoverflow.com/questions/29011893/how-to-annotate-an-array-nonnull
PS Its for eclipse.NonNull, but presume similar logic applies to the NotNull

ArtsiomCh:
By the way, Idea (193.6911.18) mark it as an error:
image

amaembo:
To annotate the array via TYPE_USE annotation (unlike array component), we should place the annotation before the brackets.

In IDEA 2020.1 we migrated our source base to TYPE_USE annotation package.
IDEA 2019.3 still uses non-TYPE_USE java5-compatible annotation package. That's why it's illegal to put NotNull before brackets when working with IDEA 2019.3 code base. So the behavior is expected.

PR #1322, opened 2020-03-24
Add a canRun overload with an executor parameter

juancnuno:
Add a canRun overload to ExecutionTarget with an executor parameter. So
Android Studio can disable the Debug and Profile toolbar buttons when
multiple devices are selected in the device drop down.


juancnuno:
@kradima You're the last one that touched ExecutionTarget.java. Could you please take a look? Thank you!

juancnuno:
We figured out a way to do what we wanted without changes to the IntelliJ platform. Thank you.

PR #1323, opened 2020-03-24
Fix bug: Cannot updating rendered document.

i-love-tadpole:
Fixed bug:
Sample code:
```java
package com.sample;

/**
 * before
 */
public class HelloWorld {
        System.out.println("hello world");
    }
}
```
Reproducible method:
1.Click "Toggle Renderd View" button,then the Hello class document is correctly rendered as "before".(Only the first click after openning the file will render document correctly).
2.Click  "Toggle Renderd View" button again and change document "before" to "after".like this:
```java
package com.sample;

/**
 * after
 */
public class HelloWorld {
        System.out.println("hello world");
    }
}
```
3.Click  "Toggle Renderd View" button again, You will see the incorrect rendered document "before".(Expect result is "after").

### DIFF
--- a/platform/lang-impl/src/com/intellij/codeInsight/documentation/render/DocRenderItem.java
+++ b/platform/lang-impl/src/com/intellij/codeInsight/documentation/render/DocRenderItem.java
@@ -269,6 +269,7 @@ class DocRenderItem {
         }
         foldingModel.removeFoldRegion(foldRegion);
         foldRegion = null;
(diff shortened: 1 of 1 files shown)


ybedrov:
Hello,

i-love-tadpole:
Yes.I am signed CLA

JB-Dmitry:
Fixed it in a little bit different way. Thanks for the report!

PR #1326, opened 2020-03-26
[Shell] IDEA-232910:  Support variables and sub-expressions in parameter expansions

jansorg:
- [x] Updated test data 
- [x] Tests are passing when build with current `master` and `201`
- [x] I tested this with popular bash scripts and couldn't find a regression

Most of these changes are generated lexer and parser source code and lexer and psi dumps for the test data.
The most important changes are in `Sh.flex` and `sh.bnf`.

The Shell plugin does not yet fully support parameter expansions.
Previously the lexer only supported nested expressions of the form `${}`. But all of `$()`, `${}`, `$var`, \`cmd\` are supported.
It also wasn't supporting the variable name, which is the first part of the expression, e.g. `${varName}` or `${#varName}`.

This pull request improves lexing and parsing of parameter expansion expressions. For example `${varName:-"my $(echo new)" value of '$varName'}` is now supported.

The first word is either a variable or a prefix string. Therefore it has to support references. 
It's now lexed as a `WORD` token to support references, not as `param_expression_body`.

This pull request tries to limit the changes to the returned tokens, therefore everything except the initial `WORD` is a `param_expression_body`.
The changes to the parser were also limited to a minimum. 
As with the previous implementation, highlighting of invalid expressions isn't fully supported.

Still missing, for a later PR:
- support for array references with references and arithmetic: `${arrayVar[expression]}`
- I noticed that `${varName:-$"content"}` isn't properly parsed. This happens because `$"content"` isn't supported by the lexer yet. I'll create an issue on YouTrack for this. This unsupported expression wasn't added to the lexer tests.


jansorg:
This was merged into master here: https://github.com/JetBrains/intellij-community/commit/acef853e72237199bd24717c50007222bda37d57#diff-b78ea8859e447852e8bf9972a5f556bc
Enhancements by JetBrains were added here: https://github.com/JetBrains/intellij-community/commit/e14315bed8f67636b0b7aee84f20f5be3cd47654#diff-b78ea8859e447852e8bf9972a5f556bc
Thank you!

PR #1330, opened 2020-03-30
fix javadoc emphasizing for Tree#setLeadSelectionPath

ArtsiomCh:
No markdown for javadoc (unfortunately).


stachenov:
This PR suddenly got some attention, but it's 6 years old and the code it modifies no longer exists. So I'll close it.

PR #1333, opened 2020-04-06
Fixed python console autocomplete by adding support for descriptors and metaclass members

smarie:
For metaclass members it is only done when the object to autocomplete is a class. 

Fixes https://youtrack.jetbrains.com/issue/PY-38151

Note: I used the following test to make sure autocompletion was ok:

```python
class MyDescriptor:
    def __get__(self, obj, objtype):
        if obj is None:
            # class-level call : return self
            return self
        else:
            # instance-level call: return field value
            def my_func(a, b):
                """hello there"""
                pass
            return my_func

class Foo2Meta(type):
    f: int = MyDescriptor()

class Foo2(metaclass=Foo2Meta):
    pass

class Foo:
    f: int = MyDescriptor()

foo = Foo()

# console autocompletion tests:
# - foo.f and Foo2.f should display as a TYPE_FUNCTION with arguments (a, b)
# - Foo.f and Foo2Meta.f should display as a TYPE_ATTR with no details (it is the descriptor)
```

### DIFF
--- a/python/helpers/pydev/_pydev_bundle/_pydev_imports_tipper.py
+++ b/python/helpers/pydev/_pydev_bundle/_pydev_imports_tipper.py
@@ -179,6 +179,17 @@ def generate_imports_tip_for_module(obj_to_complete, dir_comps=None, getattr=get
             dir_comps.append('__dict__')
         if hasattr(obj_to_complete, '__class__'):
             dir_comps.append('__class__')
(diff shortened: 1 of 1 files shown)


smarie:
Note that I did not include unit tests in the PR, as I was not able to find any for this method, but `generate_imports_tip_for_module` would definitely deserve some dedicated unit tests !

Elizaveta239:
Thank you very much for the contribution! Your pull request was merged and will be available in PyCharm 2020.2!

PR #1335, opened 2020-04-08
New default question word for boolean method name

alexengrig:
Add "non" boolean method name prefix to options for "**Java | Naming conventions | Boolean method name must start with question word**" inspection.
Usage example: `java.util.Objects#nonNull`.


akozlova:
Sorry for the late response.

I believe that adding a default question word would change the serialization of users profiles (where the inspection was enabled). In order to avoid that one need to write a lot of code actully, is it worth it?

alexengrig:
> Sorry for the late response.
> 
> I believe that adding a default question word would change the serialization of users profiles (where the inspection was enabled). In order to avoid that one need to write a lot of code actully, is it worth it?

This inspection is disabled by default, maybe it's not worth it. But I do not mind writing code.

akozlova:
Let's add a test first, e.g. add a test to `com.intellij.codeInspection.ex.InspectionProfileTest` which would read old default serial data
```
<inspection_tool class="BooleanMethodNameMustStartWithQuestion" enabled="true" level="WARNING" enabled_by_default="true">
      <option name="ignoreBooleanMethods" value="false" />
      <option name="ignoreInAnnotationInterface" value="true" />
      <option name="onlyWarnOnBaseMethods" value="true" />
      <option name="questionString" value="add,are,can,check,contains,could,endsWith,equals,has,is,matches,must,put,remove,shall,should,startsWith,was,were,will,would" />
    </inspection_tool>
```
perform empty modification, save profile and ensure that resulted xml is the same. 

Probably I was wrong and existing users won't see the change at all but we would have a test and that's always good.

Thanks

java/java-tests/testSrc/com/intellij/codeInspection/ex/InspectionProfileTest.java
```diff
@@ -618,6 +618,18 @@ public void testStoredMemberVisibility() throws Exception {
     assertEquals(mergedText, serialize(profile));
   }
 
+  public void testMergedBooleanMethodNameMustStartWithQuestionWithOldDefaultSerialData() throws Exception {
+    checkMergedNoChanges("<profile version=\"1.0\">\n" +
```
alexengrig, line 622: Not sure if it is correct.

alexengrig: > perform empty modification

For example such:

```java
profile.modifyProfile(model -> {});
```

alexengrig:
> perform empty modification

For example such:

```java
profile.modifyProfile(model -> {});
```

?

PR #1336, opened 2020-04-08
Fix the structureViewModel of markdown element

yaohui-wyh:
See the difference between the <kbd>FileStructure Popup</kbd> and the <kbd>Structure View</kbd>.

image

Fix the MkdElementInfo that is always expandable (set `isAlwaysShowsPlus()` return false)


TanVD:
Has been fixed in master a while ago.

PR #1337, opened 2020-04-09
Replace "Declare final" with bundle message

alexengrig:
In Java and Groovy.


akozlova:
Thanks, merged manually

PR #1338, opened 2020-04-10
Added retained and shallow sizes of classes to memory view

nikita-nazarov:


nikita-nazarov:
@bibaev , @gorrus , please take a look

nikita-nazarov:
Merged internally

PR #1339, opened 2020-04-13
Bro was found to be offensive in the Brotli algorithm

dandv:
See https://www.ghacks.net/2015/10/11/bro-file-extension-offensive-changed-to-br-instead/


TanVD:
Well, the excluded dictionary was moved to one of the internal libraries. We would consider adding this word there.

We are also working on infrastructure to report words, but right now I would recommend to use YouTrack issues for it.

PR #1341, opened 2020-04-15
Add introspections to english.dic

dandv:


YannCebron:
Thanks for your patches.

PR #1344, opened 2020-04-16
IDEA-219227 adding a system property to provide optional default gradle distribution type

alexlopashev:
This is a small change set to address issue mentioned in [IDEA-219227](https://youtrack.jetbrains.com/issue/IDEA-219227).

Initial idea was to adjust `org.jetbrains.plugins.gradle.service.settings.IdeaGradleSystemSettingsControlBuilder` UI component to render gradle distribution selection similar to `org.jetbrains.plugins.gradle.service.settings.IdeaGradleProjectSettingsControlBuilder`, but it turned out being pretty complicated and high risk change (to address update in message bus and listeners), as well as quite a refactoring to reuse component classes.

So instead I thought that just providing a new option system property to control default selection of distribution type which will be used if no distribution type is yet chosen, and it is absent - just fall back to previous behavior.


nskvortsov:
@alexlopashev thank you for the pull request. Unfortunately, it only provides a very local solution that will be hard to maintain in the future. We will keep working towards more product solution.

PR #1349, opened 2020-04-24
IDEA-238805 The classloader reference in a cached disabled icon prevents unloading of plugins

ingokegel:
`CachedImageIcon` has a field `myResolver` with the concrete type `MyUrlResolver`. That class has a field `myClassLoader` that is set to the plugin class loader. When the plugin creates a disabled icon for any reason, the plugin cannot be unloaded anymore.

In order to fix this, the existing `IconLoader.detachClassLoader` method is modified to remove the cached icon as well.


bibaev:
Fixed in https://github.com/JetBrains/intellij-community/commit/69ab1708450c8c429e49061fbe1bde94e6a0835c

PR #1353, opened 2020-04-30
PyCharm: add "View as ..." for geopandas GeoDataFrame/GeoSeries to the debugger

rendner:
## add "View as ..." for geopandas GeoDataFrame/GeoSeries to the debugger

Allows to display `geopandas` types `GeoDataFrame` and `GeoSeries` like the `panda` types `DataFrame` and `Series` in the `PyDataView`. 
See: https://www.jetbrains.com/help/pycharm/viewing-as-array.html

There is also a feature request: https://youtrack.jetbrains.com/issue/PY-31705

## Issues
- I couldn't add new unit tests to `PythonDataViewerTest`, because I haven't found the place where I need to add `geopandas`.
- I decided to re-use the labels `View as DataFrame`/`View as Series` for `GeoDataFrame`/`GeoSeries` in the debugger view. But if required I can add the labels `View as GeoDataFrame`/`View as GeoSeries`. 
- The changes in `pydevd_thrift.py` are not tested. I added these changes to be symmetrical with the changes in `pydevd_vars.py`. Is there a way to test the changes in `pydevd_thrift.py` with the Community Edition?
- I'm unsure if more changes are required in the `_pydevd_bundle` files to fully support `GeoDataFrame` and `GeoSeries` (for example the different `shapely.geometry` types).

### DIFF
--- a/python/helpers/pydev/_pydevd_bundle/pydevd_thrift.py
+++ b/python/helpers/pydev/_pydevd_bundle/pydevd_thrift.py
@@ -611,8 +611,13 @@ def header_data_to_thrift_struct(rows, cols, dtypes, col_bounds, col_to_format, 
     return array_headers
 
 
-TYPE_TO_THRIFT_STRUCT_CONVERTERS = {"ndarray": array_to_thrift_struct, "DataFrame": dataframe_to_thrift_struct,
-                                    "Series": dataframe_to_thrift_struct}
+TYPE_TO_THRIFT_STRUCT_CONVERTERS = {
+    "ndarray": array_to_thrift_struct,
+    "DataFrame": dataframe_to_thrift_struct,
+    "Series": dataframe_to_thrift_struct,
+    "GeoDataFrame": dataframe_to_thrift_struct,
+    "GeoSeries": dataframe_to_thrift_struct
+}
 
 
 def table_like_struct_to_thrift_struct(array, name, roffset, coffset, rows, cols, format):
--- a/python/helpers/pydev/_pydevd_bundle/pydevd_vars.py
+++ b/python/helpers/pydev/_pydevd_bundle/pydevd_vars.py
@@ -714,7 +714,13 @@ def is_able_to_format_number(format):
     return True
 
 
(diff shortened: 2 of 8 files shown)


avli:
Hello @rendner,

Thank you for the pull request! The changes look good to me, I'll take another look tomorrow and merge them to the upstream.

Answering your questions and notes:

1. Dependencies for env tests are installed with the Gradle `python.envs` plugin and should be added to [this file](https://github.com/JetBrains/intellij-community/blob/master/python/setup-test-environment/build.gradle). After that set up the test environments with `./gradlew build`, export the `PYCHARM_PYTHONS=<path to test envs>/pythons` environment variable and run tests with the IntelliJ IDEA. Note, that the path where you install the test environments is defined by the `PYCHARM_PYTHONS` variable from the Gradle file. It's a good idea to put the environments outside the sources because of indexing. If you don't want to bother with setting the things up, then don't worry, I can add tests by myself.
2. I think it's fine since geo data frames are data frames themselves and we directly use the data frame rendering machinery to show them.
3. `pydevd_thrift.py` is used for Python console communication. In order to test the changes run an example script in the PyCharm Python console. I have checked – everything works as expected.
4. Not sure what you mean here. Could you elaborate, please?

rendner:
Hi @avli,

many thanks for the feedback and explanations where I have to add the dependency. 

re-1)
I already managed to run the env tests but was unsure where to add geopandas. 
Honestly, I'm not sure if I have time to add the tests next week. If this is okay for you, you can add these. I expect that there should not so many tests since `GeoDataFrame` and `GeoSeries` extend the already tested `DataFrame` and `Series`. I will definitely look at your adjustments when you take over the tests. So that I can provide these myself next time.

re-3)
Wow so easy? I expected a more complex scenario. Will definitelly check this out to learn more about the internals.

re-4)
During my tests I came across an already known problem with [DataFrame/Series](https://youtrack.jetbrains.com/issue/PY-28279). The `PyDataView` can't display content if these objects contain tuples. Therefore I assumed that possibly other data types could also cause problems.

I had examined the tuple problem a few days later a little more closely. And could find the cause of the problem. If I have time today I will leave a comment in the issue. Otherwise the next few days.

PR #1354, opened 2020-05-01
IDEA-239871 Shell: Improve parsing of regular expression conditions

jansorg:
Closes https://youtrack.jetbrains.com/issue/IDEA-239871

The Shell parser and lexer don't support expressions in regular expressions of the form `[[ a =~ $pattern ]]`. This pull request improves the lexer and parser to handle this.

- Tests of the Shell plugin are passing (current master and 201) 
- Adds element types REGEX_CONDITION and REGEX_PATTERN and matching PsiElements. This was necessary because the right-hand-side of e.g. `a =~ a(${pattern})b` is now consisting of multiple tokens, which all form a single regular expression.
- The rhs is parsed as `regex_pattern ::= w+`, not as `regex_pattern ::= (w newlines | newlines w)+` as in `literal_condition`, because newlines within the pattern are syntax errors in Bash, afaik. Accepting newlines in `literal_condition` might be an existing bug, but I`m not sure yet.
- This PR tries to limit the changes to hopefully make it into 201.x. An improved version, e.g. for for 202.x, would add tokens for the `(` and `)` in the regular expressions tell that it's a group in a regular expression.


plugins/sh/src/com/intellij/sh/lexer/Sh.flex
```diff
@@ -248,14 +251,32 @@ EvalContent              = [^\r\n$\"`'() ;] | {EscapedChar}
...
-    "=~"                          { pushState(REGULAR_EXPRESSION); return REGEXP; }
+    "=~"                          { regexStart = getTokenEnd(); regexGroups = 0; pushState(REGULAR_EXPRESSION); return REGEXP; }
     "<"                           { return LT; }
     ">"                           { return GT; }
 }
 
 <REGULAR_EXPRESSION> {
-  {WhiteSpace}+                   { return WHITESPACE;}
-  {RegexInQuotes}                 { popState(); return WORD; }
+  "$(("                              { isArithmeticExpansion = true; yypushback(2); return DOLLAR; }
```
mmazurkevich, line 260: `REGULAR_EXPRESSION` duplicates `PARAMETER_EXPANSION`. I prefer to combine them

plugins/sh/tests/com/intellij/sh/parser/ShParserTest.java
```diff
@@ -40,6 +40,7 @@ protected String getTestDataPath() {
   public void testPipeline()      { doTest(true); }
   public void testCommands()      { doTest(true); }
   public void testRegex()         { doTest(true); }
+  public void testRegex2()        { doTest(true); }
```
mmazurkevich, line 43: We have test but we don't have corresponding `testData` and this test fails

mmazurkevich: If I use `regex2.sh` file for parser tests, the latest part of the file can't be parsed

plugins/sh/grammar/sh.bnf
```diff
@@ -370,16 +370,19 @@ condition ::=
...
+logical_and_condition ::= condition newlines '&&' newlines condition
+equality_condition ::= condition ('==' | '!=') condition
+regex_condition ::= condition '=~' regex_pattern
 comparison_condition ::= condition ('<' | '>') condition
 logical_bitwise_condition ::= '!' simple_command
 parentheses_condition ::= '(' condition ')'
-literal_condition ::= (w newlines | newlines w)
\ No newline at end of file
+literal_condition ::= (w newlines | newlines w)
+regex_pattern ::= w+
```
mmazurkevich, line 388: Why not make the `regex_pattern` as private one to avoid redundant tree nodes

plugins/sh/tests/com/intellij/sh/parser/ShParserTest.java
jansorg: My fault, I somehow managed to forget that file in the commit.  I've added the missing files.
The syntax error is unrelated to the changes of this PR. I've created https://youtrack.jetbrains.com/issue/IDEA-241708 to track this. I removed the linefeeds before `]]` to fix this for this pull requests until the YouTrack issue is fixed.

plugins/sh/grammar/sh.bnf
jansorg: I agree that the `regex_pattern` is redundant for single `w` tokens. If there's more than one element, e.g. as in `[[ "a b" =~ (a | b | c) ]]`, then I'd like to wrap them. Working with the PSI tree for this kind of structure would be quite difficult later, imo.
(A PSI element to wrap regex groups would be even better, but that's for a later improvement.)

I could try to update this to remove the `regex_pattern` for single `w` elements. Do you agree or would you prefer something different?

plugins/sh/src/com/intellij/sh/lexer/Sh.flex
jansorg: There are duplicate rules for `$((`, `$(`, `${`, `$[`, `{Variable}`, `{Quote}`, `{RawString}`, `{UnclosedRawString}. The other rules are different.

I could extract the shared rules into a seperate block above. Is that what you were suggesting?

jansorg:
@mmazurkevich Thank you for your review! I've rebased onto current master and have either addressed your notes or added questions.

I've also noticed a bug in my changes: whitespace inside of regex groups weren't properly accepted, e.g. for the whitespace around `|` in `[[ a =~ (a | b) ]]`.
To fix this I added `regexGroups <= 0` to [this line](https://github.com/JetBrains/intellij-community/pull/1354/files#diff-e91a61efadf8804c04c12879f1c7e337R279),

plugins/sh/src/com/intellij/sh/lexer/Sh.flex
mmazurkevich: Yes, I think we can do something like this:
```
< REGULAR_EXPRESSION > {
...
}
<PARAMETER_EXPANSION > {
...
}
< REGULAR_EXPRESSION , PARAMETER_EXPANSION > {
"$(("                              { isArithmeticExpansion = true; yypushback(2); return DOLLAR; }
"$("                               { pushState(PARENTHESES_COMMAND_SUBSTITUTION); yypushback(1); return DOLLAR; }
"${"                               { pushState(PARAMETER_EXPANSION); yypushback(1); return DOLLAR;}
...
}
```

plugins/sh/grammar/sh.bnf
mmazurkevich: If so, let's leave it as it is

plugins/sh/src/com/intellij/sh/lexer/Sh.flex
jansorg: Done. I had to move the shared rules before the other groups to keep the previous order how the rules are applied.

mmazurkevich:
@jansorg  Changes were merged into the master branch. Thank you for your contribution.

PR #1355, opened 2020-05-01
Shell: Use resolve scope provider instead of static scope

jansorg:
Use a resolve scope provider to provide the resolve scope to allow other extensions to use resolve scope enlargers. 
The current implementation returned a static scope and thus disabled the use of enlargers.


jansorg:
Thank you!

PR #1357, opened 2020-05-04
PY-41471: Verify that the plugin whitelist file exists before reading it

morguldir:
Spent a bit too long using git bisect to find, this, probably should have just searched for the file in the code and found the bug. Anyway this file is not created anywhere in community builds, and i assume you didn't catch it due to already having the file on your system. This PR just checks that the file exists before trying to read it.

I have tested ant build with a linux target in the python directory and it worked.


ybedrov:
Hello!

morguldir:
Yes, i have signed it electronically.

DmPanov:
@morguldir thanks for your contribution and your patience! 
The issue is already addressed [here](https://github.com/JetBrains/intellij-community/blob/3c09e18373c599e545631648eebc9a9abb1d9fd6/platform/build-scripts/src/org/jetbrains/intellij/build/impl/plugins/PluginAutoPublishList.kt#L27) so no need to merge it now. Sorry for such a delay, this PR was effectively lost indeed.

PR #1358, opened 2020-05-05
fix typo

jinwookh:
Com_pare -> Compare


amaembo:
This is not a typo but a mnemonic character marker. It should not appear in IntelliJ user interface. If it appears for you somewhere, please [file](https://youtrack.jetbrains.com/) a ticket and post the screenshot.

PR #1359, opened 2020-05-05
Fix NPE in StartupUtil involving eulaDocument

gharrma:
Fixes IDEA-240150


anstarovoyt:
Thank you for creating this pull request and for your contribution to the IntelliJ platform. We would like to apologise for not acting on it sooner.

The original issue was fixed as part of IDEA-242037 so I am closing the PR.

PR #1361, opened 2020-05-06
PY-42270 Don't detach when 'inspect' flag is set

tonyroberts:
When handling uncaught exceptions, don't detach the debugger, exit
the process, or reset the excepthook when sys.flags.inspect (Py_InspectFlag)
is set.


tonyroberts:
Link to issue: https://youtrack.jetbrains.com/issue/PY-42270

PR #1362, opened 2020-05-09
Update reademe with JavaFX requirement

MariusVolkhart:
Indicate that JavaFX is a requirement and that the Jetbrains Runtime JDK is suitable for building Intellij.

At this time, some of the popular JDK build don't include JavaFX in their JDKs anymore, AdoptOpenJDK being the biggest. Additionally, OpenJFX standalone packages are only available for versions 11 and 14.

While solutions continue to exist outside the JBR, such as Amazon Corretto, or Azul JDK FX, they are hard to find.


chashnikov:
I've added a note about that to the readme file.
> such as Amazon Corretto, or Azul JDK FX, they are hard to find.

Note that since 2020.1 IDEA allow you to download some JDK distributions (including Amazon Corretto, for example) directly from Project Structure dialog.

PR #1363, opened 2020-05-09
PyCharm: add prefix-f-string-intention

rendner:
## Intention for f-strings

This intention prefixes a string literal with `f` to make it an f-string literal (for Python versions >= 3.6). 

The caret must be placed within the string or directly after the closing quotation mark and the string must meet the following conditions:

<ul>
<li>is not a docstring</li>
<li>not prefixed with "f" or "F"</li>
<li>not prefixed with "u" or "U"</li>
<li>not prefixed with "b" or "B"</li>
<li>contains at least one unescaped "{"</li>
<li>terminated with the same type of quotes it begins with</li>
</ul>
  
I added several test cases for this new intention.

There are also some requests for such a feature: 
- https://youtrack.jetbrains.com/issue/PY-30411
- https://youtrack.jetbrains.com/issue/PY-32484


rendner:
I am not treating escaped `{`, I'll add that tonight.

rendner:
Intention now checks if there is an unescaped `{` inside the string literal.

python/python-psi-impl/src/com/jetbrains/python/codeInsight/intentions/PyPrefixFStringIntention.kt
```diff
@@ -0,0 +1,58 @@
...
+  override fun doInvoke(project: Project, editor: Editor, file: PsiFile) {
+    finStringElementAtCaret(file, editor)?.apply {
+      val expression = PyElementGenerator.getInstance(project).createStringLiteralAlreadyEscaped("f${text}")
+      replace(expression.stringElements.first())
+    }
+  }
+
+  override fun isAvailable(project: Project, editor: Editor, file: PsiFile): Boolean {
+    if (file !is PyFile) return false
+    finStringElementAtCaret(file, editor)?.let { it ->
```
east825, line 41: It's a good idea to perform a fail-fast `PyStringElement#textContains('{')` check, as you did initially, before accessing complete element text and matching it by a regex.

east825:
@rendner Honestly, I'm in doubts about this particular implementation, even though I understand a need for such action. The problem is that it's inherently ambiguous. There are two distinct scenarios for converting an existing ordinary string literal into an f-string. 

In the first one you _initially_ want an f-string. You start typing a literal, open a curly brace, see no syntax highlighting or code completion and then realize that you forgot to prefix the string with "f". You want a quick action to add it without moving the caret back and forth. It might look like that:

```python
print('value = {<caret>')
```

Your handy intention takes care of it perfectly well.

In the second scenario you already have a non-interpolated string literal but _then_ you want to embed some expression inside. You, again, need to convert it to an f-string without all the fiddling with the caret. It might look like this:

```
import re

REGEX = re.compile(r'\w{3-5} {<caret>')
```

These two are actually quite different because in the second case you also want to _escape existing curly braces_ inside the string. Already available "Convert to f-string literal" intention does that, and it seems too easy to start confusing them. 

The logic regarding the presence of a non-escaped opening brace is understandable but, I believe, also won't be clear for people who want to convert existing literals that not yet contain an incomplete  expression fragment to f-strings with this new intention. 

In general, I favor the proposal about the unification "Convert to f-string literal" so as to make it available for any string literal ([PY-32484](https://youtrack.jetbrains.com/issue/PY-32484)), but, again, it will then escape the brace that one has just typed to insert an expression. The need to remove this extra brace every time means it won't handle the first case as naturally as your version.

What do you think on this whole matter?

rendner:
@east825 thank you for the very detailed illustration of the two cases. Your point about confusing the user with the new intention seems legit. I also had this in mind when I started the implementation - that a user could expect sometimes more than this little prefix.

> The logic regarding the presence of a non-escaped opening brace is understandable but, I believe, also won't be clear for people who want to convert existing literals that not yet contain an incomplete expression fragment to f-strings with this new intention.

Good point. I added this behavior because of allowing the user to do the following:
`x = 'My name {name}'<caret>`
But it doesn't make that much sense, since most of the time the user will apply this intention while writing the string and not afterwards.

> In general, I favor the proposal about the unification "Convert to f-string literal" so as to make it available for any string literal

Sounds great. Making the already existing `PyConvertToFStringIntention` available for all string literals would be great and much more useful for most users. Especially when existing curly braces are automatically escaped if needed. I assume that the advantages of having a good way to convert string literals into f-string literals outweigh the problem with the escaped last typed brace. And maybe there are some simple ways to get rid of that problem (e.g. replacing the brace temporary with an unique placeholder before converting the string).

python/python-psi-impl/src/com/jetbrains/python/codeInsight/intentions/PyPrefixFStringIntention.kt
rendner: Make sense, and with your input I would even remove the brace-check completely.

east825:
> And maybe there are some simple ways to get rid of that problem (e.g. replacing the brace temporary with an unique placeholder before converting the string).

It's definitely one of the ways to implement such selective escaping, but the problem is _which_ brace we should omit. If we added a special case for the brace, say, immediately preceding the caret, it still would not be ideal because someone converting an existing string literal can put the caret anywhere inside before invoking the action and rightfully expect the intention to behave the same regardless of the caret position. For instance, in the example above the result should be identical for both

```
import re

REGEX = re.compile(r'\w{<caret>3-5}')
```
and
```
import re

REGEX = re.compile(r'\w{3-5}<caret>')
```
because having the caret after a brace doesn't necessarily mean that the latter belongs to an expression fragment, even that you're going to type anything there at all. 

The larger underlying problem here, I believe, is distinguishing between the cases "I'had been typing this literal and stopped midway to invoke an intention" and "I just put a caret inside this literal to invoke an intention".

On a different note, don't you want to pack this handy intention as a plugin and publish it in the plugin repository or, perhaps, just share it with other fellows in the ticket before the proper unification of "Convert to f-string" is implemented?

rendner:
@east825 
I had some time to think about the unification. Trying to not escape the "last typed" curly brace is not a good idea, as you already mentioned.

I found another case that seems to make the unification of these intentions impossible:

_User has already finished the string literal and wants to convert it into an f-string literal._
1) 
before: `"the value of {variable.name} is {variable.value}"`
expected: `f"the value of {variable.name} is {variable.value}"`
2) 
before: `"{{Hello}}, {variable.name} = {variable.value}"`
expected: `f"{{Hello}}, {variable.name} = {variable.value}"`

In these examples the "Convert to f-string literal" intention shouldn't escape any of the curly braces.

> On a different note, don't you want to pack this handy intention as a plugin and publish it in the plugin repository or, perhaps, just share it with other fellows in the ticket

I found some time and released the intention as an [IntelliJ Plugin](https://plugins.jetbrains.com/plugin/14336-python-prefix-f-string-intention).

east825:
> I found some time and released the intention as an IntelliJ Plugin. I also linked the plugin in PY-30411.

Awesome, thanks! 

Since the functionality was extracted into a plugin so far, I'm closing this PR. Thanks again for sharing the idea!

> I found another case that seems to make the unification of these intentions impossible:

My hope here is that if someone intended to write an f-string in the first place, they would invoke the intention as soon as they start typing the first expression fragment and notice missing highlighting/completion. So there just won't be existing fragments that they would want to exclude from escaping. This assumption doesn't eliminate the problem for preceding _already escaped_ curly braces as with `{{Hello}}` in your example, but I'm planning to add highlighting for these too, so that the absence of "f" prefix was immediately noticeable as well.

PR #1364, opened 2020-05-12
IDEA-204711 IDEA-204333 Add the missing plexus-utils jar for Maven 2

floating-cat:
The newest version of the plexus-utils jar is 3.3.0, but the `plexus_utils` from `.idea/libraries` is still using the 3.2.0. So I use the 3.2.0 version here.
The commit is mostly based on 821231c0.

Related issue links:
1. https://youtrack.jetbrains.com/issue/IDEA-204711
2. https://youtrack.jetbrains.com/issue/IDEA-204333


floating-cat:
I can't reproduce this problem anymore. So close it currently.

PR #1365, opened 2020-05-12
Use reflection in Gradle daemon actions

srmurguia:
... to macth with the available DaemonParamters constructor, depending
in the Gradle version.

This is causing issues similar to https://b.corp.google.com/issues/148656285
on projects that use Gradle versions greater than 6.3 and lower than
6.0.


vladsoroka:
Thanks for the contribution! Merged.

PR #1366, opened 2020-05-12
Toolwindow save bounds

jlauridsen:


jlauridsen:
IDEA-240699 Save ToolWindow position when closing Window.
    
ToolWindowManagerImpl was not saving the bounds of a ToolWindow
with a Window decorator if the was not showing.
This condition was added because of IDEA-147099 because the
code used to call Component.getLocationOnScreen which is no
longer used to compute the bounds.
    
Removing this condition will enable the bounds of a Window being
closed to be saved as expected.

stachenov:
The issue is no longer reproducible. I left a comment in the YT ticket.

PR #1367, opened 2020-05-18
Update top border

tatakir:


ForNeVeR:
This is another part of implementation for [IDEA-241054](https://youtrack.jetbrains.com/issue/IDEA-241054), I'll manage it from now.

(see also: https://github.com/JetBrains/JetBrainsRuntime/pull/18)

@tatakir, thanks for contribution!

ForNeVeR:
Current status: we're still unsure if we're okay with undocumented API in JBR, thus the PR is still under investigation.

I'll try to make the decision in about a month from now.

Sorry for the pause!

ForNeVeR:
No longer actual in modern IntelliJ: we already support full border theming on Windows 11.

Thank you so much for your help on this anyway.

PR #1368, opened 2020-05-18
[IDEA-241080] Fix issue with 'I do mind' showing up as path

erihe587:
Link to issue: https://youtrack.jetbrains.com/issue/IDEA-241080

Renamed an oddly named variable and added a method that checks if a given path equals the placeholder path which is assigned when the user clicks "No, thanks". Also added some logic for when to display the error message that notifies the user that the path is invalid, as well as when to show the path-string (which is what caused the original issue).


ybedrov:
Hello!

erihe587:
> Hello!

It should be submitted now.

mmazurkevich (changes requested):
Thank you @erihe587 for your contribution.
I'd prefer to add a separate checkbox above this input if it's enabled disable all other fields and doesn't show `I do mind` in this input anymore.  In such case, user will be able to enable shellcheck again.
image

PR #1370, opened 2020-05-21
some map iteration optimisations at todo/FileTree.java

ArtsiomCh:
Just in case you find it useful. I have to do it in my code (copied from yours) anyways.
More optimisations could be possibly done as I didn't fully get needability of those assertTrue checks.


maxmedvedev:
It's not immediately obvious to me that this patch makes performance better: there are fewer calls to `myDirectory2Children`, but more calls to `children`. So I'd like to see performance snapshots before and after the change

maxmedvedev:
I'll gladly reopen the PR if/when some proof of performance improvement is added.

PR #1372, opened 2020-05-21
Document requirement for JavaFX in JDK

russelldavis:
Without JavaFX, the build fails with errors like "java: package com.sun.javafx.embed does not exist".


chashnikov:
I've added a note about that to the readme file, thank you.

PR #1377, opened 2020-05-29
Reverse the logic to enable the inspection configurator

tonybaloney:
The implementation of this class has some logic which is the wrong way around.

The `isApplicable()` method will return `false` if the sdks list **is empty**, which the opposite of what you'd expect (and I think the intention of that guard clause).

Second, the configure environment will go and detect Python SDKs, but only if the list of SDKs **is empty**, which I think is backward.

I've tested this on a CLI environment and could never get the original project configurator to pass


traff:
Hi Anthony!

The logic here was the following: if there is no SDK configured yet, we check whether the project contains Python files and if so, then configure an SDK. Next time we don't need to configure it once again and we treat the project as Python project thereafter. Does it make sense? Maybe I am missing something. What was the problem you were fixing by this commit?

PR #1378, opened 2020-05-29
Fix Kotlin models deserialization

yuriy-solodkyy:
`emptySet()` cannot be cast to `MutableCollection<>` and `ClassUtil.isMutableCollection(result)` happens too late.

### DIFF
--- a/platform/object-serializer/src/CollectionBinding.kt
+++ b/platform/object-serializer/src/CollectionBinding.kt
@@ -76,7 +76,7 @@ internal class CollectionBinding(type: ParameterizedType, context: BindingInitia
     }
 
     @Suppress("UNCHECKED_CAST")
-    var result = property.readUnsafe(hostObject) as MutableCollection<Any?>?
+    var result = property.readUnsafe(hostObject) as? MutableCollection<Any?>?
     if (result != null && ClassUtil.isMutableCollection(result)) {
       result.clear()
       if (emptyResult != null) {


yuriy-solodkyy:
@vladsoroka

develar:
Looks like that we don't serialize empty set properly, I added test and it confirms that. I will fix it this week.

```
Index: community/platform/object-serializer/testSrc/ListTest.kt
IDEA additional info:
<+>UTF-8
--- community/platform/object-serializer/testSrc/ListTest.kt	(revision bcca1c44c7e1676edd8f8d50c1ee978776df2e09)
+++ community/platform/object-serializer/testSrc/ListTest.kt	(date 1592409878966)
@@ -100,6 +100,19 @@
     assertThat(deserializedBean.list).isSameAs(emptyList<String>())
   }
 
+  @Test
+  fun `empty set as empty set on read`() {
+    class TestBean {
+      @JvmField
+      var field: Set<String>? = null
+    }
+
+    val bean = TestBean()
+    bean.field = emptySet()
+    val deserializedBean = test(bean)
+    assertThat(deserializedBean.field).isSameAs(emptySet<String>())
+  }
+
   @Test
   fun `parameterized array`() {
     class TestBean<T> {

````

yuriy-solodkyy:
Note that it crashes when de-serializing a **NON**-empty collection into an object initialised with emptyList() since it first attempts to cast it to a mutable collection and only then tests whether it is mutable.

nskvortsov:
I can see your suggestion was applied by [this commit](https://github.com/JetBrains/intellij-community/commit/94f7837a33c8daeb00c5e5f7a89d9d65ed684212#diff-8aa6d580f18ae829109a6621ee8b0616)
so I will label this as merged and close the request.

PR #1380, opened 2020-06-04
IDEA-242521 Fix HotSwap skipping cached gradle tasks in delegated build

gaiazov:
When build is delegated to Gradle, some tasks may run as FROM-CACHE. These tasks are currently ignored by the HotSwap code.

Cached tasks can still output modified classes, so HotSwap should deploy those changes.


vladsoroka:
Thanks for the contribution, I've added the integration test for the change and merge it using the following commit https://github.com/JetBrains/intellij-community/commit/9d58fbaa9d203f343069ea6a7b6f1e44fabdbbf8

PR #1381, opened 2020-06-06
IDEA-242757: Fix parsing of Shell Script redirects

jansorg:
This fixes/improves parsing of redirections in Shell script.
https://youtrack.jetbrains.com/issue/IDEA-242757

Notable changes:
- `com.intellij.sh.parser.ShParserUtil#parseUntilSpace`: Before the rule was accepted even if nothing matched. `true` should only be returned if at least one token was consumed.
- The lexer returned a single here-string redirections token `<<<       ` (with a whitespace suffix), but that should be `<<<` and `<whitespace>`. I've updated the lexer to handle this.
- I've added a new private rule for redirection targets, to improve the error messages
- I've turned off token accessors for `ShRedirection`, because there were a lot of changes here and all seem useless to me. The methods are still not very useful, but still better than the unneeded token accessors.


plugins/sh/src/com/intellij/sh/lexer/Sh.flex
```diff
@@ -29,6 +29,7 @@ import com.intellij.openapi.util.text.StringUtil;
   private boolean isQuoteOpen;
   private String heredocMarker;
   private boolean heredocWithWhiteSpaceIgnore;
+  private int herestringStart = -1;
```
mmazurkevich, line 32: I renamed it to `herestringStartPosition` to understand clearly what it stores.
I also added the reset of this variable at the `onReset()` method.

PR #1386, opened 2020-06-09
Improve hotswap class detection when build is delegated to Gradle

gaiazov:
### What is the problem
When IntelliJ is configured to delegate the build to Gradle, the hot swap functionality does not work as well when IntelliJ is used for compilation.

One issue is that sometimes, when gradle builds a project, the hot swap agent ends up deploying **all the classes in the project**. This is a problem for our repository because we have a project with 4000 classes, so hot swap is really slow.

### What is it caused by
I debugged the hot swap/gradle code. To me it looks like the root cause is that the `HotSwapUIImpl` class only looks at gradle output directories. When gradle builds a project, the project output paths is written to a temporary file (via a gradle init script). So, when 1 or 100 classes are changed, the input to `HotSwapUIImpl` remains the same - the `classes` folder.

The `HotSwapUIImpl` class is unable to distinguish what classes _actually_ changed so it ends up deploying _all_ the classes

### Solution proposed in this PR
Instead of gradle init script outputting the list of output folders, I changed it to output the list of actual modified classes. I modified the gradle init script (which is now a separate file) to

Inside `IJGeneratedOutputs.groovy` (the new gradle init script)
1. Hash each output class file in `beforeExecute` hook
2. Hash each output class file in the `afterExecute`
3. Compare hashes, and if they are different, write the class file (and output folder) to the temporary file

Inside `GradleProjectTaskRunner` 
1. The temporary file is read same ways as before
2. Instead of using `ProjectTaskContext#addDirtyOutputPathsProvider` method, I now use `ProjectTaskContext#fileGenerated`  - which I believe is same as what IntelliJ uses when building with IntelliJ

The format in the temporary file is modified since `fileGenerated` method requires a `folder` and `relative class path`. So each line in the temporary file has 2 parts with a separator.

### DIFF
--- a/plugins/gradle/java/src/execution/build/GradleProjectTaskRunner.java
+++ b/plugins/gradle/java/src/execution/build/GradleProjectTaskRunner.java
@@ -24,6 +24,7 @@ import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.roots.OrderEnumerator;
 import com.intellij.openapi.util.UserDataHolderBase;
 import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.util.io.StreamUtil;
 import com.intellij.packaging.artifacts.Artifact;
 import com.intellij.task.*;
 import com.intellij.task.impl.JpsProjectTaskRunner;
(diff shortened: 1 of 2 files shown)


gaiazov:
Bump, is there anything I can do to expedite this?

vladsoroka:
Leonid, sorry for the long answer and many thanks for the contribution!
As soon as I add the integration test for the change I will merge it.

LifeIsStrange:
@vladsoroka any update?

gaiazov:
@LifeIsStrange I've been meaning to pick this up again in my spare time. Hopefully I'll have an update (with tests!) in a week or two

gaiazov:
Closing this in favor of newer implementation at https://github.com/JetBrains/intellij-community/pull/1551

PR #1390, opened 2020-06-12
Match schemastore file masks against project-relative paths

pete-woods:
- This change fixes schemastore masks (such as `.circleci/config.yml`, `.github/workflows/**.yaml`) that reference project-relative paths, in addition to matching only on the file name.
- I have verified that this still works for glob masks that include a `*`.
- Using the Java NIO glob implementation should also mean that double `**` globs work.
- See [here](https://www.schemastore.org/api/json/catalog.json) for the full list of masks.
- Tested on macOS only.


segrey:
@pete-woods Thanks for submitting the PR! I've reworked it and pushed as 325f9056271235733fa03dd1b5d024d819ff5322.
Related issue: https://youtrack.jetbrains.com/issue/WEB-46080.

PR #1391, opened 2020-06-15
[intellij sh] Use extension point instead of application service

jansorg:
Currently `ShSupport` is an application service. Plugins, which would like to customize it, have to override it. This override prevents it from being a dynamic plugin.

This PR changes the service into an extension point. Interface `ShSupport` and method `ShSupport.getInstance()` is unchanged. It returns the first available shSupport extension.

### DIFF
--- a/plugins/sh/resources/META-INF/plugin.xml
+++ b/plugins/sh/resources/META-INF/plugin.xml
@@ -16,8 +16,13 @@ Adds support for working with shell script files
 <li>Dedicated run/debug configuration for shell scripts</li>
 <li>Integration with external tools (<a href="https://github.com/koalaman/shellcheck">ShellCheck</a>, <a href="https://github.com/mvdan/sh">Shfmt</a>, <a href="https://explainshell.com/">Explainshell</a>)
 </li></ul>]]></description>
+
+  <extensionPoints>
+    <extensionPoint qualifiedName="com.intellij.sh.shSupport" interface="com.intellij.sh.ShSupport" dynamic="true"/>
+  </extensionPoints>
+
   <extensions defaultExtensionNs="com.intellij">
-    <applicationService serviceInterface="com.intellij.sh.ShSupport" serviceImplementation="com.intellij.sh.ShSupport$Impl"/>
+    <sh.shSupport id="shSupport" implementation="com.intellij.sh.ShSupport$Impl"/>
     <fileType language="Shell Script" extensions="sh;bash;zsh" hashBangs="sh" fieldName="INSTANCE" name="Shell Script" implementationClass="com.intellij.sh.ShFileType"/>
     <lang.syntaxHighlighter language="Shell Script" implementationClass="com.intellij.sh.highlighter.ShSyntaxHighlighter"/>
     <lang.parserDefinition language="Shell Script" implementationClass="com.intellij.sh.parser.ShParserDefinition"/>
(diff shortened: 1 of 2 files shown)


plugins/sh/resources/META-INF/plugin.xml
```diff
@@ -16,8 +16,16 @@ Adds support for working with shell script files
...
+  <extensionPoints>
+    <extensionPoint name="shSupport" interface="com.intellij.sh.ShSupport" dynamic="true"/>
+  </extensionPoints>
+
+  <extensions defaultExtensionNs="com.jetbrains.sh">
+    <shSupport id="shSupport" implementation="com.intellij.sh.ShSupport$Impl"/>
+  </extensions>
+  
   <extensions defaultExtensionNs="com.intellij">
-    <applicationService serviceInterface="com.intellij.sh.ShSupport" serviceImplementation="com.intellij.sh.ShSupport$Impl"/>
```
mmazurkevich, line 20: <sh.shSupport id="shSupport" implementation="com.intellij.sh.ShSupport$Impl" />. This can be here

plugins/sh/resources/META-INF/plugin.xml
```diff
@@ -16,8 +16,16 @@ Adds support for working with shell script files
 <li>Dedicated run/debug configuration for shell scripts</li>
 <li>Integration with external tools (<a href="https://github.com/koalaman/shellcheck">ShellCheck</a>, <a href="https://github.com/mvdan/sh">Shfmt</a>, <a href="https://explainshell.com/">Explainshell</a>)
 </li></ul>]]></description>
+
+  <extensionPoints>
```
mmazurkevich, line 20: Let's use `qualifiedName="com.intellij.sh.shSupport"`

plugins/sh/resources/META-INF/plugin.xml
```diff
@@ -16,8 +16,16 @@ Adds support for working with shell script files
...
+    <extensionPoint name="shSupport" interface="com.intellij.sh.ShSupport" dynamic="true"/>
+  </extensionPoints>
+
+  <extensions defaultExtensionNs="com.jetbrains.sh">
```
mmazurkevich, line 24: If we use `qualifiedName` this block can be removed

plugins/sh/src/com/intellij/sh/ShSupport.java
```diff
@@ -4,7 +4,9 @@
 import com.intellij.execution.Executor;
 import com.intellij.execution.configurations.RunProfileState;
 import com.intellij.execution.runners.ExecutionEnvironment;
+import com.intellij.openapi.application.ApplicationManager;
```
mmazurkevich, line 7: Unused import

plugins/sh/src/com/intellij/sh/ShSupport.java
```diff
@@ -19,7 +21,12 @@
 import org.jetbrains.annotations.Nullable;
 
 public interface ShSupport {
-  static ShSupport getInstance() { return ServiceManager.getService(ShSupport.class); }
+  ExtensionPointName<ShSupport> EP_NAME = ExtensionPointName.create("com.jetbrains.sh.shSupport");
```
mmazurkevich, line 24: Let's make it `public static final`

plugins/sh/resources/META-INF/plugin.xml
jansorg: Thanks! I didn't know about `qualifiedName`.

plugins/sh/resources/META-INF/plugin.xml
jansorg: done

plugins/sh/src/com/intellij/sh/ShSupport.java
jansorg: Isn't this already `public static final` because it's in an interface?

PR #1394, opened 2020-06-19
Add support for Python 3.10 version string

tonybaloney:
Current EAP incorrectly reports Python 3.10 as Python 3.1 (because of the LanguageLevel implementation)

Add support for Python 3.10 version string parsing and update the toString() method to work with version numbers beyond 3.9

### DIFF
--- a/python/python-psi-api/src/com/jetbrains/python/psi/LanguageLevel.java
+++ b/python/python-psi-api/src/com/jetbrains/python/psi/LanguageLevel.java
@@ -79,7 +79,8 @@ public enum LanguageLevel {
   PYTHON36(36, true, false, true, true),
   PYTHON37(37, true, false, true, true),
   PYTHON38(38, true, false, true, true),
-  PYTHON39(39, true, false, true, true);
+  PYTHON39(39, true, false, true, true),
+  PYTHON310(310, true, false, true, true);
 
   /**
    * This value is mostly bound to the compatibility of our debugger and helpers.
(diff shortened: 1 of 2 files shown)


ybedrov:
Hello,

tonybaloney:
Just signed it @ybedrov

sproshev:
@tonybaloney Hello! Since 2020.2 EAP 2 Python 3.10 is recognized as Python 3.9 https://youtrack.jetbrains.com/issue/PY-42855.

There are a little bit more cases that should be addressed before PYTHON310 introduction:

https://github.com/JetBrains/intellij-community/blob/master/python/python-psi-impl/src/com/jetbrains/python/codeInsight/typing/PyTypeShed.kt#L128 and below

https://github.com/JetBrains/intellij-community/blob/master/python/python-psi-api/src/com/jetbrains/python/psi/FutureFeature.java#L79 and below

https://github.com/JetBrains/intellij-community/blob/master/python/src/com/jetbrains/python/sdk/add/PyAddNewCondaEnvPanel.kt#L81

https://github.com/JetBrains/intellij-community/blob/master/python/helpers/tools/versions.xml

Also I was thinking about updating numbers for Python 3 language levels (30 -> 300, 31 -> 301, ..., 39 -> 309) so that 310 will be perfectly fit in the current row. I'll discuss that with a team.

I'd suggest to postpone this PR until Python 3.10a0 is released, by that moment it will be clear if we're going to change language levels numbers and then I'll be able to update `versions.xml` on merging your PR, what do you think?

PR #1395, opened 2020-06-20
IDEA-221360, IDEA-211290, IDEA-217229

pyltsin:
IDEA-221360 - add support custom annotation
IDEA-211290 - add test for this case - work
IDEA-217229 - add support org.assertj.core.util.CheckReturnValue

Hi.
I would like to help with bugs:
- IDEA-221360
- IDEA-211290
- IDEA-217229
I know that my solution may not be very accurate.
If you can, tell me how to improve it, I will be happy to fix it.

This fix is important because it covers case from AssertJ documentation (https://assertj.github.io/doc/#assertj-core):

// DON'T DO THIS ! It does not assert anything and passes
assertThat(1 == 2);


pyltsin:
Already fix with other way

PR #1399, opened 2020-06-27
Disable Mnemonics check for context menu on macOS

mhorioka:
In the issue reported at IDEA-242047, context menu on macOS does not show with error because of failing on finding location/index of Mnemonic character, which is already stripped by the change introduced in https://github.com/JetBrains/intellij-community/commit/d14e2d658b6a295cd53e2d72ee49204a7ebe491f, while Mnemonic handling should not be done on macOS.
To avoid/fix this error, necessity of Mnemonic handling should be determined by calling "UISettings.getInstance().getDisableMnemonics()" like done in other places.


kradima:
Looks good for me, thank you! @develar Could you please take a look around?

PR #1400, opened 2020-06-28
Change installer to not show an error when no config file is provided and default to admin mode

mimi89999:
Resolves the issue with the installer always requiring a config file when installing in silent mode (https://github.com/microsoft/winget-pkgs/pull/179#issuecomment-630247055)

### DIFF
--- a/build/conf/nsis/idea.nsi
+++ b/build/conf/nsis/idea.nsi
@@ -615,7 +615,7 @@ Function silentConfigReader
   ${LogText} "  config file: $R1"
 
   ${ConfigRead} "$R1" "mode=" $R0
-  IfErrors no_silent_config
+  IfErrors bad_silent_config
   ${LogText} "  mode: $R0"
   StrCpy $silentMode "user"
   IfErrors launcher_32
@@ -681,6 +681,11 @@ update_settings:
   !insertmacro INSTALLOPTIONS_WRITE "Desktop.ini" "Settings" "NumFields" "$R0"
   goto done
 no_silent_config:
+  ${LogText} "  config file was not provided"
+  ${LogText} "  defaulting to admin mode"
+  StrCpy $silentMode "admin"
+  goto done
+bad_silent_config:
   Call IncorrectSilentInstallParameters
 done:
 FunctionEnd


mimi89999:
@trespasserw Could you please review this?

trespasserw:
Sorry for the late reply.

To me, the change looks rather incomplete, designed to solve one particular need. Is it possible to get rid of the additional config file completely, and specify the mode via a command-line option?

mimi89999:
Thanks for your reply.

> Is it possible to get rid of the additional config file completely, and specify the mode via a command-line option?

It is possible to specify the mode via CLI. Bleachbit is doing that (https://github.com/bleachbit/bleachbit/blob/f13e9a86ddc92c164b80d7e752ee8e9887253886/windows/NsisInclude/NsisMultiUser.nsh).
I'm not sure what you meant by getting rid of the config file. If you meant making it optional, it's what I want to do. I think it's a bad idea to remove support for it completely, because you can set other options in it like shortcuts.

> To me, the change looks rather incomplete, designed to solve one particular need.

Indeed, it solves one particular need. Some tools and package installers like `just-install` or `winget` can't easily create a config file. They are installing packages system wide and usually just running NSIS installers with `/S`. (https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerRepositoryCore/Manifest/ManifestInstaller.cpp#L127-L133). I don't think that there are any user-wide package installers that could make use of it.

I also saw that the community is discussing going MSIX only (https://github.com/microsoft/winget-cli/issues/407). Maybe it would be worth investigating if JetBrains IDEs can be packaged that way. MSIX apps are running semi containerized and have some limitations (https://docs.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-behind-the-scenes), (https://docs.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-prepare).

trespasserw:
Thanks for the link into winget code, I now better understand how these things are supposed to work.

Okay, defaulting to admin mode probably makes sense. But what would happen to options that are jumped over - e.g. launchers?

mimi89999:
They stay default. Same as if they weren't set in the silent config file.

trespasserw:
I mean, these options _do_ have some defaults in the silent config. I afraid the script was never tested for absence of these defaults and may fail. Or it just won't create any launcher shortcuts, which doesn't seem correct.

mimi89999:
It won't create a desktop shortcut. That's the only option set to true in https://download-cf.jetbrains.com/idea/silent.config that I also see here, but I think that it should be OK. I think that probably most installers don't create desktop shortcuts if not asked to.

trespasserw:
Sorry, I was trying to sat that previously these options were read from a .config file, converted into desktop.ini records, and then executed by other parts of the script. It is these other parts that I afraid may start failing. Okay, I'll test the change.

mimi89999:
Everything should work fine. Please test. I didn't compile IntelliJ, but extracted the files from the installer and made an installer with them that worked fine.

trespasserw:
Well. The modified installer does indeed create the shortcut, and leaves other traces in the registry. In addition, the `/D=` parameter is ignored, though this is probably not related to the change.

mimi89999:
Is this PR OK to merge then?

trespasserw:
Probably yes. The main problem is that after merging I would have to request a documentation update, and for that I will need to find out what the installer does by default. And at the moment I don't have time for that :(

mimi89999:
Hello,

The _winget_ issue was reported in may when _winget_ started to become popular. There were other issues reported before for other package installers. I opened this PR two months ago. I received the last comment from @trespasserw a month ago that docs need to be changed. There was no progress since then. How long does it take to update docs? Could I help with that?

trespasserw:
@mimi89999 any metric of it's popularity? As for the help - if you figure out what the [modified] installer changes in the system, except unpacking the files, that would be very helpful.

mimi89999:
Hello,
The installer should do exactly the same as the regular installer with this config file: [silent.config](https://github.com/JetBrains/intellij-community/files/5179988/silent.config.txt) or as the regular installer with all unchecked options in the config screen:
- Don't create desktop shortcuts
- Don't add the option in context menu (Don't update context menu)
- Don't create file associations
- Don't add launchers to PATH (Don't update PATH variable)
- Don't download and install 32 bit JBR

trespasserw:
Sorry for the late reply.

I agree that your list correctly represents what the installer _should do_ in the absence of a config file. However, it's not what it _actually does_ - e.g. I have found out that the patched installer does indeed create a shortcut. I see two options here: either further modify the installer to make it work as desired, or document the new behavior.

mimi89999:
I configured a fresh env: cloned the repo at cbc460a16c7d4941d6a15f136b210b450d483030, applied this patch and built. This is the installer that I got: https://lebihan.pl/files/ideaIC-203.SNAPSHOT.exe. It does exactly what I said in my previous comment. Where was the shortcut created?

BTW, the NSIS version IntelliJ is using is from 2017. The version of SCons is from 2016. Did you consider updating them?

mimi89999:
@trespasserw Any news on this?

mimi89999:
@trespasserw Any news on this?

mimi89999:
@trespasserw Did you have time to look at this issue and my installer?

trespasserw:
Not yet, sorry. Going to do some work in the area this week and will check it again.

build/conf/nsis/idea.nsi
```diff
@@ -617,7 +617,7 @@ Function silentConfigReader
   ${LogText} "  config file: $R1"
 
   ${ConfigRead} "$R1" "mode=" $R0
-  IfErrors no_silent_config
+  IfErrors bad_silent_config
   ${LogText} "  mode: $R0"
   StrCpy $silentMode "user"
   IfErrors launcher_32
```
trespasserw, line 623: (not related to the change, just a question) The above two lines are pointless, aren't they? The NSIS script syntax looks as if it was created by aliens for predators, so I'll really appreciate a second opinion.

trespasserw (changes requested):

build/conf/nsis/idea.nsi
mimi89999: They seem useless, but they might be checking if `$silentMode` is not defined. I would need to test that.

build/conf/nsis/idea.nsi
```diff
@@ -683,6 +683,9 @@ update_settings:
   !insertmacro INSTALLOPTIONS_WRITE "Desktop.ini" "Settings" "NumFields" "$R0"
   goto done
 no_silent_config:
+  StrCpy $silentMode "admin"
```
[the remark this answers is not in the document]
mimi89999: Done

trespasserw (approved):

trespasserw:
Thanks! The change is now in the main branch (should be mirrored to GitHub eventually); I'll backport it into the current release branch (203) after internal review.

build/conf/nsis/idea.nsi
trespasserw: I see, thanks. I'll leave it as is for now.

PR #1402, opened 2020-07-02
Terminal: Handle spaces in path in .zshrc

pete-woods:
- Fixes errors opening a terminal when you have:
  - A Python venv configured.
  - A project directory path with a space in.
  - A default shell of `zsh`.
- This matches the implementation for BASH.


pete-woods:
I think this will fix zsh in any IntelliJ plugin that has shell integration, when you have a space in your project’s path.

segrey:
@pete-woods Thanks for the PR! This might also be fixed by https://github.com/JetBrains/intellij-community/pull/1408, so I'd better start with that PR first. Would be great if you could file an issue for that in the https://youtrack.jetbrains.com/issues/IDEA?q=%23%7BAssigned%20to%20me%7D%20%23%7BTools.%20Terminal%7D%20

pete-woods:
That looks like it fixes my issue

PR #1403, opened 2020-07-02
Fix SourceFolderManagerImpl.rescanAndUpdateSourceFolders

yuriy-solodkyy:
`rescanAndUpdateSourceFolders` is supposed to collect folders whose state
has change and then update then in a batch by calling
`updateSourceFolders` which will in turn send on `PROJECT_ROOTS`
notification.  However, the call seems to have been accidentally moved
inside the loop.


nskvortsov:
Merged manually, Thank you!

PR #1404, opened 2020-07-02
Fix performance issues in ProgressSuspender.freezeIfNeeded

yuriy-solodkyy:
`freezeIfNeeded` calls `isCurrentThreadHoldingKnownLocks` which is
an expensive call since it relies on
`ManagementFactory.getThreadMXBean.getThreadInfo`. The performance
problems can be reproduced by creating a new project while the IDE
indexes a lot of files. It manifests itself as very slowly progressing
"Scanning files to index..." process (hours instead od minutes).

`freezeIfNeeded` used to test `myProgresses.contains(current)` before
calling `isCurrentThreadHoldingKnownLocks` and seems to be
unintentionally updated when making `ProgressIndicator current`
parameter nullable. This change reverts the order of tests to the
original one.


donnerpeter:
Merged manually to master and 202 branches, thanks!

PR #1406, opened 2020-07-06
IDEA-245329: Handle CONSTANT_Module and CONSTANT_Package

leonardosnt:
See https://youtrack.jetbrains.com/issue/IDEA-245329

This fixes the exception by handling the CONSTANT_Module and CONSTANT_Package constants.

The decompiler will only output this:
```java
class module-info {
}
```
because it's not currently handling Modules, but I can work on this.

Edit: I managed to implement decompilation for module-info files, I just have to add tests.
Example: output of **java.base** module-info.class: https://gist.github.com/leonardosnt/8cde0dcc18ac12fe0adb463d2cbe7d2b


trespasserw:
Thanks for the patch.

Producing empty `class module-info { }` doesn't make much sense (though better than an exception of course), and the `java.base` example looks pretty good. If you need any help with tests, just let me know.

leonardosnt:
I've added a better test case: https://github.com/JetBrains/intellij-community/pull/1406/commits/d5c48afe81b0cae430eeaf645ab0b7093cd19966#diff-1ed35dbfebcc54a728a8527e0e2b3451

Everything seems to be working fine.

19h:
Is there anything still missing here besides the now-conflicting differences to master?

trespasserw:
2 or 3 hours a day are famously missing :( Will try to merge it today.

trespasserw:
Merged, with one additional change: replaced `ClassWriter#internalClassNameToJava` calls with `ExprProcessor#buildJavaClassName` (the latter better handles dollar signs in a name).

Thanks!

PR #1407, opened 2020-07-07
Make TempDirectory open

dahlstrom-g:
… after commit d1fddba made it implicitly final.

### DIFF
--- a/platform/testFramework/src/com/intellij/testFramework/rules/TempDirectory.kt
+++ b/platform/testFramework/src/com/intellij/testFramework/rules/TempDirectory.kt
@@ -21,7 +21,7 @@ import java.util.concurrent.atomic.AtomicInteger
  * An improved variant of [org.junit.rules.TemporaryFolder] with lazy init, no symlinks in a temporary directory path, better directory name,
  * and more convenient [newFile], [newDirectory] methods.
  */
-class TempDirectory : ExternalResource() {
+open class TempDirectory : ExternalResource() {
   private var name: String? = null
   private val nextDirNameSuffix = AtomicInteger()
   private var myRoot: File? = null


dahlstrom-g:
@chashnikov

chashnikov:
Why do you need to extend `TempDirectory`? If you want to change its behavior and your change is generic enough, it would be better to include to the class itself. If you want to customize for a special case it why not use aggregation instead of inheritance?

dahlstrom-g:
> Why do you need to extend `TempDirectory`? […]

I don't need to extend it; I need to update the IntelliJ platform in Android Studio where at least one class, [FakeEmulatorRule](https://android.googlesource.com/platform/tools/adt/idea/+/64f1288/emulator/testSrc/com/android/tools/idea/emulator/FakeEmulatorRule.kt#26), already extends it.

chashnikov:
I think `FakeEmulatorRule` shouldn't extend `TempDirectory`. We cannot say that FakeEmulatorRule is a kind of TempDirectory, and it's strange to have methods like newDirectory in a class named FakeEmulatorRule, so inheritance shouldn't be used here. `FakeEmulatorRule` just uses `TempDirectory` in its implementation, so it would be better to replace inheritance by aggregation, i.e. create a property of type `TempDirectory` in `FakeEmulatorRule` and delegate to it in `before`, `after` and `apply` methods.

dahlstrom-g:
I have no opinion whether `FakeEmulatorRule` is a kind of `TempDirectory`. Maybe it's just misnamed and should instead be called `FakeEmulatorTempDirectory`.

Regardless, is it important to make `TempDirectory` final when it wasn't before? Why? This is causing multiple people including you to spend significant time coordinating and changing existing code. To what benefit? Do you really want to care about the design of `FakeEmulatorRule`?

dahlstrom-g:
> `FakeEmulatorRule` just uses `TempDirectory` in its implementation, so it would be better to replace inheritance by aggregation, i.e. create a property of type `TempDirectory` in `FakeEmulatorRule` and delegate to it in `before`, `after` and `apply` methods.

Do you want to give that a try and send a patch, since otherwise commit d1fddba causes `FakeEmulatorRule` not to compile?

chashnikov:
Dana, I didn't make it final deliberately, I just converted it to Kotlin where classes are final by default. And I wouldn't do it if I know that it's extended in external code; you're right that problems caused by such changes don't outweigh the benefits.

However in general it's better to make classes final unless they are specifically designed for inheritance. Reasons for that are explained e.g. in item 18 of Bloch's Effective Java book, don't think I may describe them better. That's why classes are final by default in Kotlin.

I agree that it's better to avoid breaking compatibility. However since compatibility is already broken in this place, we have two ways to fix it: either just open `TempDirectory` for inheritance or fix classes which extend it to use composition instead. I prefer the second way; this way we make it less probable that further changes in `TempDirectory` will break compilation or behavior of `FakeEmulatorRule`.

> Do you want to give that a try and send a patch

I will give it a try.

develar:
> Is it important to make TempDirectory final when it wasn't before? Why? 

Yes. It is important. To reduce cost of supporting platform and preserving backward compatibility. `final` forces you to write a better code and force you to think is your solution good or maybe there is another one. It's like a pandora box — if `TempDirectory` is open class, then on each change of this class you have to think about inheritors (is it ok to make field private, or rename it).

That's why I also prefer to restrict unless really required.

andrei-kuznetsov:
@dahlstrom-g, could you please check contents of mirror-goog-studio-master-dev manifest (https://android.googlesource.com/platform/manifest/+/457f7ce7495aa5c0326d6f805558f65aa8eb5815/default.xml)? It points to studio-master-dev revisions. Effectively this means that we don't know how to get AOSP master to check if our fix works there.

chashnikov:
@dahlstrom-g I've prepared a [patch](https://github.com/JetBrains/intellij-community/files/4890368/FakeEmulatorRule__don.t_extend_TempDirectory._use_delegation_instead__TempDirectory_wasn.t.txt) (extension was changed from patch to txt to upload it here). Note that I didn't check if it actually works because I didn't managed to setup recent AS sources.

dahlstrom-g:
> Dana, I didn't make it final deliberately, I just converted it to Kotlin where classes are final by default. And I wouldn't do it if I know that it's extended in external code; you're right that problems caused by such changes don't outweigh the benefits.

Thanks! I understand and I'm glad we agree on the cost-benefit analysis (with hindsight). Incidentally converting code from Java to Kotlin is another thing I wish folks would do a lot less of, but that topic feels like too big a tangent to discuss here. ;-)

> However in general it's better to make classes final unless they are specifically designed for inheritance. Reasons for that are explained e.g. in item 18 of Bloch's Effective Java book, don't think I may describe them better. That's why classes are final by default in Kotlin.

Yes, I'm familiar with _Effective Java_; it's been venerated in the Java community at Google. I think what you're referring to, in the second edition anyway, is “Item 17: Design and document for inheritance or else prohibit it”. I haven't read it lately but I recall agreeing with the arguments (and appreciating the clear explanation). A lot of _Effective Java_ though reviews design and implementation choices in Java that were painful or impossible to fix later, so that authors of new code can avoid those mistakes. 

> Yes. It is important. To reduce cost of supporting platform and preserving backward compatibility. `final` forces you to write a better code and force you to think is your solution good or maybe there is another one. It's like a pandora box — if `TempDirectory` is open class, then on each change of this class you have to think about inheritors (is it ok to make field private, or rename it).

This all makes sense for new code, but for existing non-final classes, making them final does the opposite of “preserving backward compatibility”. As for when “you have to think about inheritors” it's not clear to me—shouldn't you consider them before you make a non-final class final?

> @dahlstrom-g I've prepared a [patch](https://github.com/JetBrains/intellij-community/files/4890368/FakeEmulatorRule__don.t_extend_TempDirectory._use_delegation_instead__TempDirectory_wasn.t.txt) (extension was changed from patch to txt to upload it here). Note that I didn't check if it actually works because I didn't managed to setup recent AS sources.

It almost works; `FakeEmulatorRule` needs a `newFile` function but I can add that. Thanks again!

chashnikov:
@dahlstrom-g I glad that we agree!

> This all makes sense for new code, but for existing non-final classes, making them final does the opposite of “preserving backward compatibility”

This also makes sense for existing non-final classes which aren't designed for inheritance and _aren't inherited in external code_, for the same reasons: to prevent external code from inheriting such classes in future and avoid problem which may be caused by that. If such a class already has a subclass in external code, we should not make it final to avoid breaking compatibility; however it makes sense to annotate such class with [`@ApiStatus.NonExtendable`](https://github.com/JetBrains/java-annotations/blob/a0605ed02fff37fd9566a39bc0ede1af85d5e9e1/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L106) to let IDE and the plugin verifier report a warning for a plugin where it's extended.

dahlstrom-g:
Here's the [commit](https://android.googlesource.com/platform/tools/adt/idea/+/8322701) I made starting from the [patch](https://github.com/JetBrains/intellij-community/files/4890368/FakeEmulatorRule__don.t_extend_TempDirectory._use_delegation_instead__TempDirectory_wasn.t.txt) linked in the [comment](https://github.com/JetBrains/intellij-community/pull/1407#issuecomment-655465253) above. Thanks once more @chashnikov!

PR #1409, opened 2020-07-07
Remove final keyword from InvalidDataException

dahlstrom-g:
This was added incidentally by commit d504262
breaking compilation of existing subclasses.

### DIFF
--- a/platform/util/src/com/intellij/openapi/util/InvalidDataException.java
+++ b/platform/util/src/com/intellij/openapi/util/InvalidDataException.java
@@ -1,7 +1,7 @@
 // Copyright 2000-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
 package com.intellij.openapi.util;
 
-public final class InvalidDataException extends RuntimeException {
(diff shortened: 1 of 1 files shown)


dahlstrom-g:
@develar? @SergeyZh?

dahlstrom-g:
@develar, thanks for contributing to the discussion on related #1407. Can we go ahead with this one?

develar:
Is it possible to not extend `InvalidDataException`? Where I can find usages to understand why extending is required?

dahlstrom-g:
The uses are all in Firebase code which isn't public unfortunately.

I don't understand why the `final` modifier was added in commit d504262 in the first place. Is it related to that commit somehow? In any case, with regard to this [general rationale](https://github.com/JetBrains/intellij-community/pull/1407#issuecomment-655394757), `InvalidDataException` doesn't look like much of a maintenance burden given its current state and revision history.

klikh:
Fixed in 96795c2b

klikh:
I used the original hash here, but I forgot that it has been changed when the code was mirrored to GitHub. Sorry for that, glad that you've found the commit.

PR #1411, opened 2020-07-15
Revise code to compile with JDK 11

dahlstrom-g:


platform/platform-impl/src/com/intellij/openapi/wm/impl/FrameInfoHelper.kt
```diff
@@ -77,7 +78,7 @@ private fun updateFrameInfo(frameHelper: ProjectFrameHelper, lastNormalFrameBoun
   val frame = frameHelper.frame
   var extendedState = frame.extendedState
   if (SystemInfo.isMacOSLion) {
-    val peer = AWTAccessor.getComponentAccessor().getPeer(frame)
+    val peer = AWTAccessor.getComponentAccessor().getPeer(frame) as ComponentPeer?
```
jonnyzzz, line 81: Why is this cast necessary if we do that the `is` check below?

dahlstrom-g: The commit message explains:
> In JDK 8 `AWTAccessor.getPeer` returns raw `ComponentPeer`
> but in JDK 11 it returns `<T extends ComponentPeer>`.

Without the cast Kotlin compilation fails with `Not enough information to infer type variable T`.

dahlstrom-g:
@jonnyzzz, @SergeyMalenkov: For context, in the Android Studio build we have a fairly urgent need to compile the IntelliJ platform with JDK 11. We try to minimize divergence from JetBrains/intellij-community so we'd prefer to see these commits incorporated in master and cherry-pick them from there, hopefully in the next few days.

For what it's worth I gather @chashnikov may have done some work in this area.

develar:
Thanks, fix will be pushed to master soon.

chashnikov:
@dahlstrom-g and could you please give me a link to your commit in android sources which adapts code to Java 11? I suppose we may migrate to Java 11 before we do the next merge from AS, so it would be better to avoid conflicts.

rsauciuc:
@chashnikov We're using the legacy Java toolchain from Bazel, with the JavaBuilder/error-prone javac9 compiler.  We switched the bootclasspath to 11, and we fixed compilation in this configuration. There will be more issues to fix once we make the jump to javac11, we've only taken this first step.

dahlstrom-g:
Commit af4deb0 makes a mess of this pull request: it combines commits c635239, 3cbfc52, 15f8333, and ad797f1, each of which has an accurate description we took pains to write, and unnecessarily modifies many more lines for trivial refactoring and reformatting. The result is an incoherent commit with a mostly inaccurate description arbitrarily copied from commit c635239.

Sadly commits in intellij-community often look something like this, which makes a lot of work harder, and this case is no exception. It was a pain to cherry-pick to Android Studio branches and I expect it to cause conflicts again in future merges.

More attention to SCM discipline would be much appreciated. Here's a blog post that explains a lot of what I and others would like to see and why:

https://who-t.blogspot.com/2009/12/on-commit-messages.html

dahlstrom-g:
@chashnikov: I've added commit 3f58b37 here; can you take a look?

chashnikov:
@dahlstrom-g I've [pushed](https://github.com/JetBrains/intellij-community/commit/f61d0f6a1cdf2d6fe14e35019ffc3e2c69297e33) your changes and also added suppressions for "redundant cast" warning; otherwise other developers may apply quickfix to fix such "redundant" casts. The commit will appear in Github soon.

And I'm very sorry that we messed up the previous commits from this pull request. I agree with everything you said about importance of proper commit messages. Unfortunately it takes time to ensure that all team members make it a habit to properly organize commits before pushing, we're working on it.

dahlstrom-g:
@chashnikov: Thanks too for improving the content and message of commit 3f58b37. :-)

PR #1413, opened 2020-07-17
HTTP Proxy Preferences: Store the check url while dialog is open

kullmanp:
The proxy preference page has a button to check the connection with the
settings entered in the page. Every time the button is pressed the user
is prompted for a check url. The field is always reset to 'http://'.
Sometimes proxy settings are not correct on the first attempt or one
can make a small typo in the check url. In these cases one has to
press check repeatedly and it is inconvenient to reenter the url
every time. This change stores the url for the time the preference
page is open.


MarkusTiede:
Are there any plan when / how to integrate this change? Thanks and with best regards!

BasLeijdekkers:
Merged manually https://github.com/JetBrains/intellij-community/commit/87ff47c39f1b48c4d43b29c9850b64d308c88509
Thanks for your contribution!

PR #1414, opened 2020-07-17
To log long refresh sessions details

kadim:


kadim:
Example

`2020-07-17 08:46:48,749 [  76187]   INFO - napi.vfs.newvfs.RefreshSession - Refresh session took longer than 10 seconds. Session info -> work queue size: 11118 , file types: {ARCHIVE=11117, UNKNOWN=1}; result: done, duration: 30869 ms, generated events count: 0`

ybedrov:
Hello!

kadim:
@ybedrov CLA signed

dkashyn-sfdc:
@trespasserw do you need anything else from me? Could you please also tell whether target branch is correct if this targeting 2020.2

@batya239 cc

trespasserw:
Oh, I do. I need from you to become convinced that's is not the change you want in 2020.2 :)

I see quite a few problems in the patch. Before we go into details, could you please tell what are you trying to achieve?

trespasserw:
Does it look like https://youtrack.jetbrains.com/issue/IDEA-237996? Or just one single long refresh?

kadim:
Basically https://github.com/JetBrains/intellij-community/blob/a295d3a8a2408cb60870c1ca6739c466f09f4861/platform/lang-impl/src/com/intellij/util/indexing/UnindexedFilesUpdater.java#L218 vs https://github.com/JetBrains/intellij-community/blob/14ed57d99a8699ace069c8b0938d28a260a2f98d/platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java#L114

kadim:
To add more: there is a good indicator in log for "unindexed files" which is "scanning" but no clue for "refresh session" which takes care of known files and we have strong suspicion that one causes delay we observe.

Also "refresh" is different since happens really often hence logging only of long runs but not every run

trespasserw:
Okay. I'll write down what's I do believe is wrong with the change soon - meanwhile, you can try the experimental concurrent refresh implementation by enabling `vfs.use.nio-based.local.refresh.worker` in _Help | Find Action | Registry_; see also `vfs.use.nio-based.local.refresh.worker.parallelism` nearby.

kadim:
@trespasserw thanks, I'll try, without logging though I don't have a baseline :( So to be able to see difference I need to have timings of "default" refresh we have now. So could you please comment on PR to make it to the codebase.

trespasserw:
First, the real problem. Please don't use `VirtualFile#getFileType` - it may need to read content of the files to detect their type. Use `FileTypeRegistry#getFileTypeByFileName` instead, and check for `VirtualFile#isDirectory` first (most directories are of "unknown" type).

Next, the logging definitely should be enabled for all by default. The threshold is hardcoded, but what should be considered "long" heavily depends on a project, an environment, and a current IDE state. So please make the threshold explicitly configurable via a system property. Oh, and `refreshSessionInfo` calculation should be moved under the condition.

The last thing is rather philosophical. I'd say that the proposed statistic won't help you much. Supposing the project is JVM-based, the first number in the example "{ARCHIVE=11117, UNKNOWN=1}" stands for .jar files, and the last one is for FS root, `/`. Most probably, it is the latter that takes most of the time, but without detailed traces it is impossible to tell why.

kadim:
> Oh, and refreshSessionInfo calculation should be moved under the condition.
Unfortunately the fact that session was "long" determined afterwards when work queue not available anymore... So not sure it is possible to make it conditional.

I can replace file types by extensions probably... Just wanted to see for our case what kind of input and in raw trace output it has 10k+ of jars. So any hint of what kind of input will be helpful I'd say. What would you say?

Proposed statistic will help for one of cases when we have IDE "in limbo" for longer than expected and if numbers confirmed we need to address this. If numbers are lower than we anticipate then:
1. Timings won't hurt to see anyway
2. We'll try to find real cause of "lock" if RS is a red herring

Regarding file types there is a way to make refresh less costly for them by using CRC and I suspect in our case it will help drastically.

Once again: this particular PR purpose is to demystify slowness of IDE with intent to address it based on findings from real users timings.

@nicity and Dmitry Batkovich (don't know his GitHub name) has some context for this

kadim:
`2020-07-20 15:05:31,062 [  67905]   INFO - napi.vfs.newvfs.RefreshSession - Refresh session took longer than 10 seconds. Session info -> work queue size: 11117 , file types: {zip=2, folder=1, jar=11114}; result: done, duration: 29759 ms, generated events count: 0 `

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
```diff
@@ -122,10 +127,15 @@ void scan() {
         ((LocalFileSystemImpl)fs).markSuspiciousFilesDirty(workQueue);
       }
 
-      long t = 0;
+      long t = System.currentTimeMillis();
+      String refreshSessionInfo = String.format("work queue size: %s , file types: %s", workQueue.size(),
+                                                workQueue.stream().collect(
+                                                  Collectors.groupingBy(file ->
+                                                                          file.getExtension() != null ?
+                                                                          file.getExtension() : (file.isDirectory() ? "folder" : "NA"),
```
dkashyn-sfdc, line 135: if not folder and no extension let's not guess since hard to pick good group for such case

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
```diff
@@ -122,10 +127,15 @@ void scan() {
         ((LocalFileSystemImpl)fs).markSuspiciousFilesDirty(workQueue);
       }
 
-      long t = 0;
+      long t = System.currentTimeMillis();
+      String refreshSessionInfo = String.format("work queue size: %s , file types: %s", workQueue.size(),
```
dkashyn-sfdc, line 131: we cannot do this conditional, unfortunately, since we don't know how long session might take and queue won't be there by the time we done. We can accumulate processed entries but it might be worse than compute this unconditionally

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
```diff
@@ -154,9 +164,13 @@ else if (!nvf.isDirty()) {
       }
       while (!myCancelled && myIsRecursive && count < 3 && ContainerUtil.exists(workQueue, f -> ((NewVirtualFile)f).isDirty()));
 
-      if (t != 0) {
-        t = System.currentTimeMillis() - t;
-        LOG.trace((myCancelled ? "cancelled, " : "done, ") + t + " ms, events " + myEvents);
+      t = System.currentTimeMillis() - t;
+      if (LOG.isTraceEnabled()) {
+        LOG.trace((myCancelled ?  "cancelled, " : "done, ") + t + " ms, events " + myEvents);
+      } else if (t > REFRESH_SESSION_DURATION_REPORT_THRESHOLD_SECONDS * 1000L){
```
dkashyn-sfdc, line 170: int is enough here I suppose, can flip to have it long if reqd but readability will suffer

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
trespasserw: Better do `file -> file.isDirectory() ? "directory" : Objects.requireNonNullElse(file.getExtension(), "n/a")`.
The order of conditions matter, because `VirtualFile#getExtension` is a text-based method and may return misleading results if a directory happen to have a dot in a name; and the result isn't cached, so asking once once is more environment friendly.

trespasserw:
Yep, using extensions is a good idea.

Not sure I got the idea of using CRC. For files, only metadata is used in refresh (a length and a modification timestamp).

"Timings won't hurt to see" isn't totally true. Refresh sessions happen pretty often, being >10 seconds is pretty normal, and unsolicited timings would pollute the log. In other words, it would hurt those who have to read many logs.

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
kadim: oh, my bad, it is indeed immutable. Moving under condition then

kadim:
Ok, bumping default to 30 seconds. We want to have this log line only because delay is uncomfortably long and we not sure whether it caused by RefreshSession or not. 10 seconds indeed quite low, 30 is good enough reason to speed up Refresh taking in account it has really subtle clue in UI if you know where to look :)

kadim:
New log message sample
`2020-07-20 18:14:42,773 [  66701]   INFO - .diagnostic.PerformanceWatcher - Refresh session took longer than 10 seconds. Session info -> work queue size: 11117 , file types: {zip=2, null=1, jar=11114}; result: done, generated events count: 0 took 28945ms; general responsiveness: ok; EDT responsiveness: ok`

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
kadim: well... it is not exactly immutable and more items can be added since we started session in theory (depending on in which order launch() and scan() called). But since there is no way to delete it is ok to use queue afterwards.

kadim:
> Not sure I got the idea of using CRC. For files, only metadata is used in refresh (a length and a modification timestamp).

All I know that IDE stays "in limbo" for 2-3 mins, "Refresh" takes substantial time so as I mentioned: more data we have more chances to speed it up.

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
kadim: requireNonNullElse is JDK9+ so to avoid issues with null using toString() instead

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
trespasserw: It is effectively immutable. Once `scan()` is called, the set of roots this session will process is fixed.; any new roots are simply ignored. So yes, it is perfectly okay to use the `workQueue` variable anywhere in the method.

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
trespasserw: I have lost track again which modules are at which level :(

Anyway, `"null"` is okay for files (files w/o an extension aren't very common among refresh roots), but mixing directories into the same counter makes the log look more cryptic than it should.

trespasserw:
> Ok, bumping default to 30 seconds.

Sorry, but this won't help. The logging should be off by default. There is no one universal measure for how long a refresh should take in each case, and we don't have resources to answer all questions that inevitable will arise after users start seeing that record in their logs.

platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/RefreshSessionImpl.java
kadim: there is no good "type" for directory and with 11k of jars there is no way to avoid grouping and this is what makes TRACE kind of useless

kadim:
not enabled by default now

trespasserw:
Thanks! One last question: is it super-important for you to have this in the 2020.2 release, or 2020.2.1 bugfix update is OK?

dkashyn-sfdc:
thanks @trespasserw !
we need this in 2020.x codeline to answer some questions for sure, 2020 GA will keep us busy for some time so 2020.1 is fine I think. As long as we have time to react in 2020.3 before freeze we are good.

Thanks again!

trespasserw:
👍

FYI: I took the liberty to edit the message a bit - to make it better align with the text appended by `PerformanceWatcher`.

PR #1415, opened 2020-07-17
Fix unbound variable errors with nounset mode bash terminal startup

scop:
With `set -o nounset` (or `set -u`) in bashrc, opening terminal gives:
  bash: JEDITERM_USER_RCFILE: unbound variable
  bash: JEDITERM_SOURCE: unbound variable


segrey:
@scop Thanks for the PR! Sorry for the delay. Merged now as ec113a28833e56e7f34f80dde89b264f24e472e1. It will be included in 2021.2.

PR #1417, opened 2020-07-19
Simplify redundant code

thenx-wei:
a null check followed by a method call that will definitely return false when null is passed (e.g. Class.isInstance). Such a check seems excessive as the method call always returns false


ybedrov:
Hello! 

amaembo:
Merged into the internal repository. Will appear on GitHub in a few hours. Thanks.

PR #1418, opened 2020-07-23
fsnotifier: Support mounted volumes

chrullrich:
Looking only at the root directories (C:\ etc.) means that paths on mounted volumes cannot not be monitored. Instead, the user must manually refresh both the project tree and, for example, the Git changelist, to pick up changes.

This changes fsnotifier to monitor each path given to it, rather than reducing them to the root directory which may be on another volume.

I have been using this version for about a year now with no apparent problems.

This will probably break with other kinds of paths than mounted local volumes, such as symlinks to network shares. It will also not cover mounted volumes below the watched directory, i.e. inside the project. In any case, it is better than before.


trespasserw:
Thanks!

Unfortunately, the change fails some tests ([report attached](https://github.com/JetBrains/intellij-community/files/4972663/FileWatcherTest.html.gz)). Even more unfortunately, at the moment I can't investigate the failures and propose possible solutions :(

chrullrich:
This does not yet fix all the failing tests, but it does some of them. Still working on it.

chrullrich:
This passes all tests, with two exceptions:

- testUncRoot, because the share from the test case does not exist here
- testWatchRootRenameRemove, because this test moves a directory from below the watch root to above it and expects the watcher to report both the DELETE and CREATE events. This is not possible without watching anything and everything again.

While all the other tests pass, I suspect there may be some false negatives among them.

This PR also includes my PR #1425 from today because they conflict due to changes in surrounding lines.

trespasserw:
`testUncRoot` should be easy - it expects that the watcher rejects UNC paths (reports them as "UNWATCHEABLE"). The server and share name are fictional.

`testWatchRootRenameRemove` actually passes for me. See, the test is rather an integration one, it verifies how all watcher implementations work with other parts of the IDE, and the events expected are ones from the VFS and not the raw watcher messages. The case verifies two things: a deletion event when a root itself is deleted, and a creation not long after a root comes back.

From the watcher messages, it seems that both requirements are satisfied, though the messages around last pair of assertions are somewhat suspicious: deletion of the `top` directory results in a single `DELETE, C:\...\top\d1\d2\d3\root` message (which is perfectly okay), but the subsequent `root` directory is accompanied by something that looks like a series of delayed events:
> DELETE
C:\...\top\d1\d2\d3\
DELETE
C:\...\top\d1\d2\
DELETE
C:\...\top\d1\
DELETE
C:\...\top\
CREATE
C:\...\top
CREATE
C:\...\top\d1
CREATE
C:\...\top\d1\d2
CREATE
C:\...\top\d1\d2\d3
CREATE
C:\...\top\d1\d2\d3\root

Is this expected?

chrullrich:
Thank you for explaining this.

I'll get the SMB test working; I have to look at the unwatchable response anyway, I suspect it never reports anything right now. It is also good to know that the arrival event for the upwards rename is not required; I was considering dropping all the nice logic I invented and just always watching from the closest volume root (i.e. drive root directory or mount point).

I don't think I have seen testWatchRootRenameRemove produce this DELETE cascade, but it makes sense for a recursive deletion, provided you meant `root` instead of `top` and vice versa. The size of the cascade (i.e. how many individual levels it reports) depends on a race between the watcher and whatever is deleting directories.

My code tries to stay as close as possible to the place it has been told to watch to minimize the number of unrelated events. Because of the way ReadDirectoryChangesW() works, it actually watches the parent directory of that place. When the watched directory is deleted, it goes up the tree level by level until it finds something to watch from again. (This means that the watch root can change while the program is running, and it will not go back to where it started even if that directory comes into existence again.)

If a recursive deletion is slow enough that the watcher can abort and recover while the directories are still being deleted, you get DELETE events for intermediate directories if it can recover after each directory, the cascade will look like yours above. The CREATE cascade happens when the directories are recreated. The watch root has moved up to the parent of `top` during the deletion, so you get all events down to `root`.

I'll keep on it. One more question: The current system requirements for IDEA mention Windows 8 as the oldest supported OS. Can I actually use features introduced in 8, or do you want to remain "unofficially" compatible with even earlier versions?

trespasserw:
I see, thanks. For the record, Linux implementation keeps a list of missing roots and polls them periodically.

Would be nice to keep Windows 7 compatibility for a while.

Hope I'll be able to start reviewing the code in a week or two.

trespasserw:
Sorry for the late reply.

Thanks for your contribution!

PR #1421, opened 2020-07-28
Shelf Panel: Expand/Collapse actions

alexey-anufriev:
Added two useful controls for Shelf panel: Expand All / Collapse All

image


AMPivovarov:
Merged, thanks!

PR #1422, opened 2020-07-30
[git] migration to Java 11: use explicit 'setUI' call instead of 'ui' property

dahlstrom-g:
...to fix "Val cannot be reassigned" error when compiling IntelliJ sources under Java 11 (see KT-31926).


dahlstrom-g:
@chashnikov

chashnikov:
Merged, thank you!

PR #1425, opened 2020-08-06
Fix incorrect output from fsnotifier on Windows.

chrullrich:
- Renaming a file produces a CREATE event for the old name
  and nothing for the new one.

- If the file system returns multiple changes at once,
  the last one is lost.


trespasserw:
Discovered today that I didn't yet merge the fix. For whatever reason I thought I did it long ago :/

PR #1426, opened 2020-08-13
Add system paths info to version app starter

brian-mcnamara:
We want to programmatically grab the config/system paths from IJ during our installer to be able to modify vmoptions and properties. This will allow us to know these locations over guessing the location.


dkashyn-sfdc:
@trespasserw could you please help with this? We need some predictable way to bump Xmx for any IJ install we have and only IJ itself knows all details we need to make this work

dkashyn-sfdc:
It would be great to solve https://youtrack.jetbrains.com/issue/IDEA-244523 to make this less hacky for us though ;)

PR #1429, opened 2020-08-18
Display name for JSON Schemas from schemastore where possible

artemy:
Currently Schema Mapping menu doesn't use "name" field from jsonschema.org catalog, which makes it difficult to find correct schema mapping.

I've updated menu code to use name if possible and disabled sanitizing temporary caption (which shows filename during downloading) so file extension is no more omitted.

Before|After
:-------------------------:|:-------------------------:
before | after


artemy:
@segrey do you need any additional information/clarification or other help?

segrey:
@artemy Merged. Thank you!

PR #1430, opened 2020-08-19
IDEA-248525 Support import of Javac compiler parameters per-module from Gradle

rosti-il:


nskvortsov:
Merged manually, thank you!

PR #1431, opened 2020-08-24
Use HTTPS for critical-startup-errors URL

bdukes:
I noticed this error was using HTTP in the URL, but HTTPS is supported for this domain. Hope it helps!

### DIFF
--- a/platform/bootstrap/src/com/intellij/idea/Main.java
+++ b/platform/bootstrap/src/com/intellij/idea/Main.java
@@ -125,7 +125,7 @@ public final class Main {
         "The IDE failed to install some plugins.\n\n" +
         "Most probably, this happened because of a change in a serialization format.\n" +
         "Please try again, and if the problem persists, please report it\n" +
-        "to http://jb.gg/ide/critical-startup-errors" +
+        "to https://jb.gg/ide/critical-startup-errors" +
         "\n\nThe cause: " + e.getMessage();
       showMessage("Plugin Installation Error", message, false);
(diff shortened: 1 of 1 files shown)


rosti-il:
Interesting that opening the http://jb.gg/ide/critical-startup-errors in a browser makes two redirects: (1) into the same URL but with HTTPS and then (2) into https://intellij-support.jetbrains.com/hc/en-us/articles/360007568559 that is also HTTPS. So opening [http**s**://jb.gg/ide/critical-startup-errors](https://jb.gg/ide/critical-startup-errors) i.e. with HTTPS in the first place makes one redirect less.

On the other hand the jb.gg is hosted at Amazon cloud and using it with the HTTPS (the first redirection) just for redirections to the JetBrains site (the second redirection) seems like an over-engineering. Could it be configured at Amazon cloud to make such redirections straight forward?

VladRassokhin:
>seems like an over-engineering
Contrariwise, it's just two separated simple logic blocks: 1 - convert all http to https; 2 - redirect to actual site. First block is used not only here and it's natural to reuse it.

Instead of creating another logic of redirecting http to external site it's safer and proper to get rid of http links in first place, that's done in this PR. Once merged, there would be only one redirect.

rosti-il:
Then you should change the URL in the https://intellij-support.jetbrains.com/hc/en-us/articles/360007568559 article as well.

ant-druha:
@rosti-il done, thank you.

PR #1432, opened 2020-08-24
Less noise in logs

kadim:
In some our projects we have dozens of *.run.xml files which are not run configuration and this causes noise in logs.


platform/execution-impl/src/com/intellij/execution/impl/RCInArbitraryFileManager.kt
```diff
@@ -126,7 +126,7 @@ internal class RCInArbitraryFileManager(private val project: Project) {
     }
 
     if (element.name != "component" || element.getAttributeValue("name") != "ProjectRunConfigurationManager") {
-      LOG.info("Unexpected root element ${element.name} with name=${element.getAttributeValue("name")} in $filePath")
+      LOG.trace("Unexpected root element ${element.name} with name=${element.getAttributeValue("name")} in $filePath")
```
kadim, line 129: not sure whether this logging is useful enough to have it as "info" in first place....

PR #1433, opened 2020-08-25
Check TODOs only inside committed lines in partial commits (IDEA-189479)

chylex:
This resolves https://youtrack.jetbrains.com/issue/IDEA-189479. It does not affect other 'Before commit' checks which would have been useful in general, however since the issue is marked as 'Shelved' and I run into TODO check complaints in partial commits significantly more often than the other checks, I figured I'd at least provide a simple solution to that.

I added a new interface method to `PartialLocalLineStatusTracker`, as far as I can tell it's not intended to be used by plugins and has no documentation, so hopefully that's okay.

I also was not sure if it's ever possible for the 'Before commit' `Change`s to not belong to any changelists, so in case that happens those changes *will* be processed by the TODO check, although it seems inconsistent with how `handlePartialCommit` in `PartialLocalLineStatusTracker` handles things. Let me know if you'd like me to ignore changelist-less changes instead.

I noticed that in the existing `TodoCheckinHandlerWorkerTest` tests, two tests (`testMultiLineTodoCreation` and `testMultiLineTodoRemoval`) were already failing before my changes. My changes did not cause any additional failures.

If you want me to add unit tests to cover the feature, let me know if the rest of the code is okay and I'll try to get to it soon.

### DIFF
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/checkin/TodoCheckinHandlerWorker.java
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/checkin/TodoCheckinHandlerWorker.java
@@ -5,6 +5,7 @@ import com.intellij.diff.comparison.ComparisonManager;
 import com.intellij.diff.comparison.ComparisonPolicy;
 import com.intellij.diff.comparison.DiffTooBigException;
 import com.intellij.diff.fragments.LineFragment;
+import com.intellij.diff.util.LineRange;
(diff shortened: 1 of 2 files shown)


AMPivovarov:

But I think, that a slightly different approach will be better (`wasTodoItemCommitted` might work incorrectly if `ComparisonManager` and `PartialLocalLineStatusTracker` disagree on which lines in the file are modified).
The fix is in master and should be synced with GitHub soon.

PR #1435, opened 2020-08-28
Add back log message for Gradle Java

srmurguia:
This was removed by a previous commit but those messages are useful when
debugging issues related to the JDK location that Gradle uses.

https://youtrack.jetbrains.com/issue/IDEA-249386


nskvortsov:
merged manually, thank you

PR #1440, opened 2020-09-09
File Colors Settings: Make new scopes available for colors configuration

alexey-anufriev:
**Problem**
Newly created scopes are not visible in File Colors settings as they have an empty package set by default.

**Fix**
Empty package set must be assigned by default to newly created scopes.


alexey-anufriev:
@akozlova, hi, is there any feedback on this change?

akozlova:
What is the purpose of the change? Actually, if applied, I would also tighten the check `com.intellij.ui.tabs.FileColorsConfigurableKt#getScopes` and ensure that the scope is non-empty. Otherwise, it's easy to add an invalid configuration and forget about it

alexey-anufriev:
> What is the purpose of the change?

The test case is following:
1. In "Scopes" settings section create a new scope and save settings
2. Go to "File Colors" settings section and try to configure a color for a new scope: it is not visible as related package set is null but this check https://github.com/JetBrains/intellij-community/blob/master/platform/lang-impl/src/com/intellij/ui/tabs/FileColorsConfigurable.kt#L454 accepts only non-null oces

alexey-anufriev:
> I would also tighten the check `com.intellij.ui.tabs.FileColorsConfigurableKt#getScopes`

Do you mean to replace `return list.filter { it.value != null }` with `return list.filter { it?.value != null }`?

But can the scope be null here?

akozlova:
I want to check not only nullability but also reject scopes without patterns as it makes not much sense to configure anything for such scopes. Or it's ok to show scopes with `null` values as there is no difference between these cases from the user point of view

alexey-anufriev:
I think scopes with empty packages can be a regular case. I usually create a new scope then want to assign a color for that.
So if you suggest to remove null filtering, I can do that as well. I just thought it is cleaner to have the value set.

alexey-anufriev:
@akozlova, with my use case in mind, what do you think, shall we change the implementation? or it is good to go?

akozlova:
I am not sure I understand the case, why would you go to configure colors before you have configured the scope? You would need to return to configure scopes or you won't see colors, correct? So it's two steps. 

I assume, that your actual case is different and may be related to your plugin?

alexey-anufriev:
I usually do the following:
1. go to scopes settings and create a scope
2. go to colors settings and "want but can't" configure a color for this new scope
3. go back to the project and start paining my source tree

alexey-anufriev:
hello @akozlova, any updates here?

akozlova:
I am not convinced that the change is needed, sorry. I can make the change I suggested to make the checks more strict though

alexey-anufriev:
@akozlova, let's try to follow the case I described:

1. I go to scopes settings and create a scope;
2. I go to colors settings and "want but can't" configure a color for this new scope;
3. I go back to the project and start painting my source tree.

This does not work at the moment.

Do you have any suggestions on how this can be fixed?

akozlova:
why do you need this scenario to work? you want to have a color for the undefined scope, I don't understand, sorry. Looks like it's needed for your plugin but you don't explain why...

alexey-anufriev:
> why do you need this scenario to work? you want to have a color for the undefined scope, I don't understand, sorry. Looks like it's needed for your plugin but you don't explain why...

It is not needed for my plugin. It does it from the code. I can create a scope and assign a color right away. While working on the plugin I have noticed this problem so I just wanted to contribute this fix.

The reason why I want this to work is simple, it is a pretty valid scenario: you predefine a scope the way you want it to be and then you will be using it later

Currently, it has a couple of problems:
1. IDEA limits the way scopes can be used. It does not allow assigning a color to a new scope which is weird to me. Why having this limitation is good?
2. IDEA does not show anywhere any information about this limitation, so users may be confused (as I was).

alexey-anufriev:
hi @akozlova, should we merge this one, or should I just close it?

PR #1442, opened 2020-09-16
Fix NPE in UsageViewImpl

csrhodes:
If there is exactly one Usage, getNextToSelect(usage) returns null,
giving a null pointer exception when looking null up in the toDelete
collection.  Treating it as a successful .contains() lookup (i.e. not
updating toSelect) is safe because there will be no nodes left after
the single existing node is deleted.

### DIFF
--- a/platform/usageView/src/com/intellij/usages/impl/UsageViewImpl.java
+++ b/platform/usageView/src/com/intellij/usages/impl/UsageViewImpl.java
@@ -2376,7 +2376,7 @@ public class UsageViewImpl implements UsageViewEx {
     Usage toSelect = null;
     for (Usage usage : toDelete) {
       Usage next = getNextToSelect(usage);
-      if (!toDelete.contains(next)) {
+      if (next != null && !toDelete.contains(next)) {
         toSelect = next;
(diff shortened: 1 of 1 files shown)


ybedrov:
Hello!

csrhodes:
Hi!  I work at google, so I think there should be no CLA needed?

ybedrov:
Sorry, it wasn't clear from your profile. In this case no CLA is needed.

csrhodes:
Thanks.  (I'll see if I can amend my profile to make that obvious in future.)

csrhodes:
Hi!  is there anything I can do to help this along?

Vassiliy-Kudryashov:
Thank you for the fix, it was pushed to master branch. What kind of "null-incompatible" collection do you use? Standard ArrayList can survive after contains(null) but some "optimized" collections may fail.

csrhodes:
Thank you!  The collection used in the case where I noticed it was a `ConcurrentHashMap` -- I was passing the result of `UsageView.getUsages()` in to `UsageView.removeUsagesBulk()`

PR #1443, opened 2020-09-16
Dependency resolution fixes in Gradle composite builds.

raymank26:
The PR is going to fix two issues.

The first one - https://youtrack.jetbrains.com/issue/IDEA-232283 is solved by removing "composite build substitution" string check. I guess it was left after new resolution algorithm was implemented.

The second one - https://youtrack.jetbrains.com/issue/IDEA-196240. Where transitive dependency (project3) is not added to (project1).

I'm not sure about total correctness of the fixes but please take a look :)

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleProjectResolverUtil.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleProjectResolverUtil.java
@@ -604,8 +604,7 @@ public final class GradleProjectResolverUtil {
         final ExternalProjectDependency projectDependency = (ExternalProjectDependency)mergedDependency;
 
         Collection<ProjectDependencyInfo> projectDependencyInfos = new ArrayList<>();
-        String selectionReason = projectDependency.getSelectionReason();
-        if ("composite build substitution".equals(selectionReason) && resolverCtx.getSettings() != null) {
+        if (resolverCtx.getSettings() != null) {
           GradleExecutionWorkspace executionWorkspace = resolverCtx.getSettings().getExecutionWorkspace();
           ModuleData moduleData = executionWorkspace.findModuleDataByArtifacts(projectDependency.getProjectDependencyArtifacts());
           if (moduleData != null) {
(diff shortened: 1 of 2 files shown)


ybedrov:
Hello!

raymank26:
> Hello!
> 

Hello!

Thanks, did it after your request.

plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleProjectResolverUtil.java
```diff
@@ -604,8 +604,7 @@ else if (ownerDataNode.getData() instanceof DependencyData) {
         final ExternalProjectDependency projectDependency = (ExternalProjectDependency)mergedDependency;
 
         Collection<ProjectDependencyInfo> projectDependencyInfos = new ArrayList<>();
-        String selectionReason = projectDependency.getSelectionReason();
-        if ("composite build substitution".equals(selectionReason) && resolverCtx.getSettings() != null) {
+        if (resolverCtx.getSettings() != null) {
```
vladsoroka, line 607: This will break another kind of dependency resolution, because it will ignore `sourceSetMap` and `artifactsMap` mappings.

plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/util/resolve/DependencyResolverImpl.java
```diff
@@ -266,7 +266,7 @@ private FileCollection getCompileClasspath(SourceSet sourceSet) {
 
           ProjectComponentIdentifier projectComponentIdentifier = (ProjectComponentIdentifier)artifact.getId().getComponentIdentifier();
           String projectPath = projectComponentIdentifier.getProjectPath();
-          String key = projectPath + "_" + resolvedDependency.getConfiguration();
+          String key = projectComponentIdentifier.getProjectName() + "_" + resolvedDependency.getConfiguration();
```
vladsoroka, line 269: Proposed key will not be unique for projects with the same name but from different included builds.

vladsoroka:
Hello Anton, thank you for the patch and your time!
I've found resolution issues with the change. There is also issue when projects with the same name but from different included builds will not be handled correctly (however it's actual w/o your change also). 

I've added similar fix and integration test for it in order to catch such regression in future.
You can find the commit at https://github.com/JetBrains/intellij-community/commit/c9670894ae849152ade41e0ef0a5443ee71b3260

raymank26:
> Hello Anton, thank you for the patch and your time!
> I've found resolution issues with the change. There is also issue when projects with the same name but from different included builds will not be handled correctly (however it's actual w/o your change also).
> 
> I've added similar fix and integration test for it in order to catch such regression in future.
> You can find the commit at [c967089](https://github.com/JetBrains/intellij-community/commit/c9670894ae849152ade41e0ef0a5443ee71b3260)

Great news! Thank you for your fix and commenting on the issue back. 👍

PR #1444, opened 2020-09-17
Revert usage of keymap.current.os.only key but set it to true by default

SerVB:
This doesn't change the default behavior, but reverts the feature of showing keymaps of foreign OSes via overriding Registry key.


ybedrov:
Hello!

PR #1445, opened 2020-09-21
Fix IDEA-251181 by upgrading gradle-idea-ext to 0.9

rosti-il:
https://youtrack.jetbrains.com/issue/IDEA-251181


chashnikov:

PR #1447, opened 2020-09-29
Fix order of the Trinity constructor assignment

JoeStead:
I saw a screenshot of this constructor on Twitter and it needed fixing


klikh:
Applied manually. Thanks.

PR #1454, opened 2020-10-13
Merge roots changed notifications from SourceFolderManager

yuriy-solodkyy:
It takes ages to process notifications when re-opening a 1000 module
project after `git clean -dfx`. This change merges all the notifications
produced by the SourceFolderManager.


nskvortsov:
Merged manually, thank you!

nskvortsov (approved):

yuriy-solodkyy:
@nskvortsov I'm not sure I implemented it correctly. Does `mergeRootsChangesDuring` need to run in the EDT? (to be able to call `getBatchSession(false).rootsChanged()`)?

nskvortsov:
Yes, it turned out `mergeRootsChangesDuring` is not thread safe and should be called from EDT, with write lock. I will try to implement it soon, will post link to commit here, when it is available

nskvortsov:
Re-implemented as https://github.com/JetBrains/intellij-community/commit/cd0bce508cab637d3b7be8c1d33387bd857c8704

PR #1456, opened 2020-10-14
IDEA-252935 Lookup class from MethodLocation if available

bmuskalla:
This allows to lookup the class for Spock tests as `getParentElement(PsiClass.class)`
does not work for `GrMethodImpl` elements (their parent is actually a `GrClassBody`).


bmuskalla:
Abandoning this PR as I don't intend to further work on this. Feel free to reuse the code in here.

PR #1457, opened 2020-10-19
Remove redundant .toString()

asashour:
In String concatenation, `.toString()` is not needed, since it is implicitly called by the compiler.

This PR doesn't cover all cases, but it is initial before adding more PRs.


yole:
Please see my comment to #1462.

PR #1458, opened 2020-10-22
reuse Gradle connections slightly less

csrhodes:
If we attempt to make a connection with a Gradle version which is not
currently installed, `gradleHome` in the connector parameters is null,
no matter which version of Gradle is requested.  The connector will
then download and suitably install the version of Gradle requested,
but the null value of `gradleHome` forms parts of the key for looking
up existing connections.

If there is a subsequent request to make a connection with a different
Gradle version which is also not installed, the existing stored
connection (with a different Gradle version) will be re-used rather
than installing and using the requested version.

To fix this, consider ConnectorParameter objects as different (not
equals()) if either of them has a `gradleHome` that is null.

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/GradleConnectorService.kt
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/GradleConnectorService.kt
@@ -110,7 +110,27 @@ internal class GradleConnectorService(@Suppress("UNUSED_PARAMETER") project: Pro
     val wrapperPropertyFile: String?,
     val verboseProcessing: Boolean?,
     val ttlMs: Int?
-  )
+  ) {
+    override fun equals(other: Any?): Boolean {
+      if (this === other) return true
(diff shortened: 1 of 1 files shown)


csrhodes:
(I work for Google, so I believe I do not need to sign the CLA)

csrhodes:
I have filed [IDEA-253631](https://youtrack.jetbrains.com/issue/IDEA-253631) to describe the symptoms I have seen more precisely.

csrhodes:
Hi!  Is there anything I can do to help this along?  (I'd be happy to help address any concerns).

vladsoroka:
Hi Christophe! 
Thank you for the fix. It has been merged using https://github.com/JetBrains/intellij-community/commit/4daebdb8bd0580fd748c9b0608e20c5c72c6fd97

You can also find my small refactoring change at https://github.com/JetBrains/intellij-community/commit/2c78805b342d1734b48c61fbe37f10c85769f88b

csrhodes:
Thank you very much!

PR #1459, opened 2020-10-23
Add Invoker#forBackgroundPoolWithoutReadAction to allow parallel exec…

abrooksv:
…ution without locks

We populate a few of our AsyncTreeModels with data populated from network calls, so we do not want to run it under a read lock, but we do want the calls to execute in parallel.

8 threads was chosen to match the other backgroundPool method

### DIFF
--- a/platform/platform-impl/src/com/intellij/util/concurrency/Invoker.java
+++ b/platform/platform-impl/src/com/intellij/util/concurrency/Invoker.java
@@ -571,6 +571,11 @@ public abstract class Invoker implements Disposable {
   }
 
   @NotNull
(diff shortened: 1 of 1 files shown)


abrooksv:
We are trying to move off the deprecated method:

```
    @Deprecated
    @ScheduledForRemoval(inVersion = "2021.1")
    public Background(@NotNull Disposable parent, @NotNull ThreeState useReadAction, int maxThreads) {
```

abrooksv:
@SergeyMalenkov This API is marked for removal in 211 so hoping this can get looked at before the 211 branch is cut.

SergeyMalenkov:
Done. Sorry for late response.

PR #1462, opened 2020-10-26
Replace size zero check with 'isEmpty'

asashour:
This change removes the suggestion to use `isEmpty` in kotlin.

This change is also applied to `.java`


yole:
Thanks for your contribution! The general policy in the IntelliJ team is to apply code cleanups when there's also a need to make substantial changes to the file. Applying random quickfixes to files which haven't been touched in years brings no substantial value to the project, and validating and merging each PR requires a non-zero effort from a developer (we can't simply use the Merge button because the intellij-community repo is a partial mirror of the main development monorepo). Because of that, unfortunately we won't merge this PR or other PRs of this nature. You're welcome to contribute PRs that fix bugs or usability problems; such contributions are actually valuable for us.

PR #1464, opened 2020-10-27
IDEA-252328: Consider Java installations from Gradle

bmuskalla:
Gradle 6.7 introduced support for toolchains. This allows Gradle builds to define the environment required to run the build. In case there is no matching JDK, Gradle may download a matching JDK and install it. If this happens and a project is loaded in IntelliJ, the respective JDK is not found and requires the user to manually point IntelliJ to the Gradle JDK folder.


nskvortsov:
Thank you for the contribution. I think we first need to introduce extensibility to JavaHomeFinders, so we can provide a Finder from Gradle plugin, properly referencing Gradle settings
@jonnyzzz please adivse

jonnyzzz:
The additional API is good to have, but it'd too complex. I like the simplicity of the proposed patch, let's merge it and improve once the additional EP is implemented. 

@nskvortsov do you have any statistics on how often our uses use the customized Gradle path?

nskvortsov:
merged manually, thank you!

PR #1465, opened 2020-10-27
IDEA-229823: Add more YAML settings to editorconfig

remen:
By merging the two CodeStyleSettingsProviders and renaming createSettingsPage to createConfigurable
the editorconfig plugin can now pick up the custom settings that the yaml plugin provides.

Two caveats:

* There is no code-completion for acceptable values for `ij_yaml_align_values_properties` (Wrapping and Braces > Align values in maps) which is using an integer to enumerate values. I decided to leave that out of this PR since it's not clear to me how to fix this properly.
* There was an existing failing test case in the `yaml` plugin on the `master` branch. It is still failing in this PR.

### Before
image

### After
image


plugins/yaml/resources/META-INF/plugin.xml
```diff
@@ -113,4 +112,4 @@
     </extensionPoint>
   </extensionPoints>
 
-</idea-plugin>
```
steelart, line 116: Why do you change the last line? :)

plugins/yaml/src/org/jetbrains/yaml/YAMLLanguageCodeStyleSettingsProvider.java
```diff
@@ -32,6 +27,41 @@
     };
   }
 
+  @NotNull
+  @Override
+  public CodeStyleConfigurable createConfigurable(@NotNull final CodeStyleSettings settings, @NotNull final CodeStyleSettings originalSettings) {
+    return new CodeStyleAbstractConfigurable(settings, originalSettings, YAMLLanguage.INSTANCE.getDisplayName()) {
```
steelart, line 33: I think it would be better to reuse `YAMLCodeStyleSettingsProvider` here instead of inline it.

plugins/yaml/resources/META-INF/plugin.xml
remen: Sorry my bad. Editor configured to add missing eof newlines. Will revert this change.

plugins/yaml/src/org/jetbrains/yaml/YAMLLanguageCodeStyleSettingsProvider.java
remen: Yeah, I was a bit unsure of the approach I would take. Will do as you suggest. Question though: does that mean that you alsow want the `codeStyleSettingsProvider` as part of `plugin.xml`? I looked through some of the other plugins and I didn't see a clear pattern on whether both classes are needed.

remen: Done. I made a guess that you wanted the minimal amount of changes necessary that would still solve the issue, so that is what I aimed for.

steelart:
Thank you! Will it be only the one commit? Now there is 3 of them.

plugins/yaml/src/org/jetbrains/yaml/YAMLLanguageCodeStyleSettingsProvider.java
```diff
@@ -18,6 +15,9 @@
 import static com.intellij.psi.codeStyle.CodeStyleSettingsCustomizableOptions.getInstance;
 
 public class YAMLLanguageCodeStyleSettingsProvider extends LanguageCodeStyleSettingsProvider {
+
+  private static final YAMLCodeStyleSettingsProvider YAML_CODE_STYLE_SETTINGS_PROVIDER = new YAMLCodeStyleSettingsProvider();
```
steelart, line 19: So there will be 2 instances: here and the instance is created by `plugin.xml`. Am I right?

remen: That is correct. Again, I don't quite understand the difference between the two providers, especially since `LanguageCodeStyleSettingsProvider` inherits from `CodeStyleSettingsProvider` (if I recall correctly). I don't know of any mechanism with which I could inject the instance from `plugin.xml` into this class, but to be honest I'm not extremely comfortable with this project.

remen:
> Thank you! Will it be only the one commit? Now there is 3 of them.

If you wish, I can squash them. Just say the word.

steelart:
Thank you, @remen ! I applied your first commit (it was closest to the solution) and added deprecation.

PR #1466, opened 2020-10-30
Add support for parametrized model builders injected from the IDE

yuriy-solodkyy:
Gradle since version 4.4 supports ParameterizedToolingModelBuilder
interface which allows model builders to be configured form the IDE.
This is used by the Android plugin to import only the build variant
currently selected in the IDE.

This change make it possible to configure model builders injected from
the IDE in a similar manner. This should allow the Android plugin to
communicate the currently selected build variant to the Kotlin plugin
and thus not to import unnecessary source sets.

NOTE: `ExtraModelProvider` still implements `ModelBuilder` and its 
behavior does not change. This pull requests alters the behavior when 
the IDE has already configured a parameter only. In this case Gradle
attempts to use `ParameterizedToolingModelBuilder` instead of 
`ModelBuilder`.

### DIFF
--- a/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/ExtraModelBuilder.java
+++ b/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/ExtraModelBuilder.java
@@ -10,6 +10,7 @@ import org.gradle.internal.impldep.com.google.common.collect.Lists;
 import org.gradle.internal.impldep.com.google.gson.GsonBuilder;
 import org.gradle.internal.logging.progress.ProgressLogger;
 import org.gradle.internal.logging.progress.ProgressLoggerFactory;
+import org.gradle.tooling.provider.model.ParameterizedToolingModelBuilder;
 import org.gradle.tooling.provider.model.ToolingModelBuilder;
 import org.gradle.util.GradleVersion;
 import org.jetbrains.annotations.ApiStatus;
@@ -33,6 +34,15 @@ import java.util.*;
  * @author Vladislav.Soroka
  */
 public class ExtraModelBuilder implements ToolingModelBuilder {
+
+  public static class ForGradle44 extends ExtraModelBuilder implements ParameterizedToolingModelBuilder<ModelBuilderService.Parameter> {
+    @NotNull
+    @Override
+    public Class<ModelBuilderService.Parameter> getParameterType() {
+      return ModelBuilderService.Parameter.class;
+    }
+  }
+
   private static final Logger LOG = LoggerFactory.getLogger(ExtraModelBuilder.class);
   @ApiStatus.Internal
   public static final String MODEL_BUILDER_SERVICE_MESSAGE_PREFIX = "ModelBuilderService message: ";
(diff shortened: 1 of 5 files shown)


yuriy-solodkyy:
@vladsoroka @nskvortsov this is the only change required for Kotlin single-variant sync.

nskvortsov:
Unfortunately, the `ExtraModelProvider` has to stay compatible with Gradle versions before 4.4 (IDEA 2020.3 can import projects from Gradle 2.6 onwards) and should not implement `ParameterizedToolingModelBuilder` (this interface can not be loaded on earlier Gradle versions)

yuriy-solodkyy:
> Unfortunately, the `ExtraModelProvider` has to stay compatible with Gradle versions before 4.4 (IDEA 2020.3 can import projects from Gradle 2.6 onwards) and should not implement `ParameterizedToolingModelBuilder` (this interface can not be loaded on earlier Gradle versions)

Would something like the new version I uploaded be acceptable? i verified that IDEA built from sources with this change can import projects using an older version of Gradle.

plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/ExtraModelBuilder.java
```diff
@@ -97,6 +111,8 @@ public Object buildAll(String modelName, Project project) {
             if (service instanceof AbstractModelBuilderService) {
               return ((AbstractModelBuilderService)service).buildAll(modelName, project, myModelBuilderContext);
             }
+            else if (service instanceof ModelBuilderService.Parametrized)
```
vladsoroka, line 114: we can avoid additional `instanceof` and new interface `ModelBuilderService.Parametrized` using `org.jetbrains.plugins.gradle.tooling.ModelBuilderContext`

yuriy-solodkyy: Do you mean I should add a new method to `ModelBuilderContext` instead?

vladsoroka: Yep, I'd like to deprecate `ModelBuilderService#buildAll(String modelName, Project project)` method in favour of `ModelBuilderService#buildAll(String modelName, Project project, ModelBuilderContext context)`.
At some moment in future when minimal language level of TAPI modules will be set to 1.8 and we could use "default" interface methods instead of abstract class.

yuriy-solodkyy: `AbstractModelBuilderService` is a class so I cannot just implement it in one Kotlin model builder which is derived fro their abstract model builder which implements `ModelBuilderService` directly. Since we do not have default methods right now, I can add an `Ex` interface with this method, make `AbstractModelBuilderService` implement it and then in the Kotlin plugin implement the interface independently. I'm not sure it is feasible to update all `buildAll` methods in the Kotlin plugin right now.

yuriy-solodkyy: See the second commit I uploaded to illustrate the said above.

nskvortsov:
merged manually: https://github.com/JetBrains/intellij-community/commit/7b6c1079e304b6b1bd5c4b152f288cc5031f7dc7

PR #1473, opened 2020-11-11
[EXAMPLE] Partially fix data model deserialization slowness

yuriy-solodkyy:
@nskvortsov This is the simple part which handles the case when the whole object is deserializaed via the constructor instantiation. See my comments in the issue itself for how it can be extended to other caces.


nskvortsov:
Closing an obsolete PR

PR #1474, opened 2020-11-12
Replace widescreen mode with more granular controls

MTCoster:
I’ve long been frustrated by the lack of granularity in the widescreen mode for tool windows. With this patch, those days are over.

The little bit of empty space in each corner between the stripes is now a button to toggle which of the adjacent tool windows extends to the corner. These options are also available under `Settings > Appearance & Behavior > Appearance > Tool Windows`.

To implement this, I created `ToolWindowsSplitter` to replace the existing implementation of two nested `ThreeComponentsSplitter`s; and added a new inner class `Corner` to `ToolWindowsPane`. Together, these perform all the layout operations the previous implementation did, but with the further granularity afforded by the new options.

Comments & suggestions are very welcome, this is my first contribution to the project as a long-time user.


MTCoster:
Is this likely to ever get any attention? I'm happy to update it to resolve merge conflicts, but not if it's just going to be ignored.

PR #1478, opened 2020-11-21
SystemProperties: use Boolean.getBoolean

simon04:


platform/util/src/com/intellij/util/SystemProperties.java
```diff
@@ -75,19 +75,18 @@ public static float getFloatProperty(@NotNull String key, float defaultValue) {
    * Returns the value of given property as a boolean, or {@code defaultValue} if the property is not specified or malformed.
    */
   public static boolean getBooleanProperty(@NotNull String key, boolean defaultValue) {
-    String value = System.getProperty(key);
-    return value != null ? Boolean.parseBoolean(value) : defaultValue;
+    return Boolean.getBoolean(key) || defaultValue;
```
rendner, line 78: This changes the behavior of the method. 
```
        System.setProperty("myBool", "false");

        System.out.println(getBooleanPropertyOld("myBool", true));  // returns false
        System.out.println(getBooleanPropertyNew("myBool", true));  // returns true
    }

    public static boolean getBooleanPropertyOld(String key, boolean defaultValue) {
        String value = System.getProperty(key);
        return value != null ? Boolean.parseBoolean(value) : defaultValue;
    }

    public static boolean getBooleanPropertyNew(String key, boolean defaultValue) {
        return Boolean.getBoolean(key) || defaultValue;
    }
```

simon04: Err, I'm stupid… Sure!

PR #1480, opened 2020-12-06
IDEA-238863: Allow keeping app open after WelcomeScreen is closed

nvartolomei:
Details https://youtrack.jetbrains.com/issue/IDEA-238863

It would be nice to move the focus to "Search projects" field but it
took me too much time already to figure out how to get this far :)

Also, it would be nice to implement https://youtrack.jetbrains.com/issue/IDEA-238865 as well for best experience.


nvartolomei:
I guess there should be something like FakeWelcomeFrame that still captures keyboard events. If it is hidden then keyboard events are not captured anymore. This would also help with https://youtrack.jetbrains.com/issue/IDEA-238865. Currently if I add an if statement CloseProjectWindowHelper to avoid WelcomeFrame.showIfNoProjectOpened() call then the IdeMenuBar of the project is still visible (the items are displayed but they are not functional).

nvartolomei:
Pages.app behaviour https://www.youtube.com/watch?v=kI-Kyg7cOno
Desired IntelliJ IDEA behaviour https://www.youtube.com/watch?v=lHoevzDnmjw

nvartolomei:
Updated second video from previous comment as it was trimmed incorrectly.

PR #1481, opened 2020-12-07
launcher: behave identically whether or not IDE is already running

orent:
When the IDE is not already running the launcher always waits until it exits.  This patch runs it in the background unless the --wait flag is passed.

Also redirect stderr to /dev/null to keep IDE noise from interfering with the console.
Uses the shell, but safe from any string or quoting issues as arguments are forwarded as a vector.

See IDEA-249381 (MacOS specific)

### DIFF
--- a/platform/platform-resources/src/launcher.py
+++ b/platform/platform-resources/src/launcher.py
@@ -133,9 +133,11 @@ def start_new_instance(args):
             args.insert(0, '-W')
         os.execv('/usr/bin/open', ['open', '-na', RUN_PATH] + args)
     else:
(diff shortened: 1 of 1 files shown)


orent (commented):
Oops.  I got the condition backwards. It should be `if '--wait' not in args`, of course

trespasserw:
Thanks for the contribution!

The main problem with the patch is that the script has numerous issues and is going to be replaced soon. Please vote/follow https://youtrack.jetbrains.com/issue/IDEA-205244 for the progress.

PR #1482, opened 2020-12-08
PY-45894 - fix wrong rendered cell background

rendner:
A DefaultTableCellRenderer doesn't automatically reset the foreground/background color before painting the next cell. The renderer re-uses these colors for each unselected cell, if they are not explicitly overwritten.

To fix this problem, the "colorize" method is called before "super.getTableCellRendererComponent" so that the super implementation can apply a default background color in case that "colorize" cleared the background by setting it to "null".


Elizaveta239:
Hi! Thank you very much for your pull request! I'll pass it to our QA team

Elizaveta239:
@rendner Thank you very much for your contribution! The fix will appear in the next PyCharm 2020.3 update!

PR #1484, opened 2020-12-14
Why is Google the search engine?

Hallowizer:
Ignore the commits attached to this. There's no issue button so I'm saying this here. Why is there a "Search with Google" button? Why would you promote such an invasive search engine? This is uncool. I'm not even able to change it to DuckDuckGo!


ybedrov:
Could you please clarify what button do you mean?

Hallowizer:
The one on the right click menu that appears when you select some text.

ybedrov:
Please vote for the related request on YouTrack: https://youtrack.jetbrains.com/issue/IDEA-133410

therealbluepandabear:
> The one on the right click menu that appears when you select some text.

I know I'm very late to this, but I've fought to get the issue added here: https://youtrack.jetbrains.com/issue/IDEA-289262

And, according to a member of JetBrains, this will be added soon:

image

PR #1485, opened 2020-12-15
IDEA-257798 Bump Kotlin plugin to 1.4.21

ChaseKnowlden:
Upgrade Kotlin Plugin to 1.4.21


zarechenskiy:
Hello,

Unfortunately, we can't accept your PR. With the compiler upgrade in IJ, we also have to publish new IDE Maven artifacts that will be used in several IJ parts, some of them are in the Ultimate part. So, for now, this commit will break IJ compilation.

Anyway, thank you for your effort and attention! We'll upgrade Kotlin version quite soon.

Also, to avoid misunderstandings in the future, I recommend commenting on the relevant issue before actual work

LifeIsStrange:
@zarechenskiy can you update intellij 2020.1 beta to kotlin 1.4.31? It's really a painful mismatch that intellij stable is on 1.4.31 and that intellij EAP/beta is on 1.4.20 especially since heuristically when you're on EAP you want the latest versions. 
I've seen many people complain about it on the various intellij EAP releases blog posts

PR #1487, opened 2020-12-17
Use original repositories in fallback download script

janchochol:
"project.repositories.toList().find" loops resets repositories, so
fallback "configuration.resolve()" uses only last defined repository
leading to errors, when downloaded source requires something from
different repository.

This fix first backup original repositories, and than restore them after
"project.repositories.toList().find".


jimirocks:
@vladsoroka Is something blocking this raher trivial contribution? This is effectively blocking usage of sources and docs from different repositories 😞

nskvortsov:
This PR is now obsolete. as the "download sources" feature was largely reworked. Feel free to reopen if it is still relevant.

PR #1489, opened 2020-12-24
'Embedded' attribute type should not be 'Embeddable'

isfong:
截屏2020-12-24 上午10 47 43
截屏2020-12-24 上午10 48 30

Hi sir, why?


ybedrov:
Hello! Let's continue in YouTrack issue: https://youtrack.jetbrains.com/issue/IDEA-258414.

PR #1490, opened 2020-12-30
PY-33060 Some extra options for python formatter (arguments/parameters wrapping etc)

RuRo:
This PR implements

1. `Use continuation indent for... Method declaration parameters` option (`Other` tab).
2. `Newline after '('` and `Place ')' on new line` options for `Method declaration parameters`.
3. `Do not wrap`/`Wrap if long`/`Chop down if long`/`Wrap always` options for `Method declaration parameters`.
4. `Newline after '('` and `Place ')' on new line` options for `Method call arguments`.
5. `Do not wrap`/`Wrap if long`/`Chop down if long`/`Wrap always` options for `Method call arguments`.


kkulczak:
It would be extremely helpful to include this feature int the newest release.

east825:
@RuRo Implementation looks good to me, sorry it took so long to review. The only thing that bothers me, however, is the new option for disabling a continuation indent for function parameters. I guess it's fine for consistency sake because we already have a similar option for call arguments. But using the normal indent for them almost immediately violates PEP 8, unless they also separated from the following function body with a parenthesis. Do you really encounter such a code style in practice? It wasn't mentioned as part of PY-33060, as far as I see.

RuRo:
@east825 yes, the intended use-case is when enabled together with `Place ')' on new line` for declarations.

```python
def foo(
    arg1, arg2, arg3,
    kwarg1="default",
    kwarg2="value",
):
    something.other()
    return [123, 456]


foo(
    val1, val2, val3,
    kwarg1="other",
    kwarg2=None,
)
```

The idea is to have the same indent for every kind of expression/statement, without compromising the separation between the arguments and the body.

As for whether this style is used in practice, this is the style I use for my personal projects. I don't know, if that makes me a filthy PEP8 violator :grimacing:, but I like to argue that this is an acceptable variation within the "PEP8 spirit".

east825:
Ok, got it! Any options that make us more compatible with Black are appreciated :) To be honest, we already have a number of formatting options conflicting with PEP 8. We just want to ensure that the defaults are compliant and assume that whoever tinkers with the settings knows what they're doing (and don't mind being called "filthy violators").

I'm merging the PR as-is, and it will be available in 2021.1. Actually, further PRs in this area are more than welcome, since you already have a knack. For instance, having Python versions of "Put parameters on separate lines" / "Put arguments on separate lines" intentions would be really cool (just saying 😇).

Thanks a lot for your contribution!

RickMeasham:
Thanks @RuRo! And thanks @east825 and @SergeyZh for merging!

PR #1491, opened 2021-01-11
Add telegram requirement name for python-telegram-bot package

revolter:


vladimir-koshelev:
Fix a year ago together with package prefix update

PR #1493, opened 2021-01-14
Fix a typo in ProjectRootsChangeListener

kovalpgoogle:
Typo: "==" instead of "=". This bug causes that sporadically after adding roots they are not indexed.

If we have a sequence of events there is "remove" event  before "add" even, due to a bug a final type will be `RootsChangeType.ROOTS_REMOVED` even if we have "add" event. Therefore we never trigger an indexing for these changes.


kovalpgoogle:
Ping :) It's important as it effects an indexing process

batya239:
Thanks, I've applied a patch but our tests have failed, I'm investigating the reason

PR #1494, opened 2021-01-15
update values in wsl distribution descriptor

javiersantossanchez:
fixing the wrong msdi value for PENGWIN distribution, this error generate error to user when they try to connect to wsl PENGWIN distribution. This bug is present on several tools of intellij. 

Removing unsupported wsl distribution "WLINUX" . this distribution was renamed as PENGWIN by WhitewaterFoundry project, to avoid problems with different distributions with same msdi.  (this is the root cause of wrong msdi value for PENGWIN)

adding some missed distribution. to allow to users work with Intellij on theses wsl linux distributions.

### DIFF
--- a/platform/platform-impl/src/com/intellij/execution/wsl/WSLDistributionService.java
+++ b/platform/platform-impl/src/com/intellij/execution/wsl/WSLDistributionService.java
@@ -46,6 +46,8 @@ final class WSLDistributionService implements PersistentStateComponent<WSLDistri
     new WslDistributionDescriptor("OPENSUSE42", "openSUSE-42", "opensuse-42.exe", "openSUSE Leap 42"),
     new WslDistributionDescriptor("SLES12", "SLES-12", "sles-12.exe", "SUSE Linux Enterprise Server 12"),
     new WslDistributionDescriptor("SLES15", "SLES-15", "sles-15.exe", "SUSE Linux Enterprise Server 15"),
+    new WslDistributionDescriptor("SLES15-SP1", "SLES-15-SP1", "SLES-15-SP1.exe", "SUSE Linux Enterprise Server 15 SP1"),
+    new WslDistributionDescriptor("SLES15-SP2", "SUSE-Linux-Enterprise-Server-15-SP2", "SUSE-Linux-Enterprise-Server-15-SP2.exe", "SUSE Linux Enterprise Server 15 SP2"),
     new WslDistributionDescriptor("OPENSUSE15", "openSUSE-Leap-15", "openSUSE-Leap-15.exe", "openSUSE Leap 15"),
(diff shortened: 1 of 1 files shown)


sourcedelica:
Per my comment in https://youtrack.jetbrains.com/issue/IDEA-254998#focus=Comments-27-4572866.0-0, it seems like `wsl.distributions.xml` is the wrong approach for this problem.  There are cases that it that doesn't handle, like 
1. Custom distros like the one provided by my company that are unrelated to the Windows store
2. "Distros" that are created by `wsl --import` that don't even have a distro launcher `.exe`

`wsl.distributions.xml` assumes that you need a distro launcher `.exe` to run commands in the distro but you can run any command in a distro using `wsl -d distro command`.  And you can map `\\wsl$` paths to distros using `wsl --list`.

crramirez:
I agree. This file could be omitted at all or at least use as a piece of complementary information for the well-known distros.

I found this line https://github.com/JetBrains/intellij-community/blob/8a2ff58fcd796badb74d11f5d90acc2afe35a69b/platform/platform-impl/src/com/intellij/execution/wsl/WSLCommandLineOptions.java#L19 that gives me an idea that the support for distros without exe is implemented.

Regards,
Carlos

javiersantossanchez:
> I agree. This file could be omitted at all or at least use as a piece of complementary information for the well-known distros.
> 
> I found this line
> 
> https://github.com/JetBrains/intellij-community/blob/8a2ff58fcd796badb74d11f5d90acc2afe35a69b/platform/platform-impl/src/com/intellij/execution/wsl/WSLCommandLineOptions.java#L19
> 
> that gives me an idea that the support for distros without exe is implemented.
> Regards,
> Carlos

javiersantossanchez:
I also make a search and I found this code https://github.com/JetBrains/intellij-community/blob/c8994e6afb7d15804a674934f6b00a80419d5bd7/plugins/git4idea/src/git4idea/config/GitExecutableDetector.java#L176 
Here the code looking for the executable of git, but under wsl environment, you can see they are using only well-known distros.  
I don't know is only happens with git support.

crramirez:
Looks like they are working on getting rid of this file:

https://github.com/JetBrains/intellij-community/blob/8a2ff58fcd796badb74d11f5d90acc2afe35a69b/platform/platform-impl/src/com/intellij/execution/wsl/WslDistributionManagerImpl.java#L48

Also

https://github.com/JetBrains/intellij-community/blob/8a2ff58fcd796badb74d11f5d90acc2afe35a69b/platform/platform-impl/src/com/intellij/execution/wsl/WSLUtil.java#L54

Although the old API will be removed on 2021.3 version, they should accept this PR to ensure that any plugin that still uses the old API will continue working.

segrey:
Thanks for the PR! Merged manually as 96b7d01006e8b615edf6be638b99021ce96076d7.

crramirez:
Excellent!!

javiersantossanchez:
Siiiii. Finalmenteeeee

PR #1495, opened 2021-01-16
Fix: "no module" can not be selected

hebo6:
[issue description at youtrack](https://youtrack.jetbrains.com/issue/IDEA-259609)


hebo6:
I saw that my issue IDEA-259609 on YouTrack was flagged for duplicates, but actually it wasn't. 
My issue IDEA-259609 is that: the "Scratch" configuration menu item had "no moduel" and should be able to be selected, but it can't. 
The another issue IDEA-258146 is that: the "JUnit" configuration menu shouldn't present "no moduel" menu item, but it does

anstarovoyt:
merged via 5447cd407cb276ffa8399bcde2c38ca6cd0887f7

PR #1498, opened 2021-01-21
Added checkbox for enable fixed size of editor tabs

vjmurenko:
Hello.

I think it's nice to have an option where the width of the tabs is always equal to the label. Because when the multi-row mode is enabled, the new tab is wrapped to a next line, then it stretches to full screen, jumps, this may be inconvenient for some people.

There were several issues about current behavior of multi row tabs , main is:
https://youtrack.jetbrains.com/issue/IDEA-22546 

Thank you.


bulenkov:
Hey. We're working on editor tabs at the moment and will give a status update a bit later. 
image

bulenkov:
We have fixed the problem. Thanks.

PR #1500, opened 2021-01-23
Update IStubElementType.java

ishmum123:
This throws compilation error in the implemented method of the Concrete Class


ishmum123:
@develar do you have any suggestions?

PR #1503, opened 2021-01-27
Fix memory leak in PsiFileBreadcrumbsCollector

npaige:
Pass the parent disposable to addModelListener so that the listener will
be removed when the editor tab is closed.


piotrtomiak:
Thanks for the fix!

PR #1504, opened 2021-01-28
Retain @TestDataFile annotation in .class files

gharrma:
That way test data navigation still works even when IntelliJ
is on the classpath in binary form.

cc @yole since you added the annotation :-)


yole:
Merged manually, thanks!

PR #1505, opened 2021-01-30
Fix sdkman detection in WSL

crramirez:
The java home detection in the case of sdkman, first look for the environment variables: SDKMAN_CANDIDATES_DIR, SDKMAN_DIR

In WSL it executes: 

`wsl.exe --distribution WLinux --exec /bin/sh -c printenv SDKMAN_CANDIDATES_DIR` 

It won't work because these variables in the standard installation are initialized calling a script in .bashrc. So the command should change to bash --login. Even with this change, the script is not executed before printenv

The solution is to look for the default path in Linux home, but the code checks if we are in Windows, so I added a small check telling that this piece of code is valid also in WSL

### DIFF
--- a/platform/lang-impl/src/com/intellij/openapi/projectRoots/impl/JavaHomeFinderBasic.java
+++ b/platform/lang-impl/src/com/intellij/openapi/projectRoots/impl/JavaHomeFinderBasic.java
@@ -230,7 +230,7 @@ public class JavaHomeFinderBasic {
     }
 
     // finally, try the usual location in UNIX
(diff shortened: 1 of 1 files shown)


trespasserw:
Thanks!

Environment reading was updated recently, now it works for variables set in `.profile` (and using `.bashrc` is bad idea anyway).

crramirez:
Excellent! .bashrc is the way that sdkman installs by default when you run: `curl -s "https://get.sdkman.io" | bash`

PR #1506, opened 2021-02-05
Logging message in catch block

hebo6:
Needed to log message because nothing appear in the console after it thrown IOException


yole:
A `LOG.error` would be reported as an exception to our exception analyzer (as a bug that we would need to fix). In this case there is no bug and the error has already been shown to the user, so using `LOG.error` is not appropriate.

PR #1509, opened 2021-02-06
IDEA-261503 incorrect rendering in git panel

hebo6:


hebo6:
Issue in youtrack associated this PR is [here](https://youtrack.jetbrains.com/issue/IDEA-261503)

juliabeliaeva (changes requested):
This patch has more changes than it should which complicates the code unnecessarily. The issue is caused by `graySpan.toString()` printing `<span =.../>` and not `<span =...>`, but it's not evident from the change. Extra "ifs" make the code harder to understand. One of the clauses adds a return statement, which is in contrast with others (which do not), this also makes the code harder to parse. Also commit message does not explain what was broken.

By the way, it is recommended that commit messages were written in imperative. The commit message should be able to fit into this sentence "If applied, this commit will <message>". This, of course, is just a recommendation, not a rule.

hebo6:
Thank you for review

This method ```getCommitterHtml()``` will return ```<span ...>commit by ...</span><a>...@email.com</a><span .../></span>```, which contains an empty span at the end, if ```commiter != null && emptyEmail == false && commitTime == null``` without the first return statement.

About "ifs" is too many, I haven't found a good way to extract. Can you give me some suggestions? thanks

juliabeliaeva:
I don't think the empty "span" needs fixing.

I'm going to push your change with the "<span =.../>" vs "<span =...>" fix. Thank you for the PR!

PR #1510, opened 2021-02-07
[python] Don't annotate type: ignore[error-code] syntax

FuegoFro:
This follows up on 9d34a41ee58e058e577e625fe9b21ff7d77bc9ea and c1e63b4b424b8ab830c1ae43293bebe7e112a333, which made it so that `# type: ignore` isn't annotated (eg so the IDE doesn't complain about "ignore" not being found). This extends that for the new [error code specific](https://mypy.readthedocs.io/en/stable/error_codes.html#silencing-errors-based-on-error-codes) ignore syntax, eg `# type: ignore[no-untyped-def]`. Without this change the error code is highlighted as an error itself.

This should fix PY-38213.


mano7onam:
Thank you very much for the pull request :)

PR #1511, opened 2021-02-11
Parallel Gradle sync support for Gradle 6.8+

yuriy-solodkyy:
Gradle 6.8 has introduced support for querying build models in parallel. 

It can significantly reduce the time required to import a Gradle project as it allows dependency resolution to happen in parallel. 

This series of changes makes the IDE Gradle plugin to request models via a new in Gradle 6.8 API (`org.gradle.tooling.BuildController#run`), which  executes  build action in parallel if connected to Gradle 6.8 or later. The API is backward compatible and executes the requests sequentially if connected to an older version of Gradle. 

The first two commits restructures the code to separate the phase that requests models from the phase that serializes and collects models. 

The other two commits adds thread safety to performance logging methods and migrates model fetching to the new Gradle APIs.

@nskvortsov

### DIFF
--- a/plugins/gradle/tooling-extension-api/src/org/jetbrains/plugins/gradle/model/ProjectImportAction.java
+++ b/plugins/gradle/tooling-extension-api/src/org/jetbrains/plugins/gradle/model/ProjectImportAction.java
@@ -142,8 +142,12 @@ public class ProjectImportAction implements BuildAction<ProjectImportAction.AllM
   }
 
   private void fetchProjectBuildModels(BuildController controller, final boolean isProjectsLoadedAction, GradleBuild build) {
+    List<Runnable> result = new ArrayList<Runnable>();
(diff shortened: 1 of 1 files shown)


vladsoroka (commented):
LGTM, thanks Yuriy!
I've started integration tests with the change on CI

yuriy-solodkyy:
Our tests passed with these changes ported back to Android Studio.

vladsoroka:
Merged: 53cf8d93e7a49f1c915744e9133afed401b3a16c, 6f9fffb101c447c439dcc14e918634c71b323b69, 16146d05546a101357b3e0e5375ba79f128bb689, 0b7e3e1045c2b7cf0c298ca72e57b1f513d78bf9

yuriy-solodkyy:
Thank you!

PR #1512, opened 2021-02-11
Add pep8 codes to suppression map

mickp:
This PR adds support for suppression of PEP8 naming checks with `noqa`.

* 1st commit adds codes, each with a comment describing their cause.
* 2nd commit removes the inline comments.


east825:
Thanks! Merged only the first commit with the comments explaining error codes.

PR #1514, opened 2021-02-15
IDEA-125412: Allow specifying amount of lines of context when creating a patch from a commit

zero9178:
Patches created by IntelliJ and derivatives currently have a context of precisely 3 lines. When using code review platforms such as Phabricator it is often desirable to add more context to the patch as to help the reviewing process. 

This pull request contains the following changes:
* Added an optional contextLineCount parameter to various methods allowing a patch to be created with the given context lines count
* Added a textfield to the action allowing a patch to be created from a commit which plugs into the above

Out of scope for this pull request:
Ability to specify amount of context wanted when creating a patch from other actions (from a changelist eg.)


dmitriysmirnovspb:
Thank you for the contribution. We applied the patch, but currently the number of lines is defined via IDE registry - `patch.context.line.count`
It will be eventually moved to IDE Settings, when Settings pages get updated soon.

dmitriysmirnovjb:
Thank you for the contribution. We applied the patch, but currently the number of lines is defined via IDE registry - `patch.context.line.count`
It will be eventually moved to IDE Settings, when Settings pages get updated soon.

PR #1516, opened 2021-02-19
Expose the private LineOffsets data in FileOffsetsManager with getLineNumber

jwren:
Context: Dart plugin, https://github.com/JetBrains/intellij-plugins/pull/791

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/fileEditor/impl/FileOffsetsManager.java
+++ b/platform/platform-impl/src/com/intellij/openapi/fileEditor/impl/FileOffsetsManager.java
@@ -59,6 +59,11 @@ public final class FileOffsetsManager {
     return getCorrespondingOffset(offsets.myConvertedLineOffsets, offsets.myOriginalLineOffsets, convertedOffset);
   }
 
+  public int getLineNumber(@NotNull final VirtualFile file, final int offset) {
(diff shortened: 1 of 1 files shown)


platform/platform-impl/src/com/intellij/openapi/fileEditor/impl/FileOffsetsManager.java
```diff
@@ -59,6 +59,11 @@ public int getOriginalOffset(@NotNull final VirtualFile file, final int converte
     return getCorrespondingOffset(offsets.myConvertedLineOffsets, offsets.myOriginalLineOffsets, convertedOffset);
   }
 
+  public int getLineNumber(@NotNull final VirtualFile file, final int offset) {
+    final LineOffsets offsets = getLineOffsets(file);
+    return -Arrays.binarySearch(offsets.myOriginalLineOffsets, offset) - 1;
```
alexander-doroshko, line 64: I guess the right code would be 
```
    int line = Arrays.binarySearch(offsets.myOriginalLineOffsets, originalOffset);
    return line >= 0 ? line : -line - 2;
```
so that it behaves similar to `Document.getLineNumber()`.

alexander-doroshko:
It appears that we don't need this new method (discussed in https://github.com/JetBrains/intellij-plugins/pull/791)

PR #1517, opened 2021-02-22
IDEA-207810 End align tasks input field

ceddy4395:
Resolves https://youtrack.jetbrains.com/issue/IDEA-207810

Removes the padding at the end of the task input such that all inputs align vertically at the end in the Gradle Run Configuration UI. 
## Before
image

## After
image


nskvortsov:
Thank you for the fix! Merged it manually

PR #1518, opened 2021-02-24
[IDEA-257985] Update Darcula python syntax highlighting

ceddy4395:
Fixes https://youtrack.jetbrains.com/issue/IDEA-257985

Due to the potential confusion between error highlighting and keyword argument highlighting for python in the Darcula theme. This colour has been updated from [#AA4926](https://www.color-hex.com/color/aa4926) to [#34B8A7](https://www.color-hex.com/color/34b8a7) in order to have a better contrast. 

### Screenshot 
image

### DIFF
--- a/python/pluginResources/colorSchemes/PythonDefault.xml
+++ b/python/pluginResources/colorSchemes/PythonDefault.xml
@@ -29,7 +29,7 @@
   </option>
   <option name="PY.KEYWORD_ARGUMENT">
     <value>
-      <option name="FOREGROUND" value="660099" />
+      <option name="FOREGROUND" value="34B8A7" />
     </value>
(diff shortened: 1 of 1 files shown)


east825:
Hi! Sorry, it's taking that long to process. Changing some default color for all the users is not a simple undertaking for us, even though I agree that the current color might seem falsely alarming. Another problem is that other languages featuring a similar syntactic concept, in particular Kotlin and PHP, use another color in their scheme for that. Out UI team is evaluating options for unifying all of these. We're most likely to change the color but not necessarily to the one you suggested, and already in 2021.2 as we're past the feature freeze of 2021.1 by now. I'll post the results here once we settle down.

PR #1521, opened 2021-03-06
[Performance] Reduce UnloadedModuleDescription memory usage

ardavank:
This interns module names to a StringInterner (HashSetInterner<String>), ensuring that they share memory instead of storing many copies of each module name on the heap.
In my case, this avoids allocating about 442MB of heap space that remains pinned as long as the project is open

Before the change:
Screen_Shot_2020-11-12_at_12 20 17_PM

After the change:
Screen_Shot_2020-11-12_at_12 15 45_PM


platform/projectModel-impl/src/com/intellij/openapi/module/impl/UnloadedModuleDescriptionImpl.kt
```diff
@@ -45,16 +46,17 @@ class UnloadedModuleDescriptionImpl(val modulePath: ModulePath,
   companion object {
     @JvmStatic
     fun createFromPaths(paths: Collection<ModulePath>, parentDisposable: Disposable): List<UnloadedModuleDescriptionImpl> {
+      val interner = HashSetInterner<String>()
       val pathVariables = JpsGlobalLoader.computeAllPathVariables(PathManager.getOptionsPath())
       val modules = JpsProjectLoader.loadModules(paths.map { Paths.get(it.path) }, null, pathVariables, JpsPathMapper.IDENTITY)
-      val pathsByName = paths.associateBy { it.moduleName }
-      return modules.map { create(pathsByName[it.name]!!, it, parentDisposable) }
+      val pathsByName = paths.associateBy { interner.intern(it.moduleName) }
```
chashnikov, line 52: Why do you need to intern string here? The returned value is used inside `pathsByName` map which doesn't escape from this method.

PR #1522, opened 2021-03-07
Invoke the "which mvn" using the default shell in the distro to enable finding Maven installed through SDKMAN

crramirez:
Hello,

The sdkman-init.sh script not only initializes SDKMAN but also includes in the PATH any of the installed tools like Java and Maven. This script only works with bash and zsh. It will fail if it is invoked with /bin/sh.

This addition will invoke the command which mvn using the default shell that likely will be bash or zsh. Even if the default shell is fish, it can be invoked using the bass extension.

This gives to the user the flexibility to install through SDKMAN not only Java but also Maven.

Regards


nskvortsov:
Hello and thank you for your pull request!
Could you please clean it up, so it only contains related commits?

crramirez:
Hello,

Looks like my fork is messed up. I created a new PR from another branch. 

https://github.com/JetBrains/intellij-community/pull/1524

Regards,
Carlos

PR #1523, opened 2021-03-08
[fernflower] Fix groovy.reflection.ReflectionCache: update gradle version of fernflower

ulwlu:
We can't build fernflower with JDK 14 by current version of gradle.

ss 1
ss 2

The version of gradle 6.3 or higher is required, so I fixed.

Related Issue: https://github.com/gradle/gradle/issues/10248

I signed a contribution agreement of jetbrains.


dhouck:
I've been having a lot of issues installing Fernflower and was surprised to learn it was a simple one-line change.

I did find that the latest version, 7.0, does *not* work for reasons which are probably obvious if you know more about Java development than I do.

trespasserw:
Sorry for the late reply. I'm not sure whether updating just the URL is enough - I believe scripts and the wrapper .jar should be refreshed as well.

Anyway, I just bumped the version all the way up to 7.0.

PR #1525, opened 2021-03-10
Make ReadVariableInstruction public like other InstructionBase classes

iamKunal:
Other Instruction classes inheriting `InstructionBase` are public but only ReadVariableInstruction isn't. E.g. https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/controlFlow/WriteVariableInstruction.java#L21

### DIFF
--- a/java/java-psi-impl/src/com/intellij/psi/controlFlow/ReadVariableInstruction.java
+++ b/java/java-psi-impl/src/com/intellij/psi/controlFlow/ReadVariableInstruction.java
@@ -18,7 +18,7 @@ package com.intellij.psi.controlFlow;
 import com.intellij.psi.PsiVariable;
 import org.jetbrains.annotations.NotNull;
 
-final class ReadVariableInstruction extends SimpleInstruction {
+public final class ReadVariableInstruction extends SimpleInstruction {
   @NotNull 
   public final PsiVariable variable;
 


amaembo:
Why do you need this? Are you using these APIs from your own plugin?

iamKunal:
> Why do you need this? Are you using these APIs from your own plugin?

Yes

amaembo:
Note that it's part of 'impl' module, so you are using it at your own risk. We don't guarantee that these classes won't change.

iamKunal:
> Note that it's part of 'impl' module, so you are using it at your own risk. We don't guarantee that these classes won't change.

Yeah, I don't think that would be much of a problem. Although is there any other way to do the same?

iamKunal:
What I'm trying to do is:
From a `CodeBlock` get to all read/write instructions via the `ControlFlow`
from those read/write instructions I need both the usage (`ControlFlow.getElement(<offset>)`) and the actual variable (`ReadVariableInstruction/WriteVariableInstruction.variable`)

amaembo:
Your patch is merged, thanks:
https://github.com/JetBrains/intellij-community/commit/99c6375d04d96844fa4c636aa8a6031a0c20387e

PR #1526, opened 2021-03-14
Fixed IndexNotReadyException when supportsDumbModeFolding is false

NekoCaffeine:


JB-Dmitry:
This is not a good solution for whatever problem you're attempting to solve - the code in lambda is executed on EDT, and we're trying to offload folding regions building to a background thread here.
BTW, it would be good to have more information on the original problem - it's not quite clear from issue description.

NekoCaffeine:
Yes, I know this is not a good solution.
Merging this PR was not my intention, but it worked for me.

> and we're trying to offload folding regions building to a background thread here.

Glad to know that, if nothing else, I will close this pr.

Don't worry, I can fix this problem temporarily with a runtime bytecode transformer very simply.

PR #1530, opened 2021-03-23
add "Relative Path From Current File Directory" action to Copy Path actions

k-bespalov:
I had some times when I needed to get relative path from current file directory to some file in project
Hope, this feature will be useful
image


kradima:
@k-bespalov Thank you for the PR! This additional option may be useful for part of users but for most users, it can slightly overweight the popup. So I propose to enable it by an option in the advanced settings, that we're going to provide in the nearest releases.

k-bespalov:
> @k-bespalov Thank you for the PR! This additional option may be useful for part of users but for most users, it can slightly overweight the popup. So I propose to enable it by an option in the advanced settings, that we're going to provide in the nearest releases.
@kradima 
So should I wait for those releases and finalize my pr? Or what possibilities?

kradima:
> > @k-bespalov Thank you for the PR! This additional option may be useful for part of users but for most users, it can slightly overweight the popup. So I propose to enable it by an option in the advanced settings, that we're going to provide in the nearest releases.
> 
> @kradima
> So should I wait for those releases and finalize my pr? Or what possibilities?

I propose to wait for those releases (I believe it'll be in the nearest one), provide the option and finalize the pr.

k-bespalov:
> > > @k-bespalov Thank you for the PR! This additional option may be useful for part of users but for most users, it can slightly overweight the popup. So I propose to enable it by an option in the advanced settings, that we're going to provide in the nearest releases.
> > 
> > 
> > @kradima
> > So should I wait for those releases and finalize my pr? Or what possibilities?
> 
> I propose to wait for those releases (I believe it'll be in the nearest one), provide the option and finalize the pr.

@kradima let me know then, please

kradima:
Hi! Sorry for long response. We don't sure that it's a very popular option and would like to estimate interest in it. I created an [issue](https://youtrack.jetbrains.com/issue/IDEA-277876) to collect feedback and votes. As an alternative, create a [plugin](https://plugins.jetbrains.com/docs/intellij/getting-started.html) with the provided functionality.

Anyway, thank you for contributing!

PR #1532, opened 2021-03-29
False positive SameParameterValueInspection for vararg

nsuriadi:
[Issue](https://youtrack.jetbrains.com/issue/KT-31656)

I've made a couple of assumptions here (please correct me if I'm wrong). 
- It looks like we were avoiding `vararg` comparison. I've tried to keep the same behavior here by excluding `vararg` in positions other than the last parameter from the inspection.
- The usage processor doesn't seem to take receiver argument into account (Kotlin extension) and so I've decided omit receiver parameter from being compared against its corresponding argument. There doesn't to be a way to check if a param is a receiver through UAST, so I've opted to check its name.
- I can't seem to find a way to reliably check whether a param is a `vararg` unless its the last param. I saw this [Kotlin PSI](https://github.com/JetBrains/kotlin/blob/bbf5c4412e9264f0749cc48715af052eadda0c03/compiler/light-classes/src/org/jetbrains/kotlin/asJava/classes/ultraLightParameters.kt#L174) while debugging and was wondering whether making a change in the Kotlin plugin would be necessary for this

I've also added a test for the local inspection tool. Let me know if there are any improvements I can make here.


BartvHelvert:
Hi, thank you for your pull request. I am sorry that it took so long for us to process it.

The approach taken in this pull request is wrong, the UAST API tries to emulate the way the JVM sees the code.
The JVM can't understand `vararg` parameters that aren't positioned last in the parameter list. The way Kotlin deals with this is by passing the argument as an array.

For example:
`fun foo(vararg args: String, args1: String) { }`

Is JVM equivalent to the following Java code:
`void foo(String[] args, String args1) { }`

Problem with our current implementation is that we don't correctly retrieve the arguments that are passed into the parameter. This can fairly easily be fixed by changing `call.valueArguments.get(...)` to `call.getArgumentForParameter(...)`.

I [patched](https://github.com/JetBrains/intellij-community/commit/d34ad5345eab82d6210d5b909e39e1e031725ca2) the issue today, let me know if you have any comments.

PR #1533, opened 2021-03-30
Add Gradle version as parameter for resolvers

gavra0:
This allows resolvers to adjust Gradle init scripts
based on the version of Gradle that will run.


nskvortsov:
Please take a look at this extension: https://github.com/JetBrains/intellij-community/commit/b941254f8f6005c9281a7df7da3d2fc81cbfb441

I believe, this is what you are looking for.

gavra0:
Does this work for extensions that are only supposed to contribute to init scripts, like https://github.com/JetBrains/kotlin/blob/f1cdd0e2ed0985f433715e6a69394fd48a7e2907/idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/KotlinGradleCoroutineDebugProjectResolver.kt#L13?

There is no `ProjectResolverContext` set for this extension (from the `GradleTaskManger`, so I'm not sure how to get access to Gradle settings/version.

gavra0:
I've fixed `KotlinGradleCoroutineDebugProjectResolver` using another approach in https://github.com/JetBrains/kotlin/pull/4296, so this PR is not needed any more. Thanks, and sorry for the false alarm :)

PR #1534, opened 2021-04-01
Fix IDEA-247575 Parameter names aren't used in method body when debug info is absent

Maccimo:
Fix for [IDEA-247575](https://youtrack.jetbrains.com/issue/IDEA-247575) Parameter names aren't used in method body when debug info is absent


trespasserw:
Sorry for the late reply.

I'm going to implement the fix the other way, because it amends the one made earlier that turned out to be incomplete. Have one question though: in your code, the `MethodParameters` attribute prevails over debug info. Are there any particular reasons for that? I'd rather make the other way around, because debug info is more extensive.

Maccimo:
> Sorry for the late reply.
> 
> I'm going to implement the fix the other way, because it amends the one made earlier that turned out to be incomplete. Have one question though: in your code, the `MethodParameters` attribute prevails over debug info. Are there any particular reasons for that? I'd rather make the other way around, because debug info is more extensive.

Debug info is not required for program execution in general and therefore it is often stripped or trashed by ProGuard, obfuscators, etc.  
In contrast, `MethodParameters` is designed to be available in runtime via `java.lang.reflect.Parameter` and stripping or trashing them may lead to program failure.
So `MethodParameters` looks more autрoritative source for me.

trespasserw:
So, while `MethodParameters` might be more autoritative, I'd argue that debug information is used more often. Anyway, the current implementation is able to use both sources and seem to work reasonably well.

Thanks!

PR #1536, opened 2021-04-06
Fix IDEABKL-8006 IDE hangs when decompiling class which is its own superclass

Maccimo:
Fix for [IDEABKL-8006](https://youtrack.jetbrains.com/issue/IDEABKL-8006) IDE hangs when decompiling class which is its own superclass.


gorrus:
Merged into master, thanks for your contribution!

PR #1538, opened 2021-04-19
[bug] fix a bug about multiple declarations of a same variable in decompile output.

XenoAmess:
Just a bug report today.

Will fix it when I have time.


XenoAmess:
Fixed one of the bugs, and decided to split another bug into another pr.

After all they are two different bugs.

As I'm really new to the decompiler codes, I cannot say my changes are totally 100% correct, so please take special care.

Thanks.

XenoAmess:
any news?

XenoAmess:
any news?

trespasserw:
Sorry for the late reply. The decompiler is a rather tricky piece of code, which requires some time to dig in.

I've merged the fix, but replaced test code with examples from relevant YouTrack tickets, because yours produce spellchecker complaints.

PR #1540, opened 2021-04-20
[bug] show a bug about generic wildcard function return type in decompiler

XenoAmess:
I will fix it some day,sooner or later, but not today.
Just, have no enough time...I need sleep.


XenoAmess:
any news?

trespasserw:
Thanks, but at this point, I don't see much sense in adding a failng test. JetBrains uses [YouTrack](https://youtrack.jetbrains.com) for bug reporting; this particular one looks similar to [IDEA-166071](https://youtrack.jetbrains.com/issue/IDEA-166071).

XenoAmess:
> Thanks, but at this point, I don't see much sense in adding a failng test. JetBrains uses [YouTrack](https://youtrack.jetbrains.com) for bug reporting; this particular one looks similar to [IDEA-166071](https://youtrack.jetbrains.com/issue/IDEA-166071).

get it.

I'm a person who like to solve questions one by one, so as the last bug is fixed and checked last week, I will try if I can solve this one.

will close this pr for now.

PR #1542, opened 2021-04-23
unify gradle source.

XenoAmess:


trespasserw:
Thanks, but what for?

XenoAmess:
> Thanks, but what for?

@trespasserw so what is `cache-redirector.jetbrains.com` for? I thought it be created for a reason, and as it be created, it should be used.

Anyway, 2 reasons to do this:

1. to make things unify
2. to avoid downloading 2 copies of gradle-7.0 onto everybody's computer, as gradle-wrapper cache every gradle copy not by their verson, but by url. (md5 of url actually)

trespasserw:
It's actually a redirector that serves requests from internal network via a fast local cache, and forwards external ones to a specified service (`services.gradle.org`) in this case. Two copies are virtually inevitable, because non-JB projects are unlikely to use the redirector. To the best of my knowledge, the only reliable way to avoid unwanted downloads is executing local Gradle instead of wrapper scripts.

To summarize, I believe the change is not needed. Both scripts exist for external contributors who don't need to open the entire IntelliJ project to hack on the decompiler or the shell script support plugin, and these people won't benefit from the redirector.

XenoAmess:
> It's actually a redirector that serves requests from internal network via a fast local cache, and forwards external ones to a specified service (`services.gradle.org`) in this case. Two copies are virtually inevitable, because non-JB projects are unlikely to use the redirector. To the best of my knowledge, the only reliable way to avoid unwanted downloads is executing local Gradle instead of wrapper scripts.
> 
> To summarize, I believe the change is not needed. Both scripts exist for external contributors who don't need to open the entire IntelliJ project to hack on the decompiler or the shell script support plugin, and these people won't benefit from the redirector.

Sounds reasonable. 

Thanks for your explanations.

PR #1543, opened 2021-04-25
Hg Tags get from hg tags command instead of read from tags file

balagurusurendar:
Read tags from file works perfectly but it will not work if tags stored in another branch "tags_branch".

if repo maintains tags in another branch like "tags" then it will always produce empty in current implementation.
To overcome above issue by getting tags from hg command itself. it will handle every formats.

Because of above Implementation I need to change Unit test that involves in this file. 
existing unit test implementation involves by coping existing stored hg files but it leads to hg repo corruption. so i can not able to query using hg command in that repo. so i changed to new format by manually creating repo for all unit test use cases in that file.

### DIFF
--- a/plugins/hg4idea/testSrc/hg4idea/test/repo/HgRepositoryReaderTest.java
+++ b/plugins/hg4idea/testSrc/hg4idea/test/repo/HgRepositoryReaderTest.java
@@ -45,25 +45,6 @@ public class HgRepositoryReaderTest extends HgPlatformTest {
     super.setUp();
     myHgDir = new File(myRepository.getPath(), ".hg");
     assertTrue(myHgDir.exists());
-    //File pluginRoot = new File(PluginPathManager.getPluginHomePath("hg4idea"));
-
-    //String pathToHg = "testData/repo/dot_hg";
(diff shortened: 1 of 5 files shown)


klikh:

klikh:
Please also rebase the pull request on top of the current master, to avoid the merge commit. Thanks.

plugins/hg4idea/testSrc/hg4idea/test/repo/HgRepositoryReaderTest.java
```diff
@@ -110,33 +216,35 @@ public void testLocalTags() {
...
-    Collection<String> refs = new HashSet<>();
-    String[] refsWithHashes = FileUtil.loadFile(refFile).split("\n");
-    for (String str : refsWithHashes) {
-      String[] refAndName = str.trim().split(" ");
-      assertEquals(2, refAndName.length);
-      refs.add(refAndName[1]);
-    }
-    return refs;
-  }
+  //@NotNull
```
klikh, line 239: If the method is not needed anymore, please remove it. There is no need to comment it, since there is version control history.

balagurusurendar: removed

PR #1545, opened 2021-04-26
Add aarch64 support for macOS and jna-5.8.0

vavilovm:
Platform.RESOURCE_PREFIX for aarch64 macOS is "darwin-aarch64" since jna-5.8.0


trespasserw:
Thanks for the patch, but that code is going to be removed soon (in favor of NIO 2).

PR #1546, opened 2021-04-29
[KTIJ-11532] Create member function removes selected code in target Java file

marcin-bukowiecki:
Added selection removal before setting up target editor. 

I wanted to add few test cases but I noticed that all quickfix tests are `kotlin` to `kotlin`. When I try to mix `Create member function` from `kotlin` to `java` my tests are failing (I'm trying to add test data for ` QuickFixMultiModuleTestGenerated`). 

Issue: https://youtrack.jetbrains.com/issue/KTIJ-11532

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/createFromUsage/callableBuilder/CallableBuilder.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/createFromUsage/callableBuilder/CallableBuilder.kt
@@ -895,6 +895,7 @@ class CallableBuilder(val config: CallableBuilderConfiguration) {
 
             val descriptor = OpenFileDescriptor(project, targetClass.containingFile.virtualFile)
             val targetEditor = FileEditorManager.getInstance(project).openTextEditor(descriptor, true)!!
+            targetEditor.selectionModel.removeSelection()
 
             when (newJavaMember) {
                 is PsiMethod -> CreateFromUsageUtils.setupEditor(newJavaMember, targetEditor)
--- a/plugins/kotlin/idea/tests/testData/quickfix/createFromUsage/createFunction/fromKotlin/memberFunction.after.Dependency.java
+++ b/plugins/kotlin/idea/tests/testData/quickfix/createFromUsage/createFunction/fromKotlin/memberFunction.after.Dependency.java
@@ -1,0 +1,10 @@
+public class Foo {
+
+    public void test1() {
+
+    }
+
+    public void test2() {
+
+    }
+}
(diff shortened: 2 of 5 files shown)


abelkov:
@ybedrov please pay attention: Alexey doesn't work on Kotlin anymore. CC @vladimirdolzhenko

klunnii:
@abelkov I'll take a look

klunnii:
To make test data come into effect one needs to update `plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/quickfix/QuickFixMultiFileTestGenerated.java` file. It is a generated one. File can be updated by means of `Generate Kotlin Tests` IDEA run configuration. 

After I did this on my local machine (rebased your commits on recent master) I've got `org.jetbrains.kotlin.idea.quickfix.QuickFixMultiFileTestGenerated.CreateFromUsage.CreateFunction.FromKotlin#testMemberFunction`. So far so good. The problem is though that the test passes always successfully whether the fix is applied or not.

marcin-bukowiecki:
So the problem is with testing framework which ignores </selection> in target source file?

klunnii:
You're right. <selection> from `before.Dependency.java` doesn't affect the test. You can check it evaluating `targetEditor.selectionModel.hasSelection()` near the code snippet that you added.
Here is what goes wrong.
1. `EditorTestUtil#extractCaretAndSelectionMarkersImpl` requires that a file has <caret>. Though it's required, it's not enough.
2. Files are opened in editors opened at `CodeInsightTestFixtureImpl#createEditor`. Here the magic `FileEditorManager.getInstance(project)` means that we take implementaition declared in `platform/platform-resources/src/META-INF/PlatformLangComponents.xml`
```
    <component>
      <interface-class>com.intellij.openapi.fileEditor.FileEditorManager</interface-class>
      <implementation-class>com.intellij.openapi.fileEditor.impl.PsiAwareFileEditorManagerImpl</implementation-class>
      <!--suppress PluginXmlValidity -->
      <headless-implementation-class>com.intellij.openapi.fileEditor.impl.TestEditorManagerImpl</headless-implementation-class>
    </component>
```
I.e. `TestEditorManagerImpl` (tests have no UI hence headless). Opening an editor with this manager is supplied with `editor.getSelectionModel().removeSelection()`. Not sure why.
3. Ok, we know that opening an editor clears selection. But we calculate selection afterwords (1)...
4. Selection is cleared twice: first at (2) (<selection> and <caret> are handled)  and then at `CreateCallableFromUsageFixBase#invokeImpl`.

So what I suggest is to remove your test for now. Test infrastructure should be revisited and fixed.

klunnii:
I merged the first commit. Thank you very much!

PR #1547, opened 2021-05-04
[bug] update plantuml.jar version to avoid some bug by older version

XenoAmess:
The graph that cannot be baked correctly using current plantuml.jar we used in idea is:

```plantuml
@startuml
'https://plantuml.com/component-diagram


package "Some Group" {
  HTTP - [First Component]
  [Another Component]
}

node "Other Groups" {
  FTP - [Second Component]
  [First Component] --> FTP
}

cloud {
  [Example 1]
}


database "MySql" {
  folder "This is my folder" {
    [Folder 3]
  }
  frame "Foo" {
    [Frame 4]
  }
}


[Another Component] --> [Example 1]
[Example 1] --> [Folder 3]
[Folder 3] --> [Frame 4]

@enduml
```

And after replacing the plantuml.jar in local idea folder (then clean the caches) I just can bake it.

So I do think we should upgrade the version of that jar.


ybedrov:
@TanVD Could you check please?

XenoAmess:
any news?

PR #1549, opened 2021-05-06
Started tackling NotNull anotation issue

terrarier2111:
I know that this pr is probably incomplete but i just wanted to start tackling the issue with JetBrains annotations that if you try to annotate a primitive array with NotNull or Nullable it will tell you that "Primitive type members cannot be annotated" which is obviously incorrect because arrays of primitives aren't primitive themselves. This poses a major issue for a project i am working on and trying to annotate primitive array method parameters with NotNull but Intelij just tells me that it's a primitive type.
Edits and suggestions are welcome :)

### DIFF
--- a/java/java-impl/src/com/intellij/codeInspection/inferNullity/NullityInferrer.java
+++ b/java/java-impl/src/com/intellij/codeInspection/inferNullity/NullityInferrer.java
@@ -699,7 +699,7 @@ public class NullityInferrer {
       if (field instanceof PsiEnumConstant) {
         return;
       }
-      if (field.getType() instanceof PsiPrimitiveType ||
+      if ((field.getType() instanceof PsiPrimitiveType && !(field.getType() instanceof PsiArrayType)) ||
           isNotNull(field) || isNullable(field)) {
         return;
       }


akozlova:
Oh, I am sorry, I missed the PR. 

if `field.getType() instanceof PsiPrimitiveType` then `field.getType()` can't be of `PsiArrayType` so this change should not change the behaviour.

Actually, I guess, you have `@NotNull int [] field;` which means that you want an array of not null elements of type `int`. You want probably instead a not null array of int elements `int  @NotNull[] field;`

terrarier2111:
I didn't mean fields, i meant primitive array parameters to methods, but i think it may also be helpful for fields though.

terrarier2111:
Oh, i understand what you mean, but i think this: byte @NotNull[] bytes looks rather weird especially because for any other type the @NotNull annotation has to be put in front of the actual parameter type and arrays are the only exception where you have to put the annotation in the middle, but whatever, it works and that's my only concern, thank you for your help!

akozlova:
I somehow agree with you that when you first see type annotations with array types - they look weird but on the second thought it's normally start to be ok. Anyway, it's what java spec says, not what IDE can change

PR #1550, opened 2021-05-06
[IDEA-37227] Add new "Boolean expression may be factorized" inspection

Fabrice-TIERCELIN:
[Creates a new inspection that factorizes boolean expression when possible](https://youtrack.jetbrains.com/issue/IDEA-37227):

`isValid && isFound || isValid && isActive`

becomes:

`isValid && (isFound || isActive)`


Fabrice-TIERCELIN:
@BasLeijdekkers, should I create a new cleanup in `com.siyeh.ig` package or `com.intellij.codeInspection` package?

Fabrice-TIERCELIN:
@amaembo, should I rebase my Pull Request?

amaembo:
Issue: https://youtrack.jetbrains.com/issue/IDEA-37227

amaembo:
There's also a request to do the similar thing for integer multiplication
https://youtrack.jetbrains.com/issue/IDEA-262956

Fabrice-TIERCELIN:
Hi @akozlova,

It seems that the checks (Qodana / Code Inspection, etc...) for this PR do not run the tests. Is there a way to activate the test for the checks?

BasLeijdekkers:
Hi @Fabrice-TIERCELIN,
Do you think this needs to be an inspection? That an expression can be factorized does not seem like a problem, or is it? I think this would better as an intention. What do you think?

Fabrice-TIERCELIN:
I would like to code it for mass edit. I think an intention can't be applied on a project. But a local fix looks nice too. Can we code such a development for both usages?

BasLeijdekkers:
Alright, let's leave it as an inspection, but enable it by default at INFORMATION level. Do you think this inspection could also work for numeric expressions, like it does for boolean expressions? Also it would be great if you could squash the pull request into a single commit, this makes merging easier because we can't use GitHub's Merge pull request.

Fabrice-TIERCELIN:
This PR has been recreated here: #1576

PR #1551, opened 2021-05-06
Improve hotswap class detection when build is delegated to Gradle

gaiazov:
This PR is an improved version of https://github.com/JetBrains/intellij-community/pull/1386

### What is the problem
When IntelliJ is configured to delegate the build to Gradle, the hot swap functionality does not work as well when IntelliJ is used for compilation.

One issue is that sometimes, when gradle builds a project, the hot swap agent ends up deploying **all the classes in the project**. This is a problem for our repository because we have a project with 4000 classes, so hot swap is really slow.

### What is it caused by
I debugged the hot swap/gradle code. To me it looks like the root cause is that the `HotSwapUIImpl` class only looks at gradle output directories. When gradle builds a project, the project output paths is written to a temporary file (via a gradle init script). So, when 1 or 100 classes are changed, the input to `HotSwapUIImpl` remains the same - the `classes` folder.

The `HotSwapUIImpl` class is unable to distinguish what classes _actually_ changed so it ends up deploying _all_ the classes

### Solution proposed in this PR
Instead of gradle init script outputting the list of output folders, I changed it to output the list of actual modified classes. I created a new gradle init script. The init script hashes task outputs before and after task execution, and then compares the hashes to determine which files changed

- Add an experimental flag for improved gradle hotswap detection
- Minimal changes to `GradleProjectTaskRunner` class to use new gradle hotswap detection code
- Added tests to an new class, existing tests should pass as-is

### DIFF
--- a/platform/platform-resources/src/META-INF/PlatformExtensions.xml
+++ b/platform/platform-resources/src/META-INF/PlatformExtensions.xml
@@ -974,6 +974,10 @@
       <description>New project wizard</description>
     </experimentalFeature>
 
+    <experimentalFeature id="gradle.improved.hotswap.detection" percentOfUsers="0">
+      <description>Improved Hotswap Detection when building with Gradle</description>
+    </experimentalFeature>
+
     <rawEditorTypedHandler implementationClass="com.intellij.openapi.editor.impl.EditorFactoryImpl$MyRawTypedHandler"/>
 
     <postStartupActivity implementation="com.intellij.diagnostic.AnalyzePendingSnapshotActivity"/>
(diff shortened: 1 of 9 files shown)


gaiazov:
@vladsoroka I rewrote the original implementation - added tests, experimental flag and implementation of init script using the Gradle Build Service

vladsoroka:
The change looks good to me. I've merged it using the following commits https://github.com/JetBrains/intellij-community/commit/8924a43938b4776d37de7c2ac2eaaba0f817ba1d, https://github.com/JetBrains/intellij-community/commit/e6512eb829d2f715da402b718da5787feef61134.
I've also replaced usage of [deprecated](https://github.com/JetBrains/intellij-community/commit/36c7cfd24da0cb1393ad54345a8d1c5f41ab718d) `experimentalFeature` with the registry key using the same id `gradle.improved.hotswap.detection` and fixed some windows paths specific issues in tests, see the commit at https://github.com/JetBrains/intellij-community/commit/576551aafa95e988c4c59a67a7ee8bae00c449c5

Also I've found that dirty output paths VFS refresh has been omitted in `GradleImprovedHotswapDetection#processInitScriptOutput` (I mean `CompilerUtil.refreshOutputRoots(affectedRoots)`), will fix it too.

Thank you for the contribution Leonid!

PR #1554, opened 2021-05-18
[EE-IR] Introduce more backend targets for jvm-debugger suite

kandersen:
@nikita-nazarov Here's are the first, small changes. May as well
start small, for early feedback :)

This commit starts the work of refining the notion of target
backend for the jvm-debugger suite to include the backend used to
compile _fragments_ occurring in the tests, and updates the
existing suite and infrastructure accordingly (updated all uses
of `IGNORE_BACKEND`).

The plan from here is as follows:

  1) adding another 2 targets (`_WITH_IR_BACKEND`) and stub the new fragment
     compiler backend, then mute all 250 tests on those backends.
  
  2) Start upstreaming my new fragment compiler proposal in pieces and as work
     progresses.
  
  3) There are changes to PSI2IR that lives in the kotlinc repo. We should
     co-ordinate where, how and when these land.
  
  4) Decide on how to expose the choice of fragment compiler to the end user,
     for experimental use.

Small nits:

  - removed unused variable
  - cleaned up some imports.


kandersen:
Closing in favour of a [jetbrains/kotlin PR](https://github.com/JetBrains/kotlin/pull/4425) after coordination with Mikhail Zarechenskii.

PR #1555, opened 2021-05-19
Add `hbs` as alias for `handlebars` in markdown injections

vitch:
I _think_ this will allow automatic language injection for handlebars within fenced code blocks landed by `hbs`.

I wasn't sure if there was any way to do this via configuration or if it has to be built into the plugin?


vitch:
Closing because it hasn't been looked at in nearly 4 years and I'm not even using WebStorm any more!

PR #1557, opened 2021-05-24
Allow reading zip files with more than 65,535 entries

dahlstrom-g:


dahlstrom-g:
@develar?

develar:
Already applied.

PR #1560, opened 2021-05-26
Initial support for CONSTANT_Dynamic constant pool entry type.

Maccimo:
This PR address the following issues:
 - Avoid decompiler crashing on class files with CONSTANT_Dynamic constant pool entries. Support for complete dynamic constants decompilation need further work as a separate task.
 - Fail-fast with meaningful error message on class files with unsupported constant pool entries.


trespasserw:
Sorry for the late reply.

"Fail-fast" principle isn't particularly suitable for a decompiler - on the contrary, it should strive to decompile as much as it can and leave some stubs in places where it stumbles.

Re `CONSTANT_Dynamic` support - well, I'm not sure. I would wait until compilers start using it.

What do you think?

Maccimo:
Here we have no other option but "fail-fast". When decompiler encounter constant pool entry of type they don't aware they can't go further because don't know this particular constant pool entry size and how much bytes to skip. Since class file has no information about constant pool size in bytes we can't just skip and start parsing rest of class file.

I believe decompiler should support all constant pool entries types supported by JVM. Even if not used by `javac` yet it can be used in user-/tooling-generated code such as in https://github.com/jacoco/jacoco/pull/845
Also it seems ConDy is planned to be used in Valhalla: http://cr.openjdk.java.net/~jrose/values/parametric-vm.pdf

Probably will comment more later this day.

trespasserw:
Oh. I thought that pool entries contain the size field, like attributes. Okay, I will give it a second look.

trespasserw:
Merged, thanks!

PR #1563, opened 2021-05-30
feat: createImportStatement with assigned classname

xaghoul:


akozlova:
Where do you need this api?

xaghoul:
In my case, I needed to create Import Statement by passing Autowired (spec.: org.springframework.beans.factory.annotation.Autowired). But I only have creating wild card import method. So I came to the conclusion that's it's not a bad idea to have method in pr.

akozlova:
I believe you should not create create imports statements manually and rather rely on the platform. Could you please share the code which requires creation of imports? E.g. `ref.bind(klass)` would insert required import when it's not provided yet, etc

xaghoul:
Can you please clarify, wdym by relying creating import statements on the platform?
I wanted to do something like that: 
`String autowiredPath = "org.springframework.beans.factory.annotation.Autowired";
PsiImportStatement autowiredImport = PsiElementFactory.createImportStatement(autowiredPath);`
If you're wondering, I found It out by using [this method](https://github.com/JetBrains/intellij-community/blob/85cca9ef02afc8677120871508bb9a7bac6871b9/java/java-psi-api/src/com/intellij/psi/JavaPsiFacade.java#L29); 
and then passing it with

akozlova:
what would you do with this import later? if you add an import and would add it to the import list, you need to ensure that you add it in the correct place and that you replace multiple imports under on demand import or expand on demand import if you decide to use it but a user has another setting. It's quite complicated thing and in most clients you just don't need to bother about imports, the platform will add them for you automatically. Thus I ask you what issue you fix so you want to add import manually

xaghoul:
It really was added for the possibility add it to the import list. But I don't get If we cannot ensure in the safety of this import statement adding, how this [method](https://github.com/JetBrains/intellij-community/blob/1e45ce3a2f71ab2e842e76cb68372fce7531d0e7/java/java-psi-impl/src/com/intellij/psi/impl/PsiElementFactoryImpl.java#L521) is safe? 
Sorry for bothering you in this way, I didn't actually find it in documentation

akozlova:
The call is safe. It leaves the file in the state where unused import inspection/optimize imports would have something to change when incorporated mechanism to deal with imports already should process everything for you. Thus I don't think that such api should be used at all. Could you please share the code how you create your annotation? I believe, you probably need to call `com.intellij.psi.codeStyle.JavaCodeStyleManager.shortenClassReferences(com.intellij.psi.PsiElement)` and nothing more.

xaghoul:
Thank you, I've got your point. Talking about adding annotations, I believe, there is more efficient way to do it, but that's how i figured it out: 
```
    public void addQualifiedAnnotationName(String qualifiedAnnotationName, PsiMember psiMember) {
        WriteCommandAction.runWriteCommandAction(psiMember.getProject(), (Computable<PsiAnnotation>) () ->
                Objects.requireNonNull(psiMember.getModifierList()).addAnnotation(qualifiedAnnotationName));
    }
```
I know that WriteCommandAction is a bad way to upgrade PsiMembers, but that was a University project and you know when you have to make it work and deadlines are coming :)

So, should I close this PR or it does not matter?

akozlova:
WriteCommandAction is ok, you just need to provide additional parameters, like command name so undo is nice. I would also do smith like `JavaCodeStyleManager.shortenClassReferences(Objects.requireNonNull(psiMember.getModifierList()).addAnnotation(qualifiedAnnotationName))` to get all required changes.
I am closing the PR then. Thanks for the fast update and good luck with your projects!

PR #1567, opened 2021-06-02
Optimization: Eliminate useless method call

Maccimo:
Eliminate call to `ControlFlowGraph::inlineJsr()` for class files that does not support `jsr`/`jsr_w` and `ret` instructions.

According to the JVMS11 &sect;4.9.1:
 - If the class file version number is 51.0 or above, then neither the `jsr` opcode or the `jsr_w` opcode may appear in the code array

Since `jsr` instruction is forbidden for class files version 51.0 (Java 7) or above call to `inlineJsr()` is only meaningful for class files prior to the Java 7.


trespasserw:
Merged, thanks!

PR #1570, opened 2021-06-07
KTIJ-14267 False positive intention "Expand boolean expression to 'if else'"

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-14267

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ExpandBooleanExpressionIntention.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ExpandBooleanExpressionIntention.kt
@@ -4,6 +4,7 @@ package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.psi.*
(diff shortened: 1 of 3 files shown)


nikitabobko:
You don't need to traverse all parents, you need to check only one parent.

Merged with additional patch:
```
diff --git a/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ExpandBooleanExpressionIntention.kt b/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ExpandBooleanExpressionIntention.kt
index 8325f992c9b6..c2aae9b272b5 100644
--- a/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ExpandBooleanExpressionIntention.kt
+++ b/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ExpandBooleanExpressionIntention.kt
@@ -8,7 +8,6 @@ import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.psi.*
-import org.jetbrains.kotlin.psi.psiUtil.parents
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 import org.jetbrains.kotlin.psi2ir.deparenthesize
 import org.jetbrains.kotlin.resolve.BindingContext
@@ -20,9 +19,7 @@ class ExpandBooleanExpressionIntention : SelfTargetingRangeIntention<KtExpressio
     KotlinBundle.lazyMessage("expand.boolean.expression.to.if.else")
 ) {
     override fun applicabilityRange(element: KtExpression): TextRange? {
-        if (!element.isTargetExpression()) return null
-        val target = element.parents.takeWhile { it.isTargetExpression() }.lastOrNull() ?: element
-        if (element != target) return null
+        if (!element.isTargetExpression() || element.parent.isTargetExpression()) return null
         if (element.deparenthesize() is KtConstantExpression) return null
         val parent = element.parent
         if (parent is KtValueArgument || parent is KtParameter || parent is KtStringTemplateEntry) return null
```
the commit will show up in the `master` in a few hours after it passes automatic checks

PR #1571, opened 2021-06-07
Fix spell check warnings in hexadecimal colour strings

samofcorinth:
Previously, strings such as "#33FFFF" would trigger typo warnings, since the regular expression used to split the word would detect "FFFF" as a word; more generally, unless the "#" were to be followed by a letter, the hex string would be split between the last number and the first letter, triggering a spell check on the letters. This pull request corrects the regular expressions used to split words.

I've updated one of the test files to demonstrate this.


donnerpeter:
Sorry for the delay! This scenario is already supported in the upcoming 2025.3 IDEs

PR #1574, opened 2021-06-08
KTIJ-11575 False positive inspection "Range could be replaced with '.indices' call" on map

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-11575

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceManualRangeWithIndicesCallsInspection.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceManualRangeWithIndicesCallsInspection.kt
@@ -8,13 +8,18 @@ import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
 import com.intellij.psi.search.searches.ReferencesSearch
+import org.jetbrains.kotlin.builtins.DefaultBuiltIns
 import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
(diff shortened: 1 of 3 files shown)


nikitabobko:
Merged with additional patch:
```
diff --git a/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceManualRangeWithIndicesCallsInspection.kt b/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceManualRangeWithIndicesCallsInspection.kt
index fbed3c6fc2c0..058c2b1ac625 100644
--- a/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceManualRangeWithIndicesCallsInspection.kt
+++ b/community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceManualRangeWithIndicesCallsInspection.kt
@@ -49,11 +49,12 @@ class ReplaceManualRangeWithIndicesCallsInspection : AbstractKotlinInspection()
             "rangeTo" -> right.receiverIfIsSizeOrLengthMinusOneCall()
             else -> null
         } as? KtSimpleNameExpression ?: return
-        if (collection.isMap()) return
+        if (!collection.doesHaveIndicesExtensionFunction()) return
         visitIndicesRange(holder, expression, collection)
     }
 
-    private fun KtExpression.isMap() = getType(analyze(BodyResolveMode.PARTIAL))?.isMap(DefaultBuiltIns.Instance) == true
+    private fun KtExpression.doesHaveIndicesExtensionFunction() =
+        getType(analyze(BodyResolveMode.PARTIAL))?.isMap(DefaultBuiltIns.Instance) == false
 
     private fun visitIndicesRange(holder: ProblemsHolder, range: KtExpression, collection: KtSimpleNameExpression) {
         val parent = range.parent.parent
```
the commit will show up in `master` in a few hours after our automatic checks

PR #1575, opened 2021-06-09
KTIJ-15021 Support concatenated strings for intention Convert to 'buildString'

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-15021

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertConcatenationToBuildStringIntention.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertConcatenationToBuildStringIntention.kt
@@ -3,14 +3,24 @@
 package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.idea.core.ShortenReferences
 import org.jetbrains.kotlin.idea.core.replaced
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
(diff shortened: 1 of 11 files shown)


plugins/kotlin/idea/tests/testData/intentions/convertConcatenationToBuildString/basic.kt.after
```diff
@@ -0,0 +1,11 @@
+// WITH_RUNTIME
+fun test(foo: String, bar: Int, baz: Int) {
+    val s = buildString {
+        append("foo=")
+        append(foo)
+        append(",bar=")
+        append(bar)
+        append(",baz=")
+        append(baz + baz)
```
nikitabobko, line 9: I don't like this behavior. Interpolation shouldn't have been converted. I believe that this would be much more readable:
```kotlin
fun test(foo: String, bar: Int, baz: Int) {
    val s = buildString {
        append("foo=$foo")
        append(",bar=$bar")
        append(",baz=${baz + baz}")
    }
}
```

nikitabobko (changes requested):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertConcatenationToBuildStringIntention.kt
```diff
@@ -0,0 +1,54 @@
...
+
+package org.jetbrains.kotlin.idea.intentions
+
+import com.intellij.openapi.editor.Editor
+import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.core.ShortenReferences
+import org.jetbrains.kotlin.idea.core.replaced
+import org.jetbrains.kotlin.psi.*
+
+class ConvertConcatenationToBuildStringIntention : ConvertToStringTemplateIntention() {
```
nikitabobko, line 11: Please, don't inherit from `ConvertToStringTemplateIntention`. `ConvertToStringTemplateIntention` is absolutelly different intention which is not related to `ConvertConcatenationToBuildStringIntention`. Either choose more common ancestor for inheritance or extract common code from `ConvertConcatenationToBuildStringIntention`

nikitabobko (changes requested):

PR #1580, opened 2021-06-14
Cast return of the dlopen call in case debug symbols are missing.

dylanbrodiefafard:
To fix the following error with gdb 8.3.1 when trying to attach to a Python process.

> 'dlopen@GLIBC_2.2.5' has unknown return type; cast the call to its declared return type
> No symbol "DoAttach" in current context.

`void *dlopen(const char *filename, int flags);` [Source](https://man7.org/linux/man-pages/man3/dlopen.3.html )

[See also for more details.](https://github.com/fabioz/PyDev.Debugger/pull/151)


avli:
Hello @dylanbrodiefafard,

Thank you for the PR – looks good to me 👍

You should submit the contributor license agreement [here](https://www.jetbrains.com/agreements/cla/) before the changes will be merged.

Thank you!

dylanbrodiefafard:
Done.

Thanks!

PR #1588, opened 2021-06-21
Google Maps Location icon for Git HEAD

Fabrice-TIERCELIN:
I suggest to replace the yellow label icon for Git HEAD by the [Google Maps Location icon](https://raw.githubusercontent.com/Fabrice-TIERCELIN/intellij-community/location_icon/platform/dvcs-impl/resources/icons/currentBranchFavoriteLocation.svg) because:

- The HEAD is not a tag
- The HEAD is where you are
- This icon is commonly used to be the symbol of a location, on GitHub or StackOverflow for example.

@develar, I can design the icon with a sharp, round or right-angled bottom.

- The sharp bottom is more meaningful
- The round bottom is cuter
- The right-angled bottom is a real flat design

I don't think there is any law issue.


dmitriysmirnovjb:
Thank you for your contribution. 

Could you please clarify what's the issue with the current yellow label? 

> The HEAD is not a tag

Tags, branches, and HEAD are all references, and thus we maintain the same icons of different colors to refer to them. It is actually quite common among other tools to depict HEAD with a tag sign.

Location icon has a strong association with maps and I am not sure it is the best option for git log.

> I don't think there is any law issue.

Actually, there is. Google icons are free to use only when they refer to the Google services. See the [guidelines](https://about.google/brand-resource-center/brand-elements/#product-icons)

Fabrice-TIERCELIN:
I think the law issues may be avoid by using a design different enough.

Perhaps the labels are common representations for the HEAD. I thought it was easier to understand if it is represented as a location. The git graph is like a map and the HEAD is like the _"You are here"_.

Do as you want.

dmitriysmirnovjb:
Thanks again.

We decided not to change the label, at least for now.

Fabrice-TIERCELIN:
@akozlova, and what if the [icon remains yellow](https://github.com/Fabrice-TIERCELIN/intellij-community/blob/bae9e37c7a7a2e246de6d231897a6acb37b8e9e2/platform/dvcs-impl/resources/icons/currentBranchFavoriteLocation.svg)?

PR #1592, opened 2021-06-23
[IDEA-290250] Improve the order of the editor popup menu entries

Fabrice-TIERCELIN:
Hi,

Among other entries, we have:

----
**Find Usages**
**Refactor**
----
Folding
Analyze
----
**Go to**
**Generate...**
----

_Find Usages_ and _Go to_ are both navigation features. _Refactor_ and _Generate..._ are both automatic code writing features. They should be gathered. I suggest to switch _Refactor_ and _Go to_.

@yole, I have successfully tested.

### DIFF
--- a/platform/platform-resources/src/idea/LangActions.xml
+++ b/platform/platform-resources/src/idea/LangActions.xml
@@ -337,7 +337,7 @@
               class="com.intellij.lang.documentation.ide.actions.DocumentationForwardAction"/>
       <action id="Documentation.EditSource" icon="AllIcons.Actions.EditSource" use-shortcut-of="EditSource"
               class="com.intellij.lang.documentation.ide.actions.DocumentationEditSourceAction"/>
-      
+
       <add-to-group group-id="Other.KeymapGroup"/>
     </group>
     <action id="Documentation.ViewExternal" icon="AllIcons.General.Web" use-shortcut-of="ExternalJavaDoc"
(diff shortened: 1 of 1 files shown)


Fabrice-TIERCELIN:
@ybedrov, perhaps it should be reassigned. The PR hasn't changed for 10 months.

akozlova:
The change looks logical but there are 2 considerations:
1. muscle memory
2. probably Go To actions are used by shortcuts more than refactorings, so the change would result in more "scrolling" for them.
What do you think, does it really worth it? Thanks

Fabrice-TIERCELIN:
I can invert the order if you want:

----
**Generate...**
**Refactor**
----
Folding
Analyze
----
**Find Usages**
**Go to**
----

_Muscle memory_ is important, but it's better to consider the future than the past.

akozlova:
Doesn't solve the second issue, that goto is most probably invoked with the mouse or shortcut.

Fabrice-TIERCELIN:
@akozlova, to move the debate forward, I suggest to move only one entry, let's say _Go to_:

----
**Find Usages**
**Go to**
**Refactor**
----
Folding
Analyze
----
**Generate...**
----

yole:
Thanks for your contribution! We implemented part of these changes internally; I hope the result works well for you.

Fabrice-TIERCELIN:
I had the opportunity to test it successfully on IntelliJ 2021.1.3

I can't test on a more recent installation because we can't hack the files of the installation on the further versions.

Fabrice-TIERCELIN:
I have successfully verified this feature with this configuration:
IntelliJ IDEA 2023.1 EAP (Community Edition)
Build #IC-231.4840.387, built on January 19, 2023
Runtime version: 17.0.5+8-b759.1 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Windows 10 10.0
GC: G1 Young Generation, G1 Old Generation
Memory: 1010M
Cores: 4
Non-Bundled Plugins:
    com.dubreuia (2.3.0)

Kotlin: 231-1.7.21-release-272-IJ4840.387

PR #1593, opened 2021-06-23
[kotlin] Fix visible variables in the debugger on dex VMs

sfs:
For inline functions, the Kotlin JVM debugger needs to reconstruct the currently visible variables out of metadata which the Kotlin compiler encodes in both the order and names of local variables.

On a dex VM, the order of local variables is not reliable, since locals are kept in registers which may have to be spilled. Spilling resets the start offsets of locals and changes their sorting order. This commit introduces a workaround to recover the actual variable introduction order when running on a dex VM and also makes the code for reconstructing inline call stacks more robust.

---

In particular, the previous code used only the inline depth to determine which variables are visible in the current scope. This is insufficient when there are multiple active stack frames at the same depth:
```kotlin
fun f() {
    val x = 0
    g {
        h(2)
    }
}

inline fun g(block: () -> Unit) {
    var y = 1
    block()
}

inline fun h(a: Int) {
    var z = 3
    /* breakpoint */
}
```
in `h`, both `y` and `z` are at depth `1`, but only `z` is supposed to be visible. The new code handles such examples correctly, but is not yet used consistently throughout the debugger.


sfs:
- Fixed a problem with some `this` variables being displayed twice
- Added a test for the example in the PR

yanex:
cc @nikita-nazarov

yanex:
The commit is pushed manually (8bb07102c5814016144ba4f617dbe9747f558a76). Thank you for your contribution!

PR #1594, opened 2021-06-24
update: proofread Kotlin inspections

p7nov:
Lines 104-134 from the spreadsheet


plugins/kotlin/idea/resources-en/inspectionDescriptions/RedundantRequireNotNullCall.html
```diff
@@ -1,6 +1,6 @@
 <html>
 <body>
-<p>Reports redundant <code>requireNotNull</code> or <code>checkNotNull</code> call on not-nullable expressions</p>
+<p>Reports <code>requireNotNull</code> or <code>checkNotNull</code> call on non-nullable expressions.</p>
```
nikitabobko, line 3: Do we really want to remove "redundant" word here? For it makes sense

plugins/kotlin/idea/resources-en/inspectionDescriptions/RedundantNotNullExtensionReceiverOfInline.html
```diff
@@ -1,10 +1,10 @@
...
 <body>
-<p>Reports inline functions with not nullable extension receivers which does not use the fact that extension receiver is not nullable.</p>
+<p>Reports inline functions with non-nullable extension receivers which don't use the fact that extension receiver is not nullable.</p>
 <p>
-    In Kotlin 1.0 or 1.1, calls of <code>inline fun</code> with flexible nullable extension (platform type with unknown nullability)
-    receiver did not include nullability check in bytecode, but in Kotlin 1.2 this problem is fixed
-    (see <a href="https://youtrack.jetbrains.com/issue/KT-12899">KT-12899</a>), nullability check is included into bytecode.
+    Before Kotlin 1.2, calls of <code>inline fun</code> with flexible nullable extension receiver (a platform type with an unknown 
+    nullability) did not include nullability checks in bytecode. In Kotlin 1.2 this problem was fixed
+    (see <a href="https://youtrack.jetbrains.com/issue/KT-12899">KT-12899</a>), including nullability checks into the bytecode.
```
nikitabobko, line 7: Also this paragraph is copy-pasted in two places: here and `PlatformExtensionReceiverOfInline.html`. I will update second usage

nikitabobko (changes requested):

plugins/kotlin/idea/resources-en/inspectionDescriptions/RedundantRequireNotNullCall.html
p7nov: My reason was that **all** of such checks are redundant on non-nullable expressions, so the specification isn't necessary. In UI, the shorter the better.
Although if this isn't tru, let's return `redundant`.

PR #1600, opened 2021-06-25
update: proofread kotlin intentions 109 118

p7nov:
Lines 109-118 from the spreadsheet.


zarechenskiy:
Thanks! I squashed your commits, so the resulting commit will appear in master soon

PR #1601, opened 2021-06-25
KTIJ-18901 "Create parameter" quick-fix by default doesn't add `val` keyword for missing parameter of annotation class, producing uncompilable code

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-18901


vladimirdolzhenko:
Thank you so much for your contribution.

Merged as https://github.com/JetBrains/intellij-community/commit/120dc3c1c90373d0f678d532defa0797486c4438

PR #1602, opened 2021-06-25
KTIJ-18925 False positive 'Variable should be inlined' for local delegated property

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-18925


vladimirdolzhenko:
Thank you,
it is merged as https://github.com/JetBrains/intellij-community/commit/de7e0a4b46dd712b4aade710fcbcb1c5cb5f65a9

PR #1603, opened 2021-06-27
KTIJ-12732 "Convert to sealed class" for enum doesn't handle use of `values()` method

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-12732


nikita-nazarov:
Thanks! Merged ca6497dc75c55ca565ec12c73b339b66e555b0c9

PR #1605, opened 2021-06-27
KTIJ-14780 assertNotNull(deferred) gets "Deferred result is never used" inspection

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-14780


nikita-nazarov:
Hi! In the current implementation, I'm worried about one case. Suggest you have a function:
```
fun fooAsync(param: Deferred<Int>) =
    GlobalScope.async {
        longCalculationWithSuspension() + param.await()
    }
```
Generally, it can be any function that takes one deferred parameter and returns *another*.
Then the inspection won't report that the return value is not used in this case.
```
val d  = GlobalScope.async { 42 }
fooAsync(d)
println(d.await())
```
Which is kind of not what we expect. I think it would be better to stick with
```
val fqNamesThatShouldNotBeReported = listOf("kotlin.test.assertNotNull", "kotlin.requireNotNull")
```
We can add more functions later if we need.

t-kameyama:
Fixed

nikita-nazarov:
Thanks for your contribution! Merged f84d4a6a543b0f8560ce929e70877abf59db75aa

PR #1606, opened 2021-06-27
KTIJ-17947 False positive "Use Property Access Syntax" after adding function to excluded methods in inspection's settings

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-17947


dimonchik0036 (approved):

PR #1608, opened 2021-06-28
Skip searching for symbols when starting up gdb and instead load just the required ones…

dylanbrodiefafard:
It's possible in a large enough project that searching for debug symbols throughout all the included libs will take longer than PyCharm is willing to wait for the attach code to execute. Since gdb is only being used to connect to and run a Python script in the target Python process, then we can skip this step and have the injected code execute much faster.

I observed the gdb process take over 23 seconds to execute the attach script in the target process while PyCharm shut down the debugging server after 20 seconds.


avli:
Well done! I'm ready to merge this change after submitting the contributor license agreement (see #1580) and some additional testing (https://youtrack.jetbrains.com/issue/PY-49617).

PR #1610, opened 2021-06-29
KTIJ-18387 "Redundant 'inner' modifier" inspection should warn when access containing class members of not this object

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-18387
https://youtrack.jetbrains.com/issue/KTIJ-18385


nikita-nazarov:
Thanks for the contribution! Merged e289a5690f6a0d8d2e1a009982576eca88c150d7

PR #1612, opened 2021-06-29
KTIJ-13716 Quick fix could suggest to change fully qualified name with existing import alias

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-13716


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithImportAliasInspection.kt
```diff
@@ -0,0 +1,47 @@
...
+            ReplaceWithImportAliasFix(aliasName)
+        )
+    })
+
+    private class ReplaceWithImportAliasFix(private val aliasName: String): LocalQuickFix {
+        override fun getName() = KotlinBundle.message("replace.with.0", aliasName)
+        override fun getFamilyName() = name
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            val expression = descriptor.psiElement as? KtNameReferenceExpression ?: return
+            val aliasNameIdentifier = expression.aliasNameIdentifier() ?: return
```
dimonchik0036, line 36: You should pass `aliasNameIdentifier` as a parameter (wrapped in `SmartPsiElementPointer` or use `LocalQuickFixOnPsiElement` instead of `LocalQuickFix` to avoid PSI leak) to avoid resolve on EDT

plugins/kotlin/idea/resources-en/inspectionDescriptions/ReplaceWithImportAlias.html
```diff
@@ -0,0 +1,19 @@
+<html>
+<body>
+Reports fully qualified name can be replaced with existing import alias.
```
dimonchik0036, line 3:
```suggestion
Reports fully qualified names that can be replaced with an existing import alias.
```

dimonchik0036 (changes requested):

dimonchik0036 (approved):

PR #1616, opened 2021-06-29
Fix init script generation

yuriy-solodkyy:
This seems to be another reason for the same problem.


yuriy-solodkyy:
@vladsoroka @nskvortsov

nskvortsov:
Good catch, thank you! will apply asap

nskvortsov:
Applied, will be backported to 212 branch

PR #1617, opened 2021-06-29
Treat JUnit4 runners with special annotation as parameterized

nymanjens:
## Context

When filtering individual JUnit4 tests in IntelliJ, the runner will look for a method with the name of the test. For parameterized tests, there is no 1:1 relationship between test name and test method and so this fails.

For `junit.Parameterized` and test methods with parameters, there is an existing workaround for this in `JUnit4TestRunnerUtil`.

## Problem when using TestParameterInjector

[TestParameterInjector](https://github.com/google/TestParameterInjector) is a runner developed at Google, and recently launched as open source project. In the last 2 months, [we had 3 different people asking us to fix this filtering issue in IntelliJ](https://github.com/google/TestParameterInjector/issues/6). The issue arises when fields are used for parameterization:

```java
public class MyTest {

  @TestParameter private boolean isOwner;

  @Test public void test() {...}
}
```

This will result in the following test names:

```
MyTest#test[isOwner=true]
MyTest#test[isOwner=false]
```

which will break the IntelliJ filtering.

## Proposed solution

With this pull request, the TestParameterInjector class would be changed as follows:

```java
package com.google.testing.junit.testparameterinjector

@ParameterizedRunnerIntroducingTestNamesWithSquareBrackets
static class TestParameterInjector extends Runner { ... }

@Retention(RUNTIME)
@interface ParameterizedRunnerIntroducingTestNamesWithSquareBrackets  {}
```

Other parameterized runners facing the same issue can define their own annotation with the same name and fix the same issue that way.

### DIFF
--- a/plugins/junit_rt/src/com/intellij/junit4/JUnit4TestRunnerUtil.java
+++ b/plugins/junit_rt/src/com/intellij/junit4/JUnit4TestRunnerUtil.java
@@ -224,9 +224,17 @@ public final class JUnit4TestRunnerUtil {
 
   private static boolean isParameterized(final String methodName,
                                          final Class<?> clazz) {
-    final RunWith clazzAnnotation = clazz.getAnnotation(RunWith.class);
-    if (clazzAnnotation != null && Parameterized.class.isAssignableFrom(clazzAnnotation.value())) {
+    final RunWith runWithAnnotation = clazz.getAnnotation(RunWith.class);
+    if (runWithAnnotation != null && Parameterized.class.isAssignableFrom(runWithAnnotation.value())) {
       return true;
+    }
+    if (runWithAnnotation != null) {
+      for (Annotation runnerAnnotation : runWithAnnotation.value().getDeclaredAnnotations()) {
+        if (runnerAnnotation.annotationType().getSimpleName()
+            .equals("ParameterizedRunnerIntroducingTestNamesWithSquareBrackets")) {
+          return true;
+        }
+      }
     }
     if (methodName != null) {
       try {


akozlova:
I'd like to get feedback on the existing fix first as now part of the tests are supported. 
BTW Is this annotation already included in any released version of any test framework? Supporting ghost annotations normally strikes back

nymanjens:
> I'd like to get feedback on the existing fix first as now part of the tests are supported.

If half of the use cases won't work, isn't the feedback obvious?

> BTW Is this annotation already included in any released version of any test framework? Supporting ghost annotations normally strikes back

I promise I'll add it to https://github.com/google/TestParameterInjector immediately after this PR is merged.

nymanjens:
Anna, it looks like we're set to have a long discussion about this again. We're really keen on fixing this problem though.

Would you be willing to have a discussion over videochat instead of this thread? My email address is jnyman@google.com.

akozlova:
I don't like the custom annotation as it would be exactly the same as to have dependency on one exact test runner, I'll try to come up with something else. Are you ok to change the parameters format or is it pluggable by user code (sorry, I saw too many frameworks at once), just in case?

nymanjens:
> Are you ok to change the parameters format or is it pluggable by user code

What do you mean by this? Could you give an example?

akozlova:
parameter string to look like `[fieldName=value]`; can be too long of cause

nymanjens:
To the question of whether it's feasible to change the parameters format to `[fieldName=value]`, that's unfortunately not possible because there is a hard limit on the total length of a test name (due to internal Google reasons).

PR #1618, opened 2021-07-01
ExpressionMayBeFactorized inspection: Handle numerics and eager operators

Fabrice-TIERCELIN:
Hi @BasLeijdekkers,

I have added some features you have asked for:
- Apply the rule on `&` and `|`
- Enable numbers

So I have renamed _BooleanExpressionMayBeFactorized_ to _ExpressionMayBeFactorized_. I have also merged the visit of the code because it is done twice. The `ExpressionMayBeFactorizedFix` calls the `ExpressionMayBeFactorizedVisitor` for the visit and then uses a reference on the `ExpressionMayBeFactorizedFix` to do the fix. For some cleanups I plan to integrate like [this one](https://www.eclipse.org/eclipse/news/4.18/jdt.php#hash), this part will become so huge it needs to be merged.


Fabrice-TIERCELIN:
For technical reasons, I have rebased the PR in another PR: #1628

PR #1620, opened 2021-07-06
KTIJ-11519 Call chain on collection type may be simplified: mapNotNull{}.first() -> firstNotNullOf{}

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-11519


dimonchik0036 (approved):
Thanks! Will merged manually.

PR #1624, opened 2021-07-08
KTIJ-18878 Folding: functions with expression body and comment above the function are wrongly folded

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-18878


dimonchik0036 (approved):

PR #1628, opened 2021-07-14
ExpressionMayBeFactorized inspection: Handle numerics and eager operators

Fabrice-TIERCELIN:
Hi @BasLeijdekkers,

I have added some features you have asked for:
- Apply the rule on `&` and `|`
- Enable numbers

So I have renamed _BooleanExpressionMayBeFactorized_ to _ExpressionMayBeFactorized_. I have also merged the visit of the code because it is done twice. The `ExpressionMayBeFactorizedFix` calls the `ExpressionMayBeFactorizedVisitor` for the visit and then uses a reference on the `ExpressionMayBeFactorizedFix` to do the fix. For some cleanups I plan to integrate like [this one](https://www.eclipse.org/eclipse/news/4.18/jdt.php#hash), this part will become so huge it needs to be merged.


Fabrice-TIERCELIN:
@BasLeijdekkers, I think it would be better if we can add this feature before the original one is released. Otherwise, the users will have to discover _Boolean expression may be factorized_ and then _Expression may be factorized_. It's confusing.

BasLeijdekkers:
This change has now been merged into the project here: 58f5dd08c451ea399523af98fe0eeb5f337af4ec

PR #1632, opened 2021-07-14
KTIJ-16001 False positive "Convert to primary constructor" with assignment to global variable

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-16001


dimonchik0036 (approved):

PR #1635, opened 2021-07-16
KTIJ-19110 "Enum class is never used / Safe Delete" false positive when using `values()`

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19110


dimonchik0036 (approved):

PR #1636, opened 2021-07-16
IDEA-222405: don't duplicate Gradle run VM arguments

abextm:
https://youtrack.jetbrains.com/issue/IDEA-222405


nskvortsov:
This PR does not contain neither explanation nor tests. It cannot be accepted as is. Feel free to reopen if it is still relevant.

PR #1637, opened 2021-07-19
Use `F2` for renaming instead of `shift-F6`

Da-Viper:
In vscode f2 is used for renaming but in the keymap it does not work for git rename commit so this changes it


dmitriysmirnovjb:
Merged the request manually - thank you for the contribution!

PR #1638, opened 2021-07-20
KTIJ-19139 Move variable declaration into `when`: block comment after the property to be inlined is removed

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19139


dimonchik0036 (approved):

PR #1640, opened 2021-07-21
[kotlin] Ignore errors when processing visible variables in the debugger

sfs:
This is a fix for a problem introduced in #1593.

As part of #1593 I added assertions to check the consistency of inlining locals. Unfortunately, it turns out that there are cases where the Kotlin compiler produces code which leads to an assertion error and a variable view stuck at "collecting data..."

This PR replaces the assertions with code to hide incorrect inlining locals instead.


nikita-nazarov:
Merged 71698a56269566eebe23e4a0fef265d1c9ad256f

PR #1641, opened 2021-07-21
KTIJ-19016 "Convert to lazy property" and "Convert property initializer to getter" intentions produce uncompilable code for `const val`

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19016


dimonchik0036 (approved):

PR #1642, opened 2021-07-21
KTIJ-19141 False positive "Replace size zero check with 'isEmpty'" with `list.count { predicate } == 0`

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19141


t-kameyama:
7450335d12ae8bd2b54058c621699b7c652ce764

vladimirdolzhenko:

PR #1643, opened 2021-07-22
IDEA-194699 Use actual parent class when rerunning failed tests with Gradle.

anuraaga:
Currently, the code-based enclosing class of a method is used to generate a rerun command for Gradle. This does not work well when the test is defined in a superclass though for two reasons

1) IntelliJ will display the parent class in the Tests window, even though it was showing the subclass during the failure
2) While it seems Gradle can still run e.g., `SuperClass.testMethod` when `SuperClass` is concrete, if it's abstract, it doesn't run at all. `SubClass.testMethod` works fine, and also displays the correct test name in the Tests window.

### DIFF
--- a/plugins/gradle/java/testSources/execution/test/runner/GradleRerunFailedTestsTest.kt
+++ b/plugins/gradle/java/testSources/execution/test/runner/GradleRerunFailedTestsTest.kt
@@ -81,4 +81,54 @@ class GradleRerunFailedTestsTest : GradleRerunFailedTestsTestCase() {
       """.trimIndent())
     assertFalse(performRerunFailedTestsAction())
   }
+
+  @Test
+  fun `test return failed tests abstract superclass`() {
+    createProjectSubFile("src/test/java/org/example/TestCase.java", """
+      package org.example;
+      import $jUnitTestAnnotationClass;
+      public abstract class TestCase {
+        @Test public void test1() {}
+        @Test public void test2() { throw new RuntimeException(); }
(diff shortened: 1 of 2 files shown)


anuraaga:
I wasn't able to request CLA through the website due to a server error so tried emailing. Hoping to get it resolved in time for 2021.2 release if possible :D

anuraaga:
Signed the contributor agreement - hopefully this is ok to review now :) @DeveloperHacker I noticed you created the rerunfailedtest related files, wondering if you can take a look?

plugins/gradle/java/testSources/execution/test/runner/GradleRerunFailedTestsTest.kt
```diff
@@ -81,4 +81,54 @@ class GradleRerunFailedTestsTest : GradleRerunFailedTestsTestCase() {
...
+          test4
+          test5
+          test6
+          test7
+      """.trimIndent())
+    assertTrue(performRerunFailedTestsAction())
+    assertThat(getJUnitTestsExecutionTree())
+      .isEqualTo("""
+        -[root]
+         -SubTestCase
```
anuraaga, line 128: Before the change, if `TestCase` is not abstract, this would say `TestCase` and list the tests. With it being abstract, it fails to execute anything and only lists `root`

DeveloperHacker:
@anuraaga Cool! Thank you!
I pushed this commit. Pull request will be closed when this commit becomes available on GitHub.

PR #1644, opened 2021-07-22
KTIJ-19143 False negative "Useless call on not-null type" with for isNullOrEmpty/isNullOrBlank if nullability was checked

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19143


klunnii:
Merged. Thank you!

PR #1645, opened 2021-07-24
KTIJ-15999 Incorrect `Convert to with` with nested class

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-15999


klunnii:
Fixed. Thank you!

PR #1646, opened 2021-07-26
Support arbitrary parameterized tests by looking for square brackets

nymanjens:
## Context

When filtering individual JUnit4 tests in IntelliJ, the runner will look for a method with the name of the test. For parameterized tests, there is no 1:1 relationship between test name and test method and so this fails.

For `junit.Parameterized` and test methods with parameters, there is an existing workaround for this in `JUnit4TestRunnerUtil`.

This PR supersedes https://github.com/JetBrains/intellij-community/pull/1617, which fixes the same problem in a different way.

## Problem when using TestParameterInjector

[TestParameterInjector](https://github.com/google/TestParameterInjector) is a runner developed at Google, and recently launched as open source project. In the last 2 months, [we had 6 different people asking us to fix this filtering issue in IntelliJ](https://github.com/google/TestParameterInjector/issues/6). The issue arises when fields are used for parameterization:

```java
public class MyTest {

  @TestParameter private boolean isOwner;

  @Test public void test() {...}
}
```

This will result in the following test names:

```
MyTest#test[isOwner=true]
MyTest#test[isOwner=false]
```

which will break the IntelliJ filtering.

## Proposed solution

If the test name contains a square bracket, have `isParameterized()` return `true`.

This makes sense because method names with brackets are illegal in Java. If `isParameterized()` would return `false` for a method with square brackets, its subsequent call to `clazz.getMethod()` would fail and nothing would be run. The logic triggered by `isParameterized()` is treating the test name as `methodName[otherstuff]`, which seems quite reasonable when there is a square bracket.

### DIFF
--- a/plugins/junit_rt/src/com/intellij/junit4/JUnit4TestRunnerUtil.java
+++ b/plugins/junit_rt/src/com/intellij/junit4/JUnit4TestRunnerUtil.java
@@ -235,6 +235,9 @@ public final class JUnit4TestRunnerUtil {
       }
       catch (NoSuchMethodException ignored) { }
     }
+    if(methodName.contains("[")) {
+      return true;
+    }
     return false;
   }
 


akozlova:
Ok, I'll add a test and some callback in case it would fail for some custom runner. Do you need to sign CA or google one is enough?

nymanjens:
> Do you need to sign CA or google one is enough?

I'd expect that if Google signed something on behalf of all employees, that's enough. There seem to be some other Googlers contributing to the IntelliJ codebase.

akozlova:
I wrote a test first and it passes just fine with my current implementation. Could you please describe steps which do not work for you. 
My test does the following:
1. Run test class
```

public class MyGoogleTest {

    @TestParameter({"1", "2"}) public int a;
    @TestParameter({"1", "2"}) public int b;

    @org.junit.Test
    public void simple() {
        System.out.println(a);
        System.out.println(b);
    }
}
```

2. choose node e.g. `simple[a=1,b=2]` from test and run it
3. only one test method is run.

Thanks

nymanjens:
I'm running this version:

> IntelliJ IDEA 2020.3.4 (Community Edition)
> Build #IC-203.8084.24, built on April 27, 2021

which probably doesn't have your fix, but that shouldn't matter because these methods don't have parameters anway.

If I run all tests, it works:
Screenshot from 2021-07-30 09-41-11

But if I then right-click on a single test and run it, I get the following error:
Screenshot from 2021-07-30 09-41-37

akozlova:
That version doesn't contain my fix and it works just fine now. 

When you select  the node in the tree, you get `paramString` propagated to the junit runner and then everything is correctly filtered due to the first part of the if condition:
```
if (parameterString != null || isParameterized(methodName, clazz))
```
was added in the same commit 456b5c81c6b512060d46c23aa8271bcdcf1fd4db

P.S. `parameterString` would be null if you would run one method in a class with all parameters, but this case is not supported in your runner (as far as I could check, you do not report hierarchical checks)

So I am closing this PR as everything works fine. The commit is included in master, first EAP of 2021.3 will include it. If you wish I can share a Nightly Build with you or you can build Community from sources yourself.
Thanks

nymanjens:
That's great news! I don't understand how `parameterString/programParameters` is populated, but I trust you if you say it works now.

Sorry for this dragging on for so long. I didn't realize there had been a change other than checking for method parameters.

alexjski:
Thank you for taking care of that! I wanted to make sure we are on the same page. I built intellij-community with source version from [Friday](https://github.com/JetBrains/intellij-community/commit/86611d0d248368fa82905b5982ec216e440ea320) and tested the case of selecting a single test by clicking the arrow by the method. It failed for me both in the presence or absence of the class-level parameters.

Did I miss a change which addresses this issue? I am more than happy to share more details if you are not aware of that -- just didn't want to waste your time if you already fixed that and I just failed to pick up the right version. Thank you in advance!

akozlova:
My fixes were put in mid June, so if you build from master then you must have them. So you state that it becomes worser because Jens was able to start a method.
Frankly speaking I would highly appreciate if you would file YT issue with the sample projects and exact steps to reproduce. Thanks

alexjski:
> Frankly speaking I would highly appreciate if you would file YT issue with the sample projects and exact steps to reproduce.

I filed a bug, as you requested: https://youtrack.jetbrains.com/issue/IDEA-275586.

PR #1647, opened 2021-07-28
KTIJ-19218 "Logger initialized with foreign class" does not handle object declarations

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19218


klunnii:
Merged. Thank you!

PR #1649, opened 2021-08-02
[kotlin] Test for printing variables when debugging reentrant inline functions

sfs:
This PR adds a test for the problem fixed in #1640 and adds some more comments to the code.

The underlying cause of the problem fixed in #1640 are reentrant inline functions with different variables. In this case, we don't see the scope introduction variables since they are hidden by later variables with the same name and the debug information appears to be inconsistent. Since we are only displaying the currently active inline call frame this is unlikely to be a problem in practice and we can simply hide inconsistent inlining locals.


nikita-nazarov:
Merged 8669a02aada1c7ab0de4dfa28f4a5a855e50609a

PR #1658, opened 2021-08-09
Enhance PsiAugmentProvider: support for PsiMethod

Auties00:
I was recently writing a plugin for IntelliJ and I noticed that PsiAugmentProvider only supports PsiClass. I've added support also for PsiMethod maintaining the same design choices and code style. I was also debating whether adding for PsiFIeld would be appropriate but I don't think that it makes any sense considering that PsiAugmentProvider already provides a way to change the modifiers or infer the type of a PsiElement which are the only use cases that I could think of for said element. On the other hand, adding support for augmenting the extend/implement/permit clause of a class seems to make sense also considering that it has been requested in the past on the IntelliJ forums. Before implementing that I was looking for some feedback on what has arleady been done though.


trespasserw:
Sorry for the late reply.

At the moment, we don't think extending methods his way is worth the added complexity.

PR #1659, opened 2021-08-09
FIR/UAST: update test data after migration

jsjeon:
https://github.com/JetBrains/intellij-community/commit/ad992181eb5150804894a4ea916c5c2310d39bcd adds test output for newly added test inputs in `intellj-community`, but we should run both UAST plugins; compare them; and hold identical outputs if any. They're identical, hence we need multiple runs to make sure.


jsjeon:
cc @darthorimar

darthorimar:
Merged, thank you!

PR #1660, opened 2021-08-10
FIR/UAST: call kind for array initializer in annotation call

jsjeon:
```kt
annotation class AnnotationArray(vararg val value: Annotation)

@AnnotationArray(Annotation(strings = arrayOf("sar1", "sar2")))
class C
```

Call kind for `arrayOf` is `UastCallKind.NESTED_ARRAY_INITIALIZER`, and to determine that, we need to know that:
1) `KtCallElement` is part of annotation argument, and
2) callee is `arrayOf` variant.

All necessary pieces are there in FE 1.0 UAST. We just need consolidated `arrayOf` calls so that the name from either `ResolvedCall` or `KtCall` can be checked if it's `arrayOf` call variant: [KOTLIN-CR-158](https://kotlin.jetbrains.space/p/kotlin/reviews/158/)


darthorimar (approved):

darthorimar:
Merged, thanks!

PR #1661, opened 2021-08-10
Update CoverageSuiteChooserDialog.java

chessteralbert:
Allowed folder and multiple file selection on top of single file selection for the coverage suite selection. Selected folders will be iterated through for coverage files and selected in the UI accordingly.


akozlova:
Sorry for late reaction. Though I agree that multi-selection can be useful, I'd rather avoid directories as it's not a common practice in the IDE. Could you please explain the use case in details? Thanks

chessteralbert:
> Sorry for late reaction. Though I agree that multi-selection can be useful, I'd rather avoid directories as it's not a common practice in the IDE. Could you please explain the use case in details? Thanks

Essentially, we are trying to combine multiple .ic files into 1. The root of it all is .exec files from jacoco can be merged into 1 file. .ic files cannot be merged into 1 file it looks like. If there is a tool for merging .ic files, it would be a better solution.
To work around that, we can select multiple (e.g. more than 50 and growing) .ic coverage files via the user interface. Currently, doing this would mean repetitive add -> single select -> open for the user. This pull request is trying to automate that process.

For example, if we needed to load the following one by one, it is pretty cumbersome.
image
image

akozlova:
I am fine with multiselection, I don't like the selection of the directory.

chessteralbert:
> I am fine with multiselection, I don't like the selection of the directory.

cool cheers, better than nothing in the moment. will update in a bit

akozlova:
Merged manually with some small changes. Thanks

PR #1663, opened 2021-08-12
FIR IDE: Initial version of handlers for parameter info popup

punzki:
All the `TODO`s in the `KotlinHighLevelFunctionParameterInfoHandler.kt` file must be resolved to enable all the remaining tests. I will continue working on it in a separate branch while this is being reviewed.

This is meant to go with the HL API changes in MR 161.


plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/parameterInfo/AbstractParameterInfoTest.kt
```diff
@@ -68,27 +68,36 @@ abstract class AbstractParameterInfoTest : KotlinLightCodeInsightFixtureTestCase
...
-            val mockCreateParameterInfoContext = MockCreateParameterInfoContext(file, myFixture)
-            val parameterOwner = handler.findElementForParameterInfo(mockCreateParameterInfoContext) as PsiElement
+            lateinit var handler: ParameterInfoHandler<PsiElement, Any>
+            lateinit var mockCreateParameterInfoContext: MockCreateParameterInfoContext
+            lateinit var parameterOwner: PsiElement
+            executeOnPooledThreadInReadAction {
+                val handlers = ShowParameterInfoHandler.getHandlers(project, KotlinLanguage.INSTANCE)!!
+                @Suppress("UNCHECKED_CAST")
+                handler = (handlers.firstOrNull { it.findElementForParameterInfo(context) != null }
+                    ?: error("Could not find parameter info handler")) as ParameterInfoHandler<PsiElement, Any>
```
jsjeon, line 78: This line looks like the `error` call is casted to `ParameterInfoHandler`. :p

Isn't it better to do something like:
```kt
handler = handlers.firstOrNull { ... } as? ParameterInfoHandler<...>
    ?: error(...)
```
Also, with `as?`, we may not need `@Suppress`?

punzki: `getHandlers` returns a raw type hence there's always need for an unchecked cast; can't get rid of the `@Suppress`. However I do agree it reads better the way you did it, with `error` not looking like it's casted!

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -0,0 +1,451 @@
...
+        if (context.parameterOwner !== argumentList) {
+            context.removeHint()
+        }
+        val parameterIndex = getParameterIndex(context, argumentList)
+        context.setCurrentParameter(parameterIndex)
+
+        val callElement = argumentList.parent as? KtElement ?: return
+        for (objectToView in context.objectsToView) {
+            val candidateInfo = objectToView as? CandidateInfo ?: continue
+            analyse(callElement) {
```
tgeng, line 139: Should this analyse block be hoisted out of the loop? Then I believe the first when branch can be moved out of the loop as well.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -0,0 +1,451 @@
...
+
+    private fun KtAnalysisSession.renderParameter(parameter: KtValueParameterSymbol, includeName: Boolean): String {
+        return buildString {
+            val annotationFqNames =
+                parameter.annotations
+                    .filter {
+                        // For primary constructor parameters, the annotation use site must be "param" or unspecified.
+                        it.useSiteTarget == null || it.useSiteTarget == AnnotationUseSiteTarget.CONSTRUCTOR_PARAMETER
+                    }
+                    .mapNotNull { it.classId?.asSingleFqName() }
```
tgeng, line 216: It looks like FE1.0 does not print package names for annotations. Maybe doing the same?

plugins/kotlin/fir/test/org/jetbrains/kotlin/idea/fir/parameterInfo/AbstractFirParameterInfoTest.kt
```diff
@@ -0,0 +1,16 @@
...
+import java.nio.file.Paths
+
+abstract class AbstractFirParameterInfoTest : AbstractParameterInfoTest() {
+    override fun isFirPlugin() = true
+
+    override fun doTest(fileName: String) {
+        IgnoreTests.runTestIfNotDisabledByFileDirective(Paths.get(fileName), IgnoreTests.DIRECTIVES.IGNORE_FIR) {
+            super.doTest(fileName)
+        }
+    }
```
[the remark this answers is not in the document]
punzki: I added the tear-down and for some inexplicable reason, one of the tests passed (`withLib1/useJavaFromLib.kt`). I tried to debug it a little bit, and without the tear-down, the function being called was unresolved. I didn't see the need to test further as I am sure it has something specific to the test set-up.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -0,0 +1,451 @@
...
+                    argumentMapping,
+                    valueParameters,
+                    parameterToText,
+                    isCallResolvedToCandidate
+                )
+            }
+        }
+    }
+
+    private fun areEqual(first: KtFunctionLikeSymbol, second: KtFunctionLikeSymbol): Boolean {
```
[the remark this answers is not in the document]
punzki: That would be better. Are you comparing the signatures then? That would be safer. What I'm doing works because there shouldn't be any changes in the PSI outside of the function call itself being updated.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: Yes, that works. Thanks for the suggestion!

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: I am only printing the short name. See call to `it.shortName()` a few lines below. I'll modify one of the tests `annotations/FunctionCall.kt` to put it in a package to test that only the short name is rendered.

This part is only for filtering out nullability annotations, so we need the FQ name.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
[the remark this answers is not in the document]
darthorimar: There is not need to compare symbols by the signature as `KtSymbol` may be used only inside the read action it was created in, so it should never be changed during it lifetime

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -0,0 +1,451 @@
...
+        )
+
+        return true
+    }
+
+    data class CallInfo(
+        val callElement: KtElement,
+        val valueArguments: List<KtValueArgument>?,
+        val arguments: List<KtExpression>,
+        val argumentMapping: LinkedHashMap<KtExpression, KtValueParameterSymbol>,
```
darthorimar, line 441: KtValueParameterSymbol cannnot be stored in a class, only symbol pointers may be KtSymbolPointer may be stored. It violates the contracts that KtSymbols should not be leaked outside KtAnalysisSessions they were created in

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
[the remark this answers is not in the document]
punzki: @darthorimar Not sure if I understood your comment correctly. The two symbols compared here are going to be from different read actions: one from `findElementForParameterInfo` and the other from `updateParameterInfo` (restored). So these may be two different symbols.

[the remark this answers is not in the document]
tgeng: Ah across different session would be tricky...

Actually, would it be possible to populate the `CallInfo` directly when creating `CandidateInfo`?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
tgeng: Or maybe just store the final rendered strings in here instead?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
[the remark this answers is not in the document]
punzki: At first, I thought I could populate all the `CallInfo` in `findElementForParameterInfo`. But it shouldn't be done there because `findElementForParameterInfo` is called once when the parameter info popup is requested, but `updateParameterInfo` is called once and then _also_ every time the user types something while the popup is open (the highlighting dynamically changes). So you still need to resolve again, because things like `argumentMapping` and `isCallResolvedToCandidate` can change as the user types.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: I'm storing the index as values instead. See latest version. I still need `parameterToText` map to render the parameters with no arguments.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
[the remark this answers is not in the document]
darthorimar: @punzki There should never be a case when you meet two symbols from different sessions in the same scope. Every `KtSymbol` should be used ONLY inside `analyse {..}` block and cannot be leaked outside it. So, there should not be a situation when two symbols from different `analyse {..}` blocks are available together in the same scope. So, there is no need to compare them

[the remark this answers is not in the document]
darthorimar: About restoring symbols:
Consider the following code

```
val ktFunction: KtNamedFunction = ...

val ponter = readAction {
	analyse(context) { // analysis context #1
		val symbol: KtSymbol = ktFunction.getSymbol() // symbol A
		symbol.createPointer()
	}
}

// modify psi  somehow


readAction {
	analyse(context) {  // analysis context #2
		val symbol: KtSymbol = pointer.restore() // symbol B
		val otherSymbol: ktSymbol = ...
		otherSymbol == symbol
	}
}

```

The symbol `A` you create belongs to `analysis context #1`. You can restore that symbol by pointer only in some analysis context. So, when you restore your symbol by `pointer` you get the symbol `B`. Even symbols `A` and `B` represents the same declaration, those symbols are different. The symbol `A` belongs to  `analysis context #1`, meanwhile symbol `B` belongs to `analysis context #2` in which it was restored. So, them you compare `otherSymbol` with `symbol` inside `analysis context #2` they will both belong to the same  `analysis context #2`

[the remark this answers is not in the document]
punzki: Using `==` works, but there seems to be an issue with symbol restoration for Java symbols:
```
val javaCandidateSymbol = ...
val pointer = javaCandidateSymbol.createPointer()
val restoredSymbol = pointer.restoreSymbol()
assert(javaCandidateSymbol == restoredSymbol) // BOOM
```
The root cause seems to be HL API creating new `ScopeSession()` every time it is getting unsubstituted scope during restoration, and therefore a new enhanced `FirFunction` is created and a new symbol is as well (instead of using the old enhanced `FirFunction` in the cache).

Changed to use `==`, and disabled the relevant test for now.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -0,0 +1,454 @@
...
+    override fun getActualParameterDelimiterType(): KtSingleValueToken = KtTokens.COMMA
+
+    override fun getArgListStopSearchClasses(): Set<Class<out KtElement>> = STOP_SEARCH_CLASSES
+
+    override fun getArgumentListClass() = argumentListClass.java
+
+    override fun showParameterInfo(element: TArgumentList, context: CreateParameterInfoContext) {
+        context.showHint(element, element.textRange.startOffset, this)
+    }
+
```
darthorimar, line 98: This code is the same as in FE1.0 version. Maybe it can be extracted to a base class?

darthorimar (approved):

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: I have a TODO to come up with a common base class. The structure is generally similar, but I wanted to do it separately.

darthorimar:
Merged! thank you :)

PR #1664, opened 2021-08-12
FIR IDE: fix IgnoreTests.handleTestWithWrongDirective

tgeng:
Whenever this function is called, it means the directive is wrong. The current
check actually prevents the test from failing if it's missing a positive
directive

### DIFF
--- a/plugins/kotlin/frontend-independent/tests/test/org/jetbrains/kotlin/test/utils/IgnoreTests.kt
+++ b/plugins/kotlin/frontend-independent/tests/test/org/jetbrains/kotlin/test/utils/IgnoreTests.kt
@@ -126,15 +126,10 @@ object IgnoreTests {
             true -> "passes"
         }
 
-        val directiveIsOutdated = when {
-            !testPasses && directive is EnableOrDisableTestDirective.Enable -> true
-            testPasses && directive is EnableOrDisableTestDirective.Disable -> true
-            else -> false
-        }
-
-        val directiveIsMissing = !directiveIsOutdated
-
-        if (directiveIsMissing && INSERT_DIRECTIVE_AUTOMATICALLY) {
+        if (INSERT_DIRECTIVE_AUTOMATICALLY &&
+            (directive is EnableOrDisableTestDirective.Enable && testPasses ||
+                    directive is EnableOrDisableTestDirective.Disable && !testPasses)
+        ) {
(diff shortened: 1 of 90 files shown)


punzki (approved):

jsjeon (approved):

plugins/kotlin/frontend-independent/tests/test/org/jetbrains/kotlin/test/utils/IgnoreTests.kt
```diff
@@ -126,15 +126,7 @@ object IgnoreTests {
...
-        val directiveIsOutdated = when {
-            !testPasses && directive is EnableOrDisableTestDirective.Enable -> true
-            testPasses && directive is EnableOrDisableTestDirective.Disable -> true
-            else -> false
-        }
-
-        val directiveIsMissing = !directiveIsOutdated
-
-        if (directiveIsMissing && INSERT_DIRECTIVE_AUTOMATICALLY) {
+        if (INSERT_DIRECTIVE_AUTOMATICALLY && directive is EnableOrDisableTestDirective.Enable) {
```
darthorimar, line 129: The fix seems to be incorrect as it now handles only the case when the directive is `// FIR_COMPARISON` is used to mark the test as passing, but some tests uses `// FIR_IGNORE` to mark tests as failing, in that case `directive is EnableOrDisableTestDirective.Disable`

darthorimar (changes requested):

plugins/kotlin/frontend-independent/tests/test/org/jetbrains/kotlin/test/utils/IgnoreTests.kt
tgeng: This logic is only trying to automatically fix the test if it passes. This aligns with the previous code, which doesn't offer to fix the test by removing `// IGNORE_FIR`.

I can add the ability to remove `// IGNORE_FIR` if that's desirable though.

tgeng: Thanks let me update it.

darthorimar (approved):

darthorimar:
Merged manually, thanks!

PR #1665, opened 2021-08-13
KTIJ-15194 Suggest to import extension operator function used in operator form on a literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-15194


yanex:
False positive:

```
fun usage() {
    foo(true)
}

fun foo(a: String) {}
fun foo(a: Int) {}
```

t-kameyama:
Added 34fd6ab, please take another look.

t-kameyama:
c47d1c98760330f3f568e23cd7f2ae2a2a4d3b42

PR #1669, opened 2021-08-16
FIR UAST: convert destructuring assignments

jsjeon:
When converting `KtDestructuringDeclaration`, its assignments (to destructured properties) were not done (w/ TODO left). This commit simply adds that part, which aligns with FE 1.0 UAST behavior. (Alas, expression conversion is not fully identical, and thus we can't commonize such logic yet.)


jsjeon:
cc @darthorimar

darthorimar (approved):

darthorimar:
Merged, thanks!

PR #1670, opened 2021-08-17
KTIJ-19363 Do not report "Redundant qualifier name" for `values`/`valueOf` methods of enum classes used inside itself

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19363


dimonchik0036 (changes requested):
image

```kotlin
package main.objects

enum class PrimitiveType {
    BYTE,
    CHAR,
    SHORT,
    INT,
    LONG,
    FLOAT,
    DOUBLE,
    BOOLEAN;

    companion object {
        val exceptBoolean = main.objects.PrimitiveType.values().filterNot { it == BOOLEAN }
        val exceptBoolean2 = PrimitiveType.values().filterNot { it == BOOLEAN }
    }
}
```
Expected after:
```kotlin
val exceptBoolean = PrimitiveType.values().filterNot { it == BOOLEAN }
val exceptBoolean2 = PrimitiveType.values().filterNot { it == BOOLEAN }
```

PR #1671, opened 2021-08-17
Type Migration: introducing primitive functional interfaces

SmirnovOleg:
Hey!

We are a collaborative group of researchers from JetBrains Research and Oregon State University, and we are testing our data-driven [plugin](https://github.com/JetBrains-Research/data-driven-type-migration), which is based on the IntelliJ's [Type Migration](https://www.jetbrains.com/help/idea/type-migration.html) framework and adjusts it using custom structural-replace templates that express the adaptations required to perform the type changes.

I want to apply several type changes using it and open the PR, thus introducing primitive functional interfaces in order to prevent unnecessary boxing (like `BooleanSupplier` instead of `Supplier<Boolean>`, `OptionalInt` instead of `Optional<Integer>`, `Predicate<T>` instead of `Function<T, Boolean>`, etc.), since it can affect the performance of the code (Effective Java, Items 44, 61). 

This patch is made automatically by the plugin, so it would help us a lot to evaluate the usefulness of our approach!


SmirnovOleg:
I've reverted commits that affect public methods' signatures (as @akozlova suggested, thanks!).
Also, most of the proposed code changes are local, so I hope they do not even need running test suites. Anyway, I've run all tests for the affected packages on my local machine, it passed.

platform/vcs-impl/src/com/intellij/openapi/vcs/changes/ControlledCycle.java
```diff
@@ -22,7 +22,7 @@
   private final AtomicBoolean myActive;
 
   public ControlledCycle(@NotNull Project project,
-                         final Supplier<Boolean> callback,
+                         final BooleanSupplier callback,
```
SmirnovOleg, line 25: There is only one place in the code where the instance of this public class is created, and it uses a lambda to express the `callback` parameter, so this particular change shouldn't break anything.

PR #1674, opened 2021-08-18
[kotlin] Don't suggest import quickfix for built-in functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19362


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/AbstractImportFix.kt
```diff
@@ -265,7 +267,11 @@ internal abstract class OrdinaryImportFixBase<T : KtExpression>(expression: T, f
             ) { it == name }
         )
 
-        return result
+        val importedFqNames = element?.containingKtFile?.importDirectives?.mapNotNull { it.importedFqName }.orEmpty()
+        return result.filterNot {
+            val importableFqName = it.importableFqName
+            importableFqName?.parentOrNull() in StandardNames.BUILT_INS_PACKAGE_FQ_NAMES && importableFqName !in importedFqNames
```
nikita-nazarov, line 273: Can you please clarify why the second part of the predicate is needed `importableFqName !in importedFqNames`? 
 
If we consider the following example, the problem will persist: the import is suggested and the fix does nothing, 
```
import kotlin.collections.maxByOrNull

fun <T> im(list: List<String>) {
    list.maxByOrNull<T, T> { 0 }
}
```

Maybe it is better to remove the second check? So it won't be possible to apply the import fix to functions from built in packages.

nikita-nazarov (changes requested):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/AbstractImportFix.kt
t-kameyama: The second is required for [importAliasLetAlreadyExists.kt](https://github.com/JetBrains/intellij-community/blob/35193ddc0ed27c3bb0b6125fa848256482fe0044/plugins/kotlin/idea/tests/testData/quickfix/autoImports/importAliasLetAlreadyExists.kt).

t-kameyama: Fixed second condition and added a test for the above example.

nikita-nazarov (approved):
Thanks for the contribution! Will push the pr manually.

nikita-nazarov:
5bf6cc93b116386ad5cdfd599b2b55f2a65469a8

PR #1675, opened 2021-08-18
FIR UAST: resolve to class for SAM constructor

jsjeon:
`UCallExpression.classReference` returns the class reference if the call is a constructor call. That is applied to SAM constructor too, but FIR UAST missed that.

To support that, we need a set of additions to FIR IDE/HL API (TBD) that spans:
* a new symbol kind (of course, for SAM constructor),
* its own symbol, along with pointer to restore, and
* a resolver that helps restoration.


darthorimar (approved):

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/FirKotlinUastResolveProviderService.kt
```diff
@@ -184,8 +185,10 @@ interface FirKotlinUastResolveProviderService : BaseKotlinUastResolveProviderSer
         analyseForUast(ktCallElement) {
             val resolvedFunctionLikeSymbol = ktCallElement.resolveCall()?.targetFunction?.candidates?.singleOrNull() ?: return null
             return when (resolvedFunctionLikeSymbol) {
-                is KtConstructorSymbol -> toPsiClass(resolvedFunctionLikeSymbol.annotatedType.type, ktCallElement)
-                // TODO: SAM constructor
+                is KtConstructorSymbol,
+                is KtSamConstructorSymbol -> {
```
punzki, line 189: I'm just curious. AFAICT, currently, without either of your changes, for a SAM constructor call, `resolvedFunctionLikeSymbol` would be a `KtFunctionSymbol` with `origin == SAM_CONSTRUCTOR`. Couldn't you have just checked that here, without a new kind of `KtSymbol`? You would still get the same type.

punzki: Not saying that `KtSamConstructorSymbol` (or the other API changes) wouldn't be useful, but it seemed sufficient to me to just check the origin, for what you're trying to do. (Unless I'm missing something...)

jsjeon: You're not missing anything, i.e., you're right. Without new addition of symbol/pointer, currently, it is `KtFunctionSymbol` with that origin. That was exactly the initial approach I took, and I can confirm it works.

However, somehow, I started digging further since the fact that the FIR element inside that `KtFunctionSymbol` is a synthetic function caught my eyes. Then, with some tweaks, such as symbol kind addition, reached some corner cases here and there, including restoration, and then brought that to the discussion, along with a suggestion of new symbol/pointer/kind. Note that the counter part in FE1.0 ended up with `ConstructorDescriptor` and `SamConstructorDescriptor`, so having `KtConstructorSymbol` and `KtSamConstructorSymbol` seems aligned and intuitive. Most of all, I think it's better to assign its own kind of symbol to something "synthetic."

punzki: Thanks for the explanation! I can see how it makes more sense to have a separate symbol type.

jsjeon:
Looks like we can drop the 2nd commit as https://github.com/JetBrains/intellij-community/commit/39fd40ae3f2ed11971cadb4019f84c067a605879 is done the same thing.

jsjeon:
> Looks like we can drop the 2nd commit as [39fd40a](https://github.com/JetBrains/intellij-community/commit/39fd40ae3f2ed11971cadb4019f84c067a605879) is done the same thing.

Removed

PR #1679, opened 2021-08-20
FIR/UAST: SAM type retrieval from lambda expressions

jsjeon:
To be updated: first commit is identical to #1675 and this work depends on SAM constructor addition in FIR UAST as well as FIR IDE.

------

`ULambdaExpression.functionalInterfaceType` returns SAM type if the lambda expression is used as an argument to SAM interface object creation. This PR implements that API in FIR UAST and tests SAM type retrieval from lambda expressions by sharing FE 1.0 UAST's tests.

Instead of mimicking FE 1.0 UAST [counterpart](https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin/src/org/jetbrains/uast/kotlin/internal/kotlinInternalUastUtils.kt#L208), I think it's better to integrate "expected type" retrieval from lambdas in general into FIR IDE: TBD.

In addition, we need APIs to check if given `KtType` is a functional interface type (a.k.a. SAM type) and unwrap upper/lower bound of flexible `KtType` (e.g., `Thread({ /* Runnable */})` where parameter type is returned as flexible type).


plugins/kotlin/uast/uast-kotlin/test/org/jetbrains/uast/test/comparasion/FE1UastApiTest.kt
```diff
@@ -0,0 +1,38 @@
...
+@RunWith(JUnit3RunnerWithInners::class)
+class FE1UastApiTest : AbstractFE1UastTest() {
+    override fun check(testName: String, file: UFile) {
+        // Bogus
+    }
+
+    @TestMetadata("uast-kotlin/testData")
+    @TestDataPath("/")
+    class Legacy : AbstractFE1UastTest(), UastApiTestBase {
+        override var testDataDir = File("../uast-kotlin/testData").normalize()
```
darthorimar, line 25: Could you please specify a path relative to a `KotlinRoot.DIR` (https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/tests-common/test/org/jetbrains/kotlin/test/KotlinRoot.kt#L16) 

similar to https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin-fir/test/org/jetbrains/kotlin/idea/fir/uast/FirUastResolveApiTest.kt#L18

This will help tests on the TC

darthorimar (changes requested):

plugins/kotlin/uast/uast-kotlin/test/org/jetbrains/uast/test/comparasion/FE1UastApiTest.kt
jsjeon: Done (also for existing resolve API tests)

PR #1680, opened 2021-08-20
FIR IDE: Initial version of handlers for type parameter info popup

punzki:
~~This includes commits from https://github.com/JetBrains/intellij-community/pull/1663. Only the last 4 commits are new.~~ Now rebased.

This depends on high-level API changes in [KOTLIN-MR-173](https://kotlin.jetbrains.space/p/kotlin/reviews/173/timeline).


darthorimar (approved):

plugins/kotlin/uast/uast-kotlin-fir/testData/legacyRenderLog/InnerNonFixedTypeVariable.log.fir.txt
```diff
@@ -12,7 +12,7 @@ UFile (package = )
                             USimpleNameReferenceExpression (identifier = filterIsInstance, resolvesTo = null)
                     UCallExpression (kind = UastCallKind(name='method_call'), argCount = 2))
                         UIdentifier (Identifier (mapTo))
-                        USimpleNameReferenceExpression (identifier = <anonymous class>, resolvesTo = null)
+                        USimpleNameReferenceExpression (identifier = mapTo, resolvesTo = null)
```
darthorimar, line 15: This and the following changes in UAST testdata does not seem to be consistent with FE1.0 implementation. (it has ` <anonymous class>` instead of `mapTo`). @jsjeon what do you think?

punzki: I think this is more correct than FE 1.0. If you put the code from the test file:
```kt
class Some<T : Some<T>>
fun test(list: List<Any>) {
    list.filterIsInstance<Some>().mapTo(mutableSetOf()) {

    }
}
```
...in the IDE, you get a `WRONG_NUMBER_OF_TYPE_ARGUMENTS` error on the `Some` type argument of `filterIsInstance`. But then you cannot follow the reference and go to the definition of `mapTo`. With my changes in KOTLIN-MR-173, I'm able to go to the definition for `mapTo` (in stdlib). I don't know why you can't follow it in FE 1.0, but it seems logical to me that you should be able to, since you have a candidate for `filterIsInstance`.

AFAICT, the changes in KOTLIN-MR-173 make the candidate symbol from `ConeWrongNumberOfTypeArgumentsError` available, which I believe FIR UAST picks up when getting the candidate function for a call.

jsjeon: Thank you for the investigation. Without that, perhaps I thought it'd be one of cases where FIR UAST is doing better (there are some, e.g., super type of anonymous object).

I think this rather discloses a hidden issue I overlooked. That `<anonymous class>` is quite misleading (and I need to see where such message is rendered that way), but it means the corresponding name reference is _not_ resolved. Although we can still attempt to resolve `mapTo`, an error on `filterIsInstance` means resolution of it and all other following (dot) qualified references are not resolvable. Thus, picking up `mapTo` and returning it as a _resolved_ reference seem wrong.

But, what really goes wrong is, uses of `resolveCall` in FIR UAST base service implementation didn't check whether the candidates are from error call target or successful call target. 🤦  In this case, I guess it's coming from error call target; previously no candidates at all, but now with one candidate, which is still not successful call target.

Maybe you can land this as-is. I'll revisit all of those places that retrieve candidates and revise them all together.

punzki: I guess you have a point, the return type of `filterIsInstance` has an error type argument. But you still know the return type is `List` so one could argue that you _can_ still follow the reference even if you don't have an instantiated type parameter.

darthorimar:
Merged, thank you!

PR #1684, opened 2021-08-25
feat(vscode-macos-keymap): add actions of HOME, END and control A keys to keymap

leungyauming:
This pull request is created to solve the issue below.

### Issue

HOME key and control A combination won't move the cursor to the start of the line, 
END key won't move the cursor to the end of the line.

### Expect Behavior

Home key and control A combination move the cursor to the end of the line,
END key moves the cursor to the end of the line.


yole:

PR #1685, opened 2021-08-25
FIR/UAST: commonize accessor calls to synthetic Java property

jsjeon:
`USimpleReferenceExpression` is not fully commonized because `KtSimpleNameExpression` may refer to synthetic Java property. And if so, FE 1.0 UAST creates accessor call expression on-the-fly (for the given visitor). FIR IDE didn't have HL APIs to support resolving `KtSimpleNameExpression` to accessor call that refers to synthetic Java property. [KOTLIN-CR-175](https://kotlin.jetbrains.space/p/kotlin/reviews/175/) is adding that. Then, this PR adds step-by-step commonizations that
* adds base service-based conversion of a setter value, if any
* removes redundant `ReferenceAccess` in FE 1.0 UAST (and uses the same one in frontend-independent module)
* abstracts call resolution and receiver type resolution from `KtSimpleNameExpression` if it refers to synthetic Java property
* removes unnecessary `ResolvedCall` uses since property accessor won't have type arguments
* and finally commonizes the parent expression---`USimpleReferenceExpression`


darthorimar (approved):

xiexed (approved):

darthorimar:
Merged, thank you!

PR #1686, opened 2021-08-26
[kotlin] ReplaceWith in whole project: don't add unnecessary type to lambda parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-13800
https://youtrack.jetbrains.com/issue/KTIJ-19360


dimonchik0036 (approved):
Thanks! Will be merged manually with an additional commit-rename

PR #1688, opened 2021-08-27
FIR Completion: add FirDeclarationFromUnresolvedNameContributor and priority weigher

tgeng:
`FirDeclarationFromUnresolvedNameContributor` mostly behave identical with
`FromUnresolvedNamesCompletion` but additionally provide completion entries for
qualified unresolved references.

This PR needs https://kotlin.jetbrains.space/p/kotlin/reviews/167/timeline.


plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirDeclarationFromUnresolvedNameContributor.kt
```diff
@@ -0,0 +1,149 @@
...
+                if (receiver != null) {
+                    val actualReceiverType = receiver.getKtType() ?: return false
+                    val expectedReceiverType = getReceiverType(symbol) ?: return false
+                    !(actualReceiverType isNotSubTypeOf expectedReceiverType)
+                } else {
+                    // If there is no explicit receiver at call-site, we check if any implicit receiver at call-site matches the extension
+                    // receiver type for the current declared symbol
+                    val extensionReceiverType = symbol.receiverType?.type
+                    if (extensionReceiverType != null) {
+                        !getImplicitReceiverTypesAtPosition(unresolvedRef).none { it isSubTypeOf extensionReceiverType }
```
darthorimar, line 94: `!list.none(p) <=> list.any(p)` it is just more readable without double negation

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirDeclarationFromUnresolvedNameContributor.kt
```diff
@@ -0,0 +1,149 @@
...
+            ?: currentDeclarationInFakeFile).getSymbol()) {
+            is KtCallableSymbol -> {
+                if (refExprIsType) return false
+                // If current declaration is a function, we only offer names of unresolved function calls. For property declarations, we
+                // always offer an unresolved usage because a property may be called if the object has an `invoke` method.
+                if (!refExprIsCallRef && !refExprIsCall && symbol is KtFunctionLikeSymbol) return false
+                if (receiver != null) {
+                    val actualReceiverType = receiver.getKtType() ?: return false
+                    val expectedReceiverType = getReceiverType(symbol) ?: return false
+                    !(actualReceiverType isNotSubTypeOf expectedReceiverType)
```
darthorimar, line 88: `!(actualReceiverType isNotSubTypeOf expectedReceiverType)` <=>  actualReceiverType isSubtypeOf expectedReceiverType`

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/Completions.kt
```diff
@@ -51,10 +53,12 @@ internal object Completions {
 
             is FirUnknownPositionContext -> {
                 complete(factory.keywordContributor(0), positionContext)
+                complete(factory.declarationFromUnresolvedNameContributor(0), positionContext) // for parameter declaration
```
darthorimar, line 56: Please, create new position context for such case.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/Completions.kt
```diff
@@ -18,12 +18,14 @@ internal object Completions {
...
                 complete(factory.classifierContributor(0), positionContext)
                 complete(factory.packageCompletionContributor(1), positionContext)
+                complete(factory.declarationFromUnresolvedNameContributor(1), positionContext)
             }
 
             is FirTypeNameReferencePositionContext -> {
                 complete(factory.classifierContributor(0), positionContext)
                 complete(factory.keywordContributor(1), positionContext)
                 complete(factory.packageCompletionContributor(2), positionContext)
+                complete(factory.declarationFromUnresolvedNameContributor(1), positionContext)
```
darthorimar, line 28: Why do we need `declarationFromUnresolvedNameContributor` on completion in type reference?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirDeclarationFromUnresolvedNameContributor.kt
tgeng: What am I thinking 🤣

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirDeclarationFromUnresolvedNameContributor.kt
tgeng: Thanks!

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/Completions.kt
tgeng: This is to complete an unresolved class name.

tgeng: This is for  variable and function completion. For example, with `val i<caret>`, the fake file has `val  iX.f` and hence `iX` is treated as a type reference.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/Completions.kt
tgeng: Thanks for the suggestion! Added `FirValueParameterPositionContext` for this case.

darthorimar:
Everything except of using `ToFromOriginalFileMapper` is OK

tgeng:
Hi @darthorimar I have removed usage of `ToFromOriginalFileMapper` and with https://kotlin.jetbrains.space/p/kotlin/reviews/217/timeline all HL API calls are working as expected. However, looking for unresolved references is broken if user is typing the value parameter. For example, if user has the following in the editor

```
fun test(a<caret>) {
  abc
}
```

The fake file then contains

```
fun test(aIntellijIdeaRulezzz$) {
  abc
}
```

The `$` at the end makes PSI parser think the fake file contains two declarations. See the following image for more details.
image
I think removing `$` should work. But I am not sure why it was added in the first place. What do you think?

tgeng:
> Hi @darthorimar I have removed usage of `ToFromOriginalFileMapper` and with https://kotlin.jetbrains.space/p/kotlin/reviews/217/timeline all HL API calls are working as expected. However, looking for unresolved references is broken if user is typing the value parameter. For example, if user has the following in the editor
> 
> ```
> fun test(a<caret>) {
>   abc
> }
> ```
> 
> The fake file then contains
> 
> ```
> fun test(aIntellijIdeaRulezzz$) {
>   abc
> }
> ```
> 
> The `$` at the end makes PSI parser think the fake file contains two declarations. See the following image for more details. image I think removing `$` should work. But I am not sure why it was added in the first place. What do you think?

I have made this change and it looks like it doesn't break anything.

darthorimar:
@fedochet could you please help here with handling `$` in parameters list? I reused this dummy identifier logic from FE1.0 plugin but I'm not sure it is needed here

fedochet:
A quick note: currently removing the '$' from the dummy identifier makes completion suggest unusable identifiers, for example:

```kotlin
fun foo(a: Int = <caret>, b: Int) {}
```

Here you should not be able to see `b` in the completion, but you will see it if there are no '$' (because '$' 'breaks' the parsing of the following parameters)

tgeng:
> From the other hand, we have this commit, which does approximately the same thing for the constructor parameters: https://jetbrains.team/p/ij/repositories/intellij/revision/0e2bb42852706d503250f737dfc655b3b985e4d7

Thanks Roman, unfortunately I am not able to see the linked commit. BTW, is there any other reason why `$` is used? If not, maybe we can do the filtering in a later stage during completion? Breaking a declaration into two seems very disruptive if the goal is to prevent unusable identifiers.

tgeng:
> > From the other hand, we have this commit, which does approximately the same thing for the constructor parameters: https://jetbrains.team/p/ij/repositories/intellij/revision/0e2bb42852706d503250f737dfc655b3b985e4d7
> 
> Thanks Roman, unfortunately I am not able to see the linked commit. BTW, is there any other reason why `$` is used? If not, maybe we can do the filtering in a later stage during completion? Breaking a declaration into two seems very disruptive if the goal is to prevent unusable identifiers.

I have updated FE1.0 completion logic to filter out the problematic cases.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirDeclarationFromUnresolvedNameContributor.kt
```diff
@@ -0,0 +1,142 @@
...
+        val refExprIsCallRef = refExprParent is KtCallableReferenceExpression
+        val refExprIsCall = refExprParent is KtCallExpression
+        val refExprIsType = refExprParent is KtUserType
+        val receiver = if (refExprIsCall) refExprParent.getReceiverForSelector() else unresolvedRef.getReceiverForSelector()
+        return when (val symbol = currentDeclarationInFakeFile.getSymbol()) {
+            is KtCallableSymbol -> {
+                if (refExprIsType) return false
+                // If current declaration is a function, we only offer names of unresolved function calls. For property declarations, we
+                // always offer an unresolved usage because a property may be called if the object has an `invoke` method.
+                if (!refExprIsCallRef && !refExprIsCall && symbol is KtFunctionLikeSymbol) return false
```
fedochet, line 81: I think that those three booleans make the code hard to understand

It is not clear if they are mutually exclusive (they probably are), and how the different combinations are handled

Maybe it would be better to do it like this?

```kotlin
when {
  refExprParent is KtUserType -> { ... }
  refExprParent is KtCallExpression -> { ... }
}
```

This way, those three scenarios would be cleanly separated from each other

Also, it would be clear if we want to do something at all if none of the booleans is true

If such refactoring turns out too tedious, you can leave it like that

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/Completions.kt
```diff
@@ -24,6 +24,9 @@ internal object Completions {
                 complete(factory.classifierContributor(0), positionContext)
                 complete(factory.keywordContributor(1), positionContext)
                 complete(factory.packageCompletionContributor(2), positionContext)
+                // For `val` and `fun` completion. For example, with `val i<caret>`, the fake file contains `val iX.f`. Hence a
```
fedochet, line 27: I am not sure that this commentary is needed here (at least the part about the `iX.f`)

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirDeclarationFromUnresolvedNameContributor.kt
```diff
@@ -0,0 +1,142 @@
...
+    private fun getReceiverType(symbol: KtCallableSymbol): KtType? {
+        return symbol.receiverType?.type ?: (symbol as? KtPossibleMemberSymbol)?.dispatchType
+    }
+
+
+    private fun PsiElement.getCurrentDeclarationAtCaret(): KtNamedDeclaration? {
+        return when (val parent = parent) {
+            is KtNamedDeclaration -> parent
+            is KtNameReferenceExpression -> {
+                // In a fake completion file, a property or function under construction is appended with placeholder text "X.f$", which is
```
fedochet, line 120: I suggest you to reference the code that selects the dummy identifier in the completion here, so the reader won't have to look it up by himself - it can be something like `see <method> for the details`

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/weighers/PriorityWeigher.kt
```diff
@@ -0,0 +1,16 @@
...
+
+package org.jetbrains.kotlin.idea.completion.weighers
+
+import com.intellij.codeInsight.lookup.LookupElement
+import com.intellij.codeInsight.lookup.LookupElementWeigher
+import org.jetbrains.kotlin.idea.completion.ItemPriority
+import org.jetbrains.kotlin.idea.completion.priority
+
+internal object PriorityWeigher {
+    const val WEIGHER_ID = "kotlin.priority"
```
fedochet, line 11: Do we need to have this property public?

plugins/kotlin/frontend-independent/tests/test/org/jetbrains/kotlin/test/utils/IgnoreTests.kt
```diff
@@ -282,6 +290,7 @@ object IgnoreTests {
                 firTestFile.delete()
                 originalTestFile.prependFirIdentical()
                 additionalFileToMarkFirIdentical?.prependFirIdentical()
+                if (additionalFileToDeleteIfIdentical?.exists() == true) additionalFileToDeleteIfIdentical.delete()
```
fedochet, line 293: I believe you can just call `?.delete()` here without check

fedochet (changes requested):
Some minor comments and questions

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/Completions.kt
tgeng: Hmm, I think this comment is useful. Previously Ilya was asking why we need to complete unresolved name in a type reference. So I added a comment explaining how the placeholder causes the parser to treat the current caret position as a type reference. If the placeholder text does not contain `.`, we would not need this contributor here.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/weighers/PriorityWeigher.kt
tgeng: I don't need it now, But it could potentially be used if we want to rearrange the weigers in https://github.com/JetBrains/intellij-community/blob/838bf7cbdcf0e279efc218b832345efe48e16b56/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/weighers/Weighers.kt#L11. If you like I can change all the current weigers to have this piece private unless it's used in `Weighers.kt`.

plugins/kotlin/frontend-independent/tests/test/org/jetbrains/kotlin/test/utils/IgnoreTests.kt
tgeng: Thanks! Just tried and indeed it's safe to delete an nonexistent file.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirDeclarationFromUnresolvedNameContributor.kt
tgeng: Thanks for the suggestion! Let me use when and inline these booleans. Hopefully it's clearer now.

PR #1690, opened 2021-08-28
FIR UAST: PsiType conversion for erroneous cases and primitives

jsjeon:
My first trial to mitigate the type difference between `void` (in FE 1.0 UAST) and `Unit` (in FIR UAST) wasn't successful: https://github.com/JetBrains/kotlin/pull/4507#discussion_r674710777

It turns out that FE 1.0 LC copied `KotlinType`-to-`PsiType` conversion from FE 1.0 UAST and _post-processed_ ([comment](https://github.com/JetBrains/kotlin/blob/master/compiler/light-classes/src/org/jetbrains/kotlin/asJava/classes/ultraLightUtils.kt#L160)): it's simply using `KotlinTypeMapper` and `JvmSignatureWriter`. FIR LC mimics the same behavior (which is now part of HL API), and thus my previous attempt of changing type conversion in FIR LC only resulted in FIR LC test failures.

That is,
1) type conversions (to `PsiType`) in FE 1.0 LC and FE 1.0 UAST diverge, and
2) type conversions (to `PsiType`) in FE 1.0 LC and FIR LC are quite similar, e.g.,
* FE 1.0 LC: https://github.com/JetBrains/kotlin/blob/master/compiler/light-classes/src/org/jetbrains/kotlin/asJava/classes/ultraLightUtils.kt#L258
* FIR LC (via HL API): https://github.com/JetBrains/kotlin/blob/master/idea/idea-frontend-fir/src/org/jetbrains/kotlin/idea/frontend/api/fir/components/KtFirPsiTypeProvider.kt#L94

so, the reasonable approach is to make the type conversion in FIR UAST and that in FE 1.0 UAST match (instead of touching FIR LC side).

There are two notable differences:
1) `UastErrorType` as a default type for erroneous cases
2) ad-hoc type conversion for primitives, including `Unit`, which made huge yet trivial type differences

For the former, we need HL API change ([KOTLIN-CR-179](https://kotlin.jetbrains.space/p/kotlin/reviews/179/)) so that we can pass `UastErrorType` as a default error type. For the latter, places of type conversion to `PsiType` in FIR UAST are consolidated, and there, a similar shortcut is added.


plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
```diff
@@ -42,6 +43,27 @@ internal fun KtAnalysisSession.toPsiMethod(ktCall: KtCall): PsiMethod? {
     }
 }
 
+internal fun KtAnalysisSession.toPsiType(ktType: KtType, context: KtElement): PsiType {
+    if (ktType is KtNonErrorClassType && ktType.typeArguments.isEmpty()) {
+        // TODO: unclear when we need boxed type.
```
darthorimar, line 48: @xiexed could you please help here?

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
```diff
@@ -42,6 +43,27 @@ internal fun KtAnalysisSession.toPsiMethod(ktCall: KtCall): PsiMethod? {
...
+        val psiType = when (ktType.classId.asFqNameString()) {
+            "kotlin.Int" -> PsiType.INT
+            "kotlin.Long" -> PsiType.LONG
+            "kotlin.Short" -> PsiType.SHORT
+            "kotlin.Boolean" -> PsiType.BOOLEAN
+            "kotlin.Byte" -> PsiType.BYTE
+            "kotlin.Char" -> PsiType.CHAR
+            "kotlin.Double" -> PsiType.DOUBLE
+            "kotlin.Float" -> PsiType.FLOAT
+            "kotlin.Unit" -> PsiType.VOID
```
darthorimar, line 58: We should not always convert `Unit` to `void`. It should happen only for function return types

darthorimar, line 58: Maybe let's use `org.jetbrains.kotlin.name.StandardClassIds` instead of hardcoding class ids?

darthorimar (changes requested):

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
jsjeon: But the point is, I didn't see any places/outputs where boxed primitives are used even for FE 1.0 UAST.

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
jsjeon: Good point. Changed both UAST plugins.

jsjeon: Okay, I think it's time to fix FE 1.0 UAST as well. See the latest commit.

plugins/kotlin/uast/uast-kotlin-base/src/org/jetbrains/uast/kotlin/internal/baseKotlinInternalUastUtils.kt
```diff
@@ -75,6 +75,9 @@ fun KtClassOrObject.toPsiType(): PsiType {
         PsiTypesUtil.getClassType(lightClass)
 }
 
+val KtElement.isFunctionWithImplicitUnitReturnType: Boolean
+    get() = this is KtNamedFunction && this.hasBlockBody() && !this.hasDeclaredReturnType()
```
jsjeon, line 79: FYI, the counterpart is here: https://github.com/JetBrains/kotlin/blob/master/compiler/light-classes/src/org/jetbrains/kotlin/asJava/classes/ultraLightMembersCreator.kt#L231

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
darthorimar: There are some usages of https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin/src/org/jetbrains/uast/kotlin/internal/kotlinInternalUastUtils.kt#L107 which has `boxed = true`. Type arguments and nullable types are always boxed, which seems quite reasonable.

jsjeon: I see; I noticed that I've ignored that bit while commonizing many other expression kinds. Revisited; added a flag to the base service; and updated API use places accordingly if boxed primitives are required, such as receivers and type arguments.

jsjeon:
Changed this to depend on #1685 since it adds receiver type retrieval for accessor call. Also, boxed primitive changes should be on top of it.

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
```diff
@@ -56,7 +57,9 @@ internal fun KtAnalysisSession.toPsiType(ktType: KtType, context: KtElement): Ps
             StandardClassIds.Char -> PsiType.CHAR
             StandardClassIds.Double -> PsiType.DOUBLE
             StandardClassIds.Float -> PsiType.FLOAT
-            StandardClassIds.Unit -> PsiType.VOID
+            StandardClassIds.Unit -> {
+                if (context.isFunctionWithImplicitUnitReturnType) PsiType.VOID else null
```
darthorimar: If we are using context like this, are we sure that context is always a target declaration? By the code it looks like that context is some random `PsiElement` which we use to get the module in which it lives

jsjeon: This `context` is where the given `KtType` for FIR UAST (or `KotlinType` for FE 1.0 UAST) is retrieved. Therefore, if this `context` is a function with implicit `Unit` return type, it came from either
* https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin/src/org/jetbrains/uast/kotlin/KotlinUastResolveProviderService.kt#L261 or
* https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/FirKotlinUastResolveProviderService.kt#L278

where the `context` is the declaration of interest.

I added a test to verify behaviors; see my other comments.

plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastApiTestBase.kt
```diff
@@ -61,4 +65,21 @@ interface UastApiTestBase : UastPluginSelection {
...
+            ?: throw IllegalStateException("Target function not found at ${uFile.asRefNames()}")
+        // type delegated from LC
+        TestCase.assertEquals(PsiType.VOID, m.returnType)
+        // type through the base service
+        val service = ServiceManager.getService(BaseKotlinUastResolveProviderService::class.java)
+        TestCase.assertEquals(PsiType.VOID, service.getType(m.sourcePsi as KtDeclaration, m as UElement))
+
+        val functionCall = m.findElementByText<UElement>("println").uastParent as KotlinUFunctionCallExpression
+        // type through the base service: KotlinUElementWithType#getExpressionType
+        TestCase.assertEquals("PsiType:Unit", functionCall.getExpressionType()?.toString())
```
jsjeon, line 82: This return type of function call expression _was_ converted to `void` without this change.

plugins/kotlin/uast/uast-kotlin-base/src/org/jetbrains/uast/kotlin/internal/baseKotlinInternalUastUtils.kt
jsjeon: Even if a function has explicit `Unit` return type, the generated bytecode has a signature with `void` return type, so I believe even FE 1.0 LC modeling is wrong. I'll drop this part and just convert `Unit` to `void` for function return type. I'll also prepare how to test these, like actual compiler behavior v.s. LC behavior (and of course FIR LC), in general.

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
darthorimar: Let's not relay that the `context` is always the declaration type came from. Some new usages of `toPsiType` may be added which breaks this rule. If you would like to use `context` like this, I would suggest renaming and changing it type to `declaration: KtDeclaration` so we ensure that we always pass owner declaration. Or maybe provide some additional enum/flag specifying type conversion mode for `Unit`.

jsjeon: These type conversions are used at various places, so it's not trivial to change the type of `context` to `KtDeclaration`. Instead, I prefer the idea of an additional enum (named `TypeOwnerKind`) to specify the kind of type owner, which may be different from the `context` of this type conversion. Btw, sometimes this `context` is used to retrieve the project, module, etc. as you mentioned.

In FIR UAST, type owner is always that `context`, so quite straightforward. In FE 1.0 UAST, it's quite similar for many cases, but for descriptor-based fake declaration or declaration from deserialization, it's unclear whether the `context` is really the owner of the type to be converted. Hence, hard-coded `DECLARATION` type owner kind for those cases.

darthorimar: I meant something like 

```
KtAnalysisSession.toPsiType(
    ktType: KtType,
    context: KtElement,
    typeConversionMode: TypeConversionMode,
    boxed: Boolean = false
): PsiType {
  ...
   if (typeConversionMode.convertUnitToVoid) PsiType.VOID.orBoxed() else null
   ...
}

enum class TypeConversionMode(val convertUnitToVoid: Boolean) {
	...
}
```

But your solution works well too :)

darthorimar (approved):

darthorimar:
Looks like, it depends on https://github.com/JetBrains/intellij-community/pull/1685/ . I will merge this pr together with https://github.com/JetBrains/intellij-community/pull/1700 after https://github.com/JetBrains/intellij-community/pull/1685/ is merged

darthorimar:
Merged, thank you!

PR #1693, opened 2021-08-29
[maven] IDEA-253667: detect if "--quiet" maven build was successful

bdshadow:
There is no other way other than relying on the exit code of command, which is printed to the console by ProcessTerminatedListener

Signed-off-by: Dmitrii Bocharov <bdshadow@gmail.com>


nskvortsov:
Thank you for the pull request. The issue was fixed on master in a bit another way (by properly handling Maven's SessionEnd event) here https://github.com/JetBrains/intellij-community/commit/c3500485e9d20d1da25184f4e0b10b4b0d3f85c6

PR #1695, opened 2021-08-30
[kotlin] Fix AddNamesInCommentToJavaCallArgumentsIntention

t-kameyama:
- Disable inlay hints after adding comments
- Don't suggest when there are type mismatch arguments


vladimirdolzhenko:
Nice one, while I I meant different problem:

There is a java method like `public void foo(int index){}` and it is commented in Kotlin code like `foo(/* index = */ 1`.
After that somebody came and renamed parameter as e.g. `idx` -> therefore kotlin name becomes inconsistent with a target.

ps. Do not worry on p.1 - I'm actually working on hiding inlay hints for java-commented parameters.

vladimirdolzhenko:
https://github.com/JetBrains/intellij-community/commit/fef9437a4ad2ece19c3c4de58edb0b34cac977e9 covers hiding inlay hints for java-commented parameters.

PR #1698, opened 2021-08-31
KTIJ-19518 'Convert to anonymous function' is not suggested for lambda with `_` argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19518

### DIFF
--- a/plugins/kotlin/idea/tests/testData/intentions/lambdaToAnonymousFunction/underscoreParameter.kt.after
+++ b/plugins/kotlin/idea/tests/testData/intentions/lambdaToAnonymousFunction/underscoreParameter.kt.after
@@ -2,5 +2,7 @@
 fun foo(f: (Int, String, Int) -> String) {}
 
 fun test() {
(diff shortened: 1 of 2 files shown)


plugins/kotlin/idea/tests/testData/intentions/lambdaToAnonymousFunction/underscoreParameter.kt.after
```diff
@@ -0,0 +1,8 @@
+// IS_APPLICABLE: true
+fun foo(f: (Int, String, Int) -> String) {}
+
+fun test() {
+    foo(fun(i1: Int, s: String, i: Int): String {
```
dimonchik0036, line 5: I think we shouldn't suggest names, so we should leave `_` as it is.
Expected result:
```kotlin
foo(fun(_: Int, _: String, i: Int): String {
    return ""
})
```

dimonchik0036 (changes requested):

dimonchik0036 (approved):

PR #1703, opened 2021-09-03
[KTIJ-16942] Prioritize ExpressionPSI name over valueArgument name

XiangRongLin:
https://youtrack.jetbrains.com/issue/KTIJ-16942


nikitabobko:
Close without merge since original issue was declined on YouTrack

PR #1702, opened 2021-09-03
UAST: test updates (default value param name in default setter)

jsjeon:
This is simple test update for [KOTLIN-CR-184](https://kotlin.jetbrains.space/p/kotlin/reviews/184/).

PR #1704, opened 2021-09-04
UAST: test updates (avoid @null nullability annotations)

jsjeon:
Accumulated on top of #1702, and only last two commits are additional.

The first one is simple test data update for [KOTLIN-CR-185](https://kotlin.jetbrains.space/p/kotlin/reviews/185/). The second one is minor improvement while looking at nullability annotation.


darthorimar:
Merged, thank you!

PR #1706, opened 2021-09-05
IDEA-254311 Use Windows File Explorer for displaying directory content on WSL when xdg-open is unavailable

zeko868:
Currently IDEA-Intellij reports the error message *This action isn't supported on the current platform* when it is installed and running inside of WSL that has no installed file explorer (like Nemo or Dolphin) and user attempts to view directory content in the default file manager (e.g. using *Show in File Explorer* action).

With this fix the directory content should be displayed in Windows File Explorer since it allows to browse file-system of WSL. This applies for both major versions of WSL.
In case that user already has set-up file explorer on WSL, *xdg-open* will still be used for displaying directory content.

https://youtrack.jetbrains.com/issue/IDEA-254311

### DIFF
--- a/platform/ide-core/src/com/intellij/openapi/fileTypes/NativeFileType.java
+++ b/platform/ide-core/src/com/intellij/openapi/fileTypes/NativeFileType.java
@@ -11,6 +11,9 @@ import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
 
 import javax.swing.*;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


trespasserw:
Sorry, I was sure I have already answered :(

If you're still interested in it, I'd suggest implementing it as a separate action, complimentary to the existing [`RevealFileAction`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/ide/actions/RevealFileAction.java) - it would be available for every file and directory and not only native file types, and users of GUI-capable distributions would have a choice.

PR #1707, opened 2021-09-07
FIR UAST: test updates (modality of enum class)

jsjeon:
It's on top of #1704 and only the last commit is updating test data according to [KOTLIN-CR-187](https://kotlin.jetbrains.space/p/kotlin/reviews/187/).


darthorimar:
Merged, thank you!

PR #1708, opened 2021-09-07
FIR IDE: KotlinHighLevelFunctionParameterInfoHandler follow-ups

punzki:
1. Strikeout deprecated candidates.
2. Disabled candidates when there are arguments before the current argument that don't match the corresponding parameter type (i.e., type mismatch). Added more tests.
3. Properly handle `vararg` parameter arguments. The computation for which parameter to highlight did not account for multiple arguments for `vararg` parameters, which was pretty obvious in retrospect. Added more tests. Requires [KOTLIN-MR-188](https://kotlin.jetbrains.space/p/kotlin/reviews/188/files).

### DIFF
--- a/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
+++ b/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
@@ -49,7 +49,7 @@ class KotlinHighLevelLambdaParameterInfoHandler :
 
     override fun getArgumentListAllowedParentClasses() = setOf(KtLambdaArgument::class.java)
 
-    override fun getParameterIndex(context: UpdateParameterInfoContext, argumentList: KtLambdaArgument): Int {
+    override fun getCurrentArgumentIndex(context: UpdateParameterInfoContext, argumentList: KtLambdaArgument): Int {
         val size = (argumentList.parent as? KtCallElement)?.valueArguments?.size ?: 1
         return size - 1
     }
@@ -131,8 +131,8 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
         if (context.parameterOwner !== argumentList) {
             context.removeHint()
         }
-        val parameterIndex = getParameterIndex(context, argumentList)
-        context.setCurrentParameter(parameterIndex)
+        val currentArgumentIndex = getCurrentArgumentIndex(context, argumentList)
+        context.setCurrentParameter(currentArgumentIndex)
 
         val callElement = argumentList.parent as? KtElement ?: return
         analyse(callElement) {
@@ -191,6 +191,20 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
                     }
                 }
 
+                var hasTypeMismatchBeforeCurrent = false
+                for ((index, argument) in arguments.withIndex()) {
+                    if (index >= currentArgumentIndex) break
+                    val parameterForArgument = argumentMapping[argument] ?: continue
+                    if (parameterForArgument == setValueParameter) continue
+
+                    val argumentType = argument.getKtType() ?: error("Argument should have a KtType")
+                    val parameterType = parameterForArgument.annotatedType.type
+                    if (argumentType.isNotSubTypeOf(parameterType)) {
+                        hasTypeMismatchBeforeCurrent = true
+                        break
+                    }
+                }
+
                 // TODO: This should be changed when there are multiple candidates available; need to know which one the call is resolved to
                 val isCallResolvedToCandidate = candidates.size == 1
 
@@ -201,13 +215,14 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
                     argumentToParameterIndex,
                     valueParameters.size,
                     parameterIndexToText,
-                    isCallResolvedToCandidate
+                    isCallResolvedToCandidate,
+                    hasTypeMismatchBeforeCurrent
                 )
             }
         }
     }
 
-    protected open fun getParameterIndex(context: UpdateParameterInfoContext, argumentList: TArgumentList): Int {
+    protected open fun getCurrentArgumentIndex(context: UpdateParameterInfoContext, argumentList: TArgumentList): Int {
         val offset = context.offset
         return argumentList.allChildren
             .takeWhile { it.startOffset < offset }
@@ -280,7 +295,8 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
      *
      * `setupUIComponentPresentation()` is called with `disabled = true` when any of the following are true:
      * 1. If the argument on the cursor does NOT map to a parameter (e.g., on N-th argument but there are < N parameters),
-     * 2. If any of the arguments before the cursor do NOT match the type of the corresponding parameter.
+     * 2. If any of the arguments before the cursor do NOT match the type of the corresponding parameter (ignoring arguments with type
+     *    errors, e.g., unresolved) or do NOT map to a parameter (e.g., named argument with an unknown name).
      * 3. If the cursor is after a trailing comma with no argument, AND LanguageFeature.TrailingCommas is DISABLED, AND there are
      *    already enough arguments in the call. (We assume the user is about to enter an argument in that position.)
      *
@@ -298,7 +314,8 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
         if (currentArgumentIndex < 0) return false
 
         val callInfo = itemToShow.callInfo ?: return false
-        val (callElement, valueArguments, arguments, argumentMapping, valueParameterCount, parameterIndexToText, isCallResolvedToCandidate) = callInfo
+        val (callElement, valueArguments, arguments, argumentToParameterIndex, valueParameterCount, parameterIndexToText,
+            isCallResolvedToCandidate, hasTypeMismatchBeforeCurrent) = callInfo
 
         val supportsMixedNamedArgumentsInTheirOwnPosition =
             callElement.languageVersionSettings.supportsFeature(LanguageFeature.MixedNamedArgumentsInTheirOwnPosition)
@@ -316,6 +333,7 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
         var highlightStartOffset = -1
         var highlightEndOffset = -1
         var isDisabledBeforeHighlight = false
+        var hasUnmappedArgumentBeforeCurrent = false
         val usedParameterIndices = HashSet<Int>()
         val text = buildString {
             var namedMode = false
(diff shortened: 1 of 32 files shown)


plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -191,6 +191,37 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
...
+                        // for that same vararg parameter.
+                        val parameterForLastArgument = argumentMapping[arguments.last()]
+                        if (parameterForLastArgument?.isVararg == true) {
+                            parameterToIndex[parameterForLastArgument]
+                        } else {
+                            null
+                        }
+                    }
+                    else -> null
+                }
```
darthorimar, line 209: It's better to extract calculating of `highlightParameterIndex` into a function

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -191,6 +191,37 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
...
+                    val parameterForArgument = argumentMapping[argument] ?: continue
+                    if (parameterForArgument == setValueParameter) continue
+
+                    val argumentType = argument.getKtType() ?: error("Argument should have a KtType")
+                    val parameterType = parameterForArgument.annotatedType.type
+                    if (argumentType.isNotSubTypeOf(parameterType)) {
+                        hasTypeMismatchBeforeCurrent = true
+                        break
+                    }
+                }
```
darthorimar, line 223: It's better to extract calculating of `hasTypeMismatchBeforeCurrent` into a function

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -298,7 +332,8 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
         if (currentArgumentIndex < 0) return false
 
         val callInfo = itemToShow.callInfo ?: return false
-        val (callElement, valueArguments, arguments, argumentMapping, valueParameterCount, parameterIndexToText, isCallResolvedToCandidate) = callInfo
+        val (callElement, valueArguments, arguments, argumentToParameterIndex, valueParameterCount, parameterIndexToText,
```
darthorimar, line 335: Do we really need such big destructing declaration? It is easy to accidentally mix up some parameters. What do you think about extracting the following function body intro a separate function with `callinfo` as receiver or using `with(callInfo) { .. }` syntax here?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -115,7 +115,7 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
             }
 
             val candidates = resolvedCall?.targetFunction?.candidates ?: return null
-            context.itemsToShow = candidates.map { CandidateInfo(it.createPointer()) }.toTypedArray()
+            context.itemsToShow = candidates.map { CandidateInfo(it.createPointer(), it.deprecationStatus != null) }.toTypedArray()
```
tgeng, line 118: Please consider hiding the entry if the deprecation level is `HIDDEN`. Looks like FE1.0 doesn't show any deprecated member of this level.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -191,6 +191,20 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
...
 
+                var hasTypeMismatchBeforeCurrent = false
+                for ((index, argument) in arguments.withIndex()) {
+                    if (index >= currentArgumentIndex) break
+                    val parameterForArgument = argumentMapping[argument] ?: continue
+                    if (parameterForArgument == setValueParameter) continue
+
+                    val argumentType = argument.getKtType() ?: error("Argument should have a KtType")
+                    val parameterType = parameterForArgument.annotatedType.type
+                    if (argumentType.isNotSubTypeOf(parameterType)) {
```
tgeng, line 219: Does this work for the following code? (I am having some build issue locally and so far have not been able to test this 😢)
```
fun <T> takeT(t: MutableList<T>, i: Int) {
}

fun get(): MutableList<out String> {
    return mutableListOf()
}

fun test() {
    takeT(get(), <caret>)
}
```

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: If deprecation level is `HIDDEN`, in FE 1.0, it comes out as `UNRESOLVED_REFERENCE` (the candidate is hidden for all practical purposes) but in FIR it's `INVISIBLE_REFERENCE` with a candidate and hence we show it in the popup. There are a few ways to fix this:

1. Filter out the candidate here in the handler.
2. Filter out candidates from `ConeHiddenCandidateError` in HL API.
3. Change the diagnostic to `UNRESOLVED_REFERENCE` for `HIDDEN` deprecation level.

My first inclination is to go with the last option since that matches FE 1.0 behavior. Will ask offline on whether we should do that or not.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: No it does not work correctly. I guess we need to substitute the type parameters then. :slightly_smiling_face:

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: Using `with(callInfo)` is a good idea, thanks for the suggestion!

plugins/kotlin/idea/tests/testData/parameterInfo/functionCall/NotGreen.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_FIR
```
punzki, line 1: This was only coincidentally passing before. When there are multiple candidates, we don't get the argument mapping. We use it more now with these changes, so this test is now failing.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: Updated KOTLIN-MR-188 so that candidates from `ConeHiddenCandidateError` are not returned. Also added more tests here.

plugins/kotlin/idea/tests/testData/parameterInfo/functionCall/UnmappedCurrent.kt
```diff
@@ -0,0 +1,8 @@
+fun m(x: Int, y: Boolean) = 2
+
+fun d() {
+    m(y = false, <caret>unmapped = false)
+}
+/*
+Text: (<disabled>[y: Boolean],</disabled><highlight> </highlight>[x: Int]), Disabled: false, Strikeout: false, Green: true
```
jsjeon, line 7: Just curious how to interpret `<highlight> </highlight>` (v.s., e.g., no highliht)?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
```diff
@@ -447,7 +468,8 @@ abstract class KotlinHighLevelParameterInfoWithCallHandlerBase<TArgumentList : K
         val allParametersUsed = usedParameterIndices.size == valueParameterCount
         val supportsTrailingCommas = callElement.languageVersionSettings.supportsFeature(LanguageFeature.TrailingCommas)
         val afterTrailingComma = arguments.isNotEmpty() && currentArgumentIndex == arguments.size
-        val tooManyArgs = allParametersUsed && ((!supportsTrailingCommas && afterTrailingComma) || arguments.size > valueParameterCount)
+        val isInPositionToEnterArgument = !supportsTrailingCommas && afterTrailingComma
+        val tooManyArgs = allParametersUsed && (isInPositionToEnterArgument || hasUnmappedArgument)
```
jsjeon: Just to double-check, replacement with `hasUnmappedArgument` is equivalent to the previous logic---`arguments.size > valueParameterCount` only if it's `&&`-ed with `allParametersUsed` I guess? Well, if `tooManyArgs` is only used for `isDisabled`, having any unmapped argument is a good qualifier(?), but it seems not sufficient to declare there are too many args (unless all parameters are used).

jsjeon (commented):
Very minor comments.

plugins/kotlin/idea/tests/testData/parameterInfo/functionCall/UnmappedCurrent.kt
punzki: I copied FE 1.0 behavior in certain cases, see comment in: https://github.com/JetBrains/intellij-community/blob/d94d5dc655ae54f2552c84f6ddebb906daf9a864/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt#L334-L343

I find it a little confusing, but my understanding is in order for `isDisabledBeforeHighlight` to work, you have to highlight _something_ even if it's just the empty space. My interpretation of this presentation is that `[y: Boolean]` is "disabled" because it's already used.

I'll update the comment above. When I first wrote it, I didn't understand why the presentation was that way.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: I think I understand the first part of your comment... Yes, I'm replacing `arguments.size > valueParameterCount` with `hasUnmappedArgument`. The former doesn't work when there are multiple arguments for a `vararg` parameter.

But I'm not sure I understand the second part of your comment "Well if `tooManyArgs`..." Can you clarify? There are multiple conditions for `isDisabled`, having too many args is just one of them. You can only have "too many args" if all parameters are used.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinHighLevelFunctionParameterInfoHandler.kt
punzki: Updated this PR so we're filtering candidates hidden due to visibility here on the client. Left handling of `@Deprecated(HIDDEN)` to a future change.

darthorimar:
Merged manually, thank you!

PR #1709, opened 2021-09-07
Handle column selection using SHIFT

Fabrice-TIERCELIN:
Hi @JB-Dmitry,

Currently, in column mode (Ctrl+Shift+Insert), if you click and SHIFT-click, in the editor, it makes a text selection, which is not logical. This PR makes a column selection instead.

This PR works but it can be improved. If the first click is beyond the text, it will use the last character of the line and will create the selection from this point. Also, if you make a selection when a selection already exists, it starts the selection from the opposite point of the second click of the last selection.

Those two points are not blocking but it would be better to fix them too.


platform/platform-impl/src/com/intellij/openapi/editor/impl/EditorImpl.java
```diff
@@ -4085,19 +4085,42 @@ else if (e.getSource() != myGutterComponent && isCreateRectangularSelectionEvent
...
             if (mySelectionModel.isUnknownDirection() && caretOffset > startToUse) {
               startToUse = Math.min(oldStart, oldEnd);
             }
+          }
+          if (isColumnMode()) {
+            final CaretModel caretModel = getCaretModel();
+            int selStart = oldStart;
+            int selEnd = oldEnd;
+            LogicalPosition blockStart;
+            if (caretModel.supportsMultipleCarets()) {
```
JB-Dmitry, line 4108: We know for sure that our caret model supports multiple carets. This should be simplified.

platform/platform-impl/src/com/intellij/openapi/editor/impl/EditorImpl.java
```diff
@@ -4085,19 +4085,42 @@ else if (e.getSource() != myGutterComponent && isCreateRectangularSelectionEvent
...
+              LogicalPosition logicalSelStart = offsetToLogicalPosition(selStart);
+              LogicalPosition logicalSelEnd = offsetToLogicalPosition(selEnd);
+              blockStart = selStart == caretOffset ? logicalSelEnd : logicalSelStart;
+            }
+            else {
+              blockStart = selStart == caretModel.getOffset()
+                           ? caretModel.getLogicalPosition()
+                           : offsetToLogicalPosition(selStart);
+            }
+            @NotNull LogicalPosition blockEnd = caretOffset == caretModel.getOffset()
```
JB-Dmitry, line 4118: `caretOffset == caretModel.getOffset()` is always true here.

platform/platform-impl/src/com/intellij/openapi/editor/impl/EditorImpl.java
```diff
@@ -4085,19 +4085,42 @@ else if (e.getSource() != myGutterComponent && isCreateRectangularSelectionEvent
...
+          }
+          if (isColumnMode()) {
+            final CaretModel caretModel = getCaretModel();
+            int selStart = oldStart;
+            int selEnd = oldEnd;
+            LogicalPosition blockStart;
+            if (caretModel.supportsMultipleCarets()) {
+              LogicalPosition logicalSelStart = offsetToLogicalPosition(selStart);
+              LogicalPosition logicalSelEnd = offsetToLogicalPosition(selEnd);
+              blockStart = selStart == caretOffset ? logicalSelEnd : logicalSelStart;
```
JB-Dmitry, line 4111: Why don't we calculate `blockStart` based on `startToUse` prepared above?

JB-Dmitry (commented):
Please add a test for the new functionality to `EditorImplTest`. You can find other examples of tests for mouse handling logic there.

JB-Dmitry:
The change is OK now. Waiting for the test to be added.

Fabrice-TIERCELIN:
I think I will have time for the tests this week-end.

Fabrice-TIERCELIN:
Despite I haven't coded the tests yet, I have fixed the both issues 🙂 The selection always starts from the last location you have clicked in. If a column selection was done, it starts from the corner that finished the column selection. If there was no selection, even if you have clicked in an empty location, it will start the selection from this location.

JB-Dmitry:
> The selection always starts from the last location you have clicked in.

I would argue that this is not an expected behavior - the selection shouldn't start from the last location one had Shift+clicked at. It doesn't match current behavior in normal (not column) mode, and I believe it works that way for a reason now. Suppose one has mistakenly Shift+clicked on a wrong character. With current behavior, one can easily correct that. With the new logic, one would need to start from scratch.

JB-Dmitry (changes requested):

JB-Dmitry:
The change has been pushed to master. It should be available in next week's 2021.3 EAP builds.
Thanks for your contribution.

PR #1700, opened 2021-09-03
FIR UAST: test updates (property accessor annotations)

jsjeon:
This is simple test data update for [KOTLIN-CR-183](https://kotlin.jetbrains.space/p/kotlin/reviews/183/). It's atop #1690 and only the last commit matters.

PR #1710, opened 2021-09-07
FIR/UAST: commonize expression conversion

jsjeon:
This is on top of #1707, and only the last three commits are additional ones.

As the PR title indicates, this is about commonizing expression conversion in FIR/UAST. To do so, we need to fully commonize remaining two expression kinds: array access and binary expression. The first one needs a special type resolution in FE 1.0 UAST side, and it's integrated into `getType` API. The second one needs a special call resolution for bitwise operators, and a new API is added to the base service to abstract that. (It looks like both parts are slightly wrong in FIR UAST, but those could be addressed in a follow-up, separate PR.)

After that, the entire logic to convert expressions is same, and thus commonized. As a result, literally all conversion logics are commonized, i.e., each plugin's converter object is really simple, without its own conversion logic at all! This commonization ensures that there would be no differences in UAST tree structures from FE 1.0 v.s. FIR UAST. (Note that we already entered a phase where only subtle differences came from FE 1.0/FIR LC, not UAST. This just proves that's the case.)


jsjeon:
FYI: this will complete https://kotlin.jetbrains.space/p/fir-ide/issues/27

plugins/kotlin/uast/uast-kotlin-fir/testData/legacyRenderLog/ParameterPropertyWithAnnotation.log.fir.txt
```diff
@@ -8,9 +8,7 @@ UFile (package = )
         UField (name = bar)
             UAnnotation (fqName = org.jetbrains.annotations.NotNull)
         UMethod (name = getBar)
-            UAnnotation (fqName = MyAnnotation)
```
xiexed, line 11: Hoping it will not break users code

xiexed (approved):

plugins/kotlin/uast/uast-kotlin-fir/testData/legacyRenderLog/ParameterPropertyWithAnnotation.log.fir.txt
jsjeon: No worries. I guess you left a comment on this one: https://github.com/JetBrains/intellij-community/pull/1700/commits/36ea9a579beb14a61c4e1004849ec65329aaddcc#diff-d857157f41e597027576e1886cbb8ec73f06cd51ffa0f720c1947865a82bbf0d where annotation on property accessor is discarded if no user-site is specified (from FIR LC side: https://kotlin.jetbrains.space/p/kotlin/reviews/183/). This matches FE 1.0 LC: https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin-fir/testData/legacyRenderLog/ParameterPropertyWithAnnotation.log.fe10.txt#L10

After that, with one more fix in FE 1.0 LC side this time (avoiding `@null` nullability annotation: https://kotlin.jetbrains.space/p/kotlin/reviews/185/), for this test case---`ParameterPropertyWithAnnotation`, render/log from FE 1.0 UAST and FIR UAST are identical: https://github.com/JetBrains/intellij-community/pull/1704/commits/67c8d0d744217a93712127a3c748da523e3d871e#diff-2e6af7a20b33645b7c056f567552ff90fea1ccdd8594462f737633f1924b7820

darthorimar:
Merged, thank you!

PR #1717, opened 2021-09-10
KTIJ-11380 Live template for "iter" generates illegal $VAR$ template variable name `object` for `Iterable<Object>`

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-11380

### DIFF
--- a/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/KotlinNameSuggester.kt
+++ b/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/KotlinNameSuggester.kt
@@ -100,7 +100,7 @@ object KotlinNameSuggester {
         val result = LinkedHashSet<String>()
 
         suggestNamesByExpressionOnly(collection, bindingContext, { true })
-            .mapNotNull { StringUtil.unpluralize(it) }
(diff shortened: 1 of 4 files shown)


nikitabobko (approved):
Additional patch:
```
diff --git community/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/KotlinNameSuggester.kt community/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/KotlinNameSuggester.kt
index 41f6f2a8c5f67..4900f5f9a3dd2 100644
--- community/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/KotlinNameSuggester.kt
+++ community/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/KotlinNameSuggester.kt
@@ -100,7 +100,8 @@ object KotlinNameSuggester {
         val result = LinkedHashSet<String>()
 
         suggestNamesByExpressionOnly(collection, bindingContext, { true })
-            .mapNotNull { name -> StringUtil.unpluralize(name).takeUnless { it.isKeyword() } }
+            .mapNotNull { name -> StringUtil.unpluralize(name)}
+            .filter { name -> !name.isKeyword() }
             .mapTo(result) { suggestNameByName(it, validator) }
 
         result.addNamesByType(elementType, validator)
```

PR #1719, opened 2021-09-10
FIR/UAST: commonize multi/resolve/api tests

jsjeon:
In addition to tests that compare generated golden files regarding UAST render, log, identifier, reference name, types, or values, there are tests that directly call UAST APIs regarding (multi) resolutions, constant evaluation, parameter lookup, etc. This PR commonizes such tests so that we can check if API behavior/results from FIR UAST match those from FE 1.0 UAST.

This is kind of mechanical refactoring. For those failing tests in FIR UAST, I just left several TODOs with shallow investigations. On the bright side, we have more passing tests/assertions than failing ones. :) Also, these help us (or me) identify what to do next.


plugins/kotlin/uast/uast-kotlin-fir/test/org/jetbrains/kotlin/idea/fir/uast/FirUastApiTest.kt
```diff
@@ -33,6 +33,146 @@ open class FirUastApiTest : AbstractFirUastTest() {
...
+            doCheck("uast-kotlin/testData/Anonymous.kt", ::checkCallbackForAnonymous)
+        }
+
+        @TestMetadata("TypeAliases.kt")
+        fun testTypeAliases() {
+            doCheck("uast-kotlin/testData/TypeAliases.kt", ::checkCallbackForTypeAliases)
+        }
+
+// TODO: vararg, arrayOf call inside annotation
+//        @TestMetadata("AnnotationComplex.kt")
```
darthorimar, line 127: What do you think about storing `// FIR_IGNORE` directive inside testdata file instead of directly commenting out each test?

plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastResolveApiFixtureTestBase.kt
```diff
@@ -64,25 +58,29 @@ class KotlinUastResolveApiTest : KotlinLightCodeInsightFixtureTestCase() {
...
         val resolvedDeclaration = functionCall.multiResolve()
         val resolvedDeclarationsStrings = resolvedDeclaration.map { it.element.text ?: "<null>" }
-        assertContainsElements(
-            resolvedDeclarationsStrings,
-            "fun foo(): Int = TODO()",
-            "fun foo(a: Int): Int = TODO()",
-            "fun foo(a: Int, b: Int): Int = TODO()"
-        )
+        // TODO: multiResolve
+        if (!isFirUastPlugin) {
```
darthorimar, line 62: I do not like this approach with a lot of `if (!isFirUastPlugin) { assertSmth() }` in  tests. It does not seem to be very flexible and does not allow to easily find out that some functionality started to work.

What do you think about muting non-passing tests on the `FirUastResolveApiFixtureTest.test*` side? Something like 
```
 fun testMultiResolve() {
   considerFialing {
     checkMultiResolve(myFixture)
  }
}
```
Where `considerFailing` also checks if test started to pass. 

Actually I'm for the generated tests based on a testdata as they are more well structured and can handle such cases well. But considering that there are already a lot of non-generated tests, that approach should work too.

plugins/kotlin/uast/uast-kotlin-fir/test/org/jetbrains/kotlin/idea/fir/uast/FirUastApiTest.kt
```diff
@@ -33,6 +33,146 @@ open class FirUastApiTest : AbstractFirUastTest() {
             // Bogus
         }
 
+        @TestMetadata("AnnotationParameters.kt")
```
darthorimar, line 36: This test suit looks very similar to a generated one but it is not. Maybe it can be generated? :)

darthorimar (changes requested):

plugins/kotlin/uast/uast-kotlin-fir/test/org/jetbrains/kotlin/idea/fir/uast/FirUastApiTest.kt
jsjeon: We've tried that directive before, and it turned out that such addition will alter the render, in particular, hash value inside local variables, which depends on line number or text location. So, I'd rather avoid changing (seemingly) unrelated things.

But, this reminds me of where to put the logic to handle failures, thanks.

plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastResolveApiFixtureTestBase.kt
jsjeon: I also noticed that this won't raise a flag if things start to work, thanks for pointing that out.

Added one more commit that removes plugin-specific assertions and rather adds a logic to check if it is expected to fail or notify users to fix whitelist.

plugins/kotlin/uast/uast-kotlin-fir/test/org/jetbrains/kotlin/idea/fir/uast/FirUastApiTest.kt
jsjeon: That's because I (manually) mimic generated tests so that we can easily go to the test input from test run panel. :)

Hm, regarding making it generated, resolve/api tests exercise different surfaces/aspect of APIs per test input; there won't be a unified output; plus, not all test inputs are used. So, I'm not sure what we can get from automation...

darthorimar: Got it

plugins/kotlin/uast/uast-kotlin-fir/test/org/jetbrains/kotlin/idea/fir/uast/FirUastApiTest.kt
darthorimar: Hm, will adding `// FIR_IGNORE` to the end of the tesdata file will break the line numbers/hashes? The same approach is used in find usages tests where line numbers are also important

jsjeon: Oh...... bottom of the testdata file... 🤦  Yeah, that should work.

Well, I put a whitelist in FIR UAST, along with short investigation notes. I can do the same, like putting some notes after `// FIR_IGNORE`. I can use that later. :)

One thing I'm worried is, the same testdata file is used by a number of test suites, which may use the same directive-based filtering. Adding that directive may affect all of those tests?

darthorimar: > Adding that directive may affect all of those tests?
it may. In that case, let's leave as it is now.

darthorimar (approved):

darthorimar:
Merged, thank you!

PR #1721, opened 2021-09-11
KTIJ-19591 'Property could be private' causes compile error when property is used in annotation

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19591


dimonchik0036 (approved):
Thanks! Merged manually

PR #1722, opened 2021-09-11
KTIJ-18325 ReplaceWith quickfix changes receiver variable to `this` for generic function call with `out` type parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-18325


dimonchik0036 (approved):
Thanks! Merged manually

PR #1723, opened 2021-09-13
FIR/UAST: type-aware constant conversion

jsjeon:
This is based on #1719, and only last two commits are additional.

In FIR, constant value is encoded as `long` value, along with value kind. We should not use the value as-is; rather, we need to convert to the target/expected type. Otherwise, everything will look like `42L`. This can be done in FIR UAST, but maybe better to be in FIR IDE HL API: [KOTLIN-CR-193](https://kotlin.jetbrains.space/p/kotlin/reviews/193/).

While working on it, I found FE 1.0 UAST didn't do a similar thing. Since it already uses type-based constant conversion util, all we need are just to load the type of expression and use that as a hint.


darthorimar:
Let's keep `toConst` in the Analysis API

darthorimar (approved):

darthorimar:
Merged, thank you!

PR #1726, opened 2021-09-14
FIR UAST: implement/fix argument retrieval/conversion

jsjeon:
This is on top of #1723, and only last three commits are additional.

From the commonized API tests, I found there is a good test that exercises base service's `getArgumentForParameter`, which is used by `UCallExpression`. 1st commit implements a missing part in FIR UAST. The idea is, we can retrieve the parameter index/order from the resolved call target, and then we can retrieve the matched argument or `vararg` arguments from `argumentMapping`.

That reminded me of a missing part of `vararg` handling in argument conversion. 2nd commit revisits and fixes that.

While comparing affected test data, I noticed only one diff is left in `AnnotationParameter` test input. It turns out that, in FIR, `arrayOf` call variants in annotations are desugared to `FirArrayOfCall`, instead of regular `FirCall`. [KOTLIN-CR-194](https://kotlin.jetbrains.space/p/kotlin/reviews/194/) adds a resolution for that to HL API, and the last commit is about corresponding test updates.


darthorimar:
Merged, thank you!

PR #1727, opened 2021-09-14
Read java.class.path unconditionally

gharrma:
There are some development environments in which (1) the
lib/classpath.txt file exists *and* (2) certain plugins need to be
loaded from the Java classpath rather than from the plugins directory.

Thus java.class.path should always be added in the bootstrap
classloader, even if lib/classpath.txt exists. Any duplicates
between the two are already filtered out.

Issue link: https://youtrack.jetbrains.com/issue/IDEA-278177

cc @develar


develar:
Apologies for missing this PR earlier. It should be obsolete by now.

PR #1728, opened 2021-09-15
Parcelize: Fix tests and add a test for IgnoredOnParcel annotations in objects

sfs:
- The path to `android.jar` in the parcelize tests seems to be wrong. It's possible that this is done on purpose for teamcity, in which case `getPlugins.sh` should probably generate a `community/android` symlink.
- Added a test for [PR 4579](https://github.com/JetBrains/kotlin/pull/4579) in the Kotlin repository.

### DIFF
--- a/plugins/kotlin/compiler-plugins/parcelize/tests/test/org/jetbrains/kotlin/pacelize/ide/test/parcelizeTestUtil.kt
+++ b/plugins/kotlin/compiler-plugins/parcelize/tests/test/org/jetbrains/kotlin/pacelize/ide/test/parcelizeTestUtil.kt
@@ -12,7 +12,7 @@ import java.io.File
 
 fun addParcelizeLibraries(module: Module) {
     ConfigLibraryUtil.addLibrary(module, "androidJar") {
-        addRoot(File(PathManager.getHomePath(), "community/android/android/testData/android.jar"), OrderRootType.CLASSES)
+        addRoot(File(PathManager.getCommunityHomePath(), "android/android/testData/android.jar"), OrderRootType.CLASSES)
     }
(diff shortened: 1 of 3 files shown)


plugins/kotlin/compiler-plugins/parcelize/tests/test/org/jetbrains/kotlin/pacelize/ide/test/parcelizeTestUtil.kt
```diff
@@ -12,7 +12,7 @@ import java.io.File
 
 fun addParcelizeLibraries(module: Module) {
     ConfigLibraryUtil.addLibrary(module, "androidJar") {
-        addRoot(File(PathManager.getHomePath(), "community/android/android/testData/android.jar"), OrderRootType.CLASSES)
+        addRoot(File(PathManager.getHomePath(), "android/android/testData/android.jar"), OrderRootType.CLASSES)
```
nikitabobko, line 15: It's not wrong. Inside JetBrains we have big monorepo and `intellij-community` you observe on GitHub is subdirectory in this monorepo.

I propose another fix, change `PathManager.getHomePath()` to `PathManager.getCommunityHomePath()`

nikitabobko (changes requested):

plugins/kotlin/compiler-plugins/parcelize/tests/test/org/jetbrains/kotlin/pacelize/ide/test/parcelizeTestUtil.kt
sfs: Thanks for the information, that makes a lot of sense! I've updated the code to use `getCommunityHomePath`.

nikitabobko:
Merged, thanks

PR #1729, opened 2021-09-16
KTIJ-19674 "Redundant qualifier name" should work for supertype entries and annotations

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19674


dimonchik0036 (approved):
Thanks! Merged manually

PR #1731, opened 2021-09-16
IDEA-258738: Double click doesn't select a word if it includes a sear…

wuxuanqicn:
`SearchResults#caretPositionChanged()` called after `SelectionModelImpl#fireSelectionChanged()` which repaint selection result properly, call `myEditor.getSelectionModel().removeSelection()` in `SearchResults#caretPositionChanged()` is inappropriate.

`notifyCursorMoved();` is redundant too, introduce the indicator problem reported in [IDEA-272238](https://youtrack.jetbrains.com/issue/IDEA-272238)


yole:
Note that this proposed fix would regress the fix for another issue, https://youtrack.jetbrains.com/issue/IDEA-256943, so the problem needs to be fixed differently in any case.

PR #1732, opened 2021-09-16
FIR UAST: handle annotation in annotation

jsjeon:
The motivation was an API test for `AnnotationComplex.kt` where annotation is used as annotation value in various forms.

First, unlike annotation call:
```kt
@Annotation(...) // annotation call
class C
```
which is parsed as `KtAnnotationEntry`, which is in turn abstracted as `KtAnnotationCall`, annotation as an annotation value:
```kt
annotation class Annotation(val anno: OtherAnnotation)

@Annotation(<expr>OtherAnnotation(...)</expr>) // function call to ctor
class C
```
is a regular function call. 1st commit in [KOTLIN-CR-198](https://kotlin.jetbrains.space/p/kotlin/reviews/198/) showcases those resolved calls. FIR UAST has unnecessarily strong restriction about annotation handling, and 1st commit in this PR addresses that. (2nd commit is a simple improvement while debugging the test.)

FIR IDE and FIR LC don't model not only annotations in annotations, but other kinds of annotation values, such as enum entry or array of allowed types. 3rd and 4th commits in [KOTLIN-CR-198](https://kotlin.jetbrains.space/p/kotlin/reviews/198/) add abstractions and conversions in FIR IDE and LC, respectively. While doing so, constant value returned from constant provider in Analysis API is changed. 3rd commit in this PR changes API usage.

One thing that's noteworthy is, the key of FIR LC change (which affects FIR IDE as well) is `kotlinOrigin`. Without that, a literal could trigger the conversion from UAST Java plugin, not Kotlin one.

2nd commit in [KOTLIN-CR-198](https://kotlin.jetbrains.space/p/kotlin/reviews/198/) and 4th commit in this PR are about weakening conditions to lookup nested annotation. In FE 1.0 LC util and FE 1.0 UAST, a specific class name is used, and aforementioned commits replace it with a common ancestor.

At last, the test passed, but the whitelist logic didn't notify that. And... that's because notification (via assertion error) is caught and then silently ignored. 🤦  With the fix in 5th commit in this PR, we found that one more API test is passing (guess it's also due to various `kotlinOrigin` additions and array annotation value support); and parent conversion in identifier tests is identical to FE 1.0 UAST. 🎉

### DIFF
--- a/plugins/kotlin/uast/uast-kotlin-base/src/org/jetbrains/uast/kotlin/BaseKotlinConverter.kt
+++ b/plugins/kotlin/uast/uast-kotlin-base/src/org/jetbrains/uast/kotlin/BaseKotlinConverter.kt
@@ -197,8 +197,10 @@ interface BaseKotlinConverter {
                         el<UAnnotation> { KotlinUNestedAnnotation.create(original, givenParent) }
                     } else null
                 }
-                is KtLightAnnotationForSourceEntry -> {
-                    convertDeclarationOrElement(original.kotlinOrigin, givenParent, requiredTypes)
+                is KtLightElementBase -> {
+                    original.kotlinOrigin?.let {
+                        convertDeclarationOrElement(it, givenParent, requiredTypes)
+                    }
                 }
                 is KtDelegatedSuperTypeEntry -> {
                     el<KotlinSupertypeDelegationUExpression> {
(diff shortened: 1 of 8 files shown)


darthorimar (approved):

darthorimar:
Merged, thank you!

PR #1737, opened 2021-09-18
KTIJ-19540 Keep companion object if still used after move refactor

XiangRongLin:
closes https://youtrack.jetbrains.com/issue/KTIJ-19540

I think I created the tests correctly, but I'm not sure. Is there some documentation on what the `companionHasReference.test` does exactly and what the `usages` folder does, which i left out

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/refactoring/move/moveDeclarations/MoveKotlinDeclarationsProcessor.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/refactoring/move/moveDeclarations/MoveKotlinDeclarationsProcessor.kt
@@ -61,11 +61,6 @@ import kotlin.math.min
 interface Mover : (KtNamedDeclaration, KtElement) -> KtNamedDeclaration {
     object Default : Mover {
         override fun invoke(originalElement: KtNamedDeclaration, targetContainer: KtElement): KtNamedDeclaration {
-            fun KtObjectDeclaration.hasNoUsages(): Boolean {
-                return KotlinFindUsagesHandlerFactory(project).createFindUsagesHandler(this, false)
-                    .findReferencesToHighlight(this, LocalSearchScope(containingFile)).isEmpty()
-            }
-
             return when (targetContainer) {
                 is KtFile -> {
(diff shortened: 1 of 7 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/refactoring/move/moveDeclarations/MoveKotlinDeclarationsProcessor.kt
```diff
@@ -59,6 +61,11 @@ import kotlin.math.min
 interface Mover : (KtNamedDeclaration, KtElement) -> KtNamedDeclaration {
     object Default : Mover {
         override fun invoke(originalElement: KtNamedDeclaration, targetContainer: KtElement): KtNamedDeclaration {
+            fun KtObjectDeclaration.hasNoUsages(): Boolean {
```
vladimirdolzhenko, line 64: Please inline this function to avoid unnecessary overhead - it's not good to create a local function just for a single call - if you like you can use `run` that gives same short form with a `this` reference

XiangRongLin: I inlined it.

As for my reasoning behind the function is that I found the if condition easier to read this way.

vladimirdolzhenko:
Thanks, merged as https://github.com/JetBrains/intellij-community/commit/a3cf0c3cb58539b632e0ecb979514d0bbceba055

vladimirdolzhenko:
Please pay attention to the commit message - in Kotlin have some kind of policy (gonna update https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/CONTRIBUTING.md though)

- To mark a commit that fixes some issue you need to add `#KTIJ-NNN Fixed` at the end of commit message
- Some informal rule for IJ: it's worth to start with a plugin prefix, like `[kotlin]` for the Kotlin plugin

XiangRongLin:
There is a kotlin specific readme o.O
I completly missed that, I only saw the one in the project root.

I'll keep those things in mind

PR #1738, opened 2021-09-19
KTIJ-19594 Intention to remove all argument names

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19594
https://youtrack.jetbrains.com/issue/KTIJ-19737
https://youtrack.jetbrains.com/issue/KTIJ-20090


vladimirdolzhenko:
I think this intention could be converted as inspection w/o highlighting a problem (see `com.intellij.codeHighlighting.HighlightDisplayLevel#DO_NOT_SHOW`) 

It allows to use options like `fix all in a file`, wdyt ?

nikitabobko:
Let's firstly decide whether we need such intention at all - https://youtrack.jetbrains.com/issue/KTIJ-19594#focus=Comments-27-5216368.0-0

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveAllArgumentNamesIntention.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
+import org.jetbrains.kotlin.idea.completion.placedOnItsOwnPositionInCall
+import org.jetbrains.kotlin.psi.KtCallElement
+import org.jetbrains.kotlin.psi.KtCollectionLiteralExpression
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.resolve.calls.callUtil.getParameterForArgument
+import org.jetbrains.kotlin.resolve.calls.components.isVararg
+
+class RemoveAllArgumentNamesIntention: SelfTargetingIntention<KtCallElement>(
```
nikitabobko, line 14: I expect the intention to work for cases like this:
```
fun main() {
	foo(b = 1, a = 3)
}

fun foo(a: Int, b: Int) {
}
```
It should convert it to this:
```
fun main() {
	foo(3, 1)
}

fun foo(a: Int, b: Int) {
}
```

nikitabobko, line 14: I expect the intention to work for cases like this:
```
fun main() {
	foo(a= 3, b = 1, d = 1)
}

fun foo(a: Int, b: Int, c: Int = 1, d: Int): Nothing = TODO()
```
it should remove all arguments which can be removed:
```
fun main() {
	foo(3, 1, d = 1)
}

fun foo(a: Int, b: Int, c: Int = 1, d: Int): Nothing = TODO()
```

nikitabobko (changes requested):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveAllArgumentNamesIntention.kt
```diff
@@ -0,0 +1,34 @@
...
+    KtCallElement::class.java,
+    KotlinBundle.lazyMessage("remove.all.argument.names")
+) {
+    override fun isApplicableTo(element: KtCallElement, caretOffset: Int): Boolean {
+        val arguments = element.valueArguments
+        if (arguments.count { it.isNamed() } < 2) return false
+        val collectionLiteral = arguments.firstOrNull { it.getArgumentExpression() is KtCollectionLiteralExpression }
+        if (collectionLiteral != null && collectionLiteral != arguments.last()) return false
```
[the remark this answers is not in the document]
t-kameyama: I don't think the logic can be used here since it determines if a name can be removed from a single argument.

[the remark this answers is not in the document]
nikitabobko: Sorry for long response. Yes the logic cannot be re-used "literally" but it certainly can be abstracted. Think about it: what's the difference between intention which removes one name of a named argument and all names of named arguments? The second one is just the first one but repeated for all arguments.

nikitabobko (changes requested):
Changes in reusing the logic is requested

nikitabobko:
The intention shouldn't be available for this case:
```
fun main() {
    foo(b = 1, a = 3)
}

fun foo(c: Int = 2, a: Int, b: Int) {
}
```

plugins/kotlin/completion/src/org/jetbrains/kotlin/idea/completion/NamedArgumentCompletion.kt
```diff
@@ -118,10 +116,23 @@ object NamedArgumentCompletion {
...
+    val parameter = resolvedCall.getParameterForArgument(this)
+    val isVararg = parameter?.isVararg == true
+
+    if (noOtherNamedArguments) {
+        val parameters = resolvedCall.resultingDescriptor.valueParameters
+        if (isVararg && parameter != parameters.lastOrNull()) return false
+        val firstParameter = parameters.firstOrNull()
+        if (firstParameter?.hasDefaultValue() == true && resolvedCall.valueArguments[firstParameter] is DefaultValueArgument) return false
+        return isVararg || isNamed() || placedOnItsOwnPositionInCall(resolvedCall)
+    }
```
nikitabobko, line 133: By asking you to "abstract" the code I didn't mean to introduce one boolean flag to this function which changes this function behavior completely. 

Well, indeed, `noOtherNamedArguments` boolean flag introduction made this code better because now the logic of "argument can be used without name" is localized in one function and not spread around different places in the codebase.

But still, it's not enough. Right now, if anyone wanted to fix any other bug in this subsystem (e.g. [this bug](https://youtrack.jetbrains.com/issue/KTIJ-19737)) then he/she would need to fix the bug twice -- once for "Remove argument name" intention and once for "Remove all argument names" intention.

Let me help you a bit with abstracting this API. Try it this way: change this function signature to: `fun collectAllArgumentsThatCanBeUsedWithoutName(...): List<RemoveNamedArgumentNameTransformation>`

where `RemoveNamedArgumentNameTransformation` is:
```
data class RemoveNamedArgumentNameTransformation(val argumentName: String, val argumentNewIndex: Int)
```

(this API is approximate and only to give you the overall idea)

You can use this API in both "Remove argument name" and "Remove all argument names" intentions.

Abstracting the API this way will fix [this issue](https://youtrack.jetbrains.com/issue/KTIJ-20090) too.

nikitabobko (changes requested):

nikitabobko:
https://youtrack.jetbrains.com/issue/KTIJ-19737

t-kameyama:
> https://youtrack.jetbrains.com/issue/KTIJ-19737

Added tests for [KTIJ-19737](https://youtrack.jetbrains.com/issue/KTIJ-19737) and [KTIJ-20090](https://youtrack.jetbrains.com/issue/KTIJ-20090).

PR #1739, opened 2021-09-19
KTIJ-19517 Fix false positive warning on move method without param

XiangRongLin:
When comparing methods it zips the parameters of both methods and checks if the zipped combinations are equal.
In the case where one method has no parameters, the zip returns an empty list. On that empty list the "all" check automatically returns true which is wrong.

https://youtrack.jetbrains.com/issue/KTIJ-19517

I needed to expand the testing setup for move method, because in this case there are 2 methods with the same name, which could not be differenciatet between, resulting in the wrong method being selected for moving.


XiangRongLin:
I rebased, solved the merge conflicts and adjusted the commit message to the format for kotlin changes

nikitabobko (changes requested):
Your PR breaks such cases:
```
class Bug {
	fun dummyFun() {}
	fun toMove() {}
}

class Test  {
	fun toMove(n: Bug) { // Move shouldn't be allowed
		n.dummyFun()
	}
}
```

Even though this case was broken before your PR I believe that it's part of the issue to fix it:
```
class Bug {
	fun dummyFun() {}
	fun toMove(s: String) {}
}

class Test  {
	fun toMove(n: Bug, s: String) { // Move shouldn't be allowed
		n.dummyFun()
	}
}

```

nikitabobko:
Closing as a stale Pull Request. Feel free to reopen it/leave a comment if you are going to return to it

PR #1741, opened 2021-09-20
WIP: UAST: introduce AbstractTypeMapper-based type mapper

jsjeon:
As per [KT-48532](https://youtrack.jetbrains.com/issue/KT-48532), old BE will be deprecated, and `KotlinTypeMapper` is part of it ([Kt-48773](https://youtrack.jetbrains.com/issue/KT-48773)). FE 1.0 UAST uses it to convert `KotlinType` to `PsiType`.

As part of NI (new inference), `AbstractTypeMapper` is introduced and used across BE IR (`IrTypeMapper`) and FE IR (`FirJvmTypeMapper`). This attempts to introduce yet another version of `AbstractTypeMapper` for `KotlinType`, which requires `kotlin` repo side change: https://github.com/JetBrains/kotlin/pull/4587.

------

For now, there are a couple issues left:
* `getClassInternalName` doesn't convert class name correctly: it should be FQ name, but just simple class name.
* `AbstractTypeMapper` doesn't have a handling for `typealias`. Corresponding descriptor should be handled at the client side.
* various type-related utils, such as `KClass` check, script check, etc., are not fully implemented nor tested.

### DIFF
--- a/plugins/kotlin/uast/uast-kotlin/src/org/jetbrains/uast/kotlin/internal/KotlinUastTypeMapper.kt
+++ b/plugins/kotlin/uast/uast-kotlin/src/org/jetbrains/uast/kotlin/internal/KotlinUastTypeMapper.kt
@@ -6,79 +6,48 @@ package org.jetbrains.uast.kotlin.internal
 
 import org.jetbrains.kotlin.builtins.functions.FunctionClassDescriptor
 import org.jetbrains.kotlin.builtins.functions.FunctionClassKind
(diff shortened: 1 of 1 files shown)


jsjeon:
As mentioned at https://youtrack.jetbrains.com/issue/KT-48773#focus=Comments-27-5323746.0-0, this effort seems redundant, and we may be able to re/use (parts of) KtFe10Analysis in the future.

PR #1743, opened 2021-09-20
IDEA-241674 Debug ternary operator 

marcin-bukowiecki:
New logic will try to do type conversion from the left/right conditional result to `expectedType` which is the type of conditional expression.

https://youtrack.jetbrains.com/issue/IDEA-241674


marcin-bukowiecki:
Hello sorry for the delay. I did a small code update and used the `TypeCastEvaluator`. Unfortunetly for boxed types I need to do a conversion to primitive type because `TypeCastEvaluator` was throwing exceptions. I will also do some testing for other types to make sure everything works.

gorrus:
I made slightly shorter fix in master

gorrus:
check https://github.com/JetBrains/intellij-community/commit/271643147d13e98e0f73d088c30021521963cf25

PR #1748, opened 2021-09-21
Make the IDEA run target build the KotlinPluginCommunity.

madsager:
Also, make sure plugin.path is set when running.


nikitabobko:
Elaboration: together with @darthorimar we have git-bisected the problem and the problem aroused after Ilya splitted `kotlin-compiler-for-ide.jar` to FIR and FE10 version.

This PR isn't correct because it shouldn't be needed to add KotlinPlugin artifact manually to classpath but it's not yet obvious why it has broken in the first place

madsager:
Thanks for the update Nikita. I identified the same root cause. I did the following workaround as that appears to be how the IDEA FIR target works. It seems to manually build the FIR based KotlinPlugin and add it to the classpath. I'll close this PR since you are looking into it. :)

PR #1749, opened 2021-09-21
UAST: deprecate use of KotlinTypeMapper#mapType

jsjeon:
As per [KT-48532](https://youtrack.jetbrains.com/issue/KT-48532), old BE will be deprecated, and `KotlinTypeMapper` is part of it ([Kt-48773](https://youtrack.jetbrains.com/issue/KT-48773)). FE 1.0 UAST uses it to convert `KotlinType` to `PsiType`.

As part of NI (new inference), `AbstractTypeMapper` is introduced and used across BE IR (`IrTypeMapper`) and FE IR (`FirJvmTypeMapper`). Adding yet another implementation of `AbstractTypeMapper` for `KotlinType` (#1741) is not that trivial as described there.

Instead, this PR is a short-term solution: use `descriptorBasedTypeSignatureMapping.kt` along with utils for generic type writing from `KotlinTypeMapper`.


udalov (approved):

xiexed (approved):

darthorimar:
Merged, thanks!

PR #1751, opened 2021-09-21
[kotlin] KTIJ-19714 Remove duplicated extract constant action for kotlin

XiangRongLin:
https://youtrack.jetbrains.com/issue/KTIJ-19714

Seems like I misunderstood what that thing does and thus duplicated it. The java equivalent of refactorings.xml already has the action defined.

With it removed it looks like this now
image


vladimirdolzhenko:
Thanks for the contribution.

Merged as https://github.com/JetBrains/intellij-community/commit/027351b16d8943b86245a503a80fabaf4b84d18d

PR #1752, opened 2021-09-22
Teach launcher.py about ideScript

skissane:
Make `launcher.py` aware of `ideScript` command, so it can be used to trigger running an IDE script in a existing IntelliJ IDEA instance

### Test case
1. In IntelliJ IDEA CE go to Tools > Create Command Line Launcher to save the launcher script, e.g. as `/tmp/idea`
2. Create a file called `test.groovy` with contents `com.intellij.openapi.application.ApplicationManager.getApplication().invokeLater({com.intellij.openapi.ui.Messages.showMessageDialog(null, "Hello World!", "Greeting", null)})`
3. Run `/tmp/idea ideScript ./test.groovy`
4. Switch to running IDEA app, observe "Hello World!" popup window

By comparison, prior to this change, the same steps above would not run `./test.groovy`, instead the IDEA editor would open up it up for editing.

@gregsh I thought you might be interested in this PR since you wrote the `ideScript` component

### DIFF
--- a/platform/platform-resources/src/launcher.py
+++ b/platform/platform-resources/src/launcher.py
@@ -20,7 +20,8 @@ def print_usage(cmd):
            '  {0} [project_dir] [-w|--wait]\n' +
            '  {0} [-l|--line line] [project_dir|--temp-project] [-w|--wait] file[:line]\n' +
            '  {0} diff <left> <right>\n' +
-           '  {0} merge <local> <remote> [base] <merged>').format(cmd))
+           '  {0} merge <local> <remote> [base] <merged>\n' +
+           '  {0} ideScript <script>').format(cmd))
 
 
 def write_to_sock(sock, data):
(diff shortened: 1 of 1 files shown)


trespasserw:
Thanks! There is only one problem with the change: we're trying to get rid of that script :) For the record, which OS you are using?

skissane:
I am using macOS (Big Sur x86). Is there an alternative/replacement to using `launcher.py`?

trespasserw:
Yep, macOS is the biggest obstacle. The alternative is using the app directly (`open -na IDEA --args ...`), but it is not particularly convenient. The [Toolbox app](https://www.jetbrains.com/toolbox-app/) can create such shortcut scripts.

trespasserw:
`launcher.py` is retired in favor of standard ways (`open -na IDEA.app` or `.../IDEA.app/Contents/MacOS/idea`).

PR #1757, opened 2021-09-24
FIR UAST: test updates (synthetic members of data class)

jsjeon:
This is simple test updates for [KOTLIN-CR-210](https://kotlin.jetbrains.space/p/kotlin/reviews/210/).


plugins/kotlin/uast/uast-kotlin-fir/testData/legacyRenderLog/BrokenMethod.render.fir.txt
```diff
@@ -2,17 +2,23 @@ public final class BrokenMethodKt {
     public static final fun main(@org.jetbrains.annotations.NotNull args: java.lang.String[]) : void {
         public static final class TestUsed {
             @org.jetbrains.annotations.NotNull private final var parameter: java.lang.CharSequence
-            public final fun component1() : java.lang.CharSequence = UastEmptyExpression
-            public final fun copy(@org.jetbrains.annotations.NotNull parameter: java.lang.CharSequence) : java.lang.Object = UastEmptyExpression
             public final fun getParameter() : java.lang.CharSequence = UastEmptyExpression
             public fun TestUsed(@org.jetbrains.annotations.NotNull parameter: java.lang.CharSequence) = UastEmptyExpression
+            public final fun component1() : java.lang.CharSequence = UastEmptyExpression
+            public final fun copy(@org.jetbrains.annotations.NotNull parameter: java.lang.CharSequence) : java.lang.Object = UastEmptyExpression
```
jsjeon, line 8: Now, the only difference is the return type of `copy`: containing `data` class v.s. just `Object`. Since these `data` classes in this test are _local_, during `PsiType` conversion, FIR IDE/LC turns the resolved `data` class types to `Object` as part of public type approximation. I think FIR's behavior is correct.

darthorimar: Seems not correct. As for this class, public type approximation should happen only if the type is exposed outside the scope where the local class is available.

Consider the following two examples

```
fun foo1() {
	class X {
		fun a(): X
	}
	val a = X().a() // the type of a here is X and it is correct as X is available in current scope
}


fun foo2(): Any { // type of foo() is Any here as X is not available outside foo2
	class X {
		fun a(): X
	}
	return X().a() 
}
```

darthorimar (changes requested):

plugins/kotlin/uast/uast-kotlin-fir/testData/legacyRenderLog/BrokenMethod.render.fir.txt
jsjeon: Thanks for examples along with explanations. That makes sense. Since I saw the underlying FIR/symbol has a correct type, it's just a matter of _when_ to do public type approximation here: https://github.com/JetBrains/kotlin/blob/master/analysis/analysis-api-fir/src/org/jetbrains/kotlin/analysis/api/fir/components/KtFirPsiTypeProvider.kt#L63 I'm not sure we can simply remove that or not.

Actually, while working on some other resolve API failure, I found that FIR UAST pass an incorrect type conversion context (https://github.com/JetBrains/kotlin/blob/master/analysis/analysis-api-fir/src/org/jetbrains/kotlin/analysis/api/fir/components/KtFirPsiTypeProvider.kt#L96), which can be used for type approximation context as well. This seems non-trivial, separable work item, so I'd prefer to do that in a follow-up PR (and may need a separate CR in `kotlin` repo side as well).

darthorimar: Light Classes should mimic how Kotlin/JVM compiler behaves in such cases. So, the compiler behavior should be investigated, and we should do the same for the `KtFirPsiTypeProvider.asPsiType` function as it used in LC. I'm not sure if it is applicable for the UAST

darthorimar (approved):

darthorimar:
Thank you! Merged

PR #1761, opened 2021-09-30
IDEA-279319: move updateDefaultChannel implementation to UpdateStrategyCustomization

jessecordeiro:
Moving updateDefaultChannel implementation to UpdateStrategyCustomization to allow for the default implementation to be overridden (by extending UpdateStrategyCustomization).

https://youtrack.jetbrains.com/issue/IDEA-279319


platform/platform-impl/src/com/intellij/openapi/updateSettings/UpdateStrategyCustomization.java
```diff
@@ -45,4 +55,26 @@ public boolean haveSameMajorVersion(@NotNull BuildNumber build1, @NotNull BuildN
   public boolean isNewerVersion(@NotNull BuildNumber candidateBuild, @NotNull BuildNumber currentBuild) {
     return candidateBuild.compareTo(currentBuild) > 0;
   }
+
+  public void updateDefaultChannel(UpdateSettings settings, ChannelStatus currentChannelStatus) {
```
chashnikov, line 59: Could you please describe how you're going to override this method?
(It's better to avoid putting implementation code into API classes, also this method exposes classes from *.impl package to API, so I would prefer to extract a method which does customization only, and leave the default implementation in UpdateCheckerService.

dahlstrom-g: Maybe we can pair this with a PR in [JetBrains/android](https://github.com/JetBrains/android).

chashnikov: It would be enough to post the code here in a comment :)

jessecordeiro: @chashnikov We were hoping to have completely different implementation, something like:

```
@Override
  public void updateDefaultChannel(UpdateSettings settings, ChannelStatus currentChannelStatus) {
    String versionName = getApplicationVersionName();
    ChannelStatus targetChannelStatus = parseVersionNameToChannelStatus(versionName);

    if (currentChannelStatus != targetChannelStatus) {
      if (isNotInitialLaunchForIDE()) {
        createNotificationForSwitchingChannels(targetChannelStatus);
      }
      settings.setSelectedChannelStatus(targetChannelStatus);
      LOG.info(String.format("channel set to '%s'", targetChannelStatus.getCode()));
    }
  }
```

PR #1762, opened 2021-10-01
Fix for IDEA-279024: Incorrect decompilation of indyfied string concatenation in Java 9+

Maccimo:
Fix for IDEA-279024: Incorrect decompilation of indyfied string concatenation in Java 9+


trespasserw:
Merged, thanks!

PR #1763, opened 2021-10-03
KTIJ-19680 'Replace with a for loop' and backward conversion could work for 'forEachIndexed'

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19680
https://youtrack.jetbrains.com/issue/KTIJ-19679


comitative:

It seems that the current implementation is not yet fully correct and needs some more work as well as additional unit tests. Here are some examples there the intention provides wrong results.

1. If the `for` loop has no braces, the loop body is lost after the conversion to `forEachIndexed`.

```
for ((index, element) in withIndex()) println("$index : $element")
```

results in

```
forEachIndexed { index, element ->
}
```

2. If the `forEachIndexed` call is an expression body of a function, the quick fix currently produces the syntactically incorrect result.

```
fun List<Int>.foo() = forEachIndexed { index, element -> println("$index : $element") }
```

is replaced by

```
fun List<Int>.foo() = for ((index, element) in withIndex()) {
    println("$index : $element")
}
```

The same problem occurs with the `forEach` call.

I think that the simplest solution is to check if the `forEach`/'forEachIndexed` call is an expression body, and do not generate the "Replace with a 'for' loop" action for expression bodies.

Another option would be to implicitly convert the expression body to a block body (by reusing and existing quick fix) and then to convert the function call to a `for` loop, but I think it would be more difficult to implement and may be less clear to users.

3. You did a good job to convert `continue` to `return@forEachIndexed` and vice versa, but the `break` statement is not supported.

For example, let's look at this sample.
```
fun test() {
    for ((index, element) in listOf(1, 2, 3).withIndex()) {
        println("$index: $element")
        if (index == 1) break
    }
}
```

An incorrect code is generated during the conversion, as the `break` statement is not allowed outside the `for` loop.
```
fun test() {
    listOf(1, 2, 3).forEachIndexed { index, element ->
        println("$index: $element")
        if (index == 1) break
    }
}
```

t-kameyama:
I'll resolve conflicts.

t-kameyama:
> I'll resolve conflicts.

Done

t-kameyama:
Created a new PR for [KTIJ-19679](https://youtrack.jetbrains.com/issue/KTIJ-19679).
https://github.com/JetBrains/intellij-community/pull/2311

t-kameyama:
Created a new [PR](https://github.com/JetBrains/intellij-community/pull/2317) for [KTIJ-19680](https://youtrack.jetbrains.com/issue/issue/KTIJ-19680/Replace-with-a-for-loop-and-backward-conversion-could-work-for-forEachIndexed)

PR #1764, opened 2021-10-03
KTIJ-19619 Incorrect adding brace after `when` before infix keyword on Enter autocomplete

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19619


comitative:
Thank you, merged the PR. I am sorry for a long delay.

PR #1765, opened 2021-10-06
[python] Add python 3.11 to list of python versions

KotlinIsland:
https://www.python.org/dev/peps/pep-0664/

Also fixed a small issue where `__future__.annotations` was marked as included in 3.10, when it's now marked as 3.11.


sproshev:
Merged manually in 119696be4190a68c1cb4c6fdfc11ff3f6fe313aa

PR #1766, opened 2021-10-07
Never step to fake line numbers generated by the Kotlin compiler.

madsager:
The compiler inserts a fake line number for single-line inline
function calls with a lambda parameter such as:

   `42.let { it + 1 }`

This is done so that a breakpoint can be set on the lambda and
on the line even though the lambda is on the same line. When
stepping, we do not want to stop at such fake line numbers.
They cause us to step to line 1 of the current file.

For StepOver this is simple as we just need to filter them out
in the existing Kotlin specific debugger code.

For StepInto we need to add special Kotlin handling.

^KTIJ-18645 Fixed.


nikita-nazarov:
Hi, @madsager! The code is fine, but we made many changes to Kotlin stepping actions not long ago, so I decided to finish up the pr myself. I've rebased your commit on master, refactored the code a bit, and fixed failing tests. It is now merged: 40f16aa9c964130d4dc6cc3016711723b0d0a1a4. Thank you very much for your contribution!

PR #1767, opened 2021-10-07
IDEA-279879 Highlight each line with trailing spaces in text block not just the first one

trushev:
This PR fixes IDEA-279879. 

#### Proposed solution
Continue processing lines of text block, instead of returning from the function

#### IDEA behavior after the change
text_after


artfly:
Hello! In the initial ticket you mention that the idea behind this change is to highlight first trailing space in each line since they are removed by inspection or javadoc. The problem is that it is not only the first space that is removed but all of the trailing spaces in each line. So, if we really want to show what inspection will remove then we would need to highlight all of this spaces.
And this approach "highlight all of the trailing spaces" is not really feasible cause it will introduce lots of highlighting noise (and it wouldn't look nice from user perspective). So, it seems to me that we have two options:
1. leave everything as is
2. show what actually will be removed only when user "selects" our inspection somehow
First option is easy to implement, let's discuss the second one :) We already have a common solution for it - preview inspection result (alt-enter to open warning and then f1 to show preview of what inspection quick-fix will do). Right now this feature is not implemented for inspection that we're discussing, so feel free to add it as another pull request (see LocalQuickFix#generatePreview) :)
One more option is to highlight all of the spaces that will be removed when we put caret on warning (first trailing space). This is much harder and you probably will need to ask for consult in the ticket beforehand
So, unfortunately I have to reject this pull request for now, if you want we can continue discussion in ticket.

PR #1783, opened 2021-10-15
KTIJ-13810 "Remove useless is check" quickfix should apply second quickfix to delete `if` expression with constant condition

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-13810


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/RemoveUselessIsCheckFix.kt
```diff
@@ -6,28 +6,41 @@ import com.intellij.openapi.editor.Editor
...
-import org.jetbrains.kotlin.psi.KtFile
-import org.jetbrains.kotlin.psi.KtIsExpression
-import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.idea.inspections.ConstantConditionIfInspection
+import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 
-class RemoveUselessIsCheckFix(element: KtIsExpression) : KotlinQuickFixAction<KtIsExpression>(element) {
+class RemoveUselessIsCheckFix(
```
nikitabobko, line 14: It's absolutely strange to see that "constant `is` check" is implemented as a separate quick fix. We should have only one generic quick fix which is called smth like "SimplifyExpressionQuickFix". The quick fix should combinate `org.jetbrains.kotlin.idea.inspections.ConstantConditionIfInspection.SimplifyFix`, `org.jetbrains.kotlin.idea.inspections.ConstantConditionIfInspection.RemoveFix` and probably other quick fixes, IDK, it's up for investigation.

It's not your fault that we have a bit of dirty quick fixes design. And we have separate quick fix but since you are change behaviour in this subsystem it's better to refactor it before making any changes

nikitabobko: And it would be great to have dynamic quick fix message depending on the fix it's going to do. For example it should say "Unwrap 'if' statement" in case when it's going to unwrap if statement instead of generic "Simplify boolean expression"

nikitabobko (changes requested):

nikitabobko:
Please, run the inspection on the entire `intellij-community` project ([how to run the inspection on the entire project](https://www.jetbrains.com/help/idea/running-inspections.html#run-one-inspection)) and either fix all the warnings introduced by the inspection (in separate commit but in this PR) or share your results as a comment.

t-kameyama:
> Please, run the inspection on the entire `intellij-community` project ([how to run the inspection on the entire project](https://www.jetbrains.com/help/idea/running-inspections.html#run-one-inspection)) and either fix all the warnings introduced by the inspection (in separate commit but in this PR) or share your results as a comment.

This PR does not fix inspections.

yanex:
Hello, @t-kameyama! As always, thank you for your contribution!
We inspected the state of inspections/quickfixes related to constant expressions, and what I have to say is that we decided that the whole set of inspections needs to be reworked. The goals are both improving UX and getting rid of duplicated and obsolete code. Although the PR is certainly valuable, it doesn't bring us closer to what we settled on, so I want to say sorry for closing it. Unfortunately, we don't have enough resources in the team to work on refactoring mentioned inspections right now, but we'll certainly return back to them after some time.
Again, thank you!

yanex:
Hello, @t-kameyama! As always, thank you for your contribution!
We inspected the state of inspections/quickfixes related to constant expressions, and what I have to say is that we decided that the whole set of inspections needs to be reworked. The goals are both improving UX and getting rid of duplicated and obsolete code. Although the PR is certainly valuable, it doesn't bring us closer to what we settled on, so I want to say sorry for closing it. Unfortunately, we don't have enough resources in the team to work on refactoring mentioned inspections right now, but we'll certainly return back to them after some time.
Again, thank you!

PR #1794, opened 2021-10-22
KTIJ-6817 Convert member to extension: do not preserve declaration-site variance on type parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-6817


t-kameyama:
Added e544a18

PR #1798, opened 2021-10-25
IDEA-281190: consider `MethodType.methodType` overloads in MethodHandle inspections

cac03:
Hello,

I've described a problem in [IDEA-281190](https://youtrack.jetbrains.com/issue/IDEA-281190) and I propose these changes as a fix

Please, let me know if I should something differently in this PR or maybe consider another approach, and I'll fix the code in PR

Thanks!

### DIFF
--- a/java/java-analysis-impl/src/com/intellij/psi/impl/source/resolve/reference/impl/JavaReflectionReferenceUtil.java
+++ b/java/java-analysis-impl/src/com/intellij/psi/impl/source/resolve/reference/impl/JavaReflectionReferenceUtil.java
@@ -29,8 +29,12 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
+import java.lang.invoke.MethodType;
 import java.util.*;
 import java.util.function.Function;
+
+import static com.intellij.psi.CommonClassNames.JAVA_UTIL_ARRAYS;
+import static com.intellij.psi.CommonClassNames.JAVA_UTIL_LIST;
 
 /**
  * @author Pavel.Dolgov
@@ -85,6 +89,10 @@ public final class JavaReflectionReferenceUtil {
   public static final String ATOMIC_LONG_FIELD_UPDATER = "java.util.concurrent.atomic.AtomicLongFieldUpdater";
   public static final String ATOMIC_INTEGER_FIELD_UPDATER = "java.util.concurrent.atomic.AtomicIntegerFieldUpdater";
   public static final String ATOMIC_REFERENCE_FIELD_UPDATER = "java.util.concurrent.atomic.AtomicReferenceFieldUpdater";
+
+  // Known lists: List.of and Arrays.asList
+  public static final String OF = "of";
+  public static final String AS_LIST = "asList";
 
   private static final RecursionGuard<PsiElement> ourGuard = RecursionManager.createGuard("JavaLangClassMemberReference");
 
@@ -468,6 +476,39 @@ public final class JavaReflectionReferenceUtil {
     return null;
   }
 
+  /**
+   * Tries to unwrap array or {@link List} and find its components
+   *
+   * @param maybeArrayOrList an array or {@link List} to unwrap
+   * @return list of unwrapped array or list components, some or all of them could be null if unknown (but the length is known);
+   * returns null if nothing is known.
+   */
+  @Nullable
+  public static List<PsiExpression> getArrayOrListComponents(@Nullable PsiExpression maybeArrayOrList) {
+    if (maybeArrayOrList == null) {
+      return null;
+    }
+    final List<PsiExpression> varargs = getVarargs(maybeArrayOrList);
+    if (varargs != null) {
+      return varargs;
+    }
+    if (maybeArrayOrList instanceof PsiMethodCallExpression) {
+      final PsiMethodCallExpression callExpression = (PsiMethodCallExpression)maybeArrayOrList;
+      if (isCallToMethod(callExpression, JAVA_UTIL_LIST, OF)
+          || isCallToMethod(callExpression, JAVA_UTIL_ARRAYS, AS_LIST)) {
+        final PsiExpression[] expressions = callExpression.getArgumentList().getExpressions();
+        if (expressions.length == 1) {
+          final List<PsiExpression> listFromVarArgs = getVarargs(expressions[0]);
+          if (listFromVarArgs != null) {
+            return listFromVarArgs;
+          }
+        }
+        return Arrays.asList(expressions);
+      }
+    }
+    return null;
+  }
+
   @Contract("null -> false")
   public static boolean isVarargAsArray(@Nullable PsiExpression maybeArray) {
     final PsiType type = maybeArray != null ? maybeArray.getType() : null;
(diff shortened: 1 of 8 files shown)


java/java-analysis-impl/src/com/intellij/psi/impl/source/resolve/reference/impl/JavaReflectionReferenceUtil.java
```diff
@@ -487,28 +528,94 @@ public static ReflectiveSignature composeMethodSignature(@Nullable PsiExpression
       final PsiMethodCallExpression methodCallExpression = (PsiMethodCallExpression)typeDefinition;
       final String referenceName = methodCallExpression.getMethodExpression().getReferenceName();
 
-      Function<PsiExpression[], ReflectiveSignature> composer = null;
+      final PsiMethod method = methodCallExpression.resolveMethod();
```
amaembo, line 531: Likely, `CallMatcher.staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE, GENERIC_METHOD_TYPE)` matcher will work here.

amaembo: You can even create a dynamic signature-based mapper, like this (assuming `CallMatcher.*` static import):

```java
      private static final CallMapper<ReflectiveSignature> SIGNATURE_MAPPER = new CallMapper<ReflectiveSignature>()
        .register(anyOf(
                    staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE).parameterCount(1),
                    staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE).parameterCount(3),
                    staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE).parameterTypes(JAVA_LANG_CLASS, JAVA_LANG_CLASS)),
                  call -> composeMethodSignatureFromTypes(call.getArgumentList().getExpressions()))
        .register(staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, GENERIC_METHOD_TYPE),
                  call -> composeGenericMethodSignature(call.getArgumentList().getExpressions()))
        .register(staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE).parameterTypes(JAVA_LANG_CLASS, JAVA_LANG_INVOKE_METHOD_TYPE),
                  call -> composeMethodSignatureFromReturnTypeAndMethodType(call.getArgumentList().getExpressions()))
        .register(staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE).parameterTypes(JAVA_LANG_CLASS, JAVA_LANG_CLASS + "[]"),
                  call -> composeMethodSignatureFromReturnTypeAndListOrArray(call.getArgumentList().getExpressions()))
        .register(staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE).parameterTypes(JAVA_LANG_CLASS, JAVA_UTIL_LIST),
                  call -> composeMethodSignatureFromReturnTypeAndListOrArray(call.getArgumentList().getExpressions()));
      ...
      return SIGNATURE_MAPPER.mapFirst(methodCallExpression);
```

Btw, it's easy to dispatch array and list cases to two separate methods, so probably it would be nice to split `composeMethodSignatureFromReturnTypeAndListOrArray`

java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangReflectHandleInvocationChecker.java
```diff
@@ -251,6 +255,32 @@ else if (invokeParent instanceof PsiVariable) {
...
+            // Call to MethodType.methodType(Class<?>, List<Class<?>>) or MethodType.methodType(Class<?>, Class<?>[])
+            final List<PsiExpression> signature = ContainerUtil.prepend(components, returnType);
+            return ContainerUtil.map(signature, parameter -> (() -> getReflectiveType(parameter)));
+          }
+          if (secondArgument instanceof PsiMethodCallExpression) {
+            final PsiMethodCallExpression maybeNestedMethodTypeCall = (PsiMethodCallExpression)secondArgument;
+            if (isCallToMethod(maybeNestedMethodTypeCall, JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE)) {
+              // Call to MethodType.methodType(Class<?>, MethodType)
+              final List<Supplier<ReflectiveType>> nestedSignature =
+                ourGuard.doPreventingRecursion(methodTypeExpression, false, () -> getLazyMethodSignature(maybeNestedMethodTypeCall));
```
amaembo, line 272: RecursionGuard is mainly intended to catch bugs. Here, we can probably limit recursion depth by ourselves, like passing it as a parameter to `composeMethodSignature`? It could be even outer/inner boolean flag, as I don't think we need more than two levels. Also, I'm not sure whether it's a very useful thing to support. If we created a method type right here, why would we rebind the return type? Is it useful?

amaembo (changes requested):
In general a good contribution, thanks! Please check the review comments.

java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangReflectHandleInvocationChecker.java
cac03: > RecursionGuard is mainly intended to catch bugs. Here, we can probably limit recursion depth by ourselves, like passing it as a parameter to composeMethodSignature? It could be even outer/inner boolean flag, as I don't think we need more than two levels.

1. I have added an explicit `boolean` flag to control recursion. Only `composeMethodSignatureFromReturnTypeAndMethodType` calls `composeMethodSignature` recursively and it does explicitly disallowing further recursion
2. To resolve the innermost `MethodType` in such cases I use the iterative `findInnermostMethodType` method instead of recursion as in the previous version of the PR

> Also, I'm not sure whether it's a very useful thing to support. If we created a method type right here, why would we rebind the return type? Is it useful?

To be honest, I'm not sure either. I've looked up the usages of the `methodType(Class, MethodType)` overload quickly in other projects: JRuby, Jython, Groovy, and could not find one

But here is a potential, artificial example where the inspection could be useful:

```kotlin
class Foo {
    fun bar(list: List<String>): String = "abc"
    fun bar(list: List<Int>): Int = 12
}
```

```java
public class Main {
    private static final MethodHandle BAR_STRING_METHOD_HANDLE;
    private static final MethodHandle BAR_INT_METHOD_HANDLE;


    static {
        try {
            MethodHandles.Lookup lookup = MethodHandles.lookup();
            MethodType stringListMethodType = methodType(String.class, List.class);
            BAR_STRING_METHOD_HANDLE = lookup.findVirtual(Foo.class, "bar", stringListMethodType);
            BAR_INT_METHOD_HANDLE = lookup.findVirtual(Foo.class, "bar", methodType(int.class, stringListMethodType));
        } catch (Throwable t) {
            throw new IllegalStateException(t);
        }
    }

    public void test(Foo foo) throws Throwable {
        String s = (String) BAR_STRING_METHOD_HANDLE.invokeExact(foo, List.of());
        System.out.println("s = " + s);
        long i = (long) BAR_INT_METHOD_HANDLE.invokeExact(foo, Set.of());
        //        ^^ warning                                   ^^ warning
        System.out.println("i = " + i);
    }

    public static void main(String[] args) throws Throwable {
        new Main().test(new Foo());
    }
}
```

java/java-analysis-impl/src/com/intellij/psi/impl/source/resolve/reference/impl/JavaReflectionReferenceUtil.java
cac03: > You can even create a dynamic signature-based mapper, like this (assuming CallMatcher.* static import):

Replaced with dynamic `CallMatcher` in both `JavaReflectionReferenceUtil` and `JavaLangReflectHandleInvocationChecker`

> so probably it would be nice to split composeMethodSignatureFromReturnTypeAndListOrArray

Split

java/java-analysis-impl/src/com/intellij/psi/impl/source/resolve/reference/impl/JavaReflectionReferenceUtil.java
```diff
@@ -86,6 +90,10 @@
   public static final String ATOMIC_INTEGER_FIELD_UPDATER = "java.util.concurrent.atomic.AtomicIntegerFieldUpdater";
   public static final String ATOMIC_REFERENCE_FIELD_UPDATER = "java.util.concurrent.atomic.AtomicReferenceFieldUpdater";
 
+  // Known lists: List.of and Arrays.asList
+  public static final String OF = "of";
```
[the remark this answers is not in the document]
cac03: Replaced with a `CallMatcher`

java/java-analysis-impl/src/com/intellij/psi/impl/source/resolve/reference/impl/JavaReflectionReferenceUtil.java
```diff
@@ -468,6 +476,39 @@ public static boolean isCallToMethod(@NotNull PsiMethodCallExpression methodCall
...
+    final List<PsiExpression> varargs = getVarargs(maybeArrayOrList);
+    if (varargs != null) {
+      return varargs;
+    }
+    if (maybeArrayOrList instanceof PsiMethodCallExpression) {
+      final PsiMethodCallExpression callExpression = (PsiMethodCallExpression)maybeArrayOrList;
+      if (isCallToMethod(callExpression, JAVA_UTIL_LIST, OF)
+          || isCallToMethod(callExpression, JAVA_UTIL_ARRAYS, AS_LIST)) {
+        final PsiExpression[] expressions = callExpression.getArgumentList().getExpressions();
+        if (expressions.length == 1) {
```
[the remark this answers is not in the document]
cac03: I replaced the check with `MethodCallUtils#isVarArgCall`

Now, the inspection does not give a warning for the following code:

```java
class Test2 {
    void bar() {
    }
}

public class Main2 {
    public static void main(String[] args) throws Throwable {
        MethodHandles.Lookup lookup = MethodHandles.lookup();
        MethodHandle methodHandle = lookup
                .findVirtual(Test2.class, "bar", MethodType.methodType(Test2.class, List.of(new Class<?>[]{int.class, long.class})));
        methodHandle.invoke(new Test2());
    }
}
```

Maybe, we can not support this? Because:

1. To me it seems too artificial to support
2. Arrays could be mutated if they are created previously and passed as a variable to the call

---

> 2. Arrays could be mutated if they are created previously and passed as a variable to the call

A similar issue exists for reflection inspection. Consider the following code:

```java
class Test3 {
    void bar(int unused) {
        System.out.println("bar(int)");
    }
}

public class Main3 {
    public static void main(String[] args) throws Throwable {
        Class<?> clazz = Test3.class;
        Class<?>[] parameterTypes = {long.class};
        parameterTypes[0] = int.class;
        Method method = clazz.getDeclaredMethod("bar", parameterTypes);
        //                                             ^^ Cannot resolve method 'bar' with specified argument types
        method.invoke(new Test3(), 10);
    }
}
```

IDEA produces a warning, thought the `bar` method exists

Should it be fixed too? Maybe, if this MR would merged, I can submit another one?

java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangReflectHandleInvocationChecker.java
```diff
@@ -247,41 +255,100 @@ else if (invokeParent instanceof PsiVariable) {
...
+    return null;
+  }
 
-      if (isCallToMethod(typeDefinitionCall, JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE)) {
-        final PsiExpression[] arguments = typeDefinitionCall.getArgumentList().getExpressions();
-        if (arguments.length != 0) {
-          return ContainerUtil.map(arguments, argument -> (() -> getReflectiveType(argument)));
-        }
+  @Nullable
+  private static List<Supplier<ReflectiveType>> getLazyMethodSignatureForGenericMethodType(
```
cac03, line 264: This change is an extract method

cac03:
Thanks for the comments!

1. I've rebased the PR against the latest master
2. Proceeded with review comments in a new commit. Please, let me know, if a preferred way is to squash them instead

And, please let me know if there is anything confusing in this PR, I'll address the issues

amaembo (approved):

java/java-analysis-impl/src/com/intellij/psi/impl/source/resolve/reference/impl/JavaReflectionReferenceUtil.java
```diff
@@ -43,6 +50,34 @@
 
   public static final String METHOD_TYPE = "methodType";
   public static final String GENERIC_METHOD_TYPE = "genericMethodType";
+  private static final CallMatcher LIST_FACTORY = anyOf(
+    staticCall(JAVA_UTIL_LIST, "of"),
+    staticCall(JAVA_UTIL_ARRAYS, "asList")
+  );
+  private static final CallMatcher METHOD_TYPE_MATCHER = staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE);
+  public static final CallMatcher METHOD_TYPE_WITH_METHOD_TYPE_MATCHER =
+    staticCall(JAVA_LANG_INVOKE_METHOD_TYPE, METHOD_TYPE).parameterTypes(JAVA_LANG_CLASS, JAVA_LANG_INVOKE_METHOD_TYPE);
```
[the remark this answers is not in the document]
amaembo: I will fix this by myself.

amaembo:
Changes are pushed to internal repository and will appear on GitHub soon.

PR #1799, opened 2021-10-25
FIR/UAST: revisit data/resolution of bitwise operators

jsjeon:
Expression evaluation of test data `Bitwise` suffers from div-by-zero error. It turns out that variable `one`, which is used as denominator and supposed to be literally `1`, was `0`. The value of another variable `zero` was not `0` either. 1st commit adjusts integer literals in the test data so that those variables can have literally `1` or `0`.

Next, 2nd commit rewrites FIR UAST's logic for bitwise operator resolution. As a result, all aspects (render/log/type/value) of that test data are identical.


darthorimar:
Merged, thanks :)

PR #1801, opened 2021-10-26
Try at an fix for the issue that (at least on macOS) external diff co…

monkeydom:
…mmand line and merge utilities that also accept a pipe in as file need to have their stdin closed. Otherwise the process hangs until the host app quits. As a reproduction example you can use e.g. SubEthaEdit's see tool ( https:://subethaedit.net ) or if you are quick ksdiff from kaleidoscope (although we fixed that on our end too by recognizing your caller process and not waiting as special consideration)

This is just identifiyng the places and doing what should be done in general terms, I do have a lack of build environment to verify this works correctly and or reliably or even compiles.


dmitry-zhuravlev:
See https://youtrack.jetbrains.com/issue/IDEA-281222

PR #1812, opened 2021-10-28
Allow customizing ActionMenu UI

TheNewJavaman:
Currently, the UIs for `ActionMenu` and `ActionMenuItem` are hardcoded, unlike most other Swing components in IntelliJ. This change allows plugin developers to provide their own components for action menus (File, Edit, View, etc.).

I was building my own plugin that provides custom components and noticed this issue, so I thought I would help fix it. This is my first time contributing to IntelliJ, so please let me know if I did something wrong. I made sure to read the guidelines and will sign the contributor license right after creating this PR.

I wasn't sure if tests should be included -- based on OpenJDK's Swing tests, it's not necessary. Please also let me know if I should merge to a different branch.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/actionSystem/impl/ActionMenu.java
+++ b/platform/platform-impl/src/com/intellij/openapi/actionSystem/impl/ActionMenu.java
@@ -1,4 +1,4 @@
-// Copyright 2000-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+// Copyright 2000-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
 package com.intellij.openapi.actionSystem.impl;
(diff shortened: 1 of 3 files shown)


TheNewJavaman:
@develar Any updates? It's been two weeks

TheNewJavaman:
@ybedrov Pinging you because you assigned someone to this PR several weeks ago. Do you have any updates on this?

stachenov:
Now that I got assigned to it and see it for the first time...

I see that `BegMenuItemUI` isn't a part of the public API anymore. Opening it up would mean supporting its API forever, which isn't something we're willing to commit to, given the fragile nature of UI classes instead. For example, if the overall look changes (happened twice during the last 3 years), then all extensions will start breaking because their expectations will break.

If this is still relevant, please create a YouTrack issue explaining your use case, so we can start from there, and maybe we can think of a way to support your use case in a simpler way. For example, we could add some theme keys to make menus more customizable.

PR #1814, opened 2021-10-29
FIR IDE: refactor KotlinFindUsagesSupport...

tgeng:
so that analyse can be done once for the entire container element rather
than on each element individually.

### DIFF
--- a/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/findUsages/KotlinFindUsagesSupportFirImpl.kt
+++ b/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/findUsages/KotlinFindUsagesSupportFirImpl.kt
@@ -9,41 +9,51 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiMethod
 import com.intellij.psi.PsiReference
 import com.intellij.psi.search.GlobalSearchScope
-import org.jetbrains.kotlin.idea.KotlinBundle
-import org.jetbrains.kotlin.idea.core.util.showYesNoCancelDialog
+import com.intellij.util.Processor
 import org.jetbrains.kotlin.analysis.api.analyseInModalWindow
 import org.jetbrains.kotlin.analysis.api.symbols.KtCallableSymbol
-import org.jetbrains.kotlin.lexer.KtTokens
-import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.analysis.api.symbols.markers.KtNamedSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.markers.KtSymbolWithKind
+import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.core.util.showYesNoCancelDialog
 import org.jetbrains.kotlin.idea.refactoring.CHECK_SUPER_METHODS_YES_NO_DIALOG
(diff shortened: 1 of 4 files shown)


plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/findUsages/KotlinFindUsagesSupportFirImpl.kt
```diff
@@ -9,22 +9,26 @@ import com.intellij.psi.PsiElement
...
 
 class KotlinFindUsagesSupportFirImpl : KotlinFindUsagesSupport {
-    override fun isCallReceiverRefersToCompanionObject(element: KtElement, companionObject: KtObjectDeclaration): Boolean {
-        return false
+    override fun <T> analyseIfCallReceiverRefersToCompanionObject(
+        analyseScope: KtElement,
+        companionObject: KtObjectDeclaration,
+        block: (isCallReceiverRefersToCompanionObject: (KtElement) -> Boolean) -> T
+    ): T {
+        return block { false }
```
jsjeon, line 31: Somehow we need to clarify/specify that some of `block` will be executed while some are not, depending on what `isCallReceiverRefersToCompanionObject` returns? For example, this `block { false }` looks like `block` will _not_ be executed no matter what. :\

tgeng: I am not sure I understand what you would like me to do here. This code is incomplete for now. Let me add some TODO to make this clear.

plugins/kotlin/frontend-independent/src/org/jetbrains/kotlin/idea/findUsages/KotlinFindUsagesSupport.kt
```diff
@@ -29,14 +34,18 @@ interface KotlinFindUsagesSupport {
...
-        fun getSuperMethods(declaration: KtDeclaration, ignore: Collection<PsiElement>?) : List<PsiElement> =
+        fun getSuperMethods(declaration: KtDeclaration, ignore: Collection<PsiElement>?): List<PsiElement> =
             getInstance(declaration.project).getSuperMethods(declaration, ignore)
 
         fun sourcesAndLibraries(delegate: GlobalSearchScope, project: Project): GlobalSearchScope =
             getInstance(project).sourcesAndLibraries(delegate, project)
     }
 
-    fun isCallReceiverRefersToCompanionObject(element: KtElement, companionObject: KtObjectDeclaration): Boolean
+    fun <T> analyseIfCallReceiverRefersToCompanionObject(
```
darthorimar, line 44: Name and semantic of the function is not very clear. Let's just move and make abstract to the `KotlinFindUsagesSupport` the whole `KotlinFindClassUsagesHandler.processCompanionObjectInternalReferences`

darthorimar (changes requested):

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/findUsages/KotlinFindUsagesSupportFirImpl.kt
tgeng: Per Ilya's suggestion, I have removed this and make the entire `processCompanionObjectInternalReferences` abstract.

darthorimar:
Pushed, thank you!

PR #1816, opened 2021-10-30
FIR Completion: correctly complete super member

tgeng:
Consider the following:

```
public interface A {
    public fun foo(i: Int) {}
}

public interface B {
    public fun foo(i: Int) {}
}

fun A.bar() {}

fun B.bar() {}

class C : A, B {
    override fun foo(j: Int) {
        super.<caret>
    }
}
```

Currently both FIR offers one `foo` entry. After completion, the code
becomes `super.foo()` which is ambiguous and cause a compiler error.
FE1.0 suffers from the same problem.

In addition, it offers `bar`, which is not valid for `super`

This fix does the following
1. offer two `foo` entries with tail text to differentiate them.
2. insert `<A>` or `<B>` at the end of `super` after completion so the
   completed code is correct
3. No longer offer extension functions on super (`bar`) in the example.

### DIFF
--- a/plugins/kotlin/completion/tests/testData/basic/common/ambiguousSuperMethod.kt
+++ b/plugins/kotlin/completion/tests/testData/basic/common/ambiguousSuperMethod.kt
@@ -1,0 +1,23 @@
+// FIR_COMPARISON
+interface A {
+    fun foo(i: Int) {}
+}
+
+interface B {
+    fun foo(i: Int) {}
+}
+
+fun A.bar() {}
+fun B.bar() {}
+
+class C : A, B {
+    override fun foo(j: Int) {
+        super.fo<caret>
+    }
+}
+
+// EXIST: { lookupString:"foo", tailText:"(i: Int)" }
+// EXIST: { lookupString:"foo", tailText:"(i: Int)" }
+// EXIST: { lookupString:"foo", tailText:"(i)" }
+// EXIST: { lookupString:"foo", tailText:"(i)" }
+// ABSENT: bar
--- a/plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethod.kt
+++ b/plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethod.kt
@@ -1,0 +1,18 @@
+// FIR_COMPARISON
+package foo.bar
+interface A {
+    fun foo(i: Int) {}
+}
+
+interface B {
+    fun foo(i2: Int) {}
+}
+
+class C : A, B {
+    override fun foo(j: Int) {
+        super.fo<caret>
+    }
+}
+
+// ELEMENT: "foo"
+// TAIL_TEXT: "(i: Int)"
--- a/plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethodWithArgument.kt
+++ b/plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethodWithArgument.kt
@@ -1,0 +1,18 @@
+// FIR_COMPARISON
+package foo.bar
+interface A {
+    fun foo(i: Int) {}
+}
+
+interface B {
+    fun foo(i2: Int) {}
+}
+
+class C : A, B {
+    override fun foo(j: Int) {
+        super.fo<caret>
+    }
+}
+
+// ELEMENT: "foo"
+// TAIL_TEXT: "(i)"
--- a/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirSuperMemberCompletionContributor.kt
+++ b/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirSuperMemberCompletionContributor.kt
@@ -3,45 +3,146 @@
 package org.jetbrains.kotlin.idea.completion.contributors
 
 import com.intellij.psi.util.parentsOfType
+import com.intellij.refactoring.suggested.createSmartPointer
 import org.jetbrains.kotlin.analysis.api.KtAnalysisSession
 import org.jetbrains.kotlin.analysis.api.symbols.KtCallableSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KtFunctionLikeSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KtFunctionSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KtSyntheticJavaPropertySymbol
+import org.jetbrains.kotlin.analysis.api.symbols.markers.KtSymbolWithModality
+import org.jetbrains.kotlin.analysis.api.types.KtIntersectionType
+import org.jetbrains.kotlin.analysis.api.types.KtType
+import org.jetbrains.kotlin.analysis.api.types.KtUsualClassType
+import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.idea.completion.ItemPriority
 import org.jetbrains.kotlin.idea.completion.checkers.CompletionVisibilityChecker
 import org.jetbrains.kotlin.idea.completion.context.FirBasicCompletionContext
-import org.jetbrains.kotlin.idea.completion.context.FirNameReferencePositionContext
+import org.jetbrains.kotlin.idea.completion.context.FirSuperReceiverNameReferencePositionContext
 import org.jetbrains.kotlin.idea.completion.contributors.helpers.collectNonExtensions
 import org.jetbrains.kotlin.idea.completion.lookups.CallableInsertionOptions
 import org.jetbrains.kotlin.idea.completion.lookups.CallableInsertionStrategy
 import org.jetbrains.kotlin.idea.completion.lookups.detectImportStrategy
 import org.jetbrains.kotlin.idea.completion.weighers.WeighingContext
 import org.jetbrains.kotlin.idea.completion.weighers.WeighingContext.Companion.createWeighingContext
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.name.StandardClassIds
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtSuperExpression
+import org.jetbrains.kotlin.utils.addIfNotNull
 
 internal class FirSuperMemberCompletionContributor(
     basicContext: FirBasicCompletionContext,
     priority: Int
-) : FirCompletionContributorBase<FirNameReferencePositionContext>(basicContext, priority) {
-    override fun KtAnalysisSession.complete(positionContext: FirNameReferencePositionContext) = with(positionContext) {
-        val superReceiver = explicitReceiver as? KtSuperExpression ?: return
+) : FirCompletionContributorBase<FirSuperReceiverNameReferencePositionContext>(basicContext, priority) {
+    override fun KtAnalysisSession.complete(positionContext: FirSuperReceiverNameReferencePositionContext) = with(positionContext) {
+        val superReceiver = positionContext.superExpression
         val expectedType = nameExpression.getExpectedType()
         val visibilityChecker = CompletionVisibilityChecker.create(basicContext, positionContext)
-        val possibleReceiverScope = superReceiver.getKtType()?.getTypeScope() ?: return
-        val nonExtensionMembers = collectNonExtensions(possibleReceiverScope, visibilityChecker, scopeNameFilter)
+        val superType = superReceiver.getKtType() ?: return
         val weighingContext = createWeighingContext(
             superReceiver,
             expectedType,
             emptyList(), // Implicit receivers do not match for this completion contributor.
             fakeKtFile
         )
-        collectDelegateCallToSuperMember(weighingContext, superReceiver, nonExtensionMembers)
+
+        val (nonExtensionMembers: Iterable<Pair<KtType, KtCallableSymbol>>, namesNeedDisambiguation: Set<Name>) = if (superType !is KtIntersectionType) {
+            val scope = superType.getTypeScope() ?: return
+            collectNonExtensions(scope, visibilityChecker, scopeNameFilter).map { superType to it }.asIterable() to emptySet()
+        } else {
+            getSymbolsAndNamesNeedDisambiguation(superType.conjuncts, visibilityChecker)
+        }
+        collectCallToSuperMember(superReceiver, nonExtensionMembers, weighingContext, namesNeedDisambiguation)
+        collectDelegateCallToSuperMember(weighingContext, superReceiver, nonExtensionMembers, namesNeedDisambiguation)
+
+    }
+
+    private fun KtAnalysisSession.getSymbolsAndNamesNeedDisambiguation(
+        superTypes: List<KtType>,
+        visibilityChecker: CompletionVisibilityChecker
+    ): Pair<List<Pair<KtType, KtCallableSymbol>>, Set<Name>> {
+        val allSymbols = mutableListOf<Pair<KtType, KtCallableSymbol>>()
+        val symbolsInAny = mutableSetOf<KtCallableSymbol>()
+        val symbolCountsByName = mutableMapOf<Name, Int>()
+        for (superType in superTypes) {
+            for (symbol in getNonExtensionsMemberSymbols(superType, visibilityChecker)) {
+                // Abstract symbol does not participate completion.
+                if (symbol !is KtSymbolWithModality || symbol.modality == Modality.ABSTRACT) continue
+
+                // Unlike typical diamond cases, calls to method of `Any` always do not need extra qualification.
+                if (symbol.callableIdIfNonLocal?.classId == StandardClassIds.Any) {
+                    if (symbol in symbolsInAny) continue
+                    symbolsInAny.add(symbol)
+                }
+
+                allSymbols.add(superType to symbol)
+                val name = symbol.callableIdIfNonLocal?.callableName ?: continue
+                symbolCountsByName[name] = (symbolCountsByName[name] ?: 0) + 1
+            }
+        }
+
+        val nameNeedDisambiguation = mutableSetOf<Name>()
+        for ((name, count) in symbolCountsByName) {
+            if (count > 1) {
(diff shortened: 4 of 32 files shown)


plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -167,11 +184,79 @@ internal open class FirCallableCompletionContributor(
...
+        expectedType: KtType?,
+    ) {
+        if (superExpressionType !is KtIntersectionType) {
+            val nonExtensionMembers = getNonExtensionsMemberSymbols(superExpressionType, visibilityChecker)
+            collectNonExtensionLookUpElements(nonExtensionMembers.toList(), expectedType)
+            return
+        }
+        val superTypes = superExpressionType.conjuncts
+        val allSymbols = mutableListOf<KtCallableSymbol>()
+        val symbolsByName = mutableMapOf<Name, AtomicInteger>()
```
darthorimar, line 223: `AtomicInteger` introduces overhead and should be used only in multi-threaded environments. Please, replace it with just integer.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -167,11 +184,79 @@ internal open class FirCallableCompletionContributor(
...
+                allSymbols.add(symbol)
+            }
+        }
+
+        val nameNeedDisambiguation = mutableSetOf<Name>()
+        for ((name, count) in symbolsByName) {
+            if (count.get() > 1) {
+                nameNeedDisambiguation.add(name)
+            }
+        }
```
darthorimar, line 237: All that this piece of code does is collects `nameNeedDisambiguation`. It should be extracted to function

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -44,8 +50,9 @@ internal open class FirCallableCompletionContributor(
         is KtExtensionApplicabilityResult.NonApplicable -> null
     }
 
-    protected fun KtAnalysisSession.getOptions(symbol: KtCallableSymbol): CallableInsertionOptions =
-        CallableInsertionOptions(detectImportStrategy(symbol), getInsertionStrategy(symbol))
+    protected fun KtAnalysisSession.getOptions(symbol: KtCallableSymbol): CallableInsertionOptions? {
+        return CallableInsertionOptions(detectImportStrategy(symbol), getInsertionStrategy(symbol) ?: return null)
```
darthorimar, line 54: Please, do not use inline `return/break/continue` statements. Introduce a variable for that

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -94,24 +101,32 @@ internal open class FirCallableCompletionContributor(
...
+        val availableNonExtensions =
+            collectNonExtensions(implicitScopes, visibilityChecker, scopeNameFilter) { filter(it) }
         val extensionsWhichCanBeCalled = collectSuitableExtensions(implicitScopes, extensionChecker, visibilityChecker)
 
-        availableNonExtensions.forEach { addCallableSymbolToCompletion(expectedType, it, getOptions(it)) }
+        availableNonExtensions.forEach {
+            addCallableSymbolToCompletion(
+                expectedType,
+                it,
+                getOptions(it) ?: return@forEach,
```
darthorimar, line 112: Please, do not use inline return/break/continue statements. Introduce a variable for that

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -94,24 +101,32 @@ internal open class FirCallableCompletionContributor(
...
+            getExtensionOptions(symbol, applicable)?.let { addCallableSymbolToCompletion(expectedType, symbol, it, getPriority(symbol)) }
         }
 
         if (shouldCompleteTopLevelCallablesFromIndex) {
             val topLevelCallables = indexHelper.getTopLevelCallables(scopeNameFilter)
             topLevelCallables.asSequence()
                 .map { it.getSymbol() as KtCallableSymbol }
                 .filter { with(visibilityChecker) { isVisible(it) } }
-                .forEach { addCallableSymbolToCompletion(expectedType, it, getOptions(it)) }
+                .forEach { addCallableSymbolToCompletion(expectedType, it, getOptions(it) ?: return@forEach, getPriority(it)) }
```
darthorimar, line 125: Please, do not use inline return/break/continue statements. Introduce a variable for that

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -94,24 +101,32 @@ internal open class FirCallableCompletionContributor(
...
         }
 
         collectTopLevelExtensionsFromIndices(implicitReceiversTypes, extensionChecker, visibilityChecker)
-            .forEach { addCallableSymbolToCompletion(expectedType, it, getOptions(it)) }
+            .forEach { addCallableSymbolToCompletion(expectedType, it, getOptions(it) ?: return@forEach, getPriority(it)) }
```
darthorimar, line 129: Please, do not use inline return/break/continue statements. Introduce a variable for that

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -125,13 +140,15 @@ internal open class FirCallableCompletionContributor(
...
-                        memberSymbol,
-                        CallableInsertionOptions(ImportStrategy.DoNothing, getInsertionStrategy(memberSymbol))
-                    )
-                }
+                collectNonExtensions(symbol.getStaticMemberScope(), visibilityChecker, scopeNameFilter)
+                    .forEach { memberSymbol ->
+                        addCallableSymbolToCompletion(
+                            expectedType,
+                            memberSymbol,
+                            CallableInsertionOptions(ImportStrategy.DoNothing, getInsertionStrategy(memberSymbol) ?: return@forEach),
```
darthorimar, line 148: Please, do not use inline return/break/continue statements. Introduce a variable for that

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -153,7 +170,7 @@ internal open class FirCallableCompletionContributor(
             .filter { with(visibilityChecker) { isVisible(it) } }
             .filter { filter(it) }
             .forEach { callable ->
-                addCallableSymbolToCompletion(expectedType, callable, getOptions(callable))
+                addCallableSymbolToCompletion(expectedType, callable, getOptions(callable) ?: return@forEach, getPriority(callable))
```
darthorimar, line 173: Please, do not use inline return/break/continue statements. Introduce a variable for that

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -167,11 +184,79 @@ internal open class FirCallableCompletionContributor(
         val typeOfPossibleReceiver = explicitReceiver.getKtType() ?: return
         val possibleReceiverScope = typeOfPossibleReceiver.getTypeScope() ?: return
 
+        if (explicitReceiver is KtSuperExpression) {
```
darthorimar, line 187: Please, create additional position context for super references and move all logic related to that into separate completion contributor to make the code more readable

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirSuperMemberCompletionContributor.kt
```diff
@@ -6,54 +6,46 @@ import com.intellij.psi.util.parentsOfType
...
+     *     }
+     *   }
+     * }
+     * ```
+     * The map would contain
+     *
+     * * Runnable.run -> A.run
+     * * Callable.call -> <anonymous object>.call
+     */
+    lateinit var superFunctionToContainingFunction: Map<KtCallableSymbol, KtNamedFunction>
```
darthorimar, line 40: All completion contributors should be stateless. Why do you need to extract that local variable to a class member?

darthorimar (changes requested):

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirSuperMemberCompletionContributor.kt
tgeng: I need to pass some additional data (`superFunctionToContainingFunction`) to `getInsertionStrategy`. The contributor is still stateless because `superFunctionToContainingFunction` is always updated in `complete` before used in `getInsertionStrategy`. Let me try refactoring this code to move all super related completion to this class instead. Then I probably won't need juggle all these data.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
tgeng: OK, will do!

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
tgeng: Sounds good! Thanks!

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
tgeng: OK. After the refactoring I no longer need to make `getOptions` return nullable values.

plugins/kotlin/completion/tests/testData/basic/common/ambiguousSuperMethod.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class C : A, B {
+    override fun foo(j: Int) {
+        super.fo<caret>
+    }
+}
+
+// EXIST: { lookupString:"foo", tailText:"(i: Int)" }
+// EXIST: { lookupString:"foo", tailText:"(i: Int)" }
+// EXIST: { lookupString:"foo", tailText:"(i)" }
```
darthorimar, line 20: Should not here be "j" as parameter instead?

plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethodWithArgument.kt
```diff
@@ -0,0 +1,18 @@
...
+    fun foo(i2: Int) {}
+}
+
+class C : A, B {
+    override fun foo(j: Int) {
+        super.fo<caret>
+    }
+}
+
+// ELEMENT: "foo"
```
darthorimar, line 17: What does this test check?

plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethod.kt
```diff
@@ -0,0 +1,18 @@
...
+    fun foo(i2: Int) {}
+}
+
+class C : A, B {
+    override fun foo(j: Int) {
+        super.fo<caret>
+    }
+}
+
+// ELEMENT: "foo"
```
darthorimar, line 17: What does this test check?

darthorimar (changes requested):
Screenshot 2021-11-05 at 4 09 26 PM

* Some items are duplicated in completion
* First 2 items which have direct argument present should have this argument presented in color (not grey) so they can be distinguished from later options.
* When choosing `hashCode()` from the list the following  is inserted 
```
super<kotlin.Any>.hashCode() // error: NOT_A_SUPERTYPE
```

plugins/kotlin/completion/tests/testData/basic/common/ambiguousSuperMethod.kt
tgeng: It should. This is a bug in FE1.0.

tgeng:
> Screenshot 2021-11-05 at 4 09 26 PM
> 
> * Some items are duplicated in completion
> * First 2 items which have direct argument present should have this argument presented in color like the `foo` name( white on the screenshot) so they can be distinguished from later options.
> * When choosing `hashCode()` from the list the following  is inserted
> 
> ```
> super<kotlin.Any>.hashCode() // error: NOT_A_SUPERTYPE
> ```

Ah, I totally missed these cases. Also the current logic does not work for the following
```
interface A {
    fun foo() {}
}

interface A2 : A 

interface B {
    fun foo() {}
}
class C : A2, B {
    fun test() {
        super.fo<caret> // completes with `A` instead of `A2`.
    }
```
Let me fix these issues and add tests.

plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethodWithArgument.kt
tgeng: This is to make sure the type qualifier is present when completing super member with call args.

plugins/kotlin/completion/tests/testData/handlers/basic/AmbiguousSuperMethod.kt
tgeng: This is to make sure the type qualifier is present when completing super member without call args.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirSuperMemberCompletionContributor.kt
```diff
@@ -3,38 +3,115 @@
...
+        }
+        collectCallToSuperMember(superReceiver, nonExtensionMembers, expectedType, namesNeedDisambiguation)
+        collectDelegateCallToSuperMember(superReceiver, nonExtensionMembers, expectedType, namesNeedDisambiguation)
+
+    }
+
+    private fun KtAnalysisSession.getSymbolsAndNamesNeedDisambiguation(
+        superTypes: List<KtType>,
+        visibilityChecker: CompletionVisibilityChecker
+    ): Pair<Iterable<KtCallableSymbol>, Set<Name>> {
```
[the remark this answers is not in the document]
tgeng: OK, I have changed the implementation to store both the super type and the callable symbol. Now it does not matter what I use here. After my change I am using a list since there is no need to check for duplicates (since member traversal should not return duplicates.)

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirSuperMemberCompletionContributor.kt
```diff
@@ -84,10 +161,31 @@ internal class FirSuperMemberCompletionContributor(
...
         }
     }
+
+    private fun KtAnalysisSession.wrapWithDisambiguationIfNeeded(
+        insertionStrategy: CallableInsertionStrategy,
+        callableSymbol: KtCallableSymbol,
+        namesNeedDisambiguation: Set<Name>,
+        superReceiver: KtSuperExpression
+    ): CallableInsertionStrategy {
+        var insertionStrategy1 = insertionStrategy
```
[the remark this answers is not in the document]
tgeng: I don't remember what happened but I think it was probably generated by refactoring feature and I wasn't paying attention to the result. 😞

tgeng:
I have added commit#3 trying to force the tail text color. But it appears to be a bug in IntelliJ itself since the way the current API works does not make much sense to me.

darthorimar:
> I have added commit#3 trying to force the tail text color. But it appears to be a bug in IntelliJ itself since the way the current API works does not make much sense to me.
Unfortunately, it is impossible to change IntelliJ code in kt-212-branch. All changes outside kotlin plugin will be lost

tgeng:
> > I have added commit#3 trying to force the tail text color. But it appears to be a bug in IntelliJ itself since the way the current API works does not make much sense to me.
> 
> Unfortunately, it is impossible to change IntelliJ code in kt-212-branch. All changes outside kotlin plugin will be lost

Ah.... I was hoping `platform/analysis-api` can be counted as kotlin. But I guess kotlin only include `plugins/kotlin`. Then there doesn't seem to be a way to force the color 😞

plugins/kotlin/completion/tests/testData/basic/common/ambiguousSuperMethod.kt
darthorimar: For the `ambiguousSuperMethod.fir.kt ` there is also `i` instead of `j`, but debug idea instance shows correct `j` here

tgeng: Ah... I missed `// FIR_COMPARISON`, which is why the test is passing...

darthorimar:
> But I guess kotlin only include plugins/kotlin. Then there doesn't seem to be a way to force the color
let's return to this question then we make it possible to change intellij sources :)

darthorimar:
The PR is merged, thank you!

PR #1818, opened 2021-11-01
[EE-IR] Remove reference to JvmNameProvider.

madsager:
The IR version of the code fragment compiler for evaluate
expression sets up the compiler and needs a name provider.

The JvmNameProvider is being removed from the compiler source
base and NameProvider.DEFAULT should be used.

JvmNameProvider will be unused when this compiler change
lands:

   https://github.com/JetBrains/kotlin/pull/4629


nikita-nazarov:
Thanks! Merged fb1a468e15b2a42536b647a6b82a29d07e1db40c

PR #1820, opened 2021-11-02
KTIJ-20041 Do not highlight outer function when caret on return from the anonymous function

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20041

スクリーンショット 2021-11-02 15 33 02


klunnii:
Thank you! Merged.

PR #1821, opened 2021-11-02
[gradle] Run main methods with run context action inside 'buildSrc'

fp7:
There are some problems when a main method within the special buildSrc
module is run by the run context action.

1. IJ install the main task with a init-script. But they are not applied
   to the buildSrc module
2. Even if 1. would be applied the path of the installed task would be
   ':buildSrc:<class.main>()'. But :buildSrc is not a valid project path.
3. The buildSrc module's classes and dependencies are put onto the
   classpath of the buildscript, but the injected JavaExec task
   does not have access to it.

This change solves the problem, by installing the task into the root
project when the main class is in the buildSrc module and uses the
buildscript default configuration and hardcoded name of the output
of the buildSrc module as classpath.


nskvortsov:
This part of IDEA has been reworked dramatically since your PR. 
Is the issue still present?

If yes, please provide a minimal reproducible project, so we can file an issue in our tracker.

PR #1822, opened 2021-11-03
KTIJ-20037 WRONG_ANNOTATION_TARGET: quickfix to replace unspecified annotation target at use side with applicable targets

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20037


klunnii:
Merged. Thank you!

PR #1824, opened 2021-11-03
KTIJ-20031 Add quickfix for IS_ENUM_ENTRY to replace type check with comparison

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20031


klunnii:
Merged. Thank you!

PR #1825, opened 2021-11-03
FIR UAST: return light element if reference resolves to source

jsjeon:
That matches FE 1.0 behavior: https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin/src/org/jetbrains/uast/kotlin/internal/kotlinInternalUastUtils.kt#L294 To determine whether the resolved reference is source, `getKtModule` and checking `KtSourceModule` are used. As shown below that line, though, we need to handle reference resolved to deserialized source, which I left as a TODO.


darthorimar:
LGTM, merged. Thank you!

PR #1826, opened 2021-11-03
Allow customizing the number of partial commit fetch attempts

AlexeyGy:
This change adds a System property which enables users to customize how many retries of partial commit fetches (each increasing the limit of to-be-fetched commits by a factor of 5) are done.

See [IDEA-281878](https://youtrack.jetbrains.com/issue/IDEA-281878) for background.


juliabeliaeva:
As discussed in the [youtrack issue](https://youtrack.jetbrains.com/issue/IDEA-281878), a different fix was implemented (see [131df51fa15c1a5177e8a7900e5e12d7573cce02](https://github.com/JetBrains/intellij-community/commit/131df51fa15c1a5177e8a7900e5e12d7573cce02)).

PR #1831, opened 2021-11-05
FIR UAST: run fixture tests with JDK and runtime

jsjeon:
One test failed because `mutableListOf` is not provided. 🤦


jsjeon:
I saw this is already merged, thank you!

PR #1833, opened 2021-11-09
FIR/UAST: introduce resolve everything tests

jsjeon:
Just noticed that there are so-called resolve "everything" test in FE 1.0 UAST where all `UResolvable`'s `resolve()` results are recorded. In 2nd commit, a core util to record such `resolve()` results per every `UResolvable` is utilized. Then, 3rd commit introduced tests for both UAST plugins.

It turns out that light elements' rendering can be different since each plugin's LC parts (i.e. ULC and FIR LC) have their own `toString()` implementations. 4th commit adds some massage logic to mitigate the gap. The next big chunk of differences is, as expected, resolution results for library source/deserialized. I'll work on those next.

1st commit is a simple clean up, which should have been part of #1799.


plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastResolveEverythingTestBase.kt
```diff
@@ -32,4 +34,31 @@ interface UastResolveEverythingTestBase : UastPluginSelection, UastFileCompariso
...
+        if (!isFirUastPlugin || element !is KtLightElement<*, *>) return str
+        // NB: all declarations in FIR LC are internal, so type checking won't work.
+        TAG.forEach { (firLC, uLC) -> str = str.replace(firLC, uLC) }
+        REGEX.forEach { (firRegex, uLC) -> str = firRegex.replace(str, uLC) }
+        return str
+    }
+
+    companion object {
+        private val TAG: Map<String, String> = mapOf(
+            "FirLightAnnotationClassSymbol" to "KtUltraLightClass",
```
darthorimar, line 49: Let's do not relay on implementation details :) We may use some more general names here, like KtLightClass or or KtLightMethod. Or maybe use some fancy string representation like `Kotlin Light Class`

darthorimar:
The  resolve everything seems to be rather hard to work with and debug. Are they really useful? Maybe some better approach may be used?
cc @xiexed

plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastResolveEverythingTestBase.kt
jsjeon: I see. In that case, you meant mapping each LC tag to one central, impl-agnostic name? E.g., "FirLightAnnotationClassSymbol" -> "Kotlin Light Class", "KtUltraLightClass" -> "Kotlin Light Class", and so on?

darthorimar: Yes

jsjeon:
> The resolve everything seems to be rather hard to work with and debug. Are they really useful? Maybe some better approach may be used?

At least there is one single entry point to start debugging: `resolveToDeclaration` :) Of course, I have to admit that underlying behavior differences are not that trivial to debug.

However, I can insist these are really useful. Although we're seeing many differences for now, ironically, now I'm more confident on FIR UAST. :D I meant, UAST tree structures (render/log), values if evaluatabe, types, these are well tested (even though we still have some differences), but I felt lack of resolution tests.

There are alternatives, of course. I planned to add more tests to https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastResolveApiTestBase.kt where some manual API exercises would be done per different test input, which sounds tedious and may not cover everything. https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastResolveApiFixtureTestBase.kt would be another option. These would be useful/easy-to-add e.g., when I encounter issues while integrating FIR UAST into Android Lint. But, this sounds too late, and would miss that release cycle, and wait for one more release for fixes.

That is, having this test suit of resolving everything is kinda proactive, and helps me/us figure out issues earlier than later. One example, I just learned that UAST resolve is supposed to return resolved light elements, not source-level `PsiElement`, whereas FIR UAST still returns the source-level `PsiElement` as a resolution result. This is another big difference, and I expect we can remove many noises too. Also, I found https://github.com/JetBrains/intellij-community/pull/1835 while debugging a couple differences, which I believe shows these are useful! :)

plugins/kotlin/uast/uast-kotlin-base/test/org/jetbrains/uast/test/common/kotlin/UastResolveEverythingTestBase.kt
jsjeon: Done at 2c1129d. Somehow, I decided to avoid space (` `) in tags.

jsjeon:
I accidentally(?) worked on top of [KOTLIN-CR-258](https://kotlin.jetbrains.space/p/kotlin/reviews/258/) and found subtle test updates: a8e75cf (which are technically wrong anyway, since they are still source-level `PsiElement` at the moment).

darthorimar:
No problems, I will merge KOTLIN-CR-258 first then :)

darthorimar (approved):

fedochet:
UPD: KOTLIN-CR-258 was merged 🎉

darthorimar:
merged, thank you!

PR #1834, opened 2021-11-10
FIR Completion: offer member from unstable smartcast

tgeng:
Also fix completion entry style.

### DIFF
--- a/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
+++ b/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
@@ -5,7 +5,6 @@ package org.jetbrains.kotlin.idea.completion.contributors
 import org.jetbrains.kotlin.analysis.api.KtAnalysisSession
 import org.jetbrains.kotlin.analysis.api.components.KtExtensionApplicabilityResult
 import org.jetbrains.kotlin.analysis.api.components.KtScopeContext
-import org.jetbrains.kotlin.analysis.api.impl.base.scopes.KtCompositeScope
 import org.jetbrains.kotlin.analysis.api.scopes.KtScope
 import org.jetbrains.kotlin.analysis.api.symbols.*
 import org.jetbrains.kotlin.analysis.api.types.KtFunctionalType
@@ -101,7 +100,32 @@ internal open class FirCallableCompletionContributor(
         val availableNonExtensions = collectNonExtensions(implicitScopes, visibilityChecker, scopeNameFilter) { filter(it) }
         val extensionsWhichCanBeCalled = collectSuitableExtensions(implicitScopes, extensionChecker, visibilityChecker)
 
-        availableNonExtensions.forEach { addCallableSymbolToCompletion(context, it, getOptions(it)) }
+        // TODO: consider relying on tower resolver when populating callable entries. For example
+        //  val Int.foo : ...
+        //  val Number.foo : ...
+        //  val String.foo : ...
+        //  fun String.test() {
+        //      with(1) {
+        //          val foo = ...
+        //          fo<caret>
+        //      }
+        //  }
+        //  completion should be able to all of the `foo` by inserting proper qualifiers:
+        //   foo -> foo
+        //   Int.foo -> this.foo
+        //   Number.foo -> (this as Number).foo
+        //   String.foo -> this@test.foo
+        //
+         availableNonExtensions
+            // skip shadowed variable or properties
+            .distinctBy {
+                when (it) {
+                    is KtVariableLikeSymbol -> it.name
+                    else -> it
+                }
+            }
+            .forEach { addCallableSymbolToCompletion(context, it, getOptions(it)) }
+
         // Here we can't rely on deduplication in LookupElementSink because extension members can have types substituted, which won't be
         // equal to the same symbols from top level without substitution.
         val extensionMembers = mutableSetOf<KtCallableSymbol>()
--- a/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
+++ b/plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
@@ -201,7 +201,32 @@ internal open class FirCallableCompletionContributor(
         extensionChecker: ExtensionApplicabilityChecker,
         visibilityChecker: CompletionVisibilityChecker,
     ) {
-        val typeOfPossibleReceiver = explicitReceiver.getKtType() ?: return
+        val smartCastInfo = explicitReceiver.getSmartCastInfo()
+        if (smartCastInfo?.isStable == false) {
+            // Collect members available from unstable smartcast as well.
+            collectDotCompletionForCallableReceiver(
+                smartCastInfo.smartCastType,
+                visibilityChecker,
+                implicitScopes,
+                extensionChecker,
+                context,
+                // Only offer the hint if the type is denotable.
+                smartCastInfo.smartCastType.takeIf { it.approximateToSuperPublicDenotable() == null }
+            )
+        }
+
+        val receiverType = explicitReceiver.getKtType() ?: return
+        collectDotCompletionForCallableReceiver(receiverType, visibilityChecker, implicitScopes, extensionChecker, context)
+    }
+
+    private fun KtAnalysisSession.collectDotCompletionForCallableReceiver(
+        typeOfPossibleReceiver: KtType,
+        visibilityChecker: CompletionVisibilityChecker,
+        implicitScopes: KtScope,
+        extensionChecker: ExtensionApplicabilityChecker,
+        context: WeighingContext,
+        explicitReceiverTypeHint: KtType? = null
+    ) {
         val possibleReceiverScope = typeOfPossibleReceiver.getTypeScope() ?: return
 
         val nonExtensionMembers = collectNonExtensions(possibleReceiverScope, visibilityChecker, scopeNameFilter) { filter(it) }
@@ -218,7 +243,7 @@ internal open class FirCallableCompletionContributor(
             .forEach {
                 if (it !in syntheticPropertyOrigins) {
                     // For basic completion, FE1.0 skips Java functions that are mapped to Kotlin properties.
-                    addCallableSymbolToCompletion(context, it, getOptions(it))
+                    addCallableSymbolToCompletion(context, it, getOptions(it), explicitReceiverTypeHint = explicitReceiverTypeHint)
                 }
             }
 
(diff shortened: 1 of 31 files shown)


plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCompletionContributorBase.kt
```diff
@@ -110,7 +115,40 @@ internal abstract class FirCompletionContributorBase<C : FirRawPositionCompletio
         }
         priority?.let { lookup.priority = it }
         applyWeighers(context, lookup, symbol, substitutor)
-        sink.addElement(lookup)
+        sink.addElement(lookup.boldIfImmediate())
+    }
+
+    private fun LookupElement.boldIfImmediate(): LookupElement {
```
darthorimar, line 121: This function and following enum `Style` are very specific to callable completion and should not be in `FirCompletionContributorBase`.  Let's move it (and it application) to the place which is more specific to callable completion.

darthorimar (changes requested):
Screenshot 2021-11-10 at 7 30 12 PM

Choosing the `v` option leads to the error. Can we insert a type cast there in a case of unstable smartcast: ` (pair.first as C).v`?

darthorimar:
Also, the following tests started to fail:
```
HighLevelJvmBasicCompletionTestGenerated$Common$BoldOrGrayed.testNonPredictableSmartCastWithWrongReturnType
HighLevelJvmBasicCompletionTestGenerated$Common$Shadowing.testLocals1
HighLevelJvmBasicCompletionTestGenerated$Common$Shadowing.testLocals2
HighLevelJvmBasicCompletionTestGenerated$Java$BoldOrGrayed.testNonPredictableSmartCast2
```

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCompletionContributorBase.kt
tgeng: Regarding placement of the styling logic, it would always be needed whenever `addCallableSymbolToCompletion` is invoked, so it seems this is the right place for such logic. Moving it to `FirCallableCompletionContributor` seems inappropriate because this is also needed by `FirSuperMemberCompletionContributor`, which does not inherit `FirCallableCompletionContributor` (and it can't inherit that because they are using different completion context.)

On the other hand, I do not like having this styling logic depending on `CallableWeigher` since it appears they should be totally decoupled from each other. So maybe we should move the logic computing `callableWeight` somewhere else. But then, I don't know where it should go to. Also, doing it that way may make future reader of this code wonder: why isn't the logic computing callable weight not inside the `CallableWeigher`?

In addition, in order to insert type cast, it seems natural to do it also at this place because the computation done in `CallableWeigher` already tells us whether a type cast is needed.  On the other hand, it also seems consistent to implement type cast insertion as a new insertion strategy just like inserting super type qualifier. But that means we need to duplicate the logic in `CallableWeigher` to check if receiver type matches, which can be expensive. What do you think?

tgeng: I just pushed another update with a tentative implementation of inserting type cast on demand. TBH I am not a big fan of the way it's implemented but I can't think of a better way to do it unless we restructure a lot of things. Let me know what you think and I am happy to apply changes to make things cleaner.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -101,7 +100,16 @@ internal open class FirCallableCompletionContributor(
         val availableNonExtensions = collectNonExtensions(implicitScopes, visibilityChecker, scopeNameFilter) { filter(it) }
         val extensionsWhichCanBeCalled = collectSuitableExtensions(implicitScopes, extensionChecker, visibilityChecker)
 
-        availableNonExtensions.forEach { addCallableSymbolToCompletion(context, it, getOptions(it)) }
+        availableNonExtensions
+            // skip shadowed variable or properties
```
darthorimar, line 104: Hm. Is it correct?
We should see most specific candidate here instead of the first from the scope.

Eg:
```
val Number.a: Int get() = 10
val Int.a: Int get() = 10

fun x() {
    with(1) {
        a<caret> // here we should resolve to Int.a as more specific
    }
}
```
Here we should use tower resolver logic from compiler somehow.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/weighers/CallableWeigher.kt
```diff
@@ -113,7 +148,11 @@ internal object CallableWeigher {
         // TODO: FE1.0 has logic that uses `null` for receiverIndex if the symbol matches every actual receiver in order to "prevent members
         //  of `Any` to show up on top". But that seems hacky and can cause collateral damage if the implicit receivers happen to implement
         //  some common interface. So that logic is left out here for now. We can add it back in future if needed.
-        if (bestMatchWeightKind == null) return shortcutWeight
+        if (bestMatchWeightKind == null) {
+            return if (abortOnReceiverMismatch)
```
darthorimar, line 152: Should not we return null here in case of `abortOnReceiverMismatch == true`?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/weighers/CallableWeigher.kt
```diff
@@ -83,18 +90,46 @@ internal object CallableWeigher {
...
+        val symbol = when (explicitReceiver) {
+            is KtDotQualifiedExpression -> explicitReceiver.selectorExpression?.mainReference?.resolveToSymbol()
+            is KtNameReferenceExpression -> explicitReceiver.mainReference.resolveToSymbol()
+            else -> return null
+        }
+        return when (symbol) {
+            is KtTypeAliasSymbol -> symbol.expandedType
+            is KtClassOrObjectSymbol -> buildClassType(symbol) {
+                if (symbol is KtNamedClassOrObjectSymbol) {
+                    for (typeParameter in symbol.typeParameters) {
```
darthorimar, line 119: minor: `typeParameter` is unused here. Can be rewritten as:
```
repeat(symbol.typeParameters.size) {
	argument(KtStarProjectionTypeArgument(token))
}
```

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCompletionContributorBase.kt
```diff
@@ -110,7 +119,53 @@ internal abstract class FirCompletionContributorBase<C : FirRawPositionCompletio
...
+                override fun handleInsert(context: InsertionContext) {
+                    super.handleInsert(context)
+                    if (explicitReceiverRange == null || explicitReceiverText == null) return
+                    val castType = explicitReceiverTypeHint ?: kind.fullyQualifiedCastType
+                    val newReceiver = "(${explicitReceiverText} as $castType)"
+                    context.document.replaceString(explicitReceiverRange.startOffset, explicitReceiverRange.endOffset, newReceiver)
+                    context.commitDocument()
+                    shortenReferencesInRange(
+                        context.file as KtFile,
+                        TextRange(
```
darthorimar, line 160: minor: can be rewritten as `explicitReceiverRange.grow(newReceiver.length)`

darthorimar (changes requested):

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCompletionContributorBase.kt
darthorimar: > I do not like having this styling logic depending on CallableWeigher since it appears they should be totally decoupled from each other
they both may depend on some other part instead :) So, instead of using `CallableWeigher.CallableWeightKind` in styling, some new enum entry may be introduced. And used in weighter as in styling

>  But that means we need to duplicate the logic in CallableWeigher to check if receiver type matches
can we precomute it and put into some context?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
tgeng: This is doing filtering only on non-extension variable-like members. It would includes both local variables and proper members with current implicit receiver. So with or without receiver, there is always at most one symbol that resolution should go to. There won't be any need to pick a "better" one.

I agree that we should delegate this to the tower resolver. But from my observation the current implementation seems to just do the right thing. Consider
```
class Foo {
    val length = ""
    fun String.foo() {
        val length = true
        len<caret>
    }
}
```
Both the current logic and tower resolver picks candidate from inner most and fallback to the outer one if inner one do not exist. So maybe it's sufficient to keep this as it is?

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/weighers/CallableWeigher.kt
tgeng: No, the current implementation is the intended behavior. Let me rename this parameter to make this clear.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCompletionContributorBase.kt
tgeng: OK, moved this logic to a helper named `CallableMetadataProvider`.

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
```diff
@@ -185,7 +185,32 @@ internal open class FirCallableCompletionContributor(
...
+        if (smartCastInfo?.isStable == false) {
+            // Collect members available from unstable smartcast as well.
+            collectDotCompletionForCallableReceiver(
+                smartCastInfo.smartCastType,
+                visibilityChecker,
+                implicitScopes,
+                extensionChecker,
+                context,
+                // Only offer the hint if the type is denotable.
+                smartCastInfo.smartCastType.takeIf { it.approximateToSuperPublicDenotable() == null }
```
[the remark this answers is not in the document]
tgeng: OK, added that in https://kotlin.jetbrains.space/p/kotlin/reviews/279/timeline

plugins/kotlin/fir/src/org/jetbrains/kotlin/idea/completion/contributors/FirCallableCompletionContributor.kt
tgeng: Actually after thinking about it more, it seems the filtering done here should only be applied to unqualified local variables. Ideally completion should offer all other variable-like members, whether they are extension members or normal members. Then, upon user's selection, we should add label and explicit receivers to qualify the access in case it's ambiguous. For example
```
val Int.foo : ...
val Number.foo : ...
val String.foo : ...
fun String.test() {
  with(1) {
    val foo = ...
    fo<caret>
  }
}
```
Here, we should offer all the `foo` and completion should end up with
local `foo` -> `foo`
`Int.foo` -> `this.foo`
`Number.foo` -> `(this as Number).foo`
`String.foo` -> `this@test.foo`

Obviously, to do the above correctly, we will need to rely on tower resolver. One potential way to implement this that I can think of is to enumerate all available implicit receivers and resolve `foo` on them. Then for any members available we figure out the "minimal" label/typecast needed in order to make resolution go to that symbol.

But I am guessing this implementation would be quite slow. And the benefit may not worth the effort since user should really not use the same name for many things. Maybe we can add a TODO here?

darthorimar: Yeap, let's leave TODO for now. Looks like this require separate design

darthorimar (approved):

darthorimar:
Merged! Thank you :)

PR #1835, opened 2021-11-10
FIR LL: restore library module resolve test and add tests for enum entry

jsjeon:
1st commit restores `AbstractFirLibraryModuleDeclarationResolveTest` from https://github.com/JetBrains/kotlin/commit/44f74888e0d65246ed2839142ad972cbfee3e3d8#diff-3069eab6e799ccb36040fda7ed8dd24bc7c9950fd76435517d99504a11342f7a, which is somehow dropped during repo migration. Finding top-level property in library has been broken, and fix is in [KOTLIN-CR-258](https://kotlin.jetbrains.space/p/kotlin/reviews/258/).

Also, finding enum entry in library has been handled incorrectly. `KtEnumEntry`, a subtype of `KtClass`, was handled by `FirDeclarationForCompiledElementSearcher#findNonLocalClass`, which first attempted to load class id, whereas `KtEnumEntry` doesn't have class id. Instead, we need a separate logic for enum entry, where we can load the containing enum class's id first, and then retrieve matched enum entry in resolved enum class. See [KOTLIN-CR-258](https://kotlin.jetbrains.space/p/kotlin/reviews/258/) as well.


darthorimar:
Thank you for noticing and restoring the tests :)

darthorimar (approved):

darthorimar:
Merged, thank you again for restoring the tests :)

PR #1840, opened 2021-11-16
[decompiler] Fix the default keyword being added to private interface methods

coehlrich:
Private interface method's were added in java 9 and don't compile with the `default` keyword.


trespasserw:
Merged, thanks!

PR #1841, opened 2021-11-18
KTIJ-20154 Quick-fix for UNUSED_ANONYMOUS_PARAMETER to convert forEachIndexed to forEach

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20154
スクリーンショット 2021-11-18 10 29 29


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToNonIndexedFunctionCallFix.kt
```diff
@@ -0,0 +1,76 @@
...
+        val parameterList = functionLiteral.valueParameterList ?: return
+        val call = functionLiteral.getStrictParentOfType<KtCallExpression>() ?: return
+
+        labeledReturnExpressions.forEach { it.element?.setLabel(nonIndexedFunctionName) }
+        parameterList.removeParameter(0)
+        val psiFactory = KtPsiFactory(functionLiteral)
+        val callOrQualified = call.getQualifiedExpressionForSelector() ?: call
+        callOrQualified.replace(
+            psiFactory.buildExpression {
+                appendCallOrQualifiedExpression(call, nonIndexedFunctionName)
```
zarechenskiy, line 44: We also shouldn't forget about possible type arguments:
```
fun test(list: List<Int>) {
    list.forEachIndexed<Any> { index, value ->
        println(value)
    }
}
```

Currently, transforms to `list.forEach { value -> ...`, without type argument `<Any>`

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToNonIndexedFunctionCallFix.kt
```diff
@@ -0,0 +1,76 @@
...
+
+            val functionLiteral = parameterList.parent as? KtFunctionLiteral ?: return null
+            val call = functionLiteral.parentCallExpression() ?: return null
+            val functionName = call.calleeExpression?.text ?: return null
+            val (nonIndexedFunctionFqName, nonIndexedFunctionName) = nonIndexedFunctions[functionName] ?: return null
+            val context = functionLiteral.analyze(BodyResolveMode.PARTIAL)
+            if (call.getResolvedCall(context)?.isCalling(nonIndexedFunctionFqName) != true) return null
+
+            val labeledReturnExpressions =
+                functionLiteral.collectLabeledReturnExpressions(functionName, context).map { it.createSmartPointer() }
```
zarechenskiy, line 64: I propose to avoid passing data here and there. 
Instead, let's make one function that replaces labels inside a lambda without intermediate collections?

functionLiteral.replaceLambda(...) and call this function inside `invoke`

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToNonIndexedFunctionCallFix.kt
```diff
@@ -0,0 +1,76 @@
...
+            val parameterList = parameter.parent as? KtParameterList ?: return null
+            val parameters = parameterList.parameters
+            if (parameters.size < 2 || parameters.first() != parameter) return null
+
+            val functionLiteral = parameterList.parent as? KtFunctionLiteral ?: return null
+            val call = functionLiteral.parentCallExpression() ?: return null
+            val functionName = call.calleeExpression?.text ?: return null
+            val (nonIndexedFunctionFqName, nonIndexedFunctionName) = nonIndexedFunctions[functionName] ?: return null
+            val context = functionLiteral.analyze(BodyResolveMode.PARTIAL)
+            if (call.getResolvedCall(context)?.isCalling(nonIndexedFunctionFqName) != true) return null
```
zarechenskiy, line 61: I believe we even can avoid resolving anything and just check function calls by the simple name. The compiler already checked its FQName during the warning report

zarechenskiy (changes requested):
Thanks a lot! Quick-fix looks cool, the implementation requires some optimizations and simplifications, please check my comments

plugins/kotlin/idea/tests/testData/quickfix/convertToNonIndexedFunctionCall/extension.kt
```diff
@@ -0,0 +1,7 @@
+// "Convert to 'forEach'" "true"
+// WITH_STDLIB
+fun List<String>.test() {
+    <caret>forEachIndexed { <caret>index, s ->
```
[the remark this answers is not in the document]
t-kameyama: 607c654b7cb2

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToNonIndexedFunctionCallFix.kt
t-kameyama: 3a7fd755b9b8

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToNonIndexedFunctionCallFix.kt
t-kameyama: https://github.com/JetBrains/intellij-community/blob/3a7fd755b9b8af20fb500c53b62b6ca24587c539/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/Utils.kt#L416-L417

Don't we need to check the target function of label using BindingContext?
When resolving in the `invoke` function, the test gives the following error:
```
Resolve is not allowed under the write action for `org.jetbrains.kotlin.idea.quickfix.ConvertToNonIndexedFunctionCallFix`!
```

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/Utils.kt
```diff
@@ -418,3 +420,16 @@ fun BuilderByPattern<KtExpression>.appendCallOrQualifiedExpression(
 fun KtCallExpression.singleLambdaArgumentExpression(): KtLambdaExpression? {
     return lambdaArguments.singleOrNull()?.getArgumentExpression().safeAs() ?: getLastLambdaExpression()
 }
+
+fun KtFunctionLiteral.collectLabeledReturnExpressions(label: String, context: BindingContext): List<KtReturnExpression> =
+    collectDescendantsOfType { it.getLabelName() == label && it.getTargetFunction(context) == this }
```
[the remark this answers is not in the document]
t-kameyama: Same above.

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToNonIndexedFunctionCallFix.kt
t-kameyama: 36d4ff5e229a

PR #1842, opened 2021-11-19
KTIJ-10953 Removing receiver in "Receiver parameter is never used" fix might change program behavior

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-10953


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedReceiverParameterInspection.kt
```diff
@@ -113,6 +114,16 @@ class UnusedReceiverParameterInspection : AbstractKotlinInspection() {
                     RemoveReceiverFix(inSameClass)
                 )
             }
+
+            private fun KtCallableDeclaration.isCalledWithCallOrQualifiedReceiver(): Boolean {
+                return ReferencesSearch.search(this).any {
```
zarechenskiy, line 119: I'm a bit afraid that this approach can lead us to some performance penalties... Let me think a bit more about it

zarechenskiy (commented):
I really appreciate your work but I'm not sure that reference search wouldn't bite us due to performance reasons

t-kameyama:

PR #1843, opened 2021-11-19
FIR UAST: handle resolved target to decompiled declarations

jsjeon:
With [KOTLIN-CR-267](https://kotlin.jetbrains.space/p/kotlin/reviews/267/), if the resolution target element refers to a decompiled declaration (via its containing class or file), now we can retrieve corresponding light class/facade and look up matching member if needed.


darthorimar:
@jsjeon could you please explain why do you need special handling for cases of decompiled light classes and for classes from sources? Should not they behave the same way?

jsjeon:
> @jsjeon could you please explain why do you need special handling for cases of decompiled light classes and for classes from sources? Should not they behave the same way?

I guess you're asking whether this part:
```kt
when (ktModule) {
  is KtSourceModule -> {
    resolvedTargetElement.getMaybeLightElement(ktExpression)?.let { return it }
    resolveToPsiClass(ktExpression, resolvedTargetElement)?.let { return it }
  }
  is KtLibraryModule -> {
    resolveToPsiClass(ktExpression, resolvedTargetElement)?.let { return it }
    resolveDeserialized(ktExpression, resolvedTargetElement)?.let { return it }
  }
}
```
can be simply:
```kt
  resolvedTargetElement.getMaybeLightElement(ktExpression)?.let { return it }
  resolveToPsiClass(ktExpression, resolvedTargetElement)?.let { return it }
  resolveDeserialized(ktExpression, resolvedTargetElement)?.let { return it }
```
because the counterpart: https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin/src/org/jetbrains/uast/kotlin/internal/kotlinInternalUastUtils.kt#L293 is basically:
```kt
declarationDescriptor.toSource()?.getMaybeLightElement(sourcePsi)?.let { return it }
...
resolveToPsiClass(...)?.let { ... }
...
resolveDeserialized(...)?.let { return it }
```
?

This is quite subtle. For light classes from source, not passing thru `resolveDeserialized(...)` can be seen as a simple optimization, since it won't change anything further. Under the hood, `resolveDeserialized(...)` will trigger the same LC conversion utils, where LC creation/retrieval for decompiled declarations will be used _if_ the containing file is compiled. But, it's from source, not compiled, so nothing will be created/retrieved.

For light classes from decompiled declarations, though, passing thru `getMaybeLightElement` is quite different. Also note that, in the counterpart in FE1.0 UAST, `...toSource()?...` is a guard to avoid passing thru that util. The evil part of that util is: https://github.com/JetBrains/intellij-community/blob/kt-212-master/plugins/kotlin/uast/uast-kotlin-base/src/org/jetbrains/uast/kotlin/internal/baseKotlinInternalUastUtils.kt#L95 which tries UAST conversion anyway and use the PSI from it. Since `KotlinType`-to-`PsiType` conversion in FE 1.0 UAST still uses that util, the best option for us is to use that util only for source, as FE1.0 UAST does.

Maybe I can leave some comments to clarify.

plugins/kotlin/analysis/src/org/jetbrains/kotlin/idea/caches/resolve/IDEKotlinAsJavaSupport.kt
```diff
@@ -331,22 +332,30 @@ open class IDEKotlinAsJavaSupport(private val project: Project) : KotlinAsJavaSu
         val javaFileStub = ClsJavaStubByVirtualFileCache.getInstance(project).get(classFile) ?: return null
         javaFileStub.psiFactory = ClsWrapperStubPsiFactory.INSTANCE
         val manager = PsiManager.getInstance(mirrorFile.project)
-        val fakeFile = object : ClsFileImpl(ClassFileViewProvider(manager, classFile)) {
-            override fun getNavigationElement(): PsiElement {
-                if (correspondingClassOrObject != null) {
-                    return correspondingClassOrObject.navigationElement.containingFile
-                }
-                return super.getNavigationElement()
+        val fakeFile = FakeClsFileImpl(manager, classFile, mirrorFile, correspondingClassOrObject, javaFileStub)
```
darthorimar, line 335: Could you please explain why this change in FE1.0 Plugin is needed?

jsjeon: Ugh... not quite necessary. Just doing the same refactoring while doing so in FIR LC.

jsjeon:
This, along with additional changes for migrating LC abstractions for decompiled declarations, is uploaded to [KOTLIN-CR-267](https://kotlin.jetbrains.space/p/kotlin/reviews/267) so we can review/discuss things at once.

PR #1844, opened 2021-11-19
FIR UAST: test updates (inner classes in interfaces)

jsjeon:
Interfaces (and annotations as well, technically) can have inner classes, but FIR LC so far populated inner classes for classes only. [KOTLIN-CR-268](https://kotlin.jetbrains.space/p/kotlin/reviews/268/) is a quick fix for that, and these are simple test updates in FIR UAST.


darthorimar:
Merged, thank you!

PR #1845, opened 2021-11-20
Not quoting replacement by regexp if not necessary

Miha-x64:
Avoid `Matcher.quoteReplacement` when it is not necessary


BasLeijdekkers:
I added a test and tour change can now be found here: https://github.com/JetBrains/intellij-community/commit/30d775f307290887dfbd1bfbb7ec0cb2b2f5c39e

PR #1848, opened 2021-11-23
FIR UAST: test updates (no-arg constructors)

jsjeon:
No-arg constructors are added by `kotlinc` for some cases, e.g., if there is no explicit constructor at all, or all constructors' all parameters have default values. [KOTLIN-CR-270](https://kotlin.jetbrains.space/p/kotlin/reviews/270/) adds a counterpart to FIR LC, and these are simple test updates in FIR UAST.


darthorimar:
merged, thank you!

PR #1850, opened 2021-11-27
IDEA-283404: Support Guava's ImmutableSet in 'Overwritten Map, Set, or array element' inspection

turbanoff:


amaembo:
This feature is already implemented.

PR #1852, opened 2021-11-28
KTIJ-20013 "Implement members" puts function in the wrong place with a generic base type

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20013


dimonchik0036 (approved):
Thanks! Merged manually

PR #1839, opened 2021-11-15
[EE-IR] Support mutations by evaluator fragments

kandersen:
~~!!! Dependent on #1802 and JetBrains/kotlin#4647~~ !!!

- Add the new lowering to support this to the compiler invocation.

PR #1853, opened 2021-11-29
[EE-IR] Inline Function Support

kandersen:
~~Dependent on ~~#1802~~, #1839 and JetBrains/kotlin#4657~~

~~Two~~One item~~s~~ for which I would like review comments (marked in the code, as well)

~~1) There is a "inline call depth limit" of 10. I have improved the error reporting to the user, but I am curious to the origin of this limit.~~

2) PSI2IR needs resolution for all supplied `KtFile`s, so I have the fragment frontend additionally analyze all "transitive dependencies" from inline functions. I believe there is a way to reuse the editors running analyses of these files, and would like help to ensure I have hit that.

---

- Pass all transitively dependended-on files through the frontend to facilitate
  making IR out of all it. Needed when generating the modules containing inline
  functions (and their transitive dependencies) through the IR fragment
  compiler.

- Handle inline class mangling of `GENERATED_FOR_DEBUGGER_FUN` function name.

- Improve error messages around inline functions analyzed by the IDE.

### DIFF
--- a/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/util/inlineAnalysisUtils.kt
+++ b/plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/util/inlineAnalysisUtils.kt
@@ -26,7 +26,6 @@ fun analyzeInlinedFunctions(
     val context = analyzeElementWithInline(
         resolutionFacadeForFile,
         file,
-        1,
         analyzedElements,
         !analyzeOnlyReifiedInlineFunctions, bindingContext
     )
@@ -47,7 +46,6 @@ fun analyzeInlinedFunctions(
 private fun analyzeElementWithInline(
     resolutionFacade: ResolutionFacade,
     element: KtElement,
-    deep: Int,
     analyzedElements: MutableSet<KtElement>,
     analyzeInlineFunctions: Boolean,
     fullResolveContext: BindingContext? = null
@@ -124,7 +122,7 @@ private fun analyzeElementWithInline(
 
     analyzedElements.add(element)
 
-    if (declarationsWithBody.isNotEmpty() && deep < 10) {
+    if (declarationsWithBody.isNotEmpty()) {
         for (inlineFunction in declarationsWithBody) {
             val body = inlineFunction.bodyExpression
             if (body != null) {
@@ -132,7 +130,6 @@ private fun analyzeElementWithInline(
                     analyzeElementWithInline(
                         resolutionFacade,
                         inlineFunction,
-                        deep + 1,
                         analyzedElements,
                         analyzeInlineFunctions
                     )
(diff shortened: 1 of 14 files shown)


plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/util/inlineAnalysisUtils.kt
```diff
@@ -41,7 +43,10 @@ fun analyzeInlinedFunctions(
         toProcess.add(containingFile)
     }
 
-    return Pair<BindingContext, List<KtFile>>(context, ArrayList(toProcess))
+    // TODO: Can we ensure the analysis results from the underlying project is
+    //       reused/cached to avoid duplicate work?
+    val finalContext = resolutionFacadeForFile.analyzeWithAllCompilerChecks(toProcess)
```
kandersen, line 48: Does this correctly reuse and minimize duplicate analysis work?

plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/util/inlineAnalysisUtils.kt
```diff
@@ -124,10 +129,13 @@ private fun analyzeElementWithInline(
 
     analyzedElements.add(element)
 
-    if (declarationsWithBody.isNotEmpty() && deep < 10) {
+    if (declarationsWithBody.isNotEmpty()) {
         for (inlineFunction in declarationsWithBody) {
             val body = inlineFunction.bodyExpression
             if (body != null) {
+                require (deep < TRANSITIVE_INLINE_ANALYSIS_DEPTH_LIMIT) {
```
kandersen, line 136: @yanex what is the motivation for this limit?

yanex: I don't know, this code is quite old and originally came from the bytecode tool window: https://github.com/JetBrains/kotlin/commit/715b9ad859fb70e1c18f9d95a2484b00f6a22d00#diff-2a5f5f2b5b87193d3b3c581964190c1bd577b27ef8beefa85ae1a00e53d5d4c5R315.
I'm totally ok in removing this limitation.

plugins/kotlin/core/src/org/jetbrains/kotlin/idea/core/util/inlineAnalysisUtils.kt
nikita-nazarov: `ResolutionFacade` indeed caches the analysis. So it will not be repeated twice for the same file. However I think it is better to move `resolutionFacadeForFile.analyzeWithAllCompilerChecks(toProcess)` to `org.jetbrains.kotlin.idea.debugger.evaluate.KotlinEvaluator#compileCodeFragment`. The reason is that `org.jetbrains.kotlin.idea.core.util.InlineAnalysisUtilsKt#analyzeInlinedFunctions` is an utility method, it is also used by `org.jetbrains.kotlin.idea.core.KotlinCompilerIde`, and I am a bit worried that changing a binding context and doing some extra analysis may break or slow down something in the future.

kandersen:
Resolved all comments and rebased on kt-212-master!

nikita-nazarov:
Thanks! Merged ee8bcdff7a073fe6e7cf9d58be430572568c5976

kandersen:
Thank you!

PR #1855, opened 2021-11-30
IDEA-266513: Support records for Lombok @Builder and @With

florensie:
Fixes [IDEA-266513](https://youtrack.jetbrains.com/issue/IDEA-266513)

### DIFF
--- a/plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/util/PsiClassUtil.java
+++ b/plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/util/PsiClassUtil.java
@@ -1,6 +1,7 @@
 package de.plushnikov.intellij.plugin.util;
 
 import com.intellij.psi.*;
+import com.intellij.psi.impl.light.LightRecordField;
 import com.intellij.psi.impl.source.PsiExtensibleClass;
 import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
@@ -40,7 +41,9 @@ public final class PsiClassUtil {
    */
   @NotNull
   public static Collection<PsiField> collectClassFieldsIntern(@NotNull PsiClass psiClass) {
-    if (psiClass instanceof PsiExtensibleClass) {
+    if (psiClass.isRecord()) {
+      return createRecordFields(psiClass);
+    } else if (psiClass instanceof PsiExtensibleClass) {
       return ((PsiExtensibleClass) psiClass).getOwnFields();
     } else {
       return filterPsiElements(psiClass, PsiField.class);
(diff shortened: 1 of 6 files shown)


plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/util/PsiClassUtil.java
```diff
@@ -40,7 +41,9 @@
    */
   @NotNull
   public static Collection<PsiField> collectClassFieldsIntern(@NotNull PsiClass psiClass) {
-    if (psiClass instanceof PsiExtensibleClass) {
+    if (psiClass.isRecord()) {
+      return createRecordFields(psiClass);
```
akozlova, line 45: static fields of a record class are ignored. It doesn't matter for builder but influence e.g., the name conflicts here `de.plushnikov.intellij.plugin.processor.handler.FieldNameConstantsHandler#createFields`. Moreover, calling this method I would never assume that static fields are already filtered.
I also hesitate if it's ok to created additional fields which would be different to ones create by `com.intellij.psi.impl.RecordAugmentProvider` and which are already cached in different places but looks like that in lombok doesn't compare, so this one is probably worth mentioning in javadoc

florensie: Lombok completely circumvents the augments here to prevent a recursion from happening with it's own augment.

akozlova: sorry, I don't understand your comment. I want to say that you need to include static fields ("ownFields") of a record class as well as generated record components.

florensie: My comment was in reply to this:

> I also hesitate if it's ok to created additional fields which would be different to ones create by com.intellij.psi.impl.RecordAugmentProvider and which are already cached in different places but looks like that in lombok doesn't compare, so this one is probably worth mentioning in javadoc

This doesn't actually *create* any fields, merely collects them for the plugin to create it's own augments with.

akozlova: this need to be proved: if plugin propagate those fields anywhere or try to compare them with the `aClass.getFields()`  - IDE would be in trouble. Thus @mplushnikov suggests to patch just `BuilderHandler.java` as it's clear there that nothing bad can happen. Could you please do that or should we proceed ourselves? Thanks

florensie: I've moved the logic for builder support to BuilderHandler and BuilderInfo. I've also moved the record field collection logic to AbstractFieldProcessor.

akozlova:
Merged manually. Should appear on GitHub later today. Thank you for your contribution!

PR #1859, opened 2021-12-03
bugfix StackOverflowError caused by regex

comonad:


comonad:
Hopefully this will fix some issues of plugins, i.e.
https://github.com/rikvdkleij/intellij-haskell/issues/654

I could track that bug to the regex in openapi by appending
 -XX:MaxJavaStackTraceDepth=0
in
 ~/.config/JetBrains/IdeaIC2021.2/idea64.vmoptions

alexander-lobas:
Thanks for you work, I'll just push it to master.

PR #1861, opened 2021-12-04
IDEA-284073 add support immutability of toList

pyltsin:
I tried to add new logic for "Stream.toList"

https://youtrack.jetbrains.com/issue/IDEA-284073

I hope I did it correctly,


java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/Mutability.java
```diff
@@ -64,6 +64,8 @@
   public static final @NotNull String UNMODIFIABLE_VIEW_ANNOTATION = UNMODIFIABLE_VIEW.myAnnotation;
   private static final @NotNull CallMatcher STREAM_COLLECT = CallMatcher.instanceCall(
     CommonClassNames.JAVA_UTIL_STREAM_STREAM, "collect").parameterTypes("java.util.stream.Collector");
+  private static final @NotNull CallMatcher STREAM_COLLECT_TO_LIST = CallMatcher.instanceCall(
+    CommonClassNames.JAVA_UTIL_STREAM_STREAM, "toList");
```
amaembo, line 68: I would add `.withLanguageLevelAtLeast(JDK_16)`. This matcher can match `toList()` method declared in inheritors, even if `Stream.toList()` is absent. As before JDK_16 there were no `toList()` in the Stream, inheritors were free to assign any other semantics to the result of `toList()`, including the modifiable list. This was true for older versions of `StreamEx` library and probably true for `jOOL` library (haven't checked). So if the project was not updated to Java 16 and uses one of these libraries, the libraries don't violate the contract there but we will have incorrect warnings.

java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/java/inliner/StreamChainInliner.java
```diff
@@ -69,6 +69,7 @@
     instanceCall(JAVA_UTIL_STREAM_STREAM, "reduce").parameterTypes("T", "java.util.function.BinaryOperator");
   private static final CallMatcher COLLECT_TERMINAL =
     instanceCall(JAVA_UTIL_STREAM_STREAM, "collect").parameterTypes("java.util.stream.Collector");
+  private static final CallMatcher COLLECT_TO_LIST_TERMINAL = instanceCall(JAVA_UTIL_STREAM_STREAM, "toList");
```
amaembo, line 72: Please rename it to `TO_LIST_TERMINAL`. Naming here assumes mentioning of called methods but you are not calling `collect()` here.

java/java-tests/testSrc/com/intellij/java/codeInspection/DataFlowInspection16WithMockTest.java
```diff
@@ -0,0 +1,24 @@
+// Copyright 2000-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package com.intellij.java.codeInspection;
+
+import com.intellij.JavaTestUtil;
+import com.intellij.testFramework.LightProjectDescriptor;
+import org.jetbrains.annotations.NotNull;
+
+public class DataFlowInspection16WithMockTest extends DataFlowInspectionTestCase {
```
amaembo, line 8: I think it should be ok to add a new test to the existing `DataFlowInspection16Test` or `DataFlowInspection17Test` class. It's not so important that the particular test has library class mocks, as this doesn't affect other tests in the same TestCase class and doesn't require any specific setup steps. Also, I think that nothing bad will happen if mockJDK already has the Stream class and you redefine it inside the test. The references will be naturally resolved to the redefined one, as it's in the same file. In this case, you probably don't need to mock BaseStream at all.

amaembo (changes requested):

java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/Mutability.java
pyltsin: Thank you for your explanation. Fixed it

java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/java/inliner/StreamChainInliner.java
pyltsin: Renamed

java/java-tests/testSrc/com/intellij/java/codeInspection/DataFlowInspection16WithMockTest.java
pyltsin: I moved test and created mock in this test-case.

Thanks again!

pyltsin:
I tried to fix. Thank you!

amaembo:
Pushed to internal repo, thanks.

PR #1864, opened 2021-12-05
[IDEA-37227] Exclude disordered active eager cases

Fabrice-TIERCELIN:
Hi @BasLeijdekkers,

In _Expression may be factorized_ inspection, there are still bad cases:

```Java
boolean duplicateExpression = true;
boolean newValue = false;
boolean elseExpression = false;
return (duplicateExpression | duplicateExpression = newValue) & (duplicateExpression | elseExpression);
```

After cleanup, it would return `true`.

Now, an expression with side effect may be only at the very start or at the very end, so that the order of change can't be altered. The redundant expression will be put at the start or the end to guaranty that the order of change is not altered.

So when the active expression is at the start, the duplicate expression will be put at the end. And when the active expression is at the end, the duplicate expression will be put at the start.

If the active expression is at the end, the operators have to be eager. You can also notice that we will never have several active expressions. The occurrence of an active expression determines the order of the operands at the end. If no active expressions are present, the duplicate expression will be in the same order of its order in the last operation, to improve the readability.


plugins/InspectionGadgets/testsrc/com/siyeh/ig/fixes/controlflow/ExpressionMayBeFactorizedFixTest.java
```diff
@@ -56,24 +56,11 @@ public void testFactorInMiddle() {
...
-    );
-  }
-
-  public void testFactorOnEdge() {
-    doMemberTest(InspectionGadgetsBundle.message("if.may.be.factorized.quickfix"),
-                 "boolean test(boolean duplicateExpression, boolean payload) {\n" +
-                 "    boolean reassigned = false;\n" +
-                 "    return (duplicateExpression &/**/ (reassigned = payload)) || (true & duplicateExpression);\n" +
-                 "}",
-                 "boolean test(boolean duplicateExpression, boolean payload) {\n" +
```
MrRaindrop, line 74: test

plugins/InspectionGadgets/testsrc/com/siyeh/ig/fixes/controlflow/ExpressionMayBeFactorizedFixTest.java
```diff
@@ -174,6 +171,32 @@ public void testDoNotCleanupWithXOrOperator() {
...
+  }
+
+  public void testDoNotCleanupWithNotFinalEagerActiveExpression() {
+    assertQuickfixNotAvailable(InspectionGadgetsBundle.message("if.may.be.factorized.quickfix"),
+                               "class X {\n" +
+                               "  boolean test() {\n" +
+                               "    boolean duplicateExpression = true;\n" +
+                               "    boolean newValue = false;\n" +
+                               "    boolean elseExpression = false;\n" +
+                               "    return (duplicateExpression | duplicateExpression = newValue) &/**/ (duplicateExpression | elseExpression);\n" +
```
BasLeijdekkers, line 195: This test case is red code (an expression instead of a variable on the left side in the assignment expression), is it meant to be?

plugins/InspectionGadgets/testsrc/com/siyeh/ig/fixes/controlflow/ExpressionMayBeFactorizedFixTest.java
```diff
@@ -174,6 +171,32 @@ public void testDoNotCleanupWithXOrOperator() {
...
 
+  public void testDoNotCleanupWithFinalLazyActiveExpression() {
+    assertQuickfixNotAvailable(InspectionGadgetsBundle.message("if.may.be.factorized.quickfix"),
+                               "class X {\n" +
+                               "  boolean test() {\n" +
+                               "    boolean duplicateExpression = true;\n" +
+                               "    boolean thenExpression = true;\n" +
+                               "    boolean updatedExpression = true;\n" +
+                               "    boolean newValue = false;\n" +
+                               "    boolean evaluation = (duplicateExpression || thenExpression) &&/**/ (duplicateExpression || updatedExpression = newValue);\n" +
```
BasLeijdekkers, line 182: Here also red code in the test data. Intentional?

plugins/InspectionGadgets/testsrc/com/siyeh/ig/fixes/controlflow/ExpressionMayBeFactorizedFixTest.java
Fabrice-TIERCELIN: It wasn't intentional. I fixed it.

plugins/InspectionGadgets/testsrc/com/siyeh/ig/fixes/controlflow/ExpressionMayBeFactorizedFixTest.java
Fabrice-TIERCELIN: It wasn't intentional. I fixed it.

PR #1865, opened 2021-12-06
Use AssertJ's assertions in InspectionProfileTest & SingleInspectionProfilePanelTest

bbogason:
Purpose: Use AssertJ assertions consistently in test classes that already are using AssertJ.

Rational: AssertJ's assertions are often more readable than JUnit's assertions, e.g. `assertThat(collection).isEmpty()` instead of `assertEquals(0, collection.size())`.

This was done automatically using an IntelliJ plugin: _Sensei by Secure Code Warrior®_  using the AssertJ cookbook, and then some manual tweaking done to make the tests even more readable.

I hope to make more PRs in this vein, focusing on other test classes that are currently mixing JUnit and AssertJ assertions.

Further reading:
- [Blog post: Make unit tests readable with Sensei and AssertJ](https://www.securecodewarrior.com/blog/make-unit-tests-readable-with-sensei-and-assertj)
- [AssertJ cookbook](https://sensei.securecodewarrior.com/cookbooks/scw:testing:assertj)

### DIFF
--- a/java/java-tests/testSrc/com/intellij/codeInspection/ex/InspectionProfileTest.java
+++ b/java/java-tests/testSrc/com/intellij/codeInspection/ex/InspectionProfileTest.java
@@ -36,7 +36,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.stream.Collectors;
 
 import static com.intellij.testFramework.assertions.Assertions.assertThat;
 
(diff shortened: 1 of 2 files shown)


akozlova:
I don't see AssertJ in InspectionProfileTest. So I'd accept changes in SingleInspectionProfilePanelTest and keep as is the big one.

bbogason:
> I don't see AssertJ in InspectionProfileTest. So I'd accept changes in SingleInspectionProfilePanelTest and keep as is the big one.

Hi @akozlova and thank you for taking a look!

InspectionProfileTest does actually use `assertThat` from `com.intellij.testFramework.assertions.Assertions` which extends `org.assertj.core.api.Assertions` so it does indeed come from AssertJ. And therefore InspectionProfileTest currently uses assertions both from AssertJ and from JUnit (from `junit.framework.TestCase`) and this PR is mainly about migrating to AssertJ so that all assertions come from the same source (AssertJ).

I could split this into two PRs, one per class, if that would be preferable?

bbogason:
Closing this as Won't do.

akozlova:
I am sorry, I just check it again and again and when the code becomes longer (and to my understanding not much more understandable), I would not do that myself.

PR #1866, opened 2021-12-09
Speed up DependencyGraphModelBuilder

oehme:
Use only Node#id as the criterion for equals/hashCode, greatly speeding up the
redundancy checks during serialization. The ID was not unique in the whole build
before. I fixed that by reusing the graph generator for the whole build and having
an ever-increasing ID counter. The end-result is still the same. ReferenceNodes are
only created when a dependency is found multiple times in the same Configuration.

I've also removed the usage of Gradle's internal dependency report classes, which
were an unnecessary indirection. Instead, the graph generator now uses Gradle's
ResolutionResult API directly. 

Fixes IDEA-283622

For the project mentioned in that issue, sync times are 2 times faster after this change.

### DIFF
--- a/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/tasks/DependenciesReport.groovy
+++ b/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/tasks/DependenciesReport.groovy
@@ -3,6 +3,8 @@ package org.jetbrains.plugins.gradle.tooling.tasks
 
 import com.google.gson.GsonBuilder
 import com.intellij.openapi.externalSystem.model.project.dependencies.*
+import gnu.trove.THashMap
+import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 import org.gradle.api.DefaultTask
 import org.gradle.api.Describable
@@ -11,7 +13,6 @@ import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.FileCollectionDependency
 import org.gradle.api.artifacts.component.ComponentIdentifier
-import org.gradle.api.artifacts.component.ComponentSelector
 import org.gradle.api.artifacts.component.ModuleComponentIdentifier
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.api.artifacts.result.*
@@ -19,12 +20,12 @@ import org.gradle.api.file.FileCollection
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
-import org.gradle.api.tasks.diagnostics.internal.graph.nodes.AbstractRenderableDependencyResult
-import org.gradle.api.tasks.diagnostics.internal.graph.nodes.AbstractRenderableModuleResult
-import org.gradle.api.tasks.diagnostics.internal.graph.nodes.RenderableDependency
-import org.gradle.api.tasks.diagnostics.internal.graph.nodes.RenderableModuleResult
-import org.gradle.api.tasks.diagnostics.internal.graph.nodes.RenderableUnresolvedDependencyResult
 import org.gradle.util.GradleVersion
+
+import java.util.concurrent.atomic.AtomicLong
+
+import static com.intellij.openapi.externalSystem.model.project.dependencies.ResolutionState.RESOLVED
+import static com.intellij.openapi.externalSystem.model.project.dependencies.ResolutionState.UNRESOLVED
 
 @CompileStatic
 class DependenciesReport extends DefaultTask {
@@ -36,6 +37,7 @@ class DependenciesReport extends DefaultTask {
 
   @TaskAction
   void generate() {
+    def generator = new ReportGenerator()
     Collection<Configuration> configurationList
     if (configurations.isEmpty()) {
       configurationList = project.configurations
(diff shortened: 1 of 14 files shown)


plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/tasks/DependenciesReport.groovy
```diff
@@ -54,191 +53,146 @@ class DependenciesReport extends DefaultTask {
...
-    private long value
-
-    private long getId(String prefix, String configurationName) {
-      def key = prefix + '_' + configurationName
-      def id = idMap.get(key)
-      if (id == null) {
-        idMap[key] = ++value
-        id = value
+      def iterator = root.dependencies.iterator()
+      while (iterator.hasNext()) {
```
oehme, line 117: FYI this weird iterator scheme is the only way to get fast iteration in Groovy. It would probably be best to rewrite this class in Java at some point.

plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/tasks/DependenciesReport.groovy
```diff
@@ -54,191 +53,148 @@ class DependenciesReport extends DefaultTask {
...
-        if (dependencyResult instanceof UnresolvedDependencyResult) {
-          out.add(new RenderableUnresolvedDependencyResult((UnresolvedDependencyResult)dependencyResult))
-        }
-        else {
-          out.add(new CustomRenderableDependencyResult((ResolvedDependencyResult)dependencyResult))
-        }
-      }
-      return out
+    String fun(ProjectComponentIdentifier identifier) {
+      return is45OrNewer ? identifier.projectName : identifier.projectPath
```
oehme, line 196: Why don't you use `project.displayName` which has always existed and returns the full path, even in composite builds?

plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/tasks/DependenciesReport.groovy
```diff
@@ -54,191 +54,146 @@ class DependenciesReport extends DefaultTask {
...
-    CustomRenderableDependencyResult(ResolvedDependencyResult dependency) {
-      this.dependency = dependency
+      return node
     }
 
-    ResolutionState getResolutionState() {
-      return this.dependency.isConstraint() ? ResolutionState.RESOLVED_CONSTRAINT : ResolutionState.RESOLVED
-    }
+    private static class IdGenerator {
+      private THashMap<Object, Long> idMap = new THashMap<>()
```
oehme, line 173: Using THashMap here since we still support Java 6, but I wanted an efficient putIfAbsent method.

nskvortsov:
Unfortunately, the suggested code is not compatible with Gradle 5.0 and 4.10.3 (runs fine for 5.6.2+). Our tests failed with

```
java.lang.IncompatibleClassChangeError: Class org.gradle.api.internal.artifacts.result.DefaultResolvedComponentResult does not implement the requested interface org.gradle.api.artifacts.result.ComponentResult
  at org.jetbrains.plugins.gradle.tooling.tasks.DependenciesReport$ReportGenerator.doBuildDependenciesGraph(DependenciesReport.groovy:80)
  at org.jetbrains.plugins.gradle.tooling.tasks.DependenciesReport$ReportGenerator.buildDependenciesGraph(DependenciesReport.groovy:67)
  at org.jetbrains.plugins.gradle.tooling.builder.DependencyGraphModelBuilderImpl.buildAll(DependencyGraphModelBuilderImpl.groovy:46)
  at org.jetbrains.plugins.gradle.tooling.internal.ExtraModelBuilder.buildAll(ExtraModelBuilder.java:113)
  at org.jetbrains.plugins.gradle.tooling.internal.ExtraModelBuilder.buildAll(ExtraModelBuilder.java:78)
  ```

**Comment by oehme** [comment]
Fixed, using dynamic access on older Gradle versions.

**Comment by vladsoroka** [comment]
Looks good to me! However `DependenciesReport` task can be run out of importing request. E.g. to show dependencies UML diagram. In that case we need to add `trove` lib separately.
You can add it as following:

```
Index: community/plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java
IDEA additional info:
<+>UTF-8
diff --git a/community/plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java b/community/plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java
--- a/community/plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java	(revision c3700c9d5ee3b2b21bb11b2beb1cf68dbf8911c7)
+++ b/community/plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java	(date 1639580971646)
@@ -27,6 +27,7 @@
 import com.intellij.util.ArrayUtil;
 import com.intellij.util.Consumer;
 import com.intellij.util.execution.ParametersListUtil;
+import gnu.trove.THash;
 import org.gradle.api.Task;
 import org.gradle.tooling.*;
 import org.gradle.tooling.model.build.BuildEnvironment;
@@ -362,7 +363,7 @@
                                    @Nullable TaskCallback callback) {
 
     String taskName = taskClass.getSimpleName();
-    String paths = GradleExecutionHelper.getToolingExtensionsJarPaths(set(taskClass, GsonBuilder.class, ExternalSystemException.class));
+    String paths = GradleExecutionHelper.getToolingExtensionsJarPaths(set(taskClass, GsonBuilder.class, THash.class, ExternalSystemException.class));
     String initScript = "initscript {\n" +
                         "  dependencies {\n" +
                         "    classpath files(" + paths + ")\n" +

```

vladsoroka (approved):

oehme:
Thanks for the pointer, done :)

nskvortsov:
merged into master. Thank you!

PR #1868, opened 2021-12-10
Make runtime dependency filtering faster

oehme:
Since getDependencies returned an ArrayList, the filterRuntime method would use
the inefficient List.removeAll rather than the much faster Set#removeAll. This
is now fixed by using a LinkedHashSet, like in the other resolve methods.


oehme:
@vladsoroka @nskvortsov FYI

I wonder if we should standardize this class on `Set` instead of `Collection` to prevent other problems like this.

nskvortsov:
Thanks for another optimization. Merged manually.
By "standardize" do you mean "replace all usages"?

oehme:
Yeah, specifically I mean changing the return types from `Collection` to `Set` so that it's harder to make a mistake internally by using a List where it would be inefficient to do so.

PR #1869, opened 2021-12-11
language inject with mutation

phiSgr:
If a variable is annotated with `@Language`, language injections are then also applied to writes to the variable.

Can be useful in a mutation based DSL.

### DIFF
--- a/plugins/kotlin/injection/src/org/jetbrains/kotlin/idea/injection/KotlinLanguageInjectionContributor.kt
+++ b/plugins/kotlin/injection/src/org/jetbrains/kotlin/idea/injection/KotlinLanguageInjectionContributor.kt
@@ -171,15 +171,19 @@ class KotlinLanguageInjectionContributor : LanguageInjectionContributor {
             ?: injectWithMutation(place)
     }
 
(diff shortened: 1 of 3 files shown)


plugins/kotlin/injection/src/org/jetbrains/kotlin/idea/injection/KotlinLanguageInjectionContributor.kt
```diff
@@ -167,6 +168,31 @@ class KotlinLanguageInjectionContributor : LanguageInjectionContributor {
...
+        val property = when (val left = parent.left) {
+            is KtQualifiedExpression -> left.selectorExpression
+            else -> left
+        } ?: return null
+
+        for (reference in property.references) {
+            ProgressManager.checkCanceled()
+            val resolvedTo = reference.resolve()
+            if (resolvedTo is KtProperty) {
+                val annotation = resolvedTo.findAnnotation(FqName(AnnotationUtil.LANGUAGE)) ?: return null
```
phiSgr, line 190: I want to use `injectWithExplicitCodeInstruction` here, but it works on the strings themselves, not the definition site. Changing them seems too big a change.

xiexed:
Merged! Thank you for your contribution!

PR #1877, opened 2021-12-17
IDEA-285183 add check immutability Map.of for compute and merge

pyltsin:
https://youtrack.jetbrains.com/issue/IDEA-285183
I tried to do it.
Might be, I did it wrong.
Thanks for the great software!

### DIFF
--- a/java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/java/CFGBuilder.java
+++ b/java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/java/CFGBuilder.java
@@ -632,6 +632,21 @@ public class CFGBuilder {
   }
 
   /**
+   * Add only CallInstruction without any contract
+   * <p>
+   * Stack before: ... qualifier arg1 ... argN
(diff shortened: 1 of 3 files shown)


java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/java/CFGBuilder.java
```diff
@@ -631,6 +631,21 @@ public CFGBuilder call(PsiMethodCallExpression call) {
...
+   * Add only CallInstruction without any contract
+   * <p>
+   * Stack before: ... qualifier arg1 ... argN
+   * <p>
+   * Stack after: ... method result
+   *
+   * @param call call to add
+   * @return this builder
+   */
+  public CFGBuilder simpleCall(PsiMethodCallExpression call) {
```
amaembo, line 644: Why the existing `call` method doesn't satisfy your needs?

amaembo:
It's better to do this using the Ensure instruction. I will implement it.

PR #1880, opened 2021-12-20
Remove any backslash in nonNlsCommentPattern in order for the comment to be displayed properly

tsiarasl:
If the nonNlsCommentPattern contained any escaped special characters (which is needed for e.g the '$' character so that the matching is correct) then the comment would be displayed as is (with the backslash).

### DIFF
--- a/plugins/java-i18n/src/com/intellij/codeInspection/i18n/SuppressByCommentOutAction.java
+++ b/plugins/java-i18n/src/com/intellij/codeInspection/i18n/SuppressByCommentOutAction.java
@@ -57,7 +57,7 @@ class SuppressByCommentOutAction extends SuppressIntentionAction {
         prefixFound = true;
       }
     }
-    String commentText = "//" + prefix + nonNlsCommentPattern;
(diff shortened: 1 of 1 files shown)


amaembo:
Well, it covers a very specific scenario when you use only escaped characters, like `\$`. In general, generating a sample string from regex is a non-trivial task (I see projects like [Generex](https://github.com/mifmif/Generex/) that try to solve it). However, it could be accepted as an intermittent solution.

Could you please add a unit-test that shows the desired behavior? I wonder how it would behave if the input string contains `\\` (like `NON\NLS`)...

amaembo (changes requested):
Well, it covers a very specific scenario when you use only escaped characters, like `\$`. In general, generating a sample string from regex is a non-trivial task (I see projects like [Generex](https://github.com/mifmif/Generex/) that try to solve it). However, it could be accepted as an intermittent solution.

Could you please add a unit-test that shows the desired behavior? I wonder how it would behave if the input string contains `\\` (e.g., you want to support a comment like `NON\NLS`)...

PR #1884, opened 2021-12-23
KTIJ-20499 `Put parameters on one line` intention does not work in case line exceeds hard wrap value

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20499


comitative:

Your commit fixes the problem with long lines but it seems that it changes the quick fix behavior for shorter lines as well.

The current version of the quick fix not only puts all arguments on a single line, but also formats the argument list. The formatting is consistent between the preview window and the resulting code in the editor.

The commit in this PR changes the formatting: the preview window show the formatted version, but the resulting code keeps the original formatting that may differ from the preview.

Let me illustrate my point with some screenshots.

This is a code sample whose formatting does not follow the configured style.
wrong-formatting

The existing version of the quick fix shows a preview and produces a result that are consistently formatted according to the configured code style.

Preview:
current-preview

Result:
current-result

With your proposed changes the preview is still formatted:
proposed-preview

At the same time, the resulting code keeps the original formatting that is different from the preview:
proposed-result

I suppose that the current behavior would be more convenient for users as it is consistent and automatically fixes formatting mistakes.

What do you think, is it possible to update your fix to match the current behavior (it would be the best), or at least to keep the consistency between the preview and the result?

There is another issue with the preview window for long argument strings: the preview shows only a part of the declaration, and it seems that the argument that is displayed is not on the same line as the previous one.

long-line-preview

Could you please check that as well?

Thank you again for your work, and happy new year!

t-kameyama:
> What do you think, is it possible to update your fix to match the current behavior (it would be the best), or at least to keep the consistency between the preview and the result?

Fixed in 8790147

t-kameyama:
> What do you think, is it possible to update your fix to match the current behavior (it would be the best), or at least to keep the consistency between the preview and the result?

ceb046f

t-kameyama:
> There is another issue with the preview window for long argument strings: the preview shows only a part of the declaration, and it seems that the argument that is displayed is not on the same line as the previous one.

It seems that the formating here is the cause.
https://github.com/JetBrains/intellij-community/blob/c74b06a07f96b309df74a8b1a65bc732a552f7ff/platform/lang-impl/src/com/intellij/codeInsight/intention/impl/preview/IntentionPreviewModel.kt#L42-L44

dimonchik0036 (approved):

PR #1885, opened 2021-12-24
KTIJ-20526 Change Signature refactoring with rename to invalid character for string doesn't escape name

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20526


vladimirdolzhenko:
LGTM

PR #1890, opened 2021-12-30
[EE-IR] Use new stub generation constructor parameter

kandersen:


kandersen:
@nikita-nazarov

nikita-nazarov:
This line was added by 9fa48b4c050b23adaf9e1cf772932edd64c35cbd

PR #1892, opened 2021-12-31
Java UAST: internalize Java UAST elements

jsjeon:
Some parts of massive refactoring that altered internal structures of Kotlin UAST implementations were bundled to Android Lint, which was in turn rolled out to clients, such as Metalava and Google internal checkers. The good news is that there were no behavioral regressions. The bad (or maybe good) news is that there were a couple hiccups that required Android Lint, Lint checker writers, and UAST clients to remove uses of UAST implementation details before roll outs.

Over the past couple months,
* Such uses of UAST implementations have been rewritten to use `UElement`-level abstractions instead
* Lint detectors that detects the use of UAST implementations are developed and deployed, so new uses of UAST implementations can be detected during compilation

And now, this PR `internal`izes implementation details in Java UAST. The counter part for Kotlin UAST is done in a separate space [CR](https://kotlin.jetbrains.space/p/kotlin/reviews/292).

### DIFF
--- a/uast/uast-java/src/org/jetbrains/uast/java/declarations/JavaUFile.kt
+++ b/uast/uast-java/src/org/jetbrains/uast/java/declarations/JavaUFile.kt
@@ -4,10 +4,12 @@ package org.jetbrains.uast.java
 import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiJavaFile
 import com.intellij.psi.PsiRecursiveElementWalkingVisitor
+import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.uast.*
 import org.jetbrains.uast.java.internal.JavaUElementWithComments
 import java.util.*
 
+@ApiStatus.Internal
 class JavaUFile(
   override val sourcePsi: PsiJavaFile,
   override val languagePlugin: UastLanguagePlugin
@@ -16,7 +18,7 @@ class JavaUFile(
   override val packageName: String
     get() = sourcePsi.packageName
 
-  override val imports: List<JavaUImportStatement> by lz {
+  override val imports: List<UImportStatement> by lz {
     sourcePsi.importList?.allImportStatements?.map { JavaUImportStatement(it, this) } ?: listOf()
   }
 
--- a/uast/uast-java/src/org/jetbrains/uast/java/expressions/JavaUAnnotationCallExpression.kt
+++ b/uast/uast-java/src/org/jetbrains/uast/java/expressions/JavaUAnnotationCallExpression.kt
@@ -7,6 +7,7 @@ import com.intellij.psi.PsiAnnotation
 import com.intellij.psi.PsiMethod
 import com.intellij.psi.PsiType
 import com.intellij.psi.ResolveResult
+import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.uast.*
 import org.jetbrains.uast.java.JavaAbstractUExpression
 import org.jetbrains.uast.java.JavaConverter
@@ -14,11 +15,14 @@ import org.jetbrains.uast.java.JavaUAnnotation
 import org.jetbrains.uast.java.lz
 import org.jetbrains.uast.visitor.UastVisitor
 
+@ApiStatus.Internal
 class JavaUAnnotationCallExpression(
   override val sourcePsi: PsiAnnotation,
   givenParent: UElement?
-) : JavaAbstractUExpression(givenParent), UCallExpressionEx, UMultiResolvable {
+) : JavaAbstractUExpression(givenParent), UCallExpression, UMultiResolvable {
 
+  // TODO: once there is no more external usage, this property can be `private`.
+  @get:ApiStatus.Internal
   val uAnnotation: JavaUAnnotation by lz {
     JavaUAnnotation(sourcePsi, this)
   }
--- a/uast/uast-java/src/org/jetbrains/uast/java/expressions/JavaUClassLiteralExpression.kt
+++ b/uast/uast-java/src/org/jetbrains/uast/java/expressions/JavaUClassLiteralExpression.kt
@@ -17,9 +17,11 @@ package org.jetbrains.uast.java
 
 import com.intellij.psi.PsiClassObjectAccessExpression
 import com.intellij.psi.PsiType
+import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.uast.UClassLiteralExpression
 import org.jetbrains.uast.UElement
 
+@ApiStatus.Internal
 class JavaUClassLiteralExpression(
   override val sourcePsi: PsiClassObjectAccessExpression,
   givenParent: UElement?
@@ -27,5 +29,9 @@ class JavaUClassLiteralExpression(
   override val type: PsiType
     get() = sourcePsi.operand.type
 
-  override val expression: JavaUTypeReferenceExpression by lz { JavaUTypeReferenceExpression(sourcePsi.operand, this) }
+  // TODO: return type JavaUTypeReferenceExpression doesn't have anything special,
+  //  so UTypeReferenceExpression should be good enough (or checking if the underlying sourcePsi is from Java).
+  override val expression: JavaUTypeReferenceExpression by lz {
+    JavaUTypeReferenceExpression(sourcePsi.operand, this)
+  }
 }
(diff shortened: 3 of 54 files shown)


jsjeon:
cc @darthorimar @xiexed

uast/uast-java/src/org/jetbrains/uast/java/declarations/JavaUFile.kt
```diff
@@ -8,15 +8,15 @@ import org.jetbrains.uast.*
 import org.jetbrains.uast.java.internal.JavaUElementWithComments
 import java.util.*
 
-class JavaUFile(
+class JavaUFile internal constructor(
```
jsjeon, line 11: Note that this is the only one I can't make it `internal`, hence at least `internal constructor`. There are some tests that check cached evaluation results, and those somehow need copied version of this `JavaUFile` to not alter caches I guess.

xiexed:
It is a very good idea to internalize uast implementation-classes, and it is really cool that Android and Metalave have migrated from direct implementations. But the problem is that there are other users of UAST who can still rely on implementations. Although we discourage direct usage of language-specific uast classes we cant just make it private because it will break some plugins code. At least currently our checks showed that the [JMIX] (https://plugins.jetbrains.com/plugin/14340-jmix/) plugin will be broken by this PR.

I think it will be safer to use `@org.jetbrains.annotations.ApiStatus#Internal` annotation instead of Kotlin `internal` keyword - because it will not break. What do you think?

On other hand if I understand correctly for the Kotlin case we do move classes packages and then there is no chance to keep compatibility?

xiexed (changes requested):
It is a very good idea to internalize uast implementation-classes, and it is really cool that Android and Metalave have migrated from direct implementations. But the problem is that there are other users of UAST who can still rely on implementations. Although we discourage direct usage of language-specific uast classes we cant just make it private because it will break some plugins code. At least currently our checks showed that the [JMIX](https://plugins.jetbrains.com/plugin/14340-jmix/) plugin will be broken by this PR.

I think it will be safer to use @org.jetbrains.annotations.ApiStatus#Internal annotation instead of Kotlin internal keyword - because it will not break. What do you think?

On other hand if I understand correctly for the Kotlin case we do move classes packages and then there is no chance to keep compatibility?

jsjeon:
> It is a very good idea to internalize uast implementation-classes, and it is really cool that Android and Metalave have migrated from direct implementations. But the problem is that there are other users of UAST who can still rely on implementations. Although we discourage direct usage of language-specific uast classes we cant just make it private because it will break some plugins code. At least currently our checks showed that the [JMIX](https://plugins.jetbrains.com/plugin/14340-jmix/) plugin will be broken by this PR.
> 
> I think it will be safer to use @org.jetbrains.annotations.ApiStatus#Internal annotation instead of Kotlin internal keyword - because it will not break. What do you think?

Oh well, if there are more users/clients outside than IDE features in IJ and Android Lint, I agree that `internal` would be a breaking change. Before diving into the massive change one more time, I'd like to check the following:

* How about `abstract` classes, such as `AbstractJava*`? Is it okay to keep them `internal`?

* How about `internal constructor`? Does it seem not safe either? At least users can (or should?) use `toUElement` and/or `UastLanugagePlugin` instead to create instances of `UElement`?

> On other hand if I understand correctly for the Kotlin case we do move classes packages and then there is no chance to keep compatibility?

Yeah, basic Kotlin UAST elements are in `uast-kotlin-base` module and shared across different plugins, so I thought `internal` cannot be feasible for Kotlin case when investing this issue. It turns out that everything is "fully" commonized, including a converter where Kolint `UElement` is created, so it became feasible to make those `internal` (except for those used in codegen plugin). So, in either way, we can make Java UAST and Kotlin UAST consistent.

jsjeon:
> * How about abstract classes, such as AbstractJava*? Is it okay to keep them internal?

Ah, never mind this one: if sub classes can't be `internal`, that `abstract` one can't be `internal` either. :\

xiexed:
> How about abstract classes, such as AbstractJava*? Is it okay to keep them internal?

Afaik it is not possible to extend the internal class with a "public" class, so it seems that it will not work. Ah, ok, you've already written that :)

> Yeah, basic Kotlin UAST elements are in uast-kotlin-base module

It is ok, but the question is aren't the package names changed? In other things it seems that compatibility will not be broken

> How about internal constructor?

I can't find sane reason why someone will want to create a UElement manually, but IDEA compatibility politics is that if we are not sure that no one uses something then we should go through deprecation cycle. So in general I think if we will mark class as `@ApiStatus.Internal` then there is no much sence to make constuctors `internal` additionally, at least for current iteration

jsjeon:
> > Yeah, basic Kotlin UAST elements are in uast-kotlin-base module
> 
> It is ok, but the question is aren't the package names changed? In other things it seems that compatibility will not be broken

Package names for some have been changed about 4~6 months ago in Kotlin plugin branch. And, we hit that issue in Android Lint and Google internally. All are resolved accordingly once we noticed the package relocation. Btw, that's Kotlin UAST issue, not Java UAST here.

> > How about internal constructor?
> 
> I can't find sane reason why someone will want to create a `UElement` manually, but IDEA compatibility policy is that if we are not sure that no one uses something then we should go through deprecation cycle. So in general I think if we will mark class as `@ApiStatus.Internal` then there is no much sense to make constructors `internal` additionally, at least for current iteration

I see. Changed all elements I put `internal` back to `public`, and annotated them with `@ApiStatus.Internal`. Made some return type changes for future `internal`-ization: if clients should not depend on internal implementations, specialized return types are not needed either.

uast/uast-java/src/org/jetbrains/uast/java/expressions/JavaUClassLiteralExpression.kt
```diff
@@ -17,15 +17,18 @@ package org.jetbrains.uast.java
...
 
+@ApiStatus.Internal
 class JavaUClassLiteralExpression(
   override val sourcePsi: PsiClassObjectAccessExpression,
   givenParent: UElement?
 ) : JavaAbstractUExpression(givenParent), UClassLiteralExpression {
   override val type: PsiType
     get() = sourcePsi.operand.type
 
-  override val expression: JavaUTypeReferenceExpression by lz { JavaUTypeReferenceExpression(sourcePsi.operand, this) }
```
xiexed, line 30: also there is a known external usage ((

xiexed (changes requested):

uast/uast-java/src/org/jetbrains/uast/java/declarations/JavaUFile.kt
```diff
@@ -40,4 +42,6 @@ class JavaUFile(
   @Suppress("OverridingDeprecatedMember")
   override val psi
     get() = sourcePsi
-}
\ No newline at end of file
+
+  fun copy() : JavaUFile = JavaUFile(sourcePsi, languagePlugin)
```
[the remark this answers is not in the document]
jsjeon: Ah, I thought, eventually these will become `internal`, and for that specific test, we still need a copy of `JavaUFile` to not corrupt caches. We can put this when it's the time; moved this to that test directly, along with TODO note.

uast/uast-java/src/org/jetbrains/uast/java/expressions/JavaUAnnotationCallExpression.kt
```diff
@@ -7,19 +7,21 @@ import com.intellij.psi.PsiAnnotation
...
 
+@ApiStatus.Internal
 class JavaUAnnotationCallExpression(
   override val sourcePsi: PsiAnnotation,
   givenParent: UElement?
-) : JavaAbstractUExpression(givenParent), UCallExpressionEx, UMultiResolvable {
+) : JavaAbstractUExpression(givenParent), UCallExpression, UMultiResolvable {
 
-  val uAnnotation: JavaUAnnotation by lz {
+  private val uAnnotation: JavaUAnnotation by lz {
```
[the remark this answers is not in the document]
jsjeon: Thanks, reverted. But, just in case, left TODO and getter annotation (which might be redundant, but to emphasize...?)

uast/uast-java/src/org/jetbrains/uast/java/expressions/JavaUClassLiteralExpression.kt
jsjeon: That's bad... :( there is nothing more a user can get learn from the return type here. Left TODO as well.

xiexed:
Merged! Thanks!

PR #1894, opened 2022-01-04
EqualsWhichDoesntCheckParameterClass should consider org.apache.commons.lang3.builder.EqualsBuilder

stevejagodzinski:
Fixes https://youtrack.jetbrains.com/issue/IDEA-286099


BasLeijdekkers:
Thanks for your submission, but I fixed this another way.

PR #1895, opened 2022-01-05
[gradle]2021.x cannot import

ykfs:
when update to 2021.x, gradle project can not import the jar which is install to mavenlocal:
a project:
`apply plugin: 'maven'`
then click install to mavenlocal
b project import a, but it fail, it still want find the wrong snapshot
```
repositories {
    mavenLocal()
   maven {
      credentials {
                  username xxx
                  password xxx
              }
              url xxx
      }
}
```
but when use 2020.x ,it success


ybedrov:
Hello! Please report issue on YouTrack:  https://youtrack.jetbrains.com/issues/IDEA

PR #1896, opened 2022-01-05
Protect shared sourceSet cache from concurrent access

oehme:
The sourceSet cache initializer can be called from several threads, but Gradle's
project model is not thread safe, so we need to make sure only one thread at a time
can initialize this.

This makes parallel model fetch work in the project I'm currently working on.

FYI @vladsoroka @nskvortsov

### DIFF
--- a/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/util/SourceSetCachedFinder.groovy
+++ b/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/util/SourceSetCachedFinder.groovy
@@ -15,6 +15,7 @@
  */
 package org.jetbrains.plugins.gradle.tooling.util
 
+import java.util.concurrent.ConcurrentHashMap
 import groovy.transform.CompileStatic
 import org.gradle.api.Project
 import org.gradle.api.initialization.IncludedBuild
@@ -41,7 +42,7 @@ class SourceSetCachedFinder {
   private static final DataProvider<ArtifactsMap> ARTIFACTS_PROVIDER = new DataProvider<ArtifactsMap>() {
     @NotNull
     @Override
-    ArtifactsMap create(@NotNull Gradle gradle, @NotNull MessageReporter messageReporter) {
+    synchronized ArtifactsMap create(@NotNull Gradle gradle, @NotNull MessageReporter messageReporter) {
       return createArtifactsMap(gradle)
     }
   }
@@ -49,7 +50,7 @@ class SourceSetCachedFinder {
     @NotNull
     @Override
     Map<String, Set<File>> create(@NotNull Gradle gradle, @NotNull MessageReporter messageReporter) {
-      return new HashMap<String, Set<File>>()
+      return new ConcurrentHashMap<String, Set<File>>()
     }
   }
 


nskvortsov:
I have doubts about merging this request, as it is only a very local fix.
What threads are involved in parallel model fetch ? how will it affect solutions like that `ThreadLocal` for contexts?
https://github.com/JetBrains/intellij-community/blob/fac93f2bcf61fc41717b2133e25af76e95cfe528/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/ExtraModelBuilder.java#L54

oehme:
> What threads are involved in parallel model fetch 

Gradle will run the supplied model builders against each subproject in parallel on its internal thread pool (up to --max-workers).

The ThreadLocal approach you linked works fine for that. This shared SourceSet cache is the only place I found where you reached across project boundaries, which is not safe, so I synchronized it. After that map is initialized, its safe to query. The other thing is the `SOURCES_DATA_KEY`, which can be updated from several threads and thus should be a concurrent map.

I haven't looked at all model builders, e.g. I didn't check any of the Android-related code, since the project I'm working on is a regular Java project. For that project this small fix is enough to make parallel fetch work properly.

nskvortsov:
I need to further investigate this, as we need to make sure the model builders will work in multi-threaded case.

oehme:
> I need to further investigate this, as we need to make sure the model builders will work in multi-threaded case.

Just to clarify: This PR does not change anything about activating parallel model fetch. It's still a hidden, optional feature in the registry. It just makes it work for a bunch more projects (including ours).

nskvortsov:
It turned out, the ThreadLocal was an old, deprecated workaround. It is removed from master branch now.

oehme:
I see, in that case I'll have to bring back the concurrent map :)

oehme:
I've adjusted the PR back to its original form.

LifeIsStrange:
@oehme I'm curious, how can I enable parallel model fetch?
Is this the same as setting org.gradle.parallel=true in Gradle properties?

nskvortsov:
@LifeIsStrange this can be enabled via IntelliJ IDEA Registry key `gradle.tooling.models.parallel.fetch`
Please note, this is an experimental switch and is not guaranteed to work at all

LifeIsStrange:
@nskvortsov Thanks for the info, does it make sense combining this setting with gradle.parallel=true?

oehme:
@nskvortsov I think making parallel fetch a Gradle system property would be better than the global switch in the IDE. This way individual projects could opt in and it would be easier to share that configuration along with the build.

nskvortsov:
I have reworked the shared data structures across data models. Commits will be available soon, when our cherry-pick robot wakes up )

nskvortsov:
@oehme please try this patch https://github.com/JetBrains/intellij-community/commit/2421b42394d7e9a65dcdd4c14c96504df44134e8 and see if it helps you.

oehme:
Thank you @nskvortsov. I've just verified this with the project I'm working on and it fixes the problem. Will this be in the next EAP as well? And should I send a PR for the system property idea?

LifeIsStrange:
Any benchmarck results? :0

oehme:
@LifeIsStrange it reduces the time spent resolving dependencies from 22s to 8s in my case.

nskvortsov:
@oehme yes, this should land in next EAP.  A PR for the system property would be very welcome!

PR #1897, opened 2022-01-05
Remove container callback usages from model builders

oehme:
Container callbacks like .all{} and .allprojects{} are meant to be used
during configuration by plugins and build scripts, so they can react to
newly added objects. These methods are instrumented with build operations
for detailed analytics in build scans.

When a model builder runs, all configuration has already taken place and
the simpler .each{} method can be used. This avoids creating extra "noise"
that build scans would capture. It doesn't affect performance much, but it
makes profiling IDEA sync slightly more pleasant.

FYI @vladsoroka and @nskvortsov


nskvortsov:
LGTM, running automated checks

nskvortsov:
merged manually

PR #1898, opened 2022-01-05
Remove legacy LenientConfiguration API usage

oehme:
This API is outdated and inefficient, adding overhead to the sync process.

The ArtifactView API deliberately hides concepts like classifiers, packaging and configurations,
as these are Maven/Ivy implementation details. Gradle internally uses the concept of variants
instead. It also exposes this variant information via the ResolvedVariant API that IntelliJ could use if
you wanted to provide more detail about which variant of a library was resolved.

This change is more of a basis for discussion than being intended to be merged. I think we
should finally let the old LenientConfiguration API go and the first step would be IntelliJ
no longer using it on modern Gradle versions.

@vladsoroka @nskvortsov Can you run the CI build on this so I get an idea how big the gap to
the current functionality is?

### DIFF
--- a/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/util/resolve/DependencyResolverImpl.java
+++ b/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/util/resolve/DependencyResolverImpl.java
@@ -2,25 +2,29 @@
 package org.jetbrains.plugins.gradle.tooling.util.resolve;
 
 import org.gradle.api.Action;
-import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
-import org.gradle.api.artifacts.*;
+import org.gradle.api.artifacts.ArtifactCollection;
+import org.gradle.api.artifacts.ArtifactView;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.ModuleVersionIdentifier;
 import org.gradle.api.artifacts.component.*;
 import org.gradle.api.artifacts.dsl.DependencyHandler;
-import org.gradle.api.artifacts.result.*;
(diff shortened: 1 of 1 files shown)


oehme:
@adammurdoch might be interested

nskvortsov:
Will run it and get back with the results.

nskvortsov:
So, we have quite a number of failures with this patch.
Among others, some `GradleDependenciesImportingTest` parts are affected:

* testDependencyWithDifferentClassifiers
* testNonDefaultProjectConfigurationDependencyWithMultipleArtifacts
* testSourceSetOutputDirsAsArtifactDependencies
* testTestModuleDependencyAsArtifactFromTestSourceSetOutput
* testTestModuleDependencyAsArtifactFromTestSourceSetOutput2
* testTestModuleDependencyAsArtifactFromTestSourceSetOutput3
* testTransformedProjectDependency

Lets start fixing those...

oehme:
How can I debug these? I need breakpoints in your model builder to find out what's wrong.

oehme:
I get a ton of failures, even when I'm building `master`, i.e. without any of my changes. Are there any extra steps necessary to get the tests passing?

nskvortsov:
Nothing more than described in the docs: https://github.com/JetBrains/intellij-community/blob/master/README.md#running-intellij-idea (see section "To run tests on the build,")

adammurdoch:
This would be a good change to make, as we want to deprecate the old `LenientConfiguration` API eventually and instead focus on making the `ArtifactView` and the associated result APIs work well.

We're also currently improving the `ArtifactView` API so that it can handle source and Javadoc artifacts, which would mean these would be downloaded in parallel, which the artifact query API does not do. Artifact resolution would also become variant aware (eg, could handle things like platform specific Kotlin source, for example) and there would be just a single API to use to query artifacts, instead of several.

oehme:
Closing since I didn't have time to look into this more

PR #1899, opened 2022-01-05
Convert TAPI models on a background thread

oehme:
This way the model builders can make progress while the result of the
previous one is being converted. Having a single thread for conversion
makes it safe to use the non-thread-safe converters and data structures.

FYI @vladsoroka @nskvortsov

### DIFF
--- a/plugins/gradle/tooling-extension-api/src/org/jetbrains/plugins/gradle/model/ProjectImportAction.java
+++ b/plugins/gradle/tooling-extension-api/src/org/jetbrains/plugins/gradle/model/ProjectImportAction.java
@@ -105,12 +105,14 @@ public class ProjectImportAction implements BuildAction<ProjectImportAction.AllM
   @Nullable
   @Override
   public AllModels execute(final BuildController controller) {
-    myConverterExecutor =  Executors.newSingleThreadExecutor(new ThreadFactory() {
-      @Override
-      public Thread newThread(@NotNull Runnable runnable) {
-        return new Thread(runnable, "idea-tooling-model-converter");
-      }
-    });
+    if (Boolean.getBoolean("idea.background.convert")) {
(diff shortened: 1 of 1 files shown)


plugins/gradle/tooling-extension-api/src/org/jetbrains/plugins/gradle/model/ProjectImportAction.java
```diff
@@ -141,6 +148,13 @@ public void accept(@NotNull GradleBuild includedBuild) {
     if (isProjectsLoadedAction) {
       wrappedController.getModel(TurnOffDefaultTasks.class);
     }
+    myConverterExecutor.shutdown();
+    try {
+      myConverterExecutor.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS);
```
oehme, line 153: Waiting forever is the simplest strategy, but probably not the best :) What kind of handling would you prefer here?

nskvortsov:
can we make the switch to basckground thread optional, so we can test it during the EAPs?

oehme:
Yes, for sure. Where would I put that option? In the "Registry" or are you thinking of something else?

nskvortsov:
I think a system property will be enough (so it can be configured using gradle.properties file).

oehme:
Did you want it to be on or off by default?

nskvortsov:
I think we can enable it by default in master. We can turn it off in the release branch afterward

nskvortsov:
I have enabled this by default on our master branch. Commits are merged and are in progress on our internal CI

oehme:
Thank you!

oehme:
@nskvortsov the commit with the opt-in switch doesn't seem to be in 2022.1 EAP. I only see the commits on master, where the feature is enabled by default. Any chance to get the opt-in into the EAP so people can start trying it out?

nskvortsov:
It will be available with the next EAP build, stay tuned )

PR #1902, opened 2022-01-09
Set _JAVA_AWT_WM_NONREPARENTING=1 on wayland

kressi:


trespasserw:
The IDE works fine on Wayland without this variable. Is it needed for your window manager? Which one do you use?

kressi:
My window manager is Sway, it does not work without setting this variable myself. What window manager are you using? Is that variable already set, or does run fully without this variable?

trespasserw:
Mine is GNOME Shell, works fine without the variable.

I have looked at [known issues](https://youtrack.jetbrains.com/issues?q=tag:%20de-sway) with Sway - there aren't many of them and all are closed as 3rd-party issues with references to corresponding tickets in Sway. Please take a look - maybe your problem is already fixed? If it isn't or your problem is new, please consider filing a new issue.

Thank your for the PR, but the patch doesn't look good to me: it is rather system-specific and basically is a workaround for a problem that should be fixed in other ways. I don't see how the patch could be improved.

trespasserw:
Found [IDEA-153134](https://youtrack.jetbrains.com/issue/IDEA-153134) - is this the problem you're seeing?

kressi:
Looks similar, screen is not white though. Could be different, because screenshots are taken on AwesomeWM or for different default themes.

PR #1904, opened 2022-01-20
Kotlin: Do not hide duplicate variables in InlineStackFrameVariableHolder

sfs:
Fixes KTIJ-20828

---

This is already tested in `testCapturedValues2`, where I've fixed the test expectations. There seems to be a separate problem with `@InlineOnly` functions in `testInlineStackTrace` (the inline call stack shows the wrong "this" variables), but that doesn't appear to be a new issue.

### DIFF
--- a/plugins/kotlin/jvm-debugger/util/src/org/jetbrains/kotlin/idea/debugger/InlineUtils.kt
+++ b/plugins/kotlin/jvm-debugger/util/src/org/jetbrains/kotlin/idea/debugger/InlineUtils.kt
@@ -32,7 +32,7 @@ val INLINED_THIS_REGEX = run {
  * stack frames.
  */
 class InlineStackFrameVariableHolder private constructor(
-    // Visible variables sorted by start offset.
+    // Visible variables sorted by start offset. May contain duplicate variable names.
     private val sortedVariables: List<LocalVariableProxyImpl>,
     // Map from variable index to frame id. Every frame corresponds to a call to
     // a Kotlin (inline) function.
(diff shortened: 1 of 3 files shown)


plugins/kotlin/jvm-debugger/util/src/org/jetbrains/kotlin/idea/debugger/InlineUtils.kt
```diff
@@ -147,10 +153,11 @@ class InlineStackFrameVariableHolder private constructor(
...
                     // associated with a previous active frame.
                     //
-                    // If we do encounter pending variables, then they belong to an
-                    // inline call whose scope introduction variable was shadowed by
-                    // a later call to the same function and can be hidden.
+                    // If we do encounter pending variables, then this indicates that the
+                    // debug information is inconsistent and we mark all pending variables
+                    // as invalid.
                     name.startsWith(LOCAL_VARIABLE_NAME_PREFIX_INLINE_ARGUMENT) -> {
+                        require(pendingVariables.isEmpty())
```
nikita-nazarov, line 160: What is this line needed for? Because of it, the code below doesn't make sense. Previously there was an assert here, but we decided to invalidate pending variables instead of throwing an exception.

sfs: Good catch, I added some requires for testing that these workarounds are no longer needed and forgot to take this one out. Fixed!

PR #1905, opened 2022-01-21
Improve linkProjectLibrary performance

yuriy-solodkyy:
GradleProjectResolverUtil.linkProjectLibrary is invoked at least once
for each library dependency in each module by either Gradle or Android
plugin and it each time it searches for an existing library data
matching the one being added by name.

Assuming LIBRARY_DATA nodes are only added to the collection and the
collection is never cleared, it is safe to cache this mapping and avoid
expensive linear search on each invocation.

This change adds a cache as a user data attached to PROJECT_DATA data
node and may save up to a minute of sync time on large projects.

This can be further improved, if we assume that LIBRARY_DATA nodes
should never be added in any other way than via linkProjectLibrary. In
this case we can avoid linear searching even when adding the library for
the first time.


yuriy-solodkyy:
@vladsoroka @nskvortsov

nskvortsov:
Looks great to me, thank you! 
Merged manually

PR #1908, opened 2022-01-24
KTIJ-13949 Provide quick fix for [EQUALITY_NOT_APPLICABLE] compilation error on Long (Short etc) == Int literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-13949

```kotlin
fun test(b: Byte, i: Int): Boolean {
    return b == i
}
```
スクリーンショット 2022-01-24 14 25 52

```kotlin
fun test(c: Char): Boolean {
    return c == "a"
}
```
スクリーンショット 2022-01-24 14 26 51

### DIFF
--- a/plugins/kotlin/frontend-independent/resources-en/messages/KotlinBundle.properties
+++ b/plugins/kotlin/frontend-independent/resources-en/messages/KotlinBundle.properties
@@ -2538,3 +2538,7 @@ provide.return.value=Provide the return value
 
 
 wrap.expression.in.parentheses=Wrap expression in parentheses
+
+convert.left.hand.side.to.0=Convert left-hand side to ''{0}''
+convert.right.hand.side.to.0=Convert right-hand side to ''{0}''
+convert.string.to.character.literal=Convert string to character literal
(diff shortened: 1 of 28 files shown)


t-kameyama:
Added a commit for [KTIJ-22113](https://youtrack.jetbrains.com/issues/issue/KTIJ-22113/Quick-fix-to-convert-string-literal-to-char-when-there-is-a-type-mismatch).

```kotlin
fun f1(c: Char) {
    when (c) {
        "." -> {}
    }
}

fun f2(b: Byte, i: Int) {
    when (b) {
        i -> {}
    }
}

fun f3(b: Byte, i: Int) {
    when (i) {
        b -> {}
    }
}
```

https://user-images.githubusercontent.com/1121855/176572062-b7c60901-ac0e-40b4-839b-2e3bad4d5832.mov

abelkov:
Thank you, I merged the PR.
I improved it a little:

* Added support for other diagnostics to convert string to char, see QuickFixFactoryForTypeMismatchError. And migrated your INCOMPATIBLE_TYPES factory to this existing QuickFixFactoryForTypeMismatchError factory. Unfortunately, I don't know how to easily support it also for EQUALITY_NOT_APPLICABLE diagnostic, because there we have two equivalent expressions - left and right - instead of just expected and actual types.
* Supported strings like "\"" and "'".
* Added cheaper PSI-based checks.

PR #1911, opened 2022-01-28
KTIJ-20639 "Remove explicit type" breaks code for Long variables

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20639


zarechenskiy:
I believe this issue is also true for other numeric types. For example:
```
val long: Long = 0 // 0 -> 0L
val uLong: ULong = 0u // 0u -> 0uL; 0U -> 0UL
val byte: Byte = 0 // 0 -> 0.toByte()
val short: Short = 0 // 0 -> 0.toShort()
```

Could you please support these cases as well?

zarechenskiy:
Also, while we're on this issue, please consider case when we have `null` on RHS:
```
val long: Long? = null
```

We could insert cast in this case:
```
val long = null as Long?
```

t-kameyama:
Fixed in [3b13a6e](https://github.com/JetBrains/intellij-community/pull/1911/commits/3b13a6eac52a125810ef67ff0b5c1c168cc7f4e4)

t-kameyama:

PR #1913, opened 2022-01-31
KTIJ-20627 Quick fix adding TODO for "A 'return' expression required"

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20627


nikita-nazarov (commented):
The code is fine, but the fix should look like:
```
fun test(): String {
    TODO("Provide the return value")
}
```

Instead of:
```
fun test(): String {
    return TODO("Provide the return value")
}
```

Because it better corresponds with the Intellij IDEA policy of default overridings, and adding `return` leads to unreachable code warning.
So all `add return expression` occurrences in code should be renamed to `add todo item`

nikita-nazarov:
Ok then. @t-kameyama Could you please revert your last changes, so the quick fix will add `return TODO`?

t-kameyama:
> Ok then. @t-kameyama Could you please revert your last changes, so the quick fix will add `return TODO`?

done

ilya-g (approved):
LGTM, let's wait for Nikita's review

PR #1916, opened 2022-02-03
KTIJ-17636 ReplaceWith quickfix adds wrong `this` label inside lambda with receiver for member extension function call

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-17636


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInliner/CodeToInlineBuilder.kt
```diff
@@ -357,8 +358,10 @@ class CodeToInlineBuilder(
                     if (receiver is ImplicitReceiver) {
                         val resolutionScope = expression.getResolutionScope(bindingContext, resolutionFacade)
                         val receiverExpressionToInline = receiver.asExpression(resolutionScope, psiFactory)
-                        if (receiverExpressionToInline != null) {
-                            val receiverType = receiver.type.unCapture()
+                        val receiverType = receiver.type.unCapture()
+                        val receiverIsUnnecessary = (receiverExpressionToInline as? KtThisExpression)?.labelQualifier != null &&
+                                receiverType == targetCallable.receiverType()
```
dimonchik0036, line 363: Shouldn't we do `unCapture()` for `targetCallable.receiverType()` too?

t-kameyama: b2a478825b17

dimonchik0036 (approved):
Thanks!

PR #1918, opened 2022-02-04
[maven] Navigate to managing dependency on GotoSuperAction shortcut

fp7:
The icon on the gutter already did this navigation on a click. Now the
shortcut also works.


fp7:
was merged.

PR #1919, opened 2022-02-07
KTIJ-19944 "Convert to sealed class" intention loses comments for the enum value.

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19944


klunnii:
Merged. Thank you!

PR #1920, opened 2022-02-09
use default update channel when selection combobox is not present

eugen-yakovets:
UpdateSettings#getActiveChannels always adds EAP and RELEASE channels.
so when channel selection is locked in UpdateSettingsConfigurable dialog, it always checks EAP channels for updates.

changing to be UpdateSettings.getInstance().state.updateChannelType, so that it match behavior when checking for updates from other location (e.g. Welcome screen)


eugen-yakovets:
@chashnikov
@trespasserw

chashnikov:
[Merged](https://github.com/JetBrains/intellij-community/commit/92f14ae1e39d4cc226c05486fe34b0129c0d44df) to master, thank you!

PR #1922, opened 2022-02-09
KTIJ-21022 "Call chain on collection type may be simplified" inspection leads to "Type mismatch" with nullable type

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21022


comitative:
Thank you for the fix!

PR #1923, opened 2022-02-13
[util] Fix parameter escaping for args with non space whitespace

fp7:
Fixes IDEA-288658


fp7:
Are there some problems with this pr or is this not seen as an issue?

abusalimov:
@fp7 oh, this must have lost in my email, so sorry about that! It looks good to me, I'll merge it.

PR #1925, opened 2022-02-15
KTIJ-20265 `Add full qualifier` intention on import alias produces red code

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20265


dimonchik0036 (approved):
Thanks!

PR #1927, opened 2022-02-18
KTIJ-20748 False negative "Redundant lambda arrow" with safe call

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20748

### DIFF
--- a/plugins/kotlin/idea/tests/testData/inspectionsLocal/redundantLambdaArrow/nullableReceiver.kt
+++ b/plugins/kotlin/idea/tests/testData/inspectionsLocal/redundantLambdaArrow/nullableReceiver.kt
@@ -1,0 +1,4 @@
+// PROBLEM: Redundant lambda arrow
+// WITH_STDLIB
+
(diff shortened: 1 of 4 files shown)


plugins/kotlin/idea/tests/testData/inspectionsLocal/redundantLambdaArrow/nullableReceiver.kt
```diff
@@ -0,0 +1,4 @@
+// PROBLEM: Redundant lambda arrow
+// WITH_STDLIB
+
+fun test(list: List<String>?) = list?.map { <caret>it -> it.length }.orEmpty()
```
TWiStErRob, line 4: I would use Ivan's repro and remove orEmpty, it's much simpler. The orEmpty is not important to the issue, only the `it`.

https://youtrack.jetbrains.com/issue/KTIJ-20748#focus=Comments-27-5701887.0-0

TWiStErRob (approved):
Amazingly simple fix! Thanks for contributing this to fix my reported issue yet again.

PR #1929, opened 2022-02-19
Create isAvailable with project context for SearchEverywhereContribut…

brian-mcnamara:
…orFactory


brian-mcnamara:
We have a use case where we want a search everywhere contributor to show only for a known project. In this case, project is already in context, so adding a new method to enable the use of this context within isAvailable.

PR #1931, opened 2022-02-20
PY-41061: use __aiter__ for typing in an async for loop

theY4Kman:
This PR fixes the typing of iterands in async for loops being derived from `__iter__` instead of `__aiter__`.

Previously, typing info of an iterand was retrieved in the same way for both `for` and `async for` loops — namely, if the iterable declared an `__iter__` method, typing info would be retrieved from it; otherwise, if the iterable declared an `__aiter__` method, typing info would be pulled from it. For example:

```py
from typing import AsyncIterator, Iterator

class Foo:
    def __iter__(self) -> Iterator[list]: ...
    def __aiter__(self) -> AsyncIterator[dict]: ...

for expr in Foo():
    expr.append(1)

async for expr in Foo():
    expr.update({'tacos': 'muffins'})
    # PyCharm analysis: Unresolved attribute reference 'update' for class 'list'
```
image

This PR resolves this incorrect behaviour by altering `PyTargetExpression::getIterationType` to peer into the source expression's context to determine if it's within an `async for` statement. If and only if it's within an `async for` will typing from `__aiter__` be used; if it's _not_ within an `async for` (or this could not be determined), the typing from `__iter__` will be used.

The reasoning being: `__iter__` and `__aiter__` are completely separate iterables, and Python never "falls back" — i.e. you _never_ get `__iter__`'s typing with an `async for`, and you _never_ get `__aiter__`'s typing with a sync `for`.

### DIFF
--- a/python/python-psi-impl/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
+++ b/python/python-psi-impl/src/com/jetbrains/python/psi/impl/PyTargetExpressionImpl.java
@@ -369,7 +369,10 @@ public class PyTargetExpressionImpl extends PyBaseElementImpl<PyTargetExpression
     else if (iterableType instanceof PyUnionType) {
       return ((PyUnionType)iterableType).map(member -> getIterationType(member, source, anchor, context));
     }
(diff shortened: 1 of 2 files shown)


AsafMah:
@sproshev Any updates?

PR #1933, opened 2022-02-22
KTIJ-21082 False positive "Redundant 'suspend' modifier" when invoking a suspend lambda contained in a list which is an explicit receiver

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21082


abelkov:
Merged manually.

PR #1934, opened 2022-02-23
Fix a typo for "Variable type can be explicit" inspection

Fabrice-TIERCELIN:
Hi @BasLeijdekkers,

This replaces:

_"Reports local variables of the <code>var</code> type that can be replaced **withvan** explicit type."_

by:

_"Reports local variables of the <code>var</code> type that can be replaced **with an** explicit type."_


BasLeijdekkers:
Thanks!
https://github.com/JetBrains/intellij-community/commit/07f8fa52371f3360cbadf4d7d03d9cf229109e73

PR #1935, opened 2022-02-23
Allow overriding parallel.models.fetch with a system property

oehme:
Whether this feature works depends a lot on the individual Gradle project,
so it makes sense for this to be configurable on a per-project basis rather
than globally at the IDE level.


oehme:
FYI @nskvortsov

nskvortsov:
Looks reasonable. Merging..

oehme:
@nskvortsov ping :)

nskvortsov:
already merged to master as d5b9e98e2243a972e96151ee12cf6b387a4c0f73

PR #1936, opened 2022-02-23
Make background model conversion safe

oehme:
Gradle's Tooling API models aren't thread-safe, because they use a HashMap to represent
the different views created over the same object. Since we can't retroactively fix all
Gradle versions, we can instead work around this by doing all the model access to the
root project model before we kick off any of the additional work that might access this
model as well.

See https://github.com/gradle/gradle/issues/20008


oehme:
FYI @nskvortsov

nskvortsov:
So we effectively pre-init models. Should we make it explicit in the code? E.g. extract some kind of `intialiseGradleModels()` method with javadoc saying that it needs to be called befor parallel fetch opertaions...
Otherwise we rely on fragile code ordering that may easily be lost in refactorings.

oehme:
Yes, I'll do that

oehme:
@nskvortsov this should be ready to merge

oehme:
@nskvortsov any news on this one?

nskvortsov:
merged as 2c3cb5f5bd01b916b685faac101beba74059a957

PR #1938, opened 2022-02-27
KTIJ-20924 Generate equals/hashCode/toString: don't delegate to abstract method

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20924


vladimirdolzhenko:
LGTM, Thank you so much

PR #1939, opened 2022-02-28
PY-52656 patch typeshed to trick PyCharm into using the correct type parameters

KotlinIsland:
PyCharm treats the inheritance and order of type parameters in a strange way, this hack will make PyCharm use the correct types for `dict.values()`.


east825:
Closing as it's no longer necessary, since the underlying problems with generics were fixed as part of [PY-53896](https://youtrack.jetbrains.com/issue/PY-53896).

PR #1945, opened 2022-03-03
KTIJ-21197 Postfix templates may change semantics of expressions containing `.Companion`

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21197


abelkov:
Merged manually.

PR #1946, opened 2022-03-03
KTIJ-21220 "Add abstract fun" quickfix generates broken code when called on function with default arguments

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21220


plugins/kotlin/idea/tests/testData/quickfix/override/nothingToOverride/addFunctionWithoutDefaultValue.kt.after
```diff
@@ -0,0 +1,9 @@
+// "Add 'abstract fun foo(x: String)' to 'I'" "true"
+// ERROR: An overriding function is not allowed to specify default values for its parameters
+interface I {
+    fun foo(x: String)
+}
+
+class Foo : I {
+    override fun foo(x: String = "") {}
```
nikitabobko, line 8: I'd expect the refactoring to generate default values in the base class/interface because override declarations are prohibited to have default parameters. Like this:
```
interface I {
    fun foo(x: String)
}

class Foo : I{
    override fun foo(x: String = "") {}
}
```

nikitabobko (changes requested):

t-kameyama:
Fixed:
スクリーンショット 2022-03-06 18 35 30
スクリーンショット 2022-03-06 18 35 40

nikitabobko (approved):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/AddFunctionToSupertypeFix.kt
```diff
@@ -82,6 +83,15 @@ class AddFunctionToSupertypeFix private constructor(
         }
     }
 
+    private fun KtNamedFunction.removeDefaultValueIfNeeded() {
+        if (hasModifier(KtTokens.OVERRIDE_KEYWORD)) {
```
nikitabobko, line 87: Why do we need `hasModifier(KtTokens.OVERRIDE_KEYWORD)` check? Isn't it always true?

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/AddFunctionToSupertypeFix.kt
```diff
@@ -141,12 +150,20 @@ class AddFunctionToSupertypeFix private constructor(
...
+                IdeDescriptorRenderers.SOURCE_CODE_SHORT_NAMES_NO_ANNOTATIONS.withDefaultValueOption(project).render(functionDescriptor),
                 sourceCode,
                 targetClass
             )
         }
 
+        private fun DescriptorRenderer.withDefaultValueOption(project: Project): DescriptorRenderer {
+            return withOptions {
+                this.defaultParameterValueRenderer = {
+                    OptionalParametersHelper.defaultParameterValueExpression(it, project)!!.text
```
nikitabobko, line 162:
```suggestion
                    OptionalParametersHelper.defaultParameterValueExpression(it, project)?.text
                        ?: error("value parameter renderer shouldn't be called when there is nothing to render")
```

PR #1949, opened 2022-03-05
KTIJ-20728 False positive "'foo' always returns non-null type" for function with non-local return in expression body

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20728


vladimirdolzhenko:
LGTM, thanks

PR #1953, opened 2022-03-09
IDEA-258869 Added registry option to disable forcing pinned tabs to the left

momomo:


momomo:
Note, the ticket and discussion for this is the youtrack ticket with the same name: 

https://youtrack.jetbrains.com/issue/IDEA-258869

Please just approve this pull request ASAP as it has been long overdue. It works, has been tested with on and off and defautls to the leaves the current behaviour as is but with this option to disable pinning tabs to the left.

momomo:
@yole Thank you!

PR #1954, opened 2022-03-10
KTIJ-21270 "AssertionError: Failed to create expression from text" on "Remove unnecessary non-null assertion" with line break between receiver and !!

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21270


dimonchik0036 (approved):
Thanks!

PR #1960, opened 2022-03-17
IDEA-191422 $n should be used for Back references

floating-cat:
Fix https://youtrack.jetbrains.com/issue/IDEA-191422


floating-cat:
The PR is actually wrong so I am closing this.

PR #1963, opened 2022-03-20
IDEA-290742 Gather the Java downgrading inspections

Fabrice-TIERCELIN:
Hi @BasLeijdekkers,

This PR moves all the Java downgrading inspections into a unique path. Those inspections are used very differently from the others so it would be better to identify them easily. This way, they will be gathered like this (changes into **bold**):

- Java
- - Java language level **downgrading** issues
- - - Annotation
- - - Annotation class
- - - 'assert' statement
- - - **Auto-boxing**
- - - **Auto-unboxing**
- - - **Diamond can be replaced with explicit type arguments**
- - - Enhanced 'for' statement
- - - **Enhanced 'switch'**
- - - Enumerated class
- - - **Multi-catch can be split into separate catch blocks**
- - - **Lambda can be replaced with anonymous class**
- - - **Stream API call chain can be replaced with loop**
- - - **'Optional' can be replaced with sequence of 'if' statements**
- - - **Text block can be replaced with regular string literal**
- - - **Underscores in numeric literal**
- - - Varargs method
- - - **Variable type can be explicit**

I have successfully tested it by injecting the code into my _IntelliJ 2021.2.1_ installation.

### DIFF
--- a/java/java-impl/src/inspectionDescriptions/EnhancedSwitchBackwardMigration.html
+++ b/java/java-impl/src/inspectionDescriptions/EnhancedSwitchBackwardMigration.html
@@ -26,6 +26,11 @@ statements.
 }
 </code></pre>
 <!-- tooltip end -->
+<p>
+  <i>Enhanced</i> <code>switch</code> appeared in Java 14.
+  This inspection can help to downgrade for backward compatibility with earlier Java versions.
+</p>
+<!--p>This inspection only reports if the language level of the project or module is 14 or higher.</p-->
 <p><small>New in 2019.1</small></p>
 </body>
 </html>
--- a/java/java-impl/src/inspectionDescriptions/OptionalToIf.html
+++ b/java/java-impl/src/inspectionDescriptions/OptionalToIf.html
@@ -17,6 +17,10 @@ Reports <code>Optional</code> call chains that can be replaced with a sequence o
   return getDefault();
 </code></pre>
 <!-- tooltip end -->
+<p>
+  <code>Optional</code> appeared in Java 8.
+  This inspection can help to downgrade for backward compatibility with earlier Java versions.
+</p>
 <p>This inspection only reports if the language level of the project or module is 8 or higher.</p>
 <p><small>New in 2020.2</small></p>
 </body>
--- a/java/java-impl/src/inspectionDescriptions/RemoveLiteralUnderscores.html
+++ b/java/java-impl/src/inspectionDescriptions/RemoveLiteralUnderscores.html
@@ -4,6 +4,10 @@ Reports numeric literals with underscores and suggests removing them with a quic
 This may be useful if you need to lower the language level.
 <p>The quick-fix removes underscores from numeric literals. For example <code>1_000_000</code> will be converted to <code>1000000</code>.</p>
 <!-- tooltip end -->
+<p>
+  <i>Numeric literals with underscores</i> appeared in Java 7.
+  This inspection can help to downgrade for backward compatibility with earlier Java versions.
+</p>
 <p>This inspection only reports if the language level of the project or module is 7 or higher.</p>
 <p><small>New in 2020.2</small></p>
 </body>
--- a/java/java-impl/src/inspectionDescriptions/StreamToLoop.html
+++ b/java/java-impl/src/inspectionDescriptions/StreamToLoop.html
@@ -29,6 +29,11 @@ Reports Stream API chains, <code>Iterable.forEach()</code>, and <code>Map.forEac
   the stream short-circuits.
 </p>
 <!-- tooltip end -->
+<p>
+  <i>Stream API</i> appeared in Java 8.
+  This inspection can help to downgrade for backward compatibility with earlier Java versions.
+</p>
+<!--p>This inspection only reports if the language level of the project or module is 8 or higher.</p-->
 <p>Configure the inspection:</p>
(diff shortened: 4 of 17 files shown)


akozlova:
This way, you loose the ability to understand to which language level you want to downgrade. Currently those inspections are located under corresponding level and that's make it easier to align to my taste.

Fabrice-TIERCELIN:
OK, I will create subfolders

akozlova:
In most cases with one inspection under the folder. This sounds very inconvenient to me. Could you please explain your use case first?

Fabrice-TIERCELIN:
- **Use case 1**: I have to integrate a part of Java code that have been coded with a newer Java level version than the JVM of the target project, so I only want to run the downgrading inspections
- **Use case 2**: I want to improve the quality of my code, so I'm not interested by any downgrading inspection

The main reason of this PR is that you never migrate forward and backward at the same time so it is pointless to put such eclectic features so close. And it is useful to easily identify all the downgrading inspections.

What do you prefer?

- Prefix the title of the downgrading inspections with `Java XX:...`
- Melt the downgrading inspections with the upgrading inspections (so we mostly move some downgrading inspections from _Java/Code style issues_ or _Java/Performance_)
- Melt the downgrading inspections with the upgrading inspections adding `downgrading` in their title

akozlova:
Downgrading inspections are disabled by default or have information severity so you should not have troubles in the "Use case 2" at all.
When you need to downgrade, you need to set project to the higher level, then apply transformations, then downgrade, then return - otherwise it won't work in many cases. Looks like, explicitly going through ~5 groups and checking there inspections is not that bad experience. And I think that this use case is actually quite rare, no? 
Don't get me wrong, but this would change the placement of inspections, would make some comments in different forums/help outdated. And I see not that much additional value for that cost.

Fabrice-TIERCELIN:
I'm pretty sure there is a minimal change we can do to improve the situation. @akozlova, do you think the current state is the best state ever we can have?

Fabrice-TIERCELIN:
OK, @akozlova, I have reduced the PR to a very minimal change: I have added the word _downgraded_ in the inspection names so that it is easier to find them. For instance, it was hard to guess that _Enhanced 'for' statement_ was made to remove an enhanced 'for' statement, wasn't it? The comments in the forums/help won't be outdated because the names are not replaced, they are just more detailed.

Is it OK?

Fabrice-TIERCELIN:
That's right. The sentence has no sense when you have to remove the element.

akozlova:
"may be removed to downgrade" sounds for me that the class is unused but it's not true. Sorry, I have no idea how to change this, especially that I don't think that anybody would try to downgrade now to java 4. Could we keep those old names please?

Fabrice-TIERCELIN:
And what if we only write a generic paragraph in the description like this:

> This inspection can be used for backward compatibility with Java `N-1` or earlier. `name_of_the_feature` appears in Java `N`. This inspection helps to downgrade.

The _Enhanced 'switch'_ inspection doesn't mention at all that it is a backward inspection.

akozlova:
Yes, updating descriptions sounds good to me. There is a clause about "updating" inspections, so the other way round should work well. Thanks

Fabrice-TIERCELIN:
I have updated the PR, successfully tested all the descriptions by injecting the HTMLs in my _IntelliJ_ installation and pushed it.

This way, a user can retrieve the downgrading inspections with one of those three keywords:
1. `downgrade`
2. `backward`
3. `compatibility`

java/java-impl/src/inspectionDescriptions/EnhancedSwitchBackwardMigration.html
```diff
@@ -26,6 +26,11 @@
 }
 </code></pre>
 <!-- tooltip end -->
+<p>
+  This inspection can help to downgrade for backward compatibility with Java 13 or earlier.
+  <i>Enhanced</i> <code>switch</code> appears in Java 14.
```
akozlova, line 31: I'd skip the second sentence, it's kind of clear from the first, no?

Fabrice-TIERCELIN: OK so I have inverted the two sentences and I have replaced `Java 13 or earlier` by `earlier Java versions` to remove the redundancy.

I have also replaced `appears` with `appeared` because it's the past.

Is it OK?

akozlova: Yes, thanks!

akozlova:
I am merging the request now, I would skip "annotation, enum, etc" constructs because they just highlight specific elements but they don't help to downgrade actually. On the positive side, I don't think anyone would use java 1.4 now...

Fabrice-TIERCELIN:
Thank you for merging!

For your information, my initial plan was to merge all the downgrading inspections into a single one and add a combo box to select the targeted Java version. And the initial PR was a first step for it.

You may have noticed that some inspections run even if the Java version does not fit. I will create a PR to make them run only if the Java version is high enough.

akozlova:
In many cases that's just fine, if there is no lambda expression, then the visitor with only `visitLambdaExpression` would be mostly equivalent to an empty visitor. 
BTW I think, it would be useful first to agree on the shape of the feature in YT and then start working on PR. Would save you time and for me it would be easier to suggest you smth when there is no work already done.

PR #1964, opened 2022-03-22
[decompiler] Fix enums with overrides having the sealed modifier

coehlrich:
Enum's can't become sealed from the source code but the compiler will make the enum sealed if there are constants in the enum that have overrides.


amaembo:
Looks good, thanks

amaembo (approved):

amaembo:
Merged to internal repo, thanks. For some reason, synchronization with public repo is lagging. We'll take a look.

PR #1970, opened 2022-03-24
IDEA-290939 support more default values in code generation

entrypointkr:
Writing test case... do not merge this until additional commit!
https://youtrack.jetbrains.com/issue/IDEA-290939


entrypointkr:
Ready to get review @akozlova.
Review points:
1. Defaulting for `Optional` perfectly reasonable, then also good for Collection<A>, List<A>, even String?
2. At this [code](https://github.com/JetBrains/intellij-community/pull/1970/files#diff-8f43ca2a380002447e08c9b8ce0d839138238afd21be1e0b607f786ed90c24b6R101), will find the default by `PsiType#getCanonicalText`. If not found, find again using `PsiType#equalsToText`. Is there a case that need `equalsToText` instead of `getCanonicalText`? if not, the `equalsToText` method need to be removed.

Thank you for your effort in advance.

platform/core-api/src/com/intellij/psi/CommonClassNames.java
```diff
@@ -28,6 +28,7 @@
 
   String JAVA_LANG_ITERABLE = "java.lang.Iterable";
   String JAVA_UTIL_ITERATOR = "java.util.Iterator";
+  String JAVA_UTIL_LIST_ITERATOR = "java.util.ListIterator";
```
akozlova, line 31: it's more common than navigatable collections but still not common enough for this class

java/java-psi-api/src/com/intellij/psi/util/PsiTypesUtil.java
```diff
@@ -75,6 +92,18 @@ public static Object getDefaultValue(PsiType type) {
     }
   }
 
+  // Please make this public if needed from outer
+  @Nullable
+  private static String getDefaultOfTypeFromCustom(PsiType type) {
+    String defaultExpression = ourCommonDefaults.get(type.getCanonicalText());
+    return defaultExpression == null
+           ? ourCommonDefaults.entrySet().stream()
```
akozlova, line 100: I believe this code would not find anything if no defaultExpression is found

akozlova:
I don't understand when you would use these defaults actually. To me, in mostly 99% of cases these defaults would be overridden and it's better to have nothing than smth which pretend to be ok. Could you please elaborate why you need this? While the code is ok (see my comments). Thanks

entrypointkr:
@akozlova The reason is same with `Optional`, as it basically a single size of `List`.
The use case is, if we need kind of quick test to run a code without runtime error and redundant null check:
```java
interface PersonRepository {
    List<Person> getAllPersons();
}
```

```java
class SQLPersonRepository implements PersonRepository {  }

class EmptyPersonRepository implements PersonRepository {
    // Auto gen with defaults
{
```

```java
PersonRepository repo = getRandomRepositoryWithReflection();
for (Person person : repo.getAllPersons()) {
    // at this case, commonly, we're not expecting the `getAllPersons()` will returns null.
}
```

akozlova:
Generally, it's making defaults more complicated. Optional was implemented because it's a known bad practice to return null from Optional method, there is no such problem with other types (at least other types were not invented to prevent nulls). Thus I generally don't like the change, I am sorry. 
One more argument against: if you insert `Collections.emptyList()`, you most probably insert import. If people would change the code (and most probably they would), then import would be unused and probably would be forgotten in the code (e.g. if optimize imports on commit is disabled). This way, these defaults would make the code worser. At the same time, I don't see big advantages - if you run test and fix the code - it's ok to fix NPEs as well as empty collections.
Again, I am sorry for complaining in the PR but it would land for all users. 

I think, it's possible to implement via template, do you want me to help you with it?

entrypointkr:
@akozlova I agree with the import issue, even we can use optimize import, in many place it can being leaved as unused. Just remind about the argues,

> The Optional implemented because nulling it is a known bad practice

1. Then why is it a bad practice essentially? because `Optional` invented to replace `null`.
2. Then why `Optional` invented? as Oracle says, avoid NPE and redundant null checking.

In this context, even if only `Optional` invented to avoid `null`, it's goal is actually to avoid NPE and redundant null checking.
Then why we not treat `List.empty()` to avoid NPE and redundant checking? it's essentially a same. Reminding again:

```java
List<Person> persons = personRepo.getAllPersons();
if (persons != null) {
    for (Person person : persons) {
        // TODO
    }
}
```

> Insert Collections.emptyList() causes unused imports

Even it seems trivial, I 100% agree it is worse. Then we can `List.of()` on Java9+? 

---

Please close the PR and issue still it not motivating to go.

entrypointkr:
@akozlova 
As of December 2025, when I try code generation in IntelliJ IDEA, it produces `Stream.empty()`, `Map.of()`, `""` and `List.of()` as the default return values for methods whose return types are `Stream`, `Map`, `String` and `List`, respectively. In practice, this looks very close to the idea I originally proposed. From what I can tell by looking at the public history, this behavior seems to have been introduced around commit [[java-intentions] AddVariableInitializerFix: suggest more suitable initial values instead of null for some types](https://github.com/JetBrains/intellij-community/commit/5bd02b55c32d58e7ac7057dce5025f8510c6396e#diff-8f43ca2a380002447e08c9b8ce0d839138238afd21be1e0b607f786ed90c24b6R148).

At the time, however, the feedback I received was that such defaults would not be particularly useful in real projects, would be overridden almost all of the time, and that generating seemingly “reasonable” implementations might even be worse than generating nothing, since they could leave behind unnecessary imports and add complexity. From my perspective, the current defaults now seem to follow exactly the kind of approach that was previously discouraged.

It is a bit discouraging that, although the current behavior now aligns so closely with what I suggested back then, my earlier input does not appear to be visible or credited anywhere (for example, in the issue or contributor history). Because of that, I find the situation somewhat ironic and would really like to understand how that earlier rationale fits with the present implementation. Have views on default values changed since then, or was there another consideration behind adopting this behavior after all?

PR #1977, opened 2022-04-02
[IDEA-291518] Update Maven templates

ascopes:
Existing Maven templates for new projects omit the `<project.build.sourceEncoding>` tag, which results in Maven using a system-dependant locale, and omitting the following warning:

```
[WARNING] File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
```

Ideally we probably do not want to have warnings being raised from new projects before we start adding anything to them. 

I have amended the standard Maven template to include this by default, and amended the Groovy template to always include a `<properties>` block as to be able to always include this.


nskvortsov:
Thank you for your contribution,
Please, open a ticket in https://youtrack.jetbrains.com/issues/IDEA and copy the problem description there, so we have it for future reference. Changes themselves look safe to be merged.

ascopes:
@nskvortsov opened https://youtrack.jetbrains.com/issue/IDEA-291518 in response. Thanks.

nskvortsov:
@ascopes thank you! I've launched the patch into our CI infrastructure. Looking forward to seeing it in the repo.

PR #1979, opened 2022-04-06
Set total num disks to 1 in the zip64 EOCD locator

gharrma:
...otherwise some tools get confused and think that the zip
is a split archive.

For example, when ijar sees that zip64_total_disks != 1,
it complains, "multi-disk JAR files are not supported":
github.com/bazelbuild/bazel/blob/master/third_party/ijar/zip.cc

Similarly, running `zip --test` on the lib/app.jar file in IJ 221
produces a warning: "archive name must end in .zip for splits".

---

cc @develar 
Let me know what you think. Since it is a trivial fix, I did not bother filing a YouTrack issue.


gharrma:
This issue partially blocks merging IntelliJ 2022.1 into Android Studio, because some of our build tools fail to read certain jars in IntelliJ.

develar:
Thanks for such a simple but important fix! I pushed the fix to main branch, then after several days I will push to 221 branch.

PR #1982, opened 2022-04-10
Jump to type source by Ctrl+click on a debug inlay

stachenov:
I often need to navigate to the source of the type of some runtime value, usually when the runtime type is different from compile-time type (polymorphism). This PR introduces a very convenient way to do it by simply Ctrl+clicking the appropriate debugger inlay in the editor.

It’s kind of a mix of copy-paste and cargo cult programming, with most code dragged from `XJumpToSourceActionBase` and edited accordingly. It seems to work just fine, but may be not the “right” way to do it, as I’ve just began to digging in IntelliJ IDEA code, so I don’t have thorough understanding of it yet.


gorrus:
see also [IDEA-140499](https://youtrack.jetbrains.com/issue/IDEA-140499) In debugger, jumping to source of actual type, instead of source of declared type, or having to manually choose implementation

PR #1985, opened 2022-04-12
Expose scheduling selection to sub-classes in AutoScrollFromSourceHandler

abrooksv:
Allow sub-classes of AutoScrollFromSourceHandler to schedule selection changes from other sources such as caret listeners or when the UI element becomes visible


akozlova:
Merged manually, should appear on GitHub later today/tomorrow. Thanks!

PR #1986, opened 2022-04-13
[java] IDEA-277768 Fix "never assigned" warning for org.easymock.Mock

beatngu13:
Fixes [IDEA-277769](https://youtrack.jetbrains.com/issue/IDEA-277768).

Side note: coding guidelines for [commits](https://plugins.jetbrains.com/docs/intellij/intellij-coding-guidelines.html#commits) say:

>Please avoid moving or renaming classes unless it is necessary for the fix.

I moved the `mockedField/` directory to `mockedMockitoField/` since there is now also a `mockedEasyMockField/` directory, and I thought this way it is more clear.


akozlova:
Merged manually, should appear later (hopefully today). Thanks!

PR #1990, opened 2022-04-19
[EE-IR] Support local classes

kandersen:
!! Depends on #1973 !!

---

The source of local classes needs to be included in the IR compilation in order to compute the "binary" names for the local classes in the compiled code, rather than have fragment compilation attempt to "predict" the names.


kandersen:
Rebased on kt-222-master.

nikita-nazarov:
Could you please rebase this branch on kt-222-master as #1973 has already landed?

kandersen:
Rebased!

PR #1991, opened 2022-04-20
[EE-IR] Remove unsupported configuration from test suite

kandersen:
With 2 JVM backends and 2 JVM evaluators, there are 4 combinations when
compiling code for the JVM and debugging it with the evaluator:

  * old backend, old evaluator: the old configuration, that anyone on old
    releases of IntelliJ/Kotlin will see.

  * new backend, old evaluator: the current configuration, released with the
    IntelliJ and Kotlin plugins shipping now. IR backend is default, but the new
    evaluator has not shipped. This means that the compiler used to compile the
    project is not the same used to evaluate expressions.

  * new backend, new evaluator: the next configuration, as we enable the new
    evaluator.

That leaves:

  * old backend, new evaluator: this one doesn't "naturally" occur: people will
    have to bend over backwards with deprecated command line flags and/or
    registry keys.

Let's erase from the test corpus to imporove the SNR of the test suite.


nikita-nazarov:
👍 Will merge this after #1973.

kandersen:
Rebased!

nikita-nazarov:
Hi Kris!
The code itself is fine, however I stumbled upon some test failures with this pr. The reason is that `KotlinEvaluateExpressionTestGenerated` inherits `AbstractKotlinEvaluateExpressionTest`, which transitively inherits `KotlinDescriptorTestCase`. At the same time `KotlinDescriptorTestCase` declares the default fragment compiler to be `JVM_IR`. As the result, `KotlinEvaluateExpressionTestGenerated` which should represent the 'old backend + old evaluator' combination now turns into the 'old backend + IR evaluator' one. Let's add 
```
override fun fragmentCompilerBackend() = 
    FragmentCompilerBackend.JVM
```
to `AbstractKotlinEvaluateExpressionTest`.

kandersen:
Thanks for the catch! Adjusted accordingly and rebased!

PR #1992, opened 2022-04-20
Mark code fragments as USED_AS_EXPRESSION

kandersen:
Code fragments submitted to the evaluator are analyzed by the front-end.

psi2ir relies on the `USED_AS_EXPRESSION` analysis to correctly determine the type of a when expression. If it's not used, the result is coerced to Unit.

This caused when expressions as the last statement (and thus, result) in a block submitted to the evaluator to be discarded.

Tagging the fragment as used resolves this.


nikita-nazarov:
All good 👍 Will merge after #1973

PR #1993, opened 2022-04-20
fix: VS Code shift alt down wrong action

philfontaine:
The shift alt down shortcut in vscode is much closer to EditorDuplicateLines than EditorDuplicate

## Shift alt down in VS Code (Copy Line Down):

image
image

## EditorDuplicate in Jetbrains:

image
image

## EditorDuplicateLines in Jetbrains:

image
image


batya239:
Merged, thanks for contribution!

PR #1996, opened 2022-04-23
Kotlin unwrap improvements KTIJ-21561 KTIJ-21562 KTIJ-11687

alshain:
Various fixes and improvements to Kotlin's "Unwrap" actions.

- Unwrap parameters now works for most sensible situations (with tests to prove it)
- Remove qualified receiver when unwrapping parameters / lambdas (with tests)
- Strike through text to be removed
- Usability improvements: better speed search / make choices more distinguishable

Most of the work went into improving the unwrapping of parameters. The previous implementation only supported unwrapping a single argument call when it was contained inside another function call, e.g. 

    notUnwrappable(other, canBeRemoved(123), other) 
    ↓ 
    notUnwrappable(other, 123, other)

The new implementation supports any sensible scenario I could think of.

- Top level call unwrapping
- Non-nested call unwrapping
- Delete receiver when unwrapping qualified call
- Support multi argument calls
- Allow caret on method name, paren, comma or whitespace
- Prefer parameter unwrapping over removal for top level function calls
- Fix strikethrough preview

Fixes KTIJ-21561 KTIJ-21562 KTIJ-11687

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
@@ -196,6 +196,9 @@ class KotlinFunctionParameterUnwrapper(val key: String) : KotlinUnwrapRemoveBase
         // but because we support the caret to be on LPAR/COMMA and other tokens this doesn't work
 
         val target = argumentToUnwrap(e) ?: return "<unknown>" // should not be possible
-        return message(key, getExpressionShortText(target))
+        val callee = target.getStrictParentOfType<KtCallExpression>()?.calleeExpression
+            ?.let(::getExpressionShortText) ?: "?"
+
+        return message(key, callee, getExpressionShortText(target))
     }
 }
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
@@ -2,26 +2,200 @@
 
 package org.jetbrains.kotlin.idea.codeInsight.unwrap
 
+import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElement
-import org.jetbrains.kotlin.psi.KtCallExpression
-import org.jetbrains.kotlin.psi.KtValueArgument
+import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.util.elementType
+import org.jetbrains.kotlin.idea.KotlinBundle.message
+import org.jetbrains.kotlin.idea.refactoring.getExpressionShortText
+import org.jetbrains.kotlin.idea.util.ifTrue
+import org.jetbrains.kotlin.idea.util.isComma
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getNextSiblingIgnoringWhitespaceAndComments
+import org.jetbrains.kotlin.psi.psiUtil.getPrevSiblingIgnoringWhitespaceAndComments
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
-class KotlinFunctionParameterUnwrapper(key: String) : KotlinUnwrapRemoveBase(key) {
+class KotlinFunctionParameterUnwrapper(val key: String) : KotlinUnwrapRemoveBase(key) {
 
     override fun isApplicableTo(element: PsiElement): Boolean {
-        if (element !is KtCallExpression) return false
-        if (element.valueArguments.size != 1) return false
-        val argument = element.parent as? KtValueArgument ?: return false
-        if (argument.getStrictParentOfType<KtCallExpression>() == null) return false
+        val argumentToUnwrap = argumentToUnwrap(element) ?: return false
+        deletionTarget(argumentToUnwrap) ?: return false
+
+        return true
+    }
+
+    private fun argumentToUnwrap(element: PsiElement?): KtValueArgument? {
+        if (element == null) return null
+
+        nearbySingleArg(element)?.let {
+            return it
+        }
+        when (element) {
+            is KtValueArgument -> return element
+            else -> {
+                if (element.parent !is KtValueArgumentList) return null
+
+                if (element.isComma || element.elementType == KtTokens.RPAR || element is PsiWhiteSpace || element is PsiComment) {
+                    return findAdjacentValueArgumentInsideParen(element)
+                } else return null
+            }
+        }
+    }
+
+    /**
+     * When outside a function call's parenthesized arguments but the function has a single argument, choose that
+     * ```
+     *           ┌ LPAR
+     *           │
+     *           v
+     * methodCall(arguments)
+     * ^^^^^^^^^^
+     * │
+     * └ Identifier
+     * ```
+     *
+     */
+    private fun nearbySingleArg(element: PsiElement): KtValueArgument? {
+        fun caretOnQualifier(qualifiedExpression: KtQualifiedExpression): KtCallExpression? {
+            var candidate: KtQualifiedExpression = qualifiedExpression
+
+            // a.b.c.d(123)
+            // ^^^^^^^
+            // walk up chained dot expressions until we find the one that has a call as rhs
+            // we use this complicated rather than just have isApplicableTo apply to a KtQualifiedExpression to get priority over
+            // "Remove"
+            while (candidate.selectorExpression !is KtCallExpression) {
+                candidate = candidate.parent.safeAs<KtQualifiedExpression>() ?: return null
+            }
+            return candidate.selectorExpression as KtCallExpression
+        }
+
+        val callExpression: KtCallExpression = when (element.elementType) {
+            KtTokens.LPAR -> {
+                //  methodName<caret>(123)
+                element.parent?.safeAs<KtValueArgumentList>()
+                    ?.parent?.safeAs<KtCallExpression>() ?: return null
+            }
+            KtTokens.DOT, KtTokens.SAFE_ACCESS -> {
+                element.parent?.safeAs<KtQualifiedExpression>()
+                    ?.let { caretOnQualifier(it) } ?: return null
+            }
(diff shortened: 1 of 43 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
```diff
@@ -3,25 +3,176 @@
...
+    }
+
+    private fun argumentToUnwrap(element: PsiElement?): KtValueArgument? {
+        if (element == null) return null
+
+        nearbySingleArg(element)?.let {
+            return it
+        }
+        when (element) {
+            // KtLambdaArgument is also a KtValueArgument, but it's already handled by KotlinLambdaUnwrapper
```
nikitabobko, line 34: It's not true that `KtLambdaArgument` is already handled by `KotlinLambdaUnwrapper`.

`KotlinLambdaUnwrapper` should unwrap it this way:
```
run { println("Hi") } -> println("Hi")
```

`KotlinFunctionParameterUnwrapper` should unwrap it this way:
```
run { println("Hi") } -> { println("Hi") }
```

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
```diff
@@ -3,25 +3,176 @@
...
+        return function
+    }
+
+    override fun collectAffectedElements(e: PsiElement, toExtract: MutableList<PsiElement>): PsiElement {
+        val valueArgument = argumentToUnwrap(e) ?: return e
+        val deletionTarget = deletionTarget(valueArgument) ?: e
+
+        toExtract.add(valueArgument)
+
+        return deletionTarget
```
nikitabobko, line 165:
```suggestion
        super.collectAffectedElements(e, toExtract)
        return argumentToUnwrap(e)?.let { deletionTarget(it) } ?: e
```

nikitabobko: For same reasons as `collectAffectedElements` in [KotlinLambdaUnwrapper](https://github.com/JetBrains/intellij-community/pull/1996/files#diff-a249d4cd35f4ed08fe5c4342e64d9d275f142bb0c01e5f9e918f7f6a1fb1cc18)

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
```diff
@@ -3,25 +3,176 @@
...
+    }
+
+    override fun getDescription(e: PsiElement): String {
+        // necessary because the base implementation expects to be called for KtElement
+        // but because we support the caret to be on LPAR/COMMA and other tokens this doesn't work
+
+        val target = argumentToUnwrap(e) ?: return "<unknown>" // should not be possible
```
nikitabobko, line 172:
```suggestion
        val target = argumentToUnwrap(e) ?: error("Description asked for a non applicable unwrapper")
```

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
```diff
@@ -3,25 +3,176 @@
...
+            return it
+        }
+        when (element) {
+            // KtLambdaArgument is also a KtValueArgument, but it's already handled by KotlinLambdaUnwrapper
+            is KtLambdaArgument -> return null
+            is KtValueArgument -> return element
+            else -> {
+                if (element.parent !is KtValueArgumentList) return null
+
+                if (element.isComma || element.elementType == KtTokens.RPAR || element is PsiWhiteSpace) {
```
nikitabobko, line 40: comments inside argument list are not covered:
```
fun foo(arg: Int) = arg

fun main() {
    foo(/* arg<caret> = */ 1)
}
```

nikitabobko (changes requested):
Hi! Thanks for the PR. Please, take a look at the comments I left. Let me know if something isn't clear, or you don't agree with something.

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
```diff
@@ -3,25 +3,176 @@
...
+
+              valueArgumentList
+          }
+          KtTokens.IDENTIFIER -> {
+              // method<caret>Name(123)
+              // <caret>methodName(123)
+              val parent = element.parent
+              val refExpression = parent as? KtReferenceExpression ?: return null
+              val callExpression = refExpression.parent as? KtCallExpression ?: return null
+              callExpression.valueArgumentList ?: return null
```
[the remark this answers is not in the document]
alshain: Ah cool, I didn't realize you had the `safeAs` helper function.

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
```diff
@@ -3,25 +3,176 @@
...
+          }
+          else -> return null
+        }
```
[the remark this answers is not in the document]
alshain: Fixed

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
alshain: You certainly took a good look at my change, thanks :)

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
alshain: Done

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
alshain: If IntelliJ ever reached that error, it wouldn't show any unwrap actions at all, no? I don't think throwing an Exception here is worth it.

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
alshain: Going for the luxury version, I see. Done.

alshain:
Any chance for this to be included in 2022.2? :)

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/unwrap/KotlinFunctionParameterUnwrapper.kt
nikitabobko: If you think that something isn't achievable then you should throw an exception. It's the idea of exceptions. Right now, you have unlocalized "<unknown>" string

alshain:
Yeah, that looks good to me. Thanks! 

Why the star projection in takeWhileIsInstance though? Doesn't that allow you to write non-sensical instance checks that will never match?

Regarding the exception: I agree with you. I had just seen that IntelliJ never catches these exceptions and would then not show *any* unwrap options whatsoever instead of just excluding the faulty ones.

nikitabobko:
> Why the star projection in takeWhileIsInstance though? Doesn't that allow you to write non-sensical instance checks that will never match?

`takeWhileIsInstance` is identical to `filterIsInstance` in Kotlin stdlib. Returned `Sequence<T>` contains only elements of `T`

> Also, I'd have kept the comment regarding the priority-over-"remove"-hack.

I actually, didn't get the meaning of this comment. What priority over what "remove" are you trying to get here? For me, it seems like a simple "get KtCallExpression selector of topmost KtQualifiedExpression" code

alshain:
Thanks for the prompt reply.

Regarding the comment, consider what IntelliJ does when I invoke unwrap here:
image

Given two unwrappers U1 and U2, and PSI node under the cursor (P1 with parents P2 and P3), IntelliJ will compute the following list:

* U1(P1) 
* U2(P1)
* U1(P2)
* U2(P2)
* U1(P3)
* U2(P3)

But all unwrappers are called for each node in the parentsWithSelf sequence. This populates the list of unwrap suggestions (omitting null results, of course) and the choices presented by to the user are ordered exactly in the way the unwrappers are called.

So if I(U2) want to my unwrap to affect P3, it would be natural to simply do a type check for `P3` when IntelliJ calls me and abort unless it's a P3, because IntelliJ will call me for each PSI node in the chain.

However, if the other unwrapper U1 already acts on P2, U1(P2) will be in the unwrap list before my own unwrap. But the "Remove" unwrapper seems less useful to me than the "UnwrapParameter", so I made sure "UnwrapParameter" gets to act first by acting on P1, but actually affecting P3. I hope that makes sense.

I didn't want to change the order of unwrappers themselves because I didn't know what other unpredictable effects on the order of unwrap choices that would have.




> takeWhileIsInstance is identical to filterIsInstance in Kotlin stdlib. Returned Sequence<T> contains only elements of T

I just would have expected it to be defined as `inline fun <reified S, reified T: S> Sequence<S>.takeWhileIsInstance(): Sequence<T>`. Now I wonder the same thing about `filterIsInstance` :)

nikitabobko:
Thanks for the explanation, now it makes sense.

> I just would have expected it to be defined as inline fun <reified S, reified T: S> Sequence<S>.takeWhileIsInstance(): Sequence<T>. Now I wonder the same thing about filterIsInstance :)

It would require specifying the original type on every call site. Like `sequenceOf(1, 2).takeWhileIsInstance<Number, Int>()` which is too verbose. And, unfortunately, you can't use [Kotlin 1.7.0 type argument underscore operator](https://youtrack.jetbrains.com/issue/KT-30485) because, in this case, type inference will downcast the receiver to the `Sequence<Any>`. I don't think that it can be fixed with current Kotlin type inference system

PR #1997, opened 2022-04-25
VS Code Keymap: use ctrl shift up/down to move with selection 

philfontaine:
Using shift ctrl up works fine in VS Code but does not work in Jetbrains IDE. I'm often using shift + ctrl + up/right/down/left to move around with selection, and this does not work right now for up/down.


Navnedia:
I was just about to make a pull request for this myself, I think this would be a great feature to have in the VSCode keymap extension and better work with your muscle memory. Would love to see this get merged or at least manually added.

batya239:
Merged, thanks for contribution!

PR #1999, opened 2022-04-27
[EE-IR] Support the Sequence Debugger

kandersen:
Not _strictly_ dependent on https://github.com/JetBrains/kotlin/pull/4804 but related.

---

The sequence debugger test suite revealed two flaws in the EE-IR implementation:

For the purposes of computing the JVM signature of the entrypoint to the fragment code, the EE-IR compiler _guesses_ at the correct function to account for inline class mangling. The guess was a little off, as revealed by the occurence of local functions being generated for SAM conversions.

Secondly, calls to local functions from _lambdas_ in the fragment were not being accurately rewritten to target the compiled local function on the running VM.

--- 

I have in this PR duplicated the Sequence Debugger suite across the two evaluator implementations. Let me know if we just want to include the new one, I will revise accordingly.

The fix to the newly included test (and the failing DistinctObject test in the Sequence Debugger suite) is in JetBrains/kotlin#4804


nikita-nazarov:
The code is fine. But we've got failing tests:

- org.jetbrains.kotlin.idea.debugger.test.IrKotlinEvaluateExpressionTestGenerated.SingleBreakpoint.Uncategorized#testDistinctSequence
- org.jetbrains.kotlin.idea.debugger.test.KotlinEvaluateExpressionWithIRFragmentCompilerTestGenerated.SingleBreakpoint.Uncategorized#testDistinctSequence
- org.jetbrains.kotlin.idea.debugger.test.KotlinEvaluateExpressionInMppTestGenerated.SingleBreakpoint.Uncategorized#testDistinctSequence

kandersen:
Rebased and resolved failures!

PR #2001, opened 2022-04-29
[EE] Force Evaluator tests to run in Unspecified JVM platform

kandersen:
The Kotlin expression evaluator runs _on top_ of the loaded project. Hence, it needs to reuse the resolution facade for both IDE compilation and during analysis of code fragment for compilation.

The evaluator is forced to use `unspecifiedJvmPlatform` to improve the user experience of the evaluator.

This also happened to align with all evalautor test suites until now.

The MPP tests used a different platform, hence got a different resolution facade, so the analysis was "fresh" and descriptors was not reused across project and fragment code leading to unbound references in psi2ir.

### DIFF
--- a/plugins/kotlin/jvm-debugger/test/test/org/jetbrains/kotlin/idea/debugger/test/AbstractKotlinEvaluateExpressionInMppTest.kt
+++ b/plugins/kotlin/jvm-debugger/test/test/org/jetbrains/kotlin/idea/debugger/test/AbstractKotlinEvaluateExpressionInMppTest.kt
@@ -27,8 +27,6 @@ import java.io.File
 abstract class AbstractKotlinEvaluateExpressionInMppTest : AbstractKotlinEvaluateExpressionTest() {
     override fun useIrBackend() = true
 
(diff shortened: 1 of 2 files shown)


kandersen:
@dsavvinov Hi Dmitry! I tag you in this because you introduced the `unspecifiedJvmPlatform` fix, and I would appreciate your input on this :)

Did I use the fix right, here? It seems correct to me that, since the IDEA debugger and Evaluator is evidently a JVM target, it makes sense to me to compile the IDE code being loaded on the debuggee JVM here using the `unspecifiedJvmPlatform`, too.

kandersen:
My concern is that I haven't inadvertently reduced the MPP test suite to be equivalent to the plain JVM_IR test suite by forcing the platform for the purpose of resolution :sweat_smile:

dsavvinov:
@kandersen yep, the change looks reasonable for me

kandersen:
@dsavvinov Awesome, thanks for taking a look!

nikita-nazarov:
https://github.com/JetBrains/intellij-community/commit/0fa12b4a0f979243b03b74adebeb004d8bbc5262

PR #2006, opened 2022-05-03
public and efficient SmartList#removeRange

Miha-x64:
(random holiday optimization)


cdracm:
I've pushed the patch, thank you! (except the "public" part, which I'm not sure we need for three reasons: 1) removeRange() is not in the List interface, whereas SmartList was designed to be dropin replacement as close to List as possible, 2) it would deviate from the (semi-)standard AbstractList, potentially creating unwanted dependency on private impl details and 3) removeRange() is already accessible without making it public by calling list.subList(...).clear() )

PR #2008, opened 2022-05-04
Fix resetting terminal when using "Split and Move" actions

chylex:
Fixes [IDEA-228753](https://youtrack.jetbrains.com/issue/IDEA-228753/Terminal-is-reset-when-splitting-editor), and should also fix [IDEA-252249](https://youtrack.jetbrains.com/issue/IDEA-252249/Splitting-SSH-session-produces-local-shell) although I haven't tested that since I don't know if I can enable SSH connections in the community edition repo.


chylex:
Hi, any updates on this?

PR #2009, opened 2022-05-04
[IDEA-208038] Fix missing runners for Compound configurations in Run/Debug popup

chylex:
Fixes [IDEA-208038](https://youtrack.jetbrains.com/issue/IDEA-208038).

Before:
obrazek

After: 
obrazek

I have verified that both Debug and Cover work for the compound configuration, but I'm not too familiar with this part of the codebase so please let me know:
1. If a non-empty `DataContext` is needed for `RunnerHelper.run`, and where to obtain it,
2. If I should replace all other instances of `ExecutionUtil.runConfiguration` and `ProgramRunner.getRunner` in this file, because I only replaced instances that I could find in the UI and verify that my fix worked for them.


chylex:
Hi, can someone please look at this? I have projects that exclusively use compound configurations, and this bug is completely preventing me from using this convenient menu.

PR #2010, opened 2022-05-05
IDEA-240111 Do not enforce re-running of tests in Gradle

kkocel:
Do not re-run tests when outputs did not change.

More background explanation as to why rerunning tests is the wrong approach: https://blog.gradle.org/stop-rerunning-tests

### DIFF
--- a/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/init/testFilterInit.gradle
+++ b/plugins/gradle/tooling-extension-impl/src/org/jetbrains/plugins/gradle/tooling/internal/init/testFilterInit.gradle
@@ -13,7 +13,6 @@ gradle.taskGraph.whenReady { taskGraph ->
   taskGraph.allTasks.each { Task task ->
     if (task instanceof Test || (abstractTestTaskClass != null && abstractTestTaskClass.isAssignableFrom(task.class))) {
       try {
-        task.outputs.upToDateWhen { false }
         String[] strings = ['*']
         if(ijTestIncludes.size() > 0 && ijTestIncludes != strings) {
           def filter = task.getFilter()


kkocel:
@ybedrov @DeveloperHacker Hi, wanted to ping on this

DeveloperHacker:
Hello, Chris!

I cannot accept this change because we need to restart test tasks from Idea. ( It is a feature :) )
For example if you start test from green arrow you expects that test will be excused in any case.
So we need separating two cases when we start test directly and when test tasks was executed as depend task of build tasks or other.

We can check Gradle command line in init script and find all test tasks in them. Further we should mark all these test tasks as non-up-to-date. Please see on `filterStartTasks` in `org.jetbrains.plugins.gradle.service.debugger.GradleJvmDebuggerBackend#initializationCode`

kkocel:
> Hello, Chris!
> 
> I cannot accept this change because we need to restart test tasks from Idea. ( It is a feature :) ) For example if you start test from green arrow you expects that test will be excused in any case. So we need separating two cases when we start test directly and when test tasks was executed as depend task of build tasks or other.

That's pretty simple - when running a test using green arrow just pass the` --rerun-tasks` flag. You don't need to overwrite the test tasks configuration.

DeveloperHacker:
Hi, Chris!

Could you describe your use case in YouTrack?
We cannot remove this code without implementation of basic user stories like run tests from gutter, from Gradle tool window, etc. TestLauncher support or auto injection of --rerun-tasks are a huge work which will spend a lot of time for implementation and testing.

ps. But TestLauncher support is present in our work queue for remaining year.

kkocel:
When running from Gradle tool window the test task(s) should be reported by Gradle as being UP-TO-DATE for the second execution, and should therefore be skipped.

aSemy:
If IntelliJ requires that tests are run, even though they're up to date, then other parts of the codebase should be updated to correctly use [Gradle task inputs/outputs](https://docs.gradle.org/current/userguide/lazy_configuration.html#working_with_task_dependencies_in_lazy_properties). 

It sounds like the use of `upToDateWhen { false }` here is hiding other problems, so it's not a 'feature', it's a bug. This PR should be merged and the the problems resolved.

aSemy:
I've checked out the project and from a brief experiment this PR doesn't seem to have changed anything - but then the project is so large (4GB!! Really??) and unwieldy (I encountered multiple failing tests on a fresh checkout) that I can't be sure.

nskvortsov:
@kkocel  thank you for the pull request. Unfortunately, the suggested solution is not applicable, as it breaks the individual test execution ('the green arrow case"). We are looking into using TestLauncher with the future Gradle versions.

PR #2011, opened 2022-05-05
[KTIJ-21662] FIR IDE: Implement missing Find Usages functionality

kandersen:
Initial venture in implementing Find Usages.


darthorimar (changes requested):

kandersen:
Moved to KOTLIN-CR-367

PR #2012, opened 2022-05-05
[IDEA-197202] Fix hiding undocked/unpinned tool windows when an inline editor is activated

chylex:
Fixes [IDEA-197202](https://youtrack.jetbrains.com/issue/IDEA-197202).

Example of this is in Git - Shelf when renaming a shelf:
obrazek

The inline editor's parent is `DefaultTreeCellEditor$EditorContainer` which extends `Container` and not `JComponent`, so the code thought that focus was outside the tool window and hid it.


AMPivovarov:
Thank you, merged.

PR #2013, opened 2022-05-06
support generating Android Icons with other project structures

jrlogsdon:
Allows other users to generate Android Icons when the Android project doesn't exist at intellij-community/android/artwork 

Other users can then generate Android Icons by:
- extending IconsClasses and overriding the homePath to their Android project path
- changing value of androidIcons in their build script to point to their Android project path


jrlogsdon:
@develar @DmPanov please review and let me know if there are any concerns

andrei-kuznetsov:
Merged. Thanks.

PR #2014, opened 2022-05-08
IDEA-293666 do not hardcode color of modified keybinds Tree entires to BLUE

qdbp:
This fixes https://youtrack.jetbrains.com/issue/IDEA-293666/Foreground-color-of-modified-keybinds-is-not-themable-probably-T


bulenkov:
Thank you for the patch. Fixed

PR #2015, opened 2022-05-08
KTIJ-21686 Optin quick-fix for annotation propose to add incorrect target

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21686


vladimirdolzhenko:

PR #2016, opened 2022-05-08
KTIJ-21637 "Remove annotation" intention duplicates for OptIns

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21637

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryOptInAnnotationInspection.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryOptInAnnotationInspection.kt
@@ -86,6 +86,10 @@ class UnnecessaryOptInAnnotationInspection : AbstractKotlinInspection() {
         val optInAliases = if (file is KtFile) KotlinPsiHeuristics.getImportAliases(file, OPT_IN_SHORT_NAMES) else emptySet()
 
         return annotationEntryVisitor { annotationEntry  ->
+            val annotationEntryArguments = annotationEntry.valueArguments.ifEmpty {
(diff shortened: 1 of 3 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryOptInAnnotationInspection.kt
```diff
@@ -113,7 +114,7 @@ class UnnecessaryOptInAnnotationInspection : AbstractKotlinInspection() {
             annotationEntry.getOwner()?.accept(OptInMarkerVisitor(), markerProcessor)
 
             val unusedMarkers = resolvedMarkers.filter { markerProcessor.isUnused(it.fqName) }
-            if (annotationEntry.valueArguments.size == unusedMarkers.size) {
+            if (annotationEntryArguments.isNotEmpty() && annotationEntryArguments.size == unusedMarkers.size) {
```
dimonchik0036, line 117: We shouldn't check an annotation entry without arguments at all. This fast check should be done at the begining of `annotationEntryVisitor`

dimonchik0036 (changes requested):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryOptInAnnotationInspection.kt
t-kameyama: > We shouldn't check an annotation entry without arguments at all. This fast check should be done at the begining of `annotationEntryVisitor`

Done

dimonchik0036 (approved):

PR #2020, opened 2022-05-16
[IDEA-294112] Add BranchRenameListener

i582:
Fixes [IDEA-294112](https://youtrack.jetbrains.com/issue/IDEA-294112).

This listener is needed to keep track of branch renames in plugins.


plugins/git4idea/src/git4idea/branch/GitRenameBranchOperation.java
```diff
@@ -80,6 +83,7 @@ protected void rollback() {
       myNotifier.notifySuccess(BRANCH_RENAME_ROLLBACK_SUCCESS,
                                GitBundle.message("git.rename.branch.rollback.successful"),
                                GitBundle.message("git.rename.branch.renamed.back.to", myCurrentName));
+      notifyBranchNameRollback(myNewName, myCurrentName);
```
AMPivovarov, line 86: It's better to choose and either call listener before notification or after it.
Should not change anything, but consistence is generally good.

AMPivovarov:
Could you clarify the use case for such listener?

Why 'rollback' case requires an explicit callback? How is it different from 'just a rename'?
Why won't you need information about affected VCS roots? Ex: project has two git repositories configured, both have 'some-feature' branch. User renames it to 'some-bug' in repository-2 only. Will everything work out fine for you?

plugins/git4idea/src/git4idea/branch/GitRenameBranchOperation.java
i582: Fixed.

i582:
Thanks for the great questions!
Indeed, a separate method for the case when the branch name has been rolled back is not needed.

In my case, the fact that a project can have multiple repositories and each one can be affected is not possible since we have one repository per project, however other projects can have multiple repositories in a project. I added the affected VCS root to the method signature,

i582:
Ny use case is to track the rename of a branch in order to rename it on a remote server for synchronization.

PR #2027, opened 2022-05-25
ProjectStructureMapping.groovy: close the JsonGenerator

SamB:
Otherwise, it won't know to flush the buffers, and we could end up missing the tail end of the file.


SamB:
Hmm. Sure would have been nice if GitHub had pointed out the conflict before I pushed the submit button ...

SamB:
Okay, looks like this was caught in commit 2f5482f2e95ad0e88b0324f654cdd083b0efefa0 -- just after the conversion to Kotlin.

PR #2033, opened 2022-05-30
Fix visibility icons in Package View

Yanatrill:
Currently "Tree Apperance -> Show Visibility Icons" doesn't affect Package View. No matter if it is _on_ or _off_, icons are not showing. This commit fixes it. Code and tests are similar to _ProjectViewPane_.


Yanatrill:
Looks like new UI has this working correctly

PR #2034, opened 2022-05-31
Fix Project leak through VcsToolWindowDnDTarget

npaige:
Commit 4f5f8e3c0f42d0d62b36bfdc724495c0e53be85a changed the
default parent disposable for SingleAlarm from null to
ApplicationManager.getApplication(). The runnable for the alarm in
VcsToolWindowDndTarget captures 'this', which contains a reference to
the project.

Instead, use the Content object as the parent disposable for the alarm.


AMPivovarov:
https://youtrack.jetbrains.com/issue/IDEA-295152

PR #2035, opened 2022-06-06
[IDEA-295460] ConditionalBreak: Handle else & finite loop

Fabrice-TIERCELIN:
Hi @BasLeijdekkers,

Now I'm extending an inspection: _Conditional breaks inside loop_:
**Given:**
```Java
while (i < 100) {
  if (i == 23) {
    break;
  } else {
    i++;
  }
}
```
**When:**
Applying *Conditional breaks inside loop* inspection...

**Then:**
```Java
while ((i < 100) && (i != 23)) {
  i++;
}
```

**Handle else branch**

* This feature extension only applies for loops with `break` at the beginning (i.e. not for `do`/`while` loops)
* If the `break` is in the *then*, so the *else* is pulled down from the `if` statement and the condition is negated
* If the `break` is in the *else*, so the *then* is pulled down from the `if` statement and the condition is just moved
* Loops with only one statement are handled
* We don't touch to the loops when a variable name conflict would be created like:

```Java
public class MyClass {
  private int variableWithSameName = 1;
  public void myMethod() {
    int i = 1;
    while (true) {
      if (i < 100) {
        break;
      } else {
        int variableWithSameName = -100;
      }
      i = i + variableWithSameName;
    }
  }
}
```

**Handle not-infinite loops**

* If the loop has already a condition, the existing condition and the condition of the `break` are put into a `&&` operation keeping the order
* If the loop has no condition (infinite loop), the condition of the `break` is put into the condition loop (just as before)
* For readability, we don't handle cases when the total number of logical operands would exceed 4 operands in the loop condition

**Inspection as a cleanup**

I have enabled the inspection as a cleanup. I have also renamed the inspection as it not only applies on infinite loops.

I have successfully tested the new features writing new unit tests.


plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/psiutils/BoolUtils.java
```diff
@@ -84,6 +87,31 @@ public static String getNegatedExpressionText(@Nullable PsiExpression condition,
...
+   * @param node The expression
+   * @return the number of logical operands in the expression
+   */
+  public static int getNbOperands(@Nullable PsiExpression condition) {
+    PsiExpression unparenthesizedExpression = PsiUtil.skipParenthesizedExprDown(condition);
+    if (!(unparenthesizedExpression instanceof PsiPolyadicExpression)) {
+      return 1;
+    }
+    PsiPolyadicExpression infixExpression= (PsiPolyadicExpression) unparenthesizedExpression;
+    if (!Arrays.asList(ANDAND, OROR).contains(infixExpression.getOperationTokenType())
```
sirgl, line 102: Let't avoid unnecessary allocations here, just compare the type with equals

java/java-impl/src/META-INF/JavaPlugin.xml
```diff
@@ -1373,7 +1373,7 @@
     <localInspection language="JAVA" shortName="ConditionalBreakInInfiniteLoop" enabledByDefault="true" level="WARNING"
                      groupPath="Java" groupBundle="messages.InspectionsBundle" groupKey="group.names.control.flow.issues"
                      bundle="messages.JavaBundle" key="inspection.conditional.break.in.infinite.loop.description"
-                     implementationClass="com.intellij.codeInspection.ConditionalBreakInInfiniteLoopInspection"/>
+                     implementationClass="com.intellij.codeInspection.ConditionalBreakInInfiniteLoopInspection" cleanupTool="true"/>
```
sirgl, line 1376: I would avoid making inspection cleanup, some people may prefer other code style, especially when it puts several expression in a single line of code. In such a case it will be more complicated to debug it.

sirgl (changes requested):
Hi @Fabrice-TIERCELIN 
Thank you for your contribution! In general looks nice, but I wouldn't make it cleanup inspection.

It is ok as it is, but also I may recommend to add an option for this functionality, because someone may like it in original variant, without fusing conditions of the loop and the if statement as it complicates debugging.

Fabrice-TIERCELIN:
OK, I will do it

Fabrice-TIERCELIN:
Hi @sirgl,

As we have turned off the cleanup, the option checkbox becomes useless. If the single line of code disturbs the debugging, the user will just not apply the quickfix.

Do we keep the filename `ConditionalBreakInInfiniteLoopInspection` as it also process finite loops?

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/psiutils/BoolUtils.java
```diff
@@ -84,6 +87,31 @@ public static String getNegatedExpressionText(@Nullable PsiExpression condition,
     return getNegatedExpressionText(condition, ParenthesesUtils.NUM_PRECEDENCES, tracker);
   }
 
+  /**
+   * Returns the number of logical operands in the expression.
+   *
+   * @param node The expression
+   * @return the number of logical operands in the expression
+   */
+  public static int getNbOperands(@Nullable PsiExpression condition) {
```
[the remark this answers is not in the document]
Fabrice-TIERCELIN: Done

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/psiutils/BoolUtils.java
Fabrice-TIERCELIN: Done

java/java-impl/src/META-INF/JavaPlugin.xml
Fabrice-TIERCELIN: Done

sirgl:
Thank you! Sorry for the delay. I will merge it today.

sirgl:
> Do we keep the filename ConditionalBreakInInfiniteLoopInspection as it also process finite loops?

Yes, we should keep the name, the suppression relies that it stays the same, otherwise it would require much more work to keep the suppression.

PR #2036, opened 2022-06-07
[kotlin] Convert parameter to receiver: enclose definitely not-null types in parentheses

t-kameyama:
[KTIJ-20849](https://youtrack.jetbrains.com/issue/KTIJ-20849) Definitely not-null types parameter converts to receiver type incorrectly


abelkov:
Looks like this should be fixed in the compiler, see https://github.com/JetBrains/kotlin/blob/e36091b926185ad8ddacdc13c615fdfedcdc1d6d/core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRendererImpl.kt#L368. I'll try to do it.

Also, your current fix doesn't help in case of https://youtrack.jetbrains.com/issue/KTIJ-20849/Definitely-not-null-types-parameter-converts-to-receiver-type-incorrectly#focus=Comments-27-5718225.0-0

abelkov (approved):

PR #2037, opened 2022-06-09
Add language level for Python 3.12

KotlinIsland:
I updated the python/helpers/tools/stdlib_packages.txt and versions.xml as well, I think it's correct.

YouTrack: [PY-54601](https://youtrack.jetbrains.com/issue/PY-54601)


python/python-psi-api/src/com/jetbrains/python/psi/FutureFeature.java
```diff
@@ -28,7 +28,7 @@ public enum FutureFeature {
   ABSOLUTE_IMPORT("absolute_import", LanguageLevel.PYTHON25, LanguageLevel.PYTHON30),
   PRINT_FUNCTION("print_function", LanguageLevel.PYTHON26, LanguageLevel.PYTHON30),
   UNICODE_LITERALS("unicode_literals", LanguageLevel.PYTHON26, LanguageLevel.PYTHON30),
-  ANNOTATIONS("annotations", LanguageLevel.PYTHON37, LanguageLevel.PYTHON311)
+  ANNOTATIONS("annotations", LanguageLevel.PYTHON37, LanguageLevel.PYTHON312)
```
KotlinIsland, line 31: `__future__.annotations` is turned on in the 3.11 betas so far, but it's actually been delayed again, so shouldn't be in 3.11 release.

KotlinIsland: - See this https://github.com/python/cpython/pull/93628

lada-gagina:
@KotlinIsland Thank you for your PR and sorry for the delay. I've merged it manually in 45dd0bf

PR #2039, opened 2022-06-09
[KTIJ-21907] Repro included in test suite

kandersen:
This is a test demonstrating a fix in the compiler at jetbrains/kotlin#4841. It resolves [KTIJ-21907](https://youtrack.jetbrains.com/issue/KTIJ-21907).

Files with _unrelated_ inline function dependencies (not used in fragment, not used in context of breakpoint) cause errors in codegen
because the sources are not included in the compilation unit.

Use the generation filter to avoid generating code for this.

^KTIJ-21907 open


kandersen:
kotlin#4841 has landed.

nikita-nazarov:
Thanks for the contribution! Merged into `kt-master`
https://github.com/JetBrains/intellij-community/commit/5e76316432bbd6705f6d1683425efcff9ef09cd5

PR #2041, opened 2022-06-13
Add new Git update methods for convenient fetching

stachenov:
It is often desirable to fetch upstream commits first and then decide what is best: merge, rebase, or maybe even postpone updating and discuss your changes with other developers first.

However, at the moment IDEA doesn’t provide an easy way to just fetch. There is the Fetch action, but it’s not bound to a keystroke by default, which is inconvenient. It’s possible to bind it manually, but it’s not without drawbacks: it won’t work on someone else’s PC, you need a spare keystroke for that, and it won’t be consistent with other Git GUIs, like Git Extensions, for example, which just provides a list of possible update actions in a single drop-down menu, including Merge, Rebase, Fetch, Fetch All and Fetch and Prune.

So what I did here is add a couple of new “fake” Git update methods: Fetch Default and Fetch All. Seems to be working nicely to me, except for one usability quirk: after fetching is done, an “All files are up-to-date” notification pops up, which can be a bit confusing, as something like “Fetched N new commits from ...” would be more useful. However, this isn’t possible to do without reworking the VCS Update API considerably, and I’m not sure if such a minor annoyance is enough to justify that.

### DIFF
--- a/plugins/git4idea/resources/messages/GitBundle.properties
+++ b/plugins/git4idea/resources/messages/GitBundle.properties
@@ -675,6 +675,8 @@ settings.git.update.method.merge=Merge
 settings.git.update.method.branch.default=Branch Default
 settings.git.update.method.fetch.default.description=Don't update, just fetch from the default remote
 settings.git.update.method.fetch.default=Fetch default
+settings.git.update.method.fetch.all.description=Don't update, just fetch from all remotes
+settings.git.update.method.fetch.all=Fetch all
 settings.enable.staging.area=Enable staging area
 settings.enable.staging.area.comment=This will disable changelists support and delete all changelists in the project. Only for non-modal commit interface.
 
--- a/plugins/git4idea/resources/messages/GitBundle.properties
+++ b/plugins/git4idea/resources/messages/GitBundle.properties
@@ -673,6 +673,8 @@ settings.git.update.method.rebase=Rebase
 settings.git.update.method.merge.description=Merge incoming changes into the current branch
 settings.git.update.method.merge=Merge
 settings.git.update.method.branch.default=Branch Default
+settings.git.update.method.fetch.default.description=Don't update, just fetch from the default remote
+settings.git.update.method.fetch.default=Fetch default
 settings.enable.staging.area=Enable staging area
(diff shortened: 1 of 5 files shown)


dmitriysmirnovjb:
Hello, thank you for your contribution! 
Could you please elaborate on why you want to use the Update project dialog here?
It seems assigning a default shortcut to the Fetch action is an easier and better way to achieve what you want?

stachenov:
Several reasons, although all of them are debatable, and I can’t say with confidence that this is the best way to do it.

1. As I’ve mentioned, there is no default binding for Fetch. Manual binding is possible, but not without the above-mentioned drawbacks, such as it won’t work on someone else’s PC, needs to be set up on every computer, may conflict with new standard bindings in future versions of IDEA.
2. I can’t come up with a convenient keystroke for Fetch. A lot of them are already used for something else. A lame excuse, of course, but something that’s easy to remember would be desirable.
3. If I want to never automatically merge or rebase, which is quite often the case in my experience, I can just set up the default update method to one of the fetch methods and never worry about accidentally pressing Ctrl+T and screwing up my current branch. This is the way Git Extensions work, by the way, which I find very convenient in this regard.
4. Fetching and updating are closely related tasks, so it makes some sense to put them in one place and have a single shortcut.

While these reasons are admittedly weak, I also don’t see any reasons *not* to put these methods there. It doesn’t add any new actions, doesn’t occupy new keystrokes, doesn’t lead to any confusing behavior by default. Whenever a developer that is used to dealing with Subversion, for example, migrates to Git, they just press the same keystroke and see all update options available to them.

dmitriysmirnovjb:
Thank you for the contribution and effort here, and sorry that the pull request got stalled. We appreciate the idea of simplifying the fetch workflow.

However, after internal discussions, we’ve decided not to integrate these changes.

Fetch action is now better exposed in Branches popup, as well as Update.
If your default way is the Fetch, one can re-assigne the shortcut from Update project to Fetch.
This avoids introducing new menu items that require different handling, and aligns better with existing workflows.

As a result, I’m closing this pull request unmerged. Please feel free to add follow-up suggestions if you'd like to explore alternative approaches in the future.

PR #2042, opened 2022-06-14
[PY-52971] Fix NPE when debugging SymPy user-defined Function

eagleoflqj:
The bug makes lots of SymPy code un-debuggable.
This is the same patch I uploaded https://youtrack.jetbrains.com/issue/PY-52971/NPE-when-debugging-sympy
Local test result:
Screenshot from 2022-06-13 20-59-58

Please take a look. Thank you!

### DIFF
--- a/python/src/com/jetbrains/python/debugger/PyStackFrame.java
+++ b/python/src/com/jetbrains/python/debugger/PyStackFrame.java
@@ -167,7 +167,7 @@ public class PyStackFrame extends XStackFrame {
           else if (pyValue.isIPythonHidden()) {
             groupIndex = IPYTHON_VALUES_IND;
           }
-          else if (HIDE_TYPES.contains(pyValue.getType()) || HIDE_MODULES.contains(pyValue.getTypeQualifier())) {
+          else if (isSpecialType(pyValue)) {
             groupIndex = SPECIAL_TYPES_IND;
           }
           if (groupIndex > -1) {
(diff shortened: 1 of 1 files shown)


rendner:
Hi @eagleoflqj,

since `myType` and `myTypeQualifier` where both marked as `@Nullable`, I guess it would be better to fix the wrong check here:
https://github.com/JetBrains/intellij-community/blob/b9e396a52f60f9a4bdb46625beee5822d23495be/python/src/com/jetbrains/python/debugger/PyStackFrame.java#L170
So that check could be like this:
```else if ((pyValue.getType() != null && HIDE_TYPES.contains(pyValue.getType())) || (pyValue.getTypeQualifier() != null && HIDE_MODULES.contains(pyValue.getTypeQualifier()))) {```

Or have you already tried this?

**Comment by eagleoflqj** [comment]
That should also work, but `getQualifiedType` is unnecessarily called twice if we don’t want to break current if-else structure.

**Comment by rendner** [comment]
> That should also work, but `getTypeQualifier ` is unnecessarily called twice if we don’t want to break current if-else structure.

Then you can move the check into a separate method and call that in the `else if` branch:
```
else if (isSpecialType(pyValue)) {
....

private static boolean isSpecialType(PyDebugValue pyValue) {
    final String type = pyValue.getType();
    final String typeQualifier = pyValue.getTypeQualifier();
    return (type != null && HIDE_TYPES.contains(type)) || (typeQualifier != null && HIDE_MODULES.contains(typeQualifier));
}
```

eagleoflqj:
Thanks @rendner ! I confirm locally your suggestion works. And it's in theory safer than cb3829bdb3a7.

eagleoflqj:
Closing. Will check whether the bug is fixed and make a new PR if not.

PR #2044, opened 2022-06-15
Ignore non-jar files when creating resource loaders

peterphan:
There can be cases where a non-jar file is added to the classpath / `java.class.path` system property.
See https://github.com/JetBrains/gradle-intellij-plugin/issues/1009
This diff ignores non-jar files when loading the classpath, similar to
https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/commit/9dcc6e22bcec9264cb5f06d5299e9df5fd8446fc


develar:
Is it still actual?

PR #2047, opened 2022-06-17
Add ByteBuffer's get*() methods to NotPropertiesService

stachenov:
I’m using `ByteBuffer` a lot and it’s a bit annoying when IDEA asks me to use property access syntax instead, which makes no sense, so I add ByteBuffer's methods to the list of exceptions. Figured it’s not a bad idea to include it in the standard list of exceptions as `ByteBuffer` is a standard library class.

### DIFF
--- a/plugins/kotlin/base/code-insight/src/org/jetbrains/kotlin/idea/core/NotPropertiesService.kt
+++ b/plugins/kotlin/base/code-insight/src/org/jetbrains/kotlin/idea/core/NotPropertiesService.kt
@@ -25,6 +25,9 @@ interface NotPropertiesService {
                     add("java.util.concurrent.atomic.$atomicClass.$atomicMethod")
                 }
             }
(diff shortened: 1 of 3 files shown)


vladimirdolzhenko:
Sorry, for the delay.

Could you please add tests for the change ?

vladimirdolzhenko:
it could be smth similar to https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/idea/tests/testData/intentions/usePropertyAccessSyntax/suppressedByNotPropertyList.kt

do not forget to regenerate tests after adding new test data files.

stachenov:
Added one test, similar to how the other tests were done. However, I haven’t exactly been able to figure them out, as they seem to be all concentrated around atomic types despite test names. For example, `suppressedByAtomicIntegerList` tests `AtomicInteger.getPlain()` while `suppressedByNotPropertyList` tests `AtomicInteger.getAndIncrement()`.

But it works and makes sense to me at least.

vladimirdolzhenko:
Sorry for a long delay, LGTM

Thank you for the contribution. Will be merged soon

PR #2049, opened 2022-06-19
Kotlin auto-completion: don't show aliased Java types

stachenov:
Types like `java.lang.StringBuilder` or `java.lang.AssertionError` often pop up in Kotlin auto completion, causing unnecessary ugly imports. Here, I’ve tried to fix it by disabling type aliases on the first auto completion attempt.

It could use some improvements, though.

Putting the list of aliased types in the `fqNameUtil.kt` looks inelegant. Ideally it would’ve been better to delegate it to the compiler library, but I haven’t found anything like that there.

On the second attempt those types still pop up in pretty much random order. It would’ve been better to give them a lower priority.

But it works, and to me it’s already a great help, not having to figure out which type is Kotlin and which is Java every time I see an auto-completion list pop up.

### DIFF
--- a/plugins/kotlin/completion/src/org/jetbrains/kotlin/idea/completion/AllClassesCompletion.kt
+++ b/plugins/kotlin/completion/src/org/jetbrains/kotlin/idea/completion/AllClassesCompletion.kt
@@ -14,6 +14,7 @@ import org.jetbrains.kotlin.descriptors.ClassifierDescriptorWithTypeParameters
 import org.jetbrains.kotlin.idea.base.facet.platform.platform
 import org.jetbrains.kotlin.idea.base.utils.fqname.getKotlinFqName
 import org.jetbrains.kotlin.idea.base.utils.fqname.isJavaClassNotToBeUsedInKotlin
+import org.jetbrains.kotlin.idea.base.utils.fqname.isJavaClassWithKotlinTypeAlias
 import org.jetbrains.kotlin.idea.caches.KotlinShortNamesCache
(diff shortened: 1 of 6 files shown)


vladimirdolzhenko:
Could you pls create a KTIJ ticket to describe problem.

stachenov:
Done:
https://youtrack.jetbrains.com/issue/KTIJ-22052/Auto-completion-shows-aliased-Java-types

vladimirdolzhenko:
I am so sorry for the huge time gap, 
Kotlin plugin already switched to K2 mode - is it possible to make this work for K2 mode ?

stachenov:
I think yes, but we'll have to redo the whole thing from scratch. And so it'll be easier to do it internally rather than jump through GitHub integration hoops. I'll create a new MR in Space.

FredyH:
@stachenov I think that this functionality is already more or less implemented for K2.
We have a Weigher that automatically prioritizes classes that are already imported (the Kotlin classes are implicitly imported) from ones that are not (like in the Java case).
At least in some quick tests I was unable to get a Java class to popup before the Kotlin class.

Note that completion order might be affected by ML completion sorting which can be turned off in the settings (and will actually not be enabled by default starting with 2025.3).

image

stachenov:
Sorting is good (and I haven't noticed Java classes getting in my way lately), but this PR is about hiding them altogether on the first completion invocation. Now if you type `StringBuilder`, you'll see this:
image
The second line is just noise. I'll see if I can get rid of it.

stachenov:
Sent an internal MR. I think this can be closed now.

PR #2050, opened 2022-06-19
[WIP] Pull intellij-diff-plugin into the tree

SamB:
This is an attempt to merge https://github.com/ThomasR/intellij-diff-plugin into the platform.

The primary reason for this is that I couldn't figure out how to associate a language with the PATCH filetype without adding a `language` attribute to the `fileType` element that declares the filetype and changing the FileType class, which seems ... impractical for an independent plugin to do.  If I had a way to accomplish this by just modifying the plugin, I probably would.

(The biggest reason to want this on the actual PATCH filetype instead of just adding another is that the PATCH filetype has that Apply Patch action.)

However, it would also be nice to customize the colors for diff highlighting in various themes (but this PR doesn't actually do that).

N.B. Unfortunately, I didn't write the plugin, so I can't actually sign your CLA for that code, though I will note that the code is already Apache 2.0 licensed.

P.S. I would have rebased this, but that merge is ... too hard to rebase. I've already screwed up many times while trying to rebase it once (I can't even remember if I succeeded in the end); I really don't wanna screw it up again. Fortunately, it only touches four existing files and seems to merge cleanly.


iasemenov:
Hi. I can see that there's an active fork of the plugin at https://github.com/ThexXTURBOXx/intellij-diff-plugin
Is there a reason you want us to pull it into the repo?

PR #2060, opened 2022-06-21
KTIJ-21964 False positive "unused symbol" for items of nested enum with `valueOf` call

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-21964


abelkov:
There is a failing test now `testData/inspectionsLocal/removeRedundantQualifierName/notApplicableEnumInEnum4.kt`

t-kameyama:
Fixed

PR #2065, opened 2022-06-23
KTIJ-22026 Add quick-fix to delete forbidden annotation retention SOURCE in opt-ins declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22026

```kotlin
@RequiresOptIn
@Retention(AnnotationRetention.SOURCE)
annotation class SomeAnnotation
```
スクリーンショット 2022-06-23 23 33 16

↓

スクリーンショット 2022-06-23 23 34 42


abelkov:
It doesn't make sense to keep `@Retention` annotation without arguments, in such case we can delete it too. Also, empty `@Target` can be deleted too.

t-kameyama:
https://user-images.githubusercontent.com/1121855/175463392-f291218c-9442-4a4a-8698-c503da0fcf28.mov

PR #2066, opened 2022-06-24
IDEA-296632: Add null guard to Groovy generated Object.equals()

tmoschou:


knisht:
The patch has been applied (b95fdfb6072a9bdf519349b26970ca7f68608b32).
cc @ybedrov the PR may be closed.

PR #2067, opened 2022-06-24
Always add comment in EOCD, even for a ZIP64 file

rsauciuc:
Python doesn't handle variable length comments (see https://github.com/python/cpython/blob/3.10/Lib/zipfile.py#L235). A simple way to test this is with a simple script:

```
import zipfile

with zipfile.ZipFile("c:/Users/me/AppData/Local/JetBrains/Toolbox/apps/IDEA-C/ch-0/222.3153.4/lib/app.jar") as archive:
    archive.printdir()
```
... which will result in:
```
Traceback (most recent call last):
  File "c:\src\ziptestfile.py", line 3, in <module>
    with zipfile.ZipFile("c:/Users/me/AppData/Local/JetBrains/Toolbox/apps/IDEA-C/ch-0/222.3153.4/lib/app.jar") as archive:
  File "C:\Users\me\AppData\Local\Programs\Python\Python310\lib\zipfile.py", line 1267, in __init__
    self._RealGetContents()
  File "C:\Users\me\AppData\Local\Programs\Python\Python310\lib\zipfile.py", line 1351, in _RealGetContents
    fp.seek(self.start_dir, 0)
OSError: [Errno 22] Invalid argument
```

Android Studio is built with Bazel, and certain packaging actions are implemented with Python scripts, hence the need for a working format.


gharrma:
cc @develar

develar:
Do I understand correctly that you suggest writing custom data as trailing data to the end of ZIP file? It was initially implemented in this way, but Apple `ditto` doesn't like such ZIPs. Code signinging on macOS filed. Moreover, zip with trailing data can be suspicious.

rsauciuc:
I was suggesting we write this data as a comment in EOCD, not in EOCD64. Python's zip doesn't fully parse EOCD64. This would only affect app.jar, the others would be unchanged.

PR #2069, opened 2022-06-29
Increase idle Gradle timeout in tests to 100 seconds

dahlstrom-g:
The default 10 second value results in the daemon being shut down before
the IDE completes indexing after several initial project syncs.

@vladsoroka


dahlstrom-g:
@HackerMadCat?

dahlstrom-g:
@HackerMadCat?

HackerMadCat:
Sorry. I cannot understand how it can help you. This duration affect only idea in unit test mode. Could you describe how you use that? As I know availability of Gradle daemon doesn't affect Idea indices.

Also, we have a rule that we cannot replace one magic constant by other. In this case we should add some event on which we will shutdown Gradle daemon instead unconditional shutdown.

dahlstrom-g:
Perhaps @yuriy-solodkyy can elaborate further; from what he's written I gather this is intended to make tests more reliable.

> […] we have a rule that we cannot replace one magic constant by other. In this case we should add some event on which we will shutdown Gradle daemon instead unconditional shutdown.

This is outside my area, so I don't know how that would look in detail, but for what it's worth it sounds good to me!

HackerMadCat:
@yuriy-solodkyy I will greatly appreciate if you create issue in you track and describe your problem. I cannot accept this pull request without real problem.
In this case, I would like to see the test which fails due to broken indexes caused by a short delay.

yuriy-solodkyy:
Gradle daemon shuts down while a project is set up (usually because of some
incremental indexing taking longer) and it slows down a test running after
the current one.

I think we agreed some time ago to make it configurable but I haven't got
round to preparing a pull request. I didn't plan to upstream this change.

nskvortsov:
feel free to reopen if this problem still affects you.

PR #2071, opened 2022-06-29
Replace link with button in "Invalidate Caches..." dialog

a-narsudinov:
Current implementation of the "Just restart" element as a link instead of button looks awkward to me, moreover according to the [JB UI guideline](https://jetbrains.design/intellij/controls/button/#when-to-use) element that perform immediate action should be a button, not a link.

This patch replaces the link element with a button.

Dialog menu before the patch
image

And after (checkboxes are not related to the PR changes, I have no idea why they have disappeared)
image


a-narsudinov:
Hi there! More than a year has passed and this menu is still the same and feels broken. 

@jonnyzzz @bulenkov could you please have a look?

batya239:
Merged, thanks for your contribution!

a-narsudinov:
Hmm, looks kind of strange to me, because it seems like this PR is just closed. But anyway, if it was fixed after more than 3 years, I'm glad :)

upd: I've found the [commit](https://github.com/JetBrains/intellij-community/commit/37a42969d291ace8d3f7a5e5ee3f9f75e448b74c), cool!

PR #2073, opened 2022-06-30
Add null check to JdScope equals implementation

izzyreal:
As per https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)

Reason for this PR: We actually experienced NPEs in some conditions (albeit in tests, but still...).

Credits to @unkarjedy for suggesting this fix, rather than only fixing it in https://github.com/JetBrains/intellij-scala/commit/545b9b7d90c9563f3f186ac99cff479e021ada30


izzyreal:
Moved the PR to our internal VCS.

PR #2074, opened 2022-07-01
KTIJ-22114 "Complete current statement" inside when branch should complete statement with arrow and braces

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22114


abelkov:
From such a position (note the space) the resulting caret is not located inside braces. Is that easily fixable?

```
fun foo(c: Char) {
    when (c) {
        '.' <caret>
    }
}
```
->
```
fun foo(c: Char) {
    when (c) {
        '.' -> {
        }
        <caret>
    }
}
```

t-kameyama:
I've added a commit, please review again.

abelkov (approved):

PR #2075, opened 2022-07-01
KTIJ-19509 IDE inspection to replace readLine() usages to readln() or readlnOrNull()

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19509

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceReadLineWithReadlnInspection.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceReadLineWithReadlnInspection.kt
@@ -5,21 +5,22 @@ import com.intellij.codeInspection.LocalQuickFix
 import com.intellij.codeInspection.ProblemDescriptor
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
-import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.config.ApiVersion
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.base.projectStructure.languageVersionSettings
 import org.jetbrains.kotlin.idea.inspections.collections.isCalling
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.FqName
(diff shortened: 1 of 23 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceReadLineWithReadlnInspection.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.jetbrains.kotlin.idea.inspections.collections.isCalling
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getQualifiedExpressionForSelectorOrThis
+
+class ReplaceReadLineWithReadlnInspection : AbstractKotlinInspection() {
+    companion object {
+        private val readLineFqName = FqName("kotlin.io.readLine")
+        private val readlnFqName = FqName("kotlin.io.readln")
```
abelkov, line 21: Looks like `readlnFqName` and `readlnOrNullFqName` are used only as strings, so constructing an `FqName` is not necessary.

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceReadLineWithReadlnInspection.kt
```diff
@@ -0,0 +1,54 @@
...
+        )
+    })
+
+    private class ReplaceFix(private val functionName: String) : LocalQuickFix {
+        override fun getName() = KotlinBundle.message("replace.with.0", functionName)
+
+        override fun getFamilyName() = name
+
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            descriptor.psiElement.replace(KtPsiFactory(project).createExpression("$functionName()"))
```
abelkov, line 51: What if there is a user-defined function with the same name in scope?

```
fun main() {
    val x = readLine()!!
}

fun readln() {}
```

I think it's better to insert a fully qualified name and shorten it if possible.

abelkov (changes requested):

abelkov (approved):

PR #2076, opened 2022-07-01
[gradle] Try to substitute unresolved dependencies from idea project

fp7:
When gradle is not able to resolve a dependency it may still be imported
into the current project. When dependency substitution is enabled the library
will be replaced by the module.

Notes:
- Dependencies of the non gradle module are still not resolved and are
  not usable from the gradle project as transitive dependencies.
- When the build is delegated to gradle, it will not be able to find the
  dependency because it is only resolved inside the ide.

Related to: IDEA-134885


HackerMadCat:
Thank you for fix!
I'll try to review it, but I need some time.
I tagged the issue so I don't miss this request again.

fp7:
No worries, this is not at all a pressing issue for me. And it is also not a complete fix. Just a tiny step.

nskvortsov:
The related issue [IDEA-134885](https://youtrack.jetbrains.com/issue/IDEA-134885) is fixed with 

dae918df11e4f848464fcf03c74e6ffb3cd8e659
a0b707fc23ec50a88e498b903e8bba77296c57ff
eb9289f9af732fb748b8bd4c83f0bcb481c442e9

and other commits with the issue id in the message.

please, check if it covers your case.

PR #2077, opened 2022-07-01
[KTIJ-21854] Reorder composite analysis results in fragment compiler

kandersen:
Compiler side changes in https://github.com/JetBrains/kotlin/pull/4880

---

The IR Fragment Compiler reuses the batch IR compiler infrastructure for compilation.This requires full analysis of the fragment dependencies included for compilation. 

See the comment in KotlinEvaluatorBuilder for a detailed breakdown, but, in short, the order of analysis result composition is significant for purpose of psi2ir: code fragments trigger partial analysis of the function containing the fragment, and layering this _on top_ of the complete analysis of the dependent files causes the Composite binding context to present an inconsistent view _of the dependent file_ where, in this specific case, a fresh descriptor was created for the local variable `x` and included in the symbol table while generating IR for `test`, but the reference to `x` in `bar` was pointing to a different local variable descriptor for `x`, since the local function was not reanalyzed by the analysis for the fragment context.

(Incidentally, this also explains why upgrading to `FULL` analysis for the fragment context resolves the issues: now we again have a complete and consistent view of the function `test`, including the body of the local `bar`, and all descriptors agree.)

Resolves [KTIJ-21854](https://youtrack.jetbrains.com/issue/KTIJ-21854/MPP-DebuggerIR-Evaluate-expression-fails-for-local-variable-on-first-run) and [KTIJ-21877](https://youtrack.jetbrains.com/issue/KTIJ-21877/IR-debugger-Errors-on-attempt-to-evaluate-expression)

### DIFF
--- a/plugins/kotlin/jvm-debugger/evaluation/src/org/jetbrains/kotlin/idea/debugger/evaluate/KotlinEvaluatorBuilder.kt
+++ b/plugins/kotlin/jvm-debugger/evaluation/src/org/jetbrains/kotlin/idea/debugger/evaluate/KotlinEvaluatorBuilder.kt
@@ -236,10 +236,38 @@ class KotlinEvaluator(val codeFragment: KtCodeFragment, private val sourcePositi
                         analysisResult.bindingContext
                     ).second
                 } else {
+                    // The IR Evaluator is sensitive to the analysis order of files in fragment compilation:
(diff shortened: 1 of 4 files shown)


kandersen:
CC @nikita-nazarov @dsavvinov @madsager @udalov

PR #2079, opened 2022-07-03
[IDEA-297272] Handle cleanup conflict: System.out.println("a" + "b");

Fabrice-TIERCELIN:
Hi @sirgl,

1\. Write the following statement in a Java class:

```Java
System.out.println("a" + "b"); // ab
```

2\. Right-click
3\. Go to *Analyze...* → *Code cleanup*
4\. Go to *...* → *Java* → *Performance* → *Single character concatenation*
5\. Enable *Single character concatenation*
6\. Click on *OK*

```Java
System.out.println("a" + 'b'); // ab
```

***What happens instead?***

```Java
System.out.println('a' + 'b'); // 195
```

The problem is that the inspection has identified two expressions to fix but each case is relied on the other. If one case is fixed, the other should not be fixed anymore. In other words, the context needs to be re-evaluated before applying a second fix. That's what this PR does.

The PR also adds a unit test class. All the tests successfully pass.


BasLeijdekkers:
Merged manually with some minor changes here: https://github.com/JetBrains/intellij-community/commit/0dc8bca976da2a6e3d08a33f837f9baee1a62c93

PR #2083, opened 2022-07-08
KTIJ-22186 Quick fix to convert instance of check for Array to isArrayOf intrinsic call

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22186

スクリーンショット 2022-07-08 14 19 56

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeToStarProjectionFix.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeToStarProjectionFix.kt
@@ -6,6 +6,7 @@ import com.intellij.codeInsight.intention.IntentionAction
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import org.jetbrains.kotlin.diagnostics.Diagnostic
+import org.jetbrains.kotlin.diagnostics.Errors
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
@@ -18,6 +19,7 @@ import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
 import org.jetbrains.kotlin.resolve.calls.model.ArgumentMatch
 import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+import org.jetbrains.kotlin.types.checker.SimpleClassicTypeSystemContext.isArrayOrNullableArray
 import org.jetbrains.kotlin.types.typeUtil.isTypeParameter
 
 class ChangeToStarProjectionFix(element: KtTypeElement) : KotlinQuickFixAction<KtTypeElement>(element) {
@@ -43,6 +45,7 @@ class ChangeToStarProjectionFix(element: KtTypeElement) : KotlinQuickFixAction<K
 
     companion object : KotlinSingleIntentionActionFactory() {
         override fun createAction(diagnostic: Diagnostic): IntentionAction? {
+            if (diagnostic.isArrayInstanceCheck()) return null
             val binaryExpr = diagnostic.psiElement.getNonStrictParentOfType<KtBinaryExpressionWithTypeRHS>()
             val typeReference = binaryExpr?.right ?: diagnostic.psiElement.getNonStrictParentOfType<KtTypeReference>()
             val typeElement = typeReference?.typeElement ?: return null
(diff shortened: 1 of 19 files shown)


abelkov (changes requested):
I think `ChangeToStarProjectionFix` should not be suggested at all for arrays in these cases. Because now we have two quickfixes: one correct and one incorrect. We should just suggest one correct quickfix.

Also, for code like this `ChangeToStarProjectionFix` just doesn't make sense at all:

```
fun foo(a: Array<*>) {
    if (a is Array<String>) {
    }
}
```

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeToStarProjectionFix.kt
```diff
@@ -82,6 +85,9 @@ class ChangeToStarProjectionFix(element: KtTypeElement) : KotlinQuickFixAction<K
             return null
         }
 
+        private fun Diagnostic.isArrayInstanceCheck(): Boolean =
+            factory == Errors.CANNOT_CHECK_FOR_ERASED && Errors.CANNOT_CHECK_FOR_ERASED.cast(this).a.isArrayOrNullableArray()
```
abelkov, line 89: Because the new fix is available only on jvm platform, we should leave this fix on non-jvm platforms, so at least some fix is available on every platform.

I guess it's good to leave a comment here to explain that we are depending on availability of `ConvertToIsArrayOfCallFix`.

plugins/kotlin/idea/tests/testData/quickfix/addStarProjections/checkType/array.kt
```diff
@@ -0,0 +1,9 @@
+// "Change type arguments to <*, *>" "false"
```
abelkov, line 1: `Change type arguments to <*>`

plugins/kotlin/idea/tests/testData/quickfix/convertToIsArrayOfCall/nullableArrayLhs.kt.after
```diff
@@ -0,0 +1,6 @@
+// "Convert to 'isArrayOf' call" "true"
+// WITH_STDLIB
+fun test(a: Array<*>?) {
+    if (a is Array<*> && a.isArrayOf<String>()) {
```
abelkov, line 4: Maybe this is better?

```
fun test(a: Array<*>?) {
    if (a?.isArrayOf<String>() == true) {
    }
}
```

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToIsArrayOfCallFix.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val lhsIsNonNullableArray = !lhsType.isNullable() && lhsType.isArrayOrNullableArray()
+
+    private val arrayArgumentTypeText = IdeDescriptorRenderers.SOURCE_CODE_TYPES.renderType(arrayArgumentType)
+
+    override fun getFamilyName() = KotlinBundle.message("fix.convert.to.is.array.of.call")
+
+    override fun getText() = familyName
+
+    public override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val isExpression = element ?: return
```
abelkov, line 40: Please support `!is` case:

```
fun foo(a: Any) {
    if (a !is Array<String>) {
    }
}
```

abelkov (changes requested):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeToStarProjectionFix.kt
t-kameyama: [add46db](https://github.com/JetBrains/intellij-community/pull/2083/commits/add46dbcca6ac0dc9c892ef1e080ad09a43b5283)

plugins/kotlin/idea/tests/testData/quickfix/addStarProjections/checkType/array.kt
t-kameyama: [add46db](https://github.com/JetBrains/intellij-community/pull/2083/commits/add46dbcca6ac0dc9c892ef1e080ad09a43b5283)

plugins/kotlin/idea/tests/testData/quickfix/convertToIsArrayOfCall/nullableArrayLhs.kt.after
t-kameyama: c2f5b32

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertToIsArrayOfCallFix.kt
t-kameyama: c2f5b32

abelkov (approved):
Thank you, merged with a few corrections

PR #2084, opened 2022-07-11
platform: interprate 'os.name' property as case insensitive

alonbl:
Given:

```java
public class Test {
    public static void main(String... args) {
        System.out.println(String.format("os.name=%s", System.getProperty("os.name")));
    }
}
```

Output:

```
os.name=Linux
```

Notice the mixed case.

Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>


ybedrov:
Hello! Could you please clarify why case insensitive result is needed?

alonbl:
> Hello! Could you please clarify why case insensitive result is needed?

Oh! I am so sorry, I did not notice that OS_NAME is converted to _OS_NAME with lower case.

PR #2088, opened 2022-07-16
[java-inspections] IDEA-298068 Incorrect DateTimeFormat pattern Inspection

pyltsin:
I tried to add a new inspection

https://youtrack.jetbrains.com/issue/IDEA-298068

I hope I did it correctly,

### DIFF
--- a/plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
+++ b/plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
@@ -1,0 +1,262 @@
+// Copyright 2000-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package com.siyeh.ig.bugs;
+
+import com.intellij.codeInspection.AbstractBaseJavaLocalInspectionTool;
+import com.intellij.codeInspection.LocalQuickFix;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.openapi.util.TextRange;
+import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.psi.*;
+import com.intellij.psi.util.ConstantExpressionUtil;
+import com.siyeh.InspectionGadgetsBundle;
+import com.siyeh.ig.callMatcher.CallMatcher;
+import com.siyeh.ig.psiutils.ExpressionUtils;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static com.siyeh.ig.callMatcher.CallMatcher.*;
+
+public class IncorrectDateTimeFormatInspection extends AbstractBaseJavaLocalInspectionTool {
+
+  private interface CountVerifier {
+    boolean verifier(int count);
+  }
+
+  private static class RangeVerifier implements CountVerifier {
+    private final int from;
+    private final int to;
+
+    private RangeVerifier(int from, int to) {
+      this.from = from;
+      this.to = to;
+    }
+
+    @Override
+    public boolean verifier(int count) {
+      return count >= from && count <= to;
+    }
+  }
+
+  private static RangeVerifier rangeOf(int from, int to) {
+    return new RangeVerifier(from, to);
+  }
+
+  private static class SetVerifier implements CountVerifier {
+    private final Set<Integer> set;
+
+    private SetVerifier(int... numbers) {
+      set = Arrays.stream(numbers).boxed()
+        .collect(Collectors.toSet());
+    }
+
+    @Override
+    public boolean verifier(int count) {
+      return set.contains(count);
+    }
+  }
+
+  private static SetVerifier setOf(int... numbers) {
+    return new SetVerifier(numbers);
+  }
+
+  Map<Character, CountVerifier> DATE_TIME_FORMATTER_ALLOWED = Map.ofEntries(
+    Map.entry('G', rangeOf(1, 5)),
+    Map.entry('u', rangeOf(1, 19)),
+    Map.entry('y', rangeOf(1, 19)),
+    Map.entry('Y', rangeOf(1, Integer.MAX_VALUE)), //according to refs must be until 19, but now there is no exception
+    Map.entry('Q', rangeOf(1, 5)),
+    Map.entry('q', rangeOf(1, 5)),
+    Map.entry('M', rangeOf(1, 5)),
+    Map.entry('L', rangeOf(1, 5)),
+    Map.entry('w', rangeOf(1, 2)),
+    Map.entry('W', rangeOf(1, 1)),
+    Map.entry('d', rangeOf(1, 2)),
+    Map.entry('D', rangeOf(1, 3)),
+    Map.entry('F', rangeOf(1, 1)),
+    Map.entry('g', rangeOf(1, 19)),
+    Map.entry('E', rangeOf(1, 5)),
+    Map.entry('e', rangeOf(1, 5)),
+    Map.entry('c', rangeOf(1, 5)),
+
+    Map.entry('a', rangeOf(1, 1)),
+    Map.entry('h', rangeOf(1, 2)),
+    Map.entry('H', rangeOf(1, 2)),
+    Map.entry('k', rangeOf(1, 2)),
+    Map.entry('K', rangeOf(1, 2)),
+    Map.entry('m', rangeOf(1, 2)),
+    Map.entry('s', rangeOf(1, 2)),
+    Map.entry('S', rangeOf(1, 9)), //limit as nanos
+    Map.entry('A', rangeOf(1, 19)),
+    Map.entry('n', rangeOf(1, 19)),
+    Map.entry('N', rangeOf(1, 19)),
+
+    Map.entry('B', setOf(1, 4, 5)),
+
+    Map.entry('V', rangeOf(2, 2)),
+    Map.entry('v', setOf(1, 4)),
+    Map.entry('z', rangeOf(1, 4)),
+    Map.entry('O', setOf(1, 4)),
+    Map.entry('X', rangeOf(1, 5)),
+    Map.entry('x', rangeOf(1, 5)),
+    Map.entry('Z', rangeOf(1, 5))
+  );
+
+  private static final CallMatcher PATTERN_METHODS = anyOf(
+    staticCall("java.time.format.DateTimeFormatter", "ofPattern"),
+    instanceCall("java.time.format.DateTimeFormatterBuilder", "appendPattern").parameterTypes(CommonClassNames.JAVA_LANG_STRING)
+  );
+
+  @NotNull
+  @Override
+  public PsiElementVisitor buildVisitor(@NotNull ProblemsHolder holder, boolean isOnTheFly) {
+    return new JavaElementVisitor() {
+      @Override
+      public void visitMethodCallExpression(@NotNull PsiMethodCallExpression call) {
+        if (PATTERN_METHODS.test(call)) {
+          processExpression(call.getArgumentList().getExpressions()[0]);
+        }
+      }
+
+      private void processExpression(@NotNull PsiExpression expression) {
+        Object patternObject = ConstantExpressionUtil.computeCastTo(expression,
+                                                                    PsiType.getJavaLangString(expression.getManager(),
+                                                                                              expression.getResolveScope()));
+        if (!(patternObject instanceof String)) return;
+        String pattern = (String)patternObject;
+        List<Token> tokens = new ArrayList<>();
+        int optionalDepth = 0;
+        char[] array = pattern.toCharArray();
+        boolean inQuote = false;
+        int lastQuoteIndex = 0;
+        int length = 0;
+        for (int pos = 0; pos < array.length; pos++) {
+          char c = array[pos];
+
+          //process tokens
+          if (!inQuote && isLetter(c)) {
+            length++;
+            if (pos != array.length - 1 && c == array[pos + 1]) {
+              continue;
+            }
+            else {
+              tokens.add(new Token(pos - length + 1, array, length));
+              checkPadding(expression, array, pos - length + 1, pos + 1);
+            }
+          }
+          length = 0;
+
+
+          //check if it starts quote region or one quote
+          //case with one quote ('') is processed as quote region
+          if (c == '\'') {
+            inQuote = !inQuote;
+            lastQuoteIndex = pos;
+            continue;
+          }
+          //skip whole region in quote
+          if (inQuote) {
+            continue;
+          }
+
+          checkUnsupportedSymbolsProblem(expression, pos, c);
+
+          if (c == '[') {
+            optionalDepth++;
+          }
+          if (c == ']') {
+            optionalDepth--;
+          }
+        }
+
+        checkInQuoteProblem(expression, inQuote, lastQuoteIndex);
+        checkOptionalDepthProblem(expression, optionalDepth);
+
+        for (Token token : tokens) {
+          checkAvailableCount(expression, token, holder);
+        }
+      }
+
+      private void checkOptionalDepthProblem(@NotNull PsiExpression expression, int optionalDepth) {
+        if (optionalDepth < 0) {
+          holder.registerProblem(expression, null,
+                                 InspectionGadgetsBundle.message("inspection.incorrect.date.format.message.unpaired", ']'),
+                                 LocalQuickFix.EMPTY_ARRAY);
+        }
+      }
+
+      private void checkInQuoteProblem(@NotNull PsiExpression expression, boolean inQuote, int lastQuoteIndex) {
+        if (inQuote) {
+          TextRange range = ExpressionUtils.findStringLiteralRange(expression, lastQuoteIndex, lastQuoteIndex + 1);
+          if (range != null) {
+            holder.registerProblem(expression, range,
+                                   InspectionGadgetsBundle.message("inspection.incorrect.date.format.message.unpaired", '\''),
+                                   LocalQuickFix.EMPTY_ARRAY);
+          }
+        }
+      }
+
+      private void checkUnsupportedSymbolsProblem(@NotNull PsiExpression expression, int pos, char c) {
+        if (c == '{' || c == '}' || c == '#') {
+          TextRange range = ExpressionUtils.findStringLiteralRange(expression, pos, pos + 1);
+          holder.registerProblem(expression, range,
+                                 InspectionGadgetsBundle.message("inspection.incorrect.date.format.message.unsupported", c),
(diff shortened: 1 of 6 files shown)


plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+
+  private static RangeVerifier rangeOf(int from, int to) {
+    return new RangeVerifier(from, to);
+  }
+
+  private static class SetVerifier implements CountVerifier {
+    private final Set<Integer> set;
+
+    private SetVerifier(int... numbers) {
+      set = Arrays.stream(numbers).boxed()
```
amaembo, line 51: Probably use it.unimi.dsi.fastutil.ints.IntOpenHashSet? It's available in platform and much more efficient. Or a BitSet, taken into account current uses.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+    public boolean verifier(int count) {
+      return set.contains(count);
+    }
+  }
+
+  private static SetVerifier setOf(int... numbers) {
+    return new SetVerifier(numbers);
+  }
+
+  Map<Character, CountVerifier> DATE_TIME_FORMATTER_ALLOWED = Map.ofEntries(
```
amaembo, line 65: static?

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+    Map.entry('V', rangeOf(2, 2)),
+    Map.entry('v', setOf(1, 4)),
+    Map.entry('z', rangeOf(1, 4)),
+    Map.entry('O', setOf(1, 4)),
+    Map.entry('X', rangeOf(1, 5)),
+    Map.entry('x', rangeOf(1, 5)),
+    Map.entry('Z', rangeOf(1, 5))
+  );
+
+  private static final CallMatcher PATTERN_METHODS = anyOf(
```
amaembo, line 107: com.siyeh.ig.bugs.SuspiciousDateFormatInspection matches also SimpleDateFormat.applyPattern/applyLocalizedPattern which are (hopefully) using the same syntax. Also, a constructor for SimpleDateFormat (via `visitNewExpression`). Probably it's possible to reuse some logic from there?

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+    return new JavaElementVisitor() {
+      @Override
+      public void visitMethodCallExpression(@NotNull PsiMethodCallExpression call) {
+        if (PATTERN_METHODS.test(call)) {
+          processExpression(call.getArgumentList().getExpressions()[0]);
+        }
+      }
+
+      private void processExpression(@NotNull PsiExpression expression) {
+        Object patternObject = ConstantExpressionUtil.computeCastTo(expression,
```
amaembo, line 124: You may use `com.siyeh.ig.psiutils.ExpressionUtils#computeConstantExpression(com.intellij.psi.PsiExpression)`. You don't need to materialize java.lang.String type there.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+        checkOptionalDepthProblem(expression, optionalDepth);
+
+        for (Token token : tokens) {
+          checkAvailableCount(expression, token, holder);
+        }
+      }
+
+      private void checkOptionalDepthProblem(@NotNull PsiExpression expression, int optionalDepth) {
+        if (optionalDepth < 0) {
+          holder.registerProblem(expression, null,
```
amaembo, line 184: Probably would be possible to highlight the unpaired ']' specifically?

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+                                 InspectionGadgetsBundle.message("inspection.incorrect.date.format.message.unpaired", ']'),
+                                 LocalQuickFix.EMPTY_ARRAY);
+        }
+      }
+
+      private void checkInQuoteProblem(@NotNull PsiExpression expression, boolean inQuote, int lastQuoteIndex) {
+        if (inQuote) {
+          TextRange range = ExpressionUtils.findStringLiteralRange(expression, lastQuoteIndex, lastQuoteIndex + 1);
+          if (range != null) {
```
amaembo, line 193: What if range is null? It's possible, e.g., if we are analysing the reference to a constant variable. Should we skip highlighting at all?

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+      }
+    };
+  }
+
+  private static class Token {
+    final char character;
+    final int pos;
+    final int length;
+
+    Token(int pos, char[] chars, int length) {
```
amaembo, line 251: Probably reorder parameters and put chars at the first place? It's commonly used this way in method signatures (e.g., `java.io.Reader#read(char[], int, int)`)

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,262 @@
...
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static com.siyeh.ig.callMatcher.CallMatcher.*;
+
+public class IncorrectDateTimeFormatInspection extends AbstractBaseJavaLocalInspectionTool {
+
+  private interface CountVerifier {
+    boolean verifier(int count);
```
amaembo, line 25: Let's use a verb for a method name (`verify`)

plugins/InspectionGadgets/src/inspectionDescriptions/IncorrectDateTimeFormat.html
```diff
@@ -0,0 +1,22 @@
...
+<p>The following errors are reported:</p>
+<ul>
+  <li>Unsupported patterns, like "TT"</li>
+  <li>Using reserved patterns, like "#"</li>
+  <li>Unappropriated padding.</li>
+  <li>Optional brackets without a pair.</li>
+</ul>
+<p>
+  Examples:<br>
+  <code>DateTimeFormatter.ofPattern("[][]])"<br></code>
```
amaembo, line 14: Would be nice to specify what is exactly incorrect in every sample (e.g., in parentheses after the sample)

plugins/InspectionGadgets/src/inspectionDescriptions/IncorrectDateTimeFormat.html
```diff
@@ -0,0 +1,22 @@
...
+  <code>DateTimeFormatter.ofPattern("TT")<br></code>
+  <code>DateTimeFormatter.ofPattern("[[")<br></code>
+  <code>DateTimeFormatter.ofPattern("{")<br></code>
+  <code>DateTimeFormatter.ofPattern("MMMMMM")<br></code>
+</p>
+<!-- tooltip end -->
```
amaembo, line 20: Please add `<p><small>New in 2022.3</small></p>` after tooltip end. Thanks

amaembo (changes requested):
Thanks for the great contribution. Please address the minor problems noted.

plugins/InspectionGadgets/src/inspectionDescriptions/IncorrectDateTimeFormat.html
pyltsin: Added, thank you

plugins/InspectionGadgets/src/inspectionDescriptions/IncorrectDateTimeFormat.html
pyltsin: Some short explanations were added

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Shame on me. Thank you! Fixed

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Parameters were reordered. Thank you!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Thank you for this comment. I fixed it and added one more test case.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Thank you. This part of code was rewritten and I added several test cases.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: I rewrote with `IntOpenHashSet`

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Of course. Thank you!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: This is a great question. I've thought about that, but there are several problems, which complicate this inspection
The main problem is that the syntax patterns are similar, when common cases are used, but they have completely different limits.
For example:
   - `DateTimeFormatter.ofPattern("pppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppMMpM");` 
we don't expect any exception.
But ` new SimpleDateFormat("pppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppMMpM")` produces exception, as 'p' is unsupported token.

- Patterns for DateTimeFormatter have many limits (like "MMMMMM" is too big). But SimpleDateFormat doesn' have these limits ("X" is an exception, and must be less than 4 - java.text.SimpleDateFormat#encode)

- DateTimeFormatter doesn't use optional brackets.

About `applyLocalizedPattern`, it seems to be too complex to check, because 
```
    public void applyLocalizedPattern(String pattern) {
         String p = translatePattern(pattern,
                                     formatData.getLocalPatternChars(),
                                     DateFormatSymbols.patternChars);
         compiledPattern = compile(p);
         this.pattern = p;
    }
```
and  it depends on local patterns, which can contain absolutely different tokens. 
Here, I replace standard token for year "y" to "a";
```
import java.text.DateFormatSymbols;
import java.text.SimpleDateFormat;
import java.util.Date;

public class Test {
        SimpleDateFormat p = new SimpleDateFormat();
        DateFormatSymbols newFormatSymbols = new DateFormatSymbols();
        newFormatSymbols.setLocalPatternChars("Ga"); //standard pattern - "GyMdkHmsSEDFwWahKzZYuXL". I change the second letter.
        p.setDateFormatSymbols(newFormatSymbols);
        p.applyLocalizedPattern("aaaa");
        System.out.println(p.format(new Date())); //2022 if we live in 2022 year.
    }
}
```

Honestly, I am not sure, but, perhaps, `applyLocalizedPattern` doesn't suit for `SuspiciousDateFormatInspection` because of that.

pyltsin:
Thank you for this review! I've fixed (I hope)

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
amaembo: Ok, fair enough. Let's keep it as is for now.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,264 @@
...
+
+import static com.siyeh.ig.callMatcher.CallMatcher.*;
+
+public class IncorrectDateTimeFormatInspection extends AbstractBaseJavaLocalInspectionTool {
+
+  private interface CountVerifier {
+    boolean verify(int count);
+  }
+
+  private static class RangeVerifier implements CountVerifier {
```
amaembo, line 30: As CountVerifier is a functional interface, do we need such verbose verifiers? It looks like, capturing lambdas will work in both cases, like

```java
  private static CountVerifier rangeOf(int from, int to) {
    return count -> count >= from && count <= to;
  }
  
  private static CountVerifier setOf(int... numbers) {
    return IntOpenHashSet.of(numbers)::contains;
  }
```

Or you can even go further and replace `CountVerifier` with standard `IntPredicate` which is moreless natural here.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,264 @@
...
+
+  private static SetVerifier setOf(int... numbers) {
+    return new SetVerifier(numbers);
+  }
+
+  private static final Map<Character, CountVerifier> DATE_TIME_FORMATTER_ALLOWED = Map.ofEntries(
+    Map.entry('G', rangeOf(1, 5)),
+    Map.entry('u', rangeOf(1, 19)),
+    Map.entry('y', rangeOf(1, 19)),
+    Map.entry('Y', rangeOf(1, Integer.MAX_VALUE)), //according to refs must be until 19, but now there is no exception
```
amaembo, line 70: You mean, according to the specification? I would say something like "according to the specification, must not exceed 19 but current implementation doesn't restrict this"

amaembo (changes requested):

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Thank you! Great idea!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Thank you. Replaced

pyltsin:
Thank you for this recommendations! I've fixed

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
```diff
@@ -0,0 +1,233 @@
...
+import static com.siyeh.ig.callMatcher.CallMatcher.*;
+
+public class IncorrectDateTimeFormatInspection extends AbstractBaseJavaLocalInspectionTool {
+
+  private static IntPredicate rangeOf(int from, int to) {
+    return count -> count >= from && count <= to;
+  }
+
+  private static IntPredicate setOf(int... numbers) {
+    return  count -> IntOpenHashSet.of(numbers).contains(count);
```
amaembo, line 31: This way you recreate the set every time lambda is invoked, which looks excessive. I suggested to use method reference to create it once. Alternatively, you can get rid of the set at all and use linear search (taking into account that your sets are really small):
`count -> ArrayUtil.indexOf(numbers, count) >= 0`

amaembo (changes requested):

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/IncorrectDateTimeFormatInspection.java
pyltsin: Shame on me. Thank you! `ArrayUtil.indexOf(numbers, count) >= 0` is much faster than Set.  Fixed

pyltsin:
Sorry for this error =( I've fixed

amaembo:
Merged to internal repo, thanks.

amaembo:
c0d844fd45a4fad12712f8526491e3d2f80bcb65

PR #2091, opened 2022-07-23
[java-inspections] IDEA-298475 Redundant `compareTo` for java.time classes

pyltsin:
I tried to add a new inspection

https://youtrack.jetbrains.com/issue/IDEA-298475

I hope I did it correctly,


plugins/InspectionGadgets/InspectionGadgetsAnalysis/resources/messages/InspectionGadgetsBundle.properties
```diff
@@ -1657,6 +1657,8 @@ if.may.be.factorized.problem.descriptor=<code>#ref</code> can be factorized #loc
 if.may.be.factorized.quickfix=Replace with factorized expression
 redundant.string.format.call.display.name=Redundant call to 'String.format()'
 redundant.call.problem.descriptor=Redundant call to <code>#ref()</code> #loc
+inspection.simplifiable.compare.java.time.display.name=Comparison with compareTo can be simplified
```
BasLeijdekkers, line 1660: How about: `Expression with ''java.time'' <code>#ref()</code> call can be simplified`.
#ref will expand to the highlighted area (which will have to be modified a bit as well)

plugins/InspectionGadgets/src/com/intellij/codeInspection/RedundantCompareToJavaTimeInspection.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        if (relationType == null) return;
+        if (ExpressionUtils.isZero(binOp.getLOperand())) {
+          relationType = relationType.getFlipped();
+          if (relationType == null) return;
+        }
+        else if (!ExpressionUtils.isZero(binOp.getROperand())) {
+          return;
+        }
+        holder.registerProblem(call, InspectionGadgetsBundle.message("redundant.call.problem.descriptor"),
```
BasLeijdekkers, line 65: There is an easier way to highlight only the method identifier, without specifying a separate `TextRange`. See for example com.intellij.codeInspection.java18api.OptionalGetWithoutIsPresentInspection.

platform/core-api/src/com/intellij/psi/CommonClassNames.java
```diff
@@ -132,5 +132,10 @@ public interface CommonClassNames {
 
   String CLASS_FILE_EXTENSION = ".class";
 
+  String JAVA_TIME_LOCAL_DATE = "java.time.LocalDate";
```
BasLeijdekkers, line 135: We use `CommonClassNames` only for class names that are actually common. These names are not (so far at least). Could you remove them from this class?

plugins/InspectionGadgets/src/META-INF/InspectionGadgets.xml
```diff
@@ -2690,6 +2690,11 @@
                      groupKey="group.names.verbose.or.redundant.code.constructs" enabledByDefault="true" level="WARNING"
                      editorAttributes="NOT_USED_ELEMENT_ATTRIBUTES"
                      implementationClass="com.siyeh.ig.controlflow.UnnecessaryReturnInspection" cleanupTool="true"/>
+    <localInspection groupPath="Java" language="JAVA" suppressId="RedundantCompareToJavaTimeStatement" shortName="RedundantCompareToJavaTime"
+                     groupBundle="messages.InspectionsBundle" groupKey="group.names.verbose.or.redundant.code.constructs"
+                     enabledByDefault="false" level="WARNING" editorAttributes="NOT_USED_ELEMENT_ATTRIBUTES"
```
BasLeijdekkers, line 2695: 1. I think this inspection is okay to enable by default. 
2. Please remove the `editorAttributes` attribute, because not-used highlighting is reserved for elements which can be removed without changing functionality and in this case the expression will be transformed, but not removed. 
3. Remove the `suppressId` attribute as well. The `shortName` will then be used for suppression and it is more correct in my opinion.

java/java-impl/src/inspectionDescriptions/RedundantCompareToJavaTime.html
```diff
@@ -0,0 +1,19 @@
+<!-- Copyright 2000-2022 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license. -->
+<html>
+<body>
+Reports comparisons in which the <code>compareTo</code> method can be replaced to <code>isAfter</code>, <code>isBefore</code> or <code>isEqual</code>
+of LocalTime, LocalDateTime, LocalDate, OffsetTime or OffsetDateTime.
```
BasLeijdekkers, line 5: I would replace with something like:
```
Reports <code>java.time</code> comparisons with <code>compareTo()</code> calls that can be replaced with 
<code>isAfter()</code>, <code>isBefore()</code> or <code>isEqual()</code> calls.
```
motivations:
- this mentions java.time immediately to avoid confusion about which kind of compareTo() methods it handles
- all references to methods include  parentheses to avoid confusion
- mentioning the individual classes becomes unnecessary because of the java.time package reference.

BasLeijdekkers (changes requested):
Nice work so far!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/resources/messages/InspectionGadgetsBundle.properties
```diff
@@ -1657,6 +1657,8 @@ if.may.be.factorized.problem.descriptor=<code>#ref</code> can be factorized #loc
...
+inspection.simplifiable.compare.java.time.fix.name=Replace with ''{0}'' method
```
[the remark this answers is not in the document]
pyltsin: Changed. The inspection uses `fix.replace.with.x.call` now. Thank you!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/resources/messages/InspectionGadgetsBundle.properties
pyltsin: Updated display name and problem descriptor. Thank you.

plugins/InspectionGadgets/src/com/intellij/codeInspection/RedundantCompareToJavaTimeInspection.java
pyltsin: Changed. Thank you!

platform/core-api/src/com/intellij/psi/CommonClassNames.java
pyltsin: Deleted in `CommonClassNames`

plugins/InspectionGadgets/src/META-INF/InspectionGadgets.xml
pyltsin: Changed. Thank you!

java/java-impl/src/inspectionDescriptions/RedundantCompareToJavaTime.html
pyltsin: This phrase is much better. Changed. Thank you!

pyltsin:
Thank you for this review! I've fixed (I hope)

plugins/InspectionGadgets/InspectionGadgetsAnalysis/resources/messages/InspectionGadgetsBundle.properties
```diff
@@ -1657,6 +1657,9 @@ if.may.be.factorized.problem.descriptor=<code>#ref</code> can be factorized #loc
 if.may.be.factorized.quickfix=Replace with factorized expression
 redundant.string.format.call.display.name=Redundant call to 'String.format()'
 redundant.call.problem.descriptor=Redundant call to <code>#ref()</code> #loc
+inspection.simplifiable.compare.java.time.display.name=Expression with 'java.time' 'compareTo' call can be simplified
+inspection.simplifiable.compare.java.time.family.name=Simplify expression with 'java.time' 'compareTo' call
```
BasLeijdekkers, line 1661: Please add parentheses to `compareTo` here as well:
`Expression with 'java.time' 'compareTo()' call can be simplified`

plugins/InspectionGadgets/src/com/intellij/codeInspection/RedundantCompareToJavaTimeInspection.java
```diff
@@ -0,0 +1,152 @@
...
+    CallMatcher.instanceCall(JAVA_TIME_LOCAL_TIME, "compareTo").parameterTypes(JAVA_TIME_LOCAL_TIME),
+    CallMatcher.instanceCall(JAVA_TIME_LOCAL_DATE_TIME, "compareTo")
+      .parameterTypes("java.time.chrono.ChronoLocalDateTime"),
+    CallMatcher.instanceCall(JAVA_TIME_OFFSET_TIME, "compareTo").parameterTypes(JAVA_TIME_OFFSET_TIME),
+    CallMatcher.instanceCall(JAVA_TIME_OFFSET_DATE_TIME, "compareTo")
+      .parameterTypes(JAVA_TIME_OFFSET_DATE_TIME)
+  );
+
+  @Override
+  public @NotNull String getID() {
```
BasLeijdekkers, line 37: The default ID is the `shortName`, so it is not necessary to override `getID()` if the ID is the same as the shortName.

BasLeijdekkers:
Could you squash your commits into a single one? It makes merging easier.

BasLeijdekkers (changes requested):

plugins/InspectionGadgets/InspectionGadgetsAnalysis/resources/messages/InspectionGadgetsBundle.properties
pyltsin: Fixed, thank you!

plugins/InspectionGadgets/src/com/intellij/codeInspection/RedundantCompareToJavaTimeInspection.java
pyltsin: Deleted, Thank you!

pyltsin:
Commits are squashed

BasLeijdekkers (approved):

BasLeijdekkers:
Thank you for your work! I have merged it manually [here](https://github.com/JetBrains/intellij-community/commit/794497240b696357489d9807f3d9b4bc657df311).

PR #2092, opened 2022-07-23
[IDEA-293658] Allow to use placeholders in Markdown run commands feature

i582:
# Placeholders in Markdown block code

Placeholders are piece of command like `%name%` that describe some common value in a command or set of commands.

For example ([link](https://github.com/EXL/BeGameLauncher/blob/master/ReadMe.md) to the original repository):

```sh
mkdir -p res/%name%/
mkdir -p locales/%name%/pre/
```

Here we want these two commands to be executed with some `name` value.

However, at the moment, if you try to run this command in the IDE through the icon next to the block of code, then it will simply try to execute the command, which, of course, will lead to errors.

This PR adds support for such commands by adding an intermediate dialog where the user is prompted to enter placeholder values.

Fixes [IDEA-293658](https://youtrack.jetbrains.com/issue/IDEA-293658/Allow-to-use-placeholders-in-Markdown-run-commands-feature)


GerMoranBYON:
This would be a useful feature to get merged.

PR #2095, opened 2022-07-28
Added ability to open file in active tab group regardless of the presence of the file in other tab groups

Cavantar:
This attempts to add back functionality that was present in intellij some time ago.
Previously on OpenFile/Project/* Modal you used be able to open file in active pane using Alt+Shift+Enter
It's very annoying when instead of opening file in current tab group it's changing the file in the other tab group that was used specifically for context if target file happens to be in that tab group.
In case there are no other open file modifiers and alt is pressed it opens file in selected tab group.

This is my first public pr, I don't know much about the project.
There may be significant issues with this approach. 
Please point me in the right direction if there are issues.


yole:
The main problem with this approach is that it hard-codes the modifier key used for the new behavior, whereas other open modes (Open in new window, Open in right split) allow modifying the shortcuts. The correct approach is to add a new action "Open in current tab group" and to use its modifier key to trigger the new behavior, as is done for other actions in the `getOpenMode` method.

batya239:
Hey, Jakub! Feel free to recreate PR once it will address problems mentioned above.

PR #2097, opened 2022-08-01
Startup param parsing: do not break on first param

dunno99:
Startup param parsing is erroneously returning early on the first
recognized paramter. This limits the number of startup parameters
to just one.

The fix is to remove the "break" statements, which presumably is
a left over from an optimization when there was only one parameter.


yole:
Thanks! I've manually merged your changes.

PR #2102, opened 2022-08-11
IDEA-298773: Fix memory leak in GotoSymbolModel2

npaige:
Use SymbolSearchEverywhereContributor instead of Project for the parent
disposable for the extension point listeners.


npaige:
Closing in favor of https://github.com/JetBrains/intellij-community/pull/2125, which also updates usages in tests.

PR #2104, opened 2022-08-14
Update README.md

dhruvmaindola227:
Capitalized characters at line 50 and 51 in README.md


chashnikov:
Thank you for the contribution! It appears that this part can be simplified given that IDEA 2022.2 or later is required for intellij-community project, so I've rewritten it.

PR #2108, opened 2022-08-19
[gradle] Watch libs.versions.toml

fp7:
Issue: IDEA-293328


HackerMadCat:
Thanks a lot! I'll try to pull it into master.

ps. It isn't full fix but it is better then nothing.
Gradle allows to defined any file as versions catalog.

fp7:
@HackerMadCat I completed the fix and added tests for non default version catalogs.

fp7:
@HackerMadCat Some news about merging this? I would be happy to invest some more time if this pr needs more work.

HackerMadCat:
Commit: d66a302553533e341baeef0cee912856a6db5a25

PR #2109, opened 2022-08-19
Fix SourceDirs HashSet ordering for consistent GradleOutputDir

vmadalin:
### Context
Due to the recent merge of the `intellij-community` into `studio` we figure out that one of our tests snapshot tests started to be flaky due to the following commit introduced recently: https://github.com/JetBrains/intellij-community/commit/dff59a2d1944e020dcb4e4d2c176c9b9a3b79721

### Issue
Given that Gradle model can contain multiple output directories (for each language: java, scala, kotlin, etc.) right now with the above changes the `firstExistingLang` isn't always consistent on the different runs when project have multiple source directories. This is strictly linked of `findFirst()` of an HashSet that doesn't maintains the order of the elements, having the following:

- srcDirs java first
srcDirs java first

- srcDirs kotlin first
srcDirs kotlin first

### Test
I added a specific test where this behaviour is now validated, given a project with `java library` and `kotlin` plugin applied assert that `compilerOutputUrl` and `CompilerOutputUrlForTests` are always the expected `java classesOutputPath`

### Next
For sure this part should be reconsidered in the future where probably `getGradleOutputDir` should return a list of files but until that will be great to keep consistency on the results

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/CommonGradleProjectResolverExtension.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/CommonGradleProjectResolverExtension.java
@@ -661,6 +661,7 @@ public final class CommonGradleProjectResolverExtension extends AbstractProjectR
   private static File getGradleOutputDir(@Nullable ExternalSourceDirectorySet sourceDirectorySet) {
     if (sourceDirectorySet == null) return null;
     String firstExistingLang = sourceDirectorySet.getSrcDirs().stream()
+      .sorted()
       .filter(File::exists)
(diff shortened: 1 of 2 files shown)


vmadalin:
I'm going to close this PR since those changes are already `cherry-picked` on the `master` branch by the following commits after validation:
- https://github.com/JetBrains/intellij-community/commit/95dc10510ffa2426072debf4b6ccd135f8a06e7e (original change)
- https://github.com/JetBrains/intellij-community/commit/107d5b241f2952d426cd61fdf0dba5ccea80f8d1 (small code style improvements)

Thank you

PR #2110, opened 2022-08-19
IDEA-240111 use Gradle incremental task inputs/outputs to avoid re-running test tasks

aSemy:
https://youtrack.jetbrains.com/issue/IDEA-240111

This PR demonstrates how IntelliJ can use [Gradle task inputs/outputs](https://docs.gradle.org/7.5.1/userguide/more_about_tasks.html#sec:up_to_date_checks) to avoid re-running tasks that have not changed.

I feel this change is really important, as Gradle incremental build is a really powerful tool. Implementing this change (once it is tidied up and has tests!) would have a huge impact on users, as their workflow would become faster. Additionally the power requirements of repeatedly re-running the same tests would be avoided, helping reduce energy usage worldwide.

### Summary

`IJTestEventLogger` now creates an output directory property, `ijTestEvents`. Instead of printing the XML task update to stdout, it logs it to the output directory. The files are named by timestamp and test-event type so they can be ordered. (I added an md5 checksum just in case there's a clash - but there's probably a better solution.)

The init script also registers a new task, `ijTestEventLogger`, which captures the XML output of all test tasks. It then prints the XML to stdout, which is picked up by IntelliJ.

`ijTestEventLogger` is set to always run (it is never up-to-date), and is a [finalizer task](https://docs.gradle.org/7.5.1/userguide/more_about_tasks.html#sec:finalizer_tasks) so it will always run, no matter if the test tasks pass or fail.

### Potential Improvements

* This update makes the visuals a little strange, because IntelliJ doesn't live update tests like it did before. It waits for all tests in a subproject to run before updating them all very quickly. Perhaps still continue printing lines from each test task, but in the new `ijTestEventLogger` task use incremental inputs to only print files when their status is 'MODIFIED', not 'ADDED'?
* Further performance improvements:  Perhaps use the Workers API to split the read/write load up? Find a better way to save/load files? Coroutines? AsynchronousFileChannel?

### Other changes 

* renamed `addTestListener.groovy` to `addTestListener.init.gradle` (better IDE integration)
 
### TODO

- [x] Improve performance. It's very bad at the moment when there is a lot of output logged, because this creates a large number of XML files. Gradle has to fingerprint them all Perhaps create one file per test-id+event type, and append each test event to that file? Then split the file later.
- [ ] Remove debug logging
- [ ] Add the same logic for `logTestReportLocation()`, `logConfigurationError()`
- [ ] Check if `KotlinMppTestLogger` and the Android version need to be updated too (although I think these should be done in separate PRs)
- [ ] Create a ticket to create an [init script plugin](https://docs.gradle.org/current/userguide/init_scripts.html#sec:init_script_plugins) (working with Groovy is pretty rough, and it's very difficult to test)

### DIFF
--- a/plugins/gradle/java/resources/org/jetbrains/plugins/gradle/java/addTestListener.init.gradle
+++ b/plugins/gradle/java/resources/org/jetbrains/plugins/gradle/java/addTestListener.init.gradle
@@ -1,8 +1,3 @@
-import org.gradle.api.Task
-import org.gradle.api.tasks.Sync
-import org.gradle.api.tasks.TaskProvider
-import org.gradle.api.tasks.testing.Test
-
 //file:noinspection GrPackage
 
 gradle.allprojects { project ->
--- a/plugins/gradle/java/resources/org/jetbrains/plugins/gradle/java/addTestListener.init.gradle
+++ b/plugins/gradle/java/resources/org/jetbrains/plugins/gradle/java/addTestListener.init.gradle
@@ -5,7 +5,7 @@ gradle.allprojects { project ->
     loggerTask.outputs.upToDateWhen { false }
 
     from(project.tasks.withType(Test)) {
-      include("**/*.xml")
+      include("**/ijLog*.xml")
     }
     into(loggerTask.temporaryDir)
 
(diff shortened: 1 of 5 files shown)


nskvortsov:
Logging test events to a file and reporting them after the build or in final tasks is not an acceptable solution. Tests should be reported on-the-fly.

PR #2117, opened 2022-08-23
Add options for interacting with empty commits to rebase

MaciejG604:


plugins/git4idea/src/git4idea/rebase/GitRebaseOption.kt
```diff
@@ -15,7 +15,9 @@ enum class GitRebaseOption(@NonNls private val option: String,
   REBASE_MERGES("--rebase-merges", GitBundle.message("rebase.option.rebase.merges")),
   KEEP_EMPTY("--keep-empty", GitBundle.message("rebase.option.keep.empty")),
   ROOT("--root", GitBundle.message("rebase.option.root")),
-  INTERACTIVE("--interactive", GitBundle.message("rebase.option.interactive"));
+  INTERACTIVE("--interactive", GitBundle.message("rebase.option.interactive")),
+  EMPTY_DROP("--empty=drop", GitBundle.message("rebase.option.empty.drop")),
+  EMPTY_KEEP("empty=keep", GitBundle.message("rebase.option.empty.keep"));
```
PawelLipski, line 20: Hmm no `--`? 🤔

plugins/git4idea/resources/messages/GitBundle.properties
```diff
@@ -195,7 +195,9 @@ rebase.option.onto=Specify a new base for the rebased commits
 rebase.option.interactive=Edit commits before rebasing
 rebase.option.rebase.merges=Recreate commits topology
 rebase.option.switch.to.branch=Select another branch to rebase
-rebase.option.keep.empty=Do not remove empty commits during rebase
+rebase.option.keep.empty=Do not remove initially-empty commits during rebase
+rebase.option.empty.keep=Do not remove commits that become empty during rebase
+rebase.option.empty.drop=Remove commits that become empty during rebase
```
PawelLipski, line 200: I suppose `ask` is deliberately skipped since it'd require input on stdin from user, right?

MaciejG604: Asking what to do with commits that become empty is what couses this problem in the first place (interactive mode has `--empty=ask` by default), since git4idea does not handle it correctly. But sure, it should be there for completeness sake I will open an issue since it's a bigger fix.

PawelLipski:
Hey @dmitriysmirnovjb... any chance this might get included in 2022.3? 😅

dmitriysmirnovjb:
Hello. Thank you for the contribution.

As I understand, the issue you are trying to solve is that rebase process fails if an empty commit appears while rebasing.

Unfortunately, adding these option to the list does not seem to be the best way to solve the issue.
First, we do not want to add many specific options to the list, and these two appear rather specific.
And they are supported only since git 2.26, while IDE supports older versions of git.

Moreover, the dialog is not the only way to invoke rebase, there are other, like context action, and those are also affected, as I understand, so proposed change fixes only some scenarios.

We will look into fixing this in a more generic way.
Likely the most relevant request is https://youtrack.jetbrains.com/issue/IDEA-179249/Git-rebase-fails-when-squashing-commit-with-its-reverting-commit

PawelLipski:
> As I understand, the issue you are trying to solve is that rebase process fails if an empty commit appears while rebasing.

Yes, exactly ☹️ in such case, our user gets a somewhat surprising popup error message, and then needs to press `Continue` in that popup to successfully finalize the rebase.

> while IDE supports older versions of git.

Huh out of curiosity... what's the minimum version of git supported by git4idea actually? 🤔 

> Likely the most relevant request is https://youtrack.jetbrains.com/issue/IDEA-179249/Git-rebase-fails-when-squashing-commit-with-its-reverting-commit

Yes, AFAICS that exactly exactly the same problem as we face :/

> We will look into fixing this in a more generic way.

Do you have any ideas how to approach the problem as for now? 🤔

dmitriysmirnovjb:
> what's the minimum version of git supported by git4idea actually

It's 2.19.2 for Windows, and 2.17 for others. See [here](https://github.com/JetBrains/intellij-community/blob/223ff19ba4e69df53a193c07f7a784355e8021ac/plugins/git4idea/src/git4idea/config/GitVersion.java#L61)

> Do you have any ideas how to approach the problem as for now?

Unfortunately, it does not seem there is an easy workaround. But we will try fixing it in 2022.3

PawelLipski:
@ishulgin has it been done? could you point us to the code change?

PawelLipski:
Or have you decided against implementing that?

ishulgin:
@PawelLipski I am sorry for closing it without context. This patch can't be applied as is. I will make a proper fix in `GitRebaseParams` with a version check and will introduce an advanced option to affect an interactive rebase behavior as well.

Progress will be reported under https://youtrack.jetbrains.com/issue/IJPL-84778

PR #2120, opened 2022-08-24
KTIJ-22470 False positive 'Redundant qualifier name' caused by property and object' property with the same name

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22470

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/util/expressionExt.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/util/expressionExt.kt
@@ -8,7 +8,6 @@ import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiMember
 import com.intellij.psi.PsiMethod
-import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.builtins.StandardNames
 import org.jetbrains.kotlin.descriptors.ConstructorDescriptor
 import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
@@ -55,8 +54,9 @@ fun KtNamedDeclaration.nameIdentifierTextRangeInThis(): TextRange? = nameIdentif
 fun PsiElement.hasComments(): Boolean = anyDescendantOfType<PsiComment>()
 
 fun KtDotQualifiedExpression.hasNotReceiver(): Boolean {
-    val element = getQualifiedElementSelector()?.mainReference?.resolve() ?: return false
-    return element is KtClassOrObject ||
+    val selector = getQualifiedElementSelector() ?: return false
+    val element = selector.mainReference?.resolve() ?: return false
+    return element is KtClassOrObject && selector.text !in listOf("copy", "toString", "hashCode", "equals") ||
             element is KtConstructor<*> ||
             element is KtCallableDeclaration && element.receiverTypeReference == null && (element.containingClassOrObject is KtObjectDeclaration?) ||
             element is PsiMember && element.hasModifier(JvmModifier.STATIC) ||
--- a/plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/inspections/LocalInspectionTestGenerated.java
+++ b/plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/inspections/LocalInspectionTestGenerated.java
@@ -10653,6 +10653,26 @@ public abstract class LocalInspectionTestGenerated extends AbstractLocalInspecti
             runTest("testData/inspectionsLocal/removeRedundantQualifierName/notApplicableSameNameVariable2.kt");
         }
 
+        @TestMetadata("notApplicableSameNameVariableWithCopy.kt")
+        public void testNotApplicableSameNameVariableWithCopy() throws Exception {
+            runTest("testData/inspectionsLocal/removeRedundantQualifierName/notApplicableSameNameVariableWithCopy.kt");
+        }
+
+        @TestMetadata("notApplicableSameNameVariableWithEquals.kt")
+        public void testNotApplicableSameNameVariableWithEquals() throws Exception {
+            runTest("testData/inspectionsLocal/removeRedundantQualifierName/notApplicableSameNameVariableWithEquals.kt");
+        }
+
+        @TestMetadata("notApplicableSameNameVariableWithHashCode.kt")
+        public void testNotApplicableSameNameVariableWithHashCode() throws Exception {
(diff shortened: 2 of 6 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/util/expressionExt.kt
```diff
@@ -55,8 +54,9 @@ fun KtNamedDeclaration.nameIdentifierTextRangeInThis(): TextRange? = nameIdentif
 fun PsiElement.hasComments(): Boolean = anyDescendantOfType<PsiComment>()
 
 fun KtDotQualifiedExpression.hasNotReceiver(): Boolean {
-    val element = getQualifiedElementSelector()?.mainReference?.resolve() ?: return false
-    return element is KtClassOrObject ||
+    val selector = getQualifiedElementSelector() ?: return false
+    val element = selector.mainReference?.resolve() ?: return false
+    return element is KtClassOrObject && selector.text !in listOf("copy", "toString", "hashCode", "equals") ||
```
dimonchik0036, line 59: I don't think what it should be fixed by text search – It can break valid cases. Probably `selector.resolveToCall()?.resultingDescriptor` can help here

dimonchik0036 (changes requested):
Note: I can't reproduce the issue with anything other than `copy`, but `hasNotReceiver`really needs to be fixed

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantQualifierNameInspection.kt
```diff
@@ -153,6 +155,10 @@ private fun KtDotQualifiedExpression.applicableExpression(
     val newDescriptorFqName = newDescriptor.unwrapFakeOverrideIfNecessary().fqNameSafe
     if (originalDescriptorFqName != newDescriptorFqName) return null
 
+    if (newExpression is KtQualifiedExpression && newExpression.receiverExpression.declarationDescriptor(newContext) != receiverReference) {
+        return null
+    }
```
t-kameyama, line 160: Check references old receiver and new receiver are equal.

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/util/expressionExt.kt
```diff
@@ -58,9 +58,8 @@ fun KtNamedDeclaration.nameIdentifierTextRangeInThis(): TextRange? = nameIdentif
 fun PsiElement.hasComments(): Boolean = anyDescendantOfType<PsiComment>()
 
 fun KtDotQualifiedExpression.hasNotReceiver(): Boolean {
-    val selector = getQualifiedElementSelector() ?: return false
-    val element = selector.mainReference?.resolve() ?: return false
-    return element is KtClassOrObject && selector.text !in listOf("copy", "toString", "hashCode", "equals") ||
+    val element = getQualifiedElementSelector()?.mainReference?.resolve() ?: return false
+    return element is KtClassOrObject ||
```
t-kameyama, line 62: Reverted

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/inspections/LocalInspectionTestGenerated.java
```diff
@@ -10612,21 +10612,6 @@ public void testNotApplicableSameNameVariableWithCopy() throws Exception {
...
-        @TestMetadata("notApplicableSameNameVariableWithHashCode.kt")
-        public void testNotApplicableSameNameVariableWithHashCode() throws Exception {
-            runTest("testData/inspectionsLocal/removeRedundantQualifierName/notApplicableSameNameVariableWithHashCode.kt");
-        }
-
-        @TestMetadata("notApplicableSameNameVariableWithToString.kt")
-        public void testNotApplicableSameNameVariableWithToString() throws Exception {
-            runTest("testData/inspectionsLocal/removeRedundantQualifierName/notApplicableSameNameVariableWithToString.kt");
-        }
-
```
t-kameyama: Removed as not needed

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantQualifierNameInspection.kt
```diff
@@ -153,6 +155,10 @@ private fun KtDotQualifiedExpression.applicableExpression(
     val newDescriptorFqName = newDescriptor.unwrapFakeOverrideIfNecessary().fqNameSafe
     if (originalDescriptorFqName != newDescriptorFqName) return null
 
+    if (newExpression is KtQualifiedExpression && newExpression.receiverExpression.declarationDescriptor(newContext) != receiverReference) {
```
dimonchik0036, line 158: They should be checked by `compareDescriptors`. 
Please add a valid test on it
```kotlin
package m2.second

import m2.second.Bar.foo

data class Foo(val name: String)

object Bar {
  val foo = Foo("foo")
}

object Test {
  val boo = Bar.foo.copy(name = "foo2")
}
```

dimonchik0036 (changes requested):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantQualifierNameInspection.kt
t-kameyama: `compareDescriptors` does not return the expected result because `compareDescriptorsText` returns true here.
https://github.com/JetBrains/intellij-community/blob/a29231587f989fb49f86af5246c9173413653088/plugins/kotlin/base/fe10/analysis/src/org/jetbrains/kotlin/idea/core/descriptorUtils.kt#L73

t-kameyama: Should we check `containingDeclaration`?

```kotlin
if (newExpression is KtQualifiedExpression &&
    !compareDescriptors(
        project,
        newExpression.receiverExpression.declarationDescriptor(newContext)?.containingDeclaration,
        receiverReference?.containingDeclaration
    )
) return null
```

t-kameyama: [150ebeb](https://github.com/JetBrains/intellij-community/pull/2120/commits/150ebeb4fea33ea84b5fbe8fc17c4ca730b0d0ed)

dimonchik0036 (approved):
Sorry for the long delay

PR #2123, opened 2022-08-25
[java-decompiler] DeadCodeHelper mergeBasicBlocks method (avoid dead loop )

asmjmp0:
avoid dead loop when handle a special class file
[e.class.zip](https://github.com/JetBrains/intellij-community/files/9426605/e.class.zip)
.

### DIFF
--- a/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/code/DeadCodeHelper.java
+++ b/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/code/DeadCodeHelper.java
@@ -529,7 +529,7 @@ public final class DeadCodeHelper {
     while (true) {
 
       boolean merged = false;
-
+      int originBlocksCount = graph.getBlocks().size();
       for (BasicBlock block : graph.getBlocks()) {
 
         InstructionSequence seq = block.getSeq();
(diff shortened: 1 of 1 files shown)


plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/code/DeadCodeHelper.java
```diff
@@ -565,7 +565,9 @@ public static void mergeBasicBlocks(ControlFlowGraph graph) {
           }
         }
       }
-
+      if(graph.getBlocks().size() == originBlocksCount && merged){
+        break;
+      }
       if (!merged) {
```
trespasserw, line 571: `!merged || graph.getBlocks().size() == originBlocksCount` would convey the idea better IMO.

trespasserw (changes requested):

plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/code/DeadCodeHelper.java
asmjmp0: I have changed the statement as you said.

trespasserw (approved):

trespasserw:
Merged, thanks! The commit should appear on GitHub within an hour.

PR #2124, opened 2022-08-26
IDEA-298870: fix "outdated stub in index" in tests

gharrma:
See https://youtrack.jetbrains.com/issue/IDEA-298870 for details.

This change helps fix flaky test failures that we are experiencing in Android Studio.


batya239:
Already merged

PR #2125, opened 2022-08-26
IDEA-298773: Fix memory leak in GotoSymbolModel2

npaige:
Use SymbolSearchEverywhereContributor instead of Project for the parent
disposable for the extension point listeners.


yole:

PR #2126, opened 2022-08-27
KTIJ-22610 "Generate equals() and hashCode()" for value classes produced red code

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22610


vladimirdolzhenko:
Looks good, I would recommend to make a follow up PR to prohibit overriding equals/hashCode methods for value classes.

PR #2127, opened 2022-08-27
IDEA-186628 - Provide Maven Tycho support (no POM-less)

lppedd:
This PR aims at providing a basic - but multimodule - Maven Tycho support.  
Before it can be considered worth reviewing what must be done is:
- restore `plugins/maven/maven36-server-impl/src/org/slf4j/impl/StaticLoggerBinder.java`
- restore `plugins/maven/maven36-server-impl/src/org/jetbrains/idea/maven/server/Maven3WrapperSl4LoggerFactory.java`
- cleanup `Maven3XServerEmbedder#doResolveProject`

Why do we have two deleted files? See https://youtrack.jetbrains.com/issue/IDEA-186628/Tycho-support-is-broken#focus=Comments-27-6412160.0-0

Also, a good addition would be improving the MANIFEST scanning code. I've quickly come up with a recursive solution but I don't like it. And maybe check if code duplication can be reduced (but I think it's worth it, Tycho might require additional changes in the future).


plugins/maven/maven3-server-impl/src/org/jetbrains/idea/maven/server/Maven3XServerEmbedder.java
```diff
@@ -797,6 +807,10 @@ public Collection<MavenExecutionResult> doResolveProject(@NotNull final Collecti
 
     request.setUpdateSnapshots(myAlwaysUpdateSnapshots);
 
+    if (myServerSettings.isTychoProject()) {
+      request.setDegreeOfConcurrency(2);
```
lppedd, line 811: This can be increased up to 4, but file access doesn't seem to be thread safe.

plugins/maven/src/main/java/org/jetbrains/idea/maven/project/MavenProjectResolver.java
```diff
@@ -97,6 +111,150 @@ public void resolve(@NotNull Project project,
...
+
+      final String exportedPackagesStr = manifestAttributes.getValue("Export-Package");
+
+      if (exportedPackagesStr != null) {
+        for (final String packageName : exportedPackagesStr.split(",")) {
+          exportedPackages.add(packageName.trim());
+        }
+      }
+    } catch (final IOException e) {
+      throw new RuntimeException(e);
```
lppedd, line 221: This must be replaced with a decent exception + message.

plugins/maven/src/main/java/org/jetbrains/idea/maven/project/MavenProjectResolver.java
```diff
@@ -97,6 +111,150 @@ public void resolve(@NotNull Project project,
...
+    }
+
+    return new MavenProjectManifestData(requiredBundles, importedPackages, exportedPackages);
+  }
+
+  private static void collectRequiredMavenProjects(
```
lppedd, line 227: This method uses recursion. It's fine for simplicity but it should be improved.
If by accident you have a circular dependency in MANIFEST files, you end up with a stack overflow.

plugins/maven/maven3-server-impl/src/org/jetbrains/idea/maven/server/Maven3XServerEmbedder.java
nskvortsov: this should not be tycho-specific

nskvortsov:
You did a great job of investigating Maven Tycho!

Maven server embedder was not designed with extensibility in mind. Do you think it would be possible to do all the customization of the resolved Maven projects subsets inside IDEA process (e.g., in `MavenProjectResolver.java` )?
Having a separate plugin would allow improving Tycho support outside IDEA release cycles.

Speaking of hacking the `MavenProjectResolver.java`. As @zulkar points out [here](https://youtrack.jetbrains.com/issue/IDEA-186628/Tycho-support-is-broken#focus=Comments-27-6418924.0-0), it is better to keep tycho-specific logic in a plugin. I suggest you add a new extension point in MavenProjectResolver that will allow a plugin author to customise the set of resolved maven projects.

WRT to deleted files. They should be kept intact for IntelliJ to be able to pick up Maven logging as it does now. It looks like a problem with some Tycho dependencies. Maybe detecting and excluding the colliding slf4j implementation during dependency resolution will help?

lppedd:
Hi @nskvortsov, and thanks!
I'll try to answer your question based and what I was able to undertstand.

> Do you think it would be possible to do all the customization in `MavenProjectResolver`?

I don't think so. As you could see I had to change `Maven3XServerEmbedder` a bit.  
The Tycho extension looks for `MavenSession#getProjects` only, so internally the IJ server implementation has to have a different behavior compared to the classic Maven.

Now, `Maven3XServerEmbedder#doResolveProject` only receive a list of `files`, representing the POMs. How can we describe to this method/class the different behavior to apply, in a generic way?

> I suggest you add a new extension point in MavenProjectResolver that will allow a plugin author to customise the set of resolved maven projects

This one is feasible and I could do it. But then I think on your part (and I say that because you know the code better than me) you should take some time to understand what the extension point should cover. There are multiple places that could take advantage of this new EP imho. An example is the different repository `layout` indexer, for `p2`. This now seems fixed to the standard Maven Nexus one. I had a look at the indexing process and it's pretty complex unfortunately.

> Maybe detecting and excluding the colliding slf4j implementation during dependency resolution

Might me, but speaking in technical terms, how do you imagine this implemented?

nskvortsov:
@lppedd I think we should look for a way to modify `Maven3XServerEmebedder` in such a way, that it knows nothing particular about Tycho project nature. AFAIU we need to change the `loadExtensions` behavior. That is ok. But can we change it so it works for both Tycho and "non-tycho" projects?

Indexer can also be made extendable. It does not have to share extension point with `MavenProjectResolver`, so let's concentrate on sync (import) here.

> but speaking in technical terms, how do you imagine this implemented?

Given there is a Tycho IJ plugin built with Gradle, the build script can filter out rogue transitive dependency (the best case) from the plugin distribution. Or repack this dependency, removing extra logging backend (worst case).

plugins/maven/src/main/java/org/jetbrains/idea/maven/project/MavenGeneralSettingsWatcher.kt
```diff
@@ -49,6 +54,16 @@ class MavenGeneralSettingsWatcher private constructor(
     }, parentDisposable)
   }
 
+  private fun updateImportingSettings() {
+    if (generalSettings.isTychoProject) {
+      val importingSettings = manager.importingSettings
+      val dependencyTypes = importingSettings.dependencyTypesAsSet
+      dependencyTypes.add(MavenTychoConstants.PACKAGING_ECLIPSE_PLUGIN)
+      dependencyTypes.add(MavenTychoConstants.PACKAGING_ECLIPSE_TEST_PLUGIN)
```
bric3, line 62: Why not adding the other types (in particular `eclipse-feature` or `eclipse-target-definition`) ?

plugins/maven/maven3-server-impl/src/org/jetbrains/idea/maven/server/Maven3XServerEmbedder.java
```diff
@@ -137,6 +140,15 @@ public abstract class Maven3XServerEmbedder extends Maven3ServerEmbedder {
...
   @NotNull private final MavenImporterSpy myImporterSpy;
 
+  private static final Set<String> TYCHO_BANNED_PACKAGING = new HashSet<>(Arrays.asList(
+    MavenTychoConstants.PACKAGING_ECLIPSE_FEATURE,
+    MavenTychoConstants.PACKAGING_ECLIPSE_REPOSITORY,
+    MavenTychoConstants.PACKAGING_ECLIPSE_APPLICATION,
+    MavenTychoConstants.PACKAGING_ECLIPSE_UPDATE_SITE,
+    MavenTychoConstants.PACKAGING_ECLIPSE_TARGET_DEFINITION,
+    MavenTychoConstants.PACKAGING_P2_INSTALLABLE_UNIT
+  ));
```
bric3, line 150: Why banning those?
The `eclipse_target_definition` packaging allows to define platform wide dependencies (that can be then resolved from the OSGI entries in the `MANIFEST.MF` file).

bric3 (commented):
Just noted a few things about other tycho packaging.

plugins/maven/src/main/java/org/jetbrains/idea/maven/project/MavenGeneralSettingsWatcher.kt
lppedd: Eclipse Features or Target definitions should not contain any code, they're just grouping things.
Maven does not need to scan and know about them. The real stuff happens for Plugins only.
I may be wrong tho 👀

plugins/maven/maven3-server-impl/src/org/jetbrains/idea/maven/server/Maven3XServerEmbedder.java
lppedd: This one might actually be related to how I'm using Tycho.
In my case the dependencies are simply pulled in by p2 repositories, I have no Target definition.

plugins/maven/src/main/java/org/jetbrains/idea/maven/project/MavenGeneralSettingsWatcher.kt
bric3: That is true even for other packaging, like `eclipse-repository` but then it's also true for _regular_ maven packaging like `pom` (used for _bill-of-material_ POMs).

I'm don't know how IJ handles that as POMs are not listed in the default _dependency types_ list in maven settings, but it makes sense they should be imported, at least internally ?

lppedd: @bric3 it could make sense, yeah. Main reason I excluded them is because of resolve time. Including all features and other packaging types is probably going to take much more time for the entire workspace to be resolved.

lppedd:
@nskvortsov Hi! I've returned to working a bit on this.
Before going any further, should I worry about the new workspace model? What will happen in the future?

nskvortsov:
@lppedd  You should not worry. The current API is far from being deprecated. The new workspace model is already used as underlying implementation, and now we started using it directly in some places.

lppedd:
Sad news everyone. Doesn't work anymore for Tycho 4. I'm investigating.

lppedd:
`P2ResolverFactoryImpl`: seems like an old version of the service is injected.  
Here it's trying to invoke `getService(Class)`, but we only have `getService(String)`.

image

lppedd:
@nskvortsov I have noticed IJ doesn't support toolchains when build the Maven request.  
I'm adding it quickly but would love to see it done in the main branch.

lppedd:
Proposal: instead of relying only on the client code of IntelliJ, can we open `Maven3XServerEmbedder` to plugins?
A plugin could contribute it's own extension of `Maven3XServerEmbedder`. The important thing about external contribution is they should be able to control:
1. Maven dependencies (e.g. the provided version of Maven Core/Compat)
2. How the Maven context is initialized

lppedd:
FYI I'm implementing a facade over Maven (using maven-embedded). This is going to be a pluggable system that have clear separation between Maven versions. I will then use this system to provide enhanced features for Tycho.

lppedd:
So a small good news. I've introduced a new EP at the IJ resolver level (not in the embedder) to customize the to-be-resolved projects.

```java
public interface MavenResolverCustomizer {
  boolean isEnabled();

  @NotNull
  Collection<MavenProject> customizeProjects(
    @NotNull final Collection<MavenProject> projects,
    @NotNull final MavenProgressIndicator indicator);
}
```

On the embedder level I removed every Tycho reference, and instead I always set `MavenSession#setAllProject` with the actual projects list before calling the lifecycle listeners, and not only with `Collections#singletonList`.
The lifecycle listeners get now called a single time, not once per project.

@nskvortsov do you happen to know why it was done like that?

nskvortsov:
@lppedd the current loop with singletone lists looks like a nine-years-old suboptimal implementation. Can be improved.

lppedd:
@nskvortsov I don't have much time to work on this now. But that piece of code isn't there anymore in the implementation I haven't pushed yet.

nskvortsov:
This draft is abandoned. Please feel free to pick it up and refine.

PR #2130, opened 2022-08-30
[KTIJ-11296] Code style, group declarations: align properties/variables/methods in columns

i582:
## Details

This PR adds formatting for properties, methods, functions, local functions and variables (local and global) for Kotlin.

Symbols are aligned block by block, that is, only those that follow each other without empty lines.

### Block

An inextricable sequence of the same type **single-line** elements following one after another without empty lines.

Example:

```kotlin
val long:  Int     = 100
val short: String  = "200"
val tiny:  Boolean = true
```

If the function is written in several lines, then the group ends, and the `null` alignment is returned for the function, since IDEA automatically inserts an empty line before it.

If the property is written in several lines, then the behavior is identical (except new line before by IDEA).

#### In block alignment

Alignment in a block can occur across multiple tokens. If only one of the tokens is present in a block element, then it is aligned with tokens of the same type in other blocks.

Example:

```kotlin
fun long(): String = ""
fun short()        = 3.5
```

The second function does not have a `TYPE_REFERENCE` token, so the alignment is on the `=` token.

## Settings page

Added new "Group declarations" group to "Wrapping and Braces":

Снимок экрана 2022-09-01 в 17 36 52

Fixes [KTIJ-11296](https://youtrack.jetbrains.com/issue/KTIJ-11296)
Fixes [KTIJ-19544](https://youtrack.jetbrains.com/issue/KTIJ-19544)
Fixes [KTIJ-22634](https://youtrack.jetbrains.com/issue/KTIJ-22634)


MihaelBercic:
This would be a lovely feature.

PR #2132, opened 2022-08-30
IDEA-309689 & IDEA-240111 - Configurable disabling of up-to-date checks for all test tasks

aSemy:
https://youtrack.jetbrains.com/issue/IDEA-309689

Partial fix for https://youtrack.jetbrains.com/issue/IDEA-240111

This will allow users to disable the current IntelliJ behaviour which breaks Gradle incremental task support.

This would be helpful in testing a more complete solution: https://github.com/aSemy/intellij-gradle-init-plugin/

Also rename testFilterInit.gradle to testFilter.init.gradle (better IDE support)


nskvortsov:
[IDEA-240111](https://youtrack.jetbrains.com/issue/IDEA-240111) is fixed properly. Thank you for your proposal.

PR #2135, opened 2022-09-01
Consider all compiler args when determining --preview status

j-baker:
IntelliJ discovers whether a Gradle project should have preview features
enabled by checking if --enable-preview is added to the Java compile
arguments.

In Gradle 4.5, a new mechanism of adding compiler arguments was added in
the CommandLineArgumentProvider interface. This also changes the actual
accessor to be 'allCompilerArgs' rather than 'compilerArgs'.

Right now, if a CommandLineArgumentProvider adds --enable-preview (this
is done [here](https://github.com/palantir/gradle-baseline) then
IntelliJ will not discover this, leading to preview features not being
enabled.

This PR fixes the problem in a way that doesn't affect IntelliJ's gradle
compatibility (by checking if the new approach is possible and if not,
using the old one) and modifies the tests.


plugins/gradle/java/testSources/importing/GradleJavaCompilerSettingsImportingTestCase.kt
```diff
@@ -48,18 +52,27 @@ abstract class GradleJavaCompilerSettingsImportingTestCase : GradleJavaImporting
           assignIfNotNull("sourceCompatibility", mainSourceCompatibility)
           assignIfNotNull("targetCompatibility", mainTargetCompatibility)
           if (mainSourceCompatibilityEnablePreview) {
-            call("options.compilerArgs.add", "--enable-preview")
+            if (isNotSupportedCompilerArgumentProviders) {
+              call("options.compilerArgs.add", "--enable-preview")
+            } else {
+              code("options.compilerArgumentProviders.add({ ['--enable-preview'] } as CommandLineArgumentProvider)")
```
j-baker, line 58: didn't really know a better way to do this

HackerMadCat:
Hi! Thank you for your contribution. I apologise for the delayed response. I retested your original issue, and it appears to be fixed by the fix for [IDEA-154038](https://youtrack.jetbrains.com/issue/IDEA-154038).

The test that checks the Java arguments provider.
https://github.com/JetBrains/intellij-community/blob/2e7385cd2b6c8d2b3f2f1dc264c6a2ece3cfa3ce/plugins/gradle/java/testSources/importing/GradleMiscImportingTest.java#L200

PR #2138, opened 2022-09-02
[gradle] Set apt path for sub project processor dependency correctly

fp7:
Note: This does not do dependency substitution

Issue: IDEA-293518


nskvortsov:
please rebase if the issue is still relevant.

PR #2139, opened 2022-09-03
[kotlin] Don't report "Control flow with empty body" when 'if' is used as a expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22627


abelkov (changes requested):
I think that an empty body which implicitly returns `Unit` looks confusing. It's not obvious. It's better to leave the explicit `Unit`. So, I propose that "Remove redundant Unit" inspection should not be reported in such cases.

abelkov (approved):

vanniktech:
This works really well in EAP 1, thanks for doing this!

PR #2141, opened 2022-09-04
KTIJ-22656 Introduce parameter fails when the last parameter without comma ends with a comment

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22656

```kotlin
class A(
    val a: String // test
    ) {
    fun main() {
        "".length
    }
}
```

スクリーンショット 2022-09-04 20 40 08

↓

スクリーンショット 2022-09-04 20 40 16


abelkov (changes requested):
Please support the case with trailing comma so the comment stays on the same line:

```
class A(
    val a: String, // test
) {
    fun main() {
        "".length
    }
}
```

Also with a block comment the comma is in a different place, maybe worth to fix as well:

```
class A(
    val a: String /* test */
) {
    fun main() {
        "".length
    }
}
```

abelkov (approved):

PR #2142, opened 2022-09-05
KTIJ-10025 Don't suggest adding "var" keyword to a value class's parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-10025


abelkov (approved):

t-kameyama:
Sorry for the almost meaningless PR.

abelkov:
Hopefully in https://github.com/JetBrains/intellij-community/commit/e43f11693feb78683f0f6f7699bd0e343ca176da I have unified all checks of `inline` and `value` classes, so behavior should be consistent now.

PR #2144, opened 2022-09-06
[IDEA-301271] Make it possible to ignore KDoc for override elements.

duckladydinh:
Override elements usually have KDoc defined by their parent declarations. Therefore they don't need to re-declare KDoc. However, KDocMissingDocumentationInspection doesn't always work as intended when we are overriding a method from an interface defined in some library and sources are not available. This typically happens in Bazel plugin and monorepo for example. Therefore we should allow an option to explicitly disable the KDocMissingDocumentationInspection for override elements with a checkbox.

This would give users more customizations over their IDE and make Kotlin easier to use.

You Track: https://youtrack.jetbrains.com/issue/IDEA-301271/Make-it-possible-to-ignore-KDoc-for-override-elements

### DIFF
--- a/plugins/kotlin/base/resources/resources-en/messages/KotlinBundle.properties
+++ b/plugins/kotlin/base/resources/resources-en/messages/KotlinBundle.properties
@@ -1264,6 +1264,7 @@ replace.with.kotlin.analog.function.text=Replace with ''{0}'' function
 add.documentation.fix.text=Add documentation
 missing.documentation=Missing documentation
 0.is.missing.documentation={0} is missing documentation
+ignore.override.elements=Ignore override elements
 it.s.prohibited.to.call.0.with.min.value.step.since.1.3=It''s prohibited to call {0} with MIN_VALUE step since 1.3
 obsolete.coroutine.usage.in.whole.fix.family.name=Fix experimental coroutines usages in the project
 obsolete.kotlin.js.packages.usage.in.whole.fix.family.name=Fix 'kotlin.dom' and 'kotlin.browser' packages usages in the project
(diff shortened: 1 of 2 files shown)


abelkov:
Note: there is no such option in Java "missing Javadoc" inspection too.

duckladydinh:
Hi @abelkov, I can't really talk about Java because I have never seen this problems from Java but some thoughts
* Maybe Java already handles it in a different way?
* Maybe Java can read javadoc from other classes better than Kotlin?
* Maybe Java has the same problem but somehow we just didn't see because we load everything in the class path?

In my case, I used Bazel plugin (https://plugins.jetbrains.com/plugin/8609-bazel) and the project is extremely huge. And some sources couldn't be imported in the project => the inspection couldn't find KDoc (in the interface in the dependencies) somehow => it keeps saying that it is missing documentation.

Regardless, this will just add a customization option, so the default behavior won't change.

nikitabobko:
@duckladydinh if I got it correctly some of your dependencies are missing "-sources.jar". Couldn't the issue be resolved by configuring your Bazel build so the sources would be included?

If you use `rules_jvm_external` then all you need is to add `fetch_sources = True` https://github.com/bazelbuild/rules_jvm_external#fetch-source-jars 

Or is it a type of dependency that doesn't publish "-sources.jar" at all?

> Regardless, this will just add a customization option, so the default behavior won't change.

The option you're adding is too specific. I don't think that it covers a lot of use cases. A better option though would be to add a "Ignore elements that override/implement external library members" checkbox.

duckladydinh:
@nikitabobko I don't think it works that way. I do have all the sources available locally. However, in a big monorepo, Bazel allows lazy loading. That means, you can specify a number of directories to load into your workspace. Loading all dependencies would defeat the purpose of lazy loading.

> The option you're adding is too specific. I don't think that it covers a lot of use cases. A better option though would be to add a "Ignore elements that override/implement external library members" checkbox.

That option is the way the current inspection works. If you have the interface and the implementation in your workspace, regardless of whether you have KDoc in the interface or not, there will be no warning in the implementation. The problem will only occur when the parent class is not available.

nikitabobko:
> If you have the interface and the implementation in your workspace, regardless of whether you have KDoc in the interface or not, there will be no warning in the implementation

That's not true:
image

What I mean is that we can add a checkbox (or even make it a default behavior) that if the parent class "is not available" (The definition of "is not available" is up for the inspection implementation) then we don't highlight the implementation. That sounds more generic and fair rather than not highlighting all entries that are marked with `override`, and it would cover your case

nikitabobko:
Close because of lack of activity. Feel free to open new PR

PR #2147, opened 2022-09-08
Add support for Kotlin value container assignment plugin

asodja:
This adds support for Kotlin value container assignment plugin to IntelliJ.

I see that for sam-with-receiver there are some additional things in `intellij.android.*`, but I didn't manage to integrate that for value-container-assignment yet.

/cc @demiurg906


plugins/kotlin/compiler-plugins/value-container-assignment/common/src/org/jetbrains/kotlin/idea/compilerPlugin/valueContainerAssignment/IdeValueContainerAssignmentContainerContributor.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    override fun registerModuleComponents(
+        container: StorageComponentContainer,
+        platform: TargetPlatform,
+        moduleDescriptor: ModuleDescriptor
+    ) {
+        val annotations =
+            when (val moduleInfo = moduleDescriptor.getCapability(ModuleInfo.Capability)) {
+                is ScriptModuleInfo -> moduleInfo.scriptDefinition.annotationsForSamWithReceivers
```
asodja, line 54: This needs a fix, we should not read `annotationsForSamWithReceivers`

asodja:
Replaced by https://github.com/JetBrains/intellij-community/pull/2180.

PR #2149, opened 2022-09-09
[KTIJ-19314] Debugger: Account for synthetic names of captured locals

kandersen:
Small fix to a long-standing issue in the debugger caused by the move to compile local functions to static functions instead of lambdas by the IR compiler.

Resolved by simply also looking for the equivalently named synthetic variable in the local scope.


kandersen:
CC @nikita-nazarov @madsager

PR #2151, opened 2022-09-09
[IDEA-301618] Indicate if an escaped character is in a class

andymagee:
Minor fix addressing [IDEA-301618](https://youtrack.jetbrains.com/issue/IDEA-301618).

A new method overload has been added for RegExpLanguageHost.characterNeedsEscaping(), allowing the implementation to know whether the given character is occurring within a class or not. I've added the overload with a default implementation and marked the existing method as deprecated, so that this will not be a breaking change and existing logic will continue to operate as it currently does.

### DIFF
--- a/RegExpSupport/src/org/intellij/lang/regexp/RegExpLanguageHost.java
+++ b/RegExpSupport/src/org/intellij/lang/regexp/RegExpLanguageHost.java
@@ -14,7 +14,21 @@ public interface RegExpLanguageHost {
   EnumSet<RegExpGroup.Type> EMPTY_NAMED_GROUP_TYPES = EnumSet.noneOf(RegExpGroup.Type.class);
   String[][] EMPTY_COMPLETION_ITEMS_ARRAY = new String[0][];
 
+  /**
+   * @deprecated Use {@link #characterNeedsEscaping(char, boolean)} instead.
+   */
+  @Deprecated
   boolean characterNeedsEscaping(char c);
+
+  /**
+   * Returns whether the given character needs to be escaped to be treated as a literal.
+   * @param c a character to be considered.
+   * @param isInClass whether the character is within a RegExpClass (ie, within "[...]").
+   */
+  default boolean characterNeedsEscaping(char c, boolean isInClass) {
+    return characterNeedsEscaping(c);
+  }
+
   boolean supportsPerl5EmbeddedComments();
   boolean supportsPossessiveQuantifiers();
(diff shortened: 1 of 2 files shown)


andymagee:
@BasLeijdekkers and @andrei-kuznetsov - friendly ping to ask if someone can take a look at this PR. I'm with the Google team working on Android Studio, and this will help us address [IDEA-301618](https://youtrack.jetbrains.com/issue/IDEA-301618) and [AS bug](https://issuetracker.google.com/issues/111445868).

BasLeijdekkers:
Hi @andymagee, 
Do I understand correctly that the ICU regex dialect has very few differences with the Java regex syntax? And that you are not creating a separate regex dialect for it (i.e. creating your own RegExpParserDefinition)? I ask because if you have your own regex dialect, there are better ways to get the desired result.

BasLeijdekkers:
Merged manually [here](https://github.com/JetBrains/intellij-community/commit/a89ecd6615488f2ad151eb978a8f898eb788e5fe)

Thanks!

RegExpSupport/src/org/intellij/lang/regexp/RegExpLanguageHost.java
```diff
@@ -14,7 +14,21 @@ public interface RegExpLanguageHost {
   EnumSet<RegExpGroup.Type> EMPTY_NAMED_GROUP_TYPES = EnumSet.noneOf(RegExpGroup.Type.class);
   String[][] EMPTY_COMPLETION_ITEMS_ARRAY = new String[0][];
 
+  /**
+   * @deprecated Use {@link #characterNeedsEscaping(char, boolean)} instead.
+   */
+  @Deprecated
```
hurricup, line 20: I presume default implementation should be added here. Otherwise I can't get rid of implementing deprecated method, which feels wrong. 

cc @BasLeijdekkers

andymagee: My two cents (and the reason I wrote it this way): adding the new method without a default implementation would amount to a breaking change, and anyone downstream who pulled the latest changes would immediately get a build breakage.

I suppose both methods could be given empty default implementations that just return false, which might serve both purposes.

I'm not fully aware on guidelines around making breaking changes, so don't have a specific recommendation. Just thought I'd add that context.

hurricup: Not sure I got your idea.
My point is that you should have add default implementation for deprecated method with AbstractMethodException inside.

PR #2152, opened 2022-09-10
KTIJ-22712 Inspection: use appendRange instead of append

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22712


abelkov (approved):
Thank you!

PR #2154, opened 2022-09-12
KTIJ-22737 "'StringBuilder.append(CharArray, offset, len)' call on the JVM": add a non-null assertion for nullable CharArray

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22737


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithStringBuilderAppendRangeInspection.kt
```diff
@@ -74,11 +85,25 @@ class ReplaceWithStringBuilderAppendRangeInspection : AbstractKotlinInspection()
...
+                firstArg.replace(psiFactory.createExpressionByPattern("$0!!", firstArg))
+            }
         }
 
         private fun KtPsiFactory.createCalleeExpression(functionName: String): KtExpression =
             (createExpression("$functionName()") as KtCallExpression).calleeExpression!!
 
         private fun KtExpression.toIntOrNull(): Int? = safeAs<KtConstantExpression>()?.text?.toIntOrNull()
+
+        private fun KtExpression.isNotNullType(): Boolean {
```
abelkov, line 99: I wonder does something like this function already exist in the codebase so we don't have to write it by hand? Seems like it should be pretty common functionality. (I haven't researched this).

abelkov (approved):
Thank you, now it is definitely cleaner.

PR #2155, opened 2022-09-12
KTIJ-16332 IDE: Introduce quick fix "Add fun modifier to interface declaration" for RESOLUTION_TO_CLASSIFIER error when helpful

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-16332


abelkov:
Hi, sorry for the delay. The reason is that I want to become familiar with all the new APIs we started to use to port inspections/intentions/quickfixes to K2 Kotlin plugin. For example, see all commits about "K2 inspection".

In the future, we want to use the new APIs to share implementation between K1 and K2 plugins as much as possible. Some inspections are already shared successfully, see `org.jetbrains.kotlin.idea.codeInsight.inspections.shared`.

So, I want to see if this PR can be ported to the new API.

abelkov (approved):
Thank you, I merged the PR. From now on, please don't add new K1 inspections, intentions and quickfixes until we port all existing inspections to K2. Bugfixes are still OK.

PR #2159, opened 2022-09-18
[IDEA-301967] Unified diff viewer - cannot click Revert icon immediately after reverting a previous change

chylex:
Please see https://youtrack.jetbrains.com/issue/IDEA-301967 for a video of the problem. The comments then explain why this issue happens, and from them it should be clear why I made these changes in the PR.

The `clickCount` condition was introduced in c3e755b7699eeb61272c0a6666e92b27c052c255 which is about right-clicking, not about the number of clicks, so it seems it was added reduntantly and should be safe to remove.

I would not say that the way I reset the consecutive mouse event flag is the best solution here, so if you let me know how you'd like to code that better, I'll be happy to edit the PR.


AMPivovarov:
Thank you for the investigation.

The `DiffGutterRenderer` part looks good and is merged as-is, 5b757cf6be6742e24f8466588dc7055c679cfd4f.
I've omitted the `doNotIgnoreNextConsecutiveMouseEvent` part and tried to apply a more general fix to the `myIgnoreMouseEventsConsecutiveToInitial` problem, see c89b6822462c3ea37d8d6bcab343cdd31f3fcf79. This should fix the issue, but please come back if that's not the case.

chylex:
Tried latest commit and it seems to be fine, thanks!

PR #2163, opened 2022-09-20
[fernflower] Use Paths.get() for JDK-7 compatibility

PranavPurwar:


trespasserw:
Doesn't make sense, sorry. The code already uses Java 17 language features.

PranavPurwar:
The code uses Java 17, but we should maintain JRE-8 compatibility

trespasserw:
What for?

The sources using Java 17 language features cannot be directly compiled to JRE 8 bytecode (well, there are retrofitters, but I'm not sure about their quality).

PranavPurwar:
I think you don't already know about language feature desugaring yet.

trespasserw:
I think you don't already know that desugaring tools are commonly called retrofitters :)

Seriously so, could you please provide at least one compelling reason why JetBrains should maintain the decompiler code compatible with JRE 8?

trespasserw:
You see, maintaining the compatibility requires some efforts. Accepting one particular pull request is easy, but to assure no one else changes the code again one has to set up an automated check. Do you know any such tools, per chance?

BTW, JRE 8 is pretty old already and doesn't seem to be as dominant as, say, in 2020.

PranavPurwar:
i guess running tests with Java 8 would be enough

PR #2167, opened 2022-09-23
IDEA-302320 Register Invoker in Disposer after class if fully constructed

eugen-yakovets:
Expose Invoker class to Disposer only after object is fully constructed.


batya239:
Screenshot 2025-06-16 at 10 43 48
Already applied

PR #2172, opened 2022-09-29
KTIJ-22542 False positive 'Property could be private' when another property uses it and exposes its local argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-22542


vladimirdolzhenko:
sorry for missing this PR, 
we're mostly focusing on K2 implementation.

t-kameyama:

PR #2174, opened 2022-10-01
KTIJ-23080 Redundant "Add braces to all 'if' statements" when the only one left

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23080


vladimirdolzhenko:

PR #2183, opened 2022-10-05
Fix handling of not updated VFS in coverage plugin

yuriy-solodkyy:
The Coverage plugin used to expect build output directories to be present in the virtual file system as a result of building a project. However, even making sure that build output roots are present in the VFS is an expensive operation on large projects. This is despite the coverage plugin using Java IO to deal with .class files.

Use `getCompilerOutputUrl()` instead of `getCompilerOutput()` and make sure the virtual file system is refreshed when it is really needed. This avoids refreshing on build completion.


yuriy-solodkyy:
@nskvortsov @andrei-kuznetsov

nskvortsov:
Looks great! Will run CI/CD checks now.

yuriy-solodkyy:
@nskvortsov Did it pass tests?

nskvortsov:
@yuriy-solodkyy yes, landed in master as https://github.com/JetBrains/intellij-community/commit/daa5117a659028e27b262c80c09a5dc21b530df7

PR #2184, opened 2022-10-05
[IDEA-300359] Fix default and selected checkbox bg theming in dark mode

kdrag0n:
Fixes https://youtrack.jetbrains.com/issue/IDEA-300359.

In custom themes derived from Darcula, changing the default checkbox background color (Checkbox.Background.Default.Dark) and selected color (Checkbox.Background.Selected.Dark) to different colors doesn't work. The Selected color is used in both cases because they map to the same color in Darcula (#43494a), and the Checkbox keys are just aliases for icon color replacements using SvgElementColorPatcherProvider. Since they map to the same color, one of them gets overwritten in the icon color replacement map.
https://github.com/JetBrains/intellij-community/blob/1470ece865ba879c0c5074b4cb7784858b1b35a1/platform/platform-impl/src/com/intellij/ide/ui/UITheme.java#L286-L300

Simple reproducer (test.theme.json):

```json
{
  "name": "Test",
  "dark": true,
  "author": "Test",
  "ui": {},
  "icons": {
    "ColorPalette": {
      "Checkbox.Background.Default": "#0000ff",
      "Checkbox.Background.Default.Dark": "#0000ff",
      "Checkbox.Background.Selected": "#ff0000",
      "Checkbox.Background.Selected.Dark": "#ff0000"
    }
  }
}
```

This issue doesn't occur in light mode because the default light colors are different.

This commit implements an easy workaround: change one of the colors by an imperceptible amount, e.g. #43494b, so the color replacement map works correctly with no need for more intrusive changes.


Benson665:
I like the background colour is nice and sharp to see the writing finger work

JulianSchoenbaechler:
Same with: `Checkbox.Border.Default.Dark`

What's the current status on this? Is there a better way to handle this (maybe less "hacky")?
It surprises me that there are not more reports for this issue, as this bug is visually well noticeable (e.g. in the official _Dark_ theme in the new UI) and existed now for quite some time. Checkboxes just look wrong in custom themes.

Is there maybe an alternative way of styling the elements?

PR #2185, opened 2022-10-06
KTIJ-23097 "Move Statement down" breaks out of scope when skipping over comment line

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23097

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinExpressionMover.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinExpressionMover.kt
@@ -445,6 +445,7 @@ class KotlinExpressionMover : AbstractKotlinUpDownMover() {
             editor: Editor, down: Boolean
         ): Pair<PsiElement, PsiElement>? {
             if (down && sibling.siblings().none { it is KtExpression }) return null
+            if (!down && sibling !is KtExpression) return null
 
             var currentStart = start
             var currentEnd = end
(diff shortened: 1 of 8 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinExpressionMover.kt
```diff
@@ -503,7 +506,7 @@ class KotlinExpressionMover : AbstractKotlinUpDownMover() {
 
             val movableElement = element.getParentOfTypesAndPredicate(
                 strict = false,
-                parentClasses = *MOVABLE_ELEMENT_CLASSES,
```
t-kameyama, line 506: Redundant spread operator

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinExpressionMover.kt
```diff
@@ -503,7 +506,7 @@ class KotlinExpressionMover : AbstractKotlinUpDownMover() {
...
+                parentClasses = MOVABLE_ELEMENT_CLASSES,
```
t-kameyama, line 509: Redundant spread operator

abelkov:
Can you please support a related case. On "move up", when there are multiple comment lines above, Java skips over exactly one line. But Kotlin skips all comments. I think Java behavior is better and more consistent with "move down".

```
public class JJJ {
        // Comment
        // Comment
        String s = "";
        // Comment
    }
}
```

vanniktech:
@abelkov 100% agreed. I also encountered this yesterday but I haven't had the time to comment this. (I'm the one who created the Youtrack issue)

t-kameyama:
> Can you please support a related case. On "move up", when there are multiple comment lines above, Java skips over exactly one line. But Kotlin skips all comments. I think Java behavior is better and more consistent with "move down".

[674aa47](https://github.com/JetBrains/intellij-community/pull/2185/commits/674aa472a272a77a379840b0fe6ef080a242f657)

abelkov (approved):

PR #2186, opened 2022-10-07
Fix memory leak from anonymous class in ProjectImportAction class

pablobaxter:
Fixes https://youtrack.jetbrains.com/issue/IDEA-303282/Memory-leak-in-ProjectImportAction


nskvortsov:
Thank you for your contribution! Commit will arrive soon, after the CI/CD checks/

PR #2187, opened 2022-10-07
[IDEA-176363] Shell integration for fish breaks universal variables

GrigorenkoPV:
Should fix https://youtrack.jetbrains.com/issue/IDEA-176363/

The problem with the previous implementation was that it would completely ignore `~/.config/fish/fish_variables`, where all the universal variables are stored. (Sadly, that file isn't an actual fish script, so you can't really `source` it.)

The approach used by the previous implementation (overriding `$XDG_CONFIG_HOME`) was too brutal (one had to manually re-`source` all the user configs, functions, etc. after that) and an overkill anyways: despite effectively changing the entire `$__fish_config_dir`, there was only one file there, and even it (outside of trying to revert the consequences of overriding `$XDG_CONFIG_HOME`) would only really define and run one function.

The new implementation that I suggest simply tells fish to source the custom config file before executing other commands or running in interactive mode.

### DIFF
--- a/plugins/terminal/resources/fish/config.fish
+++ b/plugins/terminal/resources/fish/config.fish
@@ -1,32 +1,5 @@
-if test -n "$OLD_XDG_CONFIG_HOME"
-  set XDG_CONFIG_HOME "$OLD_XDG_CONFIG_HOME"
-else
-  set -e XDG_CONFIG_HOME
-end
-
-if test -d ~/.config/fish/functions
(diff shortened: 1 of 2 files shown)


GrigorenkoPV:
I guess this one also closes #1048

segrey:
@GrigorenkoPV Thanks for the PR. It helped to make the fix in d5cd457615639f982adc95658774bdc931f2a143. ETA: 2023.1

PR #2188, opened 2022-10-12
Extract Fields From Continuation This Pointer as Local Variables

alonalbert:
https://youtrack.jetbrains.com/issue/KT-54426/Local-Variables-Not-Shown-in-Debugger-Inside-a-Suspend-Lambda

### DIFF
--- a/platform/editor-ui-api/src/com/intellij/openapi/actionSystem/Presentation.java
+++ b/platform/editor-ui-api/src/com/intellij/openapi/actionSystem/Presentation.java
@@ -19,6 +19,7 @@ import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
 import java.awt.*;
+import java.awt.image.RGBImageFilter;
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


madsager (commented):
This does not appear to work in the (likely common) case where developers are using kotlinx.coroutines. So, I think we have to dive deeper (see individual comments).

Also, let's make sure that we add some tests for this. I think the best place for that is `KotlinVariablePrintingTestGenerated` which is auto-generated from files in `plugins/kotlin/jvm-debugger/test/testData/variables`. So, we should create tests along the lines of:

```
// ATTACH_LIBRARY: coroutines
// SHOW_KOTLIN_VARIABLES

import forTests.builder

suspend fun bar() {
    val x = "Hello"
    barLambdaRunner {
        //Breakpoint!
        println(x)
    }
}

suspend fun barLambdaRunner(f: suspend () -> Unit) = f()

fun main() {
    builder {
        bar()
    }
}
```

However, we should also attempt to update the tests so that we can use kotlinx-coroutines in the tests to make sure that we cover all the typical cases.

madsager:
@yanex @ilmirus looping you in as you might know more about how this is supposed to work than I do. :)

madsager:
I wonder if we should just add the option of extracting the dispatch receiver from the stack frame proxy. The `this` object is confusing as it can be either an extension receiver or a dispatch receiver when looking at coroutines. The following commit is a hacked up prototype of that which allows us to extract and show the captures in the cases where `thisObject` returns the extension receiver:

https://github.com/madsager/intellij-community/commit/cba64913c2cbea198af86c48d3f64eac1103f16b

alonalbert:
I'm closing this PR because it's not the right approach anyway, and my local git clone got messed up.

I forked @madsager from above and added tests.

I think we should move the conversation to the bug instead of here.

PR #2191, opened 2022-10-14
IDEA-303699: Add a ContentManagerListener.contentRemoveQuery call

sprigogin:
This change removes a discrepancy between ToolWindowHeadlessManagerImpl and ContentManagerImpl.


batya239:
merged

PR #2195, opened 2022-10-17
Support directory locations in inspection descriptor composing

blakegallay:
My proposed fix for https://youtrack.jetbrains.com/issue/IDEA-304028/Exporting-code-inspection-results-fails-for-issues-located-at-directories. I am uncertain what specific form the fix should take - should we just add a `PsiDirectory`/`RefDirectory` case to each problematic site, or fix the broader issue of assuming that `getContainingFile()` always works?


platform/analysis-impl/src/com/intellij/codeInspection/ex/HTMLComposerImpl.java
```diff
@@ -79,6 +79,11 @@ private void appendLocation(@NotNull StringBuilder buf, final RefElement refElem
       buf.append(NBSP);
       appendElementReference(buf, refElement, false);
     }
+    else if (refElement instanceof RefDirectory){
+      buf.append("dir");
```
yopox, line 83: Please add `inspection.export.results.directory` in analysis bundle (`community/platform/analysis-api/resources/messages/AnalysisBundle.properties`)

platform/lang-impl/src/com/intellij/codeInspection/ex/DescriptorComposer.java
```diff
@@ -130,7 +131,8 @@ protected void composeDescription(@NotNull CommonProblemDescriptor description,
     VirtualFile vFile = null;
 
     if (expression != null) {
-      vFile = expression.getContainingFile().getVirtualFile();
+      vFile = (expression instanceof PsiFileSystemItem ? (PsiFileSystemItem) expression : expression.getContainingFile()).getVirtualFile();
```
yopox, line 134: Now can be written as `vFile = (expression instanceof PsiFileSystemItem item ? item : expression.getContainingFile()).getVirtualFile();`

yopox:
This PR was merged on 31/01/2023 but not closed, sorry!

PR #2197, opened 2022-10-20
[PY-56210] Add id for PyModuleNameCompletionContributor

trinhanhngoc:
Allow other contributors to take priority over it if needed.

https://youtrack.jetbrains.com/issue/PY-56210/PyModuleNameCompletionContributor-should-be-disabled-in-injected-code


trinhanhngoc:
@east825 Could you have a look at this PR?. Thanks in advance.

east825:
Sorry, my bad, I totally forgot about it. I will integrate your PR later today (it can't be done directly on Github for us).

Actually disabling this contributor inside injection fragments might be a good idea. I just need to check if it conflicts with some use cases, such as completion inside type comments. But until then having an explicit ID for the contributor is a good workaround.

trinhanhngoc:
> Sorry, my bad, I totally forgot about it. I will integrate your PR later today (it can't be done directly on Github for us).
> 
> Actually disabling this contributor inside injection fragments might be a good idea. I just need to check if it conflicts with some use cases, such as completion inside type comments. But until then having an explicit ID for the contributor is a good workaround.

Thank you very much! I almost lost hope in this PR =)

PR #2200, opened 2022-10-21
Extract Captures From Suspend Lambdas

alonalbert:
Forked from https://github.com/madsager/intellij-community/commit/cba64913c2cbea198af86c48d3f64eac1103f16b

Added tests.


yanex:
I think it's better not to bring any Kotlin knowledge to the Java debugger if it can be avoided.
Maybe we can add the `KotlinStackFrameProxy` interface or something on the Kotlin plugin side instead?

madsager:
> I think it's better not to bring any Kotlin knowledge to the Java debugger if it can be avoided. Maybe we can add the `KotlinStackFrameProxy` interface or something on the Kotlin plugin side instead?

Thanks Yan, that's a good point. Is something like this what you have in mind https://github.com/JetBrains/intellij-community/commit/c84d6907e1aa4ce04ec2f7e98acd49b791ceff6b ?

alonalbert:
Updated in response to @yanex as suggested by @madsager.

I also identified another broken use case which required a fix in CapturedValuesSearcher.

Specifically, captured values can have a `$inlined` suffix. This results in:

```
    JavaValue[field] x$inlined = 32
    DummyMessageValueNode = 'x' was optimised out
```

Instead of:
```
    JavaValue[field] x = 32
```

madsager:
@yanex Friendly ping. :)

PR #2201, opened 2022-10-24
KTIJ-23384 "Generate secondary constructor" fails with parameter names that are keywords

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23384


abelkov (approved):
Thank you. I have also improved the PR to support a similar case with properties.

PR #2202, opened 2022-10-25
[IDEA-304611] "Cast can be replaced with variable" inspection fails when the cast is in a loop

Fabrice-TIERCELIN:
Hi @amaembo,

1. Go to *File* → *Settings..*. → *Editor* → *Inspections* (this feature is not released in an official version yet)
2. Enable *Cast can be replaced with variable*
3. Enable Severity: *Warning*
4. Add this method in a Java class:

```Java
  void method(Object foo) {
    String foobar = (String)foo;
    for (int i = 0; i < 10; i++) {
      System.out.println(((String)foo).trim());
      foobar = null;
    }
  }
```

The inspection *Cast can be replaced with variable* is suggested.
***What happens instead?***
The inspection *Cast can be replaced with variable* is not suggested.

`CastCanBeReplacedWithVariableInspection` should not suggest nor directly fix the cast when the variable is reassigned in the same loop as the cast.

I have successfully run the Unit tests.

### DIFF
--- a/java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
+++ b/java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
@@ -1,6 +1,7 @@
 // Copyright 2000-2022 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
 package com.intellij.codeInspection;
 
+import com.intellij.codeInsight.daemon.impl.analysis.HighlightControlFlowUtil;
 import com.intellij.openapi.progress.ProgressIndicatorProvider;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.TextRange;
@@ -17,6 +18,7 @@ import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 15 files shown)


java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
```diff
@@ -130,10 +136,29 @@ private static boolean isChangedBetween(@NotNull final PsiVariable variable,
...
+  @NotNull
+  private static PsiElement getBroadLocation(@NotNull PsiElement scope, @NotNull PsiElement location) {
+    PsiElement broadLocation = location;
+    PsiElement currentElement = location.getParent();
+
+    while (currentElement != null && currentElement != scope) {
+      if (currentElement instanceof PsiForeachStatement
+          || currentElement instanceof PsiWhileStatement
+          || currentElement instanceof PsiForStatement
+          || currentElement instanceof PsiDoWhileStatement) {
```
amaembo, line 152: Probably `instanceof PsiLoopStatement`?

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
```diff
@@ -130,10 +136,29 @@ private static boolean isChangedBetween(@NotNull final PsiVariable variable,
       controlFlow = ControlFlow.EMPTY;
     }
     int startOffset = controlFlow.getEndOffset(start) + 1;
-    int endOffset = controlFlow.getEndOffset(end);
+    int endOffset = controlFlow.getEndOffset(broadEnd);
```
amaembo, line 139: This test starts to fail after your changes, but was correctly working before:

```java
// "Replace '(String) obj' with 's'" "true"

class X {
  void test(Object obj) {
    for (int i = 0; i < 10; i++) {
      String s = (String) obj;
      System.out.println(s);
      System.out.println((String<caret>) obj);
      s = null;
    }
  }
}
```

I will commit this test. Please make sure that it works.

amaembo (changes requested):

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
Fabrice-TIERCELIN: Yes, done

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
Fabrice-TIERCELIN: Handled, and I have integrated the test

FYI, I tried to test the inspection but I failed to make the warning appear; I don't know why

amaembo:
Merged, thanks

PR #2204, opened 2022-10-26
KTIJ-23431: Allow multiple KotlinIconProviders

andymagee:
Fix for https://youtrack.jetbrains.com/issue/KTIJ-23431.

DeclarationPresenters.kt uses the IconProvider extension point to get icons for KotlinDefaultNamedDeclarationPresentation. Currently it finds the first provider that is a KotlinIconProviders and uses it regardless of whether it returns a null icon or not.

This change will move on to check for more KotlinIconProviders in the null case, allowing plugins to register their own provider while still having fallback to the default provider.

Note that this is the same thing that's done in the other places that IconProvider.EXTENSION_POINT_NAME is used. So the standard practice is try multiple providers until a non-null icon is found.

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/presentation/DeclarationPresenters.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/presentation/DeclarationPresenters.kt
@@ -67,8 +67,7 @@ open class KotlinDefaultNamedDeclarationPresentation(private val declaration: Kt
 
     override fun getIcon(unused: Boolean): Icon? {
         for (kotlinIconProvider in IconProvider.EXTENSION_POINT_NAME.iterable.filterIsInstance<KotlinIconProvider>()) {
-            val icon = kotlinIconProvider.getIcon(declaration, Iconable.ICON_FLAG_VISIBILITY or Iconable.ICON_FLAG_READ_STATUS)
-            if (icon != null) return icon
+            kotlinIconProvider.getIcon(declaration, Iconable.ICON_FLAG_VISIBILITY or Iconable.ICON_FLAG_READ_STATUS)?.let { return it }
         }
         return null
     }
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/presentation/DeclarationPresenters.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/presentation/DeclarationPresenters.kt
@@ -66,8 +66,11 @@ open class KotlinDefaultNamedDeclarationPresentation(private val declaration: Kt
     }
 
     override fun getIcon(unused: Boolean): Icon? {
-        val instance = IconProvider.EXTENSION_POINT_NAME.findFirstSafe { it is KotlinIconProvider }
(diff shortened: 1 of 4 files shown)


vladimirdolzhenko:
Could you pls add some test ?

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/presentation/DeclarationPresenters.kt
```diff
@@ -66,8 +66,11 @@ open class KotlinDefaultNamedDeclarationPresentation(private val declaration: Kt
     }
 
     override fun getIcon(unused: Boolean): Icon? {
-        val instance = IconProvider.EXTENSION_POINT_NAME.findFirstSafe { it is KotlinIconProvider }
-        return instance?.getIcon(declaration, Iconable.ICON_FLAG_VISIBILITY or Iconable.ICON_FLAG_READ_STATUS)
+        for (kotlinIconProvider in IconProvider.EXTENSION_POINT_NAME.iterable.filterIsInstance<KotlinIconProvider>()) {
+            val icon = kotlinIconProvider.getIcon(declaration, Iconable.ICON_FLAG_VISIBILITY or Iconable.ICON_FLAG_READ_STATUS)
```
vladimirdolzhenko, line 70: nit: `kotlinIconProvider.getIcon(declaration, Iconable.ICON_FLAG_VISIBILITY or Iconable.ICON_FLAG_READ_STATUS)?.let{ return it }`

andymagee: Done.

andymagee:
Added tests, although I'm not super familiar with the structure of this code base. Please let me know if there's anything else that needs to be added to ensure they're running as part of a test suite or anything of that nature.

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/presentation/DeclarationPresentersTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import javax.swing.Icon
+
+class DeclarationPresentersTest : KotlinLightCodeInsightFixtureTestCase() {
+
+    fun testGetIconWithNoExtraIconProviders() {
+        myFixture.configureByText("Foo.kt", "class <caret>Foo")
+        val element = myFixture.elementAtCaret as KtNamedDeclaration
+
+        // By default, we expect whatever the standard Fe10KotlinIconProvider returns.
+        val defaultExpectedIcon = Fe10KotlinIconProvider().getIcon(element, Iconable.ICON_FLAG_VISIBILITY or Iconable.ICON_FLAG_READ_STATUS)
```
vladimirdolzhenko, line 24: I'd not recommend use directly `Fe10KotlinIconProvider`, use smth like `IconProvider.EXTENSION_POINT_NAME.findFirst { it is KotlinIconProvider }`

vladimirdolzhenko:
LGTM

Thank you so much for your contribution.

vladimirdolzhenko:
merged as https://github.com/JetBrains/intellij-community/commit/485cfbb9e4bf1dbcaf3b69bc2f8f9a18fc15539d

PR #2205, opened 2022-10-27
KTIJ-23042 False positive "Remove explicit type" intention: explicit type is needed for type inference

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23042


plugins/kotlin/idea/tests/testData/intentions/removeExplicitType/anonymousFunctionInitializer.kt
```diff
@@ -1 +1 @@
-val foo: <caret>() -> String = { "" }
\ No newline at end of file
+val foo: <caret>() -> String = fun(): String { return "" }
```
t-kameyama: File name correction

abelkov (approved):

PR #2208, opened 2022-10-30
PY-56361

guoci:
Fix for PY-56361
https://youtrack.jetbrains.com/issue/PY-56361/pandas-Series-iteritems-in-pydevdutilspy-FutureWarning

### DIFF
--- a/python/helpers/pydev/_pydevd_bundle/pydevd_utils.py
+++ b/python/helpers/pydev/_pydevd_bundle/pydevd_utils.py
@@ -606,7 +606,7 @@ def should_evaluate_shape():
 def _series_to_str(s, max_items):
     res = []
     s = s[:max_items]
-    for item in s.iteritems():
+    for item in s.items():
         # item: (index, value)
         res.append(str(item))
     return ' '.join(res)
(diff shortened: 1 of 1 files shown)


lancelote:
@guoci Thank you for the fix. Could you elaborate on why you use `iloc`?

guoci:
@lancelote A warning will be triggered with the following code on the python console.
image


```
import pandas as pd, numpy as np
pd.Series(range(3), index=np.arange(1,4))
```
``` FutureWarning: The behavior of `series[i:j]` with an integer-dtype index is deprecated. In a future version, this will be treated as *label-based* indexing, consistent with e.g. `series[i]` lookups. To retain the old behavior, use `series.iloc[i:j]`. To get the future behavior, use `series.loc[i:j]`.
  s = s[:max_items]```

kierke-gaard:
same here. almost for every pandas command, I'll get the warning in the pycharm interpreter. 

/mnt/c/Program Files/JetBrains/PyCharm 2021.3.2/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_utils.py:606: FutureWarning: iteritems is deprecated and will be removed in a future version. Use .items instead.
  for item in s.iteritems():

lancelote:
Thank you for the contribution. Merged in the monorepo. The fix will be available in PyCharm 2022.3.1.

PR #2211, opened 2022-11-02
KTIJ-23488 False positive redundant 'asSequence' call

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23488


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertCallChainIntoSequenceInspection.kt
```diff
@@ -298,12 +304,6 @@ internal val collectionTerminationFunctionNames = listOf(
     "reduceIndexed",
     "reduceIndexedOrNull",
     "reduceOrNull",
-    "runningFold",
-    "runningFoldIndexed",
-    "runningReduce",
-    "runningReduceIndexed",
-    "scan",
-    "scanIndexed",
```
t-kameyama, line 306: These are not termination functions.

abelkov (approved):

PR #1285, opened 2020-01-08
Handle Note Javac output lines

dansanduleac:
## Before this PR

In the javac output produced by gradle, "Note" log lines like

```
<file path>:<line number>: Note: ...
```

get interpreted as errors when compiling through gradle (either by running `classes` as a task, or having the Gradle integration configured to "Build and run using" Gradle.

It's doubly annoying because it also jumps the editor to the first "error" whenever I build.

image

These types of messages come up when compiling using the errorprone compiler, and you get a hit on an errorprone check with severity [SUGGESTION](https://github.com/google/error-prone/blob/v2.3.4/annotation/src/main/java/com/google/errorprone/BugPattern.java#L171).

## After this PR

Javac output lines of the shape `<file>:<line>: Note:` get interpreted as as info messages in the structured output when compiling with gradle.

PR #2212, opened 2022-11-03
IDEA-302026 IDEA treats javac note messages as errors during build if message includes a path

valepakh:
JavacOutputParser hardcodes prefixes from javac and there's an error in the prefix for the note.
As a result, `JavacOutputParser.parse` can't find the prefix and treats this message as an error by default.
In the JDK sources the prefix for the `DiagnosticType.NOTE` is (probably erroneously) "Note" with the capital N, while for `WARNING` and `ERROR` they start with the lowercase letter.
See [AbstractDiagnosticFormatter.java](https://github.com/openjdk/jdk/blob/master/src/jdk.compiler/share/classes/com/sun/tools/javac/util/AbstractDiagnosticFormatter.java#L112) and [compiler.properties](https://github.com/openjdk/jdk/blob/master/src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties#L1612).
So this very simple change is to keep the text in sync with the JDK sources.

### DIFF
--- a/platform/lang-impl/src/com/intellij/build/output/JavacOutputParser.java
+++ b/platform/lang-impl/src/com/intellij/build/output/JavacOutputParser.java
@@ -32,7 +32,7 @@ public class JavacOutputParser implements BuildOutputParser {
 
   private static final char COLON = ':';
   private static final @NonNls String WARNING_PREFIX = "warning:"; // default value
-  private static final @NonNls String NOTE_PREFIX = "note:";
+  private static final @NonNls String NOTE_PREFIX = "Note:";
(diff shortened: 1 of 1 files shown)


valepakh:
There was a previous attempt to fix this, which was incorrect https://github.com/JetBrains/intellij-community/pull/1285

valepakh:
@vladsoroka Is there any chance you can look at this as a reviewer of previous fix?

HackerMadCat:
d62900085a71625b319060da62dbb02a6984a0db

PR #2214, opened 2022-11-03
KTIJ-23517: Add test for implement members annotation behavior

andymagee:
Implementing an override in K1 explicitly contains a filter to remove annotations, with the exception of any annotations marked with @RequiresOptIn.

K2 does not implement this behavior, instead retaining all annotations.

This change adds a test documenting the current behavior. [KTIJ-23517](https://youtrack.jetbrains.com/issue/KTIJ-23517) tracks the difference in behavior between K1 and K2.


darthorimar:
Yeah, it's not finished K2 implementation, thanks!

darthorimar:
Merged with https://github.com/JetBrains/intellij-community/pull/2215

PR #2216, opened 2022-11-04
Create a KotlinStackFrameValueContributor Extension Point

alonalbert:
https://youtrack.jetbrains.com/issue/KTIJ-23529/Feature-Request-Add-an-Extension-Point-For-Contributing-Values-to-Debugger-View

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stackFrame/KotlinStackFrame.kt
+++ b/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stackFrame/KotlinStackFrame.kt
@@ -2,7 +2,6 @@
 
 package org.jetbrains.kotlin.idea.debugger.core.stackFrame
 
-import com.intellij.debugger.engine.DebugProcessImpl
 import com.intellij.debugger.engine.JavaStackFrame
 import com.intellij.debugger.engine.JavaValue
 import com.intellij.debugger.engine.evaluation.EvaluationContextImpl
@@ -61,8 +60,12 @@ open class KotlinStackFrame(
         if (!removeSyntheticThisObject(evaluationContext, children, existingVariables) && thisVariables.isNotEmpty()) {
             remapThisObjectForOuterThis(evaluationContext, children, existingVariables)
         }
-
         children.add(evaluationContext, thisVariables, otherVariables)
+        for (contributor in KotlinStackFrameValueContributor.EP.extensions) {
+            for (value in contributor.contributeValues(this, evaluationContext, variables)) {
+                children.add(value)
+            }
+        }
     }
 
     private fun XValueChildrenList.add(
--- a/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stackFrame/KotlinStackFrameValueContributor.kt
+++ b/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stackFrame/KotlinStackFrameValueContributor.kt
@@ -1,0 +1,27 @@
+// Copyright 2000-2022 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.kotlin.idea.debugger.core.stackFrame
+
+import com.intellij.debugger.engine.evaluation.EvaluationContextImpl
+import com.intellij.debugger.jdi.LocalVariableProxyImpl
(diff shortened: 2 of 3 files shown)


java/java-impl/src/META-INF/JavaPlugin.xml
```diff
@@ -237,6 +237,9 @@
     <extensionPoint qualifiedName="com.intellij.debugger.asyncStackTraceProvider"
                     interface="com.intellij.debugger.engine.AsyncStackTraceProvider" dynamic="true"/>
 
+    <extensionPoint qualifiedName="com.intellij.debugger.kotlinStackFrameValueContributor"
```
yanex, line 240: Please register extension points in the Kotlin plugin if they're Kotlin-only.
This particular extension point should be registered in `plugins/kotlin/plugin/common/resources/META-INF/jvm-debugger.xml`.

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stackFrame/KotlinStackFrame.kt
```diff
@@ -61,8 +60,12 @@ open class KotlinStackFrame(
         if (!removeSyntheticThisObject(evaluationContext, children, existingVariables) && thisVariables.isNotEmpty()) {
             remapThisObjectForOuterThis(evaluationContext, children, existingVariables)
         }
-
         children.add(evaluationContext, thisVariables, otherVariables)
+        KotlinStackFrameValueContributor.EP.extensions.forEach { contributor ->
+            contributor.contributeValues(this, evaluationContext, variables,).forEach {
```
yanex, line 65: Extra trailing comma.

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stackFrame/KotlinStackFrame.kt
```diff
@@ -61,8 +60,12 @@ open class KotlinStackFrame(
         if (!removeSyntheticThisObject(evaluationContext, children, existingVariables) && thisVariables.isNotEmpty()) {
             remapThisObjectForOuterThis(evaluationContext, children, existingVariables)
         }
-
         children.add(evaluationContext, thisVariables, otherVariables)
+        KotlinStackFrameValueContributor.EP.extensions.forEach { contributor ->
```
yanex, line 64: We usually use the `for {}` loop instead (unless a callable reference is passed).

java/java-impl/src/META-INF/JavaPlugin.xml
```diff
@@ -237,6 +237,9 @@
...
+                    interface="org.jetbrains.kotlin.idea.debugger.core.stackFrame.KotlinStackFrameValueContributor" dynamic="true"/>
```
yanex, line 241: By the way, there is `com.intellij.debugger.engine.FrameExtraVariablesProvider`. Is it not enough for your use-cases?

yanex (changes requested):

java/java-impl/src/META-INF/JavaPlugin.xml
alonalbert: Done

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stackFrame/KotlinStackFrame.kt
alonalbert: Done

java/java-impl/src/META-INF/JavaPlugin.xml
alonalbert: I wasn't aware of this EP but unfortunately, it's not enough for my use case.

I think it should be possible to extend FrameExtraVariablesProvider in a way that will make it work.

I'm sending you an image of my use case in a private channel.

alonalbert: My use requires adding a custom XValue to children. The existing EP doesn't allow that. It just seems allow adding a code fragment dressed as a variable.

As I said, I think I can extend it by adding another method to to that returns a list or XValue that can be added to children.

Do you want me to make this change or stick with my new EP?

yanex (approved):

yanex:
Merged to `master` manually: https://github.com/JetBrains/intellij-community/commit/c0dadceeee702d07e00ecafedf1eb41021aa9f11.
Thanks!

PR #2217, opened 2022-11-05
Fix use of `executor.id` instead of `executor.toolWindowId` in `RunContentManagerImpl`

gfszr:
Fixes [IDEA-305375](https://youtrack.jetbrains.com/issue/IDEA-305375/processToolWindowContentManagers-in-RunContentManagerImpl-class-uses-executorid-instead-of-executortoolWindowId).


hurricup:
This change causes `com.intellij.execution.ui.RunContentManagerImpl#getAllDescriptors` to return duplicates. Some executors share `toolWindowId`

PR #2218, opened 2022-11-07
IDEA-265238 - added support time conversions for `format` methods

pyltsin:
IDEA-265238 
https://youtrack.jetbrains.com/issue/IDEA-265238/Report-illegal-time-conversions-in-format-strings

Additionally, I added that `d` is not used with precision (`"%.1d`) , according to https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/Formatter.html
for Byte, Short, Integer, and Long the precision is not applicable.

Also,  checks for the most popular java.time classes about appropriate conversions are added.


amaembo (approved):
Thanks, will merge

PR #2219, opened 2022-11-07
IDEA-305522 added support java.time for DFA

pyltsin:
https://youtrack.jetbrains.com/issue/IDEA-305522/Base-support-javatime-classes-for-DataFlow
Tried to add base support java.time classes for DFA


java/java-analysis-impl/src/com/intellij/codeInspection/dataFlow/HardcodedContracts.java
```diff
@@ -319,6 +350,42 @@ static List<MethodContract> mathMinMax(boolean isMax) {
...
+    if (specialField != null) {
+      qualifier = qualifier.specialField(specialField);
+      argument = argument.specialField(specialField);
+    }
+
+    return Arrays.asList(
+      multipleConditionContract(
+        List.of(
+          ContractValue.condition(qualifier, relationType, argument),
+          ContractValue.condition(argument, RelationType.IS, qualifier)),
```
amaembo, line 366: What relation do you want to express with 'IS'? This part is unclear to me. What will happen if this clause is omitted?

amaembo:
To be honest, I'm not very happy with this change. It's quite big and may add maintenance overhead, but it I don't believe that it adds much value, as not so many code uses date-time API and not so many of them extensively use all this math. Especially I don't like the BigInteger part: you are significantly extending the Df type system to support the marginal case. Let's discuss this before rushing into the implementation. Can you estimate whether real bugs could be found by this?

PR #2220, opened 2022-11-09
IDEA-305079 show usages for record fields using lombok @builder and @with

henri5:
fixes https://youtrack.jetbrains.com/issue/IDEA-305079


henri5:
closed as some other tests broke

PR #2222, opened 2022-11-10
Add build progress indicator for Gradle builds

asodja:
This adds a progress indicator for Gradle builds to IntelliJ. The progress indicator behaves similarly as Gradle progress bar when Gradle is run via CLI. 
Progress indicator is shown only for Gradle 7.6 or greater, since events for it were added in Gradle 7.6.

progress-bar-2

The main logic, how progress state is updated is handled in [GradleProgressIndicatorEventHelper.kt](https://github.com/JetBrains/intellij-community/compare/master...gradle:intellij-community:asodja/gradle-progress-master?expand=1#diff-1edb6a73a02fa14ab39c14cfcaf36f8503f6a51fc5ccce68ac79aa6e1078fea8). It is similar to what Gradle has internally in [BuildStatusRenderer](https://github.com/gradle/gradle/blob/a410d0f6a7af60dc6c383b7cbb8f912a3cbf209d/subprojects/logging/src/main/java/org/gradle/internal/logging/console/BuildStatusRenderer.java#L72-L108).

Right now I show progress indicator, by starting a dummy background task in a different thread that waits on a build finish, check [GradleProgressIndicatorWrapper.kt](https://github.com/JetBrains/intellij-community/pull/2222/files#diff-d752c30ce7f7122e13654a47d35954e77b4fb3986e8968b3aa7d779f45b1be1e). Since I don't know IntelliJ API that much, there is probably a better way to show a progress indicator.

-----
The main project I used for testing was [gradle/gradle](https://github.com/gradle/gradle). It should work for any task (e.g. tests) and also for Gradle syncs and continuous builds. But some more testing would be nice.  

Note: I currently don't show anything after all tasks are done. That might look silly for continuous builds, since last run task will be shown in the progress indicator until next change. We could detect that all tasks are finished, by catching the `BuildPhaseFinishEvent.RUN_MAIN_TASKS` event. But I didn't know what should be shown to the user, maybe `Gradle: All work finished` or `Gradle: All tasks finished`. But that might look weird for sync, since IntelliJ does some extra work after there.


plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java
```diff
@@ -107,10 +115,16 @@ public void executeTasks(final @NotNull ExternalSystemTaskId id,
         CancellationToken cancellationToken = cancellationTokenSource.token();
         myHelper.ensureInstalledWrapper(id, rootProjectPath, effectiveSettings, listener, cancellationToken);
       }
-      myHelper.execute(projectPath, effectiveSettings, id, listener, cancellationTokenSource, connection -> {
-        executeTasks(id, tasks, projectPath, effectiveSettings, jvmParametersSetup, listener, connection, cancellationTokenSource);
-        return null;
-      });
+
+      GradleProgressIndicatorWrapper progressIndicatorWrapper = new GradleProgressIndicatorWrapper(id, effectiveSettings, listener);
```
nskvortsov, line 119: I am not sure if we need the wrapper and a new progress indicator here. The proper build event that you produce in GradleProgressListener should be enough.
Was it a workaround for some sort of issue?

nskvortsov (changes requested):
Solid work, thank you!
An availability of Gradle 7.6 is a prerequisite for merging this PR.
Also see my comments about GradleProgressIndicatorWrapper. At first glance, it is not needed

eskatos:
> Note: I currently don't show anything after all tasks are done. That might look silly for continuous builds, since last run task will be shown in the progress indicator until next change. We could detect that all tasks are finished, by catching the BuildPhaseFinishEvent.RUN_MAIN_TASKS event. But I didn't know what should be shown to the user, maybe Gradle: All work finished or Gradle: All tasks finished. But that might look weird for sync, since IntelliJ does some extra work after there.

I think showing something would be good. Something like "Gradle: All work finished" sounds good to me. Gradle is not doing more work. If the IDEs or IDE plugins are doing work after Gradle is done they should take care of showing progress, attributed to them.

plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java
asodja: Without that wrapper progress indicator is not shown when running tests or tasks. Only a progress above "Build" window is shown:
Screenshot 2022-11-25 at 17 52 24

asodja:
@eskatos I can do something like that.

plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java
nskvortsov: no, I guess it is an expected behavior.

nskvortsov:
when a progress of project evaluation is interrupted with a progress of a dependency download, it does not look good. Seems like we need to handle file download events separately, so they do not interfere with the overall progress.

leonard84:
Can we somehow show an overall progress, while also showing parallel actions, otherwise it looks like it is downloading files in sequentially instead of in parallel?

nskvortsov:
@leonard84 
this would be ideal - each artifact download should come with a separate progress bar. I will check how we can approach this.

nskvortsov:
I suggest we do the following. 

1. Process "file download" events from Gradle separately. A new component is required. It will manage multiple progress bars showing the downloads. Each progress bar can be started and stopped via `com.intellij.openapi.progress.TasksKt#withBackgroundProgress` API.  The component has to:
  * spawn new progress bars for new downloads.
  * update existing progress bars.
  * clear progress bar upon completion
2. Use "main" status update events for the Gradle's build progress indicator as suggested in this pull request.

a contributor can create a new component, as I will only have time for this in about a month.

eskatos:
@nskvortsov, @asodja, where are we with this PR?

asodja:
@nskvortsov I checked that download events and I think it's not bad. Basicallly we show download progress instead of build progress while downloading, but after download bar is updated and shows build progress. So progress bar is not in any weird state, example:
ezgif com-video-to-gif (1)

I tried to also separate download events, but I have problems with `clear progress bar upon completion` part. Basically I would need to receive onEnd/onCancel events for sync/task execution, but that might require wrapping `ExternalSystemTaskNotificationListener` code on multiple places.

plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleProgressListener.java
```diff
@@ -71,11 +83,11 @@ public GradleProgressListener(
     myTaskId = taskId;
     myGradleVersion = extractGradleVersion(settings);
     myOperationId = taskId.hashCode() + ":" + FileUtil.pathHashCode(buildRootDir == null ? UUID.randomUUID().toString() : buildRootDir);
+    myDownloadProgressListener = new GradleDownloadProgressListener(taskId);
```
nskvortsov, line 86: and here we should get GradleDownloadProgressListener service instance. It makes sense to check for nulls and wrap in read action (in case the project is getting closed right in this moment)
```
myTaskId.findProject().getService(GradleDownloadProgressListener.class)
```

nskvortsov (changes requested):

plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleProgressIndicatorWrapper.kt
```diff
@@ -0,0 +1,68 @@
...
+  fun runWithProgressIndicator(consumer: Consumer<ExternalSystemTaskNotificationListener>) {
+    if (!areGradleBuildProgressEventsSupported(effectiveSettings)) {
+      // Don't run with progress indicator if build progress events are not supported
+      consumer.accept(listener)
+      return
+    }
+
+    val progressIndicator = AtomicReference<ProgressIndicator>()
+    val countDownLatch = CountDownLatch(1)
+    ProgressManager.getInstance().run(object : Task.Backgroundable(id.findProject(), "Gradle build", false) {
```
nskvortsov, line 36: this steals a thread just to create an indicator. Please, do not do this. I think you now know better ways (as GradleDownloadProgressListenerImpl).

plugins/gradle/src/org/jetbrains/plugins/gradle/service/task/GradleTaskManager.java
```diff
@@ -102,10 +103,16 @@ public void executeTasks(final @NotNull ExternalSystemTaskId id,
...
+      progressIndicatorWrapper.runWithProgressIndicator(
```
nskvortsov, line 108: Instead of wrapping the whole thing, you can simply register the `listenerWithProgressIndicator` (created by GradleProgressIndicatorWrapper) using the following code:
```
    ExternalSystemProgressNotificationManager progressManager =
      ApplicationManager.getApplication().getService(ExternalSystemProgressNotificationManager.class);
      progressManager.addNotificationListener(id, listenerWithProgressIndicator);
```
It will be removed automatically, no need to unsubscribe.

nskvortsov (changes requested):
I believe we can simplify the way a new progress indicator is attached to the Gradle task. See my comments inline.

cobaku:
Hello, @asodja. Thank you for this great work. Unfortunately, this pull request cannot be accepted because the current implementation of Gradle TAPI progress listeners for `OperationType.BUILD_PHASE` lead to significant (~7%) performance degradation on the Gradle side.

PR #2226, opened 2022-11-13
IDEA-305934 Handle widening at declaration & assignment

Fabrice-TIERCELIN:
Hi @amaembo,

1. Go to *File* → *Settings..*. → *Editor* → *Inspections*
2. Enable *Cast can be replaced with variable* (this feature is not released in an official version yet)
3. Enable Severity: *Warning*
4. Add this method in a Java class:

```Java
  void method() {
    String anActualText = "Hello World! ";
    Object aText = anActualText;
    System.out.println(((String) aText).trim());
    aText = Integer.MAX_VALUE;
  }
```

The inspection *Cast can be replaced with variable* is suggested.
***What happens instead?***
The inspection *Cast can be replaced with variable* is not suggested.

`CastCanBeReplacedWithVariableInspection` already handles type narrowing. It may also handle type widening. The order of assignment is just the opposite. This use case is mostly as frequent as the original use case. As for the original use case, no assignment should be done between the two assignments.

I have added some unit tests that run successfully.

### DIFF
--- a/java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
+++ b/java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
@@ -72,6 +72,7 @@ public class CastCanBeReplacedWithVariableInspection extends AbstractBaseJavaLoc
     final PsiCodeBlock methodBody = method.getBody();
     if (methodBody == null) return null;
     final TextRange expressionTextRange = expression.getTextRange();
+    if (expressionTextRange == null) return null;
     PsiExpression operand = PsiUtil.skipParenthesizedExprDown(expression.getOperand());
     if (operand == null) return null;
     PsiType castType = expressionCastType.getType();
@@ -86,11 +87,7 @@ public class CastCanBeReplacedWithVariableInspection extends AbstractBaseJavaLoc
         .toList();
     PsiResolveHelper resolveHelper = PsiResolveHelper.getInstance(method.getProject());
     for (PsiTypeCastExpression occurrence : found) {
-      ProgressIndicatorProvider.checkCanceled();
-      final TextRange occurrenceTextRange = occurrence.getTextRange();
-      if (occurrence == expression || occurrenceTextRange.getEndOffset() >= expressionTextRange.getStartOffset()) {
-        continue;
-      }
+      if (!isAtRightLocation(expression, expressionTextRange, occurrence)) continue;
 
       final PsiLocalVariable variable = getVariable(occurrence);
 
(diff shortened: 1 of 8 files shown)


java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
```diff
@@ -116,7 +113,52 @@ private static PsiVariable findReplacement(@NotNull PsiMethod method,
       }
     }
 
-    return null;
+    List<PsiAssignmentExpression> narrowTypeFound =
+      SyntaxTraverser.psiTraverser(method)
+        .filter(PsiAssignmentExpression.class)
+        .filter(assignment -> EquivalenceChecker.getCanonicalPsiEquivalence().expressionsAreEquivalent(assignment.getLExpression(), operand))
+        .filter(assignment -> {
+          PsiExpression typeElement = assignment.getRExpression();
```
jqyp, line 121: Bad naming due to copy-paste.

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
```diff
@@ -116,7 +113,52 @@ private static PsiVariable findReplacement(@NotNull PsiMethod method,
...
+          return typeElement != null && InstanceOfUtils.typeCompatible(typeElement.getType(), castType, operand);
```
jqyp, line 122: Add type check for `null`, please. Otherwise, the exception 
`java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: Argument for @NotNull parameter 'instanceOfType' of com/siyeh/ig/psiutils/InstanceOfUtils.typeCompatible must not be null`
will be thrown for this code:
```
void method() {
  String anActualText = "Hello World!";
  Object aText;
  aText = unresolvedMethodCall(anActualText);
  System.out.println(((String) aText).trim());
}
```

And also for line 144.

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
```diff
@@ -116,7 +113,52 @@ private static PsiVariable findReplacement(@NotNull PsiMethod method,
...
+        || !(wideReferenceExpression.resolve() instanceof PsiLocalVariable wideVariable)) return null;
+
+    PsiExpression narrowExpression = wideVariable.getInitializer();
+    if (!(narrowExpression instanceof PsiReferenceExpression narrowReferenceExpression)
+        || !(narrowReferenceExpression.resolve() instanceof PsiVariable narrowVariable)
+        || !InstanceOfUtils.typeCompatible(narrowExpression.getType(), castType, operand)) return null;
+
+    if (!isAtRightLocation(expression, expressionTextRange, narrowExpression)) return null;
+
+    if (narrowVariable == null
```
jqyp, line 148: `null` check is redundant. The result of the `instanceof` operator is `false` if the value of the operand is `null`.

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
```diff
@@ -116,7 +113,52 @@ private static PsiVariable findReplacement(@NotNull PsiMethod method,
...
+        || resolveHelper.resolveReferencedVariable(narrowVariable.getName(), expression) != narrowVariable
```
jqyp, line 149: Pleases, add `null` check to avoid the "Argument 'narrowVariable.getName()' might be null" warning.

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
```diff
@@ -182,6 +224,16 @@ private static PsiLocalVariable getVariable(@NotNull PsiExpression occurrence) {
     return null;
   }
 
+  @Nullable
+  private static PsiLocalVariable getVariable(@NotNull PsiAssignmentExpression occurrence) {
+    if (occurrence.getRExpression() instanceof PsiReferenceExpression referenceExpression &&
```
jqyp, line 229: The expression on the right side of the assignment could be wrapped into parentheses. Please, use `PsiUtil.skipParenthesizedExprDown`. Otherwise, a false negative will be get for this code:

```
void method() {
  String anActualText = "Hello World!";
  Object aText;
  aText = (anActualText);
  System.out.println(((String) aText).trim());
}
```

jqyp (changes requested):

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
Fabrice-TIERCELIN: Fixed

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
Fabrice-TIERCELIN: Fixed

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
Fabrice-TIERCELIN: Fixed

Fabrice-TIERCELIN:
I have renamed `CastCanBeReplacedWithVariableInspectionTest` to `CastCanBeReplacedWithVariableFixTest` because it tests the fix and not the inspection.

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
Fabrice-TIERCELIN: Done

java/java-impl-inspections/src/com/intellij/codeInspection/CastCanBeReplacedWithVariableInspection.java
Fabrice-TIERCELIN: Done

Fabrice-TIERCELIN:
@amaembo, I have forgotten to tell the issues were fixed. Can you please re-open de PR?

jqyp:
Hi @Fabrice-TIERCELIN, I reviewed your changes and merged them. Would you like anything else?

Fabrice-TIERCELIN:
Everything seems to be good

PR #2227, opened 2022-11-14
Stop JavaInheritanceWeigher from promoting classes

andymagee:
Fix for https://youtrack.jetbrains.com/issue/KTIJ-23617

JavaInheritanceWeigher seems intended to give a higher weight to super classes of the current ProximityLocation as compared to other classes, and it indeed does that. However, it also ranks all PsiClass elements above any non-PsiClass elements, by virtue of returning `null` for other elements. (Null will be ranked lower than either false or true, one of which is returned for all PsiClass elements.)

This change returns false for non-PsiClass elements. The net result is that super classes will be promoted with a higher weight, whereas non-super classes and non-class elements will be left alone.


vladimirdolzhenko:
Could you please add test for that ?

andymagee:
> Could you please add test for that ?

I added a quick unit test for this specific scenario. Unfortunately I don't have bandwidth for anything more in-depth, since there's not existing test infrastructure for this weigher that I can find. (This bug isn't a priority for Android Studio since we already have a workaround; I figured I would just submit the bug fix for KTIJ-23617 since I had already located the source of the issue. If the lack of more E2E testing doesn't meet standards, please feel free to reject the PR.)

akozlova:
Hi Andy, I am sorry for the late reply. 
I'd say that in java, the default is to have methods (which can appear only as static imports) on the bottom. Thus, I am afraid, I'd keep Peter's changes for java. So, I think that kotlin should use dedicated weigher, so probably the workaround which you have for AS should be used for all kotlin files. 
WDYT?
Thanks

andymagee:
Sure, that's fine; sounds like this isn't the right fix here, and I'm happy to abandon the PR.

I think it's appropriate to keep https://youtrack.jetbrains.com/issue/KTIJ-23617 open, since the resulting symptom is still an issue. (That Kotlin "Add Imports" always ranks Java classes above other more relevant Kotlin items is definitely not optimal behavior.) I also think the fix for that should not be that every Kotlin weight has to explicitly position itself before `JavaInheritanceWeigher`, although I'm not sure what exactly; ideally there would be some way to disable the weigher across the board for Kotlin since it's not really applicable.

PR #2228, opened 2022-11-14
[Java. Inspections] IDEA-297385 - added options to ignore overloaded methods whose parameters are definitely incompatible

pyltsin:
I tried to fix https://youtrack.jetbrains.com/issue/IDEA-297385/Overloaded-varargs-method-triggers-when-not-needed

About realization.
1. I tried to follow the way from ` OverloadedMethodsWithSameNumberOfParametersInspection`, which has the additional parameter, and it seems to me, that it could be useful in OverloadedVarargsMethodInspection inspection too.

2. It seems to me, that cases like 
```java
public void method(Integer i1, Integer i2, String... strings){}
public void method(Integer i1){}
```
are obvious and there is no reason to highlight them because we have a certain indicator for these methods: Integer i2

3. I am not sure, but cases  like:
```java
public void method(Integer i1, Integer i2, String... strings){}
public void method(Integer i1, Integer i2, Integer i3){}
```
are also obvious to me and,  probably,  IDEA could skip it.

4. In this inspection, I prefer to use `isConvertibleFrom` instead of  usual `isAssignableFrom`, because I want to catch cases like

```java
public class Test<E> {
    public void test(E e) {}
    public void test(String... strings) {}
}
```

Thank you in advance

### DIFF
--- a/plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
+++ b/plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
@@ -15,9 +15,8 @@
  */
 package com.siyeh.ig.naming;
 
-import com.intellij.psi.PsiClass;
-import com.intellij.psi.PsiMethod;
-import com.intellij.psi.PsiSubstitutor;
+import com.intellij.codeInspection.ui.SingleCheckboxOptionsPanel;
+import com.intellij.psi.*;
 import com.intellij.psi.util.MethodSignatureUtil;
 import com.intellij.psi.util.TypeConversionUtil;
 import com.siyeh.InspectionGadgetsBundle;
@@ -25,7 +24,18 @@ import com.siyeh.ig.BaseInspection;
 import com.siyeh.ig.BaseInspectionVisitor;
 import org.jetbrains.annotations.NotNull;
 
+import javax.swing.*;
+
 public class OverloadedVarargsMethodInspection extends BaseInspection {
+
+  @SuppressWarnings({"PublicField"})
+  public boolean ignoreInconvertibleTypes = true;
+
+  @Override
+  public JComponent createOptionsPanel() {
+    return new SingleCheckboxOptionsPanel(InspectionGadgetsBundle.message(
+      "overloaded.vararg.method.problem.option"), this, "ignoreInconvertibleTypes");
+  }
 
   @Override
   @NotNull
@@ -46,7 +56,7 @@ public class OverloadedVarargsMethodInspection extends BaseInspection {
     return new OverloadedVarargMethodVisitor();
   }
 
-  private static class OverloadedVarargMethodVisitor extends BaseInspectionVisitor {
+  private class OverloadedVarargMethodVisitor extends BaseInspectionVisitor {
 
     @Override
     public void visitMethod(@NotNull PsiMethod method) {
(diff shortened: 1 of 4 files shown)


BasLeijdekkers:
Please add a description of the new option to `OverloadedVarargsMethod.html`. You can look at the description of other inspections with options for examples.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
```diff
@@ -65,10 +75,54 @@ public void visitMethod(@NotNull PsiMethod method) {
...
+      //otherParameterList: (Integer i1)
+      if (lengthForVarArgs > otherLength + 1) {
+        return false;
+      }
+
+      for (int i = 0; i < otherLength; i++) {
+        PsiType type;
+        if (i < lengthForVarArgs) {
+          type = getTypeForComparison(parametersWithVarArgs, i);
+        }else{
```
BasLeijdekkers, line 107: Our codestyle for `else` statements is quite different from this. Please reformat the file.

BasLeijdekkers (changes requested):
Except for the minor issues, this looks good to me.

pyltsin:
> Please add a description of the new option to `OverloadedVarargsMethod.html`. You can look at the description of other inspections with options for examples.
The description of the new option is added.


Thank you for this review!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
pyltsin: Fixed. Thank you!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
```diff
@@ -65,10 +75,55 @@ public void visitMethod(@NotNull PsiMethod method) {
...
+      //example:
+      //parameterListWithVarArgs: (Integer i1, Integer i2, String... strings)
+      //otherParameterList: (Integer i1)
+      if (lengthForVarArgs > otherLength + 1) {
+        return false;
+      }
+
+      for (int i = 0; i < otherLength; i++) {
+        PsiType type;
+        if (i < lengthForVarArgs) {
```
BasLeijdekkers, line 105: I recommend converting this `if` statement into a conditional expression (using the intention/quick-fix) and then apply the "Push conditional expression inside branch".

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
```diff
@@ -65,10 +75,55 @@ public void visitMethod(@NotNull PsiMethod method) {
...
+        if (!type.isConvertibleFrom(otherType) && !otherType.isConvertibleFrom(type)) {
+          return false;
+        }
+      }
+
+      return true;
+    }
+  }
+
+  private static PsiType getTypeForComparison(PsiParameter[] parameters, int i) {
```
BasLeijdekkers, line 122: I think it would better to just pass a single `PsiParameter` parameter into this method. This method doesn't do any bounds checking anyway, which may otherwise be expected.

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
```diff
@@ -65,10 +75,55 @@ public void visitMethod(@NotNull PsiMethod method) {
...
+    PsiType type = parameters[i].getType();
+    if (type instanceof PsiEllipsisType) {
+      type = ((PsiEllipsisType)type).getComponentType();
```
BasLeijdekkers, line 125: Perhaps just directly return instead of assigning to the variable. After that the `if` statement can be converted to a conditional expression as well.

plugins/InspectionGadgets/testsrc/com/siyeh/ig/naming/OverloadedVarargsMethodInspectionTest.java
```diff
@@ -64,4 +64,29 @@ public void testGenericMethods() {
...
+           "  public void method() {}" +
+           "  public void method(int p1) {}" +
+           "  public void method(int p1, int p2) {}" +
+           "  public void method(int p1, String p2, int p3) {}" +
+           "  public void method(int p1, String p2, String p3, int p4) {}" +
+           "  public void method(int p1, String p2, String... p3) {}" +
+           "}");
+  }
+
+  public void testWarningBecauseOfInheritance() {
```
BasLeijdekkers, line 79: I see no inheritance in the test case? Maybe rename?

BasLeijdekkers (changes requested):
Sorry, I found a few more small issues

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
pyltsin: Thank you! Changed to a conditional expression

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
pyltsin: Thank you! Fixed

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/naming/OverloadedVarargsMethodInspection.java
pyltsin: Thank you! Changed to a conditional expression

plugins/InspectionGadgets/testsrc/com/siyeh/ig/naming/OverloadedVarargsMethodInspectionTest.java
pyltsin: Thank you. Renamed

PR #2229, opened 2022-11-14
[Java. Inspections] IDEA-305106 - ignore expressions inside test containers

pyltsin:
I tried to fix https://youtrack.jetbrains.com/issue/IDEA-305106/Inspection-Java-Probable-bugs-Result-of-method-call-ignored-should-ignore-expressions-inside-assertThatThrownBy

About realization.

1. It seems to me, that IDEA shouldn't  skip all cases inside `assert` methods, because the case
```java
String text = "test";
assertThatThrownBy(() -> test.isEmpty());
```
doesn't make sense.

2. But it could be useful for `KNOWN_EXCEPTIONAL_SIDE_EFFECTS` methods because their use is close to this case.

3. Test frameworks are being developed constantly and, for me, there is no reason to have a list of all methods. Additionally, sometimes it could be useful to skip this inspection not only in methods, which check exceptions but also in methods that check a positive way. That's why, it seems to me, that it is possible to check here only package when these methods are contained and patterns of its names (for example, I found this way in `com.siyeh.ig.psiutils.SideEffectChecker#mayHaveExceptionalSideEffect`)

Thank you in advance

### DIFF
--- a/plugins/InspectionGadgets/src/com/siyeh/ig/bugs/IgnoreResultOfCallInspection.java
+++ b/plugins/InspectionGadgets/src/com/siyeh/ig/bugs/IgnoreResultOfCallInspection.java
@@ -49,6 +49,8 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 
+import static com.intellij.psi.CommonClassNames.JAVA_UTIL_FUNCTION_SUPPLIER;
+
 public class IgnoreResultOfCallInspection extends BaseInspection {
   private static final CallMatcher STREAM_COLLECT =
     CallMatcher.instanceCall(CommonClassNames.JAVA_UTIL_STREAM_STREAM, "collect").parameterCount(1);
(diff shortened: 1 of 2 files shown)


plugins/InspectionGadgets/src/com/siyeh/ig/bugs/IgnoreResultOfCallInspection.java
```diff
@@ -267,6 +267,9 @@ else if (call instanceof PsiMethodReferenceExpression) {
         exception = KNOWN_EXCEPTIONAL_SIDE_EFFECTS.mapFirst((PsiMethodReferenceExpression)call);
       }
       if (exception == null) return false;
+      if (isInTestContainer(call)) {
```
BasLeijdekkers, line 270: This does not seem to be the correct place to check this. If the inspection setting `report all ignored non-library call` is enabled non-library calls are still reported in a test container. This does not seem to be desired.

BasLeijdekkers (changes requested):

plugins/InspectionGadgets/src/com/siyeh/ig/bugs/IgnoreResultOfCallInspection.java
pyltsin: Thank you for this review!
Honestly, it is a really interesting point, because before your comment I thought that IDEA should report this case anyway. Because a case like 
```java
        class X {
          private static String twice(String s) {
            return s+s;
          }
          public void test(String abc){
            org.junit.jupiter.api.Assertions.assertThrows(RuntimeException.class, ()-> {
                twice(abc);
            });
          }
        }
```
would seem to be an error.

After reading your comment, I discovered several cases in which it would be useful. So, additionally, I added one more option 
`Don't consider non-library calls inside methods of AssertJ or JUnit as ignored` to process these cases. 

Thank you!

BasLeijdekkers:
It appears I didn't read your first comment correctly before. 
In fact your example should not be reported:
```
String text = "test";
assertThatThrownBy(() -> test.isEmpty());
```
By calling `assertThatThrownBy()` the user is indicating they are not interested in the results of the `test.isEmpty()` method call. They are only interested in the result of the test, which depends on the exception being thrown or not, not the result returned by the method.
It is not useful to only ignore known exceptional side effects, because 99.999% of the time that is not the code that will be tested.

So in my opinion the user is requesting to not report any method calls in a lambda that is an argument to the specific`org.assertj.core.api.Assertions#assertThatThrownBy()` method. We should also recognize calls to `org.junit.jupiter.api.Assertions#assertDoesNotThrow()`, `org.junit.jupiter.api.Assertions#assertThrows()` and JUnit 4's `org.junit.Assert#assertThrows()`. In other assert methods the inspection should warn as usual.

If it can be done easily we could also support anonymous classes in addition to lambdas. But this is less important, so if it is too much trouble skip it for now.  

Adding extra options is undesirable for this case in my opinion. 

Note also that I have marked the original issue as a duplicate of another older issue.

Hopefully we not understand each other correctly and you can adjust your code. Thanks for your efforts so far.

pyltsin:
Thank you for this explanation! I rewrote it according to this idea.

PR #2231, opened 2022-11-15
KTIJ-23502 Replace with field intention leads to unresolved reference

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23502


abelkov (approved):
Thank you.

I think the highlighting range of "assign backing field" is too much, consider reducing it to only `set(value)` (could be done as a separate PR).

PR #2232, opened 2022-11-15
Do not copy stream if all bindings are constructor args

yuriy-solodkyy:
When deserializing an entity which is instantiated via a constructor with parameters in the general case we need to make a copy of the input stream so that the constructor arguments can eb deserialized first. However, when all the serialized values are constructor arguments we do not need to make a copy as it is slow and inefficient.

This changes saves more than a minute of data node deserialisation time on https://github.com/gavra0/android-extra-large-benchmark


yuriy-solodkyy:
@nskvortsov

PR #2233, opened 2022-11-15
IDEA-306015 disable the annoying error icon that keeps flashing the same error every twenty minutes

aSemy:
[IDEA-306015](https://youtrack.jetbrains.com/issue/IDEA-306015)

The icon does not serve a useful purpose. It has never alerted me to something actionable. It's just plain annoying.

### DIFF
--- a/platform/platform-impl/src/com/intellij/diagnostic/IdeErrorsIcon.java
+++ b/platform/platform-impl/src/com/intellij/diagnostic/IdeErrorsIcon.java
@@ -29,35 +29,5 @@ final class IdeErrorsIcon extends JLabel {
   }
 
   void setState(@NotNull MessagePool.State state) {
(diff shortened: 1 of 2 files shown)


yole:
If the icon does not serve a useful purpose for you personally, you can disable it via the system properties on your machine. It does serve a very useful purpose for IDE developers and many EAP users.

aSemy:
> you can disable it via the system properties on your machine.

This is not possible

yole:
The specific options for that are discussed in the linked YouTrack issue. In any case, removing the icon entirely is not an acceptable solution for the problems you're experiencing.

aSemy:
hey @yole, those options don't work

image

aSemy:
The option still doesn't work, I'm constantly pinged by this annoying icon

image

yole:
I'm sorry for the inconveniences you're experiencing, but they are still not a valid reason to accept your PR.

PR #2234, opened 2022-11-16
[Java. Inspections] IDEA-156786 - new inspection to check call math rounding with `int` argument

pyltsin:
I tried to implement https://youtrack.jetbrains.com/issue/IDEA-156786/Inspection-add-warning-on-Math.roundint

This case is simple and could be implemented with Structural Search, but it seems to me, that it could be useful because sometimes this case causes unpredictable results
For example:
```java
        System.out.println(-2144864817); 
        System.out.println(Math.round(-2144864817));

//results:
//-2144864817
//-2144864768
```

About realization:
- I'm not sure that it is a good idea to enable this inspection by default, so I turned it off
- About quick-fix, I implemented the next transformation 
```java
Integer i = 0;

Math.floor(i); //-> (double) (i)
Math.rint(i); //-> (double) (i)
Math.ceil(i); //-> (double) (i)
```

because of java-doc `If the argument value is already equal to a mathematical integer, then the result is the same as the argument`

But for `Math.round`, it is impossible. I think about `Math.round(i)`->`Math.round((double) (i))`, but it's more hiding the error, so I decided to do nothing in this case.

Thank you in advance


plugins/InspectionGadgets/src/META-INF/InspectionGadgets.xml
```diff
@@ -285,6 +285,10 @@
                      key="new.string.buffer.with.char.argument.display.name" groupBundle="messages.InspectionsBundle"
                      groupKey="group.names.probable.bugs" enabledByDefault="true" level="WARNING"
                      implementationClass="com.siyeh.ig.bugs.NewStringBufferWithCharArgumentInspection"/>
+    <localInspection groupPath="Java" language="JAVA" shortName="MathRoundingWithIntArgument" bundle="messages.InspectionGadgetsBundle"
+                     key="math.rounding.with.int.argument.display.name" groupBundle="messages.InspectionsBundle"
+                     groupKey="group.names.probable.bugs" enabledByDefault="false" level="WARNING"
```
BasLeijdekkers, line 290: In my opinion you should enable it by default. This seems like a useful inspection to me, which provides few but useful warnings.

In general it is basically not worth to implement an inspection if you are not going to enable it by default. It almost won't be used.

plugins/InspectionGadgets/src/inspectionDescriptions/MathRoundingWithIntArgument.html
```diff
@@ -0,0 +1,17 @@
+<html>
+<body>
+Reports calls to <code>round</code>, <code>ceil</code>,
```
BasLeijdekkers, line 3: We try to always refer to methods with parenthesis included. That way it is not necessary to read the entire sentence to see that it refers to methods (or calls).
So, `Reports calls to <code>round()</code>..` etc. please.

plugins/InspectionGadgets/src/inspectionDescriptions/MathRoundingWithIntArgument.html
```diff
@@ -0,0 +1,17 @@
...
+<p><b>Example:</b></p>
+<pre><code>
+  double d1 = Math.floor(2);
+</code></pre>
+<p>After the quick-fix is applied:</p>
+<pre><code>
+  double d1 = (double)(2);
+</code></pre>
+<!-- tooltip end -->
+</body>
```
BasLeijdekkers, line 16: Please add:
```
<p><small>New in 2023.1</small>
```
so users can find the new inspections between the thousands of older ones.

BasLeijdekkers (changes requested):

plugins/InspectionGadgets/src/META-INF/InspectionGadgets.xml
pyltsin: Thank you! Enabled by default

plugins/InspectionGadgets/src/inspectionDescriptions/MathRoundingWithIntArgument.html
pyltsin: Thank you! I didn't know that. Changed

plugins/InspectionGadgets/src/inspectionDescriptions/MathRoundingWithIntArgument.html
pyltsin: Added

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/MathRoundingWithIntArgumentInspection.java
```diff
@@ -0,0 +1,116 @@
...
+      }
+      final PsiExpressionList argumentList =
+        callExpression.getArgumentList();
+      final PsiExpression[] arguments = argumentList.getExpressions();
+      if (arguments.length != 1) {
+        return;
+      }
+      final PsiExpression argument = arguments[0];
+      final String text = argument.getText();
+      final String nextText = "(double)(" + text + ")";
```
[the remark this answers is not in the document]
pyltsin: Used `commentTracker.text(argument, PsiPrecedenceUtil.TYPE_CAST_PRECEDENCE)`
Thank you!

plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/bugs/MathRoundingWithIntArgumentInspection.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @NotNull
+    @Override
+    public String getFamilyName() {
+      return InspectionGadgetsBundle.message("math.rounding.with.int.argument.family.name");
+    }
+
+    @Override
+    protected void doFix(@NotNull Project project, @NotNull ProblemDescriptor descriptor) {
```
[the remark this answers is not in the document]
pyltsin: I completely agree with it!  Thank you!
I added removing cast if it is necessary

BasLeijdekkers (approved):

PR #2235, opened 2022-11-16
KTIJ-23683: Sort descriptors for single imports

andymagee:
`SingleImportVariant` effectively displays the first descriptor in the list it's given (with the caveat that it prefers `ClassDescriptors` over other descriptors). This change sorts the list before it's given to `SingleImportVariant`, using the priority that's already being calculated for the descriptors. Doing this ensures the highest priority (or rather, lowest numeric value priority) is used for display.


vladimirdolzhenko:
@andymagee could you pls raise a ticket and add tests ?

andymagee:
> @andymagee could you pls raise a ticket and add tests ?

There's already a ticket filed: https://youtrack.jetbrains.com/issue/KTIJ-23683

I went to write tests originally, but needed to ask guidance from whomever this got assigned to. Looks like you're the lucky one, @vladimirdolzhenko :)

I was able to find existing tests at [AddImportTestGenerated.java](https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/addImport/AddImportTestGenerated.java), but those all just verify what the modified file in the editor looks like after the 'Add Imports' action has run. I wasn't able to find any tests that validate what the popup itself contains, including what would be rendered by the individual descriptors (which is what this change is getting at). Do you know whether any such tests exist?

Alternately, I can add a unit test specifically for [KotlinAddImportAction.kt](https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt) which would specifically test `createSingleImportAction` and `createSingleImportActionForConstructor`. That wouldn't be testing the logic in context, though, so it may not be preferable.

Appreciate any guidance you can give.

vladimirdolzhenko:
Well, it would be nice to have a simple reproducible example (even manual) w/o android and other dependencies.

So, far I tried to create `kotlinpackage.two.s1.Img` (and marked as deprecated) and `kotlinpackage.two.s2.Img` - according to alphabetical order `s1` should be 1st but it goes last (even w/o your change) as it is deprecated.

andymagee:
The scenario here is actually slightly different than what you tested - it involves overloaded functions in the same package, not in different package. I added a comment to KTIJ-23683 along with a screenshot demonstrating the issue, using IDEA CE instead of Android Studio.

I can't provide a 100% deterministic repro, because it depends on the order descriptors are returned by `KtFile.resolveImportReference(fqName: FqName)` in `ResolutionUtils.kt`, which is undefined and can change. But if multiple of these overloads exist with some being deprecated and others not, I've found I can trigger the issue by playing around with adding/removing the deprecated annotation and adding/removing various overloads.

I've verified in IDEA CE that I can reproduce this as described in [this comment](https://youtrack.jetbrains.com/issue/KTIJ-23683/Add-Imports-in-Kotlin-displays-deprecated-import-even-when-non-deprecated-versions-exist#focus=Comments-27-6643927.0-0), and that I can no longer reproduce after this patch is applied.

andymagee:
(I should note, I'm absolutely open to writing tests for this, I just am stuck as to how to do that. With most of the import logic being internal it's difficult to write unit tests, and I'm not aware of any way to validate the logic in a more end-to-end fashion.)

andymagee:
Thanks for the pointers in https://youtrack.jetbrains.com/issue/KTIJ-23683/Add-Imports-in-Kotlin-displays-deprecated-import-even-when-non-deprecated-versions-exist#focus=Comments-27-6656769.0-0. I've added tests that validate this scenario.

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/actions/KotlinAddImportActionTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.test.TestMetadata
+import org.junit.internal.runners.JUnit38ClassRunner
+import org.junit.runner.RunWith
+import java.io.File
+
+@TestRoot("idea/tests")
+@RunWith(JUnit38ClassRunner::class)
+@TestMetadata("testData/idea/actions/kotlinAddImportAction")
+class KotlinAddImportActionTest : KotlinLightCodeInsightFixtureTestCase() {
```
vladimirdolzhenko, line 19: Usually we create smth like `AbstractKotlinAddImportActionTest` that has public `doTest` function and add it to [Fe10GenerateTests](https://github.com/JetBrains/intellij-community/blob/a79368860d6d3c7d558d861d40df319eca910b6e/plugins/kotlin/util/test-generator-fe10/test/org/jetbrains/kotlin/fe10/testGenerator/Fe10GenerateTests.kt#L793) 

it this case you can't forget to add/miss newly added test data cases.

you can regenerate tests with `Generate Kotlin Tests (K1 + K2)` run configurations.

plugins/kotlin/idea/tests/testData/idea/actions/kotlinAddImportAction/MultipleMethodsAvailable.kt
```diff
@@ -0,0 +1,10 @@
+// EXPECT_VARIANT_IN_ORDER public fun bar(): kotlin.Unit defined in root.q
```
vladimirdolzhenko, line 1: It would like to suggest to wrap a string with quotes, so smth like
```
EXPECT_VARIANT_IN_ORDER "public fun bar(): kotlin.Unit defined in root.q"
```

vladimirdolzhenko: Could you please add another instruction to check that deprecated method is not present here ?

vladimirdolzhenko (changes requested):

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/actions/KotlinAddImportActionTest.kt
andymagee: Thanks for the explanation! I had seen that that class existed, but didn't understand how it was used. This helps.

plugins/kotlin/idea/tests/testData/idea/actions/kotlinAddImportAction/MultipleMethodsAvailable.kt
andymagee: That expectation doesn't make sense in this file, since the deprecated method is displayed (just displayed last). But I have indeed added such an instruction to OverloadsPresent.kt, which represents that scenario.

vladimirdolzhenko:

PR #2238, opened 2022-11-18
KTIJ-15231 'Import members from package' shouldn't modify imports for enum 'values()' and 'valueOf()'

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-15231


vladimirdolzhenko:
LGTM

Thank you for the contribution

PR #2241, opened 2022-11-19
IDEA-306900 Support plugins that want to profile Gradle run configurations

ingokegel:
External build systems map the selected executor to their own implementation. Currently, only "Run" and "Debug" executors are registered in the `ExternalSystemUtil.RUNNER_IDS` map. A profiler plugin that wants to handle Gradle run configurations somehow has to provide a runner for this purpose. Currently, this requires the use of reflection. For example, the JProfiler plugin currently does this:

    val runnerIds = ExternalSystemUtil::class.java.getDeclaredField("RUNNER_IDS")
        .apply { isAccessible = true }
        .get(null) as MutableMap<String, String>
    runnerIds[RUNNER_ID] = RUNNER_ID

to ask the external build system to actually use its profiling executor (RUNNER_ID).

To avoid this, a method should be added to allow registering a runner ID mapping for external build systems.


yole:

PR #2242, opened 2022-11-21
[Java. Inspections]  IDEA-305501 - ignore continue in finally block

pyltsin:
I tried to implement https://youtrack.jetbrains.com/issue/IDEA-305501

About realization:

I did it similar to the current option - `Ignore in then branch of 'if' statement with 'else' branch`

In this code, I included and turned on the new option - `Ignore in 'finally' block`

Thank you in advance


BasLeijdekkers (changes requested):
Hi Mikhail,
I don' t think an option is needed in this case. A `continue` statement inside a `finally` block should just be ignored unconditionally.

pyltsin:
> Hi Mikhail, I don' t think an option is needed in this case. A `continue` statement inside a `finally` block should just be ignored unconditionally.

Hello, I changed it. Thank you!

PR #2247, opened 2022-11-23
ui: Use the correct constants when opening a split window

symmb:
Judging by the implementation, `EditorWindow.split` seems to expect either `HORIZONTAL_SPLIT` or `VERTICAL_SPLIT` from `JSplitPane` for its first argument, not any of the constants in `SwingConstants`.

NB: I haven't tested this myself because I didn't set up the full build environment for IntelliJ but it should work since `JSplitPane` is already imported via `javax.swing.*`.


symmb:
If needed I can also document that method although I'm not super familiar with it + your usual documentation style myself.

yole:

PR #2249, opened 2022-11-26
KTIJ-15170 Convert to block body action does not properly handle comments

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-15170


vladimirdolzhenko:
When following code is converted to the block body 
```
public fun baz(): Int =
    /*
     * baz
     */
    42
```

it becomes (correctly)
```
public fun baz(): Int {
    /*
     * baz
     */
    return 42
}
```

if after that, convert to the expression body, formatting is a bit broken
```
public fun baz(): Int =
    /*
        * baz
        */
    42
```

t-kameyama:
> So, it would be nice to add multiline block comment `/* ... */` case to test data

Fixed

vladimirdolzhenko:
LGTM, Thanks

PR #2253, opened 2022-11-28
Step into lambdas with parameters destructuring

nikita-nazarov:
The corresponding issue: https://youtrack.jetbrains.com/issue/KTIJ-23778

When stepping into a lambda with parameter destructuring a user would stop inside the lambda's body, but the parameters would still be uninitialized and couldn't be observed in the debugger. This happened because the debugger stopped on an opcode that comes before `component*` function calls. Now after the 'step into' these `component*` calls will be skipped until all the variables are initialized.

@madsager

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
+++ b/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
@@ -2,12 +2,14 @@
 
 package org.jetbrains.kotlin.idea.debugger.core.stepping
 
+import com.intellij.debugger.SourcePosition
 import com.intellij.debugger.engine.DebugProcess.JAVA_STRATUM
 import com.intellij.debugger.engine.DebugProcessImpl
 import com.intellij.debugger.engine.MethodFilter
 import com.intellij.debugger.engine.RequestHint
 import com.intellij.debugger.engine.SuspendContextImpl
 import com.intellij.debugger.engine.evaluation.EvaluateException
+import com.intellij.debugger.jdi.StackFrameProxyImpl
 import com.intellij.debugger.jdi.ThreadReferenceProxyImpl
 import com.intellij.openapi.application.runReadAction
 import com.intellij.openapi.diagnostic.Logger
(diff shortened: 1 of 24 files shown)


nikita-nazarov:
@yanex

jreznot:
FYI: runReadAction that takes 1 sec in code insight means that users cannot type in editor this 1 sec.

Ideally, no one should ask for `runReadAction {}` without constraints. Prefer coroutine based `readAction {}` if possible, it is non-blocking

yanex:
> FYI: runReadAction that might take 1 sec in shouldStepOverParameterDestructuring means that users cannot type in editor this 1 sec.

I'm not sure this can be easily fixed without changing the logic in `RequestHint.java` to non-blocking.

yanex:
Regarding the main change, I noticed the fix does not work for one-line lambdas.

```
data class A(val x: Int, val y: Int)

fun A.foo(f: (A) -> Unit) = f(this) // Breakpoint here

fun main() {
    val a = A(0, 0);
    a.foo { (x, y) -> println("$x$y") }
}
```

nikita-nazarov:
> FYI: runReadAction that might take 1 sec in shouldStepOverParameterDestructuring means that users cannot type in editor this 1 sec.

Maybe we can use `com.intellij.openapi.application.NonBlockingReadAction`?

nikita-nazarov:
> Regarding the main change, I noticed the fix does not work for one-line lambdas.
True! Thanks for pointing that out! The solution to this may be:
1. Collect all variables declared in a method from JDI
2. Collect all names of the destructured parameters using PSI 
3. Step over to the next location until all of the destructured parameters are visible according to LVT
4. + cornercase handling like `_` names
I'll figure that out.

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
```diff
@@ -187,6 +195,15 @@ class KotlinStepIntoRequestHint(
                 lastWasKotlinFakeLineNumber = false
                 return STOP
             }
+
+            // Step over `component*` function calls to let destructured parameters be initialized
+            if (location != null && shouldStepOverParameterDestructuring(context, frameProxy, location)) {
+                // Switching to minimal stepping size allows us to not skip locations of interest
+                // in case of one-line lambdas.
+                size = StepRequest.STEP_MIN
```
nikita-nazarov, line 203: We have to change the stepping size here, otherwise for cases like `foo { (x, y) -> println("$x$y") }` we will just step on the next line. However changing the stepping size to STEP_MIN in the constructor of `KotlinStepIntoRequestHint` will just slow down the stepping for all cases. Hence the stepping size switching.

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
```diff
@@ -187,11 +196,55 @@ class KotlinStepIntoRequestHint(
...
+    }
+
+    val visibleVariablesNames = InlineStackTraceCalculator.calculateVisibleVariables(frameProxy)
+        .mapNotNull { it.name() }
+        .toSet()
+    return destructuredParametersNames.any { it !in visibleVariablesNames }
+}
+
+private fun collectDestructuredParametersNames(sourcePosition: SourcePosition): List<String> {
+    val lambda = sourcePosition.elementAt?.parentOfType<KtFunctionLiteral>() ?: return emptyList()
```
yanex, line 238: Unlimited `parentOfType<KtFunctionLiteral>()` is dangerous. In the following case, stepping into the `run()` function broke.

```
import destructurableClasses.B

fun foo(f: (B) -> Unit) {
    val b = B()
    f(b)
}

fun main() {
    foo { (x, _, y, _) ->
        val task = object : Runnable {
            override fun run() {
                print("foo")
            }
        }
        // STEP_INTO: 1
        //Breakpoint!
        task.run()
    }
}
```

yanex (changes requested):

zuevmaxim:
Closed in favor of https://github.com/JetBrains/intellij-community/pull/2834

PR #2254, opened 2022-11-28
build-scripts: fix NPE when building dmg without bundled jre

didot:
When building with `-Dintellij.build.dmg.without.bundled.jre=true` on macOS, `jreArchivePath` is null.


yole:
This code has been rewritten in master, and unfortunately we aren't able to accept PRs that target a stable branch without an equivalent change in master.

PR #2258, opened 2022-11-30
KTIJ-1614 Don't highlight property access as a var when the setter is not visible

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-1614
スクリーンショット 2022-11-30 15 29 59

### DIFF
--- a/plugins/kotlin/idea/tests/testData/highlighter/VariablesWithSetter.kt
+++ b/plugins/kotlin/idea/tests/testData/highlighter/VariablesWithSetter.kt
@@ -1,0 +1,28 @@
+// EXPECTED_DUPLICATED_HIGHLIGHTING
+
+<info textAttributesKey="KOTLIN_BUILTIN_ANNOTATION">open</info> class <info textAttributesKey="KOTLIN_CLASS">C</info> {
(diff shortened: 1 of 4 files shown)


plugins/kotlin/idea/tests/testData/highlighter/VariablesWithSetter.kt
```diff
@@ -0,0 +1,28 @@
+// EXPECTED_DUPLICATED_HIGHLIGHTING
+
+<info textAttributesKey="KOTLIN_BUILTIN_ANNOTATION">open</info> class <info textAttributesKey="KOTLIN_CLASS">C</info> {
```
t-kameyama, line 3:
```kotlin
open class C {
    var foo: Int = 0
        private set
    var bar: Int = 0
        protected set
    var baz: Int = 0
        internal set
    var qux: Int = 0
        set
}

fun test(c: C) {
    c.foo
    c.bar
    c.baz
    c.qux
}

class Test : C() {
    fun test() {
        foo
        bar
        baz
        qux
    }
}
```

abelkov (approved):
Now **that** is an old bug!

PR #2260, opened 2022-12-01
KTIJ-22046: Fix J2K import resolve scope

gharrma:
Issue link: https://youtrack.jetbrains.com/issue/KTIJ-22046

Previously, PlainTextPasteImportResolver computed a resolve scope using the temporary file that holds pasted text. Instead it should compute a resolve scope using the target file, since that is where import statements will be inserted.

After this change, both 'scope' and 'resolutionFacade' are constructed from the same target file. Thus we should no longer see issues like KTIJ-22046 [in which the resolution facade was being queried for descriptors outside its domain].

Side note: I'm not sure how to add a test for this, and the repro cases are not so simple. However, I verified locally that this change fixes the case I saw.

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/conversion/copy/PlainTextPasteImportResolver.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/conversion/copy/PlainTextPasteImportResolver.kt
@@ -45,7 +45,7 @@ class PlainTextPasteImportResolver(private val dataForConversion: DataForConvers
     private val resolutionFacade = targetFile.getResolutionFacade()
 
     private val shortNameCache = PsiShortNamesCache.getInstance(project)
-    private val scope = file.resolveScope
+    private val scope = targetFile.resolveScope
 
     private val failedToResolveReferenceNames = HashSet<String>()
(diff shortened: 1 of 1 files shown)


darthorimar:
Hi Matthew, thank you for your contribution! 
The change seems to be logical tom. For adding tests, you may try using `NewJavaToKotlinCopyPasteConversionTestGenerated` on examples from the [KTIJ-22046](https://youtrack.jetbrains.com/issue/KTIJ-22046)

gharrma:
Here is the obstacle I hit when adding a regression test:

The repro scenario requires that `file.resolveScope` contains some symbols that `targetFile.resolveScope` does not. I don't think this scenario is possible to create in a `AbstractJ2kCopyPasteTest`, because this test fixture only supports creating projects with a single module and some libraries on its classpath.

For context, note that:
* `file.resolveScope` = Project and Libraries scope
* `targetFile.resolveScope` = Module-with-dependencies scope

Given the above — do you think the fix can be submitted without a test? Or is there some better place to add a more heavyweight, multi-module test?

darthorimar:
Yes, unfortunately, right now we cannot have multimodule tests for the J2K. I've merged the PR without the tests. 

PR #2261, opened 2022-12-01
KTIJ-23751 False positive "Suspicious callable reference as the only lambda element" with some return types

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23751


abelkov (changes requested):
Can you please add tests and support related cases:

* kotlin.Function, Function0, etc
* kotlin.reflect.KFunction
* kotlin.reflect.KMutableProperty, KMutableProperty0, etc

At least KFunction doesn't work:

```
import kotlin.reflect.KFunction

fun foo() {}

fun bar(f: () -> KFunction<Unit>) {}

val someFun = bar { <caret>::foo }
```

t-kameyama:
> Can you please add tests and support related cases:

Fixed with additional tests.

abelkov (approved):
Thank you.

I did a minor refactoring of you code: moved declarations from the companion object to the top level. It's better to avoid companion objects because they affect performance. Recently we removed many such companion objects from the Kotlin plugin inspection classes.

PR #2264, opened 2022-12-03
Fix inline return rendering

Matyrobbrt:
This PR fixes the Javadoc renderer in order to properly render the description, according to the [spec](https://docs.oracle.com/en/java/javase/16/docs/specs/javadoc/doc-comment-spec.html#return):
> As an inline tag, provides content for the first sentence of a method's description, and a "Returns" section, as if `@return` description were also present. In the default English locale, the first sentence is _Returns description ._

Before:
image
After:
image

Note: in order for [IDEA-300185](https://youtrack.jetbrains.com/issue/IDEA-300185) to be properly fixed, nested inline tags would need to be properly parsed


sirgl:
Thank you for the pull request! I'll merge it today!

BasLeijdekkers:
@sirgl did you get around to merging this?

PR #2266, opened 2022-12-07
KTIJ-23831 FINAL_SUPERTYPE, SUPERTYPE_NOT_INITIALIZED quickfixes: don't suggest with annotation supertype

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23831


abelkov (approved):
Thank you. I also fixed inline classes in addition to value classes.

PR #2269, opened 2022-12-09
Groovy: Fix erroneous remove parentheses from method call suggestion

SirYwell:
Consider following code:
```groovy
class Experiment {
    static void main(String[] args) {
        def v = a([0]) // relevant
        println v
    }
    static def a = [1, 2, 3]
    static def a(o) { -1 }
}
```
IntelliJ currently provides the option to remove the parentheses of the method call in the commented line. The resulting code snippet is `a[0]`, which is valid, even compiles and runs, but yields a different result.

The suggestion is also present in cases where `a[0]` doesn't make sense, e.g., if no variable `a` exists in the current scope.

This fix checks whether the first argument is a list or map literal, in which case the intention shouldn't be suggested.

I don't know if this PR needs a bug report on YouTrack, if it needs one, I'll happily open one. Please let me know.


SirYwell:
Is there anything to do from my side here? @dovchinnikov

SirYwell:
Hey @knisht, is there anything I can do to get this integrated?

knisht:
@SirYwell The change looks fine, thank you for the contribution. Sorry for the delay.

SirYwell:
Thanks @knisht. Could you also take a look at #2271?

PR #2271, opened 2022-12-12
Groovy: Fix parenthesizing of ranges in ForToEachIntention

SirYwell:
Previously,
```groovy
for (i in (0..<10)) {
    println i
}
```
was transformed into
```groovy
0..<10.each { i ->
    println i
}
```

However, this calls the `each` method on `10` instead of the range. To fix the transformation, the range needs parentheses. It seems like returning the range precedence fixes the issue.

I didn't find any existing test for the intention, so I created a new file, similar to the `EachToForIntentionTest` one.


knisht:

PR #2273, opened 2022-12-14
code stlye: remove maximum right margin

Bixilon:
This option is really annoying, I sometimes have lines that exceed that value (e.g. minified files). I never ever want them to be hard wrapped, this option allows this. If you do not need it, simply set the maximum hard wrap not that high, leave it at the default value (120)


rvishnyakov:
Removing the limit will break some formatting algorithms which check the fact that the line length has a reasonable limit. IMO, minified files should not be formatted at all and there is a way to avoid it.

PR #2274, opened 2022-12-14
IDEA-308603 Order binary expressions

Fabrice-TIERCELIN:
Hi @amaembo,

This PR introduces the `OrderedBinaryExpression` class. I have already written this when I contributed to Eclipse IDE. It reads a PSI binary expression and then you can read it in the order you want, no matter what was the original order. It is very useful. Tell me if there is a better package than `com.intellij.psi.util`.

This PR changes the *Redundant String operation* and *String.equals() can be replaced by String.isEmpty()* inspections. Tell me if you agree the new labels.

1. Go to *File* → *Settings...* → *Editor* → *Inspections*
2. Enable *Redundant String operation*
3. Enable Severity: *Warning*
4. Add this method in a Java 11 class:

```Java
  boolean isBlank(String text) {
    return "".equals(text.strip());
  }
```

The inspection *Redundant String operation* is suggested.
***What happens instead?***
The inspection *Redundant String operation* is not suggested.

`RedundantStringOperationInspection` should write:

```Java
text.isBlank()
```

...for the following inputs:

```Java
"".equals(text.strip())
text.strip().equals("")
text.strip().length() == 0
0 == text.strip().length()
text.strip().length() < 1
```

It should also handle negative forms.

I have added and successfully run unit tests.

### DIFF
--- a/plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
+++ b/plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
@@ -17,6 +17,7 @@ import com.intellij.psi.*;
 import com.intellij.psi.util.PsiLiteralUtil;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.psi.util.PsiUtil;
+import com.siyeh.HardcodedMethodConstants;
 import com.siyeh.InspectionGadgetsBundle;
 import com.siyeh.ig.InspectionGadgetsFix;
 import com.siyeh.ig.PsiReplacementUtil;
@@ -50,15 +51,15 @@ public class RedundantStringOperationInspection extends AbstractBaseJavaLocalIns
 
   private static final CallMatcher BYTE_ARRAY_OUTPUT_STREAM_INTO_BYTE_ARRAY =
     exactInstanceCall(JAVA_IO_BYTE_ARRAY_OUTPUT_STREAM, "toByteArray").parameterCount(0);
-  private static final CallMatcher STRING_TO_STRING = exactInstanceCall(JAVA_LANG_STRING, "toString").parameterCount(0);
+  private static final CallMatcher STRING_TO_STRING = exactInstanceCall(JAVA_LANG_STRING, TO_STRING).parameterCount(0);
   private static final CallMatcher STRING_INTERN = exactInstanceCall(JAVA_LANG_STRING, "intern").parameterCount(0);
-  private static final CallMatcher STRING_LENGTH = exactInstanceCall(JAVA_LANG_STRING, "length").parameterCount(0);
+  private static final CallMatcher STRING_LENGTH = exactInstanceCall(JAVA_LANG_STRING, HardcodedMethodConstants.LENGTH).parameterCount(0);
   private static final CallMatcher STRING_SUBSTRING_ONE_ARG = exactInstanceCall(JAVA_LANG_STRING, "substring").parameterTypes("int");
   private static final CallMatcher STRING_SUBSTRING_TWO_ARG = exactInstanceCall(JAVA_LANG_STRING, "substring").parameterTypes("int", "int");
   private static final CallMatcher STRING_SUBSTRING = anyOf(STRING_SUBSTRING_ONE_ARG, STRING_SUBSTRING_TWO_ARG);
   private static final CallMatcher STRING_BUILDER_APPEND =
     instanceCall(JAVA_LANG_ABSTRACT_STRING_BUILDER, "append").parameterTypes(JAVA_LANG_STRING);
-  private static final CallMatcher STRING_BUILDER_TO_STRING = instanceCall(JAVA_LANG_ABSTRACT_STRING_BUILDER, "toString").parameterCount(0);
+  private static final CallMatcher STRING_BUILDER_TO_STRING = instanceCall(JAVA_LANG_ABSTRACT_STRING_BUILDER, TO_STRING).parameterCount(0);
   private static final CallMatcher PRINTSTREAM_PRINTLN = instanceCall("java.io.PrintStream", "println")
     .parameterTypes(JAVA_LANG_STRING);
   private static final CallMatcher METHOD_WITH_REDUNDANT_ZERO_AS_SECOND_PARAMETER =
@@ -66,7 +67,7 @@ public class RedundantStringOperationInspection extends AbstractBaseJavaLocalIns
   private static final CallMatcher STRING_LAST_INDEX_OF = exactInstanceCall(JAVA_LANG_STRING, "lastIndexOf").parameterCount(2);
   private static final CallMatcher STRING_IS_EMPTY = exactInstanceCall(JAVA_LANG_STRING, "isEmpty").parameterCount(0);
   private static final CallMatcher CASE_CHANGE = exactInstanceCall(JAVA_LANG_STRING, "toUpperCase", "toLowerCase");
-  private static final CallMatcher STRING_EQUALS = exactInstanceCall(JAVA_LANG_STRING, "equals").parameterTypes(JAVA_LANG_OBJECT);
+  private static final CallMatcher STRING_EQUALS = exactInstanceCall(JAVA_LANG_STRING, HardcodedMethodConstants.EQUALS).parameterTypes(JAVA_LANG_OBJECT);
   private static final CallMatcher STRING_EQUALS_IGNORE_CASE =
     exactInstanceCall(JAVA_LANG_STRING, "equalsIgnoreCase").parameterTypes(JAVA_LANG_STRING);
   private static final CallMatcher CHANGE_CASE = anyOf(exactInstanceCall(JAVA_LANG_STRING, "toLowerCase").parameterCount(0),
@@ -251,10 +252,10 @@ public class RedundantStringOperationInspection extends AbstractBaseJavaLocalIns
       return false;
     }
 
+    @Nullable
     private ProblemDescriptor getRedundantCaseEqualsProblem(PsiMethodCallExpression call) {
-
       PsiExpression equalTo = PsiUtil.skipParenthesizedExprDown(call.getArgumentList().getExpressions()[0]);
-
+      if (equalTo == null) return null;
       //case: "foo".equals(s.toLowerCase())
       if (equalTo instanceof PsiMethodCallExpression equalsToCallExpression &&
           isChangeCaseCall(equalsToCallExpression, call.getMethodExpression().getQualifierExpression())) {
--- a/plugins/InspectionGadgets/src/com/siyeh/ig/style/SizeReplaceableByIsEmptyInspection.java
+++ b/plugins/InspectionGadgets/src/com/siyeh/ig/style/SizeReplaceableByIsEmptyInspection.java
@@ -15,6 +15,7 @@ import com.intellij.pom.java.LanguageLevel;
 import com.intellij.psi.*;
 import com.intellij.psi.tree.IElementType;
 import com.intellij.psi.util.InheritanceUtil;
+import com.siyeh.ig.psiutils.OrderedBinaryExpression;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.psi.util.PsiUtil;
 import com.intellij.util.SmartList;
@@ -83,14 +84,11 @@ public class SizeReplaceableByIsEmptyInspection extends BaseInspection {
 
     @Override
     protected void applyFix(@NotNull Project project, @NotNull PsiElement startElement, @NotNull EditorUpdater updater) {
-      final PsiBinaryExpression binaryExpression = (PsiBinaryExpression)startElement;
-      PsiExpression operand = PsiUtil.skipParenthesizedExprDown(binaryExpression.getLOperand());
-      if (!(operand instanceof PsiMethodCallExpression)) {
-        operand = PsiUtil.skipParenthesizedExprDown(binaryExpression.getROperand());
-      }
-      if (!(operand instanceof PsiMethodCallExpression methodCallExpression)) {
+      final OrderedBinaryExpression<PsiMethodCallExpression, PsiExpression> orderedBinaryExpression = OrderedBinaryExpression.from((PsiExpression)startElement, PsiMethodCallExpression.class);
+      if (orderedBinaryExpression == null) {
         return;
       }
+      final PsiMethodCallExpression methodCallExpression = orderedBinaryExpression.getFirstOperand();
       final PsiReferenceExpression methodExpression = methodCallExpression.getMethodExpression();
       final PsiExpression qualifierExpression = methodExpression.getQualifierExpression();
       if (qualifierExpression == null) {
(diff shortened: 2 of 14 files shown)


java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+    return null;
+  }
+
+  private static <T extends PsiExpression> T as(PsiExpression input, Class<T> aClass) {
+    return tryCast(PsiUtil.skipParenthesizedExprDown(input), aClass);
+  }
+
+  private static IElementType mirrorOperator(final PsiBinaryExpression expression) {
+    if (Arrays.asList(
```
amaembo, line 100: Why allocating a list on every call? Let's extract it to a constant.

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
```diff
@@ -0,0 +1,122 @@
+// Copyright 2000-2022 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.psi.util;
```
amaembo, line 2: Please put it near com.siyeh.ig.psiutils.ExpressionUtils

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private static <T extends PsiExpression> T as(PsiExpression input, Class<T> aClass) {
+    return tryCast(PsiUtil.skipParenthesizedExprDown(input), aClass);
+  }
+
+  private static IElementType mirrorOperator(final PsiBinaryExpression expression) {
+    if (Arrays.asList(
+      JavaTokenType.AND,
+      JavaTokenType.ANDAND,
```
amaembo, line 102: I would not say that ANDAND and OROR are commutative operations. Let's skip them for now?

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
```diff
@@ -0,0 +1,122 @@
...
+      JavaTokenType.OROR,
+      JavaTokenType.EQEQ,
+      JavaTokenType.NE,
+      JavaTokenType.OR,
+      JavaTokenType.PLUS,
+      JavaTokenType.ASTERISK,
+      JavaTokenType.XOR).contains(expression.getOperationTokenType())) {
```
amaembo, line 109: Let's extract `expression.getOperationTokenType()` to a variable. It's not like a simple getter.

amaembo: Or better pass it as a parameter to this method instead of `expression`.

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
```diff
@@ -0,0 +1,122 @@
...
+   * Return the items of a binary expression in the order it is specified. It reverses the operator if needed.
+   *
+   * @param <F>          the required expression type
+   * @param <S>          the required expression type
+   * @param node         the supposed binary expression
+   * @param firstClass   the class representing the required expression type
+   * @param secondClass  the class representing the required expression type
+   * @return the items of a binary expression in the order it is specified. It reverses the operator if needed.
+   */
+  public static <F extends PsiExpression, S extends PsiExpression> OrderedBinaryExpression<F, S> from(final PsiExpression node, final Class<F> firstClass, final Class<S> secondClass) {
```
amaembo, line 75: Please add explicit Nullable/NotNull annotations to every new public method

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
```diff
@@ -131,6 +134,55 @@ public void visitMethodCallExpression(@NotNull PsiMethodCallExpression call) {
...
+      if (orderedBinaryExpression == null) {
+        return;
+      }
+      final PsiMethodCallExpression lengthCallExpression = orderedBinaryExpression.getFirstOperand();
+      if (!STRING_LENGTH.test(lengthCallExpression)) {
+        return;
+      }
+      final PsiReferenceExpression methodExpression = lengthCallExpression.getMethodExpression();
+      final PsiExpression qualifierExpression = methodExpression.getQualifierExpression();
+      if (!(qualifierExpression instanceof PsiMethodCallExpression stripCallExpression)) {
```
amaembo, line 154: Is com.siyeh.ig.psiutils.MethodCallUtils#getQualifierMethodCall suitable here?

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
```diff
@@ -131,6 +134,55 @@ public void visitMethodCallExpression(@NotNull PsiMethodCallExpression call) {
...
+        return;
+      }
+      if (!STRIP.test(stripCallExpression)) {
+        return;
+      }
+      String message = InspectionGadgetsBundle.message("expression.can.be.replaced.problem.descriptor", "isBlank()");
+      final Object object = ExpressionUtils.computeConstantExpression(orderedBinaryExpression.getSecondOperand());
+      if (!(object instanceof Integer) && !(object instanceof Long) && !(object instanceof Short)) {
```
amaembo, line 162: I believe, it cannot be Short. And I'm not sure whether it's good to issue a warning when there's a non-trivial constant computation. Let's stay with literals like 0 and 1 only.

amaembo: We also have com.siyeh.ig.psiutils.JavaPsiMathUtil#getRangeFromComparison which can help here. It will return [1..MAX_VALUE] for `>= 1` and `>0`. You can intersect the result with non-negative numbers (JvmPsiRangeSetUtil#indexRange), then check if the result is exactly 0, or it's 1..MAX_VALUE.

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
```diff
@@ -109,7 +112,7 @@ private static class RedundantStringOperationVisitor extends JavaElementVisitor
       .register(STRING_LAST_INDEX_OF, this::getLastIndexOfProblem)
       .register(STRING_IS_EMPTY, this::getRedundantCaseChangeProblem)
       .register(STRING_EQUALS, this::getRedundantSubstringEqualsProblem)
-      .register(anyOf(STRING_EQUALS, STRING_EQUALS_IGNORE_CASE), this::getRedundantCaseEqualsProblem)
```
amaembo, line 112: Do you really need to route to `getRedundantCaseEqualsProblem` by yourself? I guess, `CallMapper` would take care of this. You can register two handlers for the same CallMatcher.

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
```diff
@@ -881,6 +960,92 @@ public void applyFix(@NotNull Project project, @NotNull ProblemDescriptor descri
     }
   }
 
+  private static class StripEqualsToIsBlankFix implements LocalQuickFix {
+    private final PsiExpression wholeExpression;
```
amaembo, line 964: Quick-fix cannot store PSI elements in the fields. I believe there should be a devkit inspection that warns about this

plugins/InspectionGadgets/src/com/siyeh/ig/style/SizeReplaceableByIsEmptyInspection.java
```diff
@@ -162,28 +155,35 @@ else if (rhs instanceof PsiMethodCallExpression) {
       if (isEmptyCall == null) {
         return;
       }
-      final Object object = ExpressionUtils.computeConstantExpression(flipped ? lhs : rhs);
-      if (!(object instanceof Integer) || ((Integer)object).intValue() != 0) {
+      final Object object = ExpressionUtils.computeConstantExpression(orderedBinaryExpression.getSecondOperand());
+      if (!(object instanceof Integer) && !(object instanceof Long) && !(object instanceof Short)) {
```
amaembo, line 159: Short is unnecessary.

plugins/InspectionGadgets/src/com/siyeh/ig/style/SizeReplaceableByIsEmptyInspection.java
```diff
@@ -162,28 +155,35 @@ else if (rhs instanceof PsiMethodCallExpression) {
...
+        }
+        if (JavaTokenType.NE.equals(tokenType) || JavaTokenType.GT.equals(tokenType)) {
+          registerError(expression, '!' + isEmptyCall, qualifierExpression, referenceName);
+        }
       }
-      else if (flipped) {
+      else if (comparedValue == 1) {
+        final IElementType tokenType = orderedBinaryExpression.getTokenType();
         if (JavaTokenType.LT.equals(tokenType)) {
+          registerError(expression, isEmptyCall, qualifierExpression, referenceName);
```
amaembo, line 178: Do we have any new functionality in this inspection? Or only refactoring?

amaembo (changes requested):

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
Fabrice-TIERCELIN: Done

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
Fabrice-TIERCELIN: In this review fix, I haven't put those operators aside because if we do so, it won't work at all for those operators (we will have to return `null`) but the user will use it in full knowledge as it is the point of the method. So what do we do?

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
Fabrice-TIERCELIN: Done

java/java-psi-api/src/com/intellij/psi/util/OrderedBinaryExpression.java
Fabrice-TIERCELIN: Done

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
Fabrice-TIERCELIN: 1. Done
2. I think it's not usable here. I think it works only with a previous `if` statement, not in a boolean expression. I think it will return `[MIN_VALUE..MAX_VALUE]`.

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
Fabrice-TIERCELIN: Done

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
Fabrice-TIERCELIN: Done

plugins/InspectionGadgets/src/com/siyeh/ig/style/SizeReplaceableByIsEmptyInspection.java
Fabrice-TIERCELIN: Done

plugins/InspectionGadgets/src/com/siyeh/ig/style/SizeReplaceableByIsEmptyInspection.java
Fabrice-TIERCELIN: New cases using `1` have been added.

java/java-tests/testData/inspection/redundantStringOperation/BAOStoString.java
```diff
@@ -12,7 +12,7 @@ void charset() {
     Charset charset = Charset.defaultCharset();
 
     ByteArrayOutputStream baos = new ByteArrayOutputStream();
-    String s = new String(baos.toByteArray(), charset);
+    String s = new <warning descr="Inefficient conversion from ByteArrayOutputStream">String</warning>(baos.toByteArray(), charset);
```
amaembo, line 15: You are breaking this test. It was intentionally made for Java 9, see https://youtrack.jetbrains.com/issue/IDEA-258683. It should be executed with Java 9 language level.

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
```diff
@@ -131,6 +135,54 @@ public void visitMethodCallExpression(@NotNull PsiMethodCallExpression call) {
       myProcessors.mapAll(call).forEach(myHolder::registerProblem);
     }
 
+    @Override
+    public void visitBinaryExpression(@NotNull PsiBinaryExpression expression) {
```
amaembo, line 139: Let's remove this part. This is already covered by "s.length() == 0 -> s.isEmpty()" conversion and "s.strip().isEmpty() -> s.isBlank()" conversion, so users can perform this in two steps. No need to maintain additional chunk of code to make this conversion in a single step.

Note that SizeReplaceableByIsEmpty will be on by default in the next IntelliJ IDEA version.

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
```diff
@@ -251,10 +303,26 @@ private static boolean isNewStringFromByteArrayParams(PsiExpression[] args) {
       return false;
     }
 
-    private ProblemDescriptor getRedundantCaseEqualsProblem(PsiMethodCallExpression call) {
-
+    private ProblemDescriptor getStripAndBlankProblem(PsiMethodCallExpression call) {
       PsiExpression equalTo = PsiUtil.skipParenthesizedExprDown(call.getArgumentList().getExpressions()[0]);
+      if (equalTo == null) return null;
```
amaembo, line 308: Same here: we already have `Java | Performance | 'String.equals()' can be replaced with 'String.isEmpty()'`, so `.equals("")` will be replaced with `isEmpty()` by another inspection. No need to support both steps at once.

amaembo (changes requested):

java/java-tests/testData/inspection/redundantStringOperation/BAOStoString.java
Fabrice-TIERCELIN: Reverted

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
Fabrice-TIERCELIN: Removed

plugins/InspectionGadgets/src/com/siyeh/ig/redundancy/RedundantStringOperationInspection.java
Fabrice-TIERCELIN: Removed

amaembo:
@Fabrice-TIERCELIN have you pushed new changes? I don't see them. Sorry if I'm missing something.

Fabrice-TIERCELIN:
You've missed nothing :) Now it's pushed.

Fabrice-TIERCELIN:
Hi @amaembo,

You can re-review this PR. All changes requested are done. I don't know why Github still show pending requests.

amaembo:
Merged with minor fixes 94c43389e6a45c2a74917a0793a7eb5272ce5a8b

PR #2279, opened 2022-12-27
Introduce new heuristic to detect function literal that matches current location

nikita-nazarov:
Issues that are covered by this pr:
- https://youtrack.jetbrains.com/issue/KTIJ-23964/Debugger-smart-stepping-inside-inline-lambdas-declared-in-a-local-function-doesnt-work-for-IR-backend
- https://youtrack.jetbrains.com/issue/KTIJ-23965/Debugger-smart-stepping-into-inline-lambdas-doesnt-work-when-lambdas-are-declared-in-functions-with-the-same-name
- https://youtrack.jetbrains.com/issue/KTIJ-24090/Debugger-inline-lambda-breakpoints-sometimes-doesnt-work
- https://youtrack.jetbrains.com/issue/KTIJ-24091/Debugger-smart-stepping-steps-into-a-wrong-inline-lambda-on-the-same-line

Previously KotlinPositionManager relied on ClassNameCalculator to detect which function literal on a particular line matches the current jdi location. This lead to issues when ClassNameCalculator failed to compute class names correctly. For example, ClassNameCalculator computes class names of local functions according to the old backend naming scheme, which doesn't work for the IR backend. However, fixing ClassNameCalculator will not solve the issue, since there is no way to understand at runtime which backend was used for the compilation. Also, if the compiler naming scheme changes in the future, ClassNameCalculator will have to be updated again and again. This pr introduces a new heuristic to detect which inlined function literal corresponds to the given jdi location. The algorithm is based on inline marker variables borders, and while it doesn't solve the ClassNameCalculator issue completely, it still improves smart stepping and lambda breakpoints in a couple of cases.

Roughly speaking, the algorithm is based on the following logic:

Consider the example:
```
1.let { it }.let { it + 1 }.let { it + 2 } // Line X
```
Given a jdi location, we need to understand which inline lambda is the best fit for it. After checking variables that are visible on the location, we will see these marker variables start on line X:
```
$i$a-let-...$...$1  // 1 -- { it }
$i$a-let-...$...$2  // 2 -- { it + 1 }
$i$a-let-...$...$3  // 3 -- { it + 2 }
```
Each of these variables has visibility borders, and, for example, if the borders of variable 2 contain the given location, it means that we are in the `{ it + 1 }` lambda. 

This approach relies on inline variables borders instead of relying on ClassNameCalculator entirely. However, it still needs ClassNameCalculator to detect which inlined lambdas are 'visible' in the current context. For example:
```
inline fun foo(f: () -> Unit) = f()
fun bar(f: () -> Unit) = f()

fun main() {
    foo { bar { foo { println() } } } // Line X
}
``` 
In this case we have two inline lambdas on line X, however only one marker variable will be visible in context of `main()` and `bar.invoke()` functions. To handle cases like this, the algorithm relies on ClassNameCalculator to detect if we are inside a non-inline lambda or not.

@madsager


yanex:
It seems in some cases the new behavior results in worse stepping:

```
inline fun foo() {
    1.let { it }.let { it + 1 }.let { it + 2 }
}

fun main() {
    foo()
    // Breakpoint!
    foo()
}
```

Step into `foo()`, and then smart step into `{ it + 1 }`. Before the debugger went into `{ it }`, and after the change it jumps to `Thread.stop()`.

yanex:
One more case:

```
inline fun foo(a: Boolean, b: Boolean, c: Boolean) {
    1.letIf(a) { it }.letIf(b) { "foo $it" }.letIf(c) { "bar $it" }
}

inline fun <T, R> T.letIf(condition: Boolean, block: (T) -> R?): R? {
    return if (condition) block(this) else null
}

fun main() {
    foo(true, true, false)
}
```

As in [KTIJ-24091](https://youtrack.jetbrains.com/issue/KTIJ-24091/Debugger-smart-stepping-steps-into-a-wrong-inline-lambda-on-the-same-line), smart step into 2nd and 3rd lambdas doesn't work, yet before it worked for the first lambda.

yanex (changes requested):

nikita-nazarov:
Before fixing the comments, I would like to talk a little about issues that you mentioned, because there is a lot going on there.
Let's start with this one:
```
inline fun foo() {
    1.let { it }.let { it + 1 }.let { it + 2 } // Smart step into { it + 1 }
}

fun main() {
    foo()
    // Breakpoint!
    foo()
}
```
The stepping behaviour here changed because of this commit: 71fc686b3c57c9af5abaad141d6763e1a7d5d958. Previously we stopped in `{ it }` because while trying to smart step into `{ it + 1 }`, we crossed the fake line number, and `KotlinStepIntoRequestHint` decided to stop smart stepping because of it. The mentioned commit prevents such behaviour, that's why we won't stop in `{ it }`. 

Why it doesn't work as expected?
1. The first reason is related to https://youtrack.jetbrains.com/issue/KTIJ-24239/Debugger-stepping-out-of-inline-lambdas-on-one-line-doesnt-work, because of that issue when trying to step out of `{ it }`, we will just continue stepping until we reach the next line. This issue is very easy to fix and I will upload the commit to this pr.
2. The second reason is related to the fundamental flaw in the smart stepping mechanism. When we try to smart step into a lambda, we set a `StepIntoBreakpoint` on a location inside of this lambda. After that we start the smart stepping process, and eventually the debugger will hit that breakpoint, and the desirable lambda will be reached. This works well for Java lambdas and ordinary Kotlin lambdas. Things get worse when we are trying to smart step into inline lambda. In the above example each of the inlined sites of the `foo` function contains locations inside the lambda of interest. However, only one `StepIntoBreakpoint` is set, this can be seen here `com/intellij/debugger/ui/breakpoints/StepIntoBreakpoint.java:74`. It means that smart stepping will only work for the first inlined site of the `foo` function.

Speaking of this example:
```
inline fun foo(a: Boolean, b: Boolean, c: Boolean) {
    1.letIf(a) { it }.letIf(b) { "foo $it" }.letIf(c) { "bar $it" }
}

inline fun <T, R> T.letIf(condition: Boolean, block: (T) -> R?): R? {
    return if (condition) block(this) else null
}

fun main() {
    // Breakpoint!
    foo(true, true, false)
}
```
The two described issues persist here, and also there is another one.  
The compiler produces only two inline lambda marker variables for the main function:
```
$i$a$-letIf-SandboxKt$foo$1$iv
$i$a$-letIf-SandboxKt$foo$2$iv
```
It means, that the new heuristic, introduced in b7995f418278cf6463e64962ed00266d73b9f2e9, will not be able to detect which location belongs to which lambda. The code will just return null here https://github.com/nikita-nazarov/intellij-community/blob/b7995f418278cf6463e64962ed00266d73b9f2e9/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinPositionManager.kt#L341, since the number of marker variables doesn't match the number of lambdas on a line.

I'm not quite sure how to handle this situation. We can preserve the old smart stepping behaviour in these cases by reverting 71fc686b3c57c9af5abaad141d6763e1a7d5d958. I also can fix the two mentioned smart stepping issues, and add a fallback to `ClassNameCalculator` when the number of marker variables is smaller than the number of lambdas on a line. 

What do you think?

zuevmaxim:
All the mentioned issues were resolved since then, so this PR can be closed too. @nikita-nazarov Please, correct me if I'm wrong

PR #2280, opened 2022-12-29
Add catch for potentially missing directories

jrlogsdon:
When using the IconsClassGenerator script for generating icons and one of the modules depends on a module that gets dynamically generated, the script will crash out, as it may not exist. Needed for Android.

Change-Id: I0a2d9a0cbc26649bf429fef8329c6b0ab4de2131


andrei-kuznetsov:
@jrlogsdon, I believe the problem is not in this script, but in broken project structure. Please take a look at https://git.jetbrains.org/?p=idea/android.git;a=commitdiff;h=36d5d8a58e1a0ceb34c045c2619299f23478b7c7

Why do you have these directories in the first place?

jrlogsdon:
These will be removed when making the icon swap change, but there's still another module that relies on a Bazel generated module.

PR #2281, opened 2022-12-30
Make SourceSearcher an extension point

lppedd:
This allows third parties to contribute ways to retrieve sources for possibly non-standard JAR files.
Building block for https://github.com/JetBrains/intellij-community/pull/2127

A question: can we delegate to `SourceSearcher`s the extraction of `groupId`, `artifactId`, and `version`?
This is currently done in `InternetAttachSourceProvider`, but the logic used is not always correct.

Usage example might be `P2SourceSearcher`, which finds source JARs in p2 repositories.


lppedd:
@vladsoroka just wondering what's your opinion on this. If you think it's ok/something useful I'll rebase and solve conflicts.

PR #2282, opened 2023-01-03
Add Atom key bindings

kiasquare:
Adding Atom's default key map to be able to use my intuitive key bindings in PhpStorm.


lppedd:
I would put this into a plugin. Keymaps need maintenance too so better delegate that to a third party artifact.

Check the `com.intellij.bundledKeymap` EP.

PR #2283, opened 2023-01-04
KTIJ-24088 "Replace 'invoke' with direct call" intention: false positive on callable 'invoke' property

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-24088


abelkov (approved):
Thank you, merged. Happy new year!

PR #2285, opened 2023-01-09
[KTIJ-24126] Fix transitive module dependencies not being recognized

Nxllpointer:
Fixes [KTIJ-24126](https://youtrack.jetbrains.com/issue/KTIJ-24126)
```
Title: MPP / IDE: "UNRESOLVED_REFERENCE" Common code of multi variant library not recognized in library consumer
Project: KTIJ
Type: Bug
Subsystems: IDE. Highlighting
Priority: Major
```

So I tracked down the issue for this being is that when `LibraryDataNodeSubstitutor` (The class responsible for figuring out which dependency files are produced by which module and using the module as dependency instead) does not appear to consider transitive module dependencies of the found module being depended on.

I don't think this change conflicts with anything or breaks other parts of IntelliJ. Please keep in mind that I have no experience working with the IntelliJ codebase and this fix might be in the wrong place or a bad solution.

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/LibraryDataNodeSubstitutor.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/LibraryDataNodeSubstitutor.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.util.*;
 
 import static com.intellij.openapi.externalSystem.util.ExternalSystemApiUtil.find;
+import static com.intellij.openapi.externalSystem.util.ExternalSystemApiUtil.findAll;
 import static org.jetbrains.plugins.gradle.service.project.GradleProjectResolverUtil.attachGradleSdkSources;
 import static org.jetbrains.plugins.gradle.service.project.GradleProjectResolverUtil.attachSourcesAndJavadocFromGradleCacheIfNeeded;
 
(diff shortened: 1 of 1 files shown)


abelkov:
cc @sellmair

nskvortsov:
@Nxllpointer thank you for your help! Unfortunately, I can not apply suggested modification, as it alters the substitution algorithm, while the fix should be done around building proper library dependencies tree in [KotlinMPPGradleProjectResolver](https://github.com/JetBrains/intellij-community/blob/470eb2007e1ddeedbd2fc9816217489bd3c5fe85/plugins/kotlin/gradle/gradle-java/src/org/jetbrains/kotlin/idea/gradleJava/configuration/KotlinMPPGradleProjectResolver.kt)

@sellmair my wild guess is that some transitive information may be lost when creating the dependency data node for MPP things around here: 
https://github.com/JetBrains/intellij-community/blob/470eb2007e1ddeedbd2fc9816217489bd3c5fe85/plugins/kotlin/gradle/gradle-java/src/org/jetbrains/kotlin/idea/gradleJava/configuration/KotlinMPPGradleProjectResolver.kt#L688-L693

sellmair:
Thanks for the contribution. 
I will look into the issue today!

Nxllpointer:
@nskvortsov hm ok I see

Nxllpointer:
Awesome! Thanks @sellmair

sellmair:
Answered on the original Ticket. 
Issue is confirmed, but already fixed in upcoming releases!

PR #2287, opened 2023-01-13
Set IS_ENSURE_NEWLINE_AT_EOF = true by default

dahlstrom-g:
With this option false, the editor omits newlines at EOF
for both existing and newly created files. This is bad behavior.
Here are some explanations why:

joachimschuster.de/posts/why-inserting-newline-at-the-end-of-file
thoughtbot.com/blog/no-newline-at-end-of-file
medium.com/@alexey.inkin/fdf76d1d090e

Here are some complaints about this behavior in IntelliJ-based IDEs:

itecnote.com/tecnote/android-how-to-configure-android-studio-to-add-new-line-at-end-of-file
reddit.com/r/Hyperskill/comments/h9pyyr/warning_pep_8_w292_no_newline_at_end_of_file_in
stackoverflow.com/questions/36043061#comment67201459_37260222

Users can still override the default by unchecking the box
or by setting insert_final_newline=false in .editorconfig.

### DIFF
--- a/platform/ide-core-impl/src/com/intellij/openapi/editor/ex/EditorSettingsExternalizable.java
+++ b/platform/ide-core-impl/src/com/intellij/openapi/editor/ex/EditorSettingsExternalizable.java
@@ -53,7 +53,7 @@ public class EditorSettingsExternalizable implements PersistentStateComponent<Ed
     public boolean IS_VIRTUAL_SPACE = false;
     public boolean IS_CARET_INSIDE_TABS;
     @NonNls public String STRIP_TRAILING_SPACES = STRIP_TRAILING_SPACES_CHANGED;
-    public boolean IS_ENSURE_NEWLINE_AT_EOF = false;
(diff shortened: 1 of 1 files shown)


dahlstrom-g:
@shafirov @andrei-kuznetsov

vobal-jb:
Hello,

thank you for the constructive PR.

There are also some cons to changing this default: the IDE does something automatically which the user did not ask for and might not understand why. This automatic change can introduce undesired VCS changes as this setting is not shared through VCS. In combination with autosave such changes can become very annoying (and even impossible to revert without changing settings).

The arguments you link are compelling, however, our statistics show that the overwhelming majority of users leave this setting disabled. It seems it is better to leave it as it is.

vobal-jb:
FWIW, on the implementation: We store user settings as diffs from the defaults, meaning that your commit would also enable this setting for users who explicitly want it disabled.

PR #2288, opened 2023-01-13
Fix a potential resource leak bug.

ruba128:
Potential resource leak bug when "asdf" is a valid resource.


akozlova:
That's test data for java analysis, code won't be run anywhere. Please don't change it. Thanks

PR #2291, opened 2023-01-15
Multiple inlay author providers per language

olivernybroe:
This makes it possible for the same language to have multiple `com.intellij.codeInsight.hints.VcsCodeVisionLanguageContext`. The main scenario where this is useful is for plugin developers if the default implementation for the language does not provide support for the needed scenario.


sirgl:
I'm rather sceptical about it. Is it absolutely impossible to influence the behavior of the default implementation?

olivernybroe:
@sirgl yes, so it my case it is for php. 
The php plugin has it's own implementation of the interface and my plugin for php needs to add support for VCS Code vision in a specific context for a testing framework. 

At least from all of my testing there is not way to override the default implementation or extend on it.

PR #2299, opened 2023-01-20
KTIJ-20298 "Call chain on collection type may be simplified" to replace `map` and `sum` with `sumOf` produces "overload resolution ambiguity"

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-20298
https://youtrack.jetbrains.com/issue/KTIJ-24252

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
@@ -16,13 +16,12 @@ import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.anyDescendantOfType
-import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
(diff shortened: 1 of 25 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
```diff
@@ -97,25 +101,48 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
...
-                    ?.mapNotNull { it.returnedExpression?.getType(context) }
-                    .orEmpty()
-                val types = listOfNotNull(lastStatementType) + returnedTypes
-                types.firstOrNull { it.isNullable() } ?: types.firstOrNull()
+                val lastStatement = argumentExpression.bodyExpression?.lastBlockStatementOrThis()
+                FunctionalArgument(
+                    isLambda = true,
+                    returnType = lastStatement?.getType(context),
+                    lastStatement = lastStatement
+                )
```
t-kameyama, line 115: ReturnExpression is checked by `AbstractCallChainChecker#findQualifiedConversion`, so no need to check it here.

https://github.com/JetBrains/intellij-community/blob/4430b3bfe94a48bd1982f74b331da39c36e6373d/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/collections/AbstractCallChainChecker.kt#L47-L49

abelkov (approved):
Nice 👍

PR #2289, opened 2023-01-13
Fix the issue when debugger is stuck in an infinite loop in Android project

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-24259/Debugger-is-stuck-in-an-infinite-loop-in-an-Android-project

It could happen that the `replacements` map for an arbitrary key contained a value that is equal to that key. This resulted in an infinite loop, which affected all the debugger behaviour.

PR #2300, opened 2023-01-20
kotlin-jvm-debugger: Fix workarounds for local variables on dex

sfs:
The code in `stackFrameUtils.kt` was sensitive to the order of local variables and would get stuck in an infinite loop when locals were sorted in ascending order of start offset. This is currently the case, which breaks Kotlin debugging on Android.

This commit fixes this issue and also updates the dex specific debugger test cases.

---

I'm not sure which base branch to use currently for Kotlin plugin changes. Please let me know if I should rebase this on top of a different branch.


madsager:
@nikita-nazarov looked into the same issue and uploaded a similar fix here: https://github.com/JetBrains/intellij-community/pull/2289

@sfs and @nikita-nazarov, can the two of you coordinate which of these pull requests we should move forward with?

sfs:
I think we should go with #2289 for the fix. I'll change this PR to only include the updated tests, which are less critical.

zuevmaxim:
@sfs Is this change still relevant?

nikita-nazarov:
@zuevmaxim I think we can close this pull request now

PR #2304, opened 2023-01-22
New project wizard: add a new type of generator asset for arbitrary file content

Earthcomputer:
This PR adds `GeneratorFile` to the list of generator asset types, so that you can add more flexible types of assets with `AssetsNewProjectWizardStep`, such as files generated by code rather than from a template or a resource. Feel free to suggest a better class name, I couldn't think of one. Also feel free to suggest other improvements as this is my first intellij PR.


jreznot:
Hi,
Thank you for the contribution, please add some tests.
Read more on testing : https://plugins.jetbrains.com/docs/intellij/testing-plugins.html

In order to include such new APIs we either need significant use in our code base covered with tests or additional tests that cover new APIs

PR #2308, opened 2023-01-28
pytest: propagate test's location to locationHint

ikonst:
Currently the plugin generates a `locationHint` property linking to the test's name in Python code. This works well for Python since the location remains relevant even as code is shifted around (e.g. as the developer updates neighboring tests).

This approach however doesn't support synthetic `pytest.Item`s as used in data-driven test suites (e.g. like we have [in mypy](https://github.com/python/mypy/blob/master/test-data/unit/README.md)). Those are stored in a config format that's not Python code, and when the tests are executed, the test node's `location` points correctly to the right section of the config file but we can't double-tap the test's name to jump to its location in PyCharm. With this change, the location of the pytest item is propagated as a `locationHint` when it's referring not Python code.

Issue: https://youtrack.jetbrains.com/issue/PY-58751


EliseevEgor:
Hello! Could you please tell me your name and mailbox address? I need it for merging, I would like to add you as the author.

EliseevEgor:
I will merge your changes into the master branch!

PR #2309, opened 2023-01-29
Translate selected files to changes if no changes are selected

ReallyLiri:
If invoking the action from project tool window for example, then no changes are selected, but rather virtual files. however, these may be translated to changes.

### DIFF
--- a/platform/vcs-impl/src/com/intellij/openapi/vcs/actions/commit/AbstractCommitChangesAction.kt
+++ b/platform/vcs-impl/src/com/intellij/openapi/vcs/actions/commit/AbstractCommitChangesAction.kt
@@ -5,6 +5,7 @@ import com.intellij.idea.ActionsBundle
 import com.intellij.openapi.actionSystem.ActionPlaces
 import com.intellij.openapi.actionSystem.ActionUpdateThread
 import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.CommonDataKeys
 import com.intellij.openapi.project.DumbAwareAction
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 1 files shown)


AMPivovarov:
So the fix is to *disable* "Shelve"/"Create Patch"/"???" actions in some contexts, without changing the way they behave?
Could you share a use case?

ReallyLiri:
> So the fix is to _disable_ "Shelve"/"Create Patch"/"???" actions in some contexts (ex: ProjectView without modified files selected), without changing the way they behave? Could you share a use case?

The fix in no way should disable any action that wasn't disabled beforehand, if I missed anything in that angle, the PR should be fixed.

The action might be invoked from a virtual files context, i.e the project tool window, on files that have modifications. But since they are virtual files and not changes, the action context is assuming no changes were selected and the commit dialog is opened with all items checked, this is not the optimal ux.
Example, its using a WIP plugin, but the code behind the scenes the straight forward `AbstractVcs` infra:

changes

AMPivovarov:
The suggested patch only affects action visibility (changes `AnAction.update`) and only disables it occasionally (as `emptyList<String>().all { it == "Uniqie" } == true`.

For "default vcs commit" this problem was solved by the introduction of a separate "Commit File" action: `com.intellij.openapi.vcs.actions.commit.CommonCheckinFilesAction`. 
For "commit using some custom executor" it was not considered.

What is the source of "Diversion" menu in the gif? I do not see it in WIP plugin sources.
Maybe the fix is to use `"CheckinFiles"` action instead of `"CheckinProject"` there?

AMPivovarov:
>so I'll leave it to you guys if you want to pursue it.
Feel free to file it on https://youtrack.jetbrains.com/issues/IDEA to increase the odds.

PR #2311, opened 2023-02-01
KTIJ-19679 'Replace with a for loop' intention isn't suggested for a call with an implicit receiver

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-19679


abelkov (changes requested):
Please support the case with multiple receivers:

```
fun List<Int>.doSmth2() {
    with (Any()) {
        forEach { element ->  }
    }
}
```

abelkov (approved):

PR #2314, opened 2023-02-02
Add capability to fetch coroutine stack trace in one call

nikita-nazarov:
Depends on: https://github.com/JetBrains/kotlin/pull/5088

This commit adds a wrapper over the new compiler API that allows fetching a coroutine stack trace in one call, which will significantly speed up async stack trace fetching when debugging an android app on a phone. Also a couple of minor improvements of the coroutine debugger are added.


zuevmaxim:
The issue was fixed with an alternative approach https://github.com/JetBrains/intellij-community/commit/083a9cbbd47bc1c6ec068077ae3871d69eea5de5

PR #2315, opened 2023-02-03
IDEA-277216 show comparison view for AssertJ AssertionFailedError with Gradle test runner

gilday:
See https://youtrack.jetbrains.com/issue/IDEA-277216/Click-to-see-difference-is-not-present-in-test-output-when-running-tests-with-Gradle-and-assertj-assertions

### DIFF
--- a/plugins/gradle/java/src/execution/GradleRunnerUtil.java
+++ b/plugins/gradle/java/src/execution/GradleRunnerUtil.java
@@ -110,10 +110,7 @@ public final class GradleRunnerUtil {
       comparisonPair = parseComparisonMessage(exceptionMsg, ".*\\s*expected:\\s?<(.*)> but was:\\s?<(.*)>");
     }
     if (comparisonPair == null) {
-      comparisonPair = parseComparisonMessage(exceptionMsg, "\nExpected: \"(.*)\"\n\\s*but: was \"(.*)\"");
-    }
-    if (comparisonPair == null) {
-      comparisonPair = parseComparisonMessage(exceptionMsg, "\nexpected: \"(.*)\"\n\\s*but was: \"(.*)\"");
+      comparisonPair = parseComparisonMessage(exceptionMsg, "\n[Ee]xpected: \"(.*)\"\n\\s*but:? was:? \"(.*)\"");
     }
     return comparisonPair;
   }
--- a/plugins/gradle/java/src/execution/GradleRunnerUtil.java
+++ b/plugins/gradle/java/src/execution/GradleRunnerUtil.java
@@ -112,6 +112,9 @@ public final class GradleRunnerUtil {
     if (comparisonPair == null) {
       comparisonPair = parseComparisonMessage(exceptionMsg, "\nExpected: \"(.*)\"\n\\s*but: was \"(.*)\"");
     }
+    if (comparisonPair == null) {
+      comparisonPair = parseComparisonMessage(exceptionMsg, "\nexpected: \"(.*)\"\n\\s*but was: \"(.*)\"");
+    }
     return comparisonPair;
   }
 
--- a/.idea/google-java-format.xml
+++ b/.idea/google-java-format.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="GoogleJavaFormatSettings">
+    <option name="enabled" value="false" />
+  </component>
+</project>
--- a/plugins/gradle/java/testSources/execution/GradleRunnerUtilTest.kt
+++ b/plugins/gradle/java/testSources/execution/GradleRunnerUtilTest.kt
@@ -24,5 +24,7 @@ class GradleRunnerUtilTest {
     check("expected same:<foo> was not:<Foo >", "foo", "Foo ")
     check("the problem ==> expected same:<foo> was not:<Foo >", "foo", "Foo ")
     check("the problem ==> expected: <foo> but was: <Foo >", "foo", "Foo ")
+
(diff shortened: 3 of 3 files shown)


plugins/gradle/java/src/execution/GradleRunnerUtil.java
```diff
@@ -112,6 +112,9 @@ public static Couple<String> parseComparisonMessage(String exceptionMsg) {
     if (comparisonPair == null) {
       comparisonPair = parseComparisonMessage(exceptionMsg, "\nExpected: \"(.*)\"\n\\s*but: was \"(.*)\"");
     }
+    if (comparisonPair == null) {
+      comparisonPair = parseComparisonMessage(exceptionMsg, "\nexpected: \"(.*)\"\n\\s*but was: \"(.*)\"");
```
vlsi, line 116: Would you update the previous regexp instead so it covers both cases?

gilday: I could use a regex like

```
\n[Ee]xpected: "(.*)"\n\s*but:? was:? "(.*)"
```

but it wouldn't be semantically the same as this code, because there's a disconnect between the `E` vs `e` and the `but: was` vs `but was:`. For example, this regex would erroneously match the string

`
Expected: "foo"
but was: "Foo"
`

whereas the code (as written) would not.

vlsi: Well, that is fair, however, does it make any difference in practice?
Frankly speaking, it is demotivating that IDE has to execute the same regular expressions several times, especially when the two do not differ much.

gilday: I'm not aware of a case, such as the one I provided earlier, that is practical. I combined the regexs.

vlsi:
@gilday , I think the regexp approach looks suspicious, and it should not be required.

I believe IDEA should use `org.opentest4j.AssertionFailedError` fields instead of parsing the regular expressions: https://github.com/JetBrains/intellij-community/blob/1c41fe4eec2efd1dcbee8804c86fa542a08f112d/java/java-runtime/src/com/intellij/rt/execution/junit/ComparisonFailureData.java#L209-L211

So the issue is either `Gradle not providing a properly working AssertionFailedError` or `IDEA-Gradle runner failing to decode AssertionFailedError` properly.

gilday:
> @gilday , I think the regexp approach looks suspicious, and it should not be required.
> 
> I believe IDEA should use `org.opentest4j.AssertionFailedError` fields instead of parsing the regular expressions:
> 
> https://github.com/JetBrains/intellij-community/blob/1c41fe4eec2efd1dcbee8804c86fa542a08f112d/java/java-runtime/src/com/intellij/rt/execution/junit/ComparisonFailureData.java#L209-L211
> 
> So the issue is either `Gradle not providing a properly working AssertionFailedError` or `IDEA-Gradle runner failing to decode AssertionFailedError` properly.

@vlsi I think you're implying that the Gradle test runner catches or otherwise receives `Throwable` objects created by the user's tests, but this is not consistent with my understanding of how the Gradle Tooling API (upon which I understand the Gradle test runner is built) works. My understanding is that the IntelliJ uses the Gradle Tooling API is a facade over the Gradle daemon's RPC API, so it's unlikely that any user-defined `Throwable` types are being serialized across this API from the user's tests to the IntelliJ Gradle test runner.

The Gradle Tooling API is a much different integration than the IntelliJ JUnit test runner example you've referenced. The JUnit test runner API will propagate users' `Throwable` objects to `RunListener` objects as both exist in the same JVM.

vlsi:
> My understanding is that the IntelliJ uses the Gradle Tooling API is a facade over the Gradle daemon's RPC API, so it's unlikely that any user-defined Throwable types are being serialized across this API from the user's tests to the IntelliJ Gradle test runner.

Well, I incline that if user code produces an exception with rich metadata, then
a) Either Gradle should propagate it somehow (see https://github.com/gradle/gradle/issues/12764 )
b) Or it should just fail, so the user has a clear message: "your test has failed, however, Gradle discarded the exception".

Currently, Gradle pretends it passes the exception, however, it seems to discard custom data from exceptions.

I think it does not sound right to work around that by regexp parsing.

johnameyer:
Really looking forward to this fix thanks for picking this up.

Just to chime in, we have written our own matching function and had been making use of the fact that the IntelliJ runner matches based on regex ([as it seems many other folks have](https://stackoverflow.com/questions/10934743/formatting-output-so-that-intellij-idea-shows-diffs-for-two-texts)) - this may also be important for teams using `SoftAssertions` - so limiting to just thrown exceptions would not cover these use cases

gilday:
> I think it does not sound right to work around that by regexp parsing.

Maybe so, but I that's not a new pattern I'm introducing. I'm simply expanding an existing regex to cover another case. At this time, I'm not willing to put in the effort to both improve the Gradle Tooling API and change how IntelliJ's Gradle test runner consumes exceptions from that API.

If you'd prefer to decline this PR in favor of planning a larger refactor, then please note that in [IDEA-277216](https://youtrack.jetbrains.com/issue/IDEA-277216/Click-to-see-difference-is-not-present-in-test-output-when-running-tests-with-Gradle-and-assertj-assertions) for the folks watching.

nskvortsov:
Thank you for the PR. Will run it through the CI and merge it into the master branch if it passes.
The regexp approach is indeed very brittle and error prone. In the future, we definitely want to get more structured information in test results from Gradle TAPI

nskvortsov:
Merged as b9c612ac25618478c6d29813ef4c2446eb47f6c7

PR #2320, opened 2023-02-10
KTIJ-24559 NoSuchElementException on element removal from listOf

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-24559

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ReorderParametersFix.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ReorderParametersFix.kt
@@ -20,7 +20,9 @@ import org.jetbrains.kotlin.idea.refactoring.changeSignature.KotlinMethodDescrip
 import org.jetbrains.kotlin.idea.refactoring.changeSignature.modify
 import org.jetbrains.kotlin.idea.refactoring.changeSignature.runChangeSignature
 import org.jetbrains.kotlin.idea.search.usagesSearch.descriptor
-import org.jetbrains.kotlin.psi.*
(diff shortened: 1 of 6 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ReorderParametersFix.kt
```diff
@@ -74,15 +79,19 @@ class ReorderParametersFix(
...
+                val name = param.name ?: return@mapIndexedNotNull null
+                if (name !in parametersToReorder) {
+                    unchangedParameterCount++
+                    name
+                } else {
+                    parametersToReorder[index - unchangedParameterCount]
+                }
+            }
             if (sortedParameters.size != parameters.size) return null
             return ReorderParametersFix(function, sortedParameters)
```
nikitabobko, line 96: Can be simplified:
```
            val parametersSortedTopologically = DirectedGraph(parametersDependencyGraph).sortTopologically() ?: return null
            val parameterToIndex = parameters.asSequence().mapIndexedNotNull { index, parameter ->
                parameter.name?.let { it to index }
            }.toMap()
            val parametersToSortSorted = parametersSortedTopologically.flatMap { it.sortedBy(parameterToIndex::get) }
            val parametersToSortSortedIterator = parametersToSortSorted.iterator()
            val sortedParameters = parameters.mapNotNull {
                val parameterName = it.name
                if (parameterName in parametersToSortSorted) parametersToSortSortedIterator.next()
                else parameterName
            }
            if (sortedParameters.size != parameters.size) return null
            return ReorderParametersFix(function, sortedParameters)
```

nikitabobko (approved):
Will push to the `master` with the suggested simplification

PR #2323, opened 2023-02-13
KTIJ-24551 `Redundant value argument` inspection may also warn in case of overridden functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-24551


t-kameyama:
2cddb06468e37625d5e701088b7f8331a9d5cf8b

PR #2325, opened 2023-02-16
fix: USER_SCALE_FACTOR_PROPERTY was public before kotlin conversion

bric3:
Public field in the Java version of this file in 222 : 

https://github.com/JetBrains/intellij-community/blob/41f222c6ce05145064e1f1365fe5bf6c92fa0e90/platform/util/ui/src/com/intellij/ui/scale/JBUIScale.java#LL29C46-L29C46

### DIFF
--- a/platform/util/ui/src/com/intellij/ui/scale/JBUIScale.kt
+++ b/platform/util/ui/src/com/intellij/ui/scale/JBUIScale.kt
@@ -27,7 +27,7 @@ object JBUIScale {
   @Internal
   val SCALE_VERBOSE = java.lang.Boolean.getBoolean("ide.ui.scale.verbose")
 
-  private const val USER_SCALE_FACTOR_PROPERTY = "JBUIScale.userScaleFactor"
+  const val USER_SCALE_FACTOR_PROPERTY = "JBUIScale.userScaleFactor"
 
   /**
    * The user scale factor, see [ScaleType.USR_SCALE].


AMPivovarov:
@bric3 Why would you need this property? 
The only valid usage - the `PROPERTY_CHANGE_SUPPORT` listener, was already private in initial commit (https://github.com/JetBrains/intellij-community/blob/d6327550159355bd4e93e027b592900348f1336a/platform/util/src/com/intellij/util/ui/JBUI.java#L44).

bric3:
@AMPivovarov Hi, I don't mean to change the internal `PropertyChangeSupport` field, but `USER_SCALE_FACTOR_PROPERTY`, which is the property name that is being changed. In the commit you've mentioned this field is public too.

https://github.com/JetBrains/intellij-community/blob/d6327550159355bd4e93e027b592900348f1336a/platform/util/src/com/intellij/util/ui/JBUI.java#LL42C3-L42C67

AMPivovarov:
@bric3 Yes, but what's the use case? 

You can't fire this event nor subscribe to it directly without the `PropertyChangeSupport`.
Do you want to double-check that event that comes into `JBUIScale.addUserScaleChangeListener` callback is the "right one"? This seems unnecessary.
Use it to merge multiple `PropertyChangeListener` handlers into one? Yes, but a strange case (merging handlers for events from different sources inc. this very-guru-one).

bric3:
@AMPivovarov Sorry for the delayed response.

So the use case is when using a shared `PropertyChangeListener` used/registered in different places to react to different property changes. In order to distinguish these properties one needs to segregate by property _name_, and I believe it's better to use `SCALE_FACTOR_PROPERTY` rather than a `String`.

PR #2329, opened 2023-02-17
IDEA-310341: Respect .editorconfig on reformat code with optimize imp…

fp7:
…orts

When the ReformatCodeAction with optimize import on one file is executed, the optimize import processor will be run concurrently with `notifyCodeStyleSettingsChanged()`. Which triggers a reload of the code style settings. During that time `JavaCodeStyleSettings.getInstance(file)` for the file will yield the project wide code style settings because `myComputationLock` in CodeStyleCachedValueProvider#tryGetSettings is held by the update task and the read returns null. This is a problem with code style systems that support different formats for same files, eg. EditorConfig.

This commit changes this to run `notifyCodeStyleSettingsChanged` after the processor is done.

Note:
Writing a test for this is hard, because the reformat processor is only run concurrently when non headless. So in tests it always run correctly.


pyltsin:
I am really sorry for the long delay.
It looks like it was solved in https://youtrack.jetbrains.com/issue/IJPL-29963/Star-import-count-settings-in-editorconfig-file-are-sometimes-ignored-when-performing-reformat-code 
in different way. It is not so universal, but looks a bit safer

pyltsin:
I close it for now. If you don't agree, please, drop the comment to discuss this issue further

PR #2330, opened 2023-02-21
Fix touchpad smooth scrolling [IDEA-131174]

endorh:
I know there have been previous attempts to fix this issue before, including the many experimental Registry flags under `idea.true.smooth.scrolling`, however, none of these have managed to properly fix this problem, and any time I need to use IntelliJ on my laptop without a mouse this poor UX drives me mad.

[IntelliJ - Smooth Scrolling.webm](https://user-images.githubusercontent.com/86890810/220235467-dd032f9a-67da-46ac-8e92-2e31cd4ebcd1.webm)

I've tried to make the smallest possible fix, which perfectly works on my Windows 10 laptop, and shouldn't interfere with regular mouse scrolling in any way. I'm however unable to verify if this fix helps to mitigate this issue on MacOS devices (albeit it reasonably should).

Some implementation considerations:
- In Windows, the `MouseWheelEvent#getPreciseWheelRotation` properly returns a fractional wheel delta, except when zooming with a pinch gesture (`Ctrl + Wheel`), in which case, the `MouseWheelEvent` is accumulated until it reaches a full delta unit (unfortunately, the cause of this is Windows itself, touchpad pinch gestures are a separate event, which AWT doesn't map to `MouseWheelEvent`).
- The added `verticalCumulativeDelta` and `horizontalCumulativeDelta` fields are necessary. Otherwise, as can be tested by commenting line `MouseWheelSmoothScroll.java:60`, scrolling slowly in panels where the scrollable unit increment is small enough such that the resulting delta is smaller than a whole pixel (such as scrolling horizontally in the Project view) would result in an unresponsive UX, that would only respond after reaching a certain minimal speed, and decimals would be lost on each event, regardless of speed. This breaks the linearity of scrolling back and forth, and can drive a user mad too.
- To support proper touchpad smooth scrolling it's also necessary to suppress the mouse smooth scrolling animator when a touchpad is being used. This is detected by the use of wheel delta values smaller than a whole unit (since any other device that produced fractional wheel events would also probably prefer having no artificial smoothing). Adding inertia to the touchpad scrolling results in a laggy feeling, extremely frustrating from the user perspective.
- There's no need to implement fling gestures, since Windows (and I assume MacOS) do this already (as janky as it may be). With this fix, flinging the touchpad behaves as it'd be expected.
- `MouseWheelSmoothScroll$ScrollAnimationSettings#SETTINGS` has been renamed to `DEFAULT`, since it's a more sensible name after adding `MouseWheelSmoothScroll$ScrollAnimationSettings#NO_ANIMATION`. As this enum is package private and the only usage is within the code that has been modified, no concerns justified keeping the original name (but I'm of course willing to revert it if preferred).
- No changes have been made to the public platform API. This fix should have no effect on backwards compatibility.

Additionally, I've also extended this fix to the image viewer in a separate commit.
Unfortunately, as already mentioned, at least in Windows, touchpad pinch gestures don't generate fractional delta wheel events as scroll gestures do, so the only way to smoothly zoom in the image viewer with this fix is by holding the `Ctrl` key while scrolling with the touchpad. However, should that ever change in the future, this fix would make pinch gestures zoom smoothly.

I'm willing to make any necessary changes to get this accepted, but please consider accepting this fix. Jagged touchpad scrolling makes for very poor UX. Feedback is welcome.


endorh:
After some investigation I've found that the source of the accumulation of wheel events for pinch gestures is actually Windows, since even native applications receive them accumulated, so it's not an AWT issue as I thought initially, although a solution could be implemented there. I've updated the description to reflect this.

endorh:
Has this been forgotten, @bulenkov ? It's been more than 2 years since I opened this PR without comment.
Did I forget a step somewhere to get this reviewed?

I understand that UI changes like these are delicate and should be tested with care.
I have tested my changes today in Mac OS (in addition to Windows), and it works like a charm too.

During today's testing, I noticed that the feedback was being poorly smoothed between Windows' touchpad scroll inertia events, so I took the chance to resolve that too in today's commit.

Is there anything I may do to help get this merged? Honestly, as a frequent laptop developer, this issue drives me insane.

endorh:
Oh, I forgot to add. All my tests today were in the new UI.

PR #2331, opened 2023-02-21
feat: add create new tag action to git file history's context menu

DevDengChao:
image

See https://youtrack.jetbrains.com/issue/IDEA-313766


juliabeliaeva:
Thank you for the PR. Merged.

PR #2335, opened 2023-02-23
Parameter to control if KotlinGradleModel fetches classpath.

akerimsenol:
This should have no significant effect without changes to KGP itself as it's just changing what's called on that side.

I've added a test case KotlinGradleModelTest and also ran the KOTLIN_PLUGIN_TESTS group, which are both fine.


akerimsenol:
I forgot to add that this is in the context of: https://youtrack.jetbrains.com/issue/KTIJ-24724

darthorimar:
@Tapchicoma could you please review?

PR #2336, opened 2023-02-25
IDEA-314175: Handle relative classpath jar paths

gharrma:
We convert paths to absolute paths because JarLoader and clients thereof are unprepared for relative paths.

This change is necessary for certain environments (such as Bazel tests) in which relative paths are used for the sake of hermeticity. Previously, using PathClassLoader in Bazel would result in invalid resource paths and other runtime errors.

This change should have no effect on existing environments where absolute paths are already used.

---

Issue link: https://youtrack.jetbrains.com/issue/IDEA-314175
Test: launch the IDE in both source builds and release builds
Test: use PathClassLoader in Bazel tests

### DIFF
--- a/platform/util-class-loader/src/com/intellij/util/lang/ClassPath.java
+++ b/platform/util-class-loader/src/com/intellij/util/lang/ClassPath.java
@@ -339,7 +339,7 @@ public final class ClassPath {
         return null;
       }
 
-      Path path = files[searchOffset++];
+      Path path = files[searchOffset++].toAbsolutePath();
       try {
         Loader loader = createLoader(path);
         if (loader != null) {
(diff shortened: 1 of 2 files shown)


gharrma:
Regression test added

develar:
Thanks. The PR is in the process of the merge.

FTR: 
1. I agree that better to execute `toAbsolutePath()` where we create a loader and not in the constructor or setter. `reset` may be called, and the absolute path may be changed for the process dynamically (depending on a filesystem implementation).
2. For `UnixPath`, it costs us only `path.length > 0 && path[0] == '/'` check, looks quite safe.

The only change I will to a patch — is a comment explaining why we do call `toAbsolutePath`.

gharrma:
Thanks Vladimir, much appreciated.
I see it merged now as [c4b1c60067](https://github.com/JetBrains/intellij-community/commit/c4b1c6006714e6c9c9a1182de2ceb45fcef808fe).

PR #2337, opened 2023-02-28
Fix @Range annotations for OffsetDateTime.of

kranich:
The second and third parameters of

    public static OffsetDateTime of(int year, int month, int dayOfMonth, int hour, int minute, int second, int nanoOfSecond, ZoneOffset offset)

are month and dayOfMonth respectively. Their ranges should be 1-12 and 1-31, not vice versa.


pyltsin:
Thank you! The patch is merged into the internal repository (a6e9bc2465d177220c8f5f4b5fb162dd2ad577ea). Should appear on GitHub in a few hours.

PR #2342, opened 2023-03-06
Make KotlinRequestHint continue stepping after the Kotlin fake line number in case of smart stepping

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-24091/Debugger-smart-stepping-steps-into-a-wrong-inline-lambda-on-the-same-line

Consider the example: `1.let { it }.let { it + 1 }.let { it + 2 }`. Previously when trying to smart step into `{ it + 1 }` or `{ it + 2 }` the debugger would stop in `{ it }`, because during smart stepping the debugger would cross a fake line number that marks the end of an inline lambda, and KotlinRequestHint would stop the smart stepping process.

I must admit that merging this pr will uncover another smart stepping issue.
Consider the example:
```
inline fun foo() {
  1.let { it }.let { it + 1 }.let { it + 2 }
}

fun main() {
  foo()
  foo() // Breakpoint here
}
```
Run the debugger and stop on a breakpoint, then invoke the step into action and stop in `foo`, and then try to smart step into `{ it + 1 }` or `{ it + 2 }`. It will not work. 

When we try to smart step into a lambda, we set a StepIntoBreakpoint on a location inside of this lambda. After that we start the smart stepping process, and eventually the debugger will hit that breakpoint, and the desirable lambda will be reached. This works well for Java lambdas and ordinary Kotlin lambdas. Things get worse when we are trying to smart step into an inline lambda. In the above example each of the inlined sites of the foo function contains locations inside the lambda of interest. However, only one StepIntoBreakpoint is set, this can be seen here com/intellij/debugger/ui/breakpoints/StepIntoBreakpoint.java:74. It means that smart stepping will only work for the first inlined site of the foo function.

Before this pr we would always stop in `{ it }`, because of crossing the Kotlin fake line number. 

I believe that the described issue has to be fixed, but in a separate pr.

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
+++ b/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
@@ -180,10 +180,10 @@ class KotlinStepIntoRequestHint(
                 lastWasKotlinFakeLineNumber = true
                 return StepRequest.STEP_INTO
             }
-            // If the last line was a fake line number, the next non-fake line number
-            // is always of interest (otherwise, we wouldn't have had to insert the
-            // fake line number in the first place).
-            if (lastWasKotlinFakeLineNumber) {
+            // If the last line was a fake line number, and we are not smart-stepping,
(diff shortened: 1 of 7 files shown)


yanex:
Merged to `master` manually (https://github.com/JetBrains/intellij-community/commit/15e533f7198de56522da880389743ea28277923f). Thank you!

PR #2343, opened 2023-03-07
Make KotlinLambdaMethodFilter rely on psi while checking location inside an inline lambda

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-23965/Debugger-smart-stepping-into-inline-lambdas-doesnt-work-when-lambdas-are-declared-in-functions-with-the-same-name

This commit improves the design of KotlinLambdaMethodFilter, so instead of using ClassNameCalculator directly to detect if we should stop in a lambda, the filter now relies on KotlinPositionManager and the provided source positions.

Unfortunately this pr will not solve issues like this:
```
fun foo (x: Int) {
  x.let { it }.let { it + 1 }
}

fun foo (x: Int, y: Int) {
  x.let { it }.let { it + 1 } // Smart stepping here will not work
}

fun main() {
  foo(1)
  foo(1, 2)
}
```
The reason is that ClassNameCalculator doesn't calculate lambda names correctly in this case. You can read more about ClassNameCalculator issues [here](https://github.com/JetBrains/intellij-community/pull/2279).


yanex:
Merged to `master` manually (https://github.com/JetBrains/intellij-community/commit/1169f23dd9650a780e06eb7fd71e8de3986660ea). Thank you!

PR #2344, opened 2023-03-07
Patch path so it reflects changes in patched arguments

superjkl:
Given arguments in the form of: `os.execv("../file.py", ["../file.py", arg1, ...]) `
They will be patched to: `os.execv("../file.py", ["python", "../../pydevd.py", "--host", "--port", "--file", "../file.py", arg1, ...])`
This will end up with os.execv thinking the actual executable is  "../file.py" thus running it instead of pydevd resulting in the error `file.py: error: unrecognized arguments: host, port`

I think this bug was introduced by commit 7ec061b4182a183ef3193a24ecb9c6b0490799ad so I was careful to make sure my changes wouldn't reintroduce the bug from PY-37311.

### DIFF
--- a/python/helpers/pydev/_pydev_bundle/pydev_monkey.py
+++ b/python/helpers/pydev/_pydev_bundle/pydev_monkey.py
@@ -439,6 +439,8 @@ def create_warn_multiproc(original_name):
 
     return new_warn_multiproc
 
+def patch_path(path, args):
+    return args[0] if is_python(path) and args[0] == sys.executable else path
 
 def create_execl(original_name):
     def new_execl(path, *args):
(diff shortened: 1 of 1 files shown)


superjkl:
So sorry for leaving out that crucial info. 
OS: Centos 7.9
Python: 2.7

EliseevEgor:
Thanks for the info!

How do you make a `.py` file executable? 
Are you adding `#!/usr/bin/env python`?

superjkl:
Yes both the original script and subprocess script use `#!/usr/bin/env python`

Here's a small example I can replicate the bug with:

**tester1.py**
```
#!/usr/bin/env python
import os
args = [os.path.abspath('./tester2.py')]
os.execv(args[0], args)
```

**tester2.py**
```
#!/usr/bin/env python
import argparse

a = argparse.ArgumentParser()
a.add_argument('--option')
a.parse_args()
```

EliseevEgor:
Thank you!

EliseevEgor:
I merged your changes to the main branch, thank you for contributing!

PR #2347, opened 2023-03-10
IDEA-315065: Exclude JUnit3 from IDE classpath

gharrma:
It clashes with junit4.jar, but cannot be removed from IntelliJ entirely because an old IntelliJ quickfix once suggested to add it to the user's classpath.

---

Issue link: https://youtrack.jetbrains.com/issue/IDEA-315065


develar:
Thanks! The proposed change in the process of the merge.

gharrma:
Thanks Vladimir!

PR #2348, opened 2023-03-10
Fix KTS unresolved symbols by using the same jdk.table entry as project JDK

vmadalin:
### Description
affects IntelliJ

The reason for those unresolved symbols in the KTS, is in fact an issue with picking the JDK table entry from the jdk.table.xml in the [ScriptSdksBuilder.getScriptSdkByJavaHome](https://github.com/JetBrains/intellij-kotlin/blob/e2c105911e1fca5b7fe6b9dc10cf9f75463a3c13/core/src/org/jetbrains/kotlin/idea/core/script/ucache/ScriptSdksBuilder.kt#L71) that will end up been used to resolve the symbols for the KTS files. In particular, given a JDK path, the first entry is returned that contains the same `homePath`.

platform return first jdk table entry given  a path

Unfortunately, this logic doesn't take into consideration that may exist different entries for the same `homePath` with corrupted roots and at the same time this entry may not be the same as the one used by the `Project JDK` (entry name specified under `.idea/misc.xml`). Meaning that depending on the user `jdk.table.xml` ordering they may end up facing this issue, making it difficult to predict how many of them will be affected.

### Example:
Given a `Project JDK` using `project-jdk-name="jbr-17"`

- [OK scenario](https://user-images.githubusercontent.com/18151158/224314059-5f36d38f-169a-4137-84d4-937beede8822.png) ✅ 
```xml
<jdk version="2"> 
    <name value="jbr-17" />
    <homePath value="$USER_HOME$/jdk17/mac-arm64/Contents/Home" />
     <roots>
        <classPath>
            <root url="jrt:///USER_HOME$/jdk17/mac-arm64/Contents/Home!/java.base" type="simple" /> #valid paths
...
<jdk version="2"> 
    <name value="jbr-17-corrupted" />
    <homePath value="$USER_HOME$/jdk17/mac-arm64/Contents/Home" />
     <roots>
        <classPath>
            <root url="jrt:///INVALID/jdk17/mac-arm64/Contents/Home!/java.base" type="simple" /> #invalid paths
... 
```

- [KO scenario](https://user-images.githubusercontent.com/18151158/224314050-99c9f6fd-c5a1-4571-8aca-6b70739ebb3a.png) ❌ 
```xml
<jdk version="2"> 
    <name value="jbr-17-corrupted" />
    <homePath value="$USER_HOME$/jdk17/mac-arm64/Contents/Home" />
     <roots>
        <classPath>
            <root url="jrt:///INVALID/jdk17/mac-arm64/Contents/Home!/java.base" type="simple" /> #invalid paths
...
<jdk version="2"> 
    <name value="jbr-17" />
    <homePath value="$USER_HOME$/jdk17/mac-arm64/Contents/Home" />
     <roots>
        <classPath>
            <root url="jrt:///USER_HOME$/jdk17/mac-arm64/Contents/Home!/java.base" type="simple" /> #valid paths
...
```

### Solution
In order to keeping consistency cross project when possible we suggesting to reuse same `jdk.table` entry specified for the `Project JDK`, this will keep things aligned regarding which JDK will be used to resolve symbols.

### DIFF
--- a/plugins/gradle/testSources/org/jetbrains/plugins/gradle/extensions/SdkExtensions.kt
+++ b/plugins/gradle/testSources/org/jetbrains/plugins/gradle/extensions/SdkExtensions.kt
@@ -1,0 +1,47 @@
+// Copyright 2000-2023 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.plugins.gradle.extensions
+
+import com.intellij.openapi.projectRoots.Sdk
+import com.intellij.openapi.roots.OrderRootType
+import com.intellij.openapi.vfs.VfsUtil
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.testFramework.VfsTestUtil
+import java.io.File
+import java.io.FileNotFoundException
+
+val Sdk.rootsFiles
+  get() = OrderRootType.getAllTypes().flatMap { rootType ->
+    rootProvider.getFiles(rootType).map { it to rootType }
+  }
+
+val Sdk.rootsUrls
+  get() = OrderRootType.getAllTypes().flatMap { rootType ->
+    rootProvider.getUrls(rootType).map { it to rootType }
+  }
+
+fun Sdk.cloneWithCorruptedRoots(tempDir: File): Sdk {
+  val sdkClone = clone() as Sdk
+  val corruptedRoots = generateCorruptedJdkRoots(sdkClone, tempDir)
+  sdkClone.sdkModificator.run {
+    removeAllRoots()
+    corruptedRoots.forEach { (type, file) ->
+      addRoot(type, file)
+    }
+    commitChanges()
+  }
+  return sdkClone
+}
+
+private fun generateCorruptedJdkRoots(
+  jdk: Sdk,
+  tempDir: File,
+): List<Pair<VirtualFile, OrderRootType>> {
+  val vfsTempDir = VfsUtil.findFile(tempDir.toPath(), true) ?: throw FileNotFoundException("Unable to find ${tempDir.path}")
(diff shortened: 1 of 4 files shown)


klunnii:
The proposed solution is based on the assumption that a project JDK is set and the default one is a project one.
Please, take a look at `ScriptSdksBuilder#getDefaultSdk`. In case a project JDK is missing the first one that `ScriptSdksBuilder#canBeUsedForScript` is chosen. That is, KO scenario is still possible at least in theory.

What if instead of taking the default we change the behavior of `ScriptSdksBuilder#getScriptSdkByJavaHome` letting it jump over the broken entries? Something like this:
``` kotlin
        return runReadAction { ProjectJdkTable.getInstance() }.allJdks
            .firstOrNull { it.homeDirectory == javaHomeVF && it.canBeUsedForScript()}
            //.find { it.homeDirectory == javaHomeVF }
            //?.takeIf { it.canBeUsedForScript() }
```
`canBeUsedForScript()` probably needs to be adjusted as well.

Other than that the commit lacks tests. Can I ask you to look at `GradleBuildFileHighlightingTest` and `ScriptConfigurationHighlightingTestGenerated`? I think the case should be somehow reflected there.

vmadalin:
@klunnii Thanks for pointing this out, absolutely agree that the `project JDK` may not be defined for example when creating a new project and syncing the first time, but not only.

I also added a new test `MultiplesJdkTableEntriesWithSamePathButFirstHasCorruptedRoots` that validates the changes, using the existing `GradleBuildFileHighlightingTest` implementation but configuring the `jdk.table` properly to be able to reproduce the described scenario. I hope this makes sense to you. Thank you

klunnii:
Looks good! Merged as https://github.com/JetBrains/intellij-community/commit/46666ba96af867ab5606f46a78c4847abe4541aa. Thank you!

PR #2349, opened 2023-03-10
Fix debugger freeze while searching for inline function call sites

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-24736/Debugger-IDE-freeze-due-to-long-read-action-in-ClassNameProvider.findInlinedCalls

Previously when trying to place a breakpoint in an inline function, the debugger would have to find all of its call sites to perform this action. If you, for example, put a breakpoint in one of standard library functions, the searching would take a huge amount of time. Things became worse, when you wanted to unset a breakpoint and the searching process was still not finished. Unsetting a breakpoint required a write action, which waited for the previously launched read action to find all the inline function call sites. As a result freeze could happen and the debugger UI would be unresponsive until all of the call sites were found. This pr replaces EmptyProgressIndicator with ProgressIndicatorBase, which allows the action to be cancelled when needed.


vladimirdolzhenko:
Thanks, merged as https://github.com/JetBrains/intellij-community/commit/7284b4bdc58c84f28b56ab24245e6642cf493015

PR #2350, opened 2023-03-10
Fix stepping out of inline lambdas

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-24239/Debugger-stepping-out-of-inline-lambdas-on-one-line-doesnt-work

Previously when trying to step out of an inline lambda, the debugger would always perform a line-wise stepping out. This behaviour resulted in errors when trying to step out of one-line inline lambdas - you would always jump to the next line. This commit sets the minimal stepping size when trying to step out of an inline lambda.

I have to admit, that after merging https://github.com/JetBrains/intellij-community/pull/2342, this pull request will also improve the smart stepping experience. Right now, when trying to smart step into inline lambdas like: `1.let { it }.let { it + 1 }.let { it + 2 }`, the debugger will always stop in `{ it }` because of crossing the Kotlin fake line number. After merging the mentioned pr, we will still not be able to smart step into `{ it + 1 }` or `{ it + 2 }` because of https://youtrack.jetbrains.com/issue/KTIJ-24239/Debugger-stepping-out-of-inline-lambdas-on-one-line-doesnt-work. The debugger will always jump to the next line. Current pull request will allow performing smart stepping correctly.

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
+++ b/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/stepping/KotlinRequestHint.kt
@@ -56,8 +56,9 @@ class KotlinStepOverRequestHint(
     stepThread: ThreadReferenceProxyImpl,
     suspendContext: SuspendContextImpl,
     private val filter: KotlinMethodFilter,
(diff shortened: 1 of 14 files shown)


zuevmaxim:
Merges as bd67493ae9647d753794504b798a815e94b2f699 and 173b6f3819d8da23d803ae3c153af540b035fea9

PR #2351, opened 2023-03-13
Remove unnecessary if condition check

scruel:


AMPivovarov:
Not unnecessary, see missing 'break' in the clause above.

PR #2353, opened 2023-03-13
PY-57950 - Quick Documentation not working in Python Console if the caret is at the end of the expression

rendner:
PR for [PY-57950](https://youtrack.jetbrains.com/issue/PY-57950/Quick-Documentation-not-working-in-Python-Console-if-the-caret-is-at-the-end-of-the-expression)


EliseevEgor:
Hello, @rendner! Thank you for your PR. Your changes will be added to PyCharm with the next releases.

PR #2357, opened 2023-03-15
Add `org.openrewrite.Incubating` to known unstable API annotations

JLLeitschuh:
Link to project: https://github.com/openrewrite/rewrite

Link to annotation: https://github.com/openrewrite/rewrite/blob/main/rewrite-core/src/main/java/org/openrewrite/Incubating.java


amaembo:
It looks like, most of the [usages](https://mvnrepository.com/artifact/org.openrewrite/rewrite-java/usages) of openrewrite artifact are within the openrewrite project itself. Note that you can configure the inspection and add a custom annotation. We may consider hardcoding annotations from very popular projects but we are not inclined to support every single project in the world, as this increases the maintenance cost. Thank you for understanding.

PR #2359, opened 2023-03-20
Register WebP support as an ImageIO plugin

jgaillard85:
Currently, WebP support is manually registered on startup using the application service mechanism. For this it tries to get the IIORegistry default instance. If it had not yet been created, but is in the process of being created in another thread, that will lead to a race condition where only one of the instances will be retained and the WebP support might be lost.
Instead of manually registering the WebP support, this makes use of the plugin facility of ImageIO, where plugins that are present in the classpath will be automatically registered.
Defining an ImageIO plugin requires having files in META-INF/services which specify which services are implemented and where to find their implementations.

### DIFF
--- a/plugins/webp/resources/META-INF/plugin.xml
+++ b/plugins/webp/resources/META-INF/plugin.xml
@@ -8,6 +8,5 @@
     ]]></description>
   <extensions defaultExtensionNs="com.intellij">
     <fileType name="Image" extensions="webp"/>
-    <applicationService serviceImplementation="com.android.tools.adtui.webp.WebpMetadata$WebpMetadataRegistrar" preload="true"/>
   </extensions>
 </idea-plugin>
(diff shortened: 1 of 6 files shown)


jgaillard85:
You're right, as is this does not always work depending on who triggers the loading of ImageIO first, as it is their class loader that is going to be used for scanning for other plugins.
It is a pattern I had seen in the Images plugin (id: com.intellij.platform.images), so I think there is a problem with this one as well.

Is there a way to delay the WebP registration so that it doesn't conflict with the other format registrations that happen when during the loading of the splash screen? Or would it be possible to do the basic registration earlier (before plugins are loaded) so that the registration of the WebP plugin would correctly happen?

andrei-kuznetsov:
@jgaillard85, `com.intellij.platform.images` is a different story, because it is loaded from `platform` classloader. "Delaying a registration" sounds too vague. If we consider extreme case - delay the registration until IDE shutdown (the latest that we can delay it) - this does not make any sense either, because extensions will be registered too late. We need to find an interval when it is just perfect: not too late to be useful, not too early to avoid races.

If you have ideas - feel free to update this PR. Otherwise I suggest to submit a ticket to https://youtrack.jetbrains.com/issues/IDEA describing original problem, or containing a unit test reproducing the problem. Maybe it will be easier to move webp to the platform, or create dedicated EP in the platform to load extensions timely.

jgaillard85:
I don't know the inner workings of the platform well enough to know what is best to solve this.
I filed [IDEA-316037](https://youtrack.jetbrains.com/issue/IDEA-316037/WebP-support-registration-can-fail-due-to-ImageIO-not-being-thread-safe), I'm happy to continue the discussion there

PR #2360, opened 2023-03-20
Handle initial whitespace in Toml files

csrhodes:
If there is initial whitespace in a Toml file, it is logically next after a newline (in terms of parser states) despite that whitespace not actually containing a newline character.


csrhodes:
Thanks very much!

PR #2361, opened 2023-03-20
PY-49293 - improve install package suggestion

rendner:
https://youtrack.jetbrains.com/issue/PY-49293

This PR improves only the `Install package` suggestion and doesn't provide a solution for the whole issue.

I had to turn off the `PyPIPackageUtil.INSTANCE::isInPyPI` check when running the unit tests because I couldn't find another way to write tests. I'm unsure if this OK.


east825:
Hi! Sorry for completely forgetting about this one 😓. 
The problem was fixed a couple of years ago together with other changes related to the package installation quick fix.

PR #2362, opened 2023-03-23
 PY-37855 Use getWorkingDirectorySafe for getting a correct directory

heckad:
https://youtrack.jetbrains.com/issue/PY-37855/Working-directory-is-automatically-set-to-PyCharm-bin-folder-when-empty-in-Run-Debug-configuration


vladimir-koshelev:
The fix has been merged

PR #2363, opened 2023-03-24
Suppress K2 warning in Android Studio

gharrma:
Users cannot disable the Android plugin in Android Studio, so `AndroidPluginIncompatibilityCheckerStartupActivity` should not suggest it (especially now that Android Studio has started migrating to K2).


darthorimar:
@gharrma thank you for your PR :)

PR #2366, opened 2023-03-28
Add support to the new #LOCAL_PROPERTIES_JDK gradleJvm macro

vmadalin:
### Description
We have planned to land the support for this new macro `#LOCAL_PROPERTIES_JDK` at a certain point in `Android Studio H`, becoming a new standard for the users to specify their JDK path on android projects. The main goal is to improve the existing JDK flow that requires a valid `jdk.table.xml` entry to trigger sync but also mitigate some of the existing [issues](https://issuetracker.google.com/issues/214428183#comment6). Meaning that all new created projects will configure their JDK using this new macro and the existing projects will be recommended to be migrated. In any case, studio will still supporting the existing `gradleJvm` configurations like macros and `jdk.table` entries.

Right now, IntelliJ already supports this macro to be able to trigger project sync, introduced by this [changes](https://github.com/JetBrains/intellij-community/pull/2193), however, the `Gradle Settings` doesn't recognise it. For this reason, these changes are adding support to #LOCAL_PROPERTIES_JDK macro on the `GradleJvmComboBox` resolving the JDK using the `jdk.dir` property defined on the project `local.properties` file. 

In any case, the current [GradleJvm resolution](https://github.com/JetBrains/intellij-community/blob/master/plugins/gradle/src/org/jetbrains/plugins/gradle/util/GradleJvmResolutionUtil.kt#L36) will not propose this macro as a possible way to resolve the JDK, keeping this only for a project that is already configured with `#LOCAL_PROPERTIES_JDK` macro. The idea is just to be able to give the users the best experience in case they open their android project in IntelliJ. 

- Before
     Screenshot 2023-03-27 at 15 16 36

- After
      Screenshot 2023-03-27 at 17 44 26


vmadalin:
Merged 1508556a4c6d61cf39dbd74c239d3fce150dc1b0

PR #2369, opened 2023-03-30
Added @Nullable annotation for PsiFile methods

mlytvyn:
To ensure that developers will not forget to add `null` check for PsiFile methods corresponding `@Nullable` annotation has been added.


mlytvyn:
@karollewandowski , updated with latest `master`

mlytvyn:
@chashnikov , any chance this PR will be merged and available with the next release?

chashnikov:
I'm sorry for the late reply. These methods may return `null` indeed, but it happens in rare cases, and in many situations you may be sure that they won't return `null`. Adding these annotations will help to find some bugs indeed, but in many other cases it'll require adding unnecessary `null` checks, and I'm not sure that the benefits are worth it.

Anyway, in its current form the patch cannot be applied because it'll cause compilation failures in some Kotlin files. If we really want to mark returns types of these methods as `@Nullable`, we need to analyse their usages to find out how many places miss `null` checks, and fix at least Kotlin code. And this cannot be done as a pull request, because we have a lot of code not included into intellij-community, and it also needs to be updated.

PR #2370, opened 2023-03-30
[PY-57582] Add support for `dataclasses.KW_ONLY` type

heckad:
https://youtrack.jetbrains.com/issue/PY-57582/dataclasses-field-with-default-value-preceding-KWONLY-shows-false-positive-warning.


mmews-n4:
@heckad Thanks for your fix! We are sorry it took us that long to incorporate it. It will be available with version 252.3.

PR #2371, opened 2023-03-31
KTIJ-23957 ReplaceWith doesn't add explicit generic type argument inferred from deprecated declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-23957


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/replaceWith/DeprecatedSymbolUsageFixBase.kt
```diff
@@ -159,7 +159,6 @@ abstract class DeprecatedSymbolUsageFixBase(
                 patternTypeArgumentCount == typeArgumentList?.arguments?.size
             ) {
                 if (typeArgumentList != null) expressionFromPattern.replaceOrCreateTypeArgumentList(typeArgumentList.copy() as KtTypeArgumentList)
-                else patternTypeArgumentList?.delete()
```
t-kameyama, line 162: Unnecessary type arguments are deleted at:
https://github.com/JetBrains/intellij-community/blob/3840f95449cbc4e5758d58ebc0d7c5bd23f93170/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/codeInliner/CodeInliner.kt#L553

abelkov (approved):

PR #2375, opened 2023-04-03
IDEA-316269 Fix project leak via GotoSymbolModel2 EP listener

npaige:


aydarin:
Thank you @npaige !
The fix has landed to intellij main branch with you as an author.

PR #2377, opened 2023-04-04
Make TextEditorWithPreview#getTabs() non-final

amaurymedeiros:
This method is overridden in the Android Plugin to control the tabs visibility.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/fileEditor/TextEditorWithPreview.java
+++ b/platform/platform-impl/src/com/intellij/openapi/fileEditor/TextEditorWithPreview.java
@@ -441,7 +441,7 @@ public class TextEditorWithPreview extends UserDataHolderBase implements TextEdi
   }
 
   @Override
-  public final @NotNull ActionGroup getTabActions() {
+  public @NotNull ActionGroup getTabActions() {
     AnAction[] actions = createTabActions();
     return new ConditionalActionGroup(actions, () -> isShowActionsInTabs());
   }


dahlstrom-g:
@KonstantinHudyakov
This follows commit 8d16b9d0.

Incidentally the method name is malformed and misstated in the commit title; it's `TextEditorWithPreview.getTabActions`.

amaurymedeiros:
@KonstantinHudyakov Please let me know if you need me to update the commit header for correctness or if it's OK. Thanks!

KonstantinHudyakov:
@amaurymedeiros hi! Commit header is ok.
What versions are you targeting by this change? Unfortunately, 2023.1 is released, so I can't include this change to it.
We can include it to the future 2023.2 release.
But anyway It would be better to override `createTabActions` method if it is possible.

amaurymedeiros:
@KonstantinHudyakov 2023.1 would be ideal because that's the version that breaks us. However, I was hoping that, depending on the scope of this change, we could cherry-pick it into tools/idea even if on your end it's only included in 2023.2.

Overriding `createTabActions()` doesn't actually work for us because we control the tabs visibility depending on some constraints (e.g. showing it if there are `android.compose` or Android custom view imports in code, and hiding it otherwise).

We achieved that by wrapping `getTabActions()` as follows:

``` kotlin
  override fun getTabActions(): ActionGroup? {
    val parentActions = super.getTabActions() ?: return null
    return tabsAction ?: TabsActionWrapper(parentActions).also { tabsAction = it }
  }
```

`TabsActionWrapper` has the `isHideGroupIfEmpty` property set to true, and whenever there is a code change affecting the tabs visibility, we add/remove the actions of `parentActions` to/from it.

Since `createTabActions()` returns a list that's used to create an `ActionGroup` that's not controllable by `TextEditorWithPreview` subclasses,  this solution doesn't work with the new design.

I thought about playing around with overriding `createViewActionGroup()`, but it's used by both toolbar and tabs actions. Do you have any suggestion for us to keep the current behavior using the new design other than exposing `getTabsAction()`? 

Thanks!

KonstantinHudyakov:
@amaurymedeiros we can cherry-pick this fix to 2023.1, but it will be included only in next minor updates (2023.1.1, 2023.1.2 and so on). So, for current version it still will be broken :(

In 2023.1 `getTabActions` method returns `ConditionalActionGroup`. This group is empty if `isShowActionsInTabs` returns false. So, you can override `isShowActionsInTabs` and check whether you need to show the actions.

lppedd:
I also confirm that `isShowActionsInTabs` is the right method to override. Just tested it out for your use case and it works as @KonstantinHudyakov described.

amaurymedeiros:
Thanks you both for the input! Unfortunately, overriding `isShowActionsInTabs` is not enough for us because of our UX requirements regarding this action's visibility.

You only check for this value in `ConditionalActionGroup#getChildren()`, which is called once in `createRightToolbar()` and once in `getTabActions()`. Correct me if I'm wrong, but these methods are only called when you open the editor, and that's when we evaluate if you should display the actions or not. Furthermore,  the `ConditionalActionGroup` created in `createRightToolbar` used the inverse condition (i.e. `!isShowActionsInTabs()`), so only changing the `isShowActionsInTabs()`  return value to hide the tabs means the toolbar will show instead.

In our case, we want to dynamically show and hide the toolbar (for old UI) and the tabs (for new UI). If the user opens an empty Kotlin file in a Compose project, for example, we open the editor without showing the controls. As soon as they add a `@Composable` function to this file, we then display the controls. Similarly, if the user deletes all the Compose-related code from the file, we hide the toolbar again. Note that `getTabActions()`, `isShowActionsInTabs()` and related methods are not called again at this point, and that's why we need to wrap `getTabActions()` in an `ActionGroup` controlled by the subclass at the time it's created.

@dahlstrom-g if this change of removing the `final` modifier is only merged in 2023.2, would it be possible to cherry-pick it when merging 2023.1 into tools/idea? I don't think we would break at runtime because our scenario is different than the one described by @lppedd in the comment above. Is that right, or am I missing something? If that's possible, I think this is the best solution for both parts.

Thank you all.

KonstantinHudyakov:
@amaurymedeiros thank you for so detailed explanation!
`isShowActionsInTabs` called on each action group update, that happen every 500ms when IDE is focused. I think that it is enough often for your case.
Regarding `!isShowActionsInTabs()` in `createRightToolbar` - that's the problem. But you are free to override `createRightToolbar` and specify your custom condition whether to show actions in the floating toolbar.

But anyway, if merging this change to 2023.2 will definitely solve your problems with compatibility, I don't mind doing it.

KonstantinHudyakov:
Also, the best (from the view of design) way to control whether some action need to be shown or not, is specifying visibility state using `Presentation#setEnabledAndVisible` in `AnAction#update`. If you fully control the implementation of your custom actions, you can specify the condition in this method. The services (like editor, and file editor) required for calculation of your condition can be obtained using `AnActionEvent#getData`.

amaurymedeiros:
> `isShowActionsInTabs` called on each action group update, that happen every 500ms when IDE is focused

Thanks for letting me know! If that's the case, I think this might be enough for us then. I certainly can work around the `createRightToolbar` to handle the old UI as well.

Also, thanks for the guidance on controlling visibility state. I think that opens yet another possibility for us. Know that the tabs action are three individual actions (as opposed to an action to change mode + popup action, as it was when they were introduced), we might be able to make it work with something like:

``` kotlin
override fun createTabActions(): Array<AnAction> {
    return super.createTabActions().map {
      object : AnActionWrapper(it) {
        override fun update(e: AnActionEvent) {
          super.update(e)
          e.presentation.isEnabledAndVisible = <our condition here>
        }
      }
    }.toTypedArray()
}
```


What's your deadline for including API changes to 2023.2? I think Dana is planning to merge 2023.1 in the upcoming weeks, and then I can experiment with these solutions and get back to you to confirm whether removing the `final` modifier would still be needed. Does that timeline work for you?

Thanks!

KonstantinHudyakov:
@amaurymedeiros great! Looks like `AnActionWrapper ` could also work here.
For 2023.2 we have a plenty of time. So, you are free to experiment with this.

amaurymedeiros:
Thanks again for the help! I'm closing this PR, as the suggestions mentioned above worked on our end.

PR #2378, opened 2023-04-04
PY-58672 - add support for highlighting global and local variables

rendner:
https://youtrack.jetbrains.com/issue/PY-58672/The-Python-Color-Scheme-Settings-preview-incorrectly-shows-that-colors-for-things-like-local-variables-will-be-applied-when-in


ybedrov:
Hello! Have you checked the issue with 2023.2 EAP?

rendner:
@ybedrov Thank you for the hint. Works for me :).

PR #2385, opened 2023-04-06
fix: Never archive .DS_Store files

bric3:
Without this change, the builder will likely encounter .DS_Store files created automatically by the Finder.
This results in the following error

```
FATAL: java.lang.IllegalStateException: Only Info.plist file is allowed in IntelliJ IDEA CE.app/Contents directory but found IntelliJ IDEA CE.app/Contents/.DS_Store
```

It's unlikely the distribution wants to have .DS_Store files, so let's filter them out.

Should fix [IDEA-317409 Only Info.plist file is allowed in IntelliJ IDEA CE.app/Contents directory but found IntelliJ IDEA CE.app/Contents/.DS_Store ](https://youtrack.jetbrains.com/issue/IDEA-317409/Only-Info.plist-file-is-allowed-in-IntelliJ-IDEA-CE.app-Contents-directory-but-found-IntelliJ-IDEA-CE.app-Contents-.DSStore)


DmPanov:
[Merged](https://github.com/JetBrains/intellij-community/commit/94d6d4aa9d5f24ce6be28ef1444af75d5eea1103), thank you!

bric3:
You're welcome, thank you ! @DmPanov

PR #2387, opened 2023-04-07
ControlFlow: Generate WRITE for field initializers

SirYwell:
Previously, the `ControlFlowAnalyzer` did not emit a `WRITE` instruction for field initializers. This is rather unexpected, as it generates instructions for other field writes. Looking at the `processVariable` method, it also looks like such usage was already intended: It checks if the variable is either a local variable or a field.


SirYwell:
It seems like this PR didn't like the force-push. I don't know if that's relevant for you as you don't merge the PR via GitHub. The only relevant commit is [JetBrains/intellij-community@`fa84d78` (#2387)](https://github.com/JetBrains/intellij-community/pull/2387/commits/fa84d789dd8a4ca2fc3bdbbc3e609ed604343841). 

Just let me know if I need to do anything :)

amaembo:
Please create a clean PR that contains only your changes. Thanks.

PR #2391, opened 2023-04-07
PY-60057 - fix endless resolve for nonlocal target

rendner:
https://youtrack.jetbrains.com/issue/PY-60057/Semantic-Highlighting-and-nonlocal-target-result-in-java.lang.StackOverflowError


east825:
Merged! Thanks for investigating the problem and submitting a patch. Sorry it took so long to process it.

PR #2393, opened 2023-04-10
KTIJ-25176 False positive `Redundant nullable return type` inspection in case of getter with nullable condition

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-25176


abelkov (commented):
@t-kameyama Why is it necessary to extract `dataFlowValueFactory`, `moduleDescriptor` and `languageVersionSettings`?

t-kameyama:
@abelkov 
Since `actualReturnTypes` function may be called twice, I thought it should be reused.
No need to reuse?

abelkov (approved):
Thank you, makes sense.

PR #2394, opened 2023-04-11
Suggest to run all inherited tests for KMP common source set

vmadalin:
### Description
This issue was reported on studio side https://issuetracker.google.com/issues/275409155, where given a KMP project  containing the bellow code in `commonTest` source set, the run gutter icon on the `myTest()` doesn't display the sub-class to run the test, instead asks for the target (jvm, android, etc). At the same time, the `Child1` and `Child2` aren't marked as test classes (displaying the run gutter icon), since they are implementing a class that contains a test.

```kotlin
abstract class BaseTestCommon {
    @Test
    fun myTest() {}
}

class Child1 : BaseTestCommon()

class Child2 : BaseTestCommon()
```

As a note this works properly when the same test is under `androidTest` or `jvmTest` source set, meaning that is `common` specific. 

### Changes
The `InheritorChooser.runMethodInAbstractClass()` is responsible to display the popup regarding which implementation of abstract class to use for the test execution, been triggered from:
 - when is class: `AbstractKotlinMultiplatformTestClassGradleConfigurationProducer`
 - when is method: `AbstractKotlinMultiplatformTestMethodGradleConfigurationProducer`
 
At the same time, each source set implements its own `GradleConfigurationProducer`. This validates given a method if its `PsiClass` contains a modified property `abstract`. Due that in case of `common` the `getPsiMethodForLocation()` and `getPsiClassForLocation()` respectively returning `fakes` objects causing to be unable to determine given a dummy class/method if contains or not `abstract` modifier. Those changes where introduced by the following [commit](https://github.com/JetBrains/intellij-community/commit/d051ad23f886d5101a20542e12131038d8548f83) addressing  [KTIJ-15137](https://youtrack.jetbrains.com/issue/KTIJ-15137/) that is still required to be considered due that the `expect` and `actual` modifiers aren't  supported by the existing `KotlinAsJavaSupport` returning `null` on those cases, when trying to create the `lightClass` or `lightMethod`. 

Regarding the fact that `Child1` and `Child2` aren't marked as test classes is because the existing implementation of `Fe10GenericTestIconProvider.isKotlinTestDeclaration()` doesn't take into consideration inherence of any type interfaces or abstract classes, for this reason, now the `isKotlinTestDeclaration()` will be invoked for all class super classifiers.

### Overview
- Before
Screenshot 2023-04-08 at 20 39 20

- After

https://user-images.githubusercontent.com/18151158/231167616-a0b44cab-0c78-43e3-a8ba-58529c360007.mov

### DIFF
--- a/plugins/kotlin/base/fe10/code-insight/src/org/jetbrains/kotlin/idea/base/fe10/codeInsight/tooling/Fe10GenericTestIconProvider.kt
+++ b/plugins/kotlin/base/fe10/code-insight/src/org/jetbrains/kotlin/idea/base/fe10/codeInsight/tooling/Fe10GenericTestIconProvider.kt
@@ -9,11 +9,14 @@ import org.jetbrains.kotlin.idea.base.codeInsight.KotlinTestAvailabilityChecker
 import org.jetbrains.kotlin.idea.base.codeInsight.tooling.AbstractGenericTestIconProvider
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
 import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperClassifiers
 
 object Fe10GenericTestIconProvider : AbstractGenericTestIconProvider() {
     override fun isKotlinTestDeclaration(declaration: KtClassOrObject): Boolean {
         val descriptor = declaration.resolveToDescriptorIfAny() ?: return false
-        return isKotlinTestDeclaration(descriptor)
+        return descriptor.getAllSuperClassifiers().any {
+            isKotlinTestDeclaration(it)
+        }
     }
 
     private tailrec fun isIgnored(descriptor: DeclarationDescriptor): Boolean {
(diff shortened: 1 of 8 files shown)


terrakok:
Hi! Thank you for the contribution!
Fix looks valid from my point of view. But could you please avoid `mockito` test framework? It shouldn't be difficult to write a test without it.

vmadalin:
> 

Unfortunately, I didn't find any better way of avoiding getting the below exception because `JBPopupFactory` is trying to display the popup on the tests. Probably having a flag to disable it will be much more desirable.
```
com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: Access from Event Dispatch Thread (EDT) is not allowed; see https://jb.gg/ij-platform-threading for details
Current thread: Thread[AWT-EventQueue-0,6,main] 1033856359 (EventQueue.isDispatchThread()=true)
SystemEventQueueThread: (same)
```

terrakok:
Use the following code instead of mockito:

```kotlin
private fun mockInheritorPopup() {
    application.replaceService(
        JBPopupFactory::class.java,
        object : PopupFactoryImpl() {
            override fun <T : Any?> createPopupChooserBuilder(list: MutableList<out T>): PopupChooserBuilder<T> {
                @Suppress("UNCHECKED_CAST")
                list as? List<PsiClass>
                return PopupChooserBuilder(JList())
            }
            override fun guessBestPopupLocation(dataContext: DataContext) =
                RelativePoint(Point())
        },
        testRootDisposable
    )
}
```

Could you please make related issue in [YouTrack](https://youtrack.jetbrains.com/newIssue?project=KTIJ) and rename commits to `issue_id: title`?

PR #2397, opened 2023-04-12
Don't compare Android Studio and IDEA versions

dahlstrom-g:
@jreznot, @yanex
This comparison causes PluginStartupActivity.checkPluginCompatibility to fail because Android Studio's version numbers don't match IDEA's.
See for example android.googlesource.com/platform/tools/idea/+/efacdfcb.

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/PluginStartupActivity.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/PluginStartupActivity.kt
@@ -65,7 +65,8 @@ internal class PluginStartupActivity : ProjectPostStartupActivity {
         val rawVersion = KotlinIdePlugin.version
         val kotlinPluginVersion = KotlinIdePluginVersion.parse(rawVersion).getOrNull() ?: return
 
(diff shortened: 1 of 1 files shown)


dahlstrom-g:
The exact versions are, for example, those built from the commit given above, efacdfcb.
Here's another example: android.googlesource.com/platform/tools/idea/+/9ff2225e

yanex:
Do I understand right that the build version (`ApplicationInfo.getInstance().build.baselineVersion`) is correct in that case?

dahlstrom-g:
Yes, the baseline version in the build number seems correct to me. Thanks!

dahlstrom-g:
Can you put a commit on the master branch implementing the check as you envision, or do you want me to try that in a new commit on this pull request?

dahlstrom-g:
@yanex?

dahlstrom-g:
@yanex: For context, I'm updating the intellij-community repository in Android Studio right now and I'd like to have a shared solution to this in the next few days. Can you push a commit to the master branch?

dahlstrom-g:
@yanex: ping?

yanex:
Merged the alternative fix (https://github.com/jetbrains/intellij-community/commit/ba2f6ac5).
Thanks for noticing the problem!

PR #2400, opened 2023-04-13
[gradle] Don't mark includeBuild modules as buildSrc modules

fp7:
Issue: IDEA-317967

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleBuildSrcProjectsResolver.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleBuildSrcProjectsResolver.java
@@ -308,7 +308,7 @@ public final class GradleBuildSrcProjectsResolver {
         }
       }
       else {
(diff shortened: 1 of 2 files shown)


nskvortsov:
I apologize for the delay
@fp7 there are some problems with the pull request (too much commits included). Would you mind cleaning it up a bit?

fp7:
> I apologize for the delay 

No worries.

> there are some problems with the pull request (too much commits included). Would you mind cleaning it up a bit?

Yeah, sure. I rebased it.

nskvortsov:
I took another look at the problem and the suggested solution and found it viable this time. I made a slight modification for the proposed pull request (removed the replacement, simply dropped the call to `setBuildSrcModule`) and now run it through the CI.

fp7:
I investigated this some time ago again, too and came up with a different pull request that would solve the same problem i have in a different way. It also solves the rebuild problem if the module is a buildSrc module for real.

https://github.com/JetBrains/intellij-community/pull/2503.

PR #2403, opened 2023-04-15
PY-39352 - use the more accurate term "CapWords" instead of "CamelCase"

rendner:
https://youtrack.jetbrains.com/issue/PY-39352


east825:
Thanks for the pull request!

PR #2404, opened 2023-04-17
Replace with safe (.?) call" does not add right value of elvis operator when passing expression as parameter to infix function call 

gyoussef55:
https://youtrack.jetbrains.com/issue/KTIJ-114/Replace-with-safe-.-call-does-not-add-right-value-of-elvis-operator-when-passing-expression-as-parameter-to-infix-function-call


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ReplaceCallFixUtils.kt
```diff
@@ -15,10 +15,36 @@ import org.jetbrains.kotlin.resolve.calls.util.getType
...
 fun KtExpression.shouldHaveNotNullType(): Boolean {
     val type = when (val parent = parent) {
-        is KtBinaryExpression -> parent.left?.let { it.getType(it.analyze()) }
+        is KtBinaryExpression ->{
+            val default = parent.left?.let { it.getType(it.analyze()) }
+            val resolvedCall = parent.operationReference.resolveToCall()
+            val function = resolvedCall?.resultingDescriptor as? FunctionDescriptor
+            val isInfix = function?.isInfix ?: false
+
+            if (isInfix) {
```
abelkov, line 30: Does it work correctly if the function is not infix?

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ReplaceCallFixUtils.kt
```diff
@@ -15,10 +15,36 @@ import org.jetbrains.kotlin.resolve.calls.util.getType
...
 import org.jetbrains.kotlin.types.typeUtil.isTypeParameter
 import org.jetbrains.kotlin.types.typeUtil.isUnit
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.types.typeUtil.makeNullable
+
 
 fun KtExpression.shouldHaveNotNullType(): Boolean {
     val type = when (val parent = parent) {
-        is KtBinaryExpression -> parent.left?.let { it.getType(it.analyze()) }
+        is KtBinaryExpression ->{
```
abelkov, line 24: Please format code

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/quickfix/K1QuickFixTestGenerated.java
```diff
@@ -13112,6 +13112,22 @@ public void testNormal() throws Exception {
         public void testReturn() throws Exception {
             runTest("testData/quickfix/replaceWithSafeCall/return.kt");
         }
+        
+           @TestMetadata("Test1.kt")
```
abelkov, line 13116: Test should be generated automatically with "Generate Kotlin Tests (K1 + K2)" run configuration

plugins/kotlin/idea/tests/testData/quickfix/replaceWithSafeCall/Test1.kt
```diff
@@ -0,0 +1,12 @@
+// "Replace with safe (?.) call" "true"
```
abelkov, line 1: Please use proper formatting in tests (and no redundant empty lines) and meaningful file names.

abelkov (changes requested):

abelkov:
Please squash all commits into one and write a proper commit message. Example: https://github.com/JetBrains/intellij-community/commit/45c4a243e61478d1951a945af2f9a5a8a093d309

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/quickfix/ReplaceCallFixUtils.kt
gyoussef55: Thank you for the feedback. I will work on the requested changes.
I only handled the problem of infix function call, however. 
I did check a few non-infix functions and found that they worked properly. Additionally, my solution passed all of the provided tests.
Should I fix the bug also in K2 ? 
 
 
 
I am also discovered another problem that if the infix function has before it '.' and parenthesized argument it would work correctly in some cases, it happens in k1 and k2 compiler.  
intellij

gyoussef55: Thank you for the feedback. I will work on the requested changes.
I only handled the problem of infix function call, however. 
I did check a few non-infix functions and found that they worked properly. Additionally, my solution passed all of the provided tests.
Should I fix the bug also in K2 ? 
 
I am also discovered another problem that if the infix function has before it '.' and parenthesized argument it would work correctly in some cases, it happens in k1 and k2 compiler.  

intellij

abelkov:
Thank you, I will be on vacation until May 15, I will take a look later.

abelkov:
@gyoussef55 Currently this PR doesn't have changes. Can you please actualize it (and squash into one commit)?

PR #2405, opened 2023-04-17
KTIJ-24962 "Unused unary operator": add quickfix to move operator to the previous line to make it binary

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-24962

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedUnaryOperatorInspection.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedUnaryOperatorInspection.kt
@@ -7,11 +7,14 @@ import com.intellij.codeInspection.LocalQuickFix
 import com.intellij.codeInspection.ProblemDescriptor
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiDocumentManager
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
+import com.intellij.psi.codeStyle.CodeStyleManager
 import com.intellij.psi.util.prevLeafs
+import com.intellij.refactoring.suggested.endOffset
 import com.intellij.refactoring.suggested.startOffset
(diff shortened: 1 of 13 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedUnaryOperatorInspection.kt
```diff
@@ -56,7 +82,29 @@ class UnusedUnaryOperatorInspection : AbstractKotlinInspection() {
             prefixExpression.replace(baseExpression)
         }
     }
+
+    private class MoveUnaryOperatorToPreviousLineFix : LocalQuickFix {
```
abelkov, line 86: Probably worth to make it a `HighPriorityAction`?

plugins/kotlin/idea/tests/testData/inspectionsLocal/unusedUnaryOperator/moveToPrevLine/bigDecimal.kt.after
```diff
@@ -0,0 +1,6 @@
+// FIX: Move unary operator to previous line
+// WITH_STDLIB
+fun main() {
+    val a = "2".toBigDecimal() -
+    "4".toBigDecimal()
```
abelkov, line 5: I think it would be nice to automatically format this line.

abelkov (changes requested):

plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedUnaryOperatorInspection.kt
t-kameyama: 74a9c0d03e5ebf96e19e384be3d1ad6d13b397c4

plugins/kotlin/idea/tests/testData/inspectionsLocal/unusedUnaryOperator/moveToPrevLine/bigDecimal.kt.after
t-kameyama: d1ab4bc2970f6466696a17aee69371dbc8d8f12d

abelkov (approved):

PR #2408, opened 2023-04-19
update scipy/numpy docs map

guoci:
Update external documentation map for scipy and numpy.
https://youtrack.jetbrains.com/issue/PY-36368


east825:
Thanks for updating the documentation mapping! Sorry it took so awfully long to get to the PR.

PR #2413, opened 2023-04-21
[kotlin] fix nested lambda breakpoint skipping

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-25056/Debugger-lambda-breakpoint-is-skipped-in-case-it-has-only-another-lambda-call-inside

Previously, KotlinLineBreakpointType used a special heuristic to detect if we should stop on a location. This heuristic picked a lambda on a current source position by an ordinal encoded in the breakpoint, and then it tried to match this lambda with the one the breakpoint was put in. However, it turned out that this heuristic picks wrong lambdas in some cases, and simply using the `getContainingMethod` function works everywhere. So this commit deletes the heuristic in favour of the mentioned function.


zuevmaxim (approved):

yanex:
Change in behaviour for this case:

```
fun main() {
    lambda("first", fun(it: String) {
        // ← lambda breakpoint
        lambda("second", fun(it: String) {
            println()
        })
    })
}

fun lambda(s: String, l1: (String) -> Unit) {
    l1(s)
}
```

There are a few things that go weird, though:

1. It's unclear if lambda breakpoints should be suggested for anonymous functions. Probably not. It's out of the scope of this PR, though.
2. Even though (1), lambda breakpoints stopped working.
3. Line breakpoints also don't work on the `lambda("second", fun(it: String) {` line, so it became harder to get inside the functions.

I don't have a strong opinion on whether this needs to be fixed in this PR. However, I strongly suggest documenting it somewhere in YouTrack so it won't silently become a degradation.

nikita-nazarov:
I conducted some research on the example above and found out some interesting things. 
So upon reaching the `org.jetbrains.kotlin.idea.debugger.breakpoints.KotlinLineBreakpointType#getContainingMethod` method call, the breakpoint that is passed in will have its source position set to the opening bracket: 
`lambda("second", fun(it: String) /* This one -> */ {` 
Because of that, the `getContainingMethod` call will return the second lambda as expected, and since we placed the breakpoint in the first lambda, it will be skipped. This happens, because when `KotlinPositionManager` tries to guess the element that we stopped on for this line: `lambda("second", fun(it: String) {`, instead of picking the `lambda` element, it picks the opening bracket `{`. This method is to blame: `org.jetbrains.kotlin.idea.debugger.KotlinPositionManager#calculatedClassNameMatches`.
I will investigate further on how it would be best to fix this issue.

zuevmaxim:
Merged as 80a64312ad538b464cd861bf255ed8d3519525b9

zuevmaxim:
https://youtrack.jetbrains.com/issue/IDEA-320295

PR #2415, opened 2023-04-22
KTIJ-25345 Intention 'Replace with a 'for' loop' on nullable receiver should automatically surround the call with a null check

t-kameyama:
https://github.com/JetBrains/intellij-community


abelkov (approved):
Thank you. I will remove `safeAs` usage, because it is less safe than pure `as?` operator.

PR #2417, opened 2023-04-25
Fix EditorLineStart home shortcut in VSCode on OSX

benediktwerner:
Currently, both start and end of line are mistakenly bound to the "end" key.


NikitkoCent:
@dmimat, looks like it's for you. Unfortunately I can't assign the issue to you

dmimat:
@NikitkoCent , I'll fix that. Thanks!

PR #2420, opened 2023-04-30
PY-60408 - format numpy arrays

rendner:
https://youtrack.jetbrains.com/issue/PY-60408


rendner:
Issue is marked as fixed.

PR #2421, opened 2023-05-01
PY-60372 - use formatted min and max

rendner:
https://youtrack.jetbrains.com/issue/PY-60372


vladimir-koshelev:
The original issue was fixed.

PR #2424, opened 2023-05-05
ControlFlow: Generate WRITE for field initializers

SirYwell:
Fresh PR after #2387 broke due to the force-push.
Original description:

Previously, the ControlFlowAnalyzer did not emit a WRITE instruction for field initializers. This is rather unexpected, as it generates instructions for other field writes. Looking at the processVariable method, it also looks like such usage was already intended: It checks if the variable is either a local variable or a field.

cc @amaembo


amaembo:
Which problem are you trying to solve? Do you experience incorrect behavior of IDEA somewhere? Or is it necessary for your plugin? In the latter case, could you please shed more light on what are you trying to achieve? Thanks.

amaembo:
Thank you for explanantion. Merged. Hopefully, this won't introduce regressions.

amaembo:
e9572c832412482f83e1e4b017ff27434d57d8d8

PR #2425, opened 2023-05-09
IDEA-319974 Expose updateModifiedProperty more generally

andymagee:
Some editors other than TextEditorImpl need to have their modified property updated when a file is dropped from the unsaved file list. An example is a split editor used for markdown files.

Bug: https://youtrack.jetbrains.com/issue/IDEA-319974


andymagee:
Huh. In the day since I last pulled, TextEditorImpl.java was deleted... closing this PR and will look again.

PR #2427, opened 2023-05-10
Expose externalProjectPath on ExternalSystemNotificationExtension.customize()

vmadalin:
This was added to be able to support multiple gradle root projects, providing the `externalProjectPath` on the  `ExternalSystemNotification.customize` allowing to provide better error message to users


HackerMadCat:
Hi! @vmadalin I cannot merge this pull request. There are a lot of compilation and usages conflicts.

1. ExternalSystemNotificationExtension#customize without path is used inside Pants external plugin. So we need to deprecate old one method.
2. It has conflicts with android part. Android also has a lot of usages of customize and createFailureResult. Inside: `GradleBuildInvoker`, `GradleJvmNotificationExtension` and `GradleJvmNotificationExtensionTest`

Please fix it and attach minimized Android commits.

PR #2428, opened 2023-05-11
KTIJ-11579 Recursive call icon in gutter not shown if function name is not a valid java identifier (but still a valid Kotlin&JVM identifier)

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-11579


abelkov (approved):

LouisCAD:
Thank you for fixing that!

PR #2429, opened 2023-05-11
IDEA-318834 Fix processing Unix-style path with system specific separator on SourceFolderManagerImpl

vmadalin:
Vfs event processing in `SourceFolderManagerImpl` does not handle windows paths correctly. More specifically, [here](https://github.com/JetBrains/intellij-community/blob/0b02d9193dbe991ae4b677b4fc2d9c2706bab8b8/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/project/manage/SourceFolderManagerImpl.kt#L160) we try to find source folders nested under the changed path.

`VfsUtilCore.pathToUrl(event.path)` is used to get the changed path, which is Unix-style.

But, when processing this path, we split it based on file system specific separator, so prefix-style processing does not work on windows:

```kotlin
private object UrlPrefixFactory : AbstractPrefixTreeFactory<String, String>() {
    override fun convertToList(element: String): List<String> {
      return element.removeSuffix(File.separator).split(File.separator)
    }
  }
```


dahlstrom-g:
@HackerMadCat: Any update here?

HackerMadCat:
I am going to merge it before next release. But now it isn't available in EAP.

dahlstrom-g:
Thanks! Meanwhile does it make sense on the `master` branch?

PR #2430, opened 2023-05-11
Expose extensions for the GradleJvm resolvers to allow custom macros

vmadalin:
To be able to trigger Gradle sync the project requires to contain a  valid `jdk.table` entry defined on the `gradleJvm` under `.idea/gradle.xml` file. This isn't necessary when macros are in place like `#JAVA_HOME` that are able to bypass this requirement and the validation defined on [LocalGradleExecutionAware.prepareJvmForExecution()](https://github.com/JetBrains/intellij-community/blob/06a40bca3c613b5290153473655afa95d299c268/plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/LocalGradleExecutionAware.kt#L99). 

Until now, the resolution of the `gradleJvm` was happening between [GradleJvmUtil.nonblockingResolveGradleJvmInfo](https://github.com/JetBrains/intellij-community/blob/master/plugins/gradle/src/org/jetbrains/plugins/gradle/util/GradleJvmUtil.kt#L32) and [ExternalSystemJdkUtil.nonblockingResolveJdkInfo](https://github.com/JetBrains/intellij-community/blob/master/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/execution/ExternalSystemJdkUtil.kt#L14) where each one of them were responsible to handle some supported macros, otherwise the specified `gradleJvm` will validated as a `jdk.table` entry name.

This logic isn't properly exposed to the API in order to support new resolvers for macros but also other special cases. For this reason,  a dedicated abstract class called `GradleJvmResolver` as been created as an `extensionPoint` allowing to create easily define any new resolvers or override the current implementation for the existing ones.

The existing logic was moved to the following resolvers each one responsible for a specific macro:
- `GradleJvmGradlePropertiesResolver` (#GRADLE_JAVA_HOME)
- `GradleJvmGradleLocalPropertiesResolver` (#GRADLE_LOCAL_JAVA_HOME)


plugins/gradle/testSources/org/jetbrains/plugins/gradle/util/ExternalSystemJdkNonblockingUtilTest.kt
nskvortsov, line 1: Relocation of this test looks suspicious to me. Why would it be necessary, given that the test checks platform code? (ExternalSystem*  is considered a part of the IntelliJ platform). It should not be Gradle-only

plugins/gradle/testSources/org/jetbrains/plugins/gradle/util/ExternalSystemJdkNonblockingUtilTestCase.kt
nskvortsov, line 1: Same suspicious rellocation

nskvortsov (changes requested):
see the comments inline

plugins/gradle/testSources/org/jetbrains/plugins/gradle/util/ExternalSystemJdkNonblockingUtilTest.kt
vmadalin: Thank you for pointing this out, seems I missed cleaning up this part. The original implementation [ExternalSystemJdkUtil.nonblockingResolveJdkInfo](https://github.com/JetBrains/intellij-community/blob/master/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/execution/ExternalSystemJdkUtil.kt#L14) was only used on tests and from the `GradleJvmUtils`.
Screenshot 2023-06-01 at 09 17 35

Due to that, I consider it necessary to just move everything under the `gradle` package because resolves JDK info from the specified `gradleJvm` field under `.idea/gradle.xml` file. Avoiding in this way having not used logic exposed unnecessarily for `External System`. With the latest update, I merged both classes with the existing tests and removed them completely.

For sure in the future, we can consider potentially renaming the [GradleJvmResolver.kt](https://github.com/JetBrains/intellij-community/pull/2430/files#diff-ecd3dbe55e4f6c82979f1e5165694688670787815aa2ef032b4d496d0bfea58f) to something more generic and place it under `ExternalSystem` allowing to define some common `Resolvers` for different build systems. In any case, since this isn't exactly the case right now probably this is a much clean option for now.

nskvortsov: Indeed `nonblockingResolveJdkInfo` is used only by Gradle plugin at the moment. But `ExternalSystemJdkUtil` overall has about 100 usages across IntelliJ Community, some usages in IDEA ultimate and Fleet, and is used by around 7 third-party plugins.
Merging the whole class with the `GradleJvmUtils` is not acceptable at this point.

nskvortsov:
Here is the related issue: [IDEA-339715](https://youtrack.jetbrains.com/issue/IDEA-339715)

nskvortsov:
running the CI now.

nskvortsov:
Merged https://github.com/JetBrains/intellij-community/commit/2aeebdce1d4bb943c6b46bd58ea2bea37d0120ed

PR #2432, opened 2023-05-12
IDEA-320384 Display Gradle unsupported Java version based on SupportMatrix

vmadalin:
https://youtrack.jetbrains.com/issue/IDEA-320384/

The existing implementation of [IncompatibleGradleJdkIssueChecker](https://github.com/JetBrains/intellij-community/blob/master/plugins/gradle/src/org/jetbrains/plugins/gradle/issue/IncompatibleGradleJdkIssueChecker.kt) is able to handle properly this kind of exceptions when Gradle version is `<= 5.0` but when using `4.10.1` the following exception is thrown:

```
* What went wrong:
Unable to make protected void java.net.URLClassLoader.addURL(java.net.URL) accessible: module java.base does not "opens java.net" to unnamed module @7a187f14

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
java.lang.reflect.InaccessibleObjectException: Unable to make protected void java.net.URLClassLoader.addURL(java.net.URL) accessible: module java.base does not "opens java.net" to unnamed module @7a187f14
	at org.gradle.internal.reflect.JavaMethod.<init>(JavaMethod.java:42)
	at org.gradle.internal.reflect.JavaMethod.<init>(JavaMethod.java:32)
	at org.gradle.internal.reflect.JavaMethod.<init>(JavaMethod.java:36)
```
- After
Screenshot 2023-05-12 at 12 10 20

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/issue/IncompatibleGradleJdkIssueChecker.kt
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/issue/IncompatibleGradleJdkIssueChecker.kt
@@ -65,20 +65,15 @@ class IncompatibleGradleJdkIssueChecker : GradleIssueChecker {
       }
     }
 
-    var isJavaGroovyCompatibilityIssue = false
-    var isJavaByteCodeCompatibilityIssue = false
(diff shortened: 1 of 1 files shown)


nskvortsov:
lgtm, running CI checks

PR #2435, opened 2023-05-14
[maven] IDEA-320329 - non-case sensitive unique artifactId in maven project import

seregamorph:
## Motivation

Fixes https://youtrack.jetbrains.com/issue/IDEA-320329

## Proposed fix
Do a non-case sensitive comparison when calculating possible repeated module names.

## Notes
This implementation does not fix the conflicting module names if according IDEA module already exists, it will only work if the module is new or project is under initial import.
This change can break shared run configurations for multi-module projects with non-unique artifactId names as some modules will now have another name on import from scratch.

### DIFF
--- a/plugins/maven/src/main/java/org/jetbrains/idea/maven/importing/MavenModuleNameMapper.java
+++ b/plugins/maven/src/main/java/org/jetbrains/idea/maven/importing/MavenModuleNameMapper.java
@@ -13,7 +13,10 @@ import org.jetbrains.idea.maven.project.MavenProject;
 import java.io.File;
 import java.util.*;
 
+import static java.util.Locale.ROOT;
+
 public final class MavenModuleNameMapper {
+
   public static void map(Collection<MavenProject> projects,
                          Map<MavenProject, Module> mavenProjectToModule,
                          Map<MavenProject, String> mavenProjectToModuleName,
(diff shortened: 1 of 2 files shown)


plugins/maven/src/main/java/org/jetbrains/idea/maven/importing/MavenModuleNameMapper.java
```diff
@@ -74,19 +78,19 @@ private static void resolveModuleNames(Collection<MavenProject> projects,
...
+          nameCountersLowerCase.put(nameItem.originalName.toLowerCase(ROOT), c + 1);
         }
 
         do {
           String name = nameItem.getResultName();
           if (existingNames.add(name)) break;
 
           nameItem.number++;
-          nameCounters.put(nameItem.originalName, nameItem.number + 1);
+          nameCountersLowerCase.put(nameItem.originalName.toLowerCase(ROOT), nameItem.number + 1);
```
seregamorph, line 93: it could make sense to store new field `originalNameLowerCase` in the class `NameItem`

plugins/maven/src/test/java/org/jetbrains/idea/maven/importing/MavenModuleNameMapperTest.java
zulkar, line 1: we try to avoid using mocks in tests as possible. Most tests in maven plugin are integration tests by their nature.

I'll rewrite the tests.

PR #2443, opened 2023-05-18
Fix adding existing conda interpreters edge cases

tl-hbk:
There's two edge cases where adding an existing interpreter fails.

Assuming envs_dirs is `/home/user/miniconda3/envs` in `.condarc`:
* If `envPath` is in a subdirectory of `envs_dirs` (`/home/user/miniconda3/envs/foo/bar`)
* If `envPath` starts with `envs_dirs` but isn't actually in the directory (`/home/user/miniconda3/envsfoo/bar`)

Trying to add the above environments through PyCharm would attempt to add `NamedEnvs` that don't exist since `NamedEnv` takes the default `envs_dirs` and appends the name to calculate the expected path.

To replicate the second example

```
conda create --prefix ~/miniconda3/envsfoo/pycharm python=3.8
```

Then if you try adding the environment in PyCharm this gets logged in idea.log

```
EnvironmentLocationNotFound: Not a conda environment: /home/user/miniconda3/envs/pycharm 
```


tl-hbk:
Hi I was wondering if there was an ETA on when this is expected to be looked at?

throwable-one:
Hello @tl-hbk 
Thank you for this change!

Could you please add test to ``PyCondaTest`` for that?

You might need to refactor ``PyCondaEnv`` a little to fetch ``envs_dirs``

throwable-one:
@tl-hbk 
``PyCondaEnv.createEnv`` should be able to do so. 

See ``testCondaCreateEnv``. But first install conda and set ``CONDA_PATH`` env var (See ``LocalCondaRule``)

tl-hbk:
@throwable-one I've added a test that checks for the two edge cases that I mentioned.  I'm not at all familiar with Kotlin so please let me know if there's any issues

throwable-one:
Hello @tl-hbk 
Thank you.

Here are some comments:

``Path.of`` is ok for local paths, but not for remote ones (SSH, WSL etc). Please use ``TargetCommandExecutor.local`` and only use ``Path`` for local paths. Use old approach for remote. We are going to support ``nio.Path`` for remote targets and after we do so, I will remove this check. 

Kotlin supports operators overloading, so you may use equals sign.

```
if (command.local) Path.of(it) == Path.of(envPath).parent else envPath.startsWith(it)
```

In Kotlin, use ``val`` whenever possible. Variable names are in "lowerCamelCase". (``envDirs``)

No need to call ``PyCondaEnv.getEnvs`` several times, and since Kotlin supports smart casts and functional style programming, we can do something like

```
val envs = PyCondaEnv.getEnvs(condaRule.commandExecutor, condaRule.condaPathOnTarget).getOrThrow()
      .map { it.envIdentity }
      .filterIsInstance<PyCondaEnvIdentity.UnnamedEnv>()
Assert.assertTrue("No child $childEnvPrefix in $envs", envs.any { it.envPath == childEnvPrefix })
Assert.assertTrue("No sibling $siblingEnvPrefix in $envs", envs.any { it.envPath == siblingEnvPrefix })
```

We did some changes in this are, so please pull the latest version. Thank you

tl-hbk:
I've made the suggested changes and merged master into the branch.  

Currently away from the computer I have the dev environment set up on though so I haven't been able to re-run the tests yet.  Does this repo have CI setup to run against PRs?

throwable-one:
I just checked it, and tests passed. The patch looks good to me.  

I will amend two last commits to  3d26b26443d6af761ef198d8f729267a3081d4a5 so we have one commit (instead of 3) and a good commit message, ok?

tl-hbk:
Sure sounds good thanks for taking a look at this!

PR #2449, opened 2023-05-22
IDEA-320889 New clean up: Lombok @Getter may be used

Fabrice-TIERCELIN:
Hi @amaembo,

***What steps will reproduce the feature?***

1. Go to *File* → *Settings...* → *Editor* → *Inspections*
2. Enable *Lombok @Getter may be used*
3. Enable Severity: *Warning*
4. Add Lombok in the project dependencies
5. Write this Java class:

```Java
  public class MyClass {
    /**
     * The process date.
     */
    private Date processDate;

    /**
     * Returns the date.
     *
     * @return The date
     */
    public Date getProcessDate() {
      return processDate;
    }
  }
```

6. Run clean up on the class

The inspection *Class MyClass may use Lombok @Getter* is suggested  and the class is rewritten like this:

```Java
  import lombok.Getter;

  @Getter
  public class MyClass {
    /**
     * The process date.
     * -- GETTER --
     * Returns the date.
     *
     * @return The date
     */
    private Date processDate;
  }
```

***What happens instead?***
The inspection *Class MyClass may use Lombok @Getter* is not suggested.

This new feature lombokify the code with the `@Getter` annotation. It adds the annotation on the fields or on the class when possible. It handles the static fields, the boolean fields, the inner classes and it also moves the JavaDoc to the field. It only processes if the Lombok library is detected.

I have written some unit tests that run successfully.


Fabrice-TIERCELIN:
Merged

PR #2453, opened 2023-05-25
Fix smart stepping into lambdas inlined multiple times

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-25635/Debugger-smart-stepping-into-inline-lambdas-doesnt-work-when-they-are-inlined-multiple-times

The smart stepping machinery is designed in a way that it needs to put a breakpoint in a place of interest, and then it is trying to step until the breakpoint is hit. However, when trying to smart step into inline lambda that is inside an inline function that is inlined several times, it may happen that multiple locations in a class file are sufficient candidates for the smart stepping breakpoint. In that case the breakpoint will be placed only on the first sufficient location, others will be ignored. That's why when trying to step into inline lambda in its first inlined call site, smart stepping works, but it never succeeds after that. This commit adds a workaround for this issue, it makes smart stepping machinery stop if we think that we are already inside a lambda of interest PSI wise.


zuevmaxim:
Merged as 15dfe0dee49333de4d791ec32eafd7b9b27be164

PR #2454, opened 2023-05-26
Disallow Java in-place rename in non-Java files

andymagee:
The Java in-place renamer currently allows in-place renaming for members in any file where a reference appears. But this doesn't always work when the reference to the member is in a non-Java file.

This change prevents Java's in-place rename from running when the reference being renamed is not in a Java file. The target language still has the chance to provide their own renamer if desired.

This change addresses https://youtrack.jetbrains.com/issue/IDEA-321110


amaembo:
Unfortunately, this PR cannot be accepted in its current state. We have proper inplace rename support in many non-Java files. For example, if you develop an IntelliJ IDEA plugin, and have an extension class name declared inside plugin.xml, you can rename it inline right inside the XML without any problems. Disallowing this feature would be a regression.

PR #2457, opened 2023-06-01
Add Protected Hooks to PositionManagerImpl

alonalbert:
Allows AndroidPositionManager to support static methods in interfaces prior to API 24.

See https://youtrack.jetbrains.com/issue/IDEA-321559/Feature-Request-Add-Protected-Hooks-to-PositionManagerImpl


gorrus:
Have you seen `com.intellij.debugger.NameMapper`? Maybe it is enough?

alonalbert:
Have not seen it. Will try.

alonalbert:
I don't think `NameMapper` solves our issues. Sent you a message on Slack.

alonalbert:
Abandoning this approach. Will update another PR

PR #2459, opened 2023-06-03
KTIJ-25715 "Range can be converted to indices or iteration": remove unnecessary parentheses if possible

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-25715


abelkov (approved):
Thank you!

PR #2460, opened 2023-06-04
IJPL-126: Migrate public APIs to use "Predicate" instead of "com.intellij.openapi.util.Condition"

RivanParmar:
This is a WIP pull request. It addresses: https://youtrack.jetbrains.com/issue/IJPL-126

This pull request migrates the IntelliJ platform public APIs to use `java.util.function.Predicate` instead of `com.intellij.openapi.util.Condition`.

Currently I have only migrated functions that have `Condition` as their parameter type, and variables, as there isn't much information given as to how to migrate abstract functions, interfaces, etc. I would like to migrate these too if anyone can provide me with information.

If there are classes that shouldn't have been migrated, then please let me know and I will revert the changes. Also do let me know of classes which I may have left and need to be migrated as well.

### DIFF
--- a/java/java-analysis-impl/src/com/intellij/codeInsight/daemon/impl/analysis/HighlightClassUtil.java
+++ b/java/java-analysis-impl/src/com/intellij/codeInsight/daemon/impl/analysis/HighlightClassUtil.java
@@ -913,7 +913,7 @@ public final class HighlightClassUtil {
           if (!PsiUtil.isInnerClass(base)) return;
 
           if (resolve == resolved && baseClass != null && (!PsiTreeUtil.isAncestor(baseClass, extendRef, true) || aClass.hasModifierProperty(PsiModifier.STATIC)) &&
-              !InheritanceUtil.hasEnclosingInstanceInScope(baseClass, extendRef, psiClass -> psiClass != aClass, true) &&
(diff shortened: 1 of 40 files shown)


chashnikov:
Hi Rivan! 
First of all, thank you for your work, it's important to clean up the sources to get rid of obsolete code. 

I have a few comments about your changes:
* I understand that you changed order of parameters to avoid necessity to explicitly cast to lambda interface, but it doesn't work well in some cases; e.g. it's strange to have parameters in that order in `removeSelectedItems(predicate, list)`, also it's convenient to have the lambda parameter as the last one when calling a method from Kotlin;
* when you mark a method with a javadoc as deprecated, it's better to move the javadoc to the newly added method;
* there is no need to keep package-private methods as deprecated, they cannot have external usages so it's safe to remove them.


But the main problem with such pull requests is that external contributors cannot update usages located in the closed part of intellij Git repository. So if we merge this, we get many new warnings about usages of deprecated code in our closed sources. And actually it's easier for us to add replacements for these methods, and at the same time replace all usages of obsolete methods across all sources in semi-automatic way.

So I'm sorry, but we won't merge this pull request.

RivanParmar:
That's alright! Thanks for the response!

PR #2462, opened 2023-06-05
KTIJ-13683 Override quickfix allows to override package-private methods from Java

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-13683


abelkov (approved):
Thank you!

Looks like for FIR_IDENTICAL tests it's not necessary to add a separate Impl.kt.fir.after file, I will remove it.

PR #2465, opened 2023-06-05
PY-60660 - conditional stub definition (typeshed)

rendner:
Hi, I would like to implement a fix for conditional typeshed stubs in PyCharm ([PY-60660](https://youtrack.jetbrains.com/issue/PY-60660)).

Some stubs are defined by using checks for a specific Python version or the operating system as you can [see here](https://github.com/search?q=repo%3Apython%2Ftypeshed+sys.platform+sys.version_info+repo%3Apython%2Ftypeshed+path%3A*.pyi&type=code).

I already started a PoC which can identify the correct stub depending on the os-platform and Python version.
In case one of the conditions for a stub can't be evaluated, the code falls back to the current behavior - resolves to multiple definitions if there is more then one defined.

Can you take a look at the current implementation and hopefully answer the following questions?

1) Where would you expect/implement such a check for conditional stub definitions if the current one is not a good choice?

2) The current implementation requires excluding of pyi-files from [`assertNotParsed`](https://github.com/rendner/intellij-community-1/commit/d5b39360f9ccc8d15e0db07009b926bcfe77ab2e). Because, these files are now parsed to check if there are conditional definitions. Is this OK for you? If not, what would be the way to go to implement checks for conditional definitions without breaking the current behavior of `assertNotParsed` (some codepointer could help)?

3) To resolve the correct stub definition, the implementation needs to know the Python version (checked by `sys.version_info`) and the operating system (checked by `sys.platform`). What would be the correct way to get the necessary information to implement a correct check with `sys.version_info` and `sys.platform` (which also works for all kind of Python interpreters supported by PyCharm)? I use [`SystemInfo`](https://github.com/JetBrains/intellij-community/blob/275cd68d843eb238535638aedd13c09251c61909/platform/util/src/com/intellij/openapi/util/SystemInfo.java#L21) to get the os-platform. But this class can't be used for a Python remote interpreter where the remote os is a different one.

**Note:**
The test [Py3QuickDocTest::testTypingExtensionsSelf](https://github.com/JetBrains/intellij-community/blob/bb882b63f5f310be487a93d1e8da0e5483ef2f5b/python/testSrc/com/jetbrains/python/Py3QuickDocTest.java#L855-L857) now fails because PyCharm doesn't understand [the following construct](https://github.com/JetBrains/intellij-community/blob/18590c07814601081cfce5dba3bd3ee2d674dc88/python/helpers/typeshed/stdlib/typing_extensions.pyi#L206-L221). Therefore, the correct matching [`Self`](https://github.com/JetBrains/intellij-community/blob/18590c07814601081cfce5dba3bd3ee2d674dc88/python/helpers/typeshed/stdlib/typing_extensions.pyi#L212) can't be detected. Instead [this one is resolved](https://github.com/JetBrains/intellij-community/blob/18590c07814601081cfce5dba3bd3ee2d674dc88/python/helpers/typeshed/stdlib/typing.pyi#L189-L190) (which is wrong).

I verified this by adding `Self: Self` in front of the [`from typing import (`](https://github.com/JetBrains/intellij-community/blob/18590c07814601081cfce5dba3bd3ee2d674dc88/python/helpers/typeshed/stdlib/typing_extensions.pyi#L206) and the test `Py3QuickDocTest::testTypingExtensionsSelf` completes as expected.


east825:
Hi! Thanks for having interest in fixing the issue. The underlying problem is actually a part of a larger, long requested issue [PY-34617](https://youtrack.jetbrains.com/issue/PY-34617) we're working on right now. We can try integrating your PoC with our prototype, but it might be difficult depending on the current state of the latter. @Elizaveta239 can tell more. 

As for your questions,

> Where would you expect/implement such a check for conditional stub definitions if the current one is not a good choice?

PyResolveProcessor is one possible place. I don't see any glaring problems with it, except that some resolve API (such as PyClass.find* family of methods) might ignore it.

> The current implementation requires excluding of pyi-files from [assertNotParsed](https://github.com/rendner/intellij-community-1/commit/d5b39360f9ccc8d15e0db07009b926bcfe77ab2e)

This part is problematic. Our idea was that such checks should become part of PSI-stubs so that they could be persisted and extracted from the indexes without re-parsing the corresponding physical files, similarly to PyExceptPart (see PyExceptPartElementType). 

> What would be the correct way to get the necessary information to implement a correct check with sys.version_info and sys.platform?

LanguageLevel can be extracted from any PSI element in a project with LanguageLevel#forElement. The underlying OS of a configured project SDK in theory could be retrieved from the corresponding Run Target request TargetEnvironmentRequest#getTargetPlatform, but it seems that it doesn't distinguish between macOS and linux. We actually didn't want to support `sys.platform` checks right away and focus on `sys.version_info` first, because those are causing more problems.

Another difference with how we were going to support these is that we planned to consider the checks in .py files as well, not only in .pyi stubs. Also, there is a harder problem of supporting them in the compatibility inspection that should deal with a few possible Python versions, not only the one currently configured for the project, but we wanted to postpone this one for the time being.

Overall, let's check up with @Elizaveta239 first, to make sure your proposal is compatible with what we already have in progress.

Elizaveta239:
Hi @rendner! Thank you very much for your pull request! 
Yes, we're currently working on this issue, on a more wide solution which was mentioned by @east825. This solution doesn't parse corresponding files (and stores info inside Psi-stubs) and works not only with `.pyi`, but also with `.py` files.
Currently, I'm not ready to give any ETA and I'm not sure if we manage to re-use any parts of your implementation, but we'll try! (with saving authorship, of course).
Thank you again!

rendner:
Hi @east825 and @Elizaveta239,

>This part is problematic. Our idea was that such checks should become part of PSI-stubs so that they could be persisted and extracted from the indexes without re-parsing the corresponding physical files, similarly to PyExceptPart (see PyExceptPartElementType).

Your solution will then implement such a stub for the whole if-statement or at least for the if and elif parts of a such a statement.
This sounds good. I was already thinking that there must be a better way than my current approach (from a performance standpoint).

>LanguageLevel can be extracted from any PSI element in a project with LanguageLevel#forElement.

I started with that method, but it doesn't work as expected for elements of a pyi-file, because [`PyiFile.getLanguageLevel`](https://github.com/JetBrains/intellij-community/blob/d070837f27ae125ba6e8b44331e8affc17974c9e/python/python-psi-impl/src/com/jetbrains/python/pyi/PyiFile.kt#L39) returns always the latest LanguageLevel.
For these files [`PyiUtil.getOriginalLanguageLevel`](https://github.com/JetBrains/intellij-community/blob/f456ed06049d735da09cd981043e17e8d3c3e90a/python/python-psi-impl/src/com/jetbrains/python/pyi/PyiUtil.java#L93) does the right job. In the end I decided to take the information from the project Sdk, because I expected all elements would return the same language level as the used Sdk.


>Yes, we're currently working on this issue, on a more wide solution which was mentioned by @east825. This solution doesn't parse corresponding files (and stores info inside Psi-stubs) and works not only with .pyi, but also with .py files.

I will keep an eye on the original problem to see when you have finished your solution. Maybe in the meantime I will also find some time and try to implement the psi-stubs approach for myself, at least for pyi files.


Thank you very much for your detailed answers and the extended insight into the overall problem. 

The PR can therefore be closed, right?

Elizaveta239:
> The PR can therefore be closed, right?
Let's leave it opened so far, until the solution is ready. 

Thank you very much for your contribution!

Elizaveta239:
Hi! I'm not working on the project anymore, so will not be reviewing the changes - but I can't remove myself from the reviewers list. Please, feel free to remove me

vladimir-koshelev:
The issues is fixed.

PR #2467, opened 2023-06-06
IDEA-321906 Exclude module-info.java files from sources zip

gharrma:
zipSourcesOfModules() aggregates sources from many modules and libraries, resulting in conflicting module-info.java files in the output.

These module-info.java files break Kotlin debugging functionality occasionally because they cause JvmModuleAccessibilityChecker to compute an incorrect Java module name for app.jar.

---

Issue link: https://youtrack.jetbrains.com/issue/IDEA-321906


chashnikov:

PR #2469, opened 2023-06-07
IDEA-292947 Skip checking of UNC root in FsRoot

paulthomson:
LocalFileSystemBase.extractRootPath will happily return a UNC path root like "//server/..", but this is rejected by FsRoot because it contains "/..". Change FsRoot to skip this check for UNC roots.

Bug: https://youtrack.jetbrains.com/issue/IDEA-292947

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/impl/FsRoot.java
+++ b/platform/platform-impl/src/com/intellij/openapi/vfs/newvfs/impl/FsRoot.java
@@ -2,6 +2,7 @@
 package com.intellij.openapi.vfs.newvfs.impl;
 
 import com.intellij.openapi.util.SystemInfo;
+import com.intellij.openapi.util.SystemInfoRt;
 import com.intellij.openapi.util.io.FileAttributes;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.vfs.newvfs.NewVirtualFileSystem;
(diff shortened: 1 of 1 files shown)


paulthomson:
As explained in the bug, some strings cause the embedded IDE terminal to "crash" (an exception is thrown, and the terminal is closed) on Windows. It seems to be caused by UNC paths (`//server/share/filepath`). Some strings to try:

* `//test/../filecpp: 3` -> crash
* `//test/..: 3` -> crash (different path through `extractRootPath`)
* `//../aa/filecpp: 3` -> no crash, but causes a log message:

```
 WARN - #c.i.o.v.n.VfsImplUtil - LocalFileSystem has extracted incorrect root '' from '//../aa/filecpp' (original '//../aa/filecpp')
```


Alternative fix: we could change extractRootPath, like this: https://github.com/paulthomson/intellij-community/commit/9abacb103e519e76e1d65c745069c33e97950a86

I have tested both fixes locally on Windows, and they both seem to work. Possibly you could argue that `//server/../` is a valid UNC root (unlike `//../share/`) because the share part is not really restricted by the [UNC reference](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/62e862f4-2a51-452e-8eeb-dc4ff5ee33cc), and thus skipping the check in `FsRoot` (as in this PR) is possibly more correct.

paulthomson:
Is there any update on this?

paulthomson:
Any update?

trespasserw:
I'll take a look. But at the first glance, the fix seems to be at the wrong place.

PR #2472, opened 2023-06-08
IDEA-322117 Convert InspectionProfilerDataHolder to non-light service

radomirm-g:
This converts InspectionProfilerDataHolder from light service to non-light service. This allows for a different implementation of the service in Android Studio.


cdracm:
hi, this class is not intended for overloading, please see if the fix in https://youtrack.jetbrains.com/issue/IDEA-322117 works

PR #2473, opened 2023-06-08
IDEA-316270 Fix leaking HierarchyBrowsers

npaige:
Reuse of a Content across multiple HierarchyBrowsers led to a leak where a lambda which captured the HierarchyBrowser was registered to the Content in the disposer.

HierarchyBrowsers were also registered to a very long-lived (project lifetime) ContentManagerImpl.

Furthermore, the project was used as the parent disposable in the message bus connection in the DumbUnawareHider wrapper for the HierarchyBrowser.

This change removes this reuse of the Content object and addresses all of the above leak paths.


npaige:
@cdracm

PR #2477, opened 2023-06-11
KTIJ-25795 "Variable declaration could be moved inside 'when'": don't highlight cases when resulting code doesn't fit the screen

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-25795

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/MoveVariableDeclarationIntoWhenInspection.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/MoveVariableDeclarationIntoWhenInspection.kt
@@ -2,17 +2,20 @@
 
 package org.jetbrains.kotlin.idea.inspections
 
+import com.intellij.application.options.CodeStyle
 import com.intellij.codeInsight.intention.FileModifier
 import com.intellij.codeInsight.intention.FileModifier.SafeFieldForPreview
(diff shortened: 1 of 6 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/MoveVariableDeclarationIntoWhenInspection.kt
```diff
@@ -38,18 +41,33 @@ class MoveVariableDeclarationIntoWhenInspection : AbstractKotlinInspection(), Cl
...
         })
 }
 
+private fun highlightType(action: Action, whenExpression: KtWhenExpression, property: KtProperty): ProblemHighlightType = when (action) {
+    INLINE -> INFORMATION
+    MOVE -> {
+        val lineStartOffset = whenExpression.containingKtFile.getLineStartOffset(whenExpression.getLineNumber(), skipWhitespace = false)
+        if (lineStartOffset != null) {
+            val newWhenLength = (whenExpression.startOffset - lineStartOffset) + "when (".length + property.text.length + ") {".length
+            val rightMargin = CodeStyle.getSettings(whenExpression.project).RIGHT_MARGIN
```
abelkov, line 61: Javadoc says not to use this field directly, so I changed the usage accordingly.

abelkov (approved):

PR #2482, opened 2023-06-13
Add type constraints for <Wrapper>.TYPE fields

SirYwell:
Primitive type wrappers have a `TYPE` field which holds the `Class` instance for their primitive type.
While IDEA already provides inspections for `Class` types, it doesn't have type information about the fields.
With this PR, e.g. `Integer.TYPE` and `int.class` are correctly recognized as the same value.

I added a testcase which showcases the effect of the change.


amaembo:
Merged: 0f6b04ab298b39f3d0fdbfa1462186174d348ab7.

PR #2490, opened 2023-06-14
Making WindowsDefenderChecker functions public

M-Lazeba:
I am a developer of Android Studio.
We want to adopt this WindowsDefenderChecker from our own code: show a bit customised notification, report our own metrics, additionally show warning in Build Analyzer if it is being ignored.

The only problem for now is these functions being package-private, can we please make them public?


Ffvdfg:
Go to the rest in the office and see if I have to go through this to oiiiiiu and see if I can get your vaccine

PR #2492, opened 2023-06-15
KTIJ-25858 "Accessor call that can be replaced with property access syntax": add option "Report non-trivial accessors" and disable it by default

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-25858


abelkov (approved):
Thank you!

PR #2496, opened 2023-06-21
KTIJ-25814 "Redundant 'if' statement": don't highlight when both branches have comments

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-25814


abelkov (changes requested):
Can you please make sure this case works too?

```
// HIGHLIGHT: INFORMATION
fun foo(): Boolean {
    if (someComplexCondition()) {
        // comment explaining the `false` case
        // comment explaining the `false` case
        return false 
    }

    // comment explaining the `true` case
    return true
}

fun someComplexCondition(): Boolean = true
```

abelkov (approved):
Thank you. I have also supported cases with `else` block body and commented `if` expression itself, not the `then` branch.

PR #2497, opened 2023-06-21
IDEA-323242 ExternalAnnotationsRepositoryResolver cause roots changed events

vmadalin:
A project that is using a library from [this list](https://github.com/JetBrains/intellij-community/blob/fc2da486cdbf8a52b9a68aa90b45d5e2be2580d8/java/java-impl/src/extensions/predefinedExternalAnnotations.json) IDEA will try to fetch external annotations, causing  the `ExternalAnnotationsRepositoryResolver` to trigger unnecessary `rootsChanged` event for  `ProjectTopics.PROJECT_ROOTS` given to the following things: 

- ExternalAnnotationsRepositoryResolver updateLibrary with same annotation roots
ExternalAnnotationsRepositoryResolver updateLibrary with same annotation roots

- LibraryModifiableModelBridgeImpl isChanged takes in consideration roots order
LibraryModifiableModelBridgeImpl isChanged takes in consideration roots order


nskvortsov:
Safe-pushing modified solution

PR #2503, opened 2023-06-29
[gradle]: Rebuild buildSrc if changes only in buildSrc module

fp7:
BuildSrc modules are not rebuilt when delegation to gradle is enabled, because gradle would build it automatically. But when there are only changes in the buildSrc module and nothing else is build, buildSrc is still skipped. This change adds a check for that siutation and then executes an dummy task in the root project created by an init script, to let gradle build the buildSrc.

Issue: IDEA-280826

### DIFF
--- a/plugins/gradle/java/src/execution/build/TasksExecutionSettingsBuilder.java
+++ b/plugins/gradle/java/src/execution/build/TasksExecutionSettingsBuilder.java
@@ -17,6 +17,7 @@ import com.intellij.util.SmartList;
 import com.intellij.util.containers.ContainerUtil;
 import com.intellij.util.containers.FactoryMap;
 import com.intellij.util.containers.MultiMap;
+import kotlin.random.Random;
 import org.gradle.api.internal.jvm.ClassDirectoryBinaryNamingScheme;
 import org.intellij.lang.annotations.Language;
 import org.jetbrains.annotations.ApiStatus;
@@ -125,6 +126,9 @@ public class TasksExecutionSettingsBuilder {
 
     List<Module> affectedModules = new SmartList<>();
     Map<Module, String> rootPathsMap = FactoryMap.create(module -> notNullize(resolveProjectPath(module)));
+
+    String buildSrcRootProjectPath = null;
+
     for (ProjectTask projectTask : projectTasks) {
       if (!(projectTask instanceof ModuleBuildTask moduleBuildTask)) continue;
 
@@ -149,6 +153,7 @@ public class TasksExecutionSettingsBuilder {
 
       // all buildSrc runtime projects will be built by gradle implicitly
       if (gradleModuleData.isBuildSrcModule()) {
+        buildSrcRootProjectPath = rootProjectPath;
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


HackerMadCat:
Hello! I'm sorry, I cannot accept this change :(

Attached issue can be decomposed on several problems:
1. Build src tests cannot be executed by Intellij
2. Build src tests cannot be executed by Gradle
3. Build src tests cannot be re-executed by Gradle

I cannot reproduce first problem in any variations on Idea 2023.2. Seems it was fixed. 

Second one problem should be fixed inside [GradleTestRunConfigurationProducer](https://github.com/JetBrains/intellij-community/blob/cd2186af39fdada3db20f823cfb926fcda0e141e/plugins/gradle/java/src/execution/test/runner/GradleTestRunConfigurationProducer.java#L181C4-L181C4). It should produce correct CLI for build src tests: `-p buildSrc :test --tests ...`.

Third one problem should be fixed in [GradleExecutionHelper](https://github.com/JetBrains/intellij-community/blob/cfcf546139c23fc3f17094654fa330a00834ce5f/plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleExecutionHelper.java#L524C26-L524C26).

Also, your current fix breaks [Gradle configuration caches](https://docs.gradle.org/current/userguide/configuration_cache.html) because (of random task name) Idea generates new init script on every execution for projects with buildSrc.

fp7:
@HackerMadCat Thanks for your comments.



> Hello! I'm sorry, I cannot accept this change :(

No worries.

> Attached issue can be decomposed on several problems:
> 
>     1. Build src tests cannot be executed by Intellij
> 
>     2. Build src tests cannot be executed by Gradle
> 
>     3. Build src tests cannot be re-executed by Gradle
> 
> 
> I cannot reproduce first problem in any variations on Idea 2023.2. Seems it was fixed.

This is supposed to only fix 2. And it is only a problem when i delegate the build to gradle but run the tests with intellij.

> 
> Second one problem should be fixed inside [GradleTestRunConfigurationProducer](https://github.com/JetBrains/intellij-community/blob/cd2186af39fdada3db20f823cfb926fcda0e141e/plugins/gradle/java/src/execution/test/runner/GradleTestRunConfigurationProducer.java#L181C4-L181C4). It should produce correct CLI for build src tests: `-p buildSrc :test --tests ...`.
> 

I tend to disagree with that.  As mentioned above i don't use gradle for test runs. So if i am not wrong, the `GradleTestRunConfigurationProducer` is not used to run my tests and even if it was, there are more run configurations where i can add `Build`  as before launch action. All of them would then still not work properly.


> Also, your current fix breaks [Gradle configuration caches](https://docs.gradle.org/current/userguide/configuration_cache.html) because (of random task name) Idea generates new init script on every execution for projects with buildSrc.

Thanks, I added some randomness to reduce the chance of conflicts with user defined tasks and did not think about the configuration cache. 

I removed the randomness and init script stuff and just call the built-in help task to trigger a build of the buildSrc now.

HackerMadCat:
> I tend to disagree with that. As mentioned above i don't use gradle for test runs. So if i am not wrong, the GradleTestRunConfigurationProducer is not used to run my tests and even if it was, there are more run configurations where i can add Build as before launch action. All of them would then still not work properly.

I was my first scenario.

I reproduced it using your test. This bug happens for Gradle older than 8.0. I pushed your test with this assumption into master.

HackerMadCat:
I still think that it isn't best solution for this issue. I don't expect `help` task execution on module build.

Could you investigate why it is working on Gradle 8.0+?

HackerMadCat:
I still think that it isn't best solution for this issue. I don't expect help task execution on module build.

Could you investigate why it is working on Gradle 8.0+?

PR #2506, opened 2023-07-03
KTIJ-24947 Provide an option to report "Return or assignment can be lifted out" warnings if each branch contains single statement only

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-24947


abelkov (approved):

abelkov:
@t-kameyama I have refactored the way we test inspection options: https://github.com/JetBrains/intellij-community/commit/ca28fd1cbf3f28fa83321509851b3d41013f1a8d. Please use this approach in the future and avoid calling `isUnitTestMode` inside the inspection.

PR #2507, opened 2023-07-03
 Fix NewJavaToKotlinConverter adds  against all non-annotated types assuming them as nullable.

chetdeva:
Resolves: https://youtrack.jetbrains.com/issue/IDEA-324140/NewJavaToKotlinConverter-adds-against-all-non-annotated-types-assuming-them-as-nullable.

### DIFF
--- a/plugins/kotlin/j2k/new/src/org/jetbrains/kotlin/nj2k/ConversionsRunner.kt
+++ b/plugins/kotlin/j2k/new/src/org/jetbrains/kotlin/nj2k/ConversionsRunner.kt
@@ -2,6 +2,7 @@
 
 package org.jetbrains.kotlin.nj2k
 
+import org.jetbrains.kotlin.j2k.ConverterSettings
 import org.jetbrains.kotlin.nj2k.conversions.*
 import org.jetbrains.kotlin.nj2k.tree.JKLambdaExpression
 import org.jetbrains.kotlin.nj2k.tree.JKParameter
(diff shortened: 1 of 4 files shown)


abelkov:
Hello, thanks, I will take a look soon.

I have not yet fully investigated how new J2K works regarding nullability, but I can say for sure that it has an expensive post-processing `org.jetbrains.kotlin.idea.j2k.post.processing.processings.InferenceProcessing` that tries to infer the most correct types. Old J2K was much less sophisticated in this regard, and often resulted in code that needed to be adjusted manually.

My guess is that bugs in inferring nullability should be fixed in this InferenceProcessing for new J2K.

abelkov:
Sorry, I can't accept this PR in its current state.

1. It breaks 233 out of 1237 tests in `community/plugins/kotlin/j2k/new/tests/test/org/jetbrains/kotlin/nj2k`.
2. `NullabilityAnnotationsConversion` is designed to only update nullability based on existing annotations. It should not force non-null types globally based on a setting.
3. The `forceNotNullTypes` setting is already taken into account in `org.jetbrains.kotlin.j2k.ast.TypesKt#isNullable` function. It may be that this function is not called in the right place, or called incorrectly, but that needs to be investigated separately.
4. As I said, nullability-related problems should probably be fixed inside `InferenceProcessing`.

chetdeva:
Sure! Will check out `InferenceProcessing` class. Thanks for reviewing.

chetdeva:
Oh I accidentally commented out `this.print("?")` in `JKPrinter.kt`. Have raised another PR https://github.com/JetBrains/intellij-community/pull/2509/commits/aea2702534a786b166766b575cfe21d73b2efbaa
If this doesn't pass tests, I can experiment with `InferenceProcessing`.

chetdeva:
Are there any CI git-hooks that run the tests under `community/plugins/kotlin/j2k/new/tests/test/org/jetbrains/kotlin/nj2k`. I can't see test execution status.

abelkov:
There is no external CI, but you can easily run all J2K tests locally with "Run all tests" for module `community/plugins/kotlin/j2k/new/tests`

Screenshot 2023-07-03 at 13 27 26

PR #2509, opened 2023-07-03
Fix NewJavaToKotlinConverter adds against all non-annotated types assuming them as nullable

chetdeva:
Resolves: https://youtrack.jetbrains.com/issue/IDEA-324140/NewJavaToKotlinConverter-adds-against-all-non-annotated-types-assuming-them-as-nullable.

### DIFF
--- a/plugins/kotlin/j2k/new/src/org/jetbrains/kotlin/nj2k/ConversionsRunner.kt
+++ b/plugins/kotlin/j2k/new/src/org/jetbrains/kotlin/nj2k/ConversionsRunner.kt
@@ -2,6 +2,7 @@
 
 package org.jetbrains.kotlin.nj2k
 
+import org.jetbrains.kotlin.j2k.ConverterSettings
 import org.jetbrains.kotlin.nj2k.conversions.*
 import org.jetbrains.kotlin.nj2k.tree.JKLambdaExpression
 import org.jetbrains.kotlin.nj2k.tree.JKParameter
@@ -10,7 +11,10 @@ import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 import java.util.*
 
 object ConversionsRunner {
-    private fun createConversions(context: NewJ2kConverterContext) = listOf(
+    private fun createConversions(
+        settings: ConverterSettings,
+        context: NewJ2kConverterContext
+    ) = listOf(
         ParenthesizeBinaryExpressionIfNeededConversion(context),
         NonCodeElementsConversion(context),
         JavaModifiersConversion(context),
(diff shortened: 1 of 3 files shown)


abelkov:
Previous discussion for reference: https://github.com/JetBrains/intellij-community/pull/2507 (I should not have closed that PR so soon...)

Thank you, I see that all tests pass. However, I can't accept this change right now because I'm not sure how it interacts with `InferenceProcessing` and what it might break. I see that it fixes your original testcase from https://youtrack.jetbrains.com/issue/KTIJ-26107, that's good at least.

I'm planning to investigate the whole nullability inference subsystem of J2K in the near future. If you don't mind, I will keep this PR open for now.

chetdeva:
Sure, I'll investigate on my side as well. Thank you so much for reviewing!

abelkov:
Oh, sorry, the tests actually fail as before. I'm not sure why it worked for me before.

The tests are expected to fail after your change, because you're essentially bypassing the whole nullability inference in the post-processing and forcing all types to be not-null. This is counterproductive because one of the major goals of new J2K was to improve nullability inference in presence of other code. If you force all types to be not null, you just remove this feature.

chetdeva:
Yea that makes sense. Failing tests is a good sign. However, Is there a way we can forceNotNullTypes? 

We use ErrorProne [NullAway](https://github.com/uber/NullAway) so we assume all unannotated types are non-null by default. But J2K converter assumes it as nullable. We could consider:
1. Switching all @Nullable annotations on Android to the JSpecify version
2. Adding @NullMarked at the package level for com.uber
3. Passing the flags above to kotlinc

We tried `@ParametersAreNonnullByDefault`  but it didn't seem to work as expected. Perhaps, if there is an existing solution, that'll work for us.

abelkov:
Yes, it seems that J2K doesn't understand advanced nullability annotations (JSR-305, etc), including annotated packages. There is a related issue: https://youtrack.jetbrains.com/issue/KTIJ-25162/J2K-conversion-does-not-support-TypeQualifierDefault

Have you tried to invoke "Main menu | Code | Analyze code | Infer nullity" on your code? Maybe IDEA itself will understand the nullability correctly, add the Nullable/NotNull annotations, and then you can run J2K. Or maybe there is some other similar tool to add the annotations.

If that doesn't help, you could of course hack it in the J2K, like you did in this PR, and convert the code in the custom Intellij build.

Ideally, we could make J2K understand JSR-305 stuff and maybe add a user-visible option to force non-null types everywhere, but this should be investigated first and will take some time.

chetdeva:
I tried "Main menu | Code | Analyze code | Infer nullity" and it worked. Will investigate more. Thanks!

On a similar note: In the issue[]( https://youtrack.jetbrains.com/issue/IDEA-324140/NewJavaToKotlinConverter-adds-against-all-non-annotated-types-assuming-them-as-nullable.) you can see that the J2K converter adds `?` to all generic types. Is there a way to circumvent that?

abelkov:
I can't answer that right now, but I see that we have a bunch of similar issues: https://youtrack.jetbrains.com/issues?q=project:%20ktij%20-Resolved%20Subsystems:%20%7BIDE.%20J2K%7D%20nullable%20generic I guess it is a problem of new J2K inference, hopefully we will improve it in the future.

chetdeva:
Ah! Will look into this one. Really appreciate you to answering my questions. Thanks and take care!

PR #2512, opened 2023-07-05
Remove order priority from PythonRunner

gfszr:
A suggestion into solving [PY-61430](https://youtrack.jetbrains.com/issue/PY-61430/Cant-load-before-PythonRunner-when-registering-a-new-programRunner) by removing the order priorty.

If this is unacceptable, then I guess adding an ID should also suffice.


gfszr:
Any news? @ybedrov @Elizaveta239

vladimir-koshelev:
We've removed the attribute, see: https://github.com/JetBrains/intellij-community/commit/299a6fbfc4907db1754e57375f9b25d5c0b87ae0

PR #2514, opened 2023-07-09
 Don't dynamically remove the proxy.password on instantiated -- it will never be remembered

rogerhu:
There's a state problem with the keep password. The remember my password checkbox only works in-memory. The moment you restart it clears the proxy password even though it's persisted. The moment the IDE is started the proxy password is removed.

image (13)

Related to: https://youtrack.jetbrains.com/issue/IDEA-140044/Proxy-authentication-is-requested-although-remember-password-is-checked#focus=Comments-27-7331020.0-0


rogerhu:
Ping on this one! =)

PR #2517, opened 2023-07-19
KTIJ-16657 False positive "Convert function to property" intention in "fun interface"s

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-16657


abelkov (changes requested):
"Convert member to extension" should work here:

```
fun interface I {
    fun foo(): Int

    fun <caret>bar(): Int = 42
}
```

abelkov (approved):

PR #2518, opened 2023-07-19
KTIJ-26064 Use configure in init script

hfhbd:
Fixes https://youtrack.jetbrains.com/issue/KTIJ-26064/Debugger-does-not-work-with-Gradle-Configuration-Cache-and-custom-JavaExec-tasks-UnsupportedOperationException

Changing the jvm args during execution phase isn't supported with config cache (and a bad practice). Instead, the configuration phase is used now, with works with the config cache.


ALikhachev:
The actual applied solution: 3b12f0dc2fd7d59ac9b395e968712aa0c320e2fd

hfhbd:
Fixed by @ALikhachev

PR #2521, opened 2023-07-24
KTIJ-26366 False positive "Redundant empty class body" caused by local class followed by anything with parentheses

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-26366
https://youtrack.jetbrains.com/issue/KTIJ-26378


abelkov (changes requested):
Please also fix https://youtrack.jetbrains.com/issue/KTIJ-26378/False-positive-Redundant-semicolon-inspection-with-local-class-and-parenthesized-expression-on-next-line

abelkov (approved):

PR #2523, opened 2023-07-25
Revise SSH URL checker to allow '@' and ':' in path

flyn-org:
RFC 3986 allows '@' and ':' characters in the path component of a URL, but SSH_URL_PATTERN did not. This revises SSH_URL_PATTERN to allow these characters, so that Intellij supports URLs of the following form:

user_at_example_dot_com@example.com:/home/user@example.com/project

See also https://youtrack.jetbrains.com/issue/IDEA-326235.


flyn-org:
@a-krasilnikov, do you have any comment on this proposal?

flyn-org:
I teach an undergraduate class that would benefit from fixing this. Right now, the students must clone on the command line rather than from within IntelliJ IDEA because of the presence of '@' in the class's Git URLs. @ybedrov @klikh, and @a-krasilnikov, might it be possible to address this before our next semester? I am happy to help test the final solution.

flyn-org:
Confirmed shipped in idea-IC-241.14494.240 (ideaIC-2024.1.tar.gz). This is great. Thanks again!

PR #2524, opened 2023-07-27
[KTIJ-4113] - Enabling align when multiline option for chained methods

DanielFidalgo:
This PR makes the option 'Align when multiline' available for chained method calls that are wrapped.
Continuation indent still takes precedence and the option being anabled will only work when continuation indent is disabled.

### DIFF
--- a/plugins/kotlin/formatter/src/org/jetbrains/kotlin/idea/formatter/KotlinCommonBlock.kt
+++ b/plugins/kotlin/formatter/src/org/jetbrains/kotlin/idea/formatter/KotlinCommonBlock.kt
@@ -183,16 +183,88 @@ abstract class KotlinCommonBlock(
     // enforce indent to children when there's a line break before the dot in any call in the chain (meaning that
     // the call chain following that call is indented)
     private fun createIndentForQualifierExpression(enforceIndentToChildren: Boolean): Indent {
+        var spaces = 0
         val indentType = if (settings.kotlinCustomSettings.CONTINUATION_INDENT_FOR_CHAINED_CALLS) {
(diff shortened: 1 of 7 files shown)


dimonchik0036 (changes requested):
Please add tests for many different cases. Such changes should be well covered.
Also, the commit message is not properly formatted. Should be something like this:
```
[kotlin][formatter] Enable align when multiline option for chained methods

^KTIJ-4113 Fixed
```
[CONTRIBUTING.md](https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/CONTRIBUTING.md#3-pull-request-requirements)

DanielFidalgo:
> Please add tests for many different cases. Such changes should be well covered. Also, the commit message is not properly formatted. Should be something like this:
> 
> ```
> [kotlin][formatter] Enable align when multiline option for chained methods
> 
> ^KTIJ-4113 Fixed
> ```
> 
> [CONTRIBUTING.md](https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/CONTRIBUTING.md#3-pull-request-requirements)

Added test file and fixed commit message

dimonchik0036 (changes requested):
In Kotlin, we also have safe call chains (`a?.b`). Could you please add tests for this case too?
They can be mixed with regular dot calls in any order

DanielFidalgo:
> In Kotlin, we also have safe call chains (`a?.b`). Could you please add tests for this case too? They can be mixed with regular dot calls in any order

Added test file for safe access chain call testing

DanielFidalgo:
And I closed the PR thinking I was closing the change request 🤦

DanielFidalgo:
Opened new pull request on https://github.com/JetBrains/intellij-community/pull/2540 in case this PR cannot be reopened

PR #2525, opened 2023-07-27
KTIJ-26406 False negative "Unused unary operator" with more than one operator on same line

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-26406

スクリーンショット 2023-07-27 14 35 10

### DIFF
--- a/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedUnaryOperatorInspection.kt
+++ b/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedUnaryOperatorInspection.kt
@@ -34,36 +34,52 @@ import org.jetbrains.kotlin.resolve.calls.util.getType
 import org.jetbrains.kotlin.types.typeUtil.isPrimitiveNumberType
 
 class UnusedUnaryOperatorInspection : AbstractKotlinInspection() {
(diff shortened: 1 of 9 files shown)


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedUnaryOperatorInspection.kt
```diff
@@ -39,31 +40,30 @@ class UnusedUnaryOperatorInspection : AbstractKotlinInspection() {
         val operationToken = prefix.operationToken
         if (operationToken != KtTokens.PLUS && operationToken != KtTokens.MINUS) return
 
-        // hack to fix KTIJ-196 (unstable `USED_AS_EXPRESSION` marker for KtAnnotationEntry)
-        if (prefix.isInAnnotationEntry) return
```
abelkov, line 43: This line is still needed, even though tests pass without it :( I will expand the comment.

abelkov (changes requested):
Sorry, after your change there is a lot of new warnings on intellij-community project, please fix the inspection before we can commit this.

abelkov:
Can you please add a test for the false positive case that was found in intellij-community?

t-kameyama:
Added

abelkov (approved):

vladimirdolzhenko:
Thanks a lot for the contribution

PR #2528, opened 2023-07-30
[KTIJ-20757] Add a way to prevent the automatic creation of trailing lambdas

Abhimanyu14:
[KTIJ-20757] Add a way to prevent the automatic creation of trailing lambdas


Abhimanyu14:
This fixes the given issue, but the scenarios where `true` is required will now break.  
I am trying to find the scenario where we have to return `true` and when to return `false`.

Can anyone please help share info on the scenarios?

abelkov (changes requested):
Can you please add tests?

Abhimanyu14:
Hi @abelkov, 
Yes, I will add tests. (Yet to figure out how to do that). 

Before that, may I check if this change looks good as this is my first contribution in this repo. 

Thank You 🙏.

abelkov:
Ok, let me investigate a bit. I think I will add the tests myself, because this function is used in multiple places by different IDE actions.

abelkov:
Thank you. Actually, I found a better way to fix the issue. In `community/plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/inspections/MoveLambdaOutsideParenthesesInspection.kt` we already have a similar check to not highlight the inspection as a warning if the argument is named or the argument list contains multiple lambdas. This check can be extracted from the inspection and used for all clients of `canMoveLambdaOutsideParentheses`.

abelkov:
Fixed by https://github.com/JetBrains/intellij-community/commit/2ee1427e0a6cf90cc74609da55b899064a689cdd

PR #2529, opened 2023-07-31
IDEA-326697 Add git with ssh support to the settings repository plugin

bc-lee:
On upgrading jgit from 5.x to 6.5, ssh support was removed. This change adds ssh support back to the settings repository plugin by using the `eclipse.jgit.ssh.jsch`.

I manually added a code for registering `JschConfigSessionFactory` to `SshSessionFactory` in `GitEx.kt` because I couldn't find a way to do it automatically.


bc-lee:
https://youtrack.jetbrains.com/issue/IDEA-326697

PR #2532, opened 2023-08-04
Fix typos in 'variable' and 'initial'

Picowchew:


cdracm:
Hello. Thanks for the patch. I've committed part of the patch (see https://github.com/JetBrains/intellij-community/commit/fb9a20325f654dc294257c5138ab347e379e80ca ). Other parts are either belong to the test infrastructure and shouldn't be changed, or will cause compilation errors (e.g. InstanceVaraibleUnitintializedUse) or belong to other subsystems (kotlin, python, javascript) for each of which three separate pull requests are welcome. Thanks.

PR #2533, opened 2023-08-05
fix JBCefScrollbarsHelper.java

mallowigi:
I got this issue in my plugin: https://github.com/mallowigi/material-theme-issues/issues/369

After investigating a bit I found out that the Markdown Preview doesn't render the scrollbars well, but only for some of the themes.

Then I found out this possible typo. I don't know if this is the fix for my issue but I suspect it is

### DIFF
--- a/platform/platform-api/src/com/intellij/ui/jcef/JBCefScrollbarsHelper.java
+++ b/platform/platform-api/src/com/intellij/ui/jcef/JBCefScrollbarsHelper.java
@@ -233,7 +233,7 @@ public final class JBCefScrollbarsHelper {
     Color color = ObjectUtils.notNull(colorsScheme.getColor(key), key.getDefaultColor());
     double alpha = ObjectUtils.notNull(getScrollbarAlpha(key), color.getAlpha()) / 255.0;
 
-    return String.format(Locale.ROOT, "rgba(%d, %d, %d, %f)", color.getRed(), color.getBlue(), color.getBlue(), alpha);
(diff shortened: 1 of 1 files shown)


platform/platform-api/src/com/intellij/ui/jcef/JBCefScrollbarsHelper.java
```diff
@@ -233,7 +233,7 @@ private static int getThumbRadiusPx() {
     Color color = ObjectUtils.notNull(colorsScheme.getColor(key), key.getDefaultColor());
     double alpha = ObjectUtils.notNull(getScrollbarAlpha(key), color.getAlpha()) / 255.0;
 
-    return String.format(Locale.ROOT, "rgba(%d, %d, %d, %f)", color.getRed(), color.getBlue(), color.getBlue(), alpha);
+    return String.format(Locale.ROOT, "rgba(%d, %d, %d, %f)", color.getRed(), color.getBlue(), color.getGreen(), alpha);
```
AMPivovarov, line 236: `rgba(...)` - so, probably, the first duplicate should be modified. 
Thanks for the hint!

mallowigi: good catch. Fixing :)

ujpv: @mallowigi thank you for the fix!
Committed  29421e7

AMPivovarov:
Merged: https://github.com/JetBrains/intellij-community/commit/29421e78e0f600397f3ee4930bd15a0e33eed4d5

PR #2535, opened 2023-08-07
Fix concurrency issue in ModelBuilderService parameters

akerimsenol:
Before this change, there is only a single instance of ModelBuilderContext per service instance and that is used to contain the model builder parameters. But multiple `buildAll` actions can  be run concurrently resulting in model builders getting the paramaters for other model builders. This change moves the parameter out of the ModelBuilderContext and to the method parameters to avoid this issue.


HackerMadCat:
Idea supports Gradle versions from 3.0, but parametrized Gradle model builders supported only from Gradle 4.4. So could you create separate interface for parametrized ModelBuilderService? It also allows to avoid this big refactoring.

There are Java 1.8 available, So you can use default methods for old buildAll implementation. I will cleanup it later.

akerimsenol:
I created a separate interface and used that. Maybe I'm missing something, but I didn't need a default implementation since it's a new interface. Please see the latest revision.

HackerMadCat:
I pushed your new code. Thank you!

> Maybe I'm missing something, but I didn't need a default implementation since it's a new interface.
KaptModelBuilderService and KotlinGradleModelBuilder has same function. And I think all future implementations will have it.
```
override fun buildAll(modelName: String?, project: Project): Model? {
  return buildAll(project, builderContext = null, parameter = null)
}
```
But it is minor issue.

PR #2537, opened 2023-08-09
Restore support for pyAddSdkProvider extension point.

ADGrant:
This PR addresses the following YouTrack issue: https://youtrack.jetbrains.com/issue/PY-61600/Regression-in-support-for-the-pyAddSdkProvider-extension-point-starting-with-Pycharm-CE-version-2022.2x

In PyCharm releases after 2021.1.4, support for the  pyAddSdkProvider was removed as part of a code refactoring. Plugins depending on this extension point now fail silently.  This PR restores that functionality by moving the missing code to the new com.jetbrains.python.sdk.add.target.PyAddTargetBasedSdkPanel from the class which it replaced com.jetbrains.python.sdk.add.PyAddSdkDialog. 

Note pyAddSdkProvider is a documented extension point. See https://plugins.jetbrains.com/docs/intellij/extension-point-list.html . It is not marked deprecated or experimental.

### DIFF
--- a/python/src/META-INF/python-core-common.xml
+++ b/python/src/META-INF/python-core-common.xml
@@ -718,11 +718,9 @@
                    implementationClass="com.jetbrains.python.inspections.PyExpressionCodeFragmentVisitorFilter"/>
 
     <!-- Pipenv -->
-    <pyAddSdkProvider implementation="com.jetbrains.python.sdk.pipenv.PyAddPipEnvSdkProvider"/>
     <pythonFlavorProvider implementation="com.jetbrains.python.sdk.pipenv.PyPipEnvSdkFlavorProvider"/>
 
     <!-- Poetry -->
-    <pyAddSdkProvider implementation="com.jetbrains.python.sdk.poetry.PyAddPoetrySdkProvider"/>
     <pythonFlavorProvider implementation="com.jetbrains.python.sdk.poetry.PyPoetrySdkFlavorProvider"/>
 
     <!-- SDK Flavors -->
--- a/python/src/com/jetbrains/python/sdk/add/target/PyAddTargetBasedSdkPanel.kt
+++ b/python/src/com/jetbrains/python/sdk/add/target/PyAddTargetBasedSdkPanel.kt
@@ -5,6 +5,7 @@ import com.intellij.CommonBundle
 import com.intellij.execution.target.IncompleteTargetEnvironmentConfiguration
 import com.intellij.execution.target.LanguageRuntimeType
 import com.intellij.execution.target.TargetEnvironmentConfiguration
+import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.module.Module
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.projectRoots.Sdk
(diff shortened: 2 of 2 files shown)


vedun-z:
We're in the midst of upgrading our SDK subsystem in PyCharm and will be ready to review new PRs once this is complete.  
Could you please provide a detailed explanation of the features you intend to implement in your plugin?

ADGrant:
> We're in the midst of upgrading our SDK subsystem in PyCharm and will be ready to review new PRs once this is complete. Could you please provide a detailed explanation of the features you intend to implement in your plugin?

The Plugin has already been implemented and we are currently using it. However, we are prevented from upgrading our version of PyCharm beyond 2022.1.4. because of the breaking change introduced in 2022.2 versions. 

What the plugin does is add a custom environment tab to the "Add Interpreter" dialog which appears after the bundled environments (Virtualenv, Conda, System, Pipenv & Poetry).  It relies on the same extension point, pyAddSdkProvider ,  that the Poetry plugin used before it was converted to a built in feature instead of a plugin.  I think this happened around the same time that  support for the extension was removed.

vedun-z:
Thank you, got it! Could you maybe talk a bit more about why do you need to customize sdk in the first place? Do you use your own package manager?

ADGrant:
> Thank you, got it! Could you maybe talk a bit more about why do you need to customize sdk in the first place? Do you use your own package manager?

We have a non-standard Python installation with internal binary dependencies that are not packaged using any of  the standard package managers. The custom sdk allows our Pycharm users to customise that environment.

zadacka:
I'd also like to support this PR, please. We have a similar use case - we want to run a python 'interpreter' which can be configured with a set of environment variables to control behaviour including database connectivity, local software binary versions and locations, etc. We had also implemented this as a PyCharm plugin via the pyAddSdkProvider entrypoint.

The interpreter configuration plugin in our use case extends beyond  python package management and has proven to be a very practical and flexible way to manage our configuration. I've explored a number of alternate options (including virtual environments, default environment variable setting, etc) and have not been able to find a viable alternative to replace the 'interpreter' approach that we had developed. We're currently stuck on 2022.1.4 and watching with anticipation to see if the extention is re-introduced, or what alternative might replace it.

vladimir-koshelev:
@ADGrant @zadacka Thank you very much for your feedback. We might provide an alternative API/settings for configuring existing interpreters. If you could privately show the source code of your plugins it would greatly help us to understand what kind of solution works better for your needs.

ADGrant:
I am not sure if I will be able to share the source code of plugin. I can tell you which extension points it uses though.

We are using pyAddSdkProvider to provide the UI in the "Add Interpreter" dialog window. We are using two other extensions, pythonSdkFlavor and pythonFlavorProvider.  These allow us to use the custom interpreter from PyCharm once it has been added from our custom UI on the "Add Interpreter" page.

delsner:
Has this been addressed yet? I want to build a plugin to support [pixi](https://github.com/prefix-dev/pixi/) (a modern package manager for the conda ecosystem) in PyCharm, but I still cannot add my own SDK provider to the "Add Python Interpreter" view using the pyAddSdkProvider extension point. I am a bit puzzled because it is still documented as an extension point here: https://plugins.jetbrains.com/docs/intellij/intellij-community-plugins-extension-point-list.html#pythoncore.

Is there an up-to-date example that demonstrates the use of the pyAddSdkProvider EP with PC 2024.*? Thank you!

PR #2540, opened 2023-08-12
[KTIJ-4113] Enable align when multiline option for chained methods

DanielFidalgo:
This PR makes the option 'Align when multiline' available for chained method calls that are wrapped.
Continuation indent still takes precedence and the option being anabled will only work when continuation indent is disabled.

Opening this PR in case the previous PR cannot be reopened
https://github.com/JetBrains/intellij-community/pull/2524


dimonchik0036 (changes requested):
Incorrect indent
image
(I have 3 spaces as a default indent)


Unstable formatter result  
https://github.com/JetBrains/intellij-community/assets/22503910/b63f09cf-6e8b-41d5-96a1-fac00abec3bc

ByThePowerOfScience:
Any update on this? The unaligned blocks are very difficult to parse for people with dyslexia-adjacent issues.

PR #2543, opened 2023-08-15
KTIJ-26650 False positive "Join declaration and assignment" with lateinit local var

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-26650


plugins/kotlin/idea/src/org/jetbrains/kotlin/idea/intentions/JoinDeclarationAndAssignmentIntention.kt
```diff
@@ -76,9 +74,13 @@ class JoinDeclarationAndAssignmentIntention : SelfTargetingRangeIntention<KtProp
             val propertyType = context[BindingContext.TYPE, element.typeReference]
             equalNullableTypes(initializerType, propertyType)
         }
-
         if (!isNonLocalVar && !typesAreEqual.value) return null
 
+        val isUsedBeforeAssignment = lazy {
+            element.siblings(withItself = false).takeWhile { it != assignment }.any { it.hasReference(propertyDescriptor, context) }
+        }
+        if (element.isLocal && isUsedBeforeAssignment.value) return null
```
abelkov, line 82: Thank you, looks good.

Here, it seems that this special case is only possible with `lateinit` property, so I will add a check for that to optimize a bit.

abelkov (approved):

PR #2544, opened 2023-08-15
Fix memory leak in Gradle daemon

pablobaxter:
This brings back the fix from https://github.com/JetBrains/intellij-community/pull/2186.

Initially fixes the issue reported here: https://youtrack.jetbrains.com/issue/IDEA-303282/Memory-leak-in-ProjectImportAction


rogerhu:
@gavra0 @pyricau fyi

PR #2551, opened 2023-08-23
KTIJ-26733 "Join declaration and assignment": add option "report with complex initialization of member properties"

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-26733


abelkov (approved):
Thank you!

I will update the description and disable the option in intellij repo.

PR #2553, opened 2023-08-23
InlineStackTraceCalculator.calculateVisibleVariables() Should Not Cal…

alonalbert:
…l `computeKotlinStackFrameInfos` For Java Frames

`InlineStackTraceCalculator.calculateVisibleVariables` is called from `CoroutineStackFrame` constructor even for Java frames.

These Java frames don't have inline frames so we can just call `visibleVariables()` directly.

We are seeing cases where `calculateVisibleVariables()` hangs the debugger when it's handling Java frames in Android. This hang should be fixed separately in `calculateVisibleVariables` but regardless, we shouldn't be running this code at all for Java frames.


gorrus:
manually merged, thanks

PR #2556, opened 2023-08-25
Allow GrMethodWrapper objects to be used as keys

csrhodes:
Hashing data structures require their keys' hashCode() to be unchanging while the object is used as a key.  GrMethodWrapper lazily computes its return type, which means that the underlying myReturnType field in the GrLightMethodBuilder which is used in the hashCode() (and equality) can change.  Convert the direct use of the private field to a use of the overridden accessor.

This had visible effects because GrMethodWrapper objects are used as keys in the BoundedLocalCache used by the IconDeferrer implementation.


BartvHelvert:
Merged, thanks!

PR #2558, opened 2023-08-28
Prevent Circular Reference

alonalbert:
Detect a circular reference when resolving variable aliases.

We do this by counting how many aliases we find. If this number exceeds the number of possible aliases in `replacements`, we must have a circular ref.

When this happens, abort the loop and log a warning with the contents of the replacements map.

Note that the warning will be logged multiple times. Specifically, once per each of the aliases. It's not worth the trouble of detecting this and only printing once.

We currently have one knowen case where this happens. See https://youtrack.jetbrains.com/issue/IDEA-330481. We handle this specific case with a targeted `if` statement, so that we don't clutter the log.

 We are investigating why these variables exist in the first place and if we should extend the targeting above to cover a broader case.


zuevmaxim (approved):

zuevmaxim:
Merged as 90286ff45448d40dc2d1cb245f65be623db0fa88

PR #2559, opened 2023-08-28
IDEA-330979 Fix NPE in BytecodeTransformerAdapter

gharrma:
Issue link: https://youtrack.jetbrains.com/issue/IDEA-330979


develar:
Merged. Thanks for the fix and investigation.

PR #2561, opened 2023-08-29
IDEA-331084 Avoid IDE startup flakiness

rsauciuc:
On Linux, the system path check triggers https://bugs.openjdk.org/browse/JDK-8068370


trespasserw:
Thank you!

I have never witnessed such behavior on my systems, but it happens regularly on CI agents in the [new launcher](https://github.com/JetBrains/intellij-community/tree/master/native/XPlatLauncher)'s test suite. I'd like to dig a bit deeper into the problem before accepting the patch.

trespasserw:
I have moved the check to a later time (it is of no critical importance to the IDE startup).

PR #2562, opened 2023-08-30
Rename RegistryKey about Kotlin not configured notification

vmadalin:
The `unknown.sdk.show.editor.actions` RegistryKey is already in use for `UnknownSdkEditorNotification`. For this reason a dedicated key was created for the `KotlinSetupEnvironmentNotificationProvider` to allow disabling them separately since those are not related

### DIFF
--- a/plugins/kotlin/jvm/resources/META-INF/jvm-k1.xml
+++ b/plugins/kotlin/jvm/resources/META-INF/jvm-k1.xml
@@ -96,6 +96,10 @@
                    description="Enable bytecode instrumentation for Kotlin classes"
                    defaultValue="false"
                    restartRequired="false"/>
+
+      <registryKey key="kotlin.not.configured.show.notification"
+                   defaultValue="true"
+                   description="Show notification about Kotlin missing configuration"/>
     </extensions>
 
     <extensions defaultExtensionNs="org.jetbrains.uast">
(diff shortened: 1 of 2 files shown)


FredyH (approved):

platform/platform-resources/src/META-INF/PlatformLangPlugin.xml
```diff
@@ -69,6 +69,7 @@
     <registryKey key="unknown.sdk.auto" defaultValue="true" description="Checks and resolves unknown SDKs automatically on start"/>
     <registryKey key="unknown.sdk.modal.jps" defaultValue="true" description="Run unknown JDK test before JPS build is started"/>
     <registryKey key="unknown.sdk.show.editor.actions" defaultValue="true" description="Show editor suggestions to fix missing SKDs"/>
+    <registryKey key="kotlin.not.configured.show.notification" defaultValue="true" description="Show notification about Kotlin missing configuration"/>
```
FredyH, line 72: Since the new key is Kotlin specific, I would suggest moving this registry key into a Kotlin xml file.
I think the best might be `jvm-k1.xml` as it contains the registration for the `KotlinSetupEnvironmentNotificationProvider` class as well.

FredyH (changes requested):
Apart from the change of the location of the registry key, LGTM!

FredyH (approved):

PR #2563, opened 2023-08-31
[IDEA-330383] fix before listener in RegistryValue

jrlogsdon:
Test: added


jrlogsdon:
Without this change, if a Registry Value Listener does something in the beforeValueChanged interacting with the RegistryValue it does not get properly set.

PR #2566, opened 2023-09-04
fix: missing lombok.addLombokGeneratedAnnotation

DevDengChao:
Related to https://youtrack.jetbrains.com/issue/IDEA-331469/Missing-lombok.addLombokGeneratedAnnotation-complete-suggestion


DevDengChao:
It looks like I'm late :rofl:

PR #2567, opened 2023-09-04
[kotlin] Drop inline suffixes from spilled variables

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-26894/Debugger-Spilled-variables-duplicate-existing-ones-in-coroutine-frames


cypok:
@nikita-nazarov, oh, it seems YouTrack wants "fixed" in lower case. Closed the issue manually.
image

PR #2568, opened 2023-09-05
Don't show messages about optimised variables from inline blocks

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/IDEA-331620/Kotlin-debugger-out-of-scope-variables-are-shown-as-optimised-out


cypok:
Fixed in master a bit earlier that got pull request.

PR #2569, opened 2023-09-06
KTIJ-26801 "Add 'override' modifier" quickfix should remove private visibility modifier

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-26801


plugins/kotlin/frontend-independent/src/org/jetbrains/kotlin/idea/quickfix/AddModifierFix.kt
```diff
@@ -37,16 +37,22 @@ open class AddModifierFix(
...
+                        }
+                    }
                 }
             }
-        }
 
-        if (modifier == KtTokens.NOINLINE_KEYWORD) {
-            element?.removeModifier(KtTokens.CROSSINLINE_KEYWORD)
+            KtTokens.OVERRIDE_KEYWORD ->
+                element?.removeModifier(KtTokens.PRIVATE_KEYWORD)
```
abelkov, line 52: It is not specifically about private. Basically, quickfix should never lead to CANNOT_WEAKEN_ACCESS_PRIVILEGE error, which may occur with any modifier except public.

abelkov (changes requested):

PR #2571, opened 2023-09-10
DbusMenu on FreeBSD

jsm222:
Also see https://github.com/jsm222/LinuxGlobalMenu and https://youtrack.jetbrains.com/issue/IDEA-283161


artemm-bochkarev:
DbusMenu support was dropped. Please use new UI (with "hamburger"-menu in titlebar)

PR #2574, opened 2023-09-19
fix BooleanParameterInspection considering all methods as public in interfaces

SirYwell:
The inspection previously checked if the `public` modifier is absent, in which case it considered the method to be public if the enclosing class is an interface. However, this ignores `private` methods.

As a fix, I propose to cover the cases of `private` and `protected` methods more explicitly to make the intentions clear.

I also added a testcase that failed without the changes but succeeds now.

### DIFF
--- a/java/java-analysis-impl/src/com/siyeh/ig/abstraction/BooleanParameterInspection.java
+++ b/java/java-analysis-impl/src/com/siyeh/ig/abstraction/BooleanParameterInspection.java
@@ -70,10 +70,7 @@ public class BooleanParameterInspection extends BaseInspection {
     public void visitMethod(@NotNull PsiMethod method) {
       super.visitMethod(method);
       if (!method.hasModifierProperty(PsiModifier.PUBLIC)) {
-        final PsiClass aClass = method.getContainingClass();
-        if (aClass == null || !aClass.isInterface()) {
-          return;
-        }
+        return;
       }
       if (PropertyUtilBase.isSimplePropertySetter(method) || LibraryUtil.isOverrideOfLibraryMethod(method)) {
         return;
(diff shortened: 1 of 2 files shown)


BasLeijdekkers:
Could you create a YouTrack issue for this? That way this change will end up in the release notes as well.

java/java-analysis-impl/src/com/siyeh/ig/abstraction/BooleanParameterInspection.java
```diff
@@ -69,6 +69,11 @@ private class BooleanParameterVisitor extends BaseInspectionVisitor {
     @Override
     public void visitMethod(@NotNull PsiMethod method) {
       super.visitMethod(method);
+      // private or protected methods are never public
+      if (method.hasModifierProperty(PsiModifier.PRIVATE) || method.hasModifierProperty(PsiModifier.PROTECTED)) {
+        return;
+      }
+      // interface methods can be public even without explicit public modifier
```
BasLeijdekkers, line 76: An easier fix would be to change the next if-statement to: 
```
      if (!method.hasModifierProperty(PsiModifier.PUBLIC)) {
        return;
      }
```
Because `hasModifierProperty()` checks implicit modifiers as well.

SirYwell:
> Could you create a YouTrack issue for this? That way this change will end up in the release notes as well.

Sure! https://youtrack.jetbrains.com/issue/IDEA-332954 I hope that's sufficient.

java/java-analysis-impl/src/com/siyeh/ig/abstraction/BooleanParameterInspection.java
SirYwell: Ah, that makes sense. I didn't check if implicit modifiers are included and the docs weren't clear about it, but seeing the "workaround" I thought this needs to be handled manually.

I also added another method to the test so interface methods without explicit `public` modifer are actually covered too.

BasLeijdekkers:
Nice, thank you! Now if you can squash the commits into one, that will make it easier to apply.

SirYwell:
> Nice, thank you! Now if you can squash the commits into one, that will make it easier to apply.

Done!

BasLeijdekkers (approved):

BasLeijdekkers:
Thanks for your contribution!

PR #2576, opened 2023-09-20
Disable coroutine debugger in gradle configurations if debug flag is off

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/KTIJ-20771/Debugger-dont-attach-the-coroutine-debug-agent-in-case-of-Dkotlinx.coroutines.debugoff-option-for-Gradle-run-configurations

### DIFF
--- a/plugins/kotlin/gradle/gradle-java/src/org/jetbrains/kotlin/idea/gradleJava/configuration/KotlinGradleCoroutineDebugProjectResolver.kt
+++ b/plugins/kotlin/gradle/gradle-java/src/org/jetbrains/kotlin/idea/gradleJava/configuration/KotlinGradleCoroutineDebugProjectResolver.kt
@@ -5,11 +5,15 @@ package org.jetbrains.kotlin.idea.gradleJava.configuration
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.externalSystem.rt.execution.ForkedDebuggerHelper
 import com.intellij.util.Consumer
+import kotlinx.coroutines.DEBUG_PROPERTY_NAME
+import kotlinx.coroutines.DEBUG_PROPERTY_VALUE_OFF
 import org.jetbrains.kotlin.idea.extensions.KotlinJvmDebuggerFacade
 import org.jetbrains.plugins.gradle.service.project.AbstractProjectResolverExtension
 
 class KotlinGradleCoroutineDebugProjectResolver : AbstractProjectResolverExtension() {
-    val log = Logger.getInstance(this::class.java)
+    companion object {
+        val log = Logger.getInstance(this::class.java)
+    }
 
     override fun enhanceTaskProcessing(taskNames: MutableList<String>, initScriptConsumer: Consumer<String>, parameters: Map<String, String>) {
         try {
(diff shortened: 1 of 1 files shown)


abelkov:
cc @ALikhachev

plugins/kotlin/gradle/gradle-java/src/org/jetbrains/kotlin/idea/gradleJava/configuration/KotlinGradleCoroutineDebugProjectResolver.kt
```diff
@@ -9,7 +9,11 @@ import org.jetbrains.kotlin.idea.extensions.KotlinJvmDebuggerFacade
 import org.jetbrains.plugins.gradle.service.project.AbstractProjectResolverExtension
 
 class KotlinGradleCoroutineDebugProjectResolver : AbstractProjectResolverExtension() {
-    val log = Logger.getInstance(this::class.java)
+    companion object {
+        val log = Logger.getInstance(this::class.java)
+        private const val KOTLINX_COROUTINES_DEBUG_PROPERTY_NAME = "kotlinx.coroutines.debug"
```
mvicsokolova, line 14: It'd be better to use const properties `kotlinx.coroutines.DEBUG_PROPERTY_NAME` and `kotlinx.coroutines.DEBUG_PROPERTY_VALUE_OFF` here instead of hardcoding their values.

plugins/kotlin/gradle/gradle-java/src/org/jetbrains/kotlin/idea/gradleJava/configuration/KotlinGradleCoroutineDebugProjectResolver.kt
```diff
@@ -29,6 +33,12 @@ class KotlinGradleCoroutineDebugProjectResolver : AbstractProjectResolverExtensi
             gradle.taskGraph.whenReady { TaskExecutionGraph taskGraph ->
                 taskGraph.allTasks.each { Task task ->
                     if (!(task instanceof Test || task instanceof JavaExec)) return
+                    for (arg in task.getAllJvmArgs() + task.getJvmArgs() + task.getArgs()) {
```
mvicsokolova, line 36: Could we just check `System.getProperty(KOTLINX_COROUTINES_DEBUG_PROPERTY_NAME) == KOTLINX_COROUTINES_DEBUG_OFF_VALUE`?

Why do we need to run through all the task arguments?

nikita-nazarov: Fetching `System.getProperty(KOTLINX_COROUTINES_DEBUG_PROPERTY_NAME)` will fetch the property for the gradle VM, not the target one. Another this is that the property we are looking for can be returned by one of the args methods depending on how it was added to the task. For example if we add this block to the build.gradle file:
```
application {
    mainClassName = 'MainKt'
    applicationDefaultJvmArgs = ['-Dkotlinx.coroutines.debug=off']
}
```
The property will be returned by `getJvmArgs()`.
For this block:
```
test {
    useJUnitPlatform()
    jvmArgs '-Dkotlinx.coroutines.debug=off'
}
```
The property will be returned by `getAllJvmArgs()`, but not by `getJvmArgs()`

I actually found out that `getArgs()` is not needed, so I removed it.

mvicsokolova: Ok, thank you for the clarification

mvicsokolova (approved):

nikita-nazarov:
Since the changes are approved, can we now merge them?

PR #2577, opened 2023-09-20
Improve: Dependency Analyzer/Add english case-insensitive support for search bar

huanmingcn:
h3. Background
Currrently, the function **Dependency Analyzer** does not support case-insensitive, even though most of maven/gradle artifacts are typically named in lowercase English, project module names might be given in uppercase (e.g. Pascal case), which can be less user-friendly while analysising dependency (A project with numerous modules).

h3. What I have done
Compare the inputted string with dependencies in a case-insensitive manner.

h3. Coverage
No new tests were added due to this class was never covered.


huanmingcn:
### Before
before

### After
after

PR #2580, opened 2023-09-28
fix: forward mouseClicked() to correct delegate method (IDEA-333538)

jfree:
Fixes the issue reported in https://youtrack.jetbrains.com/issue/IDEA-333538/ScaledDelegatedPresentation-does-not-forward-mouse-clicks


zinoviy23:
@jfree Thank you for the fix! 
Soon our robot will merge your commit.

PR #2581, opened 2023-09-29
Collaboration wrong focus TokenLoginDialog and GitLab wrong isDefault condition

Jonatha1983:
### Summary:
This Pull Request addresses two minor issues within the collaboration module and the GitLab plugin.

##### Issue 1:
When users click the "Log In" button, the focus is incorrectly set to the server instead of the token, as illustrated in the attached screenshot.

image

##### Issue 2:
There seems to be an incorrect condition checking whether the GitLab server is the default one.

### DIFF
--- a/platform/collaboration-tools/src/com/intellij/collaboration/auth/ui/login/TokenLoginInputPanelFactory.kt
+++ b/platform/collaboration-tools/src/com/intellij/collaboration/auth/ui/login/TokenLoginInputPanelFactory.kt
@@ -6,7 +6,6 @@ import com.intellij.collaboration.ui.SingleValueModel
 import com.intellij.collaboration.util.URIUtil
 import com.intellij.openapi.ui.DialogPanel
 import com.intellij.openapi.util.NlsContexts
-import com.intellij.openapi.wm.IdeFocusManager
 import com.intellij.ui.AnimatedIcon
 import com.intellij.ui.DocumentAdapter
 import com.intellij.ui.components.fields.ExtendableTextComponent
@@ -64,7 +63,7 @@ class TokenLoginInputPanelFactory(
           }
       }
       row(CollaborationToolsBundle.message("login.field.token")) {
-        val tokenField = passwordField()
+        passwordField()
           .bindText(model::token)
           .align(AlignX.FILL)
           .resizableColumn()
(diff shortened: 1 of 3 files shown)


plugins/gitlab/src/org/jetbrains/plugins/gitlab/api/GitLabServerPath.kt
```diff
@@ -34,8 +34,13 @@ class GitLabServerPath : ServerPath {
...
     get() = toURI().resolveRelative("api/v4/")
 
+  /**
+   * I think the first part of this condition was wrong and always false - since the default String uri is always https://gitlab.com and so it's
+   *  never start with gitlab.com
+   *
+   */
   val isDefault: Boolean
-    get() = uri.startsWith("gitlab.com", true) || uri.contains("/gitlab.com", true)
+    get() = uri.startsWith("https://gitlab.com", true) || uri.contains("/gitlab.com", true)
```
iasemenov, line 43: I would say that it would be better to strip the schema from `uri` for this comparison. One can specify schema as `http`

platform/collaboration-tools/src/com/intellij/collaboration/auth/ui/login/TokenLoginInputPanelFactory.kt
```diff
@@ -80,12 +79,10 @@ class TokenLoginInputPanelFactory(
           .apply {
             onReset { component.text = "" }
           }
-          .component
 
         if (model is LoginTokenGenerator) {
           button(CollaborationToolsBundle.message("login.token.generate")) {
             model.generateToken(serverTextField.text)
-            IdeFocusManager.findInstanceByComponent(tokenField).requestFocus(tokenField, false)
```
iasemenov, line 88: Not sure if dropping this is necessary

iasemenov (changes requested):

platform/collaboration-tools/src/com/intellij/collaboration/auth/ui/login/TokenLoginInputPanelFactory.kt
Jonatha1983: I think that this is redundant because the focus should be on the Token password field in any case with or without the generate button but maybe I am missing something in here

iasemenov: Yeah. I agree, for the main use-case focus will be returned to the field

plugins/gitlab/src/org/jetbrains/plugins/gitlab/api/GitLabServerPath.kt
iasemenov: I think I can do this during the "merge", so it's fine

iasemenov (approved):

iasemenov:
Merge, thanks you!

PR #2582, opened 2023-09-29
IDEA-333730: Persist compiler plugins model in data nodes

gavra0:
... to allow data importers to always produce the same result, even if data nodes are retrieved from cache.

Also, variant switching in Android Studio relies on this. It restores data nodes from the cache, to avoid invoking full Gradle sync.


nskvortsov:
`UserDataHolder` content of a `DataNode` is not persisted regardless of copyable vs non-copyable API. 
If plugins model data is to be persisted, it should be a `data` field of a `DataNode` instance

CherepanovAleksei:
Thank you for your PR
Merged to master with the fix (Lombok key) --> https://youtrack.jetbrains.com/issue/KTIJ-27235/AnnotationBasedPluginProjectResolverExtension-stores-transient-data

PR #2583, opened 2023-10-02
[python] Add pyAssignedTo element pattern for language injection rules

Bradan:
Ability to check for specific property and attribute assignments in language injection rules.


Bradan:
[Here](https://bradan.eu/temp/intellij-community-screenshot-1.png) is a screenshot showing what I want to achieve. Why should language injections only be possible in method arguments. If you assign a string literal to a specific property or attribute it should be possible to inject a language as well.

PR #2586, opened 2023-10-08
Fix handling of line and column in ClassSource and CompositeTestSource.

LeMikaelF:
# Description of the issue

I work with a proprietary test library that uses JUnit 5 and sets a `ClassSource` on each `DynamicTest` it creates. But IntelliJ doesn't currently handle tests with an associated `ClassSource` correctly. Instead it fails in one of two ways:
1. If there is a line but no column set on the `ClassSource`, clicking the test's name in the `Run` window simply redirects to the top of the class declaration, instead of the correct line.
2. If there is a column set, IntelliJ fails to read the test results and displays the raw output from `JUnit5TestExecutionListener` (see second screenshot below).

This is because `JUnit5TestExecutionListener` prints the column of the `ClassSource`'s `FilePosition` without unwrapping in from the enclosing `Optional`, resulting in lines that contain output like `metainfo='123:Optional.empty'` in the first case, or `metainfo='123:Optional[456]'` in the second.

The first scenario (no column) fails because `JavaTestLocator` simply ignores the line and column if they can't be parsed into ints. The reason for the more obvious failure in the second case is that the message parser looks for an unescaped `]` to determine the end of a parseable event (`ServiceMessagesParser` line 62). It therefore stops parsing at the `]` in `Optional[456]`, instead of at the end of the test event. This second failure mode logs this:

```
2023-10-08 15:04:56,600 [  92063] SEVERE - #c.i.e.t.s.ServiceMessageUtil - Failed to parse service message
Details: ##teamcity[testStarted id='|[engine:junit-jupiter|]/|[class:com.mikaelfrancoeur.MyTest|]/|[test-factory:bootstrap()|]/|[dynamic-test:#1|]' name='it passes' nodeId='|[engine:junit-jupiter|]/|[class:com.mikaelfrancoeur.MyTest|]/|[test-factory:bootstrap()|]/|[dynamic-test:#1|]' parentNodeId='|[engine:junit-jupiter|]/|[class:com.mikaelfrancoeur.MyTest|]/|[test-factory:bootstrap()|]' locationHint='java:suite://com.mikaelfrancoeur.MyTest' metainfo='11:Optional[456]
testFramework:JUnit
java.text.ParseException: Error while parsing TeamCity service message: Value should end with "'". Valid service message has a form of "##teamcity[messageName name1='escaped_value' name2='escaped_value']" where escaped_value uses substitutions: '->|', [->|[, ]->|], |->||, newline->|n
	at jetbrains.buildServer.messages.serviceMessages.MapSerializerUtil.stringToProperties(MapSerializerUtil.java:134)
	at jetbrains.buildServer.messages.serviceMessages.ServiceMessage.parseAttributes(ServiceMessage.java:237)
	at jetbrains.buildServer.messages.serviceMessages.ServiceMessage.init(ServiceMessage.java:371)
	at jetbrains.buildServer.messages.serviceMessages.ServiceMessagesParser.doParse(ServiceMessagesParser.java:136)
	at jetbrains.buildServer.messages.serviceMessages.ServiceMessagesParser.parse(ServiceMessagesParser.java:54)
	at com.intellij.execution.testframework.sm.ServiceMessageUtil.parse(ServiceMessageUtil.java:33)
        ...
```
## Demonstration of the issue

When the test has a `ClassSource` with no column, the tree displays correctly but you can't click it to navigate to the test sources:

image

When the test has a `ClassSource` with a column:

image

# Description of the solution

I solved this by adding unwrapping the `getColumn()` optional with `.orElse(1)` (`JUnit5TestExecutionListener` line 369). I also added a test to cover this and used the opportunity to add coverage to the other `TestSource` subclasses for which location data is emitted by `JUnit5TestExecutionListener` (`FileSource`, `MethodSource`, and `CompositeTestSource`). This revealed an other issue: the `metainfo` of `CompositeTestSource` wasn't being emitted, even though its `locationHint` was being emitted, so I added support for this as well (`JUnit5TestExecutionListener` line 346).

Fixing this also surfaced another bug. JUnit treats its line and column numbers as 1-based (see the `line > 0` and `column > 0` preconditions in `org.junit.platform.engine.support.descriptor.FilePosition`). This is coherent with the `FileUrlProvider` used to handle line numbers for instances of `FileSource` (URIs starting in `file:`). But `JavaTestLocator`, used by IntelliJ for test URIs starting in `java:`, uses the 0-based `OpenFileDescriptor`. This meant that clicking on a test node in the `Run` window would bring you to the line after the test. I fixed this by subtracting 1 from the line and column when the text string is emitted (`JUnit5TestExecutionListener` line 368), to avoid impacting other code paths, since things like breakpoint also used the 0-based behaviour. The method in `DocumentUtil` that was responsible for this 0-based logic wasn't directly tested, so I added a unit test to further document the 0-based behaviour.

# Results

IntelliJ community, test using a `ClassSource` with a column, after this patch (for the "before" version, see above):

image

Moreover, you can now click on a test node that's using either or a `ClassSource` or an applicable `TestSource` wrapped in `CompositeTestSource`, and it will take you to the correct line. In this case, you can click `it passes`. The contextual menu is also accessible for test nodes, with options to run/debug a certain test.

-----

I'll be happy to address any comments on this PR. N.B. the new test will fail when run from IntelliJ, unless you rename `JUnit5TestExecutionListener` to something like `JUnit5TestExecutionListener2` to avoid java using the bundled class instead.

### DIFF
--- a/platform/platform-tests/testSrc/com/intellij/openapi/editor/DocumentUtilTest.java
+++ b/platform/platform-tests/testSrc/com/intellij/openapi/editor/DocumentUtilTest.java
@@ -18,4 +18,15 @@ public class DocumentUtilTest {
     assertEquals(" \t ", DocumentUtil.getIndent(doc, doc.getLineStartOffset(1) + 2).toString());
     assertEquals("", DocumentUtil.getIndent(doc, doc.getLineStartOffset(2)).toString());
   }
+
+  @Test
+  public void calculateOffsetIsZeroBased() {
+    final Document doc = new DocumentImpl("""
+                                            line1
+                                            line2
+                                            """);
+
+    int offset = DocumentUtil.calculateOffset(doc, 0, 0, 0);
+    assertEquals(0, offset);
+  }
 }
--- a/plugins/junit5_rt/src/com/intellij/junit5/JUnit5TestExecutionListener.java
+++ b/plugins/junit5_rt/src/com/intellij/junit5/JUnit5TestExecutionListener.java
@@ -25,6 +25,7 @@ import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.*;
+import java.util.stream.Stream;
 
 import static com.intellij.rt.execution.TestListenerProtocol.CLASS_CONFIGURATION;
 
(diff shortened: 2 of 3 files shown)


BartvHelvert (changes requested):

LeMikaelF:
Replying here to https://github.com/JetBrains/intellij-community/pull/2586#discussion_r1377595579, because GitHub isn't letting me comment on the thread:

-----

I didn't think of that. It would also require a `.filter(Optional::isPresent)`, but you're right it's way simpler. Unfortunately, this made me realize that there's a bug in my implementation and fixing it might require some heavier refactoring.

What I did in my initial commit was pick the leftmost leaf node of a `CompositeTestSource` tree that had a supported `metaInfo`. This works fine in isolation. But there's actually 2 separate depth-first searches on `CompositeTestSource` trees in this class, the second being `getLocationHintValue()`, to get the `locationHint` property. The problem is that `locationHint` and `metaInfo` work together to indicate a unique location, so if they get desynchronized, their combination will be invalid. Consider this, using the setup from `JUnit5NavigationTest`:

```java
@Test
void locationHintForCompositeSourcePickTheFirstSupportedSourceInTheList() {
  myTestSource = CompositeTestSource.from(
    Arrays.asList(anyUnsupportedTestSource(),
                  ClassSource.from(String.class),
                  ClassSource.from(Object.class, FilePosition.from(4, 5)),
                  MethodSource.from("className", "methodName")));

  String locationHint = locationHint();

  // oops! locationHint and metainfo are inconsistent
  Assertions.assertEquals(
    "locationHint='java:suite://java.lang.String' metainfo='3:4'", locationHint);
}
```
One way to solve this would be to unify `getLocationHintValue()` and `getMetaInfo()` behind a single `getLocationHint()` method. `getLocationHintValue()` is only visible for testing anyways, so the change would be contained to the `JUnit5TestExecutionListener` class and the `JUnit5NavigationTest` test. This seems like the best option, but it involves some heavier refactoring, and since it was just a side quest to the main goal of this PR, I reverted the changes that were related to `CompositeTestSource`, keeping the pre-existing behaviour of providing only a `locationHint` and no `metainfo`.

If you think merging the retrieval of `locationHint` and `metainfo` would be an appropriate solution, maybe that's something that can be addressed in another PR.

BartvHelvert:
Thanks! I think it will be ok to merge the current fixes for line and column offset. We can address `CompositeTestSource` in the future.

BartvHelvert:
Merged, thanks!

PR #2588, opened 2023-10-11
IDEA-334617: Suggest installing Azure plugin for users who have Azure CLI installed

Flanker32:
### Changes
- Add `az` to the `ALLOWED_EXECUTABLES` in `EnvironmentDependencyCollector`, so that IDEA could detect azure cli installation and recommend Azure plugins

### Todo
Register executable dependencySupport in Azure Toolkit for IntelliJ and Azure Toolkit for Rider
- https://github.com/JetBrains/azure-tools-for-intellij/pull/721
- https://github.com/microsoft/azure-tools-for-java/pull/7933

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/pluginsAdvertisement/EnvironmentDependencyCollector.kt
+++ b/platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/pluginsAdvertisement/EnvironmentDependencyCollector.kt
@@ -15,6 +15,7 @@ import kotlin.io.path.isRegularFile
 
 internal class EnvironmentDependencyCollector : DependencyCollector {
   private val ALLOWED_EXECUTABLES: List<String> = listOf(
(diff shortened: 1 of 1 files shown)


platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/pluginsAdvertisement/EnvironmentDependencyCollector.kt
```diff
@@ -15,6 +15,7 @@ import kotlin.io.path.isRegularFile
 
 internal class EnvironmentDependencyCollector : DependencyCollector {
   private val ALLOWED_EXECUTABLES: List<String> = listOf(
+    "az",
```
wangmingliang-ms, line 18: suffix of `az` is `cmd`, which is not in the supported ext list.

jreznot:
We already added az as allowed binary some time ago

PR #2590, opened 2023-10-12
Gifs now work when viewing plugin media

evictoria:
- Removed the code appending .png to the file name.
- Added the component itself as the ImageObserver when the plugin media is drawn.
- IDEA-334480


meiMingle:
The automatic playback of Gifs has a little bit of performance waste. As far as I know, gif playback requires a separate thread to execute. It might be better to play the gif when the user clicks it to make it full screen. It would be great if you could add 'pause', 'resume', 'stop', and 'restart' buttons to control the playback progress.

meiMingle:
Then I think it's worth supporting

evictoria:
Hello @alexander-lobas any feedback about the changes? 😁

alexander-lobas:
Thanks for your work, I'll just push it in the master.

PR #2592, opened 2023-10-19
IDEA-335494 Fixed clashing mnemonics

evictoria:
Protected's mnemonic changed from o to t:
mnemonics-visibility-fixed

A certain limitation is if the JavaVisibilityPanel is used twice in the same dialog:
mnemonics-visibility-limitation


evictoria:
Hello @BasLeijdekkers any feedback about the change? 😁

PR #2594, opened 2023-10-23
KTIJ-27441 False positive "Redundant overriding method" when override toString() method in data class which has parent class

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-27441


abelkov (approved):
Thank you!

PR #2596, opened 2023-10-26
upgrade gradlew to 7.6.3

jcrben:
When downloading over a proxy the wrapper times out and 7.6 and above fixes that. https://stackoverflow.com/questions/74222309/gradle-wrapper-times-out-when-using-a-proxy


trespasserw:
Hi! Thanks for the patch, but I'm hesitant about adding the timeout property to the project, because it doesn't belong to it and reflects peculiarities of a specific build environment. Would it be enough to just upgrade the Gradle wrapper, so that one can specify the timeout via a command-line parameter or an environment variable or something?

jcrben:
@trespasserw that configuration change came from upstream script when I ran this:

`./gradlew wrapper --gradle-version=7.6`

It's a bit confusing since I also want a timeout configuration change but I didn't hand edit the file. Do you want me to edit it even tho it came from the upstream automated update? I'm happy to remove it.

jcrben:
Thank you!!

PR #2598, opened 2023-10-27
Show local plugin's description & change notes when in plugin sandbox mode

evictoria:
As a plugin developer, I want to see locally the changes I make for my plugin's description and change notes.
The issue this is trying to fix is documented in IDEA-336003.


alexander-lobas:
Fixed.

PR #2599, opened 2023-10-27
[java] Pull up EditorFactoryImpl methods to interface

sadv1r:
I needed to use this methods in my plugin and I searched how to use it on IDE code base and I found it a little scary and not future proofed to just cast an interface to specific implementation without checks


chashnikov:
I've rebased the commit on the top of the 'master' branch, fixed conflicts, made other necessary changes and pushed it. I'm sorry that it took so long. Thank you for the contribution!

PR #2600, opened 2023-10-30
KTIJ-26237: Don't ever return duplicates of classes

omega697:
This change ensures that `getClassNames` and `findClassesWithoutJvmFacade` never return duplicates, even if multiple `PsiElementFinder`s find the same `PsiClass`es.

### DIFF
--- a/plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
+++ b/plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
@@ -1,0 +1,29 @@
+// Copyright 2000-2023 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.kotlin.idea.editor.quickDoc
+
+import com.intellij.codeInsight.javadoc.JavaDocInfoGenerator
+import com.intellij.psi.PsiPackage
+import org.jetbrains.kotlin.idea.test.KotlinLightCodeInsightFixtureTestCase
+import org.junit.Assert
+
+class PackageQuickDocTest: KotlinLightCodeInsightFixtureTestCase() {
+    fun testNoDuplicates() {
+        val psiFile = myFixture.addFileToProject(
+            "com/example/nodupes/Fantastic.kt", """
+            package com.example.nod<caret>upes
+            class Fantastic
+        """.trimIndent()
+        )
+        myFixture.configureFromExistingVirtualFile(psiFile.virtualFile)
+
+        val psiPackage = myFixture.elementAtCaret as PsiPackage
+
+        val stringBuilder = StringBuilder()
+        JavaDocInfoGenerator(myFixture.project, psiPackage).generateDocInfoCore(stringBuilder, false)
+        val html = stringBuilder.toString()
+
+        val hrefPattern = Regex(Regex.escape("psi_element://com.example.nodupes.Fantastic"))
+        val numMatches = hrefPattern.findAll(html).count()
+        Assert.assertEquals(1, numMatches)
+    }
+}
--- a/java/java-psi-impl/src/com/intellij/psi/impl/JavaPsiFacadeImpl.java
+++ b/java/java-psi-impl/src/com/intellij/psi/impl/JavaPsiFacadeImpl.java
@@ -167,7 +167,7 @@ public class JavaPsiFacadeImpl extends JavaPsiFacadeEx {
       }
     }
 
-    return result == null ? Collections.emptyList() : result;
+    return result == null ? Collections.emptyList() : new ArrayList<>(result);
   }
 
   private static Predicate<PsiClass> getFilterFromFinders(@NotNull GlobalSearchScope scope, @NotNull List<? extends PsiElementFinder> finders) {
@@ -273,7 +273,7 @@ public class JavaPsiFacadeImpl extends JavaPsiFacadeEx {
     for (PsiElementFinder finder : finders) {
       PsiClass[] classes = finder.getClasses(psiPackage, scope);
       if (classes.length == 0) continue;
-      if (result == null) result = new LinkedHashSet<>(classes.length);
+      if (result == null) result = new HashSet<>(classes.length);
       filterClassesAndAppend(finder, classesFilter, classes, result);
     }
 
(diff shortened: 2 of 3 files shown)


abelkov:
@omega697 Could you please add tests if possible

omega697:
Test added.

vladimirdolzhenko:
I think it would be better to create a test that replicates the original problem. 

Minor: [KTIJ-27237](https://youtrack.jetbrains.com/issue/KTIJ-27237) specified in the title of PR is fixed and relates to a completely different issue.

omega697:
Added a test for the original problem as well.

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
```diff
@@ -0,0 +1,29 @@
+// Copyright 2000-2023 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.kotlin.idea.editor.quickDoc
```
vladimirdolzhenko, line 2: this test targets K1 plugin, it would be better to make it shared for both K1 and K2.

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+class PackageQuickDocTest: KotlinLightCodeInsightFixtureTestCase() {
+    fun testNoDuplicates() {
+        val psiFile = myFixture.addFileToProject(
+            "com/example/nodupes/Fantastic.kt", """
+            package com.example.nod<caret>upes
+            class Fantastic
+        """.trimIndent()
+        )
+        myFixture.configureFromExistingVirtualFile(psiFile.virtualFile)
```
vladimirdolzhenko, line 17: I think it would be nice to make it more generic and use `testData` as many other tests do.

omega697: I don't understand how that makes it more generic. This test is small and contained and minimizes external dependencies. It's clear what's being tested directly from the test.

I assume those tests use `testData` because it would be unwieldy to set up complex file structures/contents within a unit test. But this test only needs to add a single, empty class.

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
omega697: Can you clarify how this targets K1? I don't see any references to K1 or K2 anywhere - I must be missing something.

vladimirdolzhenko: actually, two plugins are built from `plugins/kotlin/` codebase - for K1 and for K2. 

They share some common code, but some parts are completely different. 
`plugins/kotlin/idea/tests/test/` are tests for K1. You can see that some modules have `-k2` in their name, or shared if it is used for both plugins. 

In some cases, K2 test utilizes K1 test (both logic and test data) but specifies that it is K2-version:  for instance, you can compare `AbstractAutoImportTestBase` and `AbstractK2AutoImportTest`

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+        val psiPackage = myFixture.elementAtCaret as PsiPackage
+
+        val stringBuilder = StringBuilder()
+        JavaDocInfoGenerator(myFixture.project, psiPackage).generateDocInfoCore(stringBuilder, false)
+        val html = stringBuilder.toString()
+
+        val hrefPattern = Regex(Regex.escape("psi_element://com.example.nodupes.Fantastic"))
+        val numMatches = hrefPattern.findAll(html).count()
+        Assert.assertEquals(1, numMatches)
```
vladimirdolzhenko, line 27: it would be better to check not just the number of elements, but what exact elements they are:

when test fails, it gives more useful information about the problem.

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
omega697: Ahh, OK. So do I just need to move this test to a different module? Which module would you suggest?

plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/editor/quickDoc/PackageQuickDocTest.kt
omega697: I agree it would be better to have more robust testing for quick documentation.

This test case, however, is only testing that there are no duplicates (`testNoDuplicates`) - if the test fails, it will be clear that there is a problem with duplicates. IMO, if we want to test a different aspect of the quickdoc, (e.g. that all classes are represented), that seems like a different test case outside the scope of this change, which is specifically to address the duplicate issue. 

There is a more comprehensive unit test for the changes I made to `JavaPsiFacadeImpl` in the other test I added.

vladimirdolzhenko:
Sorry for the long delay,
I think I cannot accept this PR at this stage:
- I would rather try to address the problem on the level of quick documentation provider rather in `JavaPsiFacadeImpl`: `JavaPsiFacadeImpl` is heavily used from many parts, esp. resolve. It means collecting classes to set with further conversion to a list creates extra overhead and noticeable performance penalty. 
- We usually do not practice mocking items in tests, moreover `JavaPsiFacadeImplTest` just enforces and validates the contract of `JavaPsiFacadeImpl`, do not see a particular reason for it.
- Minor: commit message format. In Kotlin plugin we still follow https://github.com/JetBrains/kotlin/blob/master/docs/contributing.md#rules-for-commit-messages + some IntelliJ style: you can check out on other commits — usually they have `[kotlin]` prefix and reference to the ticket at the end of it like `#KTIJ-26237`

PR #2602, opened 2023-11-01
KTIJ-27688: Use the Kotlin source instead of the decompiled source if available

omega697:
https://youtrack.jetbrains.com/issue/KTIJ-27688 explains how we do not use the available source for rendering quick documentation when it is available, instead using decompiled source that does not have defaults. This change detects when we are rendering documentation text for a `PsiElement` in a Kotlin `.class` file and redirects us to the source, if available.


vladimirdolzhenko:
please add tests

omega697:
Tests added.

akozlova:
Please try to use unconditional `element.navigationElement` instead. It will automatically "navigate" to sources from decompiled code and if nothing found, keep `this`. Thanks

akozlova:
LGTM

akozlova:
When Vladimir tried to merge, we get failed test. The problem comes from elements which are not kotlin elements but have kotlin navigatables, e.g. different Spring beans. So I suggest to check explicitly `(element as? KtElement)?.navigationElement`, otherwise let other providers do their job.
Sorry that I haven't spot it earlier.

omega697:
Done :D

vladimirdolzhenko:
Thanks a lot, merged as https://github.com/JetBrains/intellij-community/commit/d5edfd31000dd7e7e433d79b3ac9d33ad2ca9114

PR #2603, opened 2023-11-02
KTIJ-27132 IDEA shouldn't suggest opt-in/propagate quick-fix if the annotation is invisible

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-27132


abelkov (approved):
Thank you, merged with small changes

PR #2611, opened 2023-11-16
Add navigate/diff command to jetbrains:// protocol

0x6368656174:
The command `jetbrains://{IDE_NAME}/navigate/diff` supporting was added. 

The command opens a simple diff view between two files in two VCS revisions. It could be useful for opening the diff directly from Bitbucket/Github/Gitlab PR/MRs.

The command expects the following arguments:
- `path_left` the path to the file relative to the project root that should be opened on the left side of the diff view.
- `path_right` the path to the file relative to the project root that should be opened on the right side of the diff view.
- `revision_left` the VCS revision of file that should be opened on the left side of the diff view. 
- `revision_right` the VCS revision of file that should be opened on the right side of the diff view.

Parameters `path_left/path_right` and `revision_left/revision_right` use the same logic as the parameters `path` and `revision` from `navigate/reference` command.

The command requires the same arguments `project` or `origin` so as the `/navigate/reference` command for finding the project of files.

Also, `/navigate/reference` command was refactored for allowing to share their logic for finding a project by `project/origin`, find `VirutalFile` or `PSIElement` with other commands. The logic code is not changed, only split to multiple files.

Example of using the command: `jetbrains://idea/navigate/diff?project=intellij-community&path_left=README.md&path_right=README.md&revision_left=302af0a2cebbd2b90be04f67c3cf46c715745f73&revision_right=c82d7fe1df0a7deb0e6198226f7a1c0dd66b5326`


iasemenov:
Hi. I'm sorry it took so so long.
Could you please clarify your use-case for this functionality?

PR #2612, opened 2023-11-20
Expose comparator option on TreeFileChooserFactory and TreeClassChooserFactory

vmadalin:
The existing `TreeFileChooserFactory` and `TreeClassChooserFactory` don't expose a way to sort the existing project tree in a different order than the one defined by `AlphaComparator.INSTANCE`. For this reason, and in order to keep consistency with other representation of the project tree representation a optional comparator was considered to be exposed, allowing in this way to have same Project pane structure cross different views. 

### Overview
Screenshot 2023-11-17 at 17 44 07


vmadalin:
> I can see reasoning behind this, but why did you remove a bunch of methods from the `TreeClassChooserFactory` class? It's a part of the public API, and we can't just randomly remove public API methods for no good reason.

The reason of removing those methods from `TreeClassChooserFactory` is because they are already present on `TreeFileChooserFactory`.  In this way avoid having duplication and especially force the usage of a proper chooser factory when creating the `TreeFileChooser`, since the existing implementation is mixing concepts considering that `TreeClassChooserFactory` as is name suggest should be tree class specific. However, those changes are completely optional and if we prefer to keep the API I'm happy to just revert them.

> Also, could you please create a [YouTrack](https://youtrack.jetbrains.com/) ticket on what you're trying to achieve here and add the ticket ID (in the form of IDEA-XXXXXX) to the commit message? This would ensure that these changes go through the proper channels and are accounted for everywhere they should.

Done https://youtrack.jetbrains.com/issue/IDEA-339513

stachenov:
OK, that makes sense, but it still needs some improvement.

First, this definitely should be done in a separate commit. Let's keep unrelated changes separate.

Second, yes, we do need to preserve API. Especially since this one is actually used externally. So instead of removing them, please add a `@Deprecated` annotation instead, add Javadoc explaining why they're deprecated and rewrite the methods themselves to delegate to `TreeFileChooserFactory` methods instead of duplicating their code.

It's OK if both commits refer to the same YT ticket, it just falls under the "cleaned up this mess while working on that ticket" category.

platform/lang-impl/src/com/intellij/ide/util/AbstractTreeClassChooserDialog.java
```diff
@@ -55,6 +56,7 @@ public abstract class AbstractTreeClassChooserDialog<T extends PsiNamedElement>
   private final @NotNull Project myProject;
   private final GlobalSearchScope myScope;
   private final @NotNull Filter<T> myClassFilter;
+  private final @Nullable Comparator<? super NodeDescriptor<?>> myComparator;
```
stachenov, line 59: Looks like this should be `@NotNull` instead because you set it to `AlphaComparator.INSTANCE` if `null` is passed into the constructor.

platform/lang-impl/src/com/intellij/ide/util/TreeFileChooserDialog.java
```diff
@@ -66,6 +67,7 @@ public final class TreeFileChooserDialog extends DialogWrapper implements TreeFi
   private final @Nullable PsiFile myInitialFile;
   private final @Nullable PsiFileFilter myFilter;
   private final @Nullable FileType myFileType;
+  private final @Nullable Comparator<? super NodeDescriptor<?>> myComparator;
```
stachenov, line 70: Ditto.

stachenov (changes requested):
Looks mostly good now, just one nit that should probably be fixed with a fixup commit squashed into the existing one. Or feel free to create one more commit fixing that if you're more comfortable that way.

stachenov (approved):
Looks good now. I'll start our internal process of merging, this PR should close automatically once it's over.

platform/lang-impl/src/com/intellij/ide/util/AbstractTreeClassChooserDialog.java
```diff
@@ -80,14 +82,15 @@ public AbstractTreeClassChooserDialog(@NlsContexts.DialogTitle String title,
...
-    this(title, project, scope, elementClass, classFilter, null, initialClass, false, true);
+    this(title, project, scope, elementClass, classFilter, null, null, initialClass, false, true);
   }
 
   public AbstractTreeClassChooserDialog(@NlsContexts.DialogTitle String title,
                                         @NotNull Project project,
                                         GlobalSearchScope scope,
                                         @NotNull Class<T> elementClass,
                                         @Nullable Filter<T> classFilter,
+                                        @Nullable Comparator<? super NodeDescriptor<?>> comparator,
```
stachenov, line 93: Looks like we need yet another constructor overload here.

platform/lang-impl/src/com/intellij/ide/util/TreeFileChooserDialog.java
```diff
@@ -76,12 +78,14 @@ public TreeFileChooserDialog(@NotNull Project project,
                                final @Nullable PsiFile initialFile,
                                @Nullable FileType fileType,
                                @Nullable PsiFileFilter filter,
+                               @Nullable Comparator<? super NodeDescriptor<?>> comparator,
```
stachenov, line 81: And here.

stachenov (changes requested):
Oops, no, no luck this time. Looks like a couple of classes in the lang-impl module are used by some external plugins, so instead of modifying the signature, looks like we need yet another constructor overload. I've marked those places. Without this fix, our internal API watcher doesn't accept these changes, and for a good reason.

PR #2613, opened 2023-11-21
IDEA-299661 Run console input is reset with every output

ItsFlare:
Breaking this loop effectively discards all `ConsoleViewContentType.USER_INPUT` tokens buffered before any other type of token, which prevents interleaving of `System.out` / `System.in` and does not align with the behavior of typical terminals. In order to discover all user input, we simply keep iterating.

[YouTrack Issue](https://youtrack.jetbrains.com/issue/IDEA-299661)


cdracm:
pushed, thank you

PR #2614, opened 2023-11-21
IDEA-338696: Resolve GRADLE_USER_HOME environment variable defined on macOS

vmadalin:
Following the original documentation https://www.jetbrains.com/help/idea/gradle-settings.html, users should be able to override their `GRADLE_USER_HOME` environment variable and this should be taken into consideration for the Gradle Daemon configuration.

The reason is because platform relies on Gradle implementation of `BuildLayoutParameters.findGradleUserHomeDir()` which uses `System` class causing on macOS to not be able to obtain the defined environment variable. For this reason will be required to move this logic into the platform directly and use `Environment` class, being this aligned with https://github.com/JetBrains/intellij-community/commit/474b49db890e986170ce91655a80d65ea36c63ce change about `JAVA_HOME`.

#### Overview 
Platform issue


nskvortsov:
LGTM, thank you. Running the CI with suggested changes.

PR #2615, opened 2023-11-22
KTIJ-27745 K2: Specify type in constructor

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-27745


vladimirdolzhenko:
sorry for the delay, 
this PR in breaks [propertyWithGetterWithBlockBody](https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/idea/tests/testData/quickfix/typeAddition/propertyWithGetterWithBlockBody.kt)

`"Specify type explicitly"` becomes unavailable for this case:

```
class My {
    val <caret>x
        get() {
            return 3.14
        }
}
```

t-kameyama:
Fixed. Please review.

vladimirdolzhenko:

PR #2616, opened 2023-11-25
KTIJ-27737 "Replace with for loop" result is incompilable if receiver is nullable function call

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-27737


abelkov (approved):
Thank you!

PR #2624, opened 2023-12-08
Feature: JUnit patcher supports excluding module dependency

VictoryWangCN:
In the previous version, when I used classpathDependencyExclude with MavenId configured as a Module Dependency, the JUnit plugin did not exclude this dependency's path (e.g., module-dependency/target/classes). For more details, please refer to the test case in the PR.

I have attempted a fix and would greatly appreciate it if you could perform a code review on my code.

### DIFF
--- a/plugins/maven/src/test/java/org/jetbrains/idea/maven/execution/MavenJUnitPatcherTest.kt
+++ b/plugins/maven/src/test/java/org/jetbrains/idea/maven/execution/MavenJUnitPatcherTest.kt
@@ -25,6 +25,8 @@ import org.intellij.lang.annotations.Language
 import org.jetbrains.idea.maven.project.MavenProjectSettings
 import org.junit.Test
 import java.io.File
+import java.nio.file.Paths
+import kotlin.io.path.name
 
 class MavenJUnitPatcherTest : MavenMultiVersionImportingTestCase() {
   override fun runInDispatchThread() = true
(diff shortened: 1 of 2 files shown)


BartvHelvert:
@VictoryWangCN Hello! Sorry for the delay. Could you maybe also create an issue in our [issue tracker](https://youtrack.jetbrains.com/issues/IDEA) for this issue?

plugins/maven/src/test/java/org/jetbrains/idea/maven/execution/MavenJUnitPatcherTest.kt
```diff
@@ -36,6 +36,82 @@ class MavenJUnitPatcherTest : MavenMultiVersionImportingTestCase() {
...
+      if (path.startsWith(projectPath)) {
+        path.replace(projectPath, "")
+      }
+      else {
+        path.substringAfterLast('/')
+      }
+    }
+
+    javaParameters.configureByModule(module, JavaParameters.CLASSES_AND_TESTS, IdeaTestUtil.getMockJdk18())
+    assertEquals(listOf("/m/target/classes", "/dep/target/classes", "junit-4.0.jar"),
```
BartvHelvert, line 107: The test fails for me here because `classPath` is the full path and you assert it to the relative path here

BartvHelvert (changes requested):

plugins/maven/src/test/java/org/jetbrains/idea/maven/execution/MavenJUnitPatcherTest.kt
VictoryWangCN: fixed

VictoryWangCN:
https://youtrack.jetbrains.com/issue/IDEA-346507/

created

plugins/maven/src/test/java/org/jetbrains/idea/maven/execution/MavenJUnitPatcherTest.kt
BartvHelvert: It doesn't look fixed for me on the first assert:
```
Expected :[dep/target/classes, junit-4.0.jar, m/target/classes]
Actual   :[dep\target\classes, junit-4.0.jar, m\target\classes]
```

But even if I fix the forward/backslash issue here on second assert the result is:
```
Expected :[junit-4.0.jar, m\target\classes]
Actual   :[dep\target\classes, junit-4.0.jar, m\target\classes]
```

VictoryWangCN: Currently, I am unable to reproduce the second issue on my end. Could you provide some additional information, such as the environment and how you are executing the test case?

BartvHelvert: Looks like it's ok on Linux/MacOS but not on Windows. I have found the issue, will merge your change and the issue with local paths myself.

BartvHelvert:
Merged, thanks!

PR #2625, opened 2023-12-11
Cleaning and refactoring strange code

TheLivan:
Hi. I simple corrected the oddities in the code.


xiexed (approved):

amaembo:
Thank you. We will apply each of theses changes separately, after investigation. Some are already applied.

PR #2626, opened 2023-12-13
IDEA-288455 - Add registry option for ignoring force push warning

UnderKoen:
This change addresses https://youtrack.jetbrains.com/issue/IDEA-288455

It is not a full fix, just a registry option to achieve the same result.


a-krasilnikov:
I closed PR as it was implemented like a separate setting
[https://youtrack.jetbrains.com/issue/IJPL-73008 ](https://youtrack.jetbrains.com/issue/IJPL-73008)

PR #2628, opened 2023-12-17
(🐞) PyTypeProvider: add new method `getAnnotaionType`

KotlinIsland:
[youtrack](https://youtrack.jetbrains.com/issue/PY-65388/PyTypingTypeProvidergetParameterType-getReturnType-doesnt-return-null)


east825:
Closing because we've introduced the experimental `com.jetbrains.python.codeInsight.typing.PyTypeHintProvider` extension point for the same purpose.

PR #2630, opened 2023-12-28
Add registry property to allow user to disable run tool window from navigating to first error

mgroth0:
This is intended to resolve. https://youtrack.jetbrains.com/issue/IDEA-298449/Options-related-to-exception-handling-in-the-Run-Tool-Window


nskvortsov:
Thank you.
We are working on a proper solution that should not require the registry key.

PR #2634, opened 2024-01-04
IDEA-287090 update plugin parent classloader when optional dependency is loaded dynamically

jansorg:
This is a fix for https://youtrack.jetbrains.com/issue/IDEA-287090

When an optional plugin dependency `optionalPlugin` of plugin `mainPlugin` is loaded dynamically at runtime, then the features of `mainPlugin` using `optionalPlugin` were enabled, but the Java classpath of `mainPlugin` did not include the classes of `optionalPlugin`. 
Further details have been documented in the YouTrack issue above.

The bug is a major nuisance for plugins, which make use of optional dependencies.

This PR changes the `parents` property of `PluginClassLoader` to be mutable to allow changing it when an optional dependency is loaded.
I can't tell if that's how you'd like it, but IMHO it should be safe because it's volatile and only updated in a WriteAction, AFAIK.

The test only validates the handling with old-style plugins, because AFAIK the new-style module approach is already handling this case, i.e. the module depending on `optionalPlugin` is loaded when `optionalPlugin` is loaded and unloaded again when `optionalPlugin` is removed again.
I don't have any experience with the new-style module descriptors, so I may have misunderstood the code.

Thank you for your consideration!

### DIFF
--- a/platform/core-impl/src/com/intellij/ide/plugins/cl/PluginClassLoader.kt
+++ b/platform/core-impl/src/com/intellij/ide/plugins/cl/PluginClassLoader.kt
@@ -62,7 +62,8 @@ private val parentListCacheIdCounter = AtomicInteger()
 
 @ApiStatus.Internal
 class PluginClassLoader(classPath: ClassPath,
-                        private val parents: Array<IdeaPluginDescriptorImpl>,
+                        @Volatile
+                        private var parents: Array<IdeaPluginDescriptorImpl>,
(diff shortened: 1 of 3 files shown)


jansorg:
This has been rejected on YouTrack because the new plugin model offers a better way to do this, https://youtrack.jetbrains.com/issue/IJPL-1115/Classloader-fails-to-load-class-of-dynamically-loaded-optional-dependency

PR #2637, opened 2024-01-15
IDEA-343109 - add MavenPropertyDIalog size remembering

kamilkrzywanski:
I've added method to save EditMavenPropertyDialog size on close.


zulkar:
merged

PR #2638, opened 2024-01-16
JavaHomeFinder: add mise install directories

muescha:
# Where

- JavaHomeFinder

# Changed 

- add mise install directories

# Documentation

https://mise.jdx.dev/configuration.html#mise-data-dir

> `MISE_DATA_DIR`
> Default: `~/.local/share/mise` or `$XDG_DATA_HOME/mise`
> 
> This is the directory where mise stores plugins and tool installs. These are not supposed to be shared across machines.

# About Mise 

`mise` claim to be a drop in replacement for `asdf` (see [Mise FAQ: How do I migrate from asdf?](https://mise.jdx.dev/faq.html#how-do-i-migrate-from-asdf)) written in Rust.


yopox:
Hi @muescha, thanks for the patch! I added support for `~/.local/share/mise/java` in February but your solution is better. I will merge it.

muescha:
Thx :)

PR #2639, opened 2024-01-19
Don't add orphaned labels to LocalHistory RevisionData

csrhodes:
If the list of local history revisions is prefixed by one or more labels, simply drop them to avoid spurious revisions showing up in the Local History Dialog.

Addresses IDEA-343512


juliabeliaeva:
@csrhodes thank you for the fix!

PR #2640, opened 2024-01-19
IDEA-304758 - Add support for Nushell when loading environment

shyim:
This fixes https://youtrack.jetbrains.com/issue/IDEA-304758, by adjusting the script when the shell is Nushell to `| save <path>`. Also escaped paths requires a `^` in nushell, otherwise it just outputs the path as a string.

I tested the change with Nushell and Bash. For every other shell, the generated script is exactly the same as before.

### DIFF
--- a/platform/util/src/com/intellij/util/EnvironmentUtil.java
+++ b/platform/util/src/com/intellij/util/EnvironmentUtil.java
@@ -264,10 +264,23 @@ public final class EnvironmentUtil {
         }
         readerCmd.append(SHELL_SOURCE_COMMAND).append(" \"").append(file).append("\" && ");
       }
-
-      readerCmd.append("'").append(reader).append("' > '").append(envDataFile.toAbsolutePath()).append("'");
-
       List<String> command = getShellProcessCommand();
+      boolean isNushell = command.get(0).endsWith("/nu");
+
+      if (isNushell) {
+        readerCmd.append("^");
+      }
+
+      readerCmd.append("'").append(reader);
+
+      if (isNushell) {
+        readerCmd.append("' | save '");
+      } else {
+        readerCmd.append("' > '");
+      }
+
+      readerCmd.append(envDataFile.toAbsolutePath()).append("'");
+
       int idx = command.indexOf(SHELL_COMMAND_ARGUMENT);
       if (idx >= 0) {
         // if there is already a command append command to the end


ultinous-kszegi:
Any reason it can't be merged?

blended-bram:
I'd love for nushell support to get added; but I think there is a smaller adjustment that can fix the bug in a more platform-agnostic manner.

The flaw which is at fault for causing nushell not to be supported is that the process being spawned in order to extract the environment is expecting the user's shell to be POSIX compliant. The mechanism for extracting the environment, for some reason, uses shell programming to forward the output of the executable into a temporary file. For some reason this is done through the command passed into the shell, rather than performing this task in java land.
Because of that the forward-info-file notation (`program > file`) must be supported by the user's shell. Not the case with nushell.

Here the fix proposed changes the shell command into one that is nushell compliant. This means that the same problem will occur with any other non-posix shell. I'd argue that instead the problem that needs to be addressed is that the IDE is relying on shell programming to extract the environment data into a file, while it could do this perfectly itself.

For example:
1. Passing the output location as an argument to the environment extraction program.
2. Reading the extraction program's stdout into the IDE memory and not using a temp file at all.

I haven't been able to think any reason why the environment extraction is written the way it is; relying on posix for the shell while you could also just read the program's stdout into java seems much more sane.

maritaria:
Having looked at it some more; the section of code that this belongs is a bit more complicated than just reading out the env vars. It seems to be a more general utility that sees some re-use when needing to execute some shell scripts.

The executable ran is either `/usr/bin/env` or a compatible program that prints the environment variables split by `\0`.
The stdout, stderr are collected during the shell execution but not used in the output of the `readShellEnv` function.

I think that we can change the implementation to read stdout of the env program directly, still collect stderr into the log but assume that the regular program.

maritaria:
Another idea, which would be much simpler, but doesn't work for batch is to launch `/bin/sh` from within the user shell and then reliably execute a shell script. This would work because as you launch `sh`, the environment is already correctly configured.

Side note; the reliance of the code on `posix` shell parsing is a little odd given that `env` is a posix program and has a known output format.
OR, if there is going to be a native c program for extracting the environment anyway, why not use that always? Then you may even avoid some compat problems with windows; but I'm no posix platform interop expert :/

trespasserw:
Sorry this took so long. The ability to add support was dependent on a task of refactoring messy environment reading code.

PR #2645, opened 2024-01-25
[kotlin] K2 J2K: Port JavaMapForEachInspection to BuiltinMembersConversion

ermattt:
NB: The test `testJava8MapForEachWithFullJdk` still fails, but I thought I'd put up a PR to get feedback before I dig deeper and/or start moving tests around.

I was able to switch the expression to use the more Kotlin-y curly braces (i.e. from `map.forEach((key, value) -> foo(key, value))` to `map.forEach { key: String?, value: String? -> foo(key, value) }`), but *not* to use the destructuring (i.e. `key: String?, value: String?` vs. `(key: String?, value: String?)`).

If I understand correctly, I *think* that change might be impossible without 2 refactorings to JKTree classes:
1. Add a JKTree class that's analogous to `KtDestructuringDeclaration`
2. Modify `JKParameter` to also have a `JKDestructuringDeclaration` property (or something similar)

Is there another way to do this that I'm missing?

```
internal class Test {
    fun test(map: HashMap<String?, String?>) {
        map.forEach { key: String?, value: String? -> foo(key, value) }
    }

    fun foo(key: String?, value: String?) {
    }
}
```
@darthorimar @abelkov @jocelynluizzi13

### DIFF
--- a/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
+++ b/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
@@ -99,7 +99,6 @@ private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
     RemoveExplicitPropertyTypeProcessing(),
     RemoveRedundantNullabilityProcessing(),
     inspectionBasedProcessing(FoldInitializerAndIfToElvisInspection(), writeActionNeeded = false),
-    inspectionBasedProcessing(JavaMapForEachInspection()),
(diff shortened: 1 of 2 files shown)


abelkov:
Thank you.

AFAIU your current `withArgumentsProvider` block doesn't really do anything: it just creates another lambda expression with the same parameters. It looks like we indeed need a new JK tree element to represent a (restricted form of) `KtDestructuringDeclaration` specifically to support this case.

> I was able to switch the expression to use the more Kotlin-y curly braces

Well, the curly braces are printed as usual in `JKCodeBuilder.Visitor#visitLambdaExpressionRaw`. If you only remove the post-processing, without adding any other code, the test fails in exactly the same way.

After you add a new JK tree node, you will be able to print it specially (inside parentheses) in `visitLambdaExpressionRaw`.

ermattt:
Replaced by another PR

PR #2648, opened 2024-01-26
[kotlin] J2K: Convert RemoveRedundantModalityModifierProcessing post-processing to a JK Tree printing phase

jocelynluizzi13:
Move RemoveRedundantModalityModifierProcessing to JKTree printing.

Additionally add interface checks in the ConvertGettersAndSettersToPropertyProcessing postprocessing step since this is where some of the properties get created and ABSTRACT is added. This will be migrated during group 2

### DIFF
--- a/plugins/completion-ml-ranking/resources/experiment.json
+++ b/plugins/completion-ml-ranking/resources/experiment.json
@@ -127,7 +127,7 @@
       "id": "java",
       "experimentBucketsCount": 4,
       "includeGroups": [ 13, 22, 24 ],
-      "shouldLogElementFeatures": true
+      "shouldLogElementFeatures": false
     },
     {
       "id": "python",


abelkov:
Please use Git rebase to not introduce redundant merge commits. Roughly speaking, https://patch-diff.githubusercontent.com/raw/JetBrains/intellij-community/pull/2648.patch should be very small and show only the changes actually made by you. This will greatly ease the process of merging PRs. In the end, this PR should consist of exactly one commit with your changes.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -965,4 +964,4 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
             printer.print(label.render())
         }
     }
-}
+}
```
abelkov, line 967: Not needed change

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -64,14 +63,14 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
...
+            val isOpenAndAbstractByDefault = (modifiersListOwner is JKClass && modifiersListOwner.classKind == INTERFACE) ||
                     (modifiersListOwner is JKMethod && modifiersListOwner.parentOfType<JKClass>()?.classKind == INTERFACE)
+                    || (modifiersListOwner is JKField && modifiersListOwner.parentOfType<JKClass>()?.classKind == INTERFACE)
             val hasOverrideModifier = modifiersListOwner
                 .safeAs<JKOtherModifiersOwner>()
                 ?.hasOtherModifier(OVERRIDE) == true
-
             fun Modifier.isRedundant(): Boolean = when {
-                this == OPEN && isOpenByDefault -> true
+                (this == OPEN || this == ABSTRACT) && isOpenAndAbstractByDefault -> true
```
abelkov, line 73: I'm a little confused. We have both "or" and "and" here. Which one is it? Maybe it's better to split this into two branches?

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -64,14 +63,14 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
         }
 
         private fun renderModifiersList(modifiersListOwner: JKModifiersListOwner) {
-            val isOpenByDefault = (modifiersListOwner is JKClass && modifiersListOwner.classKind == INTERFACE) ||
+            val isOpenAndAbstractByDefault = (modifiersListOwner is JKClass && modifiersListOwner.classKind == INTERFACE) ||
                     (modifiersListOwner is JKMethod && modifiersListOwner.parentOfType<JKClass>()?.classKind == INTERFACE)
+                    || (modifiersListOwner is JKField && modifiersListOwner.parentOfType<JKClass>()?.classKind == INTERFACE)
```
abelkov, line 68: The initializer is getting complicated, probably it's worth extracting common subexpressions and/or simplifying in other ways.

abelkov (changes requested):

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
jocelynluizzi13: the interface would be both open and abstract by default, so if it has either modifier we should remove it- i can separate it into two branches if that makes it clearer

PR #2649, opened 2024-01-29
[kotlin] K2 J2K: Port ConvertToIsArrayOfCallFix to a JK conversion

ermattt:
Context from the spreadsheet:
"KTIJ-26655. Looks like ConvertToIsArrayOfCallFix can be just ported to a JK conversion"

NB: The test `CheckForInstanceOfErasedArrayType` still fails, but I thought I'd put up a PR to get feedback before I dig deeper. The test fails because the parameters are still treated as non-nullable. Should this step remain as a postprocessing, since it seems like nullability analysis needs to happen first?

Contents of the failed test:
```
class J {
    fun testObject(obj: Any): Boolean {
        return obj is Array<*> && obj.isArrayOf<String>()
    }

    fun testArray(array: Array<Any?>): Boolean {
        return array.isArrayOf<String>()
    }
}
```
Expected output:
```
class J {
    fun testObject(obj: Any?): Boolean {
        return obj is Array<*> && obj.isArrayOf<String>()
    }

    fun testArray(array: Array<Any?>?): Boolean {
        return array?.isArrayOf<String>() == true
    }
}
```

cc @darthorimar  @abelkov @jocelynluizzi13

### DIFF
--- a/plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/IsArrayOfCallConversion.kt
+++ b/plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/IsArrayOfCallConversion.kt
@@ -1,0 +1,47 @@
+// Copyright 2000-2022 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.kotlin.nj2k.conversions
+
+import org.jetbrains.kotlin.j2k.Nullability
+import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
+import org.jetbrains.kotlin.nj2k.RecursiveApplicableConversionBase
+import org.jetbrains.kotlin.nj2k.identifier
+import org.jetbrains.kotlin.nj2k.symbols.JKFieldSymbol
+import org.jetbrains.kotlin.nj2k.tree.*
+import org.jetbrains.kotlin.nj2k.types.*
+import org.jetbrains.kotlin.nj2k.unboxFieldReference
+
+internal class IsArrayOfCallConversion(override val context: NewJ2kConverterContext) :
+    RecursiveApplicableConversionBase(context) {
+
+    override fun applyToElement(element: JKTreeElement): JKTreeElement {
+        if (element !is JKIsExpression) return recurse(element)
+        if (!element.type.type.isArrayType()) return recurse(element)
+        val operand = element.expression
+        val operandType = operand.calculateType(typeFactory) ?: return recurse(element)
+        val type = element::type.detached().type
+
+        val isArrayOfExpression = JKQualifiedExpression(
+            operand.detached(element), JKCallExpressionImpl(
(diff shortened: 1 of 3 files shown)


plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/IsArrayOfCallConversion.kt
```diff
@@ -0,0 +1,47 @@
...
+        if (element !is JKIsExpression) return recurse(element)
+        if (!element.type.type.isArrayType()) return recurse(element)
+        val operand = element.expression
+        val operandType = operand.calculateType(typeFactory) ?: return recurse(element)
+        val type = element::type.detached().type
+
+        val isArrayOfExpression = JKQualifiedExpression(
+            operand.detached(element), JKCallExpressionImpl(
+                symbolProvider.provideMethodSymbol("kotlin.isArrayOf"),
+                JKArgumentList(), JKTypeArgumentList(type.arrayInnerType()!!)
```
abelkov, line 26: It's better to avoid `!!`, if possible. I'd introduce a variable with an early return.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/IsArrayOfCallConversion.kt
```diff
@@ -0,0 +1,47 @@
...
+                        symbolProvider.provideClassSymbol("kotlin.Array"),
+                        listOf(JKStarProjectionTypeImpl),
+                        Nullability.NotNull
+                    ).asTypeElement()
+                ),
+                isArrayOfExpression,
+                JKKtOperatorImpl(JKOperatorToken.ANDAND, typeFactory.types.boolean)
+            )
+        }
+        return recurse(replacement)
```
abelkov, line 45: Please add a test case with some comments and make sure to preserve them with `withFormattingFrom` call.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/IsArrayOfCallConversion.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
```
abelkov, line 47: I don't know if the test differences in nullability are significant or kind of accidental. I think it makes sense to try to replicate the behavior of the original quick fix with regards to nullability (the `== true` part) and negation. And then hope that the future J2K nullability refactoring will provide the proper nullability of all types for this conversion.

So, I propose that the conversion should assume that types already have the correct nullability and properly handle both cases: nullable and not-null.

See the original commit with inspection and all relevant tests: https://github.com/JetBrains/intellij-community/commit/d1f935490a3d3bf3295ec75d0c06d0d8ffefe725

abelkov (changes requested):

PR #2650, opened 2024-01-29
[kotlin] J2K: K2 Convert RemoveRedundantModalityModifierProcessing post-processing to a JK Tree printing phase

jocelynluizzi13:
Move RemoveRedundantModalityModifierProcessing to JKTree printing.

Additionally add interface checks in the ConvertGettersAndSettersToPropertyProcessing postprocessing step since this is where some of the properties get created and ABSTRACT is added. This will be migrated during group 2


jocelynluizzi13:
@darthorimar  can I get a review of this when you get a chance?

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -64,14 +63,17 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
         }
 
         private fun renderModifiersList(modifiersListOwner: JKModifiersListOwner) {
-            val isOpenByDefault = (modifiersListOwner is JKClass && modifiersListOwner.classKind == INTERFACE) ||
-                    (modifiersListOwner is JKMethod && modifiersListOwner.parentOfType<JKClass>()?.classKind == INTERFACE)
+            var isOpenAndAbstractByDefault = false
```
abelkov, line 66: In general, it's better to avoid `var`s. This code doesn't look so complicated as to justify a `var` (probably).

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -64,14 +63,17 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
...
+            if (modifiersListOwner is JKElement) {
```
abelkov, line 67: If you need a smart cast to a `JKElement` to have the code work both for `JKMethod` and `JKField`, maybe you could preserve the structure of the existing code and just change `modifiersListOwner is JKMethod` to `modifiersListOwner is JKDeclaration`?

abelkov (changes requested):
Sorry for the delay. In general, the solution looks OK to me, but maybe the code can be improved a little.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
jocelynluizzi13: made the change to `JKDeclaration`!

abelkov (approved):
LGTM. cc @darthorimar

darthorimar (approved):

PR #2652, opened 2024-01-31
Add Arena#ofAuto and Arena#global to list of ignored AutoCloseable producers

SirYwell:
Both methods return implementations that will throw UOEs when calling `close()`.
IntelliJ should therefore not warn when those methods are used without TWR. (It might make sense to instead warn if used *with* TWR?)

Java 22: https://cr.openjdk.org/~mcimadamore/jdk/FFM_22_PR/javadoc/java.base/java/lang/foreign/Arena.html
Java 21 (Preview): https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/foreign/Arena.html


sirgl:
Thank you for the contribution. Your PR is getting merged.

PR #2655, opened 2024-02-01
Keep Balloon open when hovered, even if it has no buttons

albertvaka:
Removes a condition so the code that keeps the Balloon visible when hovered is always executed.

This is the expected behavior for floating UI components, so the user can keep it open if they are reading (or deciding whether to interact with) the contents.

Previously, this was conditioned to the Balloon being created with a close button (adding other buttons manually, or using a text with clickable links wouldn't trigger this behavior either).


albertvaka:
@alexander-lobas ping?

albertvaka:
Reping @bulenkov @alexander-lobas

albertvaka:
Pingy pongy ping @bulenkov @alexander-lobas ?

albertvaka:
Closing because of being utterly ignored.

PR #2657, opened 2024-02-03
Github Plugin: Disable comment button when text is empty

purofle:


plugins/github/src/org/jetbrains/plugins/github/pullrequest/ui/review/GHPRSubmitReviewPopup.kt
```diff
@@ -44,6 +44,7 @@ internal object GHPRSubmitReviewPopup : CodeReviewSubmitPopupHandler<GHPRSubmitR
         toolTipText = GithubBundle.message("pull.request.review.submit.comment.description")
       }.apply {
         bindDisabledIn(cs, vm.isBusy)
+        bindDisabledIn(cs, vm.text.map { it.isBlank() })
```
iasemenov, line 47: This will disable the previous condition and also submission should not be blocked if there are draft comments, since the user may want to submit whose without making a new comment.
Could you describe your concern with the current functionality?

iasemenov (changes requested):

PR #2658, opened 2024-02-03
IDEA-345026: Fix plugin.xml recognition when comment contains angle brackets

RedNesto:


YannCebron (approved):
Thanks for the report and the patch. I've augmented your patch with some additional required changes and it should land in master shortly.

PR #2666, opened 2024-02-07
[kotlin] J2K: K2 Convert WRONG_MODIFIER_TARGET Post Processing to InternalDeclarationConversion and Printing Phase

jocelynluizzi13:
the WRONG_MODIFIER_TARGET diagnostic consists of incorrectly having "inner" on objects and "internal" on local classes

fix the incorrect inner on objects in the JK Printing phase by checking whether or not we are in an outer object
fix the incorrect internal on local classes in this phase as well

For copy/pasting classes that become local, we need to make changes to `InternalDeclarationConversion`. This should default to public if there is a class in the PSI tree but not in the element's parents since this indicates something is changing in the visibility structure of the code upon copy/paste.


abelkov:
Thank you. Looks reasonable, but I need to do some additional testing around the copy-paste bit.

Before we proceed however, I want to clarify: did you try to investigate any additional cases when the `WRONG_MODIFIER_TARGET` diagnostic may appear? Here is the relevant code:

https://github.com/JetBrains/kotlin/blob/e12e78d96b77ae96d5639535b92abec017c7b789/compiler/frontend/src/org/jetbrains/kotlin/resolve/ModifiersChecker.kt#L122 - the place where diagnostic is reported from.

https://github.com/JetBrains/kotlin/tree/master/compiler/testData - you can search this directory for `WRONG_MODIFIER_TARGET` string to find various examples.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -67,6 +69,9 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
                 (it is JKClass && it.isInterface()) ||
                         (it is JKDeclaration && it.parentOfType<JKClass>()?.isInterface() == true)
             }
+            val isOuterObject = modifiersListOwner is JKClass && modifiersListOwner.classKind == OBJECT && modifiersListOwner.parents()
```
abelkov, line 72: Observation: this function is becoming complicated. I think we shouldn't have complicated logic in the code printer. Maybe it makes more sense to move this whole redundancy check to a conversion (for example, to an existing `RemoveWrongExtraModifiersForSingleFunctionsConversion`).

abelkov (changes requested):

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -67,6 +69,9 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
...
+                .none { parent -> parent is JKClass && parent.classKind == OBJECT }
+            val isLocalClass = modifiersListOwner is JKClass && modifiersListOwner.isLocalClass()
```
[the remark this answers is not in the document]
jocelynluizzi13: You're right the `isLocalClass` is actually unnecessary, removing

jocelynluizzi13:
Made changes to clean up the Printing phase and place the modifier modifier into "removewrongmodifiers"

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/InternalDeclarationConversion.kt
```diff
@@ -13,15 +13,21 @@ internal class InternalDeclarationConversion(context: NewJ2kConverterContext) :
...
 
-        val containingClass = element.parentOfType<JKClass>()
-        val containingClassKind = containingClass?.classKind ?: element.psi<PsiMember>()?.containingClass?.classKind?.toJk()
+        val containingClass: JKClass? = element.parentOfType<JKClass>()
+        val psiTreeContainingClass = element.psi<PsiMember>()?.containingClass
+        val containingClassKind = containingClass?.classKind ?: psiTreeContainingClass?.classKind?.toJk()
 
         val containingClassVisibility = containingClass?.visibility
             ?: element.psi<PsiMember>()
                 ?.containingClass
```
abelkov, line 22: Can be replaced with `psiTreeContainingClass` now

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
```diff
@@ -6,14 +6,23 @@ import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
...
+internal class RemoveWrongExtraModifiersConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
     override fun applyToElement(element: JKTreeElement): JKTreeElement {
         if (element !is JKOtherModifiersOwner) return recurse(element)
         if (element.parentOfType<JKClass>() == null) {
+            // single function --> remove STATIC
             element.elementByModifier(OtherModifier.STATIC)?.also { modifierElement ->
                 element.otherModifierElements -= modifierElement
             }
         }
+        else if ( element is JKClass && element.classKind == JKClass.ClassKind.OBJECT && element.parents()
```
abelkov, line 18: This whole section needs to be reformatted

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
```diff
@@ -6,14 +6,23 @@ import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
 import org.jetbrains.kotlin.nj2k.RecursiveConversion
 import org.jetbrains.kotlin.nj2k.tree.*
 
-internal class RemoveWrongExtraModifiersForSingleFunctionsConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
+internal class RemoveWrongExtraModifiersConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
     override fun applyToElement(element: JKTreeElement): JKTreeElement {
         if (element !is JKOtherModifiersOwner) return recurse(element)
         if (element.parentOfType<JKClass>() == null) {
+            // single function --> remove STATIC
```
abelkov, line 13: Is it really applicable to functions only? If yes, it's better to add an explicit check in the code.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
abelkov: Condition is very complicated, it's better to extract the second part into a variable.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
```diff
@@ -6,14 +6,23 @@ import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
...
             element.elementByModifier(OtherModifier.STATIC)?.also { modifierElement ->
                 element.otherModifierElements -= modifierElement
             }
         }
+        else if ( element is JKClass && element.classKind == JKClass.ClassKind.OBJECT && element.parents()
+                .none { parent -> parent is JKClass && parent.classKind == JKClass.ClassKind.OBJECT }) {
+            // Object not nested in another object --> remove INNER
+            element.elementByModifier(OtherModifier.INNER)?.also { modifierElement ->
```
abelkov, line 21: You need to determine which modifier to remove. The removal part itself is duplicated code, can be refactored.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
```diff
@@ -6,14 +6,23 @@ import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
...
+                element.otherModifierElements -= modifierElement
+            }
+        }
         return recurse(element)
     }
+
```
abelkov, line 27: Redundant

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
```diff
@@ -6,14 +6,23 @@ import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
 import org.jetbrains.kotlin.nj2k.RecursiveConversion
 import org.jetbrains.kotlin.nj2k.tree.*
 
-internal class RemoveWrongExtraModifiersForSingleFunctionsConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
+internal class RemoveWrongExtraModifiersConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
```
abelkov, line 9: I guess we should move here all logic from the code builder about redundant modifiers. I can do it later, it's not necessary for this PR.

abelkov (changes requested):

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
```diff
@@ -0,0 +1,25 @@
...
+package org.jetbrains.kotlin.nj2k.conversions
+
+import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
+import org.jetbrains.kotlin.nj2k.RecursiveConversion
+import org.jetbrains.kotlin.nj2k.tree.*
+
+internal class RemoveWrongExtraModifiersConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
+    override fun applyToElement(element: JKTreeElement): JKTreeElement {
+        if (element !is JKOtherModifiersOwner) return recurse(element)
+        val objectClassParentExists = element.parents().none { parent -> parent is JKClass && parent.classKind == JKClass.ClassKind.OBJECT }
```
abelkov, line 12: Looks like it should be "doesn't exist" instead of "exists". I will clean this up myself along with some other things possibly, to speed up the review.

abelkov (approved):
cc @darthorimar

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
abelkov: AFAIU objects always can't be `inner`, so additional check of parents is redundant, I will fix this.

plugins/kotlin/j2k/k1.new/src/org/jetbrains/kotlin/nj2k/conversions/RemoveWrongExtraModifiersConversion.kt
abelkov: On the other hand, the cases are a little bit different. In this conversion, we can just delete "other" modifiers. But it doesn't work for modality and visibility: we can't really delete them, we can only "skip" printing them. So, I think I will just extract the logic out of the code builder into a helper function.

PR #2667, opened 2024-02-08
KTIJ-28762 False negative 'Redundant explicit this' inspection with overloaded extension functions of Int

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-28762


abelkov (changes requested):
Thank you.

This change also fixes an [unrelated case](https://youtrack.jetbrains.com/issue/KTIJ-18546/Unexpected-behaviour-change-in-new-J2K#focus=Comments-27-7267408.0-0). Please add a test for this.

Also, please run this inspection on intellij-community repo before and after the change and make sure there are no new false positives.

PR #1626, opened 2021-07-08
PY-49650 Fix for static member function skeletons

HunterTracer:
Fix for static member function skeletons. See https://youtrack.jetbrains.com/issue/PY-49650 and https://youtrack.jetbrains.com/issue/PY-49144 for details.

PR #2668, opened 2024-02-08
FIxed bug of generating bad native py module skeletons

swigger:
I recently developed a native python module and found pycharm faield to generate the good sketelon for it. 
However, it can handle classmethod well.
I addressed the problem and found its an issue in module_redecalarator.py.
So I created the pull request here. the code here has the same bug as pycharm, hope pycharm is using code from here.


east825:
Hi! Thanks for investigating the issue and sorry for completely overlooking about this PR. This problem was reported in [PY-49650](https://youtrack.jetbrains.com/issue/PY-49650) and happened to be addressed in an earlier PR, namely #1626. That fix is on its way to 2025.2.1.

PR #2669, opened 2024-02-09
IDEA-332969 - Fix source attachments when a dependency has several matching source jars

pantherdd:
Fixes https://youtrack.jetbrains.com/issue/IDEA-332969

> The issue happens when we're trying to import a Gradle project into IntelliJ, where one of the dependencies has multiple JARs. So, a single dependency coordinate (e.g. `foo:bar:1.2.3`) resolves to multiple code JARs, which is possible with an Ivy repository. In this case, IntelliJ only attaches a single source JAR, even if all code JARs have matching source JARs.

### DIFF
--- a/plugins/gradle/testSources/org/jetbrains/plugins/gradle/importing/GradleDependenciesImportingTest.java
+++ b/plugins/gradle/testSources/org/jetbrains/plugins/gradle/importing/GradleDependenciesImportingTest.java
@@ -1734,42 +1734,53 @@ public class GradleDependenciesImportingTest extends GradleImportingTestCase {
   }
 
   @Test
-  public void testSourcesForDependencyWithMultipleArtifactsWithIvyLayout() throws Exception {
+  public void testJavadocAndSourcesForDependencyWithMultipleArtifactsWithIvyLayout() throws Exception {
     GradleSettings.getInstance(myProject).setDownloadSources(true);
-    // IVY_ARTIFACT_PATTERN = "[organisation]/[module]/[revision]/[type]s/[artifact](.[ext])"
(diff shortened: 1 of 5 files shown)


pantherdd:
@cobaku, can you please run my changes through a full CI build, whether it causes any regression in other tests? I couldn't find a way to do that.

pantherdd:
I improved the tests that execute actual Gradle imports, now they all fail without my production code changes.
* In `testJavadocAndSourcesForDependencyWithMultipleArtifacts`, I only had to change the order of the artifacts in the Ivy descriptor
* In `testJavadocAndSourcesForDependencyWithMultipleArtifactsWithIvyLayout`, the important thing is that now there is an `<artifact name="depArtifact" .../>` entry whose name is the same as the `<info module="depArtifact" .../>` value (this situation is quite likely to happen in practice)

These will all be fixed by the second commit in the PR.

pantherdd:
Unfortunately I have to revert my "artifact reordering" change to `testJavadocAndSourcesForDependencyWithMultipleArtifacts`, because if `depArtifact-api-1.0-SNAPSHOT-sources.jar` is moved up to be the first artifact in the list, the test fails even after my fix is applied. It's strange because my `chooseAuxiliaryArtifactFile` algorithm returns the correct file, `depArtifact-1.0-SNAPSHOT-sources.jar`, both before and after `importProjectUsingSingeModulePerGradleProject()` call, but after that call something overrides the chosen source JAR with the wrong one. Without the fix, it's the first `assertModuleLibDep` call that fails the test, with the fix it can go further but it still fails at the last `assertModuleLibDep` call.

PR #2674, opened 2024-02-11
Fix identifier highlight in injected languages.

Earthcomputer:
Identifier highlighting was disabled for single-line editors, which accounts for injected languages inside single-line Java string literals. According to git blame, the original intention of this was to disable identifier highlighting in dialog editors, so I switched the check to explicitly check that.

My use case is that I'm developing a plugin which makes use of a `MultiHostInjector` for a DSL spread across multiple Java string literals. It would significantly improve UX if identifier highlighting was possible in this use case. I could not find any clean workaround which doesn't involve modifying IntelliJ in this way. Besides, I consider this a bug and it's nice to get it fixed :)

Before the fix:
image

After the fix:
image

### DIFF
--- a/platform/lang-impl/src/com/intellij/codeInsight/daemon/impl/IdentifierHighlighterPassFactory.java
+++ b/platform/lang-impl/src/com/intellij/codeInsight/daemon/impl/IdentifierHighlighterPassFactory.java
@@ -7,6 +7,7 @@ import com.intellij.openapi.Disposable;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.impl.NonBlockingReadActionImpl;
 import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.editor.ex.EditorEx;
 import com.intellij.openapi.project.DumbService;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Key;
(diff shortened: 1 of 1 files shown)


platform/lang-impl/src/com/intellij/codeInsight/daemon/impl/IdentifierHighlighterPassFactory.java
```diff
@@ -23,7 +24,7 @@ public final class IdentifierHighlighterPassFactory {
   public IdentifierHighlighterPass createHighlightingPass(@NotNull PsiFile file,
                                                           @NotNull Editor editor,
                                                           @NotNull TextRange visibleRange) {
-    if (!editor.isOneLineMode() &&
+    if (!((EditorEx) editor).isEmbeddedIntoDialogWrapper() &&
```
BasLeijdekkers, line 27: Now  multi-line editors contained in a dialog don't get identifier highlighting anymore. Perhaps the old and new check should be combined?

BasLeijdekkers:
Hi @Earthcomputer ,
Thank you for your contribution. Could you create a YouTrack issue for this? That way it becomes part of the release notes and such.

BasLeijdekkers (changes requested):

Earthcomputer:
Hi, sorry for the delay, I've been away. I can take a look this evening.

Earthcomputer:
I addressed the review, and made a YouTrack issue: https://youtrack.jetbrains.com/issue/IDEA-346533/Identifier-highlighting-doesnt-work-in-single-line-language-injections

BasLeijdekkers:
Hello, sorry for the delay. Could you squash the two commits into one? That makes it easier to apply the changes to our internal repository.

Earthcomputer:
@BasLeijdekkers done

BasLeijdekkers (approved):
I have applied your fix with some small changes. It should appear on GitHub within a couple of days.

PR #2675, opened 2024-02-12
IDEA-347543 - Fix Incorrect File Selection in Merge Conflict Window Directory Tree by Preserving Expanded State

Khukhunashvili:
### Problem:

When users collapse directories in the Merge Conflict window pop up and a refresh occurs (triggered by updateModelFromFiles(), for example when I open the file for merge and close it immediately), the previously selected file could be reselected incorrectly. 

This issue stems from `table.selectionModel.minSelectionIndex` not accounting for the visibility (expanded/collapsed state) of files in the directory tree. Consequently, when the updateTree() function expands all directories during a refresh, the indices of files change, leading to a mismatch between the intended and actual selections.

Here is the visualization of the problem:
https://imgur.com/gallery/sjJe84Y

### Solution:
To address this issue, I implemented a strategy to preserve the user's context in the directory tree across updates. The solution involves the following steps:

1. Save Expanded Paths
2. After `updateTree()` function is invoked, collapse the entire tree.
3. Restore expanded paths

This approach ensures that the directory tree's structure post-update remains identical to its state prior to the update. Consequently, the index saved in `selIndex` remains consistent.

### DIFF
--- a/platform/platform-api/src/com/intellij/util/ui/tree/TreeUtil.java
+++ b/platform/platform-api/src/com/intellij/util/ui/tree/TreeUtil.java
@@ -21,6 +21,7 @@ import com.intellij.ui.tree.DelegatingEdtBgtTreeVisitor;
 import com.intellij.ui.tree.TreeVisitor;
 import com.intellij.ui.treeStructure.CachingTreePath;
 import com.intellij.ui.treeStructure.Tree;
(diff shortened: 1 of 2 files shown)


AMPivovarov:
The issue shall be fixed in https://github.com/JetBrains/intellij-community/commit/31fe4188d5f0d728d2c5a5839bb7389181c87f49

PR #2677, opened 2024-02-18
Gradle Sync cleanup: Move dependency resolver test to same package as the tested class

pantherdd:
[Gradle|Sync] cleanup: moved dependency resolver test to dependencyModel package
* This is a follow-up to commit c4be347449156b43b1fb10b7b1b671ea4538525a that moved the resolver class itself, and allows reducing the visibility of method `chooseAuxiliaryArtifactFile` to package private again.

@HackerMadCat , please review.


pantherdd:
Looks like this has already been merged: https://github.com/JetBrains/intellij-community/commit/2c6b68fbe1f49f5ab097c9c5cb18e798b08acbc3

Thanks :)

PR #2678, opened 2024-02-18
[kotlin] J2K: K2 Port SimplifyAssertNotNullInspection to JKTree processing

ermattt:
Context from spreadsheet:
"Try to move to conversions.JavaStatementConversion#convertAssert"

There are no functional changes in the first commit--I just expanded the tests here to include cases covered by the [inspection's tests](https://github.com/JetBrains/intellij-community/tree/45d0890f92f15bd9bf7c181a939affff166da55e/plugins/kotlin/idea/tests/testData/inspectionsLocal/simplifyAssertNotNull).

The second commit contains the actual porting plus--as the tests indicate--two changes in behavior compared to the original inspection:
1. Better preservation of the original spacing 
2. Replacing `assert` with `checkNotNull` instead of `!!` (more on this choice below)

I actually feel pretty strongly that it's preferable to replace `assert(s != null)` with `checkNotNull(s)` rather than `s!!`, though I may be missing some important details. My reasoning is that `assert` and `checkNotNull` are both strong indications that a previous author wanted to make a condition very explicit, whereas `!!` feels more like a stopgap that crops up when converting Java code with wonky nullability. From that perspective, `checkNotNull` feels like a more natural replacement. (Admittedly I'm also biased in favor of the `checkNotNull` approach because adding JKTree support for `!!` expressions would be more work.)

### DIFF
--- a/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
+++ b/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
@@ -115,7 +115,6 @@ private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
         it.getExplicitLabelComment() == null
     },
     DestructureForLoopParameterProcessing(),
-    inspectionBasedProcessing(SimplifyAssertNotNullInspection()),
     LiftReturnInspectionBasedProcessing(),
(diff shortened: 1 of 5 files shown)


abelkov (approved):
Thank you, I will merge this with minor cleanup.

I agree that since Kotlin 1.3 introduced contracts, it makes more sense to use `checkNotNull` instead of `!!`. I filed an issue to update the inspection too: https://youtrack.jetbrains.com/issue/KTIJ-28848.

PR #2684, opened 2024-02-21
CertificateManager: set default SSLSocketFactory

gharrma:
The problem:

  If an HTTPS connection is created before CertificateManager.<init>
  calls SSLContext.setDefault(sslContext), then HttpsURLConnection
  will cache the wrong value for its field `defaultSSLSocketFactory`,
  and thus all future HTTPS connections will use the wrong
  SSLSocketFactory by default. This can lead to issues such as
  https://issuetracker.google.com/319147212.

The solution:

  Explicitly call setDefaultSSLSocketFactory() to ensure any
  previously cached value is overwritten.

---

Issue link: https://issuetracker.google.com/319147212

Let me know what you think. Also, concerned that this initialization code
happens within `coroutineScope.launch` and thus has no guarantees of
finishing before the IDE starts making HTTPS requests. Let me know if you
would like to YouTrack issue filed for further discussion. Thanks!

### DIFF
--- a/platform/platform-api/src/com/intellij/util/net/ssl/CertificateManager.kt
+++ b/platform/platform-api/src/com/intellij/util/net/ssl/CertificateManager.kt
@@ -29,6 +29,7 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicReference
 import javax.crypto.BadPaddingException
+import javax.net.ssl.HttpsURLConnection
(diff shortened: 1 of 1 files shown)


paksv:
related commit: https://github.com/JetBrains/intellij-community/commit/7bc2c622b1534531d78694dbc1e7fe68c9540d01
related ticket: https://youtrack.jetbrains.com/issue/IJPL-171446/Unable-to-find-valid-certification-path-to-requested-target-exception-in-Settings-Sync-when-proxy-is-used

PR #2687, opened 2024-02-26
[kotlin] J2K: K2  Remove CAST_NEVER_SUCCEEDS Post Processing 

jocelynluizzi13:
`CAST_NEVER_SUCCEEDS` and `ReplacePrimitiveCastWithNumberConversionFix` are redundant and no longer affect anything because the number casting is done in other steps

Remove them from post-processing so that we don't have to convert it over to K2


jocelynluizzi13:
Question— I'm not sure how to regenerate tests so that `K1QuickFixTestGenerated`, since the tests related to these cases need to be removed

abelkov:
Classes from `org.jetbrains.kotlin.idea.quickfix` are Kotlin IDE plugin quickfixes, they are completely unrelated to J2K and it is not our goal to remove them. If it is redundant for the J2K, we should just remove the J2K post-processing registration (one line). I will do that.

You can see the quickfix in action on the following code:

```
fun foo(c: Char) {
    val a = c as Short // Replace cast with call to 'toShort()'
}
```

To regenerate all Kotlin tests (including the J2K ones), run the "Generate Kotlin Tests (K1 + K2)" run configuration.

PR #2688, opened 2024-02-27
IDEA-346717: Decompiler fails to decompile switch statements containing static calls

h4ckm310n:
[IDEA-346717](https://youtrack.jetbrains.com/issue/IDEA-346717/Decompiler-fails-to-decompile-switch-statements-containing-static-method-calls)

When a switch statement contains a static call like **switch(Cls.foo())**, the decompiler throws java.lang.NullPointerException.

The field **instance** in InvocationExprent is null when the method call is static (InvocationExprent.java:121):
```
else if (opcode == CodeConstants.opc_invokestatic) {
  isStatic = true;
}
else {
  instance = stack.pop();
}
```

However, the decompiler does not check whether **instance** is null before getting its field **type** (SwitchHelper.java:427):
```
if (switchSelector.getInstance().type != Exprent.EXPRENT_VAR) return null;
```

To fix this,  it is neccessary to check whether switchSelector.getInstance() is null first.

### DIFF
--- a/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
+++ b/plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
@@ -424,7 +424,7 @@ public final class SwitchHelper {
       @Override
       public SwitchOnStringCandidate recognize(@NotNull SwitchStatement firstSwitch,
                                                @NotNull InvocationExprent switchSelector) {
-        if (switchSelector.getInstance().type != Exprent.EXPRENT_VAR) return null;
+        if (switchSelector.getInstance() == null || switchSelector.getInstance().type != Exprent.EXPRENT_VAR) return null;
         if (!switchSelector.isInstanceCall(ClassNameConstants.JAVA_LANG_STRING, "hashCode", 0)) return null;
 
         Set<Object> realCaseValueHashCodes = findRealCaseValuesHashCodes(firstSwitch);


plugins/java-decompiler/engine/src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
RooftopJoe, line 1: I'm sorry but shouldn't the condition instead be `switchSelector.getInstance() == null || switchSelector.getInstance().type != Exprent.EXPRENT_VAR`? I didn't even check the code but that immediately looks like a red flag to me because the whole point seems to be to return null when you don't have the same type.

h4ckm310n: Thank you for your revision. To check `type`, your condition is right. However, my condition also works in this case, because in the next line, there is another `if` statement to check whether the switch statement contains a `String.hashCode()` virtual call.

This method (`public SwitchOnStringCandidate recognize()`) is used to handle a switch statement containing a String, looks like `switch (str.hashCode())` in the bytecode. `hashCode()` is a virtual call, so `getInstance()` is always not null is this case, and the result of the condition depends on `type`. In the opposite, if `getInstance()` is null, the invocation is a static call instead of a virtual call, this makes the next `if` statement return null.

In conclusion, if `getInstance()` is not null and `type` is not the same (as in the original code), this method returns null; if `getInstance()` is null, this method also returns null in the next `if` statement. But I think it would be better to apply your revision instead.

h4ckm310n:
Update: Changed 
`switchSelector.getInstance() != null && switchSelector.getInstance().type != Exprent.EXPRENT_VAR` 
to 
`switchSelector.getInstance() == null || switchSelector.getInstance().type != Exprent.EXPRENT_VAR`

PR #2690, opened 2024-02-28
[kotlin] J2K: K2 Replace Getters in JK Tree Stage Instead of Postprocessing

jocelynluizzi13:
this diff is designed as a first step to replace `ReplaceGetOrSetInspection` in postprocessing by moving it to the Psi->JKTree and printing stage.

Currently, this diff only changes the "get" part of the post-processing but I plan to employ a similar strategy for "set"

There are a couple of issues that still need to be addressed for this diff to work:
1. Nullability— We are now marking parameters as Nullable when previously we would make them NonNull because they were dereferenced. We need to make a change in nullability to make brackets consistent with dereferences
2. Enums / Casting— In cases where we add `toTypedArray()`, we should not add `toTypedArray()` if we are accessing the enum. Right now we are adding it extraneously. This is _probably_ a similar issue as the dereferences above

### DIFF
--- a/platform/platform-impl/src/com/intellij/platform/ide/bootstrap/DirectoryLock.java
+++ b/platform/platform-impl/src/com/intellij/platform/ide/bootstrap/DirectoryLock.java
@@ -62,6 +62,8 @@ final class DirectoryLock {
 
   private static final Logger LOG = getLogger();
   private static final AtomicInteger COUNT = new AtomicInteger();  // to ensure redirected port file uniqueness in tests
+  private static final long TIMEOUT_MS = Integer.getInteger("ij.dir.lock.timeout", 5_000);
+  private static final List<String> ACK_PACKET = List.of("<<ACK>>");
 
   private static Logger getLogger() {
     Logger logger = Logger.getInstance(DirectoryLock.class);
@@ -76,7 +78,6 @@ final class DirectoryLock {
   private final @Nullable Path myRedirectedPortFile;
   private final Function<List<String>, CliResult> myProcessor;
 
-  private long myTimeoutMs = Integer.getInteger("ij.dir.lock.timeout", 5_000);
   private volatile @Nullable ServerSocketChannel myServerChannel = null;
 
   DirectoryLock(@NotNull Path configPath, @NotNull Path systemPath, @NotNull Function<List<String>, CliResult> processor) {
@@ -160,10 +161,10 @@ final class DirectoryLock {
       throw new CannotActivateException(e);
     }
 
-    if (LOG.isDebugEnabled()) LOG.debug("Deleting " + myPortFile);
+    if (LOG.isDebugEnabled()) LOG.debug("deleting " + myPortFile);
     Files.deleteIfExists(myPortFile);
     if (myRedirectedPortFile != null) {
-      if (LOG.isDebugEnabled()) LOG.debug("Deleting " + myRedirectedPortFile);
+      if (LOG.isDebugEnabled()) LOG.debug("deleting " + myRedirectedPortFile);
       Files.deleteIfExists(myRedirectedPortFile);
     }
 
@@ -178,7 +179,7 @@ final class DirectoryLock {
     var serverChannel = myServerChannel;
     myServerChannel = null;
     if (serverChannel != null) {
-      if (LOG.isDebugEnabled()) LOG.debug("Cleaning up");
+      if (LOG.isDebugEnabled()) LOG.debug("cleaning up");
       Suppressions.runSuppressing(
         () -> serverChannel.close(),
         () -> {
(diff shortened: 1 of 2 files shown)


plugins/kotlin/code-insight/impl-base/src/org/jetbrains/kotlin/idea/codeinsights/impl/base/inspections/ReplaceGetOrSetInspectionUtils.kt
```diff
@@ -10,9 +10,9 @@ import org.jetbrains.kotlin.util.OperatorNameConventions
 
 object ReplaceGetOrSetInspectionUtils {
     private val operatorNames: Set<Name>
-        get() = setOf(OperatorNameConventions.GET, OperatorNameConventions.SET)
```
abelkov, line 13: The code of inspection should not be changed, we should just not apply it in a post-processing.

plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/inference/nullability/NullabilityDefaultStateProvider.kt
```diff
@@ -6,6 +6,7 @@ import org.jetbrains.kotlin.descriptors.TypeParameterDescriptor
 import org.jetbrains.kotlin.idea.j2k.post.processing.inference.common.*
 import org.jetbrains.kotlin.idea.refactoring.isAbstract
 import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.nj2k.tree.JKArrayGetterExpression
```
abelkov, line 9: I suggest to ignore all nullability issues for now, because it may be challenging to fix the current nullability analysis to recognize array access expressions properly.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -403,9 +402,16 @@ class JavaToJKTreeBuilder(
                         .qualified(qualifier ?: JKThisExpression(JKLabelEmpty()))
                 }
 
-                symbol is JKMethodSymbol ->
-                    JKCallExpressionImpl(symbol, arguments.toJK(), typeArguments)
-                        .qualified(qualifier)
+                symbol is JKMethodSymbol -> {
+                    val fqName = target?.kotlinFqName.toString()
+                    if ((fqName.endsWith("charAt") || fqName.endsWith("get")) && arguments.expressionCount == 1) {
```
abelkov, line 407: To properly convert 'get' method calls to operator index access, we have to know that either:

1. This 'get' is a Java method that will not be converted to Kotlin in this session
2. This is a Kotlin method that has an 'operator' modifier and a proper signature or it overrides such a method (see logic in ReplaceGetOrSetInspection)

And `fqName.endsWith("get")` is not correct: it has to be exactly 'get'.

If these conditions are hard to support, I suggest to concentrate on the array case only (at first).

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -548,12 +554,19 @@ class JavaToJKTreeBuilder(
                 it.withFormattingFrom(this)
             }
 
-        private fun PsiArrayAccessExpression.toJK(): JKExpression =
-            arrayExpression.toJK()
+        private fun PsiArrayAccessExpression.toJK(): JKQualifiedExpression {
+            if (indexExpression?.text != null) {
```
abelkov, line 558: Is this condition important? If an empty index expression is an error in Java, we may ignore it and convert it into a similarly erroneous empty index expression in Kotlin

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/tree/expressions.kt
```diff
@@ -72,6 +72,15 @@ class JKQualifiedExpression(
     override fun accept(visitor: JKVisitor) = visitor.visitQualifiedExpression(this)
 }
 
+class JKArrayGetterExpression(
```
abelkov, line 75: Probably `JKArrayAccessExpression` is a better name, which is consistent with Java and Kotlin PSI names.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/tree/expressions.kt
```diff
@@ -72,6 +72,15 @@ class JKQualifiedExpression(
...
+    receiver: JKExpression,
+    literal: JKExpression,
```
abelkov, line 77: An index expression can be any expression, not only a literal.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/tree/expressions.kt
```diff
@@ -72,6 +72,15 @@ class JKQualifiedExpression(
...
+    override val expressionType: JKType? = null,
+) : JKQualifiedExpression(receiver, literal) {
```
abelkov, line 79: Both `KtArrayAccessExpression` and `PsiArrayAccessExpression` are not qualified expressions.

abelkov:
About issue 2 (enums): please see `EnumSyntheticValuesMethodConversion`, it has to be adapted to your changes.

PR #2693, opened 2024-02-29
On Windows, inline environment variable values case-insensitively

jansorg:
On Windows, when inlining environment variable values,`$PATH$` and `$Path` should both refer to the same value.
On macOS and Linux, names of environment variables are case-sensitive and must be treated as before.

On Windows, this prints `second`:
```bat
set myVALUE=first
set MYvalue=second
echo %myvalue%
```


trespasserw:
Sorry, I apparently forgot to give the PR the right assignee :/

dmitry-avdeev:
Merged, thank you!

PR #2694, opened 2024-03-01
[kotlin] J2K: K2 Replace Getters in JK Tree Stage Instead of Postprocessing

jocelynluizzi13:
this diff is designed as a first step to replace ReplaceGetOrSetInspection in postprocessing by moving it to the Psi->JKTree and printing stage.

see conversation in https://github.com/JetBrains/intellij-community/pull/2690 for more info


abelkov:
The part about arrays is merged in https://github.com/JetBrains/intellij-community/commit/19e9839d656126fee10964b9242a3f10d8498a83 for K2 mode.

PR #2695, opened 2024-03-04
[kotlin] J2K: Begin merging JKCodeBuilder and JKVisitorWithCommentsPrinting (1/7)

ermattt:
The motivation for this stack of PR's comes from trying to do an item in the spreadsheet, "MoveLambdaOutsideParenthesesProcessing," which required modifying the printing phase. The way the printing logic is currently spread out across two different classes, JKCodeBuilder and JKVisitorWithCommentsPrinting, makes it hard to the implement the change. It's also, in my very subjective opinion, a little hard to read. 

I think the outcome of this PR stack is good in terms of brevity (we start out with ~2K lines spread across 2 files, and end up with ~1.2K lines in just one file), but if this refactoring goes against your style guidelines, or has unintended side effects, or is just highly annoying, please feel free to tell me to stop! Alternatively, if you like the change but want me to compress it into one giant commit or PR, I can do that too.

NB: We lose a bunch of newlines in JKVisitorWithCommentsPrinting thanks to my stupidity with the formatter, but since we end up deleting that file later in this stack anyway, I think we can live with it here.


ermattt:
See https://github.com/JetBrains/intellij-community/pull/2697 instead

PR #2696, opened 2024-03-04
[kotlin] J2K: Remove some unnecessary indirection in JKCodeBuilder (2/7)

ermattt:
For functions with only 1 callsite and <=10 lines of code, I merged them into their callsite (or deleted them entirely if they did nothing!). I'm guessing the pattern with the raw methods was originally intended to facilitate more subclassing, but since it looks like that never got used, I think it's slightly better to move those short chunks of code inline.

2 of 7 in the stack. Depends on https://github.com/JetBrains/intellij-community/pull/2695


ermattt:
See https://github.com/JetBrains/intellij-community/pull/2697 instead

PR #2698, opened 2024-03-04
[kotlin] [kotlin] J2K: K2 Remove IndentRawStringIntention and ConvertToRawStringTemplateIntention from Postprocessing

jocelynluizzi13:
now that we can use complex strings in Java there's no need for us to change string literals to binary expressions all the time. The one case where we need to do this still is in DocTag comments. Adjust the stringliteral for doctag comments to not use binary expression and instead use a complex expression.

Also modify the test case to match what we are trying to convert and ensure we're not accidentally converting when we should not still

delete the `IndentRawStringIntention` from postprocessing now that it's no longer necessary


plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/expressions.kt
```diff
@@ -142,6 +142,10 @@ fun JKAnnotationList.annotationByFqName(fqName: String): JKAnnotation? =
 
 fun stringLiteral(content: String, typeFactory: JKTypeFactory): JKExpression {
     val lines = content.split('\n')
+    if (lines.size > 1 && lines.last().isNotEmpty()) {
```
abelkov, line 145: I don't understand the significance of the condition `&& lines.last().isNotEmpty()`. When I comment it out, no tests fail. Is it necessary? Also, can this function be simplified now? It looks like we care about two cases: when there is a single line, and when there are many lines. If some cases are missing from the test data, please add them.

Note: make sure to run K2 J2K tests and remove the "IGNORE_K2" directive where it is not necessary now, as described in the meetings document.

plugins/kotlin/j2k/shared/tests/testData/newJ2k/postProcessing/ConvertConcatenationToRawString.java
```diff
@@ -1,15 +1,21 @@
...
-               "asdfasdf";
-    String s2 = "asdf\n" +
-               "asdfasdf\n";
-    String s3 = "asdf\n" +
-               "nadfadsf\n" +
-               "asdfasdf";
-    String s4 = "asdf\n" +
-               "nadfadsf\n" +
-               "asdfasdf\n";
+    String s1 = """
```
abelkov, line 3: Now you are converting Java multiline strings to Kotlin raw strings.

1. The test data name should be changed.
2. It seems like now this test mostly duplicates `testData/newJ2k/newJavaFeatures/textBlocks". Please migrate important cases there or remove this test entirely.

abelkov (changes requested):

abelkov:
@darthorimar Do you agree with the general approach that we don't need to convert every concatenation to a raw string?

plugins/kotlin/j2k/shared/tests/testData/newJ2k/postProcessing/ConvertConcatenationToRawString.java
abelkov: Actually, it makes more sense to leave the original test data to assert that Java string concatenations are left as concatenations.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/expressions.kt
jocelynluizzi13: Originally I was writing this because in the current test cases, we don't want to convert:
`val x = "hello" + "\n"`
to a raw string. However, on further thought I think it might make sense to convert to a raw string for `@deprecated`  doc tags regardless. I will simplify this section.

abelkov (approved):

jocelynluizzi13:
Updated with regenerated tests and correct `// IGNORE_K2` flags

abelkov:
I think I will merge the part about `stringLiteral`, so it works the same way in K1 and K2.

But let's leave the cleanup post-processings as is in K1, and just not enable them in K2. We could try to support them later as "cleanup only" post-processings in K2.

PR #2701, opened 2024-03-06
[kotlin] J2K: Port "move lambda outside parentheses" processing to printing phase

ermattt:
I'm not sure whether it's possible to port this postprocessing step earlier for 2 reasons:
1. When trying to resolve a method call expressions via `JKSymbol::target`, the returned method definition node has the right name and return type, but not necessarily the right parameters--i.e. it's oblivious to overloaded methods. For example, resolving a call like `checkNotNull(x, "oops")` returns the `KtNamedFunction` node for `checkNotNull` with only one parameter. The same is true when resolving local functions, like the call to `bar` with 2 args in the test below (`bar(f, (i) -> "default g")`).
2. When trying to resolve a constructor call expression via `JKSymbol::target`, we get the whole class definition rather than the constructor definition with matching parameters. With some extra work I could implement very basic resolution for constructors in classes without overloading or inheritance (like in the test below where we call `new Base("Base", () -> "base")`), but that's hardly a replacement for the postprocessing step that currently handles everything.

Is there another way to do symbol resolution on JKTree nodes that I'm missing?

### DIFF
--- a/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
+++ b/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
@@ -82,7 +82,6 @@ private val removeRedundantElementsProcessingGroup = InspectionLikeProcessingGro
 )
 
 private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
-    MoveLambdaOutsideParenthesesProcessing(),
     intentionBasedProcessing(ConvertToStringTemplateIntention(), writeActionNeeded = false) {
         ConvertToStringTemplateIntention.Holder.shouldSuggestToConvert(it)
     },
--- a/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
+++ b/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
@@ -82,6 +82,7 @@ private val removeRedundantElementsProcessingGroup = InspectionLikeProcessingGro
 )
 
 private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
+    MoveLambdaOutsideParenthesesProcessing(),
     intentionBasedProcessing(ConvertToStringTemplateIntention(), writeActionNeeded = false) {
         ConvertToStringTemplateIntention.Holder.shouldSuggestToConvert(it)
     },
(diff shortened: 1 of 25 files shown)


abelkov (changes requested):
You can try to use `jkElement.psi.resolve()` and/or Kotlin Analysis API, instead of JK symbols.

The main problem is that this processing implicitly depends on `RemoveRedundantSamAdaptersProcessing`, because for many lambdas the SAM constructor is unconditionally printed in `visitLambdaExpressionRaw`. To fully migrate `MoveLambdaOutsideParenthesesProcessing` we will have to migrate (some cases of) `RemoveRedundantSamAdaptersProcessing` as well.

Also, you can avoid the problem for now if you just don't try to move the lambda outside parenthesis if `lambdaExpression.functionalType.present()`. That way, we will have at least partial support in JK tree.

ermattt:
So I tried the `jkElement.psi.resolve()` approach and it worked for most cases, but I forgot that JKElements that were created during an intermediate conversion phase won't have a psi element (e.g. a JKElement corresponding to a `checkNotNull` that replaces a Java assert expression). So next I'll see if I can hunt down the places where we're creating a JKElement without a psi node and make sure that those at least have the correct target linked.

abelkov:
Actually, for manually generated elements like `checkNotNull` this is very easy - you already know that the lambda can be definitely moved outside parentheses. So, we should have a flag in the lambda node that it can be moved and set it to `true` when creating the lambda.

ermattt:
> You can try to use `jkElement.psi.resolve()` and/or Kotlin Analysis API, instead of JK symbols.
> 
> The main problem is that this processing implicitly depends on `RemoveRedundantSamAdaptersProcessing`, because for many lambdas the SAM constructor is unconditionally printed in `visitLambdaExpressionRaw`. To fully migrate `MoveLambdaOutsideParenthesesProcessing` we will have to migrate (some cases of) `RemoveRedundantSamAdaptersProcessing` as well.
> 
> Also, you can avoid the problem for now if you just don't try to move the lambda outside parenthesis if `lambdaExpression.functionalType.present()`. That way, we will have at least partial support in JK tree.

Hello @abelkov, happy Friday!
I think I've addressed these comments by 
- adding a field, `canExtractLastArgumentIfLambda` to `JKNewExpression` and `JKCallExpressionImpl` to handle the case where no psi is available to be resolved
- Punting on the`RemoveRedundantSamAdaptersProcessing`-dependency problem by leaving the lambda inside parentheses when an unknown functional type is present (and letting the still-present `MoveLambdaOutsideParenthesesProcessing` handle those cases)
Did I miss anything?

abelkov (approved):
Thank you, this seems to be generally the right direction. I will take this PR from here, because I need to tweak and test it anyway. I will merge it hopefully soon with some changes.

General comments:

1. For such change, a single commit is enough. Especially, if one commit undoes the changes from a previous commit. This is just hard to cherry pick automatically and to review per-commit.

2. Please don't exceed the multiple line size of 140 characters.

3. Code builder should be as simple as possible, all PSI logic should be extracted to some utility file.

ermattt:
@abelkov I was worried that a call to `Map::forEach` might end up losing a (not actually redundant) constructor call. For example, I thought this code:
```
Map<String, Runnable> mapOfRunnables = new HashMap();
map.forEach((key, value) -> new Runnable() {
        @Override
        public void run() {
            System.out.println("hello");
        }
    });
```
might get turned into this:
```
val mapOfRunnables: Map<String?, Runnable?> = HashMap<Any?, Any?>()
        map.forEach { { println("hello") } }
```
But actually I was wrong--it gets turned into this correct code:
```
val mapOfRunnables: Map<String?, Runnable?> = HashMap<Any?, Any?>()
map.forEach { (key: String?, value: String?) -> Runnable { println("hello") } }
```

PR #2702, opened 2024-03-06
[kotlin] K2 J2K: Move VarToValProcessing LocalVarToValInspectionBasedProcessing and fixValToVarDiagnosticBasedProcessing  to JKTree

jocelynluizzi13:
Removing `VarToValProcessing` `LocalVarToValInspectionBasedProcessing` and `fixValToVarDiagnosticBasedProcessing ` from the postprocessor to make it more compatible with K2. 

To do this, we will add mutability logic to the `JKField` class member conversion phase. Now, we will return `IMMUTABLE` for any private vars that do not have writable logic and are not JPA qualified. Note that StubExpressions are allowed to have 1 write, since this would be the initial assignment of the field. 

To check if a field was writable, added more logic to `expressions.kt`:
1. count the number of writable usages as opposed to simply returning "yes" or "no". This allows us to account for JKStubExpression
2. Add `asQualifiedAssignmentFromTarget` expression. This checks for cases such as `this.x = x` and `this.x++` that we were previously missing when using `asAssignmentFromTarget` alone. As such, we explicitly check that we are in either a constructor or a setter when applying this.


jocelynluizzi13:
It makes more sense to combine this with the `fixValToVarDiagnosticBasedProcessing` step given they can be fixed in the same place in the code. 

Switching to draft until I complete that conversion

plugins/kotlin/j2k/shared/tests/testData/newJ2k/field/initializedBothWithInitializerAndConstructor.kt
```diff
@@ -1,5 +1,6 @@
 class TestFieldInitializer(string: String) {
     var string: String = ""
+        private set
```
jocelynluizzi13, line 3: I believe this test case was initially wrong. It seems like `private set` needs to be there else `string` could be set from outside the class

plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
```diff
@@ -131,15 +131,13 @@ private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
...
     removeUnnecessaryNotNullAssertionDiagnosticBasedProcessing,
-    fixValToVarDiagnosticBasedProcessing
 )
 
 private val inferringTypesPostProcessingGroup = NamedPostProcessingGroup(
     KotlinNJ2KServicesBundle.message("processing.step.inferring.types"),
     listOf(
         InspectionLikeProcessingGroup(
             processings = listOf(
-                VarToValProcessing(),
```
abelkov, line 142: Please don't forget that `VarToValProcessing` is also used in K2. You should remove it from K2 and fix the failing tests. Currently tests are failing in K2 and not in K1 because in K1 we have an additional `LocalVarToValInspectionBasedProcessing` that is now doing more work.

jocelynluizzi13:
Modified this commit to fix things you pointed out earlier but also to add fixes for `LocalVarToValInspectionBasedProcessing` since while doing that I noticed additional test cases that should have been accounted for

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/expressions.kt
```diff
@@ -201,10 +213,79 @@ fun JKVariable.hasUsages(scope: JKTreeElement, context: NewJ2kConverterContext):
...
+        while (parent?.parentOfType<JKIfElseStatement>() != null) {
+            parent = parent.parentOfType<JKIfElseStatement>()
+        }
+        parent
+    }.distinct()
+    if (parentStatement.size != 1) {
+        return false
     }
+    // ensure no nested statements
+    return ifElseStatements.all { parentStatement[0]?.psi?.text == it?.psi?.text || parentStatement[0]?.psi?.text?.contains("else " + it?.psi?.text) == true }
```
abelkov, line 250: Again, access to psi just for the text looks suspicious. Is it really necessary?

abelkov (changes requested):
Please take a look at a failing test org.jetbrains.kotlin.nj2k.NewJavaToKotlinCopyPasteConversionTestGenerated#testComments

abelkov:
Also, please recheck testdata of `org.jetbrains.kotlin.idea.inspections.CanBeValInspection`, maybe there is something in there that we miss in J2K.

jocelynluizzi13:
Tested in IDEA with ~10 files, things seem to be working well!

abelkov (changes requested):
Some tests are failing in

TextNewJavaToKotlinCopyPasteConversionTestGenerated
NewJavaToKotlinCopyPasteConversionTestGenerated

Please make sure to run all tests from `j2k/k1.new/tests/test` and `j2k/k2/tests/test`.

PR #2706, opened 2024-03-11
Use decoded form instead of external form for URLs

omega697:
We currently call .toString() on these URLs when converting. This causes the '#' character (amongst others) to be escaped. This causes those URLs to not work properly. Because these were already part of HTML content, we don't need to worry about re-escaping the URL contents.

### DIFF
--- a/java/java-impl/src/com/intellij/codeInsight/javadoc/JavaDocExternalFilter.java
+++ b/java/java-impl/src/com/intellij/codeInsight/javadoc/JavaDocExternalFilter.java
@@ -61,7 +61,7 @@ public class JavaDocExternalFilter extends AbstractExternalFilter {
             String[] parts = href.substring(2).split("/", 2);
             if (parts.length != 2) return href;
             Url relativeUrl = Urls.newUrl(scheme, parts[0], parts[1]);
-            return relativeUrl.toString();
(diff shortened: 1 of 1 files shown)


java/java-impl/src/com/intellij/codeInsight/javadoc/JavaDocExternalFilter.java
```diff
@@ -61,7 +61,7 @@ else if (href.startsWith("//")) {
             String[] parts = href.substring(2).split("/", 2);
             if (parts.length != 2) return href;
             Url relativeUrl = Urls.newUrl(scheme, parts[0], parts[1]);
```
yopox, line 63: Thanks for the patch!

I am writing a test for this method and I was wondering about this logic. `Urls.newUrl("scheme", "autority", "path")` gives `scheme://autoritypath`, is that really what we want here?

`<a href="//example.com/about">` -> `<a href="https://example.comabout">`

Probably we should write `Urls.newUrl(scheme, parts[0], '/' + parts[1]);`?

omega697: Sorry somehow I lost track of this. Thank you for following up on it :)

PR #2708, opened 2024-03-12
More idiomatic use of Kotlin for generating a diagnostic map

yorlov:
Cleanup code


jreznot:
Thank you so much for the effort! Unfortunately, for such places we barely want to make automatic cleanups, as it makes harder reading of Annotate and History

yorlov:
Could you please provide more context? 

I'm not quite understanding how a commit with minor local refactoring could complicate the reading of history.

PR #2709, opened 2024-03-14
[kotlin] J2K: Port replace guard clause with function call inspection postprocessing to jk tree phase

ermattt:
Screenshot 2024-03-13 at 7 56 40 PM
In the spreadsheet, this one is listed as "ReplaceGuardClauseWithFunctionCallInspection." 

I added some more test cases inspired by those under `code-insight/inspections-shared/tests/testData/inspectionsLocal/simplifyNegatedBinaryExpression` (even though it's a different inspection, it's similar enough to provide useful inspiration). 

Unfortunately, I'm still having trouble getting the line breaks formatted correctly in the test (see the attached screenshot). I think it has something to do with the nested blocks that result when replacing the IfElseExpression with a block containing the method call + whatever statements used to live in the else branch. If you have any ideas for a fix I'm all ears, otherwise I'll resume debugging tomorrow.

### DIFF
--- a/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
+++ b/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
@@ -117,7 +117,6 @@ private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
     LiftAssignmentInspectionBasedProcessing(),
     MayBeConstantInspectionBasedProcessing(),
     RemoveForExpressionLoopParameterTypeProcessing(),
-    inspectionBasedProcessing(ReplaceGuardClauseWithFunctionCallInspection()),
     intentionBasedProcessing(ConvertToRawStringTemplateIntention(), additionalChecker = ::shouldConvertToRawString),
     intentionBasedProcessing(IndentRawStringIntention()),
     intentionBasedProcessing(JoinDeclarationAndAssignmentIntention()),
--- a/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/JavaStatementConversion.kt
+++ b/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/JavaStatementConversion.kt
@@ -5,9 +5,7 @@ import org.jetbrains.kotlin.analysis.api.KtAnalysisSession
 import org.jetbrains.kotlin.j2k.Nullability.NotNull
 import org.jetbrains.kotlin.nj2k.*
 import org.jetbrains.kotlin.nj2k.tree.*
-import org.jetbrains.kotlin.nj2k.types.JKJavaDisjunctionType
-import org.jetbrains.kotlin.nj2k.types.isNull
-import org.jetbrains.kotlin.nj2k.types.updateNullability
+import org.jetbrains.kotlin.nj2k.types.*
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
 class JavaStatementConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
@@ -19,6 +17,7 @@ class JavaStatementConversion(context: NewJ2kConverterContext) : RecursiveConver
                 is JKJavaAssertStatement -> convertAssert(element)
                 is JKJavaSynchronizedStatement -> convertSynchronized(element)
                 is JKJavaTryStatement -> convertTry(element)
+                is JKIfElseStatement -> convertGuardStatement(element)
                 else -> element
             }
         )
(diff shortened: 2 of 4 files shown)


plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/JavaStatementConversion.kt
```diff
@@ -167,4 +169,69 @@ class JavaStatementConversion(context: NewJ2kConverterContext) : RecursiveConver
...
+
+        val expressionComparedToNull = ifElseStatement.condition.expressionComparedToNull()
+        val exceptionName = thrownExpression.identifier?.name
+        val correspondingMethodName = when (exceptionName) {
+            "IllegalArgumentException" -> if (expressionComparedToNull != null) "kotlin.requireNotNull" else "kotlin.require"
+            "IllegalStateException" -> if (expressionComparedToNull != null) "kotlin.checkNotNull" else "kotlin.check"
+            else -> return ifElseStatement
+        }
+        val exceptionArgument = thrownExpression.arguments.arguments.firstOrNull()
+        if (exceptionArgument != null && exceptionArgument.value !is JKLiteralExpression) {
```
abelkov, line 192: The original inspection instead checks that argument is of type String. Should we preserve this logic?

1. It may work not only with literal strings
2. It will future-proof the code against potential new overloads of the Java exception constructors.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/JavaStatementConversion.kt
```diff
@@ -167,4 +169,69 @@ class JavaStatementConversion(context: NewJ2kConverterContext) : RecursiveConver
                 tryStatement.catchSections.flatMap(::convertCatchSection)
             )
         ).withFormattingFrom(tryStatement)
+
+    private fun convertGuardStatement(ifElseStatement: JKIfElseStatement): JKStatement {
```
abelkov, line 173: It is not obvious what is a "guard statement". Please document the function.

plugins/kotlin/j2k/shared/tests/testData/newJ2k/postProcessing/GuardClause.java
```diff
@@ -1,8 +1,60 @@
...
+            throw new IllegalArgumentException();
+        }
+
+        if (!b1 && b2) throw new IllegalArgumentException();
+        else {
+            System.out.println(1);
+            System.out.println(2);
+        }
+
+        if (s1.length() < 3) {
```
abelkov, line 24: Please make sure it works correctly with floating point. See

* https://youtrack.jetbrains.com/issue/KTIJ-13841

* `org.jetbrains.kotlin.idea.codeinsight.utils.NegatedBinaryExpressionSimplificationUtils#canBeSimplifiedWithoutChangingSemantics`

plugins/kotlin/j2k/shared/tests/testData/newJ2k/postProcessing/GuardClause.kt
```diff
@@ -1,5 +1,47 @@
...
+        requireNotNull(s1) { "s should not be null" }
+
+        require(b1)
+        println("never mind")
+        // comment above b2
+        require(!b2)
+
+        require(!(!b1 && b2))
+        println(1)
+        println(2)
```
abelkov, line 12: To not insert an extra newline, it seems that you should set `lineBreaksAfter` to zero of the last statement in a block `else` branch.

abelkov (changes requested):

plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
```diff
@@ -117,7 +117,7 @@ private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
     LiftAssignmentInspectionBasedProcessing(),
     MayBeConstantInspectionBasedProcessing(),
     RemoveForExpressionLoopParameterTypeProcessing(),
-    inspectionBasedProcessing(ReplaceGuardClauseWithFunctionCallInspection()),
+    //inspectionBasedProcessing(ReplaceGuardClauseWithFunctionCallInspection()),
```
abelkov, line 120: Remove line

abelkov (changes requested):

plugins/kotlin/j2k/shared/tests/testData/newJ2k/postProcessing/GuardClause.kt
ermattt: Turns out the last statement inside the else block was the culprit!

plugins/kotlin/j2k/shared/tests/testData/newJ2k/postProcessing/GuardClause.kt
```diff
@@ -1,5 +1,52 @@
 class Test {
-    fun test(s: String?) {
-        requireNotNull(s) { "s should not be null" }
+    fun testRequire(s1: String, b1: Boolean, b2: Boolean) {
+        requireNotNull(s1) { "s should not be null" }
```
abelkov, line 3: This is now redundant, because `s` has a non-null type. I met the same issue during my experiments with nullability, so I will probably fix it later (the whole `requireNotNull` call can be removed).

PR #2710, opened 2024-03-14
[junit] IDEA-349194 Can't resolve @EnumSource("names") for @ParameterizedTest in Kotlin

qeeqez:
# JUnit: Can't resolve @EnumSource("names") for @ParameterizedTest in Kotlin

## Fixes [IDEA-349194](https://youtrack.jetbrains.com/issue/IDEA-349194)

Now EnumSourceReference is handled with Uast, so Kotlin is supported

**Bonus:** enum fields now has color (looks like PsiMemberReference was missed)


BartvHelvert:
Hi! Sorry for the delay, could you add a test for this problem?

qeeqez:
> Hi! Sorry for the delay, could you add a test for this problem?

Hi, it was not so straightforward how to create those tests but looks like I've figured out. Pushed!

BartvHelvert:
We already have some tests in `JUnit5ParameterizedReferencesTest` but they seem Java only. I'll try to move those tests into the regular JUnit plugin and merge them with your Kotlin tests. Thanks a lot for your contribution!

PR #2714, opened 2024-03-18
[IDEA-349449] UpdateCheckerService should display results

jrlogsdon:
I think this was meant to be true, which allows the other noisy checks to still not display results but `UpdateCheckerService` to still display results.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/UpdateChecker.kt
+++ b/platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/UpdateChecker.kt
@@ -73,7 +73,7 @@ private class UpdateCheckerHelper(private val coroutineScope: CoroutineScope) {
   /**
    * For scheduled update checks.
    */
-  fun updateAndShowResult(showResults: Boolean = false): ActionCallback {
+  fun updateAndShowResult(showResults: Boolean = true): ActionCallback {
     val callback = ActionCallback()
     coroutineScope.launch(limitedDispatcher) {
       doUpdateAndShowResult(


jreznot:
@jrlogsdon If you need to show results you have to pass true explicitly there. By default we definitely do not want to

jrlogsdon:
There are only two calls call to `updateAndShowResults` that do not override `showResults=false` to therefore not display results. 

1) plugins are failing and a plugin check is needed
2) called from the `UpdateCheckerService` where there is already logic to check it isn't spamming results

I think we want to show the results in these two cases.

jreznot:
@jrlogsdon Still, why you need this? Do you call it somewhere? What is the use case?

jrlogsdon:
`UpdateCheckerService` should check for available updates. This should be broken on IntelliJ as well after that commit. This has been used implicitly by Android Studio for detecting updates (when the toolbox isn't used to manage updates).

From what I can tell,  `a47d10bada18f84479088f75345fc6440573024e` accidentally changed the default behavior.

What's the purpose of `UpdateCheckerService` if this is not the case?

jreznot:
@jrlogsdon Do you mean show updates in notifications? Because we do not usually do this nowadays, it is quite disturbing and we show them inside Settings button on toolbar instead.

This code does this for us:
```
  if (!showResults) {
    UpdateSettingsEntryPointActionProvider.newPluginUpdates(updatedPlugins, customRepoPlugins)
```

jreznot:
So by default IDEs do this:

Screenshot 2024-03-18 at 21 55 28


Screenshot 2024-03-18 at 21 55 33

jreznot:
Am I correct that you were looking specifically for notifications? If so, we just can close this as Works As Intended

jrlogsdon:
Alright, makes sense. 

So this did change in 2024.1 and it was intentional?

jrlogsdon:
Yes this was for the notification that used to appear.

jreznot:
> Yes this was for the notification that used to appear.

Yes, we are trying to decrease cognitive load of our products and got rid of the notifications for updates everywhere if they were not triggered by users explcitly

jrlogsdon:
Thanks!

jreznot:
You can also track [IDEA-185204](https://youtrack.jetbrains.com/issue/IDEA-185204) for easier updates of plugins

PR #2715, opened 2024-03-19
Open relative files to the project root from markdown preview

raddatzk:
I have a project hosted in Azure DevOps and for proper navigation within markdown files I have to use absolute links relative to the project, like `/documentation/service.md`. 

A link like `[documentation](/documentation/service.md)` opens properly from the editor, but from the preview window this results in an error "Browser configuration problem - The file /documentation/service.md does not exist". While the path is relative to the project root from the editor, it seems like this is treated as an absolute path from the preview window.

With this PR the current behavior is extended to also check if there is also a file that is relative from the project root when it could not be found as an absolute file.


ybedrov:
@FirstTimeInForever Could you please check PR?

PR #2717, opened 2024-03-20
[kotlin] K2 J2K: Move SimplifyNegatedBinaryExpressionInspection to JKTree Phase

jocelynluizzi13:
We can move `SimplifyNegatedBinaryExpressionInspection` to the JKTree phase by running the checks for a negated expression in `JavaToJKTreeBuilder`. 

In this section, there's 3 potential cases:
1. `!(foo is bar)` - Fixed by adding a flag to `JKIsExpression` that handles the `!` in printing
2. `!(foo == bar)` or `!(foo != bar)` - Simplify this when converting the PSI to a JKBinary expression. Note that I'm only simplifying equals/not equals, not "greater than" or any other symbols because it looks like we ignore those in `NegatedBinaryExpressionSimplificationUtils#canBeSimplifiedWithoutChangingSemantics`
3. `Object.equals(foo, bar)` or `foo.equals(bar)`— we can convert this directly to a `JKBinaryExpression` with `JavaTokenType.NE `

### DIFF
--- a/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
+++ b/plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
@@ -103,9 +103,6 @@ private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
     inspectionBasedProcessing(IfThenToElvisInspection(highlightStatement = true, inlineWithPrompt = false), writeActionNeeded = false) {
         it.shouldBeTransformed()
     },
-    inspectionBasedProcessing(KotlinInspectionFacade.instance.simplifyNegatedBinaryExpression) {
-        it.canBeSimplifiedWithoutChangingSemantics()
-    },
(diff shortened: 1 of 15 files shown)


abelkov (changes requested):
Please look through all testdata of the original inspection (like it was done in https://github.com/JetBrains/intellij-community/pull/2709, for example) and port this testdata to J2K testdata, because now that we have ported the inspection's logic, we need to port the tests too.

> it looks like we ignore those in NegatedBinaryExpressionSimplificationUtils#canBeSimplifiedWithoutChangingSemantics

No, the set of all supported operators is listed in `org.jetbrains.kotlin.idea.codeinsight.utils.NegatedBinaryExpressionSimplificationUtils#negate`. We probably need to support all of them too. But be careful with floating point (see `canBeSimplifiedWithoutChangingSemantics` function).

> Object.equals(foo, bar) or foo.equals(bar)

I believe this is covered by EqualsOperatorConversion?

Also, this processing looks like a thing that we would like to disable in "basic" mode, so the code has to be written in such a way that it will be easy to disable later. I propose to move all this logic to a separate JK processing, and don't mix anything into the tree builder. Generally, the builder should be as dumb as possible (and it is already too big as it is).

abelkov (approved):
Thank you, merged with some improvements.

PR #2720, opened 2024-03-22
[gradle] Mark buildSrc modules for Gradle 8.0+

scott-pollom:
After 1fa206f, GradleModuleData.isBuildSrcModule() would incorrectly return false for buildSrc modules when using Gradle 8.0 and higher.

[Related Android Studio bug](https://issuetracker.google.com/issues/328871352)

### DIFF
--- a/plugins/gradle/java/src/execution/build/TasksExecutionSettingsBuilder.java
+++ b/plugins/gradle/java/src/execution/build/TasksExecutionSettingsBuilder.java
@@ -147,11 +147,6 @@ public class TasksExecutionSettingsBuilder {
         rootProjectPath = gradleModuleData.getDirectoryToRunTask();
       }
 
-      // all buildSrc runtime projects will be built by gradle implicitly
-      if (gradleModuleData.isBuildSrcModule()) {
-        continue;
-      }
-
       String gradlePath = gradleModuleData.getGradlePath();
       List<TaskData> taskDataList =
         ContainerUtil.mapNotNull(gradleModuleData.findAll(ProjectKeys.TASK), taskData -> taskData.isInherited() ? null : taskData);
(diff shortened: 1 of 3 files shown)


scott-pollom:
Filed [IDEA-349902](https://youtrack.jetbrains.com/issue/IDEA-349902/GradleModuleData.isBuildSrcModule-incorrectly-returns-false-for-Gradle-8.0) to track.

nskvortsov:
Unfortunately, the proposed patch breaks the test `org.jetbrains.plugins.gradle.compiler.GradleDelegatedBuildTest#testDelegationBuildsBuildSrc`. Could you please fix it?

scott-pollom:
I removed the following lines from `TasksExecutionSettingsBuilder`, and that fixes the broken `org.jetbrains.plugins.gradle.compiler.GradleDelegatedBuildTest#testDelegationBuildsBuildSrc`:

```
if (gradleModuleData.isBuildSrcModule()) {
    continue;
}
```

That will not effect the current IDE behavior for Gradle 8.0+, but might affect the behavior for lower Gradle versions.

I haven't been able to run the full test suite locally, so I'm not sure if that change breaks other tests.

More generally, I'm not sure what the desired behavior of `TasksExecutionSettingsBuilder` is.

nskvortsov:
This `if` statement was a kind of a micro-optimization that would avoid calling compilation tasks on `buildSrc` modules explicitly. Author implies that `buildSrc` will be compiled implicitly anyway. I believe we can safely sacrifice it.
The pull request has passed our CI pipeline and was merged into the main branch. Soon to be mirrored to GitHub.

nskvortsov:
Closing. Thank you for the contribution!

scott-pollom:
Thanks!!

PR #2721, opened 2024-03-22
IDEA-349803: Eliminate the gap before the tab actions' icons

sprigogin:
The gap was caused by ToolWindowContentUi.TabPanel pushing ToolWindowHeader.toolbarWest to the far right when the preferred width of TabPanel was larger than the sum of widths of its visible children. Since reducing the preferred width would prevent TabPanel from growing when its container becomes wider, we move ToolWindowHeader.toolbarWest inside ToolWindowContentUi.TabPanel instead, This way toolbarWest can be positioned adjacent to the rightmost visible tab regardless of the TabPanel's preferred width.


sprigogin:
Superseded by https://github.com/JetBrains/intellij-community/pull/2723.

PR #2727, opened 2024-03-27
[kotlin] J2K: Port JavaMapForEachInspection functionality to BuiltinMembersConversion 

ermattt:
Replaces previous [PR](https://github.com/JetBrains/intellij-community/pull/2645).

It might make sense to generalize some of this logic when porting `DestructureForLoopParameterProcessing`, but that's a job for another PR!

Note that in addition to porting over the logic for destructuring lambda parameters in calls to `Map::forEach`, I also had to introduce two JKTree element classes, which are similar to `KtDestructuringDeclaration` and `KtDestructuringDeclarationEntry`.  The JKTree versions are a little simpler though, with the most significant drawback being that they don't (currently) support nested destructuring declarations. I figured it's pretty unlikely we'll encounter that case during a Java-to-Kotlin conversion, however, so hopefully this implementation is sufficient.


abelkov:
Sorry, I will probably have to review next week, as I'm trying to get my nullability prototype done!

abelkov (approved):

PR #2729, opened 2024-03-29
[kotlin] Delegating to 'var' property: suggest "Change to val" if parameter is not used in function

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-24692/Delegating-to-var-property-Remove-var-quick-fix-produces-incompilable-code-if-property-belongs-to-outer-class#focus=Comments-27-9603616.0-0

### DIFF
--- a/plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/inspections/AbstractLocalInspectionTest.kt
+++ b/plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/inspections/AbstractLocalInspectionTest.kt
@@ -52,6 +52,8 @@ abstract class AbstractLocalInspectionTest : KotlinLightCodeInsightFixtureTestCa
 
     private val fixTextDirectiveName: String = "FIX"
 
+    private val noFixTextDirectiveName: String = "NO_FIX"
+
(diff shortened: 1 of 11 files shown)


plugins/kotlin/idea/tests/test/org/jetbrains/kotlin/idea/inspections/AbstractLocalInspectionTest.kt
```diff
@@ -161,7 +163,8 @@ abstract class AbstractLocalInspectionTest : KotlinLightCodeInsightFixtureTestCa
         expectedProblemString: String?,
         expectedHighlightString: String?,
         localFixTextString: String?,
-        inspectionSettings: Element? = null
+        inspectionSettings: Element?,
+        noLocalFixTextStringList: List<String> = emptyList(),
```
PetrakovichVictoria, line 167: It's better not to add to variable names their type, for example, `List`. I renamed it to `noLocalFixTextStrings`.

PetrakovichVictoria:
Thank you, @t-kameyama! It's being merged.

t-kameyama:
cf575605b856e870eaba65fbc2cfa6790dc9c0c0

PR #2730, opened 2024-03-30
Avoid bundling duplicate Windows binaries

gharrma:
IntelliJ 2024.1 contains duplicate Windows binaries. For example:

* bin/restarter.exe
* bin/amd64/restarter.exe
* bin/aarch64/restarter.exe

Only the first binary is used at runtime.

This appears to have been caused by commit c8910659b4, which changed `.include("*.*")` to `.includeAll()`. The latter includes files recursively, thereby including subdirectories such as 'amd64' and 'aarch64'.


trespasserw:
Thanks! I have rewrote that place, so there is no recursive copying anymore.

PR #2731, opened 2024-03-30
Fix incorrect comparison warnings.

rishabhsdev:
IDE gives incorrect warning messages while comparing Arrays, Strings, and Objects.

I have modified the affected files to display correct warning messages.


amaembo:
The current warnings are correct. They say what's written in the code, not what should be written. For example, when you use `==` to compare arrays, it says: "Array objects are compared using '==', not 'Arrays.equals()'", which describes the current erroneous state of the program. Swapping the expected and actual operator is incorrect. One could alternatively write: "Array objects _should be_ compared using 'Arrays.equals()', not '=='", but this is in general a matter of taste and uniformity. In our messages, we prefer to describe the current state of the code, rather than the expected one, because it's not always possible to say what is expected. E.g., probably the '==' operator is correct, but the author of the code mistakenly compares the wrong variable. So, I think, the current messages are fine.

PR #2732, opened 2024-04-01
KTIJ-29263 Add quickfix for VOLATILE_ON_VALUE to change property to 'var'

t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-29263


abelkov (approved):
Thank you!

PR #2733, opened 2024-04-02
[kotlin] K2 J2K: Move MayBeConstantInspection to JKTree

jocelynluizzi13:
**draft commit because it currently fails on cases where mutability is not yet determined (will be fixed when mutability is migrated earlier) and when we have cases of static init blocks in java that get merged into one expression in Kotlin, because it hasn't happened yet at this stage.

This diff adds `const` modifiers to fields that meet the given criteria: immutable, initialized immediately, primitive type or string, and top level or in a companion / object. 

This does not currently handle @JVMField cases -> do we add the `const` and remove the annotation in these cases?


abelkov:
Thank you, I will merge the partial support right now, since it fixes 11 tests. Other cases can be supported later.

> fields that meet the given criteria: immutable, initialized immediately

You missed an important rule: the initializer should be a constant (i.e., not a method call or something like that).

I will enable this conversion only for literal expressions for now. It won't support compositions of constant expressions, but it's better than nothing.

PR #2735, opened 2024-04-03
New framework for displaying images in rendered documentation

alex-bel:
- Supports raster and SVG images
 - Re-rasterizes SVG images when view dimensions are changed
 - Supports changes tracking of local files via VirtualFileSystem; automatically reloads images if they are changed
 - Supports base64 data urls

DocRenderer and DocumentationEditorPane now properly handle content height change due to images being loaded lazily

Fixes IDEA-256643

Marking it as draft for now since this is a proof-of-concept and I need initial feedback on the code architecture and whether you want to have this feature/fix merged at all.
I will be adding unit-tests soon


platform/lang-impl/src/com/intellij/codeInsight/documentation/CachingAdaptiveImageManager.kt
```diff
@@ -0,0 +1,568 @@
...
+internal fun determineCacheSize(sizeRegistryKey: String): Long {
+  var memorySizeMb = 750 // default value, if something goes wrong
+  try {
+    memorySizeMb = VMOptions.readOption(VMOptions.MemoryKind.HEAP, true)
+  }
+  catch (e: Throwable) {
+    logger<CachingAdaptiveImageManagerService>().error("Failed to get Xmx", e)
+  }
+
+  var cacheSize = 20_000_00L // minimum value
```
stachenov, line 84: That's a rather unusual way to write 2 MB. Or was it intended to be 20 MB and one zero at the end was lost?

alex-bel: Indeed, it is a typo, this was intended to be 20MB. I don't rememeber the reason though why I wanted to set it to 20 (original value from AbstractDocRenderMemoryManager was 10MB). Most likely I opened couple of tabs of code with lots of images in javadocs (like Project Reactor's [Flux](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html)) and somehow estimated amount of  pixels in visible rasterized images
Are you going to merge this stuff? Want me to rebase these changes and test them on newer code base?

stachenov: I've already rebased and merged locally. I'll push it soon, but it'll be disabled in 2025.3 because we haven't planned for this, so there are no resources for the QA to test it and so on. It'll be under a registry key, which we'll then enable internally and hopefully enable it in 2026.1.

There are no big issues with this, and I really like how it's done, but I've seen a few glitches even as I cleaned up code a bit. Once I faced a disco-like endless resize of a documentation popup, apparently popup resize caused image resize, which caused the popup to resize again and so on. I'll have to investigate it if I can reproduce it.

But overall I hope to get it released. I see that much work was put into it and I really like the outcome.

alex-bel: Thank you for pushing this forward
FYI, I also encountered couple of glitches when I was testing this code back then. I tracked one of them down to Swing's HTML internals and submitted bug report to OpenJDK: https://bugs.openjdk.org/projects/JDK/issues/JDK-8329849 
It is still not fixed unfortunatelly :(
I believe I also saw the endless resize loop you described, but I don't remember if it is related to that bug in OpenJDK or not

stachenov:
Merged manually, will be available in 2025.3 under the `doc.render.adaptive.image` registry key (off by default) and hopefully in 2026.1 by default.

PR #2737, opened 2024-04-09
PY-42693 (wip): Infer completitions for lambda parameters in PyCharm

Christopher-Chianelli:
A partial fix for https://youtrack.jetbrains.com/issue/PY-42693/PyCharm-cannot-infer-type-for-lambda-parameters

Added a new TypeProvider that tries to determine the callable type of a lambda in Python. In particular:

- If the lambda is part of an argument list that has a call expression with a known callee
- Then it finds the position of the lambda inside the argument list, and resolves the callee. The type of the entire lambda expression is determined to be the same as the type of the Nth argument of the callee (where N is the lambda's position in the argument list).

Modified PyTypingTypeProvider to determine type of lambda named parameters. In particular:

- If a NamedParameter is in a parameter list belonging to a lambda expression
- Then get the type of the Nth parameter (where N is the index of the NamedParameter in the lambda expression)

With these changes, code completions work; i.e., given

```python
from typing import Callable

class A:
    my_attribute: int

def function(a: A, fun: Callable[[A], int]):
    return fun(a)

a = A()
a.my_attribute = 10
function(a, lambda parameter: parameter.my_
```

it would suggest the completion "my_attribute", which has type "int".

However, code insights still suggests parameter is `Any`, and doesn't seem to check the return type of the lambda expression.

Needs tests.

(I am unfamiliar with the code base and unsure where to go from here)

### DIFF
--- a/python/python-psi-impl/src/com/jetbrains/python/codeInsight/lambda/PyLambdaTypeProvider.java
+++ b/python/python-psi-impl/src/com/jetbrains/python/codeInsight/lambda/PyLambdaTypeProvider.java
@@ -1,0 +1,42 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.jetbrains.python.codeInsight.lambda;
+
+import com.jetbrains.python.psi.*;
+import com.jetbrains.python.psi.types.PyType;
+import com.jetbrains.python.psi.types.PyTypeProviderBase;
+import com.jetbrains.python.psi.types.TypeEvalContext;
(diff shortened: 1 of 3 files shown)


trinhanhngoc:
Hope this feature will be landed in PyCharm 2026. Hope hope hope...

vladimir-koshelev:
Thanks for the contribution!
However, we’ve already started working on a proper implementation of this functionality. The proposed approach with a custom TypeProvider doesn’t really fit our architecture — it’s better to implement type inference directly in the relevant PSI elements where it naturally belongs. This ensures the solution is consistent, maintainable, and integrates well with the existing type inference system.

PR #2739, opened 2024-04-12
[kotlin] K2 J2K: Move RemoveRedundantVisibilityModifierProcessing to JKTree

jocelynluizzi13:
Move `RemoveRedundantVisibilityModifierProcessing` to the JKTree stage. 

To do this, added some small checks to `modifiers.kt` whenever possible. For constructors, made changes to the printing phase to avoid printing the constructor itself when it's unnecessary. For overrides, used symbol checking in a separate conversion file to find the visibility of the overridden method. 

This still has two failures—there's a nullability problem with enum constructors, and it fails when there is nested inheritance and we're overriding a method not present in the in-between class.

Before I invest more time in fixing these errors, I wanted to post this to get feedback on the implementation


plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RedundantModifiersInOverrideConversion.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.jetbrains.kotlin.nj2k.tree.*
+import org.jetbrains.kotlin.nj2k.tree.OtherModifier.OVERRIDE
+import org.jetbrains.kotlin.nj2k.tree.Visibility.*
+import org.jetbrains.kotlin.nj2k.types.JKClassType
+import org.jetbrains.kotlin.nj2k.types.fqName
+import org.jetbrains.kotlin.nj2k.visibility
+import org.jetbrains.kotlin.psi.psiUtil.visibilityModifierTypeOrDefault
+
+
+class RedundantModifiersInOverrideConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
```
abelkov, line 24: I wonder, can we simplify this by essentially reimplementing `org.jetbrains.kotlin.resolve.OverridingUtil#findMaxVisibility` over Java PSI (using `com.intellij.psi.PsiMethod#findSuperMethods()`). (`findMaxVisibility` is used in `org.jetbrains.kotlin.idea.core.PsiModificationUtilsKt#implicitVisibility`).

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -759,10 +760,22 @@ class JKCodeBuilder(context: NewJ2kConverterContext) {
             renderTokenElement(method.rightParen)
         }
 
+        private fun redundantModifiersForConstructor(constructor: JKConstructor): Boolean {
```
abelkov, line 763: Can this be merged into `org.jetbrains.kotlin.nj2k.tree.ModifiersKt#isRedundant`?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -777,9 +790,13 @@ class JKCodeBuilder(context: NewJ2kConverterContext) {
 
         override fun visitKtPrimaryConstructorRaw(ktPrimaryConstructor: JKKtPrimaryConstructor) {
             ktPrimaryConstructor.annotationList.accept(this)
-            renderModifiersList(ktPrimaryConstructor)
+            val redundantConstructor = redundantModifiersForConstructor(ktPrimaryConstructor)
+            if (!redundantConstructor) {
+                renderModifiersList(ktPrimaryConstructor)
```
abelkov, line 795: `redundantModifiersForConstructor` checks only visibility modifiers, but here we skip rendering any modifiers. Is this intended?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -777,9 +790,13 @@ class JKCodeBuilder(context: NewJ2kConverterContext) {
...
+            }
 
-            val needConstructorKeyword = ktPrimaryConstructor.hasAnnotations || ktPrimaryConstructor.visibility != PUBLIC
+            val needConstructorKeyword =
+                (ktPrimaryConstructor.hasAnnotations || ktPrimaryConstructor.visibility != PUBLIC) && !redundantConstructor
```
abelkov, line 799: `ktPrimaryConstructor.visibility != PUBLIC` should probably be moved out of this place now

abelkov (changes requested):

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RedundantModifiersInOverrideConversion.kt
abelkov: I mean, `org.jetbrains.kotlin.idea.inspections.RedundantVisibilityModifierInspection.Holder#getRedundantVisibility` looks a lot simpler than the new conversion, and the hardest part is `OverridingUtil#findMaxVisibility`.

jocelynluizzi13:
> NullabilityConversion

I don't think that will work in this case, because the enum constructor is `Internal` not private, and the parameter is never used in the class. Specifically, `plugins/kotlin/j2k/shared/tests/testData/newJ2k/enum/oneLinePerEntry.kt` is failing because both `name` and `s` are appearing as nullable strings. I can't figure out what happened to change it to nullable in this case.

Otherwise, the rest of the suggested changes have been made and `IGNORE_K2` flags have been removed

abelkov:
> the enum constructor is Internal not private

You're thinking about the enum itself. But I mean the enum _constructor_, which must be private.

In such code, the constructor is private, and we can easily analyze all its three invocations:

```
internal enum class BigEnum(name: String?) {
    ENUM_ONE("SOME_NAME_TO_PUT_HERE"),
    ENUM_TWO("SOME_NAME_TO_PUT_HERE"),
    ENUM_THREE("SOME_NAME_TO_PUT_HERE"),
}
```

abelkov:
Also, there are still failing tests, please take a look.

org.jetbrains.kotlin.nj2k.TextNewJavaToKotlinCopyPasteConversionTestGenerated#testComments
org.jetbrains.kotlin.nj2k.NewJavaToKotlinCopyPasteConversionTestGenerated#testComments

abelkov (changes requested):
Don't forget to delete `RemoveRedundantVisibilityModifierProcessing`

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -904,6 +905,24 @@ class JavaToJKTreeBuilder(
...
+            val methodVisibility = visibility().visibility
+
+            // ensure none of the found super methods violate redundancy conditions
+            return superMethods.all { superMethod ->
+                val superVisibility = superMethod.visibility().visibility
+                when {
+                    name == "clone" && superMethod.containingClass?.name == "Object"
+                            && superMethods.size == 1 -> methodVisibility == PROTECTED // special built-in type
+                    superVisibility == methodVisibility -> true
+                    superVisibility == INTERNAL -> methodVisibility == PUBLIC
```
abelkov, line 920: I noticed that in `K2JavaToKotlinConverterSingleFileTestGenerated.ToKotlinClasses#testMutableIterator`, the overridden methods of the public `java.util.Iterator` are converted to `internal`, but actually they should be public. It looks like this is a bug in `org.jetbrains.kotlin.nj2k.PsiUtilsKt#visibility` for members with default visibility. Can you please investigate how we can do better here? Probably there is some method to compute the default visibility of Java declarations somewhere in the Java plugin that we can reuse.

Also, what happens if our method overrides a Kotlin method?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -904,6 +905,24 @@ class JavaToJKTreeBuilder(
...
+            val superMethods = findSuperMethods()
+            if (superMethods.isEmpty()) return false // Unknown super method
+            val methodVisibility = visibility().visibility
+
+            // ensure none of the found super methods violate redundancy conditions
+            return superMethods.all { superMethod ->
+                val superVisibility = superMethod.visibility().visibility
+                when {
+                    name == "clone" && superMethod.containingClass?.name == "Object"
+                            && superMethods.size == 1 -> methodVisibility == PROTECTED // special built-in type
```
abelkov, line 918: A similar check also exists in `JKModifierElement.isRedundant`, is it necessary in both places?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/tree/modifiers.kt
```diff
@@ -144,15 +146,36 @@ inline fun JKModifiersListOwner.forEachModifier(action: (JKModifierElement) -> U
...
         OPEN, ABSTRACT -> isOpenAndAbstractByDefault
+        PROTECTED -> owner is JKMethod && owner.name.value == "clone"
+                && owner.parameters.isEmpty() && owner.returnType.type.fqName == "kotlin.Any"
         else -> false
     }
-}
\ No newline at end of file
+}
+
+private fun JKMethod.isRedundantPublicBuiltIn() = when {
```
abelkov, line 176: Does `PsiMethod#findSuperMethods()` not find methods from `java.lang.Object`?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/tree/modifiers.kt
```diff
@@ -144,15 +146,36 @@ inline fun JKModifiersListOwner.forEachModifier(action: (JKModifierElement) -> U
...
+    name.value == "toString" && parameters.isEmpty() && returnType.type.fqName == "kotlin.String" -> true
+    name.value == "equals" && parameters.size == 1 && returnType.type.fqName == "kotlin.Boolean" -> true
```
abelkov, line 178: I believe the parameter type should also be checked, otherwise it may not be an override. We should also add a test for this case.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/NullabilityConversion.kt
```diff
@@ -21,6 +19,14 @@ class NullabilityConversion(context: NewJ2kConverterContext) : RecursiveConversi
         if (element is JKTypeCastExpression) {
             element.updateNullability()
         }
+        if (element is JKMethod) {
```
abelkov, line 22: `element` can't be both `JKTypeCastExpression` and `JKMethod`. Better to use "else if" or "when".

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/NullabilityConversion.kt
```diff
@@ -33,4 +39,19 @@ class NullabilityConversion(context: NewJ2kConverterContext) : RecursiveConversi
...
     }
-}
+    private fun JKMethod.updateNullabilityOfParameters() {
+        val enumConstants = mutableListOf<JKEnumConstant>()
+        parentOfType<JKClassBody>()?.forEachDescendantOfType<JKEnumConstant> {
+            enumConstants.add(it)
+        }
+        for (i in parameters.indices) {
+            val isNotNull = enumConstants.none {
+                val argument = it.arguments.arguments.getOrNull(i)
```
abelkov, line 49: Probably we should check that the number of arguments matches the number of parameters, because there can be varargs?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/NullabilityConversion.kt
```diff
@@ -33,4 +39,19 @@ class NullabilityConversion(context: NewJ2kConverterContext) : RecursiveConversi
...
+                argument == null || argument.value !is JKLiteralExpression
```
abelkov, line 50: It's better to directly check the nullability of the expression's type and not hardcode `JKLiteralExpression`.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -786,9 +791,9 @@ class JKCodeBuilder(context: NewJ2kConverterContext) {
 
         override fun visitKtPrimaryConstructorRaw(ktPrimaryConstructor: JKKtPrimaryConstructor) {
             ktPrimaryConstructor.annotationList.accept(this)
-            renderModifiersList(ktPrimaryConstructor)
+            val renderedModifiersCount = renderModifiersList(ktPrimaryConstructor)
 
-            val needConstructorKeyword = ktPrimaryConstructor.hasAnnotations || ktPrimaryConstructor.visibility != PUBLIC
+            val needConstructorKeyword = ktPrimaryConstructor.hasAnnotations || renderedModifiersCount > 0
```
abelkov, line 796: Count is used as a boolean, it's better to make it an actual Boolean.

abelkov (changes requested):

jocelynluizzi13:
Made some changes based on comments. the `org.jetbrains.kotlin.nj2k.PsiUtilsKt#visibility` function needed to be updated so that it read text strings if there were no children of `modifierList`. In Kotlin-specific cases, this is the only way to access the individual modifiers. 

Also tested on some live files and ran through the test cases.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/psiUtils.kt
```diff
@@ -49,22 +50,43 @@ fun canKeepEqEq(left: PsiExpression, right: PsiExpression?): Boolean {
 fun PsiMember.visibility(
     referenceSearcher: ReferenceSearcher,
     assignNonCodeElements: ((JKFormattingOwner, PsiElement) -> Unit)?
-): JKVisibilityModifierElement =
-    modifierList?.children?.mapNotNull { child ->
+): JKVisibilityModifierElement {
+    if (parentOfType<PsiClass>()?.isInterface == true) {
+        // interface methods are always public
```
abelkov, line 55: What about this?

```
interface I {
    private void method() { // private
    }
    
    void newMethod(); // package private
}
```

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/psiUtils.kt
```diff
@@ -49,22 +50,43 @@ fun canKeepEqEq(left: PsiExpression, right: PsiExpression?): Boolean {
...
         }?.let {
             JKVisibilityModifierElement(it)
         }?.also { modifier ->
             assignNonCodeElements?.also { it(modifier, child) }
         }
-    }?.firstOrNull() ?: JKVisibilityModifierElement(Visibility.INTERNAL)
+    }?.firstOrNull()
+
+    if (visibility != null) return visibility
+    // for Kotlin classes, modifierList.children() returns null
```
abelkov, line 74: I think it's better to have a separate special method for Kotlin visibility, because it is only needed in one special case (computing visibility of Kotlin overridden methods), and it is kinda unexpected to have this code mixed in here.

plugins/kotlin/j2k/shared/tests/testData/newJ2k/function/overrideExtensionFunction.java
```diff
@@ -1,5 +1,4 @@
 // IGNORE_K2
-// Not supported yet: KTIJ-7583
```
abelkov, line 2: KTIJ-7583 is not about visibility, it is about converting extension functions

plugins/kotlin/j2k/shared/tests/testData/newJ2k/visibility/overrideEqualsBuiltIn.java
```diff
@@ -0,0 +1,9 @@
+// IGNORE_K2
+import org.jetbrains.annotations.Nullable;
+
+public class Test {
+    @Override
+    public Boolean equals(@Nullable Any other) {
```
abelkov, line 6: Please don't use Kotlin `Any` in Java tests. The test should use types that are natural to use in Java code.

abelkov (changes requested):

abelkov:
Due to my recent optimization, the compiler errors are checked only in testdata that already has "// ERROR" directives. I now realize that this is a mistake, because we can accidentally introduce some new errors in new or modified testdata. I will revert this soon. After you run the test suite, you may find some new errors in testdata.

plugins/kotlin/j2k/shared/tests/testData/newJ2k/function/extendsBaseWhichExtendsObject.kt
```diff
@@ -42,7 +42,7 @@ internal open class Base : Cloneable {
     }
 
     @Throws(Throwable::class)
-    protected open fun finalize() {
+    open fun finalize() {
```
abelkov, line 45: I think 'protected' should be preserved for 'finalize' that directly overrides built-in 'finalize'. I will fix this in `JavaStandardMethodsConversion` before merging this PR.

abelkov (approved):

PR #2740, opened 2024-04-17
[kotlin] J2K: Port RemoveUnnecessaryParentheses to JKTree

ermattt:
In the spreadsheet, this one is listed as "RemoveUnnecessaryParenthesesIntention".

There were a few small changes in formatting required to make the tests pass, but they seemed harmless enough.

Adding this JKTree processing step unblocked 17 K2 tests, but wasn't sufficient to remove the `RemoveUnnecessaryParenthesesIntention` postprocessing step because some other postprocessing steps add parentheses, too (mostly, it seems, while appending `!!`).

Additional test cases included in this commit are largely based on the contents of [`kotlin/code-insight/intentions-shared/tests/testData/intentions/removeUnnecessaryParentheses`](https://github.com/JetBrains/intellij-community/tree/0d2d78699f0f2d8be32292375e10160727e74d12/plugins/kotlin/code-insight/intentions-shared/tests/testData/intentions/removeUnnecessaryParentheses).


plugins/kotlin/j2k/shared/tests/testData/newJ2k/binaryExpression/operatorPrecedence.kt
```diff
@@ -28,7 +28,8 @@ internal object J {
 
         val insideMultiline2 = (1
                 + 1
-                - 0x1234) and (0x1234 ushr 1)
+                - 0x1234
+                ) and (0x1234 ushr 1)
```
abelkov, line 32: Is it possible to preserve original formatting?

plugins/kotlin/j2k/shared/tests/testData/newJ2k/methodCallExpression/stringConcatFormatting.kt
```diff
@@ -6,9 +6,9 @@ internal object A {
         val result = s
             .replace("_", "/") + "=="
 
-        val result2 = (s
-            .replace("_", "/") // comment
-                + "==")
+        val result2 = s
+            .replace("_", "/") + "==" // comment
+
```
abelkov, line 11: Would be nice to drop the extra line break

plugins/kotlin/j2k/shared/tests/testData/newJ2k/assignmentExpression/expressionAssignmentWithParentheses.kt
```diff
@@ -7,10 +7,10 @@ internal object Test {
         var d = true
         val e = true
 
-        if (e.let { d = d and it; d }.also { (c) = it }.let { b = b or it; b }.also { a = it });
+        if (e.let { d = d and it; d }.also { c = it }.let { b = b or it; b }.also { a = it });
         while ((((((b.also { a = it }))))));
         do {
-        } while (((b.let { (a) = (a) xor it; (a) })))
+        } while (((b.let { a = (a) xor it; (a) })))
```
abelkov, line 13: I know it is not strictly related, but is it possible to clean up this mess? It seems like it should not be hard to drop straight nested parentheses.

plugins/kotlin/j2k/k1.new/tests/test/org/jetbrains/kotlin/nj2k/NewJavaToKotlinConverterSingleFileTestGenerated.java
```diff
@@ -5064,6 +5064,11 @@ public void testParenthesized2() throws Exception {
         public void testParenthesized3() throws Exception {
             runTest("../../shared/tests/testData/newJ2k/parenthesizedExpression/parenthesized3.java");
         }
+
+        @TestMetadata("parenthesizedAddedAndRemovedDuringConversion.java")
+        public void testParenthesizedAddedAndRemovedDuringConversion() throws Exception {
```
abelkov, line 5069: K2 test should be also regenerated

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+            ) {
+                return false
+            }
+            return parent.right == expression
+        }
+
+        return innerPriority < parentPriority
+    }
+
+    private fun isAtomic(expression: JKExpression): Boolean {
```
abelkov, line 119: Is this the negation of checks in `org.jetbrains.kotlin.nj2k.ExpressionsKt#parenthesizeIfCompoundExpression`?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+        return expression is JKQualifiedExpression ||
+                expression is JKKtWhenExpression ||
+                expression is JKCallExpression ||
+                expression is JKFieldAccessExpression ||
+                expression is JKLiteralExpression
+    }
+
+    private fun recursivelyContainsNewlines(expression: JKExpression): Boolean {
```
abelkov, line 127: Unused

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+        return false
+    }
+
+    private fun recursivelyEndsWithNewline(expression: JKExpression): Boolean {
+        if (expression.hasLineBreakAfter) return true
+        val lastChild = expression.children.lastOrNull()?.safeAs<JKExpression>() ?: return false
+        return recursivelyEndsWithNewline(lastChild)
+    }
+
+    private fun recursivelyStartsWithNewline(expression: JKExpression): Boolean {
```
abelkov, line 150: unused

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+import org.jetbrains.kotlin.nj2k.types.fqName
+import org.jetbrains.kotlin.parsing.KotlinExpressionParsing
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+import org.jetbrains.kotlin.utils.filterIsInstanceAnd
+
+/**
+ * Analogous to `RemoveUnnecessaryParenthesesIntention`.
+ * Removes parentheses (conservatively) added by other steps.
+ */
+class RemoveUnnecessaryParenthesesConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
```
abelkov, line 21: Have you considered the opposite approach, i.e., do a call site check before generating a set of parentheses (the check can possibly be performed inside `org.jetbrains.kotlin.nj2k.ExpressionsKt#parenthesize`). In this case, any following conversions won't be able to circumvent this conversion and accidentally introduce redundant parentheses. But I understand that it can be harder to do an "in place" check while we are modifying the tree and maybe have incomplete information.

abelkov (changes requested):
Looks good overall! I did not dig into the nuances of priorities though, will have to take another look later.

plugins/kotlin/j2k/shared/tests/testData/newJ2k/assignmentExpression/expressionAssignmentWithParentheses.kt
abelkov: Oh, sorry, it was in the original file, and we just preserve the parentheses, right? Then no problem here.

abelkov:
So, now `org.jetbrains.kotlin.nj2k.printing.JKCodeBuilder.Visitor#visitParenthesizedExpressionRaw` can drop this code?

```
            if (parenthesizedExpression.shouldBePreserved) {
                printExplicitLabel(parenthesizedExpression)
            }
```

plugins/kotlin/j2k/shared/tests/testData/newJ2k/binaryExpression/operatorPrecedence.kt
ermattt: In this case I think the updated formatting is actually closer to the original java which was:
```
int insideMultiline2 = 1
                               + 1
                               - 0x1234
                               & 0x1234 >>> 1;
```

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    private fun areParenthesesNecessary(expression: JKParenthesizedExpression): Boolean {
+        val innerExpression = expression.expression
+        val parent = expression.parent
+
+        // We can end up with silly "preexisting" parentheses when converting expressions that take significantly different forms in Java
+        // vs. Kotlin. For example, we can have unnecessary parentheses wrapping the left side of an assignment when turning an expression
+        // like `a = b += (c) = ((d *= e))` into `(c) = ((e.let { d *= it; d })) ...`
+        if ((parent is JKKtAssignmentStatement && parent.field == expression) || (parent is JKAssignmentChainLetLink && parent.receiver == expression)) return false
```
abelkov, line 39: This is very specific to `AssignmentExpressionUnfoldingConversion`, I would prefer this case to be fixed there (don't add redundant parentheses), if it's not too hard.

abelkov:
> some other postprocessing steps add parentheses, too

Yes, looks like `org.jetbrains.kotlin.idea.j2k.post.processing.processings.DiagnosticBasedProcessingsKt#getFixTypeMismatchDiagnosticBasedProcessing` is the culprit. This is easy to fix though. After we do `element.replace(psiFactory.createExpressionByPattern("($0)!!", element.text))`, we can immediately remove the possibly redundant parentheses. We can call `RemoveUnnecessaryParenthesesIntention` here directly.

I will do this change myself, because some refactoring of the plugin code may be required.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+        return recursivelyStartsWithNewline(lastChild)
+    }
+
+    /**
+     * Based on `getPriority` in KtPsiUtil.java
+     */
+    private fun getPriority(expression: JKElement): Int {
+        val maxPriority = KotlinExpressionParsing.Precedence.entries.size + 1
+
+        if (expression is JKPostfixExpression || expression is JKQualifiedExpression || expression is JKCallExpression || expression is JKMethodReferenceExpression) {
```
abelkov, line 162: JKArrayAccessExpression?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+            ) {
+                return maxPriority - precedence.ordinal - 1
+            }
+        }
+
+        return maxPriority
+    }
+
+    companion object {
+        val STD_INFIX_FUNCTIONS = setOf(
```
abelkov, line 191: Do we need all of them? Is it enough to check that the operator is `JKKtWordOperatorToken`?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    private fun recursivelyStartsWithNewline(expression: JKExpression): Boolean {
+        if (expression.hasLineBreakBefore) return true
+        val lastChild = expression.children.firstOrNull()?.safeAs<JKExpression>() ?: return false
+        return recursivelyStartsWithNewline(lastChild)
+    }
+
+    /**
+     * Based on `getPriority` in KtPsiUtil.java
```
abelkov, line 157: Adding this reference is great, but using the FQN is even better: `org.jetbrains.kotlin.psi.KtPsiUtil#getPriority`.

abelkov (changes requested):

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
ermattt: Not quite: we can't replace one with the negation of the other because there are a bunch of expression types that are neither compound nor atomic (e.g. JKPrefixExpression)

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
ermattt: I did, but like you said it's difficult/unwieldy for 2 reasons:
1. Like you mentioned, while we're building the JKTree, we don't have all the information we need about parent and descendent nodes
2. Sometimes parentheses become unnecessary in a later step (e.g. `((int) x).dec()` -> `(x as Int).dec()` -> `x.toInt().dec()`). We could require that every step take into consideration that it might need to remove now-unnecessary parentheses, but that seems laborious and error-prone, or at least not strictly better than this approach, IMO

plugins/kotlin/j2k/shared/tests/testData/newJ2k/methodCallExpression/stringConcatFormatting.kt
ermattt: I definitely agree, but unfortunately it's pretty hard to fix because of the way the printing phase is set up. In this case, the problem is that the `JKLocalVariable` node has 2 trailing newlines, and the comment node has 1 trailing newline.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
ermattt: I don't think it's possible to do this easily for every corner case in `AssignmentExpressionUnfoldingConversion`. For example, when running `AssignmentExpressionUnfoldingConversion` on the java expression `a = b |= (c) = ((d &= e))`, we get to a state in `JKJavaAssignmentExpression.toExpressionChainLink` where the receiver is still `((d &= e))` (and therefore needs a layer of parentheses). Later that binary expression gets transformed into a let expression, but the opportunity to remove the parentheses easily is already gone.

All of that being said, I made a change to `AssignmentExpressionUnfoldingConversion` which covers most cases (e.g. instead of `e.let { d = d and it; d }.also { c = it }.let { b = b or it; b }.also { a = it }` we get `(e.let { d = d and it; d }).also { c = it }.let { b = b or it; b }.also { a = it }`, but that's still better than the original `((e.let { d = d and it; d })).also { c = it }.let { b = b or it; b }.also { a = it }`)

abelkov (approved):
Thank you, nice work! I will try to solve the post-processing problem with `!!` and merge this with a little cleanup.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/RemoveUnnecessaryParenthesesConversion.kt
```diff
@@ -0,0 +1,213 @@
...
+        // like `a = b += (c) = ((d *= e))` into `(c) = ((e.let { d *= it; d })) ...`
+        if ((parent is JKKtAssignmentStatement && parent.field == expression) || (parent is JKAssignmentChainLetLink && parent.receiver == expression)) return false
+
+        // If parentheses were present in original Java code, leave them be
+        if (expression.shouldBePreserved) return true
+
+        if (parent == null) return true
+
+        // Nested parentheses are always unnecessary
+        if (innerExpression is JKParenthesizedExpression || parent is JKParenthesizedExpression) return false
```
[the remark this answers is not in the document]
abelkov: Without the `parent` check `K2JavaToKotlinConverterSingleFileTestGenerated.BinaryExpression#testMultiLine2` fails, so I will add it back.

PR #2741, opened 2024-04-22
[kotlin] J2K: Add JKTree processing step to remove redundant type projections

ermattt:
In the spreadsheet, this one is listed as "REDUNDANT_PROJECTION."

Unfortunately this JKTree processing step isn't sufficient to remove the corresponding diagnostic-based postprocessing step: we need mutability analysis to run before we can remove redundant type projections from collections' type parameters (e.g. we can't remove the `out` from `MutableList<out Foo>`, but we can from `List<out Foo>`). If I omit the postprocessing step, there are 8 failing K1 tests (see attached screenshot).

I also tried implementing these changes by modifying JavaToJKTreeBuilder, but that had a bunch of unpleasant side effects, and didn't seem like a natural place to stick this logic.

NB: The test cases added here are largely based on my imagination, and might be overkill.

Screenshot 2024-04-21 at 9 49 32 PM

### DIFF
--- a/plugins/kotlin/j2k/shared/tests/testData/newJ2k/detectProperties/GetterWithRedundantProjection.kt
+++ b/plugins/kotlin/j2k/shared/tests/testData/newJ2k/detectProperties/GetterWithRedundantProjection.kt
@@ -3,7 +3,35 @@ import java.util.Arrays
 class J {
     interface Element
 
+    inner class Box<T>(var t: T) {
+        fun putT(t: T) {
+            this.t = t
+        }
+    }
+
     class Container(vararg children: Element?) : Element {
         val children: List<Element> = ArrayList(Arrays.asList(*children))
+        private val containerChildren: List<in Container> = ArrayList<Container>()
+
+        private fun testAssignment1(elements: List<in Element>) {
+            var elements: List<in Element>? = elements
+            elements = children
+        }
+
+        private fun testAssignment2(elements: List<in Element>) {
+            var elements: List<in Element>? = elements
+            elements = containerChildren
+        }
+
+        fun mergeWithChildren(other: MutableList<in Element>): List<in Container> {
+            other.addAll(children)
+            return other
+        }
+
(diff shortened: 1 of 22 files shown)


plugins/kotlin/j2k/shared/tests/testData/newJ2k/typeParameters/classDoubleParametrizationWithTwoBoundsWithExtending.kt
```diff
@@ -1,4 +1,4 @@
 internal interface INode
 internal interface Node
 internal open class A
-internal class CC<T, K> : A() where T : INode?, T : Comparable<T?>?, K : Node?, K : Collection<K?>?
```
ermattt, line 4: IMO the nullability change on `T` is good--I think removing the out/in earlier allows some of the nullability postprocessing steps to function at their best, since it's still `T?` before any postprocessing steps run (see attached screenshot). That being said, maybe I'm missing some important cases where it should stay nullable?
Screenshot 2024-04-21 at 9 16 29 PM

plugins/kotlin/j2k/shared/tests/testData/newJ2k/detectProperties/GetterWithRedundantProjection.kt
```diff
@@ -3,7 +3,35 @@ import java.util.Arrays
...
+        fun putT(t: T) {
+            this.t = t
+        }
+    }
+
+    inner class Container(vararg children: Element?) : Element {
         val children: List<Element> = ArrayList(Arrays.asList(*children))
+        private val containerChildren: List<in Container> = ArrayList<Container>()
+
+        private fun testAssignment1(elements: List<in Element>) {
```
abelkov, line 16: `List<in Element>` is invalid code. Also, it's better to create a separate test file.

abelkov (changes requested):

PR #2744, opened 2024-04-26
Add Ability to Run Tests on Android ART VM

alonalbert:
Add a way to run Kotlin debugger tests an an ART VM.

At least for now, this will only be possible inside Google since it depends on an internal repo.

The behavior is controlled by a few environment variables and JVM properties:

#### INTELLIJ_DEBUGGER_TESTS_ART / intellij.debugger.tests.art:
  When "true" will run tests under ART.

#### INTELLIJ_DEBUGGER_TESTS_STUDIO_ROOT /  intellij.debugger.tests.studio.root
  The path to an internal Google studio-main repo

#### INTELLIJ_DEBUGGER_TESTS_TIMEOUT_MILLIS / intellij.debugger.tests.timeout.millis

  A timeout in millis. Many tests time out under ART right now so the default timeout (5 minutes) means the take too long to run.

### DIFF
--- a/plugins/kotlin/jvm-debugger/test/test/org/jetbrains/kotlin/idea/debugger/test/KotlinDescriptorTestCase.kt
+++ b/plugins/kotlin/jvm-debugger/test/test/org/jetbrains/kotlin/idea/debugger/test/KotlinDescriptorTestCase.kt
@@ -5,6 +5,7 @@ package org.jetbrains.kotlin.idea.debugger.test
 import com.intellij.debugger.DebuggerManagerEx
 import com.intellij.debugger.DefaultDebugEnvironment
 import com.intellij.debugger.engine.DebugProcessImpl
+import com.intellij.debugger.engine.RemoteStateState
 import com.intellij.debugger.impl.*
 import com.intellij.debugger.settings.DebuggerSettings
 import com.intellij.execution.ExecutionTestCase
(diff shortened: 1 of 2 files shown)


plugins/kotlin/jvm-debugger/test/test/org/jetbrains/kotlin/idea/debugger/test/KotlinDescriptorTestCase.kt
```diff
@@ -405,6 +398,69 @@ abstract class KotlinDescriptorTestCase : DescriptorTestCase(), IgnorableTestCas
...
+                .create(javaCommandLineState.javaParameters)
+
+        val env = javaCommandLineState.environment
+        env.putUserData(DefaultDebugEnvironment.DEBUGGER_TRACE_MODE, traceMode)
+
+        return attachVirtualMachine(javaCommandLineState, env, debugParameters, false)
+    }
+
+    private fun createArtLocalProcess(javaParameters: JavaParameters, environment: ExecutionEnvironment): DebuggerSession {
+        println("Running on ART VM")
```
zuevmaxim, line 428: Could you please replace it with logging?

zuevmaxim:
Merged e620ecf363d41d15cfaa22e2505dcf470eaffeda

PR #2745, opened 2024-04-26
AsmCodeGenerator.java replace try/finally with try with resources

IlyaSekunov:


alexander-lobas:
Our code base a little changed and I apply a part of your commit directly. Thanks.

PR #2746, opened 2024-04-27
[kotlin] J2K: `String.lines()` call should be converted to `lineSequence().asStream()`

isuckatcs:
^KTIJ-29317 Fixed

### DIFF
--- a/plugins/kotlin/j2k/k1.new/tests/test/org/jetbrains/kotlin/nj2k/NewJavaToKotlinConverterSingleFileTestGenerated.java
+++ b/plugins/kotlin/j2k/k1.new/tests/test/org/jetbrains/kotlin/nj2k/NewJavaToKotlinConverterSingleFileTestGenerated.java
@@ -3814,6 +3814,21 @@ public abstract class NewJavaToKotlinConverterSingleFileTestGenerated extends Ab
         public void testStreamOperations() throws Exception {
             runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/streamOperations.java");
         }
+
+        @TestMetadata("stringLines.java")
+        public void testStringLines() throws Exception {
+            runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/stringLines.java");
+        }
+
+        @TestMetadata("stringLinesCustomClass.java")
+        public void testStringLinesCustomClass() throws Exception {
+            runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/stringLinesCustomClass.java");
+        }
+
+        @TestMetadata("stringLinesRef.java")
+        public void testStringLinesRef() throws Exception {
+            runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/stringLinesRef.java");
+        }
     }
 
     @RunWith(JUnit3RunnerWithInners.class)
--- a/plugins/kotlin/j2k/k2/tests/test/org/jetbrains/kotlin/j2k/k2/K2JavaToKotlinConverterSingleFileTestGenerated.java
+++ b/plugins/kotlin/j2k/k2/tests/test/org/jetbrains/kotlin/j2k/k2/K2JavaToKotlinConverterSingleFileTestGenerated.java
@@ -3814,6 +3814,21 @@ public abstract class K2JavaToKotlinConverterSingleFileTestGenerated extends Abs
         public void testStreamOperations() throws Exception {
             runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/streamOperations.java");
         }
+
+        @TestMetadata("stringLines.java")
+        public void testStringLines() throws Exception {
+            runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/stringLines.java");
+        }
+
+        @TestMetadata("stringLinesCustomClass.java")
+        public void testStringLinesCustomClass() throws Exception {
+            runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/stringLinesCustomClass.java");
+        }
+
+        @TestMetadata("stringLinesRef.java")
+        public void testStringLinesRef() throws Exception {
+            runTest("../../shared/tests/testData/newJ2k/javaStreamsApi/stringLinesRef.java");
+        }
     }
 
     @RunWith(JUnit3RunnerWithInners.class)
--- a/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/JKImportStorage.kt
+++ b/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/JKImportStorage.kt
@@ -5,23 +5,23 @@ package org.jetbrains.kotlin.nj2k
 import com.intellij.psi.CommonClassNames
 import org.jetbrains.kotlin.config.LanguageVersionSettings
 import org.jetbrains.kotlin.idea.base.psi.kotlinFqName
+import org.jetbrains.kotlin.idea.base.utils.fqname.isImported
 import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.load.java.NULLABILITY_ANNOTATIONS
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtQualifiedExpression
+import org.jetbrains.kotlin.resolve.ImportPath
 import org.jetbrains.kotlin.resolve.calls.util.getCalleeExpressionIfAny
 import org.jetbrains.kotlin.resolve.jvm.platform.JvmPlatformAnalyzerServices
 
 class JKImportStorage(languageSettings: LanguageVersionSettings) {
     private val imports = mutableSetOf<FqName>()
 
-    private val defaultImports: Set<FqName> =
+    private val defaultImports: Set<ImportPath> =
         JvmPlatformAnalyzerServices.getDefaultImports(
             languageSettings,
             includeLowPriorityImports = true
-        ).mapNotNull { import ->
-            if (import.isAllUnder) null else import.fqName
-        }.toSet()
+        ).toSet()
 
     fun addImport(import: FqName) {
         if (isImportNeeded(import, allowSingleIdentifierImport = false)) {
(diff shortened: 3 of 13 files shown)


plugins/kotlin/j2k/shared/tests/testData/newJ2k/stream/stringLines.java
```diff
@@ -0,0 +1,8 @@
+// IGNORE_K2
```
isuckatcs, line 1: With K2, the generated method is `fun foo(s: String): Stream<String?>` and because of it the test is failing.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
```diff
@@ -0,0 +1,38 @@
...
+class StreamConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
+
+    context(KtAnalysisSession)
+    override fun applyToElement(element: JKTreeElement): JKTreeElement {
+        if (element !is JKQualifiedExpression) return recurse(element)
+        if (element.selector !is JKCallExpressionImpl) return recurse(element)
+
+        val target = (element.selector as JKCallExpressionImpl).identifier.fqName
+
+        // TODO: Handle other cases too.
```
abelkov, line 19: Do you have some cases in mind?

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
```diff
@@ -0,0 +1,38 @@
...
+        )
+    }
+
+    private fun convertStringLines(element: JKQualifiedExpression): JKTreeElement {
+        val lineSequence = "kotlin.text.lineSequence"
+        val asStream = "kotlin.streams.asStream"
+
+        context.importStorage.addImport(asStream)
+
+        return element.receiver.copyTreeAndDetach()
```
abelkov, line 34: It's better to avoid copying elements, if possible. See `plugins/kotlin/docs/j2k/j2k.md`

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
```diff
@@ -0,0 +1,38 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.kotlin.nj2k.conversions
+
+import org.jetbrains.kotlin.analysis.api.KtAnalysisSession
+import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
+import org.jetbrains.kotlin.nj2k.RecursiveConversion
+import org.jetbrains.kotlin.nj2k.callOn
+import org.jetbrains.kotlin.nj2k.tree.*
+
+class StreamConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
```
abelkov, line 10: For a simple conversion such as this, `BuiltinMembersConversion` should be enough.

plugins/kotlin/j2k/shared/tests/testData/newJ2k/stream/stringLines.java
abelkov: That is to be expected, because K2 J2K type argument nullability inference is not implemented yet.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
```diff
@@ -0,0 +1,38 @@
...
+                else -> element
+            }.withFormattingFrom(element)
+        )
+    }
+
+    private fun convertStringLines(element: JKQualifiedExpression): JKTreeElement {
+        val lineSequence = "kotlin.text.lineSequence"
+        val asStream = "kotlin.streams.asStream"
+
+        context.importStorage.addImport(asStream)
```
abelkov, line 32: Should not be necessary, I think

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        return element.receiver.copyTreeAndDetach()
+            .callOn(symbolProvider.provideMethodSymbol(lineSequence))
+            .callOn(symbolProvider.provideMethodSymbol(asStream))
```
abelkov, line 36: It's better to add a test with some comments, they should be preserved. `withFormattingFrom` call may be necessary.

abelkov (changes requested):

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
isuckatcs: Not exact cases, but I've seen that you have this ticket, which I assume, means the handling of other cases too: https://youtrack.jetbrains.com/issue/KTIJ-14393/Convert-Java-8-stream-API-operations-to-Kotlin-sequence-operations

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
isuckatcs: I was trying to keep other cases in mind too, so I called `withFormattingFrom` on a higher level. Should I move it down here?

```kotlin
return recurse(
            when (target) {
                "java.lang.String.lines" -> convertStringLines(element)
                else -> element
            }.withFormattingFrom(element)
        )
```

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
isuckatcs: Without the `import`, we get an `Unresolved reference: asStream` error, and the statement is not added automatically.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
isuckatcs: I added a test for formatting.

abelkov: Yep, I have missed the call on a higher level. All good.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/StreamConversion.kt
abelkov: Well, this issue is a small bug that leads to red code. [KTIJ-14393](https://youtrack.jetbrains.com/issue/KTIJ-14393) is a big feature request to map Stream API to Sequence API. So, it will be handled separately.

abelkov:
Can you please squash your changes into a single commit and write a proper commit message?

abelkov (approved):
Thank you. I want to investigate the problem with imports and then merge this soon with some cleanup.

isuckatcs:
Should imports happen automatically? Let me take a look into it then. Also please let me know what cleanups you have in mind, so I can do those too.

isuckatcs:
After a quick check, it seems like we want to tweak `fun renderSymbol(symbol: JKSymbol, owner: JKTreeElement?): String`, so that it adds the import for this function too.

abelkov:
> After a quick check, it seems like we want to tweak `fun renderSymbol(symbol: JKSymbol, owner: JKTreeElement?): String`, so that it adds the import for this function too.

Yeah, it seems that there is some problem with extension functions here. I believe `testData/newJ2k/function/mathStaticMethods.java` test has the same issue. It would be great if you can investigate it.

abelkov:
I think tests can be put into `javaStreamsApi` directory. And probably we should reduce the number of new test files (some cases can be put into a single file). We are basically testing the use of an existing conversion, not brand new code.

isuckatcs:
It seems like the connection between resolved and unresolved functions is the way, they get printed initially.

```kotlin
class A {
    fun /*@@ljbpzy@@*/foo(x: /*@@ywrlyl@@*/kotlin.Double) {
        kotlin.math.abs(x)                // resolved
        x.nextDown()                      // not resolved
        "test".lineSequence().asStream()  // not resolved
    }
}
```

What I came up with is to add the import if needed inside `renderSymbol()`, though I think `importStorage.isImportNeeded()` is incomplete, because `JKImportStorage.defaultImports` only contains a limited amount of entries. 

Our [docs](https://kotlinlang.org/docs/packages.html#default-imports) says that e.g.: `kotlin.collections.*` is imported, but in `JKImportStorage.defaultImports`, we miss some entries, including `kotlin.collections.reversed`, which leads to J2K adding redundant import statements.

isuckatcs:
Regarding the import statements, we only get the `ImportPath`s from the compiler, which doesn't seem to be enough information for J2K.

```kotlin
private val defaultImports = ... {
    ArrayList<ImportPath>().apply {
        listOf(
            "kotlin.*",
            "kotlin.annotation.*",
            "kotlin.collections.*",
            "kotlin.ranges.*",
            "kotlin.sequences.*",
            "kotlin.text.*",
            "kotlin.io.*"
                ).forEach { add(ImportPath.fromString(it)) }
        ...
    }
}
```
I wonder if it makes sense to return every concrete import instead.

Also we filtered out every "all under" import, which includes everything in the list above, so we completely lost track of these imports. I removed this filtering, and didn't see anything breaking. I guess it was some artifact from the past.

@abelkov please have a look at this new possible way of checking if an import is needed, and if you're happy with the result, I will squash the commits once again.

abelkov (approved):
Thanks! The change looks fine to me. I will separate the commits and merge soon.

isuckatcs:
I split the commits based on the ticket descriptions. I initially created a [different PR](https://github.com/isuckatcs/intellij-community/pull/1/files) for it, but I'm not sure how cross-repo stacked PR-s work on github, so I merged it into this one instead.

abelkov:
Yes, I reassembled the commits manually. You can try to run the K2 J2K test suite and make sure that the tests are passing (and the tests have passed CI). So it should be all good.

PR #2747, opened 2024-04-28
[PY-61315] Fix unresolved function references in stub files

trinhanhngoc:
https://youtrack.jetbrains.com/issues?q=&preview=PY-61315

@east825


trinhanhngoc:
Hello @east825 ,

Can you take a look at my PR and tell me what you think? Thanks!

east825:
~~Merged! Thanks for contributing and sorry for getting to it so late.~~

My bad, I mixed it up with a different patch merged yesterday. This particular problem was addressed by a more general fix in name resolution for overloads implemented around the same time as you commented. Sorry for not getting back and checking yours in time :sweat:

trinhanhngoc:
Hello @east825 ,

No problem. I just feel a bit sad because PyCharm’s type inference is lagging behind other type checkers in both correctness and performance. However, I still hope that you and other developers in the PyCharm team will make a better story for the type inference in PyCharm in the future 😊.

east825:
@trinhanhngoc Thanks. Yes, that's true but lately we've considerably beefed up the team and now moving towards being fully typing spec compliant and overall catching up with other type checkers. But a lot of things need to be revamped and old design choices back from the pre-typing era reconsidered, it's not a quick process, but we're getting there.

PR #2748, opened 2024-04-30
[toml] add `pdm.lock`

mkniewallner:
[PDM](https://github.com/pdm-project/pdm) is a fairly popular Python package manager. It can generate a [lock file named pdm.lock](https://pdm-project.org/en/stable/usage/lockfile/) that uses TOML format ([example](https://github.com/pdm-project/pdm)).


jreznot (approved):

jreznot:
Thank you so much, on the way to master

PR #2750, opened 2024-05-03
[kotlin] K2 J2K: Move LiftReturnInspectionBasedProcessing to JKTree

jocelynluizzi13:
This PR moves `LiftReturnInspectionBasedProcessing` for nonnested return statements into the tree by marking conditionals that always return with `hasLiftedReturn`, then changing the `JKReturnStatement` to a non-return statement, and then printing `return` in the printing phase.

Notably, I ran into errors trying to make the Conditionals into JKReturnStatements prior to printing, because of issues converting between statements and expressions. 

**This does not yet handle nested return statements, so I haven't removed the `LiftReturnInspectionBasedProcessing` post-processing from K1. Putting this up now anyways since it fixes 6 K2 test cases

### DIFF
--- a/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/MarkConditionalStatementsWithLiftedReturnsConversion.kt
+++ b/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/MarkConditionalStatementsWithLiftedReturnsConversion.kt
@@ -1,0 +1,32 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+
+package org.jetbrains.kotlin.nj2k.conversions
+
+import org.jetbrains.kotlin.analysis.api.KtAnalysisSession
+import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
+import org.jetbrains.kotlin.nj2k.RecursiveConversion
+import org.jetbrains.kotlin.nj2k.statements
+import org.jetbrains.kotlin.nj2k.tree.*
+
+class MarkConditionalStatementsWithLiftedReturnsConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
+    context(KtAnalysisSession)
+    override fun applyToElement(element: JKTreeElement): JKTreeElement {
+        if (element !is JKIfElseStatement && element !is JKKtWhenBlock) return recurse(element)
+        if (element is JKIfElseStatement) {
+            if (element.elseBranch.statements.isNotEmpty() && element.thenBranch.statements.isNotEmpty() &&
+                element.thenBranch.statements.all {it is JKReturnStatement} && element.elseBranch.statements.all {it is JKReturnStatement}
+            ) {
+                element.hasLiftedReturn = true
+            }
+        }
+        else if (element is JKKtWhenBlock) {
+            element.hasLiftedReturn = when {
+                element.cases.isEmpty() -> false
+                element.cases.none { it.labels.isNotEmpty() && it.labels.any { label -> label is JKKtElseWhenLabel } } -> false
+                else -> element.cases.all { case -> case.statement is JKReturnStatement }
+            }
+        }
+
+        return recurse(element)
+    }
+}
--- a/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
+++ b/plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
@@ -13,7 +13,10 @@ import org.jetbrains.kotlin.nj2k.tree.JKClass.ClassKind.*
 import org.jetbrains.kotlin.nj2k.tree.Modality.FINAL
 import org.jetbrains.kotlin.nj2k.tree.Visibility.PUBLIC
 import org.jetbrains.kotlin.nj2k.tree.visitors.JKVisitorWithCommentsPrinting
-import org.jetbrains.kotlin.nj2k.types.*
+import org.jetbrains.kotlin.nj2k.types.JKContextType
+import org.jetbrains.kotlin.nj2k.types.isAnnotationMethod
+import org.jetbrains.kotlin.nj2k.types.isInterface
+import org.jetbrains.kotlin.nj2k.types.isUnit
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
 class JKCodeBuilder(context: NewJ2kConverterContext) {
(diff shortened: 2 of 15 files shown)


plugins/kotlin/j2k/k1.new.post-processing/src/org/jetbrains/kotlin/idea/j2k/post/processing/allProcessings.kt
```diff
@@ -105,7 +105,6 @@ private val inspectionLikePostProcessingGroup = InspectionLikeProcessingGroup(
     inspectionBasedProcessing(ReplaceGetOrSetInspection(), checkInspectionIsEnabled = false),
     intentionBasedProcessing(ObjectLiteralToLambdaIntention(), writeActionNeeded = true),
     DestructureForLoopParameterProcessing(),
-    LiftReturnInspectionBasedProcessing(),
```
abelkov, line 108: Just in case, have you checked all the tests of LiftReturnOrAssignmentInspection (plugins/kotlin/idea/tests/testData/inspectionsLocal/liftOut)? I see that you have not added any new J2K tests, but I doubt that we have great existing test coverage here. Some tests should probably be ported.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/MarkConditionalStatementsWithLiftedReturnsConversion.kt
```diff
@@ -0,0 +1,33 @@
...
+
+package org.jetbrains.kotlin.nj2k.conversions
+
+import org.jetbrains.kotlin.analysis.api.KtAnalysisSession
+import org.jetbrains.kotlin.nj2k.NewJ2kConverterContext
+import org.jetbrains.kotlin.nj2k.RecursiveConversion
+import org.jetbrains.kotlin.nj2k.statements
+import org.jetbrains.kotlin.nj2k.tree.*
+
+class MarkConditionalStatementsWithLiftedReturnsConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
```
abelkov, line 11: It is better to add documentation to all new conversions.

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/MarkConditionalStatementsWithLiftedReturnsConversion.kt
```diff
@@ -0,0 +1,33 @@
...
+    override fun isEnabledInBasicMode(): Boolean = false
+    context(KtAnalysisSession)
+    override fun applyToElement(element: JKTreeElement): JKTreeElement {
+        if (element !is JKIfElseStatement && element !is JKKtWhenBlock) return recurse(element)
```
abelkov, line 15: Original inspection also handles `try`

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
```diff
@@ -367,8 +370,10 @@ class JKCodeBuilder(context: NewJ2kConverterContext) {
             }
         }
 
-
         override fun visitIfElseStatementRaw(ifElseStatement: JKIfElseStatement) {
+            if (ifElseStatement.hasLiftedReturn) {
+                printer.print("return ")
```
abelkov, line 375: This is very questionable. `return` statements should be represented as real JK elements, not printed out of thin air in the code builder.

Let's try to refactor this into a single conversion with an obvious algorithm. It should create a new return statement wrapping the if/while/try and detach all inner return statements (preserving the comments/formatting). If you meet some particular problem with this approach, let's discuss it with a concrete code example.

plugins/kotlin/j2k/shared/tests/testData/newJ2k/switch/noLiftedReturnInBasicMode.java
```diff
@@ -0,0 +1,12 @@
...
+public String foo(int i, int j) {
+        switch (i) {
+        case 0:
+        return "1";
+        case 1:
+        return "3";
+default:
+        return "4";
+        }
+        }
```
abelkov, line 12: Formatting should be fixed

abelkov (changes requested):

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/LiftReturnsConversion.kt
```diff
@@ -0,0 +1,58 @@
...
+ */
+class LiftReturnsConversion(context: NewJ2kConverterContext) : RecursiveConversion(context) {
+    override fun isEnabledInBasicMode(): Boolean = false
+
+    context(KtAnalysisSession)
+    override fun applyToElement(element: JKTreeElement): JKTreeElement {
+        return when (element) {
+            is JKIfElseStatement -> {
+                element.hasLiftedReturn = element.liftReturnStatement()
+                // ideally we want to return JKReturnStatement() instead of using the "hasLiftedReturn" printer
```
jocelynluizzi13, line 26: @abelkov Do you have any advice/suggestions to achieve this goal? Right now anytime I try to return `JKReturnStatement(element)` or some variation of this I'm running into "check" issues, or else `element` is not in the correct format

plugins/kotlin/j2k/shared/tests/testData/newJ2k/issues/kt-824-isDir.kt
```diff
@@ -11,8 +11,6 @@ object Test {
             return false
         }
         val result = true
-        return if (parent.isDirectory) {
-            true
-        } else false
+        return if (parent.isDirectory) true else false
```
abelkov, line 14: All curly blocks, formatting and comments of nested elements need to be preserved (and we should make sure to have tests to check that)

plugins/kotlin/j2k/shared/tests/testData/newJ2k/liftReturns/tryCatchStatement.kt
```diff
@@ -0,0 +1,9 @@
+internal class Test {
+    fun test(): String {
+        return try {
```
abelkov, line 3: This is questionable from a code style perspective. I believe the inspection doesn't highlight such code as a warning, so it should not be automatically converted by J2K as well. I think lifting the return out of `try` only makes sense if both `try` and `catch` blocks contain exactly one return statement and nothing else.

abelkov (changes requested):

plugins/kotlin/j2k/shared/src/org/jetbrains/kotlin/nj2k/conversions/LiftReturnsConversion.kt
abelkov: Maybe @ermattt can help here as well? You can try debugging the code together and see if "check failed" errors can be avoided somehow.

PR #2751, opened 2024-05-03
[kotlin] K2 J2K: Move Explicit API Mode Visibility Checks to JKTree

jocelynluizzi13:
Delete `NO_EXPLICIT_VISIBILITY_IN_API_MODE` from the postprocessing step and handle it in the `modifers.kt` redundancy check. `PUBLIC` should not be redundant in explicit API mode, so mark it as such.


abelkov (approved):
LanguageVersionSettings can be passed down from the converter context, I will simplify it.

PR #2753, opened 2024-05-10
fix wrong reported conflicting PsiElement

rendner:
Hi, I noticed a wrong reported `PsiElement` in `RenamePyElementProcessor::findExistingNameConflicts`.


east825:
Merged, thanks for noticing! Sorry for not getting to the PR for so long.

PR #2754, opened 2024-05-10
Check for project disposal before adapter.createInstance

jenna-h:


jenna-h:
I think I may not have been following appropriate procedure before, so I apologize for that! I encountered this problem through my work at Google and recently asked my teammates about how best to upstream changes to IntelliJ.

At their suggestion, I filed a bug on YouTrack that has some more context about the issue: https://youtrack.jetbrains.com/issue/IDEA-354912/adapter.createInstance-in-EditorNotificationsImpl-can-throw-AlreadyDisposedException 

Thank you very much for your consideration! Please let me know if you have any questions, comments, or concerns.

develar:
Thanks, sorry for the delay, the full fix is applied as a part of https://youtrack.jetbrains.com/issue/IJPL-156594.

PR #2755, opened 2024-05-12
Add JAR manifest header folding

mkem114:
Add folding support to make it easier to wrestle with large files. After this PR I'll add structure support though

### DIFF
--- a/java/manifest/src/org/jetbrains/lang/manifest/folding/ManifestFoldingBuilder.java
+++ b/java/manifest/src/org/jetbrains/lang/manifest/folding/ManifestFoldingBuilder.java
@@ -1,0 +1,78 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.lang.manifest.folding;
+
+import com.intellij.lang.ASTNode;
+import com.intellij.lang.folding.FoldingBuilderEx;
+import com.intellij.lang.folding.FoldingDescriptor;
+import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.project.DumbAware;
+import com.intellij.openapi.util.TextRange;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiRecursiveElementVisitor;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.lang.manifest.psi.Header;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ManifestFoldingBuilder extends FoldingBuilderEx implements DumbAware {
+
+  /**
+   * The <a href="https://docs.oracle.com/en/java/javase/22/docs/specs/jar/jar.html#notes-on-manifest-and-signature-files">JAR file
+   * specification</a> states each line should be no longer than 72 bytes
+   */
+  private static final int MAX_BYTES_PER_LINE = 72;
+  private static final String ELLIPSIS = "...";
+
+  @Override
+  public FoldingDescriptor @NotNull [] buildFoldRegions(@NotNull PsiElement root, @NotNull Document document, boolean quick) {
+    List<FoldingDescriptor> descriptors = new ArrayList<>();
+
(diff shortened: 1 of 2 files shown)


mkem114:
How should I go about testing? It's a bit of mish-mash looking around at other files

java/manifest/src/org/jetbrains/lang/manifest/folding/ManifestFoldingBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+
+    return descriptors.toArray(FoldingDescriptor.EMPTY_ARRAY);
+  }
+
+  @Nullable
+  @Override
+  public String getPlaceholderText(@NotNull ASTNode node) {
+    if (node.getPsi() instanceof Header) {
+      String usefulText = node.getText().trim();
+      if (usefulText.length() <= MAX_BYTES_PER_LINE) {
```
mkem114, line 53: Seems like a useful proxy, the viewer cuts off placeholders that are too long with soft-wrapping (well you can still horizontally scroll).

In this case, adding more detail in the placeholder shouldn't add more cognitive load

java/manifest/src/org/jetbrains/lang/manifest/folding/ManifestFoldingBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+      int endIndex = Math.min(MAX_BYTES_PER_LINE - ELLIPSIS.length(), usefulText.length());
+      return node.getText().substring(0, endIndex) + ELLIPSIS;
+    }
+
+    return null;
+  }
+
+  @Override
+  public boolean isCollapsedByDefault(@NotNull ASTNode node) {
+    return false; // no change in existing behaviour
```
mkem114, line 65: I think we can avoid this discussion for now. What the heuristic should be, should probably be discussed in an issue IMO

mkem114:
@sirgl Can I please get a review?

java/manifest/src/org/jetbrains/lang/manifest/folding/ManifestFoldingBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private static int getTextLengthWithoutFinalNewline(PsiElement psiElement) {
+    int lengthOfTextWithoutFinalNewline = psiElement.getTextLength();
+    if (psiElement.getText().endsWith("\n") || psiElement.getText().endsWith("\r")) {
```
amaembo, line 70: I think our text file model normalizes line breaks already, so you don't need to check for `\r` or `\r\n`. Am I missing something? Do you know examples where handling \r\n is necessary?

java/manifest/src/org/jetbrains/lang/manifest/folding/ManifestFoldingBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+  @Nullable
+  @Override
+  public String getPlaceholderText(@NotNull ASTNode node) {
+    if (node.getPsi() instanceof Header) {
+      String usefulText = node.getText().trim();
+      if (usefulText.length() <= MAX_BYTES_PER_LINE) {
+        return node.getText();
+      }
+      int endIndex = Math.min(MAX_BYTES_PER_LINE - ELLIPSIS.length(), usefulText.length());
+      return node.getText().substring(0, endIndex) + ELLIPSIS;
```
amaembo, line 57: We have methods like `com.intellij.openapi.util.text.StringUtil#shortenTextWithEllipsis`. Could they be useful here?

amaembo (commented):
Hello and sorry that the review took so long.

It would be great if you submit an issue to our [YouTrack](https://youtrack.jetbrains.com/newIssue?project=IDEA), showing screenshots how it will look after your changes, so we can discuss with our UX team whether we want such a change in IntelliJ IDEA. Thank you in advance.

If the change is approved, it will also be great if you can add at least one unit-test? You can use `com.intellij.openapi.editor.impl.JavaFoldingTest` as an example.

BartvHelvert:
Since there was no activity here for a long time, I am going to close this PR. Feel free to open another one if you still want to address the review comments.

PR #2757, opened 2024-05-13
Add OSGi headers to JAR manifest parser

mkem114:
Makes the IDE squiglies more useful since common headers will be recognized


amaembo:
I think similar functionality is provided by an OSGi plugin
https://github.com/JetBrains/intellij-community/pull/2757
I don't feel that we need to duplicate it right in the IntelliJ Community
Though probably we need to support more headers in OSGi plugin. @trespasserw please take a look, as I see that OSGi plugin was maintained by you.

mkem114:
> I think similar functionality is provided by an OSGi plugin

Did the performance and bugginess problems with that plugin get resolved? Honestly seemed kinda dead.

trespasserw:
Admittedly, the OSGi plugin is rather on life support due to lack of resources and the technology being not wildly popular. But as for me, it's an insufficient reason for duplicating functionality.

PR #2760, opened 2024-05-17
[IJPL-149990] Allow custom defaults for CodeVisionSettings

andymagee:
Fixes https://youtrack.jetbrains.com/issue/IJPL-149990. This solution is proposed as one way to fix the issue. For Android Studio we would be open to other mechanisms if there's a better way to handle this.


trushev:
Thank you for your contribution, the commit is merged into the master branch https://github.com/JetBrains/intellij-community/commit/6de7daf202db864746269ec32b0f23abf7e80ab0

trushev (approved):

PR #2761, opened 2024-05-17
KTIJ-29850: Do not suggest privatizing functions referenced by callable references

freya022:
Fixes https://youtrack.jetbrains.com/issue/KTIJ-29850

Callable references are considered as usages outside of the class, as calling them reflectively causes an `IllegalAccessException`


cdracm:
thanks, pushed

PR #2765, opened 2024-05-21
Add Android Framework and Internal Classes to DEFAULT_STEPPING_FILTERS

alonalbert:
Android Studio team would like these classes to be added to `DebuggerSetting.getSteppingFilters()`.

```
android.*
com.android.*
androidx.*
libcore.*
dalvik.*
```

While it's possible to them dynamically only when Android Plugin is enabled (via a plugin ProjectService for example, it would simplify things if we added it to the default because then, the Debugger Tests running on ART which was added in https://github.com/JetBrains/intellij-community/pull/2744 would work out of the box without having to dynamically apply these filters.


zuevmaxim:
I think that it would be better to add these filters in Android specific code, rather than platform code.
Please consider implementing something similar to `org.jetbrains.kotlin.idea.debugger.core.filter.JvmDebuggerAddFilterStartupActivity`

PR #2767, opened 2024-05-24
Migrate from FileComparisonFailure to FileComparisonFailedError in testFramework

abrooksv:
The testFrameworks package was using `FileComparisonFailure` for asserts which comes from the `intellij.java.rt` module. 

In the newer broken up test frameworks in intellij-platform-gradle-plugin, the usage of the test frameworks could lead to ClassNotFoundExceptions on this class when testing against the smaller IDEs.

This PR tries to replace the usages of the deprecated class to cut down on the reliance of `intellij.java.rt` inside of `intellij.platform.testFramework`.

`DynamicExtensionPointsTester` also appears to reach into `intellij.java.rt` for `com.intellij.rt.execution.junit.MapSerializerUtil` and that usages looked unsafe to touch for someone with no knowledge of what those print statements power

`FileComparisonData` was dropped from `FileComparisonFailedError` to rely only upon opentest4j

See https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/1632

### DIFF
--- a/java/testFramework/src/com/intellij/codeInsight/JavaCodeInsightTestCase.java
+++ b/java/testFramework/src/com/intellij/codeInsight/JavaCodeInsightTestCase.java
@@ -28,6 +28,7 @@ import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.vfs.newvfs.impl.VfsRootAccess;
+import com.intellij.platform.testFramework.core.FileComparisonFailedError;
 import com.intellij.pom.java.LanguageLevel;
 import com.intellij.psi.PsiClass;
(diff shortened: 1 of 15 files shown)


HackerMadCat:
Unfortunately, I cannot accept this pull request, because this migration is a lot more complicated thing than the simple refactoring. We need to recheck several things.

For example:
* Our remote tests control their classpath explicitly;
* The pure FileComparisonFailedError has been supported only since IDEA 2024.1;
* The classes in the idea-rt.jar can be resolved in runtime to any idea-rt.jar on the computer;

Please follow up on the issue https://youtrack.jetbrains.com/issue/IJPL-156277
The first step is done in the issue https://youtrack.jetbrains.com/issue/IJPL-155979

abrooksv:
@HackerMadCat I figured, it was more just to call attention to it then to be able to finish the migration without insider knowledge. Thank you for working on IJPL-156277

PR #2768, opened 2024-05-26
KTIJ-14684: Allow sam-with-receiver Kotlin compiler module to work in test source sets

eviltak:
Resolves https://youtrack.jetbrains.com/issue/KTIJ-14684. 

Quite a straightforward fix - Previously the implementation of `IdeSamWithReceiverComponentContributor::registerModuleComponents` added a `SamWithReceiverResolverExtension` to the component container only `ModuleProductionSourceInfo` modules. I changed it to add the extension for general `ModuleSourceInfo` modules (which includes both `ModuleProductionSourceInfo` and `ModuleTestSourceInfo` modules).

To test, open the [sam-receiver-kotlin-bug](https://github.com/big-guy/sam-receiver-kotlin-bug) (see https://youtrack.jetbrains.com/issue/KTIJ-22591) Gradle project in the IDE and navigate to the `ImplTest.kt` file in the `test` source root. The references in the lambda body in the `onFoo` method call should resolve properly.


vladimirdolzhenko:
Sorry, we focus on K2 part of plugin, K1 support and its further development is quite limited.

Could you please check it out for K2 ?

fedochet:
Hi @eviltak ! 

Have you been able to try out K2 Mode in Kotlin IntelliJ Plugin and see if the original problem persists there?

As @vladimirdolzhenko said, we're currently focused on K2 Mode, and there the "default" compiler plugins (including sam-with-receiver) actually work "out of the box" in the IDE, so the experience should be better already

PR #2771, opened 2024-06-05
Don't group delete line actions for undo

kazariad:
Currently consecutive "Delete Line" actions are grouped and reverted together by a single Undo, however this is unintuitive and inconsistent with the behavior of other line-based actions like duplicate line or cut, which are reverted 1-by-1.  A practical example  of this problem is tapping the delete line hotkey to quickly remove multiple lines but accidentally deleting 1 extra line, then hitting undo and having the entire block restored instead of just the last line.

https://youtrack.jetbrains.com/issue/IJPL-52352

Before:
before

After:
after


cmeeren:
Would love to see this merged and published!

PR #2773, opened 2024-06-06
[Kotlin debugger] Add bytecode inspection for ART in KotlinSyntheticTypeComponentProvider 

nikita-nazarov:
When trying to detect synthetic methods, KotlinSyntheticTypeComponentProvider uses bytecode inspection to verify that a method simply delegates to a default implementation in an interface. This inspection didn't work for ART, because bytecode on ART is different. This commit implements ART specific bytecode inspection and by that improves stepping behaviour in Android Studio. For example in this test case: `afterDefaultParameterValuesIntf.kt`.

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/kotlin.jvm-debugger.core.iml
+++ b/plugins/kotlin/jvm-debugger/core/kotlin.jvm-debugger.core.iml
@@ -53,5 +53,7 @@
     <orderEntry type="module" module-name="intellij.java.debugger.impl" />
     <orderEntry type="module" module-name="intellij.java.psi.impl" />
     <orderEntry type="module" module-name="kotlin.code-insight.utils" />
+    <orderEntry type="library" name="google-dexlib2" level="project" />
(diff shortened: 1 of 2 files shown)


plugins/kotlin/jvm-debugger/core/kotlin.jvm-debugger.core.iml
```diff
@@ -53,5 +53,7 @@
     <orderEntry type="module" module-name="intellij.java.debugger.impl" />
     <orderEntry type="module" module-name="intellij.java.psi.impl" />
     <orderEntry type="module" module-name="kotlin.code-insight.utils" />
+    <orderEntry type="library" name="google-dexlib2" level="project" />
```
zuevmaxim, line 56: This library is currently used only in Android plugin. What do you think of adding an extension point with implementation in android plugin?

nikita-nazarov:
We personally discussed that we want to add a bytecode inspection extension point, so I close this pr as the issue will be resolved in another one

PR #2774, opened 2024-06-06
jps-bootstrap: allow customizing user.home

estebandlc:
... by using the HOME environment variable

### DIFF
--- a/platform/jps-bootstrap/jps-bootstrap.sh
+++ b/platform/jps-bootstrap/jps-bootstrap.sh
@@ -155,7 +155,7 @@ set -x
 
 if [ "x${JPS_BOOTSTRAP_PREPARE_SKIP_DOWNLOAD:-0}" != "x1" ] ; then
   # Download and compile jps-bootstrap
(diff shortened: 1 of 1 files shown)


estebandlc:
Hi, Android Studio engineer here. Given that the build scripts download data to the home directory, our build servers need to be able to control downloads and locations. This seemed like a non intrusive way of controlling where the maven downloads go. Thanks!

abusalimov:
@estebandlc Thanks for the contribution!
Could you wrap the `$HOME` args into double quotes to avoid escaping issues in case the variable has whitespaces in it? Also I'll appreciate if you add an explanatory comment for why it's needed, I think adding right before the last call to java should suffice. Thank in advance!

abusalimov (changes requested):
@estebandlc Thanks for the contribution!
Could you wrap the `$HOME` args into double quotes to avoid escaping issues in case the variable has whitespaces in it? Also I'll appreciate if you add an explanatory comment for why it's needed, I think adding right before the last call to java should suffice. Thank in advance!

estebandlc:
Thank you for the review! Done.

PR #2775, opened 2024-06-07
Fix PY-44858

MaxLevitsky:
Added method `isInNamespacePackage` and variable `isNamespace` to classes in "Move", that used for separation between two cases (Namespace and ordinary python package)


python/src/com/jetbrains/python/refactoring/move/moduleMembers/PyMoveModuleMembersProcessor.java
```diff
@@ -81,8 +83,9 @@ protected void performRefactoring(final UsageInfo @NotNull [] usages) {
     for (UsageInfo usage : usages) {
       usagesByElement.putValue(((MyUsageInfo)usage).myMovedElement, usage);
     }
+    boolean isNamespace = isInNamespacePackage(this.mySourceFiles);
```
east825, line 86: `this` qualifier is not necessary here.

python/python-psi-impl/src/com/jetbrains/python/refactoring/classes/PyClassRefactoringUtil.java
```diff
@@ -518,8 +521,31 @@ public static void optimizeImports(@NotNull final PsiFile file) {
     PyImportOptimizer.onlyRemoveUnused().processFile(file).run();
   }
 
+
+  public static PsiFile placeFile(Project project, String path, String filename) throws IOException {
```
east825, line 525: This overload of `placeFile` is still used in `PyExtractSuperclassHelper` and in the "Extract superclass" refactoring, which still causes the same problem with unnecessary `__init__.py` created inside namespace packages. Please take namespace packages into account there as well, add regression tests and remove this overload entirely.

python/testSrc/com/jetbrains/python/refactoring/PyMoveTest.java
```diff
@@ -427,6 +427,38 @@ public void testMoveSymbolDoesntCreateInitPyInSourceRoot() {
...
+  //PY-44858
+  public void testMoveNotCreateInitPyForNamespacePackagesToParentDirectory() {
+    doComparingDirectories(testDir -> {
+      final VirtualFile sourceRoot = testDir.findFileByRelativePath("src");
+      runWithSourceRoots(Collections.singletonList(sourceRoot), () -> moveSymbols(testDir, "src/pkg/subpkg/B/module_b.py", "myfunc"));
+    });
+  }
+
+  //PY-44858
+  public void testMoveNotCreateInitPyForNamespacePackagesToChildDirectory() {
```
east825, line 455: For this kind of test scenarios you don't need namespace packages to be inside an explicit source root. So inside the test data intermediate "src" root could be removed and the test code could be simplified to:
```java
//PY-44858
public void testMoveNotCreateInitPyForNamespacePackagesToChildDirectory() {
  doMoveSymbolsTest("pkg/subpkg/A/B/module_b.py", "myfunc");
}
```
I mean the following "flattened" hierarchy in test data:
```
community/python/testData/refactoring/move/moveNotCreateInitPyForNamespacePackagesToChildDirectory/
├── after
│   └── src
│       └── pkg
│           └── subpkg
│               └── A
│                   ├── B
│                   │   └── module_b.py
│                   └── module_a.py
└── before
    └── src
        └── pkg
            └── subpkg
                └── A
                    └── module_a.py

```

python/src/com/jetbrains/python/refactoring/move/makeFunctionTopLevel/PyBaseMakeFunctionTopLevelProcessor.java
```diff
@@ -89,7 +89,7 @@ protected final void performRefactoring(UsageInfo @NotNull [] usages) {
 
     assert ApplicationManager.getApplication().isWriteAccessAllowed();
```
east825, line 91: Same problem as with "Extract superclass". This processor is responsible for refactorings moving a method or a local function to the top-level of a file (triggered by the normal Move action). If the original function is inside a namespace package and is going to be moved to a different file, unnecessary `__init__.py` will be created. Please add tests on that (ideally both for "Make method top-level" and "Make local function top-level").

east825 (changes requested):

python/src/com/jetbrains/python/refactoring/move/makeFunctionTopLevel/PyBaseMakeFunctionTopLevelProcessor.java
```diff
@@ -241,6 +244,16 @@ private boolean isFromEnclosingScope(@NotNull PsiElement element) {
            !(ScopeUtil.getScopeOwner(element) instanceof PsiFile); 
   }
 
+  private static boolean isFunctionInNamespacePackage(PyFunction myFunction){
```
mano7onam, line 247: here need @NotNull check for PyFunction myFunction

python/testSrc/com/jetbrains/python/refactoring/classes/extractSuperclass/PyExtractSuperclassTest.java
```diff
@@ -172,15 +234,15 @@ private void doSimpleTest(final String className,
   }
 
   public void testMultifileNew() {
-    String baseName = "refactoring/extractsuperclass/multifile/";
-    myFixture.configureByFile(baseName + "source.py");
+    String baseName = "refactoring/extractsuperclass/multifileNew/";
```
east825, line 237: Why was this test updated? It looks unrelated.

python/testSrc/com/jetbrains/python/refactoring/classes/extractSuperclass/PyExtractSuperclassTest.java
```diff
@@ -134,6 +135,67 @@ public void testAbstractMethodDocStringIndentationPreserved() {
     });
   }
 
+  // PY-44858
+  public void testExtractNotCreateInitInAnotherDir(){
```
east825, line 139: These tests don't test anything. If you revert all the changes from `PyExtractSuperclassHelper`, they still pass. The reason is that they assert not the structure of the *actual* "Extract superclass" result, but the structure of their own *expected* result in `doTestNotCreateInitCommon`. 

For instance, in each of these tests the file where a superclass is expected to be placed is called "suppa.py" according to the test data, but it's not where the test actually puts it. It's always an `__init__.py` file inside a new package, e.g. in this test the superclass ends up in in `mypkg/b/__init__.py`, but you check the structure of the directories `a` and `b` copied from the test data verbatim. Please add tests that behave like scenarios in PyMoveTest, comparing the project directory before and after the refactoring.

east825: After the refactoring is performed the test project directory looks like this (with the bug fix code reverted):
```
mypkg/
  b/
    __init__.py # Suppa is here
  __init__.py
a/
  source.py
b/
  suppa.py      
```
and this test doesn't fail.

east825 (changes requested):

east825:
Closing since the changes have been squashed and integrated into master in https://github.com/JetBrains/intellij-community/commit/90bbc21f3fed4f43ef8d97f21335412e44cfeb61.

burnpanck:
Has this already found its way into a released version of PyCharm? I just hit this **%ç&*çç%*"ç% bug again in PyCharm 2024.2.3 Professional, Build #PY-242.23339.19, built on September 25 2024...

burnpanck:
@east825 I already don't remember anymore with 100 % accuracy, but I think it was a move refactor of some python modules between sub-packages within the same namespace package.

However, I realised now that PyCharm has the option to explicitly mark a folder as a namespace package; the corresponding folder wasn't marked as such. I presume that this fix here only applies to explicitly marked namespace packages. While I am still convinced that it is always wrong for PyCharm to (*silently!*) create `__init__.py` silently in directories that are not directly touched by a user-initiated operation (especially if intermediate folders already have an `__init__.py`), explicitly marking directories as namespace package is workable for me, and may be good enough.

burnpanck:
Ok, so now it has hit me again.

Before the refactor:
```
src/
# marked as library root in PyCharm
- company_namespace/
    # no __init__.py here; marked as namespace package in PyCharm
  - componentX/
    - __init__.py
    - partA/
      - __init__.py
      - module.py
    - partB/
      - __init__.py
```
Then, I initiated a move refactor by dragging `module.py` from `partA` to `partB`. PyCharm creates an unwanted `__init__.py` in `company_namespace/`. Luckily, it asks me if I want to add it to git - because otherwise, I would have not noticed, and only learned about this after `componentX` makes it into a release, gets installed, and now imports from `componentY` (managed in another project) fail. At this point you assume the problem is with `componentY`, and there is absolutely no hint to `componentX`.

Seriously people - the namespace package feature is now over 10 years old and this still doesn't work 🤬😡. The fallout of this bug is immensely painful.

PR #2778, opened 2024-06-12
Don't call `get` on a null `myType2Facets` map.

jenna-h:


ikorennoy:
Hi @jenna-h, thank you for the fix.

PR #2779, opened 2024-06-13
[kotlin] J2K: Always add Override Keyword when @Override Present

jocelynluizzi13:
Meta is frequently having issues where `@Override` gets stripped when it cannot find the method we're overriding. (`ERROR: 'onFragmentCreate' overrides nothing`)

However, we should still include these `override` modifiers here and assume the Java is correct in it being an `@Override`


abelkov:
1. What happens if the user has a custom `foo.bar.Override` annotation? Can we check by FQN?
2. How do we handle these Override annotations in general? Do they become annotations in JK tree and then removed somewhere?

jocelynluizzi13:
Updated to use fqName and modify the file in the location we are already modifying. Rather than removing the `@Override` annotation in one step and adding the `override` modifier later, simply do it all in the same step (as well as checking if there's a superMethod that wasn't annotated)

abelkov (approved):

abelkov:
Merged in https://github.com/JetBrains/intellij-community/commit/c3d7158d97ec0ccdceaa0c964acaa3221af39759

PR #2780, opened 2024-06-17
Generalize tab pinning logic

YustusKevyn:
This commit moves some of the logic for pinning tabs from `com.intellij.openapi.fileEditor` to `com.intellij.ui.tabs` and generalizes it.

Pinning tabs is a fundamental functionality that should be implemented in a more generalized manner, allowing other parts of the codebase or plugins utilizing tabs (via JBTabs) to leverage and expand upon this feature. Since the logic for sorting and grouping pinned tabs already resides within JBTabsImpl, maintaining parts of the core pinning logic across multiple packages seems somewhat illogical and inconsistent.

This change would also facilitate the implementation of additional tab-related features. For instance, adding new navigation actions to JBTabsImpl that allow tabs to be moved via keyboard shortcuts could require setting whether a tab is pinned in certain circumstances (such as moving an unpinned tab behind a pinned tab when grouping is enabled).

Additionally, this change resolves a bug where the previous position of pinned tabs was ignored when restoring closed tabs. I did not create a report on YouTrack, as this is a minor issue and the fix is simply a nice side effect of this change.

### DIFF
--- a/platform/platform-api/src/com/intellij/ui/tabs/TabInfo.kt
+++ b/platform/platform-api/src/com/intellij/ui/tabs/TabInfo.kt
@@ -6,7 +6,10 @@ import com.intellij.openapi.actionSystem.ActionGroup
 import com.intellij.openapi.editor.markup.TextAttributes
 import com.intellij.openapi.ui.Queryable
 import com.intellij.openapi.util.NlsContexts
-import com.intellij.ui.*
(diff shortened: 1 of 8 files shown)


YustusKevyn:
Mentioning @develar, as he seems to have done some extensive work on the code in question recently

PR #2781, opened 2024-06-17
[git-stash] navigate to the first stash under the root from the context when "Unstash Changes..." action is called

juliabeliaeva:
This PR addresses the problem described in [IJPL-156748](https://youtrack.jetbrains.com/issue/IJPL-156748/Unstash-Changes-does-not-navigate-to-the-corresponding-root) by selecting the first stash in the target repository root in the stashes tree.


juliabeliaeva:
Api dumps need to be regenerated, since there is no way for me to do this for the community.

PR #2786, opened 2024-06-27
[platform] IJPL-832 Explicitly indicate in editor that the file is in generated sources root

wgasiorek:
Fixes issue described here: [[IJPL-832] Explicitly indicate in editor that the file is in generated sources root](https://youtrack.jetbrains.com/issue/IJPL-832/Explicitly-indicate-in-editor-that-the-file-is-in-generated-sources-root)

### DIFF
--- a/platform/analysis-api/src/com/intellij/openapi/roots/GeneratedSourcesFilter.kt
+++ b/platform/analysis-api/src/com/intellij/openapi/roots/GeneratedSourcesFilter.kt
@@ -1,25 +1,45 @@
 // Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
 package com.intellij.openapi.roots
 
-import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.ReadAction
 import com.intellij.openapi.extensions.ExtensionPointName
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.NlsContexts
 import com.intellij.openapi.util.NlsSafe
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.psi.PsiElement
+import com.intellij.util.concurrency.annotations.RequiresReadLock
 import org.jetbrains.annotations.ApiStatus
+
 
 abstract class GeneratedSourcesFilter {
   companion object {
     @JvmField
     val EP_NAME: ExtensionPointName<GeneratedSourcesFilter> = ExtensionPointName("com.intellij.generatedSourcesFilter")
 
+    /**
+     * Checks whether the given files is considered a generated source file in the context of the given project.
+     *
+     * This method can block, if a write action is currently pending.
+     *
+     * @return {@code true} if the file is considered a generated source, {@code false} otherwise.
+     */
     @JvmStatic
     fun isGeneratedSourceByAnyFilter(file: VirtualFile, project: Project): Boolean {
-      return ApplicationManager.getApplication().runReadAction<Boolean, RuntimeException> {
-        !project.isDisposed && file.isValid && EP_NAME.extensionList.any { it.isGeneratedSource(file, project) }
-      }
+      return ReadAction.compute<Boolean, RuntimeException> { findFirstMatchingFilter(file, project) != null }
+    }
+
+    /**
+     * Returns the first source filter (in order of generated source filter extension traversal order) that considers this file
+     * to be generated. This method is non-blocking itself, but has to be called in the context of a read lock.
+     *
+     * @return the first `GeneratedSourcesFilter` that considered this file to be generated, `null` if none matches.
+     */
+    @RequiresReadLock
+    fun findFirstMatchingFilter(file: VirtualFile, project: Project): GeneratedSourcesFilter? {
+      if (project.isDisposed || !file.isValid) return null
+
+      return EP_NAME.extensionList.find { it.isGeneratedSource(file, project) }
     }
   }
 
@@ -54,4 +74,3 @@ data class GeneratedSourceFilterNotification(@NlsSafe val text: String, val acti
 @ApiStatus.Experimental
 @ApiStatus.Internal
 data class GeneratedSourceFilterHyperLinkAction(@NlsSafe val text: String, val link: String)
-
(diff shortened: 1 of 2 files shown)


platform/analysis-api/src/com/intellij/openapi/roots/GeneratedSourcesFilter.java
```diff
@@ -16,16 +16,31 @@
...
+    return ReadAction.compute(() -> findFirstMatchingFilterNonBlocking(file, project) != null);
+  }
+
+  /**
+   * Returns the first source filter (in order of generated source filter extension traversal order) that considers this file
+   * to be generated. This method is non-blocking.
+   *
+   * @return the first {@code GeneratedSourcesFilter} that considered this file to be generated.
+   */
+  public static GeneratedSourcesFilter findFirstMatchingFilterNonBlocking(@NotNull VirtualFile file, @NotNull Project project) {
```
trushev, line 36: Let's mark it returning `@Nullable`

trushev, line 36: Since it is not a private method now, it is better to either acquire read action or mark `@RequiresReadLock`

platform/lang-impl/src/com/intellij/ide/GeneratedFileEditingNotificationProvider.java
```diff
@@ -16,33 +15,25 @@
...
   @Override
   public @Nullable Function<? super @NotNull FileEditor, ? extends @Nullable JComponent> collectNotificationData(@NotNull Project project,
                                                                                                                  @NotNull VirtualFile file) {
-    if (!GeneratedSourceFileChangeTracker.getInstance(project).isEditedGeneratedFile(file)) return null;
+    var matchingFilter = GeneratedSourcesFilter.findFirstMatchingFilterNonBlocking(file, project);
+
+    if (matchingFilter == null) return null;
 
-    String notificationText = getText(file, project);
+    String notificationText = firstNonNull(
```
trushev, line 28: Would be nice to use platform's util method `com.intellij.util.ObjectUtils.notNull`

trushev (changes requested):

platform/lang-impl/src/com/intellij/ide/GeneratedFileEditingNotificationProvider.java
jaen: I assume this is because of the `Contract`s for static verification?

platform/analysis-api/src/com/intellij/openapi/roots/GeneratedSourcesFilter.java
jaen: Any opinion on which is better? I have originally explicitly made it non-blocking (hence the name), because I assumed that notification providers should do as little as possible. Would `@RequiresReadLock` be a better option in that case? Also, is there anything I would have to do to check this?

platform/lang-impl/src/com/intellij/ide/GeneratedFileEditingNotificationProvider.java
```diff
@@ -16,33 +15,25 @@
 import javax.swing.*;
 import java.util.function.Function;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
+
 final class GeneratedFileEditingNotificationProvider implements EditorNotificationProvider, DumbAware {
   @Override
   public @Nullable Function<? super @NotNull FileEditor, ? extends @Nullable JComponent> collectNotificationData(@NotNull Project project,
                                                                                                                  @NotNull VirtualFile file) {
-    if (!GeneratedSourceFileChangeTracker.getInstance(project).isEditedGeneratedFile(file)) return null;
```
jaen, line 23: @trushev I see @wgasiorek forgot to mention this when he published this PR, but the `GeneratedSourceFileChangeTracker` is now dead code. I haven't removed it as a part of this PR, but it probably should be removed. Thoughts?

platform/analysis-api/src/com/intellij/openapi/roots/GeneratedSourcesFilter.java
trushev: > Also, is there anything I would have to do to check this?

Sorry, I'm not sure what you mean by that. `RequiresReadLock` throws an assertion if read access is not allowed

Well, I think it is better to mark it as `RequiresReadLock` and remove `NonBlocking` suffix

platform/lang-impl/src/com/intellij/ide/GeneratedFileEditingNotificationProvider.java
trushev: You are right. The dead code shall be removed. Let's do it as a separate patch just in case to be able to easily revert it

platform/lang-impl/src/com/intellij/ide/GeneratedFileEditingNotificationProvider.java
trushev: I just think that in the context of developing the platform, it is slightly easier to understand what intellij's `notNull` does than google's `firstNonNull`

trushev (changes requested):

platform/analysis-api/src/com/intellij/openapi/roots/GeneratedSourcesFilter.java
jaen: Should be marked as such now.

jaen: Ah, I kind of assumed it's some static contract annotation and I would have to run something that checks this invariant. If it just inserts an assertions then yeah, I don't have to do anything else.

The method should now have the annotation.

jaen:
@trushev sorry for not having time to do it earlier, but it should now be up to date and your comments should be addressed. Still seems to work correctly in my testing.

trushev:
I'm sorry, `GeneratedFileEditingNotificationProvider` was reworked right at the moment I was merging this commit.
Just 30 mins early and it could be merged:( Could you please adopt your changes for the latest version?

jaen:
@trushev try now?

trushev:
@jaen Nice work, merged into master

PR #2787, opened 2024-06-27
[platform] IJPL-11679 Fix run widget doesn't use last run config if its stored externally in .run folder

wgasiorek:
Fixes issue described here: [[IJPL-11679] Run widget doesn't use last run config if it's stored externally in .run folder](https://youtrack.jetbrains.com/issue/IJPL-11679/Run-widget-doesnt-use-last-run-config-if-its-stored-externally-in-.run-folder)

### DIFF
--- a/platform/execution-impl/src/com/intellij/execution/impl/RunManagerImpl.kt
+++ b/platform/execution-impl/src/com/intellij/execution/impl/RunManagerImpl.kt
@@ -943,9 +943,16 @@ open class RunManagerImpl @NonInjectable constructor(val project: Project, share
     }
 
     if (selectedConfiguration == null) {
+      // NOTE: we use a local variable, because `notYetAppliedInitialSelectedConfigurationId` is a mutable property that can change
+      //       between the assignment and the conditional
+      val configurationToApply = selectedConfigurationId ?: ""
+
+      notYetAppliedInitialSelectedConfigurationId = configurationToApply
+
       // Empty string means that there's no information about initially selected RC in workspace.xml => IDE should select any.
-      notYetAppliedInitialSelectedConfigurationId = selectedConfigurationId ?: ""
-      selectAnyConfiguration()
+      if (configurationToApply.isEmpty()) {
+        selectAnyConfiguration()
+      }
     }
   }
 


alexander-doroshko:
Thank you for the pull request and sorry for the delay with the review. I'm now trying to push a different fix for this issue. If it lands, I'll write a detailed comment tomorrow explaining the root cause of the problem, corner cases, and the approach that I used to fix it.

alexander-doroshko:
Ok, so the fix has [landed](https://github.com/JetBrains/intellij-community/commit/a5a139dd05c24353b4852faaeacfdfc145bf94b0), and the tests that cover this functionality have become green and stable (`RunConfigurationSchemeManagerTest`)

Here's the task, the state of affairs, and the problem:
(RC = run configuration)
- the knowledge about the name of the "RC to be selected" is loaded when project is opening (`RunManagerImpl`: `loadState()` or `noStateLoaded()`)
- the knowledge about RCs available in the project comes from 3 sources:
  - .idea/workspace.xml file
  - .idea/runConfigurations/*.xml files
  - arbitrary `*.run.xml` files that may be located anywhere within the project content
- if the RC that should be selected on project opening comes from `workspace.xml` or `.idea/runConfigurations` then it just works because all the required knowledge is already available when `loadState()` or `noStateLoaded()` are running
- the things are trickier if the "RC to be selected" is not there
  - we may hope that the "RC to be selected" will eventually be loaded from some `*.run.xml` file. (The RCs from arbitrary `*.run.xml` files are loaded by either `RunManagerImpl#loadRunConfigsFromArbitraryFiles` or `RunConfigurationInArbitraryFileScanner`)
  - so we set *any* RC as the selected one and remember the name of the "RC to be selected"
  - when/if the "RC to be selected" is loaded from some `*.run.xml` file, we change the selected RC to the correct one
  - corner case: assume, no RCs have been loaded from `workspace.xml` or `.idea/runConfigurations`; some RCs have been loaded from `*.run.xml` files; but the  "RC to be selected" has not been loaded. In this case we should select *any* RCs from any `*.run.xml` file as the selected one


The previous (buggy) solution was based on this check:
```
if (!StartupManager.getInstance(project).postStartupActivityPassed()) {
    StartupManager.getInstance(project).runAfterOpened { ... }
}
```

This was not correct because the code that loads RCs from `*.run.xml` files is not directly related to "post startup activities" and their life cycle. The problem was caught by `RunConfigurationSchemeManagerTest`: two of the testcases were flaky.

The pushed fix switches to using `InitialVfsRefreshService.awaitInitialVfsRefreshFinished()`. This seem to work perfectly.

Feel free to have a look at the pushed fix and provide any feedback.

jaen:
I'm not sure I entirely understand everything about your explanation, but it certainly seems like a better choice to solve the root cause and I probably wouldn't even know to use `InitialVfsRefreshService.awaitInitialVfsRefreshFinished()` as so far I have only very, very cursory understanding of the codebase. Thanks for providing a more comprehensive fix!

PR #2792, opened 2024-06-27
[platform] IJPL-149806 Sort File Type mappings from the longest to shortest matcher

wgasiorek:
Fixes ticket described here: [[IJPL-149806] More specific File Types needs to have a higher priority](https://youtrack.jetbrains.com/issue/IJPL-149806/More-specific-File-Types-needs-to-have-a-higher-priority)

### DIFF
--- a/jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
+++ b/jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
@@ -20,13 +20,6 @@ import java.util.stream.Collectors;
 @ApiStatus.Internal
 public final class FileTypeAssocTable<T> {
 
-  // A comparator for pattern specificity added to resolve IJPL-149806. See the ticket and tests for this class for more details.
-  private final Comparator<Pair<FileNameMatcher, T>> myMatchingMappingComparator = Comparator
-    .comparing((Pair<FileNameMatcher, T> pair) -> pair.first.getPresentableString().length(), Comparator.reverseOrder())
-    .thenComparing(pair -> pair.first.getPresentableString()
-      .replace("?", "\uFFFE")
-      .replace("*", "\uFFFF"));
-
   private final Map<CharSequence, T> myExtensionMappings;
   private final Map<CharSequence, T> myExactFileNameMappings;
   private final Map<CharSequence, T> myExactFileNameAnyCaseMappings;
@@ -106,7 +99,14 @@ public final class FileTypeAssocTable<T> {
       myMatchingMappings.add(Pair.create(matcher, type));
     }
 
-    Collections.sort(myMatchingMappings, myMatchingMappingComparator);
+    // A comparator for pattern specificity added to resolve IJPL-149806. See the ticket and tests for this class for more details.
+    var mostSpecificPatternFirstComparator = Comparator
+      .comparing((Pair<FileNameMatcher, T> pair) -> pair.first.getPresentableString().length(), Comparator.reverseOrder())
+      .thenComparing(pair -> pair.first.getPresentableString()
+        .replace("?", "\uFFFE")
+        .replace("*", "\uFFFF"));
+
+    Collections.sort(myMatchingMappings, mostSpecificPatternFirstComparator);
 
     return Pair.getSecond(previousAssociation);
   }
--- a/jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
+++ b/jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
@@ -19,6 +19,14 @@ import java.util.stream.Collectors;
 
 @ApiStatus.Internal
 public final class FileTypeAssocTable<T> {
+
+  // A comparator for pattern specificity added to resolve IJPL-149806. See the ticket and tests for this class for more details.
+  private final Comparator<Pair<FileNameMatcher, T>> myMatchingMappingComparator = Comparator
+    .comparing((Pair<FileNameMatcher, T> pair) -> pair.first.getPresentableString().length(), Comparator.reverseOrder())
+    .thenComparing(pair -> pair.first.getPresentableString()
+      .replace("?", "\uFFFE")
+      .replace("*", "\uFFFF"));
+
   private final Map<CharSequence, T> myExtensionMappings;
   private final Map<CharSequence, T> myExactFileNameMappings;
   private final Map<CharSequence, T> myExactFileNameAnyCaseMappings;
@@ -28,14 +36,15 @@ public final class FileTypeAssocTable<T> {
 
   @FunctionalInterface
   public interface ConcurrentCharSequenceMapBuilder<T> {
-    @NotNull Map<CharSequence, T> build(@NotNull Map<? extends CharSequence, ? extends T> source, boolean caseSensitive);
+    @NotNull
+    Map<CharSequence, T> build(@NotNull Map<? extends CharSequence, ? extends T> source, boolean caseSensitive);
   }
 
   private FileTypeAssocTable(@NotNull Map<? extends CharSequence, ? extends T> extensionMappings,
                              @NotNull Map<? extends CharSequence, ? extends T> exactFileNameMappings,
                              @NotNull Map<? extends CharSequence, ? extends T> exactFileNameAnyCaseMappings,
                              @NotNull ConcurrentCharSequenceMapBuilder<T> concurrentCharSequenceMapBuilder,
-                             @NotNull Map<? extends String, ? extends T> hashBangMap,
+                             @NotNull Map<String, ? extends T> hashBangMap,
                              @NotNull List<? extends Pair<FileNameMatcher, T>> matchingMappings) {
     myExtensionMappings = concurrentCharSequenceMapBuilder.build(extensionMappings, false);
     myExactFileNameMappings = concurrentCharSequenceMapBuilder.build(exactFileNameMappings, true);
@@ -78,25 +87,34 @@ public final class FileTypeAssocTable<T> {
       String extension = ((ExtensionFileNameMatcher)matcher).getExtension();
       return myExtensionMappings.put(extension, type);
     }
+
     if (matcher instanceof ExactFileNameMatcher) {
       ExactFileNameMatcher exactFileNameMatcher = (ExactFileNameMatcher)matcher;
 
       Map<CharSequence, T> mapToUse = exactFileNameMatcher.isIgnoreCase() ? myExactFileNameAnyCaseMappings : myExactFileNameMappings;
       return mapToUse.put(exactFileNameMatcher.getFileName(), type);
     }
-    int i = ContainerUtil.indexOf(myMatchingMappings, p -> p.first.equals(matcher));
-    if (i == -1) {
-      myMatchingMappings.add(Pair.create(matcher, type));
-      return null;
+
+    Pair<FileNameMatcher, T> previousAssociation = null;
+    int previousAssociationIndex = ContainerUtil.indexOf(myMatchingMappings, a -> a.first.equals(matcher));
+
+    if (previousAssociationIndex >= 0) {
+      previousAssociation = myMatchingMappings.get(previousAssociationIndex);
+      myMatchingMappings.set(previousAssociationIndex, Pair.create(matcher, type));
     }
-    Pair<FileNameMatcher, T> old = myMatchingMappings.get(i);
-    myMatchingMappings.set(i, Pair.create(matcher, type));
-    return Pair.getSecond(old);
+    else {
+      myMatchingMappings.add(Pair.create(matcher, type));
+    }
+
+    Collections.sort(myMatchingMappings, myMatchingMappingComparator);
+
+    return Pair.getSecond(previousAssociation);
   }
 
   void addHashBangPattern(@NotNull String hashBang, @NotNull T type) {
     myHashBangMap.put(hashBang, type);
   }
+
   void removeHashBangPattern(@NotNull String hashBang, @NotNull T type) {
     myHashBangMap.remove(hashBang, type);
   }
@@ -298,7 +316,8 @@ public final class FileTypeAssocTable<T> {
   }
 
   // todo drop it, when ConcurrentCollectionFactory will be available in the classpath
-  private static @NotNull <T> Map<CharSequence, T> createCharSequenceConcurrentMap(@NotNull Map<? extends CharSequence, ? extends T> source, boolean caseSensitive) {
+  private static @NotNull <T> Map<CharSequence, T> createCharSequenceConcurrentMap(@NotNull Map<? extends CharSequence, ? extends T> source,
+                                                                                   boolean caseSensitive) {
     Map<CharSequence, T> map = CollectionFactory.createCharSequenceMap(caseSensitive, source.size(), 0.5f);
     map.putAll(source);
     return Collections.synchronizedMap(map);
@@ -314,11 +333,11 @@ public final class FileTypeAssocTable<T> {
 
   @Override
   public String toString() {
-    return "FileTypeAssocTable. myExtensionMappings="+myExtensionMappings+";\n"
-      +"myExactFileNameMappings="+myExactFileNameMappings+";\n"
-      +"myExactFileNameAnyCaseMappings="+myExactFileNameAnyCaseMappings+";\n"
-      +"myMatchingMappings="+myMatchingMappings+";\n"
-      +"myHashBangMap="+myHashBangMap+";";
+    return "FileTypeAssocTable. myExtensionMappings=" + myExtensionMappings + ";\n"
+           + "myExactFileNameMappings=" + myExactFileNameMappings + ";\n"
+           + "myExactFileNameAnyCaseMappings=" + myExactFileNameAnyCaseMappings + ";\n"
+           + "myMatchingMappings=" + myMatchingMappings + ";\n"
+           + "myHashBangMap=" + myHashBangMap + ";";
   }
 
   @TestOnly
--- a/jps/model-impl/testSrc/com/intellij/openapi/fileTypes/impl/FileTypeAssocTableTest.java
+++ b/jps/model-impl/testSrc/com/intellij/openapi/fileTypes/impl/FileTypeAssocTableTest.java
@@ -1,0 +1,117 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.openapi.fileTypes.impl;
+
+import com.intellij.openapi.fileTypes.FileNameMatcher;
+import com.intellij.openapi.util.Pair;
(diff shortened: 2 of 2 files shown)


cdracm (changes requested):
please add a test

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
```diff
@@ -42,7 +51,7 @@ private FileTypeAssocTable(@NotNull Map<? extends CharSequence, ? extends T> ext
     myExactFileNameAnyCaseMappings = concurrentCharSequenceMapBuilder.build(exactFileNameAnyCaseMappings, false);
     myConcurrentCharSequenceMapBuilder = concurrentCharSequenceMapBuilder;
     myHashBangMap = new ConcurrentHashMap<>(hashBangMap);
-    myMatchingMappings = ContainerUtil.createLockFreeCopyOnWriteList(matchingMappings);
+    myMatchingMappings = new CopyOnWriteArrayList<>(matchingMappings);
```
cdracm, line 54: please revert

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
```diff
@@ -84,14 +93,24 @@ public T addAssociation(@NotNull FileNameMatcher matcher, @NotNull T type) {
       Map<CharSequence, T> mapToUse = exactFileNameMatcher.isIgnoreCase() ? myExactFileNameAnyCaseMappings : myExactFileNameMappings;
       return mapToUse.put(exactFileNameMatcher.getFileName(), type);
     }
-    int i = ContainerUtil.indexOf(myMatchingMappings, p -> p.first.equals(matcher));
-    if (i == -1) {
-      myMatchingMappings.add(Pair.create(matcher, type));
-      return null;
+
+    // Since we need to sort on insertion for IJPL-149806, we allow only one thread to modify the mapping list at a time
+    myMatchingMappingsLock.lock();
```
cdracm, line 98: please get rid of this lock, it's unnecessary

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
```diff
@@ -84,14 +93,24 @@ public T addAssociation(@NotNull FileNameMatcher matcher, @NotNull T type) {
...
+    try {
+      int i = ContainerUtil.indexOf(myMatchingMappings, p -> p.first.equals(matcher));
+      if (i == -1) {
```
cdracm, line 101: please rewrite to have only one call to sort

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
```diff
@@ -120,7 +139,15 @@ void removeAssociation(@NotNull FileNameMatcher matcher, @Nullable T type) {
       }
       return;
     }
-    myMatchingMappings.removeIf(assoc -> matcher.equals(assoc.getFirst()) && (type == null || type.equals(assoc.getSecond())));
+
+    // Since we need to sort on insertion for IJPL-149806, we allow only one thread to modify the mapping list at a time
+    myMatchingMappingsLock.lock();
```
cdracm, line 144: please don't sync, that's why COWAL was here

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
```diff
@@ -14,17 +14,26 @@
...
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 @ApiStatus.Internal
 public final class FileTypeAssocTable<T> {
+  // Comparator to sort wildcard mappings from the longest to shortest matcher for IJPL-149806.
+  // NOTE: this is not perfect as sometimes wildcard matchers of the same length can still be considered to be of different "specificity"
+  //       (e.g. you'd probably want `*.pdf` to take precedence over `*.p?f`), but is still an improvement on the status quo.
+  private static final Comparator<Pair<FileNameMatcher, ?>>
+    MY_MATCHING_MAPPINGS_COMPARATOR = Comparator.comparing((Pair<FileNameMatcher, ?> pair) -> pair.first.getPresentableString()).reversed();
```
cdracm, line 28: please rename to meaningful name

jaen: Saying "it's a comparator for matching mappings" seems meaningful enough to me? Unless you want the name to say something about how it sorts? So maybe something like `MY_LONGEST_MATCHING_MAPPING_FIRST_COMPARATOR` is what you have in mind?

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
jaen: TL;DR The JB custom lock-free COW List is not sortable, which is why I reverted to the built-in Java one, which is sortable. If you want to stay with the JB one, I'm not quite sure how you want me to implement that? Implement atomic sorting for JB's COW arraylist as a part of this ticket?

For a longer explanation, here's the full context:

Since I didn't have any direct line of contact to you to ask questions about the implementation, I went with my best effort of understanding it from the code and relevant tickets. So I'll take this opportunity to explain it and ask questions, hope you don't mind.

So yes, while investigating this I've noticed that the code first was not thread-safe, then made thread-safe with `Collections.synchronizedList` (there was no relevant ticket linked, but I assume that's a self-evident issue), then later to Java's built-in COW arraylist (the ticket does not explain why, I'd assume it would be contention issues?) and then finally a custom JB COW arraylist implementation was made by you, as a part of this ticket — https://youtrack.jetbrains.com/issue/KTIJ-27475.

I figured the cleanest solution would've been a sorted datastructure (such as a sorted map), but given it would require benchmarking and none of the changes in the place had any test data to benchmark with, I decided against that.

So then next best idea was to just sort on insertion — it's kind of crude, but maybe modifications don't happen often enough for it to matter. But it turns out that JB COW arraylist is not sortable — Java's sort requires a mutable iterator by default (c.f. https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/List.java#L505-L513) while COW arraylist (both stdlib's and JB's) have immutable iterators. The stdlib list works around that by defining it's own `sort` implenetation (c.f. https://github.com/openjdk/jdk/blob/jdk-17%2B35/src/java.base/share/classes/java/util/concurrent/CopyOnWriteArrayList.java#L886-L898) but JB's list doesn't. This makes sorting fail using JB's COW arraylist.

So the next best thing I could think of without having any direct line of contact to you is to try reverting to stdlib's COW arraylist and try benchmarking if it makes any difference. Your PR for KTIJ-27475 makes it look like the introduction of JB's COW arraylist wasn't really related to `FileTypeAssocTable.java`, but some other issue (hard to tell for sure though, without more context that's unavailable to me) so I figured it's worth a try.

So this is the implementation I ended up with and I have benchmarked it by building the IDE with and without that change and using those built IDEs to compile IntelliJ again, with profiler enabled. In both cases full IntelliJ compilation took ~12min 30s and churned about ~50GB of memory and seemed to behave similarly. So I figured it's okay to revert that and use stdlib's COW arraylist in this case.

If you have any projects I can test this that this code behaved pathologically with stdlib's implementation, I would appreciate being able to test with them to see if I'm not breaking anything here. Otherwise it's hard to see how this change has a negative impact. It's quite possible I might be missing something, but I'm not sure what.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
jaen: Is it? Granted, I made a wrong assumption about stdlib's COW arraylist, it seems that it also overrides `removeIf` to be atomic (c.f. https://github.com/openjdk/jdk/blob/jdk-17%2B35/src/java.base/share/classes/java/util/concurrent/CopyOnWriteArrayList.java#L814-L869), so at least from the sorting perspective it seems to be unnecessary — my bad.

But what if some other thread modifies the list between `indexOf` and `add` — I see there's `addIfAbsent` that's atomic, but it would need a predicate version, because we want to compare by `first.equals` instead of just `equals` — won't we duplicate items then?

Or what if some other thread modifies the list between `indexOf` and `set`? Won't we overwrite a wrong matching then?

Say, one thread runs `addAssociation` while another runs `removeAssociation` and it clobbers the list for `addAssociation`?

If you're sure that can _never_ happen, then the lock if probably not needed, but I don't have enough idea about what can use this class and if modifications can happen concurrently from different threads, so I thought it's better to be defensive. If you're sure that's not necessary, then I guess it can be removed. I just want to be sure I'm not missing something here.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
jaen: But there is already only one call ever due to the return? Unless you mean something like this instead?
```java
myMatchingMappingsLock.lock();
try {
  Pair<FileNameMatcher, T> old = null;
  int i = ContainerUtil.indexOf(myMatchingMappings, p -> p.first.equals(matcher));

  if (i >= 0) {
    old = myMatchingMappings.get(i);
    myMatchingMappings.set(i, Pair.create(matcher, type));
  else {
    myMatchingMappings.add(Pair.create(matcher, type));
  }

  Collections.sort(myMatchingMappings, MY_MATCHING_MAPPINGS_COMPARATOR);
  
  return Pair.getSecond(old);
}
finally {
  myMatchingMappingsLock.unlock();
}
```
It seems to be more self-explanatory to read that way, I suppose.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
jaen: See my comment for https://github.com/JetBrains/intellij-community/pull/2792/files#r1677694484. Maybe I'm wrong, but my understanding is that if `addAssociation` and `removeAssociation` can ever get called from different threads they could mess up the matcher list.

Also, there's one more thing I noticed — but probably benign.

The JB's COW arraylist doesn't overwrite `removeIf` to make it atomic, so it uses the `Collection` version that gets the iterator (which as a copy of the array), iterates over it testing the predicate, and if the predicate matches it calls `remove` on the iterator (c.f. https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/Collection.java#L576-L587).

Then JB's COW arraylist just runs `remove` for each single element (c.f. https://github.com/JetBrains/intellij-community/blob/0fa7fb3581d4f15365c9365bb68885595bfc9d67/platform/util/src/com/intellij/util/containers/LockFreeCopyOnWriteArrayList.java#L899-L908), so other threads can see the state of the list in the middle of deletion. It's probably not a big deal if we're only removing stuff, but figured I'd mention it anyway, if I started explaining what I've done here and why.

jaen (commented):
@cdracm responded to your comments.

My responses are a bit long because I try to explain why I've arrived at the solution you see — because I had no way to ask why certain changes were made, I had to guess from the context and maybe misunderstood something — but here's a TL;DR so you can at least have an idea of what is there before you read:

  * I can't just use JB's COW arraylist as–is because it's not sortable,
  * I reverted to stdlib's COW arraylist implementation, because your change to add it seemed to address a different issue,
  * I have profiled IntelliJ compilation with stock IDE and IDE with my change and it took roughly the same time (~12m 30s) and churned roughly the same amount of memory (~50GB) and I didn't see anything egregiously wrong in the profile, so it seems like it's okay to revert to the stdlib's COW arraylist to me. But maybe I was testing the wrong thing, but it was hard to know since the tickets didn't have any context or reproducer projects for the issue to ensure nothing broke, so let me know if there's a way to test for the issue you fixed but this change specifically,
  * if you prefer to stay with JB'S COW arraylist, then it would need adding at least an atomic `sort` override and ideally making `removeIf` atomic as well as adding something for atomic `addOrSetIfMatching` so that another thread can't change the list between `indexOf` and `set` (I think it can happen and it's why I added a lock).

Let me know where do you want to go from here. Of course, if something incorrect then I'll certainly change it — I'd just like to understand why it is incorrect.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
cdracm: Hi. It's a valid point about ContainerUtil.createLockFreeCopyOnWriteList() not being (atomically) sortable, thanks for noticing. I've pushed a version with overridden sort() method, so now it's sortable. About testing the difference: it's hard to spot any differencein performance of ContainerUtil.createLockFreeCopyOnWriteList(), off-the-shelf COWAL or synchronizedLIst here, since this code is not on any critical path. I'm advocating  ContainerUtil.createLockFreeCopyOnWriteList here for consistency/simplicity reasons, not the performance.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
cdracm: myMatchingMappings is changed only during showing of the modal dialog in Settings|file types. there's no need to guard the modification any further, for simplicity/readability reasons.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
cdracm: yes, precisely, that variant is better, thanks

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
cdracm: now, that we have only one usage of this constant, could you please inline it to a local var closer to the usage, and add there this comment above? I don't think anyone would use that constant ever more, so it doesn't make sense to have this comparator as a separate constant. the local var name can be "byNameLengthComparator" or smth similar.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
cdracm: yeah, we don't have to worry about multithreading here, addAssociation/removeAssociation don't get called in parallel.

jps/model-impl/src/com/intellij/openapi/fileTypes/impl/FileTypeAssocTable.java
jaen: Makes sense, now that those operations were added it should be okay to use the JB COW arraylist variant and drop the locks. I'll update the PR accordingly.

jaen (commented):
@cdracm I have updated the code – it should now use JB's COW arraylist and not have locks. I have also added some tests, as requested.

While writing tests I realised the comparator was a bit too simple, because you'd probably want to sort dots as more specific than wildcards (and in ASCII dot has a lower ordinal than `?` and `*`) — for example `a.bc.de` would get sorted as less specific than `a?bc?de`, even though intuitively you'd expect it to be the other way round — so I have changed the the comparator to be more intuitive.

But that makes it have two more outstanding questions:

This (still relatively simple) comparator treats `*.foo.cpp` as less specific than `env.foo.*`, which might be unintuitive (I'd expect the `*.foo.cpp` pattern to win, personally). I have written a test to document this behaviour. Do we think this behaviour is good enough for now, or would we want to improve it?

A different comparator I tried is by writing a more low-level one that does more-or-less the same comparison, but going from the end of the string rather than from the start. It handles the above case more intuitively, but it probably will have other corner cases (I think only using a suffix trie here is a real solution for no corner cases, but it'd be significant implementation change that's possibly not worth it).

<details>
<summary>Alternative comparator code</summary>

```java
class PatterSpecificityComparator<T> implements Comparator<Pair<FileNameMatcher, T>> {
  @Override
  public int compare(Pair<FileNameMatcher, T> lhs, Pair<FileNameMatcher, T> rhs) {
    var lhsIterator = new StringCharacterIterator(lhs.first.getPresentableString());
    var rhsIterator = new StringCharacterIterator(rhs.first.getPresentableString());

    // Start from the end, since we're dealing with file extensions
    var lhsChar = lhsIterator.last();
    var rhsChar = rhsIterator.last();

    while (true) {
      if (lhsChar == CharacterIterator.DONE) {
        if (rhsChar == CharacterIterator.DONE) {
          return Character.compare(lhsIterator.next(), rhsIterator.next());
        }
        else {
          return 1; // LHS is shorter, so probably less specific, so sort it later in the list
        }
      }
      else if (rhsChar == CharacterIterator.DONE) {
        return -1; // RHS is shorter, so probably less specific, so sort it later in the list
      }

      if (lhsChar == '?') lhsChar = '\uFFFE';
      if (lhsChar == '*') lhsChar = '\uFFFF';
      if (rhsChar == '?') rhsChar = '\uFFFE';
      if (rhsChar == '*') rhsChar = '\uFFFF';

      var comparisonResult = Character.compare(lhsChar, rhsChar);

      if (comparisonResult != 0) return comparisonResult;


      lhsChar = lhsIterator.previous();
      rhsChar = rhsIterator.previous();
    }
  }
}

```
</details>

I think other than those two issues, that is:
  * deciding where to instantiate the comparator,
  * deciding which comparator to use,
this should be more or less good enough now, right?

cdracm:
> @cdracm I have updated the code – it should now use JB's COW arraylist and not have locks. I have also added some tests, as requested.

Thanks!,

> 
> This (still relatively simple) comparator treats `*.foo.cpp` as less specific than `env.foo.*`, which might be unintuitive (I'd expect the `*.foo.cpp` pattern to win, personally). I have written a test to document this behaviour. Do we think this behaviour is good enough for now, or would we want to improve it?


Yes, that's the problem with this simple heuristic, which is mentioned in the original issue. I think we don't want to implement the general "regex nesting problem" algorithm, so this is good enough.

> 
> A different comparator I tried is 

... too complex

> I think other than those two issues, that is:
> * deciding where to instantiate the comparator,

in the usage site

> * deciding which comparator to use,

the simplest one

>   this should be more or less good enough now, right?

yes, when the tests are added

jaen:
@cdracm
> in the usage site
Ok, if you're sure about it then https://github.com/JetBrains/intellij-community/pull/2792/commits/957db7558dee085bfc7e138f0334413864629b63 moves it where I understood you want it.

> yes, when the tests are added
But I have already added tests in `jps/model-impl/testSrc/com/intellij/openapi/fileTypes/impl/FileTypeAssocTableTest.java` when I posted the above comment. Is there something wrong with them?

cdracm:
> But I have already added tests 

oh, i didn't see them. Great, thank you!

PR #2794, opened 2024-06-27
[settings-sync] IJPL-13931 Ensure settings categories are respected in the initial sync

jaen:
Resolves: https://youtrack.jetbrains.com/issue/IJPL-13931/Settings-Sync-ignores-the-category-state-changing-when-enabling-sync-for-the-first-time

As far as I can tell, the Settings Sync code _does_ send only the files you requested, but:

  * it does only send files as they are on the disk – when setting up the sync, the
    `SettingsSyncSettings` state won't get persisted immediately (the IDE only does
    it later, after you exit settings), so another IDE wanting to use that state
    will show all categories enabled (even though the sync will not contain that
    date) until the settings get flushed,
  * settings for plugins didn't get properly categorised – this mean that you could
    either disable the category wholesale, or you would get all the plugins when
    syncing.
    
Those issues were fixed by — respectively — adding an explicit settings flush before
the initial sync and adding code to categorise plugins properly. I think this should
fix the issue as perceived in the UI by an user.

### DIFF
--- a/plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncBridge.kt
+++ b/plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncBridge.kt
@@ -77,6 +77,10 @@ class SettingsSyncBridge(
   internal fun initialize(initMode: InitMode) {
     coroutineScope.launch {
       try {
+        // Always explicitly flush settings – if this is not done before sending sync events, then remotely synced settings
+        // might not contain the most up–to–date settings state (e.g. sync settings will be stale).
+        saveIdeSettings()
+
         settingsLog.initialize()
 
         // the queue is not activated initially => events will be collected but not processed until we perform all initialization tasks
@@ -109,10 +113,8 @@ class SettingsSyncBridge(
     }
   }
 
-  private fun saveIdeSettings() {
-    runBlockingCancellable {
-      saveSettings(ApplicationManager.getApplication(), forceSavingAllSettings = true)
-    }
+  private suspend fun saveIdeSettings() {
+    saveSettings(ApplicationManager.getApplication(), forceSavingAllSettings = true)
   }
 
   private suspend fun applyInitialChanges(initMode: InitMode) {
--- a/plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncFiltering.kt
+++ b/plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncFiltering.kt
@@ -2,6 +2,7 @@ package com.intellij.settingsSync
 
 import com.intellij.configurationStore.getPerOsSettingsStorageFolderName
 import com.intellij.configurationStore.schemeManager.SchemeManagerFactoryBase
+import com.intellij.ide.plugins.PluginManagerCore
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.*
 import com.intellij.openapi.editor.colors.impl.AppEditorFontOptions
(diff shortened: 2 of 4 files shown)


plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncBridge.kt
```diff
@@ -77,6 +77,10 @@ class SettingsSyncBridge(
   internal fun initialize(initMode: InitMode) {
     coroutineScope.launch {
       try {
+        // Always explicitly flush settings – if this is not done before sending sync events, then remotely synced settings
+        // might not contain the most up–to–date settings state (e.g. sync settings will be stale).
+        saveIdeSettings()
```
paksv, line 82: that might potentially lead to problems, such as https://youtrack.jetbrains.com/issue/IJPL-1245/NullPointerException-Cannot-invoke-com.intellij.openapi.application.impl.ReadMostlyRWLock.isReadAllowed-because-this.myLock-is (https://github.com/JetBrains/intellij-community/commit/2c938259247aaa1b6bc7a0c93b48e6c067192960)

Since, it happens inside the coroutine, I don't expect it to happen. But still a thing to keep in mind and watch

plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncFiltering.kt
```diff
@@ -82,18 +78,30 @@ private fun getSchemeCategory(fileSpec: String): SettingsCategory? {
...
-  if (fileSpec == AppEditorFontOptions.STORAGE_NAME)
-    return EDITOR_FONT_SUBCATEGORY_ID
-  else
-    return null
+private fun getSubCategory(fileSpec: String, category: SettingsCategory, componentClass: Class<PersistentStateComponent<Any>>): String? {
+  // NOTE: we do not cache the subcategory here, as it's cached in the `categoryCache`
+  val subcategory = when (category) {
+    SettingsCategory.UI ->
+      if (fileSpec == AppEditorFontOptions.STORAGE_NAME) EDITOR_FONT_SUBCATEGORY_ID else null
+    SettingsCategory.PLUGINS -> {
```
paksv, line 94: That PLUGINS category is a bit tricky. It isn't in fact used as it is written here, but rather very rarely as a category (see `com.intellij.openapi.updateSettings.impl.pluginsAdvertisement.GlobalIgnoredPluginSuggestionState` as an example), so I will revert the method to the original one.

paksv (approved):

plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncBridge.kt
jaen: Just to double check I understand — nothing needs to be done this, just a note for the future?

plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncFiltering.kt
jaen: Hmmm, it's been a while since I wrote this so maybe I misremember it, but as far as I could tell without this change you can't really disable syncing for a single plugin (both sync code and apply code consult `isSyncCategoryEnabled` as far as I could tell). This makes the `Configure` dropdown by the `Plugins` checkbox end up being dead and misleading UI — no matter what you choose there, you can only either sync all plugins or none.

If you prefer to keep this broken behaviour for now and fix it separately then sure, we can do it like that. Could you just clarify what exactly you want me to do here? You gave an approve and said "I will revert" — do I have to do something there, or does that mean you will finish this PR yourself?

paksv: @jaen indeed, there's some complication and inconsistency here, but I'd like to address this separately

paksv:
merged 3b3ef02

plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncFiltering.kt
paksv: >  You gave an approve and said "I will revert" — do I have to do something there, or does that mean you will finish this PR yourself?
nope, I did this myself

plugins/settings-sync/src/com/intellij/settingsSync/SettingsSyncBridge.kt
paksv: @jaen yes

PR #2800, opened 2024-06-30
Allow configuring test maven repositories

estebandlc:
Hi! Android Studio engineer here! We are starting to run debugger tests on Android, and we need to be able to specify our own custom maven url/directory. Hope this change makes sense. Thanks!


amaembo:
Merged, thanks: 3b7c729f5ac5d7bd1a82cf037487c14681233d7a

PR #2801, opened 2024-07-01
IDEA-355762: Propagate additional title actions to nested decorators

sprigogin:


platform/platform-impl/src/com/intellij/toolWindow/InternalDecoratorImpl.kt
```diff
@@ -536,6 +547,11 @@ class InternalDecoratorImpl internal constructor(
     secondDecorator?.setTabActions(actions)
   }
 
+  private fun updateTitleActions() {
+    val actions =  if (ClientProperty.get(this, HIDE_COMMON_TOOLWINDOW_BUTTONS) == true) emptyList() else titleActions
```
stachenov, line 551: Overall everything looks fine, but I don't like this part. There was one place when this property was checked, now there are two, so the logic is spread out and became more complicated. As far as I can tell, since these actions never worked to begin with for the split mode, we can just modify the other place to skip these if this property as set, in the header's toolbar, like this:
```
Index: community/platform/platform-impl/src/com/intellij/toolWindow/ToolWindowHeader.kt
IDEA additional info:
<+>UTF-8
diff --git a/community/platform/platform-impl/src/com/intellij/toolWindow/ToolWindowHeader.kt b/community/platform/platform-impl/src/com/intellij/toolWindow/ToolWindowHeader.kt
--- a/community/platform/platform-impl/src/com/intellij/toolWindow/ToolWindowHeader.kt	(revision Staged)
+++ b/community/platform/platform-impl/src/com/intellij/toolWindow/ToolWindowHeader.kt	(date 1721805523576)
@@ -132,7 +132,7 @@
 
           val tabListAction = e.actionManager.getAction("TabList")
           if (hideCommonActions == true) {
-            return arrayOf(tabListAction, *extraActions.toTypedArray())
+            return arrayOf(tabListAction)
           }
           return arrayOf(tabListAction, *extraActions.toTypedArray(), commonActionsGroup)
         }
```
This way the additional call in `reshape` won't be necessary as well.

stachenov (changes requested):

stachenov:
I could fix it myself, but I can't easily check it, as this combination of a split tool window with custom title actions doesn't seem to be used anywhere I can just access locally without setting up a whole infrastructure (like Android emulators).

platform/platform-impl/src/com/intellij/toolWindow/InternalDecoratorImpl.kt
sprigogin: Good idea.

sprigogin:
Superseded by https://github.com/JetBrains/intellij-community/pull/2818.

PR #2802, opened 2024-07-02
[settings-sync] IJPL-13108 Merge plugin.json and settings providers files manually

Florian3k:
Fixes: https://youtrack.jetbrains.com/issue/IJPL-13108/Dont-let-Git-merge-plugins.json-or-settings-providers-files


paksv:
@Florian3k the fix requires some tests as well. `com.intellij.settingsSync.GitSettingsLogTest` should be the best place for them

PR #2803, opened 2024-07-02
IDEA-355763: Fix tab dragging for windowed and floating tool windows

sprigogin:


platform/platform-impl/src/com/intellij/toolWindow/InternalDecoratorImpl.kt
```diff
@@ -835,16 +836,22 @@ class InternalDecoratorImpl internal constructor(
     }
 
     val divider = divider
-    disposable = Disposer.newCheckedDisposable()
-    HOVER_STATE_LISTENER.addTo(this, disposable!!)
+    val disposable = Disposer.newCheckedDisposable()
```
stachenov, line 839: I'll merge it like this to avoid spending too much time on it, but the next time please separate refactoring/cleanup like this and the actual functional changes into separate commits. It's easier to follow the logic and not to get lost in changes that don't change any behavior.

PR #2804, opened 2024-07-04
Fix false warnings on signature polymorphic method lookup

SirYwell:
`MethodHandle.invoke(Exact)` and `VarHandle.{access-mode}` behave differently than other methods when using `MethodHandle.Lookup#findVirtual`. According to the API specification (https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/invoke/VarHandle.html#interoperation-between-varhandles-and-the-core-reflection-api-heading):

> The [Lookup.findVirtual](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html#findVirtual(java.lang.Class,java.lang.String,java.lang.invoke.MethodType)) API is also able to return a method handle to call an access mode method for any specified access mode type and is equivalent in behavior to [MethodHandles.varHandleInvoker(java.lang.invoke.VarHandle.AccessMode, java.lang.invoke.MethodType)](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/invoke/MethodHandles.html#varHandleInvoker(java.lang.invoke.VarHandle.AccessMode,java.lang.invoke.MethodType)).

As a consequence, the signature mismatch warning should not be emitted for those methods.

### DIFF
--- a/java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangInvokeHandleSignatureInspection.java
+++ b/java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangInvokeHandleSignatureInspection.java
@@ -1,6 +1,7 @@
 // Copyright 2000-2023 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
 package com.intellij.codeInspection.reflectiveAccess;
 
+import com.intellij.codeInsight.AnnotationUtil;
 import com.intellij.codeInsight.daemon.JavaErrorBundle;
 import com.intellij.codeInsight.intention.preview.IntentionPreviewInfo;
 import com.intellij.codeInsight.lookup.*;
@@ -25,7 +26,6 @@ import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.lang.invoke.VarHandle;
 import java.util.*;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


amaembo:
Hello! Probably it would be more robust to check the annotation? Like this (applied to the original code):

```java
diff --git a/community/java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangInvokeHandleSignatureInspection.java b/community/java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangInvokeHandleSignatureInspection.java
--- a/community/java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangInvokeHandleSignatureInspection.java	(revision a7add089214429079c20e2101f8591a392c50ef5)
+++ b/community/java/java-impl-inspections/src/com/intellij/codeInspection/reflectiveAccess/JavaLangInvokeHandleSignatureInspection.java	(date 1702548929463)
@@ -225,6 +225,10 @@
         return;
       }
     }
+    PsiMethod onlyMethod = ContainerUtil.getOnlyItem(filteredMethods);
+    if (onlyMethod != null && onlyMethod.getModifierList().hasAnnotation("java.lang.invoke.MethodHandle.PolymorphicSignature")) {
+      return;
+    }
 
     final ReflectiveSignature methodSignature = composeMethodSignature(methodTypeExpression);
     if (methodSignature == null) return;
```

What do you think?

SirYwell:
> What do you think?

Agree, that's cleaner. I decided to use the same method that is used in other places to check for the annotation.

amaembo:
Merged to internal repo, thank you. Will appear on GitHub within one hour.

SirYwell:
Hey @amaembo, I assume https://github.com/JetBrains/intellij-community/commit/f49ee0495791c1ee5704e01e36b102f40d77fe9e is the commit? It looks like it contains outdated changes too, is that intentional?

amaembo:
b22bb694b41e16455f601bd32bbf8f3d1b335e05

PR #2805, opened 2024-07-04
IDEA-355899 show the run options menu for common modules

karimai-gg:
When running from multiplatform package context, where multiple configurations can be available.


sellmair:
Thank you for your contribution: Will dispatch to the relevant team 👍

PR #2807, opened 2024-07-05
[kotlin] J2K: Correctly reference inner class types in outer class header

ermattt:
J2K is being a little ovezealous when adding imports for nested, locally defined types. In K1, postprocessing handles most of the clean up, though it still misses type references in outer class headers. In addition to fixing the references-in-outer-class-headers bug (which produces unbuildable kotlin code), we can improve output for K1 & K2 modes by applying a few rules during type symbol printing:
1. If a type's definition and reference live in the same file, no import is needed as long as we follow the next 3 rules:
2. If a type's definition is a parent node of the reference, we can use the type's short name
3. If a type's definition and reference share the same direct parent node, we can use the type's short name
4. If a type's definition and reference share a non-direct parent node, we can use the type's semi-qualified name

I wish I could move this logic out of the printing phase and somewhere earlier, but I'm not sure how to do that without a significant refactor of how type symbols are handled.


abelkov:
> which produces unbuildable kotlin code

Can you please add a test for this? So far I don't quite understand what exactly are we fixing in this PR, apart from the style issue.

PR #2808, opened 2024-07-08
[PY-27985] Add pygame package alias for pygame-ce

zoldalma999:
pygame-ce is a fork and drop in replacement of pygame and to make migration easier, uses the pygame package name (but on pypi it is up as pygame-ce, and so pycharm does not detect it in requirements.txt)   
Related issues: [PY-27985](https://youtrack.jetbrains.com/issue/PY-27985) and https://github.com/pygame-community/pygame-ce/issues/2478


vladimir-koshelev:
We’ll be using pygame by default rather than pygame-ce, since it’s the original project rather than a fork. The development of pygame is still ongoing, so there’s no need to switch to an alternative version.

PR #2811, opened 2024-07-12
PY-34626 Fix isMethodContext to exclude inner functions as methods

somethingnew179:


east825:
LGTM! I've updated the tests a little and put everything into master. Thanks for your contribution! :rocket:

PR #2816, opened 2024-07-19
Fixed condition for plugin updates check

fkorotkov:
It seems in this scenario we'd like to check for plugin updates and no the IDE updates itself.


fkorotkov:
@jreznot just a gentle ping. 😅

jreznot:
Thank you! Sorry for the delay, it is on the way to master

PR #2820, opened 2024-07-25
IDEA-356832 - method reference inspection for java cast method

kamilkrzywanski:
Fix for issue https://youtrack.jetbrains.com/issue/IDEA-356832/Inspections-ignore-redundant-Classcast-method-call


amaembo:
Your approach is somewhat limited, as it covers `Stream.map` only. I'll implement it in a more general way to report redundant `Class::cast` call in any position.

PR #2821, opened 2024-07-29
PY-49174 make Match and Case expressions folding

somethingnew179:


east825:
LGTM. I've started a Safe Push and will close the ticket once the changes are in master. Thanks for taking care of this!

PR #2822, opened 2024-07-29
ThreadLeakTracker: avoid busy-waiting (IJPL-159321)

gharrma:
Busy-waiting burns some CPU and potentially slows down the background thread by continuously requesting its stack trace.

More importantly, if ThreadLeakTracker.checkLeak() happens to be called on the EDT inside computePrioritized(), then busy-waiting can cause other threads to get stuck yielding priority to the EDT (in sleepIfNeededToGivePriorityToAnotherThread), leading to a false-positive thread leak error.

Issue link: https://youtrack.jetbrains.com/issue/IJPL-159321

### DIFF
--- a/platform/testFramework/common/src/common/ThreadLeakTracker.java
+++ b/platform/testFramework/common/src/common/ThreadLeakTracker.java
@@ -13,6 +13,7 @@ import com.intellij.util.containers.ContainerUtil;
 import com.intellij.util.io.FilePageCacheLockFree;
 import com.intellij.util.ui.EDT;
 import com.intellij.util.ui.UIUtil;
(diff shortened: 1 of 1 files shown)


gharrma:
Side note: there is an [alternative](https://github.com/gharrma/intellij-community/commit/670abcbae188905d8c09fb9b5b76cb4b5c3b8a4b) fix that additionally avoids delays in event queue processing. However, the alternative fix seemed complicated, and I noticed that existing utilities such as `PlatformTestUtil.waitForFuture` already do the simpler thing anyway (i.e., just sleeping for 10 milliseconds).

PR #2823, opened 2024-08-01
Add Android Studio icons to newer load API calls

jrlogsdon:
This allows for slightly faster startup by not requiring reading in the IconMapper's json file when the plugin is loaded. Additionally, it allows for removal of the iconMapper extension from IntelliJ side once all plugins are off of it.


jrlogsdon:
This allows for us to create those newer loadApi calls on the Android Studio side (and get rid of our existing mapping file). If/when this is accepted, I'll create the corresponding changes in our android repo.

jreznot:
Merged!

PR #2824, opened 2024-08-09
[IDEA-347310] Run GradleDependencyReportTask without configuration cache

wiyarmir:
On projects with configuration cache enabled, the `ConfigurationContainer` object will not contain any of the configuration info needed to generate the dependency report. This is true even if we removed now-forbidden accesses to `getProject()` and injected it through the Gradle services. So it just runs the task without configuration cache.

`runCustomTask` is only ever used by `GradleDependencyAnalyzerContributor` but still extended the API just in case.


HackerMadCat:
Hi! Thank you for the contribution. I'm sorry for the long response. Your suggestion helped me choose the best solution for the fix. :)

The original issues were closed
[IDEA-368196](https://youtrack.jetbrains.com/issue/IDEA-368196) Mark GradleDependencyReportTask as non-compatible with configuration cache
[IDEA-347310](https://youtrack.jetbrains.com/issue/IDEA-347310) GradleDependencyReportTask not compatible with configuration cache

wiyarmir:
@HackerMadCat 🙇🏻 thanks for fixing and replying!

PR #2825, opened 2024-08-16
Add ExpandCollapseToggleAction to IdeActions

The1xDeveloper:
Main driver is to add `za` toggle fold support to ideaVim https://youtrack.jetbrains.com/issue/VIM-566/Support-more-Vim-folding-commands

IdeaVim imports these actions [here](https://github.com/JetBrains/ideavim/blob/master/src/main/java/com/maddyhome/idea/vim/helper/IjActionExecutor.kt#L48). Technically IdeaVim can pass the raw string `ExpandCollapseToggleAction` to use the toggling but that's flimsy


AlexPl292:
Hi there! Thank you for the update. This constant will be available in the platform since 2024.3 release.

Regarding the IdeaVim: since we support the range of IDE versions, we won't be able to use this constant until [the minimal supported version](https://github.com/JetBrains/ideavim/blob/5fb4c10f881d4864f886a427bdbc611db8aae128/.teamcity/_Self/Constants.kt#L13) will be 2024.3.
I left a note in the code to update the code when the change will be available.

PR #2826, opened 2024-08-16
IDEA-80032 Support classifiers with Maven sources

nrayburn-tech:
This resolves the issue when using a Maven dependency with a classifier.  Previously, the sources without the classifier would be downloaded.  This modifies the behavior to download the sources with the classifier.

I didn't find any tests to edit for WorkspaceModuleImporter.kt, but if there's something existing I can update it.

A few notes before merging.
1. I noticed that may be related to my changes is that when viewing the source files, the breadcrumbs at the bottom right show a different source jar depending on where you click within the source file on screen. Sometimes it shows as `primefaces-14.0.4-jakarta.jar` and sometimes it shows as `primefaces-14.0.4-jakarta-sources.jar`.
2. This only adds the prefix for sources, it does not do it for javadocs.  I'm not sure if it's the right approach or not for javadocs.
3. If the sources don't exist with the artifact classifier prefix, then there is not any fallback to the sources without the classifier prefix.

### Manual Test Steps
1. Create a new Java -> Maven project within IntelliJ.
4. Modify the existing pom.xml with the below dependency.
  ```xml
<dependency>
    <groupId>org.primefaces</groupId>
    <artifactId>primefaces</artifactId>
    <version>14.0.4</version>
    <classifier>jakarta</classifier>
</dependency>
  ```
5. Reload the Maven project and download sources.
6. Expand the "External Libraries" in the file explorer.
7. Navigate to `Maven: org.primefaces:primefaces:jakarta:14.0.4` and expand the library.
8. Open the `org.primefaces.application.exceptionhandler.PrimeExceptionHandler` class and review the imports at the top of the file, confirm they are the `jakarta` imports.
9. Remove the `classifier` from the dependency.
10. Reload the Maven project and download sources.
11. Repeat the same general steps to confirm that the `javax` imports are used now.
12. Navigate to `Maven: org.primefaces:primefaces:14.0.4` and expand the library.
13. Open the `org.primefaces.application.exceptionhandler.PrimeExceptionHandler` class and review the imports at the top of the file, confirm they are the `javax` imports.

### DIFF
--- a/plugins/maven/src/main/java/org/jetbrains/idea/maven/importing/workspaceModel/WorkspaceModuleImporter.kt
+++ b/plugins/maven/src/main/java/org/jetbrains/idea/maven/importing/workspaceModel/WorkspaceModuleImporter.kt
@@ -187,7 +187,7 @@ internal class WorkspaceModuleImporter(
     val libraryRootsProvider = {
       val classes = MavenImportUtil.getArtifactUrlForClassifierAndExtension(artifact, null, null)
       val javadoc = MavenImportUtil.getArtifactUrlForClassifierAndExtension(artifact, "javadoc", "jar")
-      val sources = MavenImportUtil.getArtifactUrlForClassifierAndExtension(artifact, "sources", "jar")
+      val sources = MavenImportUtil.getArtifactUrlForClassifierAndExtension(artifact,  if (artifact.classifier.isNullOrBlank()) "sources" else (artifact.classifier + "-sources"), "jar")
 
       // Keep the list of roots sorted by url to avoid extra "roots changed" events after loading the `.iml` files.
       // The `.iml` files keep the order of roots sorted.
--- a/plugins/maven/src/main/java/org/jetbrains/idea/maven/project/MavenProject.java
+++ b/plugins/maven/src/main/java/org/jetbrains/idea/maven/project/MavenProject.java
@@ -1197,6 +1197,9 @@ public class MavenProject {
       Pair<String, String> result = each.getExtraArtifactClassifierAndExtension(artifact, type);
       if (result != null) return result;
     }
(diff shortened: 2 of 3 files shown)


zulkar:
Thanks for the patch, unfortunately it cannot be accepted
Classifierm by definition is **some optional and arbitrary string** that - if present - is appended to the artifact name just after the version number. The classifier distinguishes artifacts that were **built from the same POM** but differ in content.

So, one cannot simply calculate the source classifier as classifier + "sources".
For your dependency, `org.primefaces:primefaces:14.0.4` and your classifier `jakarta` the source `jakarta-sources` is correct, indeed. But this is not true for the common case, since it is an arbitrary string, I don't know about such convention and can't find any in official maven docs, please let me know if I'm wrong.

For example, oauth.signpost:signpost-commonshttp4:2.1.1 has an artifact with `test` classifier, but does not have with `test-sources`. 

One more example:  Almost every other maven package has an artifact with `javadoc` classifier. You can even depend on it, but this does not mean that there should be an artifact with`javadoc-sources` classifier. Again, it is an arbitrary string, maven-source-plugin do not generate source artifact automatically for you project with classifier, there is no such convention

zulkar:
To address the issue, we should probably test both artifacts, with `sources` classifier and with `classifier+"sources"`, but it is pure heuristic thing.

Also, regarding for ther code. DependenciesImportingTest#testDependencyWithClassifier is a test for this.

Also, MavenAttachSourcesProvider should be fixed as well probably.

nrayburn-tech:
@zulkar thanks for reviewing.

I think that `javadoc` and `sources` are unique classifiers.  These classifiers are the default values for `maven-javadoc-plugin` and `maven-sources-plugin` so they will be very common.

Given that this only solves a single possible way of naming additional sources, maybe a different approach would be better.  Otherwise, there may end up being multiple naming styles requested (`${classifier}-sources` and `sources-${classifier}` for example).

Any improvement to using alternative sources would be helpful.  Would it be better to discuss ideas within the youtrack issue?

PR #2827, opened 2024-08-16
[kotlin] K2 J2K: Move VarToValProcessing LocalVarToValInspectionBasedProcessing and fixValToVarDiagnosticBasedProcessing to JKTree

jocelynluizzi13:
Clean commit of mutability inference for k2

This relies on the Java inference in order to decide whether a field should be `var` or `val` and then propagates this down the stack. There are some slight differences / things that would need to be refined for K1 and therefore I left the K1 postprocessing in place. It significantly improves K2 as evinced by the number of test cases that are changed.

(continuation of #2702 and #2726 and #[2743](https://github.com/JetBrains/intellij-community/pull/2743))


darthorimar:
@jocelynluizzi13 The MR is merged with a few commits on top
- daf3765276f54dac415034b9f9db087529d47b2f
- c615e2eb1d4f3e934fbcc56237775c3ad770b4cd

Thank you a lot for your contribution!

PR #2828, opened 2024-08-19
Support signature polymorphic methods in debugger evaluation

SirYwell:
Before, debugger evaluation did not work with signature polymorphic methods as the invocation behaves like reflective invocation:
<details>
<summary>Invocation of signature polymorphic methods throwing exceptions:</summary>

vh-broken
mh-invoke-broken

</details>

With this change, the methods are now invoked indirectly using the `MethodHandles` `invoker/exactInvoker/varHandleInvoker/varHandleExactInvoker` methods plus `invokeWithArguments`.

I made sure the code works with invoke behavior and invokeExact behavior, as well as properly handles primitive types and return type casts:

<details>
<summary>Simple VarHandle#getAndSet</summary>

vh-working

</details>

<details>
<summary>Type mismatch with invokeExact behavior VarHandle#getAndSet</summary>

evh-working-mismatch

</details>

<details>
<summary>Return type casting with invokeExact behavior VarHandle#getAndSet</summary>

evh-working

</details>

<details>
<summary>Simple MethodHandle#invoke with type conversion (unboxing, boxing due to Object return type)</summary>

mh-working-invoke

</details>

<details>
<summary>MethodHandle#invokeExact with type mismatch</summary>

mh-working-invokeExact-mismatch

</details>

<details>
<summary>MethodHandle#invokeExact with proper types</summary>

mh-working-invokeExact-working

</details>

I wasn't able to find out where to add tests covering this, please let me know if you want me to add tests for it.

I also made sure that methods returning a `VarHandle` are only invoked once, even though we need to dynamically check whether the VarHandle has invokeExact behavior or not. I also explored creating a string expression and let that get converted instead, but due to the single-invocation requirement, a single expression would be rather difficult, so I decided to manually build the needed elements instead.

### DIFF
--- a/java/debugger/impl/src/com/intellij/debugger/engine/evaluation/expression/EvaluatorBuilderImpl.java
+++ b/java/debugger/impl/src/com/intellij/debugger/engine/evaluation/expression/EvaluatorBuilderImpl.java
@@ -6,6 +6,7 @@
  */
 package com.intellij.debugger.engine.evaluation.expression;
 
+import com.intellij.codeInsight.AnnotationUtil;
 import com.intellij.codeInsight.daemon.JavaErrorBundle;
 import com.intellij.codeInsight.daemon.impl.HighlightInfo;
 import com.intellij.codeInsight.daemon.impl.analysis.HighlightUtil;
@@ -41,6 +42,7 @@ import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.lang.invoke.VarHandle;
 import java.util.*;
 
 public final class EvaluatorBuilderImpl implements EvaluatorBuilder {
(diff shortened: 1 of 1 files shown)


cypok:
I've created [IDEA-358378](https://youtrack.jetbrains.com/issue/IDEA-358378) Java evaluator incorrectly invokes signature polymorphic methods.

cypok:
Hi @SirYwell,

thank you for your pull request. It's great that you made us aware of the problem with invoking MH/VH methods, however, we had to reject your pull request.

The code is too complex and a bit fragile (e.g., it would fail to evaluate `(int)(mh.invokeExact(37))`), and relies heavily on JVM internals. We had another way to evaluate such complex stuff — send it to an actual compiler. E.g., lambdas instantiation works exactly like this. The [trivial fix](https://github.com/JetBrains/intellij-community/commit/a5afd574559c283a9e6d1904df36b2228f4a5e87) was already pushed.

Regarding unit tests. Unfortunately, the debugger’s unit tests are located in the closed-source part of IntelliJ.

SirYwell:
Hey @cypok thanks for taking a look.

Finding the correct return type is indeed too fragile, but the code doesn't rely on any JVM internals. I wasn't aware of the fallback with the compiler; that is actually pretty cool. However, running the code in a different method can result in different behavior in at least two scenarios I'm aware of:

First scenario: `StackWalker#getCallerClass()`
```java
MethodHandle getCallerClass = MethodHandles.lookup()
    .findVirtual(StackWalker.class, "getCallerClass", MethodType.methodType(Class.class))
    .bindTo(StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE));
```
Running `getCallerClass.invoke()` returns different results depending if it is called in the context of the current method or in a different method (note that the method returns the class that called the method we are currently in).

Second scenario: Lookup privileges
As your fix means the whole expression is compiled, it also means that a potential `MethodHandles.lookup()` call is moved into a fresh method in another class. This method is caller-sensitive too. Consider a class `ClassInPackage` with a private static field `i`. `(int) MethodHandles.lookup().findStaticGetter(ClassInPackage.class, "i", int.class).invoke();` will work in code, but not in the debugger expression evaluation due to an `IllegalAccessException` being thrown during evaluation in the latter case.

I didn't spend more time on this now, but I think the second scenario could be addressed by generating a hidden class in the nest of the current class instead (I assume only the defining mechanism would need to be changed then), but that only works on Java 15+.

By the way I also noticed that your current fix does not work when calling e.g. `MethodHandles.lookup().findStaticGetter(Inner.class, "i", int.class).invoke()` in an inner class if the enclosing class is in the default package, not sure how much of a problem that is and where it comes from.

cypok:
> but the code doesn't rely on any JVM internals
Sorry, wanted to say JLS/JVMS/Standard Library — we don't have control over it and it tends to change occasionally.

True, calling context-sensitive methods in the debugger evaluation window is not our top priority. :)
I will try to think if we can respect the original context better.

PR #2830, opened 2024-08-19
Fix race in IStubElementType (IJPL-160566)

gharrma:
Previously it queried isLazilyRegistered() before isInitialized(), leading to false-positive errors if initialization happens inbetween.

(Note that isLazilyRegistered() returns false if initialization has not happened yet.)

Issue link: https://youtrack.jetbrains.com/issue/IJPL-160566


maxmedvedev (commented):
LGTM

PR #2831, opened 2024-08-20
Fix error message warning about the wrong missing file in PathManager

Fox2Code:
The commit https://github.com/JetBrains/intellij-community/commit/70afc86a2730f37989dae2c4ee8f2538e9fa6328 changed which files where checked to find IDE home from  `bin/idea.properties`  
to `product-info.json` but the actual error message was never updated to account for that.

Related youtrack issue: https://youtrack.jetbrains.com/issue/IJPL-160565/Start-Failed-Could-not-find-installation-home-path.-Please-make-sure-bin-idea.properties-is-present-in-the-installation


ybedrov:
Issue was fixed

ybedrov:
Issue was fixed

PR #2833, opened 2024-08-20
Leveraging IntelliJ IDEA for Android Studio's Cloud Settings Sync

kunshenks:
Android Studio aims to implement settings synchronization using its own cloud service. To avoid redundant development, we propose that IntelliJ IDEA, which already possesses a robust settings sync plugin, introduce minor modifications to facilitate the replacement of the underlying sync service. This would allow Android Studio to leverage IntelliJ IDEA's existing sync functionality without altering the core features or user experience.


jreznot:
Thank you so much, we seems integrated most of suggestions here already and there is no API that supports both JetBrains and Google providers. Closing this one as superseded.

PR #2838, opened 2024-08-23
Making junit plugin using `default` test runner in case Spock testing framework is detected in classpath.

viktor-yengovatov:


BartvHelvert:
Hi! Thanks for your patch, would you be able to add a test?

viktor-yengovatov:
Hi, sure. I am new in the community and would like to see and example for tests here. Also scope of what is testable outcome.

BartvHelvert:
Looking at it again, it seems like testing would be challenging here. I'll run your patch on CI, if there aren't any problems, it should be merged today. The fix will then be available starting from the next 2024.3 EAP.

PR #2840, opened 2024-08-30
Only create FixAllIntention when quick fix is available for batch.

jenna-h:
I'm not sure if this breaks anything else, but I was trying to write a LocalInspection. I didn't like the behavior of the Fix All action, so I was able to disable it by specifying `availableInBatchMode = false`. However, the Fix All action is still generated, and then it gives an error ("Unfortunately [action] is currently not available for batch mode User interaction is required for each problem found") when the user tries to use it.


jenna-h:
I work at Google on Android Studio and wanted to request this feature as a result of my work.

jenna-h:
I filed a bug in YouTrack: https://youtrack.jetbrains.com/issue/IJPL-161378/FixAllIntention-is-generated-even-when-quick-fix-is-not-available-for-batch

cdracm:
pushed

PR #2841, opened 2024-09-02
Pin tooltip popups for keypress actions

kazariad:
When hovering the mouse cursor over a code element, the resulting tooltip/popup is opened in a temporary window which closes automatically when the mouse cursor moves away from it.  However, the behavior for popups activated via hotkey is inconsistent.

For example, the Quick Documentation action (Ctrl + Q) pins the tooltip and requires pressing Esc or clicking elsewhere to close (simply moving the mouse away isn't enough).  But other actions, such as Error Description (Ctrl + F1), don't pin the popup window and function like the normal mouse hover.  This patch changes the behavior of the following keypress actions to also pin their windows:

Next Highlighted Error (F2)
Previous Highlighted Error (Shift + F2)
Error Description (Ctrl + F1)
Hover Info

Before:
before

After:
after


yopox:
Thank you! This looks great for consistency.

PR #2842, opened 2024-09-02
[kotlin][debugger] Remove bytecode inspection for coroutine method switch check

nikita-nazarov:
When checking if a suspend method has a state machine, the debugger inspected bytecode trying to find an instruction like `getfield com/example/MainKt$foo$1.label:I`. This approach didn't work in Android Studio, which resulted in a stepping difference between IDEs. This commit introduces a simpler way of detecting if a state machine is generated for a suspend function.

@zuevmaxim

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/debuggerUtil.kt
+++ b/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/debuggerUtil.kt
@@ -5,6 +5,7 @@
 package org.jetbrains.kotlin.idea.debugger.core
 
 import com.intellij.debugger.engine.DebugProcessImpl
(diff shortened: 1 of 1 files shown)


mvicsokolova:
Probably this approach will not work for the cases with tail call optimization, when no state machine is generated for a suspend lambda. I'll check the cases like this.

@nikita-nazarov maybe you could share an example of code, where the behavior was different?

nikita-nazarov:
Hi @mvicsokolova! Can you please take a look at this pr?

mvicsokolova:
Hi! @nikita-nazarov, I've very sorry for the endless review of this PR, I've merged the fix for this issue into the master branch of ultimate. 
I've added your implementation for the case of Android in `doesMethodHaveSwitcher` method, and left the bytecode traversal looking for the `label` field in case of JVM.

PR #2845, opened 2024-09-04
Fix javadoc info generator array init printing

SirYwell:
Printing of array initializers in the javadoc info generator was wrong:
wrong0
(multiple dimension expressions are comma-separated in the last [])

wrong1
(dimension expressions are generally moved into the last [])

The problem stems from how the full expression is a combination of type information and dimension expressions. This patch doesn't use the generic type printing code but instead a mix of printing dimension expressions and type information.

> [!NOTE]  
> I initially wanted to include annotations on the dimensions too, but I noticed that IntelliJ generally seems to attach annotations in the wrong order to the dimensions of the type. I assume this should be addressed separately first.
> 
wrong2
(order of annotations is reversed - this problem remains as it is not caused by this part of the code)

One test needed to be adjusted because the formatting now wasn't stick together manually anymore. The visuals are equal though I think.


SirYwell:
Hey @dovchinnikov is there anything to do from my side here? Thanks.

BasLeijdekkers:
Hello @SirYwell ,

Is there a YouTrack issue for this problem? If not, could you create one?

SirYwell:
> Hello @SirYwell ,
> 
> Is there a YouTrack issue for this problem? If not, could you create one?

I didn't find anything related, so I created <https://youtrack.jetbrains.com/issue/IDEA-366529/>. This does not cover the problem with annotations, as that seems to be separate. Let me know if you need anything else.

BasLeijdekkers:
Thanks @SirYwell , I have applied your pull request with some minor modifications.
It would be great if you can file an issue about the annotations problem as well.

SirYwell:
> Thanks @SirYwell , I have applied your pull request with some minor modifications. It would be great if you can file an issue about the annotations problem as well.

Hi, I created https://youtrack.jetbrains.com/issue/IDEA-366918/ now. Please let me know if you need any more information.

PR #2847, opened 2024-09-16
Implement `MetaPathFinder.find_spec` method in `MyFinder` subclass

fumoboy007:
The `MyFinder` subclass of [`MetaPathFinder`](https://docs.python.org/3.11/library/importlib.html#importlib.abc.MetaPathFinder) currently only implements the deprecated `find_module` method. That deprecated method was [removed](https://github.com/python/cpython/pull/98059) in Python 3.12. This change adds an implementation for its replacement method called `find_spec`.

### DIFF
--- a/python/helpers/generator3/core.py
+++ b/python/helpers/generator3/core.py
@@ -592,6 +592,11 @@ def imported_names_collected():
             imported_names.add(fullname)
             return None
 
+        # noinspection PyMethodMayBeStatic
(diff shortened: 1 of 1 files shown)


east825:
Merged, thanks for the fix! Sorry it took so long.

east825:
Merged. Thanks for noticing and contributing the fix!

PR #2848, opened 2024-09-16
Update `six.py` to version 1.16.0

fumoboy007:
The included `six.py` is from version 1.9.0. This version is not compatible with Python 3.12, which [removed](https://github.com/python/cpython/pull/98059) the deprecated `MetaPathFinder.find_module` method. This change updates `six.py` to version [1.16.0](https://github.com/benjaminp/six/blob/65486e4383f9f411da95937451205d3c7b61b9e1/six.py), which uses the replacement `MetaPathFinder.find_spec` method.

Fixes [PY-75977](https://youtrack.jetbrains.com/issue/PY-75977/Skeleton-Generator-No-module-named-six.moves).

### DIFF
--- a/python/helpers/six.py
+++ b/python/helpers/six.py
@@ -1,6 +1,4 @@
-"""Utilities for writing code that runs on Python 2 and 3"""
-
-# Copyright (c) 2010-2015 Benjamin Peterson
+# Copyright (c) 2010-2020 Benjamin Peterson
 #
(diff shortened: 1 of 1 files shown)


DmPanov:
@fumoboy007 thank you for the PR, applied in https://github.com/JetBrains/intellij-community/commit/5fc17ccbb3de4ee5b9cb27cd9dac30f796be64fc

PR #2849, opened 2024-09-20
[kotlin][debugger] Add capability to fetch @kotlin.Metadata in the debugger

nikita-nazarov:
As personally discussed with @zuevmaxim

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/kotlin.jvm-debugger.core.iml
+++ b/plugins/kotlin/jvm-debugger/core/kotlin.jvm-debugger.core.iml
@@ -57,5 +57,7 @@
     <orderEntry type="module" module-name="intellij.platform.statistics" />
     <orderEntry type="module" module-name="intellij.platform.util" />
     <orderEntry type="module" module-name="kotlin.base.resources" />
+    <orderEntry type="library" name="gson" level="project" />
+    <orderEntry type="library" name="jetbrains.kotlinx.metadata.jvm" level="project" />
   </component>
 </module>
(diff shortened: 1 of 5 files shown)


plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/metadataUtil.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.intellij.debugger.engine.evaluation.EvaluationContext
+import com.sun.jdi.*
+import kotlinx.metadata.jvm.KotlinClassMetadata
+import org.jetbrains.kotlin.idea.debugger.base.util.wrapEvaluateException
+import org.jetbrains.kotlin.idea.debugger.base.util.wrapIllegalArgumentException
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
+
+fun ReferenceType.fetchKotlinMetadata(context: EvaluationContext): KotlinClassMetadata? {
+    val classObject = classObject() ?: return null
```
zuevmaxim, line 16: We can cache the result of this function to avoid redundant computations

nikita-nazarov: I also thought about creating a metadata cache. We need a cache that would invalidate itself when the debug process stops. The only issue is that I didn't find any 'debug process wise' caches in Intellij IDEA. Only project wide ones.

nikita-nazarov:
Closing in favor of another pr

PR #2851, opened 2024-09-27
Fix encoding of project name in doc url

andymagee:
JavaDocExternalFilter.getExternalDocInfoForElement does a comparison to determine if the given doc URL is coming from the built in server. However, it does this using a constructed string that fails to correctly encode the project name, so the comparison does not work for any project names that contains characters needing encoding (like a space).

Using Urls.newHttpUrl() to make the comparison string fixes the issue, since this is the same API used in BuiltInWebBrowserUrlProvider to create the doc URL in the first place.

### DIFF
--- a/java/java-impl/src/com/intellij/codeInsight/javadoc/JavaDocExternalFilter.java
+++ b/java/java-impl/src/com/intellij/codeInsight/javadoc/JavaDocExternalFilter.java
@@ -103,8 +103,9 @@ public class JavaDocExternalFilter extends AbstractExternalFilter {
     String externalDoc = null;
     myElement = element;
 
+    String authority = "localhost:" + BuiltInServerOptions.getInstance().getEffectiveBuiltInServerPort();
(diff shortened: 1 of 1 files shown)


BartvHelvert:
@andymagee hey, sorry for ignoring the review. Could you maybe create an issue in our issue tracker, describing the problem and also add a test?

andymagee:
Almost a year later I don't really have much context that I remember other than what I can quickly look up, so I'm not able to add any tests at this time.

From what I recall, I was able to reproduce this only in Android Studio; the code is in the IJ platform, but I vaguely recall that the documentation pathway used in IDEA was different and didn't trigger the same bug. Or something of the sort. The bug is there, so maybe it's triggered in a different way.

The Android Studio bug is at https://issuetracker.google.com/issues/369323271. Whenever a project name had a space in it, documentation would be misrendered because the documentation URL is malformed:

image

We've already fixed this in Android Studio with this same fix in our fork of the platform code, so it's up to you if you want to take the change or not.

PR #2853, opened 2024-10-01
Add EP to customize "Add Import" weights

andymagee:
Fix for https://youtrack.jetbrains.com/issue/KTIJ-31471

### DIFF
--- a/plugins/kotlin/code-insight/fixes-k2/resources/kotlin.code-insight.fixes.k2.xml
+++ b/plugins/kotlin/code-insight/fixes-k2/resources/kotlin.code-insight.fixes.k2.xml
@@ -2,6 +2,12 @@
   <dependencies>
     <module name="kotlin.refactorings.k2"/>
   </dependencies>
+  <extensionPoints>
+    <extensionPoint
+            qualifiedName="org.jetbrains.kotlin.expressionImportWeigherFactory"
+            interface="org.jetbrains.kotlin.idea.k2.codeinsight.fixes.imprt.ExpressionImportWeigherFactory"
+            dynamic="true"/>
+  </extensionPoints>
   <extensions defaultExtensionNs="org.jetbrains.kotlin">
     <codeinsight.quickfix.registrar implementation="org.jetbrains.kotlin.idea.k2.codeinsight.fixes.KotlinK2QuickFixRegistrar"/>
   </extensions>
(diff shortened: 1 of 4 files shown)


andymagee:
Friendly ping on this - any updates?

fedochet:
Hi @andymagee !

We actually already have a simlar EP for callables - `KotlinAutoImportCallableWeigher`

Could you please see if it is possible to extend it or introduce a new EP which would look similar to it?

andymagee:
> Hi @andymagee !
> 
> We actually already have a simlar EP for callables - `KotlinAutoImportCallableWeigher`
> 
> Could you please see if it is possible to extend it or introduce a new EP which would look similar to it?

Thank you, I had missed that existing EP!

The existing weigher uses `CallExpressionWeigher`/`CallExpressionImportWeigher` for all name references, even though the name references are not always call expressions (eg, in the case of type name references like `val foo: SomeUnimportedType`. I've updated the logic to more accurately handle type names that aren't callables, and added a number of test cases to validate those.

I'm also removing the `@Internal` markings for the EP, since the goal would be to use this in the Android Studio codebase.

Let me know what you think, thanks.

andymagee:
Hi there, any comments on the updated changes?

fedochet:
Sorry for the delay, I will take a look at it this week

andymagee:
Hi @fedochet , have you had a chance to take a look at this PR?

fedochet:
Hi again Andy! 

I have looked at your PR and the original issue once again, and I see that in the AS you've previously used `ComposeProximityWeigher` to achieve the same goal, and then it stopped working at some point

Unfortunately, I am not sure that I am (as the owner of the auto-import subsystem) ready to make the `KotlinAutoImportCallableWeigher` EP public in the current state - it definitely needs some extra-work IMO

My main concern is that in the current form the EP seems a little bit clumsy, and it is also somewhat under-used in our own code. Another concern is that it is not related to weighers, which seems like an overlooking from our part

If it's possible, I would like to postpone this PR for some time. I am currently woking on auto-import subsystem, and I would also pay attention to the existing EPs, and will take your paricual problem (https://youtrack.jetbrains.com/issue/KTIJ-31471) into consideration as well

WDYT?

andymagee:
That sounds like a reasonable plan, and I agree that it makes sense to have a single mechanism rather than multiple different types of weighers.

Do you have any timeline in mind for this work? Depending on when we would expect any updates, we may need to make some workarounds on our side for this in the meantime.

PR #2855, opened 2024-10-06
KTIJ-10950 Language injection only works for the first argument to a vararg parameter

Thunderblade73:
## Tracker
[KTIJ-10950](https://youtrack.jetbrains.com/issue/KTIJ-10950/Language-injection-only-works-for-the-first-argument-to-a-vararg-parameter)
## Issue
The issue was that in ```injectionForKotlinCall``` the index of the argument gets misaligned with the index of the ktFunction valueParameters.

## Solution
if the argumentIndex > index of the vararg then it needs to be the vararg. This works, since the Kotlin spec only allows for one vararg and any parameter after a vararg needs to be named, which means it does not need the index.

## Improvement
It would be better to save the varargIndex to the KtFunction object. But that change needs to be on the kotlin compiler side.

### DIFF
--- a/plugins/kotlin/injection/base/tests/test/org/jetbrains/kotlin/idea/base/injection/KotlinInjectionTestBase.kt
+++ b/plugins/kotlin/injection/base/tests/test/org/jetbrains/kotlin/idea/base/injection/KotlinInjectionTestBase.kt
@@ -330,7 +330,7 @@ abstract class KotlinInjectionTestBase : AbstractInjectionTest() {
         languageId = HTMLLanguage.INSTANCE.id, unInjectShouldBePresent = false
     )
 
-    fun testInjectionOffVarargParameterWithAnnotation() = doInjectionPresentTest(
+    fun testInjectionOfVarargParameterWithAnnotation() = doInjectionPresentTest(
         """
         import org.intellij.lang.annotations.Language
 
@@ -343,7 +343,7 @@ abstract class KotlinInjectionTestBase : AbstractInjectionTest() {
         languageId = HTMLLanguage.INSTANCE.id, unInjectShouldBePresent = false
     )
 
-    fun testInjectionOffVarargParameterWithOutAnnotation() = doInjectionPresentTest(
+    fun testNoFalseInjectionOfVarargParameter() = doInjectionPresentTest(
         """
         import org.intellij.lang.annotations.Language
 
--- a/plugins/kotlin/injection/base/tests/test/org/jetbrains/kotlin/idea/base/injection/KotlinInjectionTestBase.kt
+++ b/plugins/kotlin/injection/base/tests/test/org/jetbrains/kotlin/idea/base/injection/KotlinInjectionTestBase.kt
@@ -330,6 +330,33 @@ abstract class KotlinInjectionTestBase : AbstractInjectionTest() {
         languageId = HTMLLanguage.INSTANCE.id, unInjectShouldBePresent = false
     )
 
+    fun testInjectionOffVarargParameterWithAnnotation() = doInjectionPresentTest(
+        """
+        import org.intellij.lang.annotations.Language
+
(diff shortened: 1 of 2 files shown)


vladimirdolzhenko:
it does not work for the reported issue

image

or like this
image

plugins/kotlin/injection/base/tests/test/org/jetbrains/kotlin/idea/base/injection/KotlinInjectionTestBase.kt
```diff
@@ -330,6 +330,33 @@ abstract class KotlinInjectionTestBase : AbstractInjectionTest() {
...
+        )
+        """,
+        languageId = HTMLLanguage.INSTANCE.id, unInjectShouldBePresent = false
+    )
+
+    fun testInjectionOffVarargParameterWithOutAnnotation() = doInjectionPresentTest(
+        """
+        import org.intellij.lang.annotations.Language
+
+        fun bar(vararg s: String,@Language("HTML") s2 : String){}
```
vladimirdolzhenko, line 350: for the sake of clearness it would be better to use the same function signature as in previous test
```
fun bar(@Language("HTML") vararg s: String){}
```

Thunderblade73:
Oh, the last minute addition of the skip does eaxtly the opposite of what it should do.

plugins/kotlin/injection/base/tests/test/org/jetbrains/kotlin/idea/base/injection/KotlinInjectionTestBase.kt
Thunderblade73: That is intended to be a different signature. Since this test is for wrong injection of the second vararg with the injection of the 2 argument.

Thunderblade73: I renamed the test case to be more descriptive of what the test should test.

vladimirdolzhenko:
do you mind if I change a bit code style and namings before the merge of this PR ?

vladimirdolzhenko:
Tests are failing

Thunderblade73:
> do you mind if I change a bit code style and namings before the merge of this PR ?

No problem with it.

Thunderblade73:
> Tests are failing

I would fix them, but I can't run test at all. I won't bother you to much with that issue, but can you at least link me to someone/somewhere where I can get help?

Does one of the new test fail or old ones? Since when the new test fail I could maybe fix it, otherwise I would need to close the PR since I can't really work on failed test if I can't check them myself.

<details>
<summary>Details of the problem I have with the Testing</summary>

I either get a 
```Exception in thread "main" java.lang.NoSuchMethodError: 'java.util.Set org.junit.platform.engine.TestDescriptor.getAncestors()'```
 (Which is an error for mismatching junit versions) or I get
```java.lang.RuntimeException: Cannot find service org.jetbrains.kotlin.idea.base.plugin.KotlinPluginModeProvider (classloader=com.intellij.util.lang.UrlClassLoader@3d075dc0, client=null)```.
And I already invested 30h now into trying to fix those errors while not even seeing any slight change on the errors. I already tried to clone and setting it up again, with no effect. I followed everything in the readme. And I'm out of ideas how I could fix it.

(Windows 10, IntelliJ IDEA 2024.2.3, last commit from master ab1c28fe90b1da62680c47116f29db3fa59ca9ab)
</details>

vladimirdolzhenko:
Thank you so much for the contribution.

Merged as https://github.com/JetBrains/intellij-community/commit/b7a27fe7c328994f8b531ac511e4aca29cd7f89e

vladimirdolzhenko:
There are several small problems in infra `KotlinInjectionTestBase.kt` - it uses already injected file in `editor.file` that causes some issues. Will address it later.

PR #2856, opened 2024-10-08
Attempt to short-circuit directory enumeration

csrhodes:
In case-insensitive filesystems, canonizing a given name to the one in the filesystem is expensive if we need to enumerate directories. Attempt to compute an answer earlier by generating a full realpath, which should succeed in the absence of symbolic links.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/LocalFileSystemBase.java
+++ b/platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/LocalFileSystemBase.java
@@ -514,8 +514,25 @@ public abstract class LocalFileSystemBase extends LocalFileSystem {
     var originalFileName = file.getName();
     var t = LOG.isTraceEnabled() ? System.nanoTime() : 0;
     try {
+      var nioFile = convertToNioFileAndCheck(file, false);
       if (SystemInfo.isWindows) {
-        return convertToNioFileAndCheck(file, false).toRealPath(LinkOption.NOFOLLOW_LINKS).getFileName().toString();
+        return nioFile.toRealPath(LinkOption.NOFOLLOW_LINKS).getFileName().toString();
+      }
+      else {
+        // Not Windows (i.e. probably Unix, probably probably MacOS X with its case-insensitive FS).
+        //
+        // On OS X, we would like to query the filesystem directly for metadata, using an interface to the
+        // getattrlist() system call, which would return directly the metadata associated with the existing file, including
+        // its filesystem name.  However, we don't have an interface to that in the JVM as far as I am aware.
+        //
+        // Instead, handle one common case as quickly as possible: compute the file's realpath, resolving links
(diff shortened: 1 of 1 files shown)


csrhodes:
This needs testing before merge.  (I can't easily test it: I don't have access to a case-insensitive Unix filesystem, at least in these terms).

It's also far from ideal.  Although on OS X `realpath(3)` via `getattrlist(2)` is constant-time in the size of each directory's contents, it is linear-time in the depth of the directory.  That's a problem for this use case, which I believe calls `getCanonicallyCasedName` directly for each level of the directory hierarchy itself, so we end up with quadratic-time in the depth of the directory overall.

I can't see any proper fix other than a specialized implementation for OS X involving `getattrlist(2)` directly.  (on the assumption that OS X's default filesystem is the only case-insensitive filesystem worth supporting with any efficiency.)

platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/LocalFileSystemBase.java
```diff
@@ -517,6 +517,22 @@ public boolean markNewFilesAsDirty() {
...
+        //
+        // On OS X, we would like to query the filesystem directly for metadata, using an interface to the
+        // getattrlist() system call, which would return directly the metadata associated with the existing file, including
+        // its filesystem name.  However, we don't have an interface to that in the JVM as far as I am aware.
+        //
+        // Instead, handle one common case as quickly as possible: compute the file's realpath, resolving links
+        // (to avoid quadratic behaviour from directory listing).  In general this is not a suitable canonical name for the
+        // file, because links have been resolved, but if the result compares with case-insensitive equality with the given
+        // file's path, then the return value is suitable for use as a canonical name.
+        var resolvedRealPath = nioFile.toRealPath();
```
gottagofaster236, line 531: I think you want to add `Path nioFile = convertToNioFileAndCheck(file, false);` on line 516 for the code to compile.
Then you can replace the calls to `convertToNioPath...` on lines 518 and 537 with `nioFile`.

gottagofaster236:
This change does speed up the method significantly on my side with macOS, thanks a lot!
image
image

platform/platform-impl/src/com/intellij/openapi/vfs/impl/local/LocalFileSystemBase.java
csrhodes: Heh, yes, thank you.  (I did warn you that I hadn't tried compiling it...).  Updated to something hopefully a bit more correct.  (Not the one on line 537 which gets the nio file for `parent`).

csrhodes:
This might also speed up canonizing names of files which don't exist (yet) on the filesystem, if this ever happens: in a large directory, without `realpath`, we would have to enumerate everything in the directory before concluding that we should just return whatever the user gave us...

gottagofaster236 (approved):

PR #2857, opened 2024-10-08
[launcher] make CEF dependency optional

eugen-yakovets:
make CEF optional feature, enabled by default.
Use `cargo build --no-default-features` to build without CEF


eugen-yakovets:
@trespasserw

trespasserw:
Merged, thanks!

PR #2858, opened 2024-10-08
[launcher] Use /MANIFEST linker option only with MSVC

eugen-yakovets:


trespasserw:
I made a simpler change. Thanks!

eugen-yakovets:
Awesome, Thanks.

PR #2862, opened 2024-10-15
IDEA-113632 - Fix Geb Page content not being available

Poundex:
Fixes [IDEA-113632](https://youtrack.jetbrains.com/issue/IDEA-113632/Cannot-resolve-symbol-in-Geb-Spec)

When using Page objects, Geb makes the content from the Page definition available. However, IJ is currently not recognising this:

notworking

This change makes content from the current detected Page available:

working

Any method which changes the Page (`to()`, `via()`, `page()`) or asserts the current page (`at()`) will modify the current page, and only content from the last detected page will be in scope. The dynamic variables have the correct type, and navigate correctly back to their definitions in the Page. Completion is also working as expected:

completion

### DIFF
--- a/plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/geb/GebUtil.java
+++ b/plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/geb/GebUtil.java
@@ -2,21 +2,29 @@
 package org.jetbrains.plugins.groovy.geb;
 
 import com.intellij.psi.*;
+import com.intellij.psi.impl.source.PsiImmediateClassType;
 import com.intellij.psi.scope.PsiScopeProcessor;
 import com.intellij.psi.util.CachedValueProvider.Result;
 import com.intellij.psi.util.CachedValuesManager;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


plugins/groovy/groovy-psi/src/org/jetbrains/plugins/groovy/geb/GebUtil.java
```diff
@@ -37,12 +45,81 @@ public static boolean contributeMembersInsideTest(PsiScopeProcessor processor,
...
+        }
+      }
+    }
+  }
+
+  
+  private static @Nullable PsiClass findPageChange(PsiElement place, PsiClassType pageType) {
+    PsiElement currentElement = place;
+    PsiClass ourPage = null;
+    while (currentElement != null) {
```
Hardes1, line 89: It seems that this method will be invoked for each reference inside the method, but you actually traverse the whole method starting from the innermost block to the outermost. Could you, please, rewrite it in the way like:
```
GrCodeBlock block = PsiTreeUtil.getParentOfType(place, GrCodeBlock.class):
for (var statement : block.getStatements()) {
   ...
}
```

Hardes1 (changes requested):

PR #2866, opened 2024-10-21
Introduced `setupEditor` parameter to allow for editor customization

Jonatha1983:
Refactor 'createIn' to 'create' for clarity and consistency. Introduced `setupEditor` parameter to allow for editor customization and condensed formatting in the `mapToValueModel` function.


Jonatha1983:
I had to change the internal `creatIn` function because of overloading, and since no one is using it outside this Class, I made that private.

PR #2868, opened 2024-10-24
IDEA-340866 [gradle] fix converting maven dependencies when pasted into gradle kotlin DSL buildscript

Flame239:
Fixes [IDEA-340866](https://youtrack.jetbrains.com/issue/IDEA-340866/Maven-to-gradle-snippet-conversion-doesnt-work-with-kotlin-DSL)

IntelliJ has a feature to convert copy-pasted maven dependency into gradle syntax, but as for now it doesn't support Kotlin DSL buildscripts, only Groovy.

This PR adds support for Kotlin DSL buildscripts as well as allows pasting dependencies without version (which are still valid as long as BOM exists).

Preview of how it works:
IDEa-dependencies-pasting

@nskvortsov 
Tagging you because you are assignee of the ticket


nskvortsov:
The PR was successfully merged. I added a couple more commits on top, mostly moving the solution to Kotlin. Thank you for your contribution!

PR #2871, opened 2024-10-28
[kotlin][debugger] Don't throw when exception occurred in the completable future

nikita-nazarov:
Sometimes when computing classes with inlined code, an ObjectCollectedException could be thrown in on of the futures. This resulted in failure of `CompleatableFuture.allOf(*futures)`. Because of this, smart stepping and breakpoints in inline functions sometimes didn't work, which was easy to observe when debugging on Android.

@zuevmaxim

### DIFF
--- a/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinPositionManager.kt
+++ b/plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinPositionManager.kt
@@ -588,9 +588,12 @@ class KotlinPositionManager(private val debugProcess: DebugProcess) : MultiReque
     private fun getClassesWithInlinedCode(candidatesWithInline: List<String>, line: Int): List<ReferenceType> {
         val candidatesWithInlineInternalNames = candidatesWithInline.map { it.fqnToInternalName() }
         val futures = debugProcess.virtualMachineProxy.allClasses().map { type ->
-            hasInlinedLinesToAsync(type, line, candidatesWithInlineInternalNames).thenApply { hasInlinedLines ->
-                type.takeIf { hasInlinedLines }
-            }
+            hasInlinedLinesToAsync(type, line, candidatesWithInlineInternalNames)
+                .thenApply { hasInlinedLines -> type.takeIf { hasInlinedLines } }
+                .exceptionally { e ->
+                    val exception = DebuggerUtilsAsync.unwrap(e)
+                    if (exception is ObjectCollectedException) null else throw e
+                }
         }.toTypedArray()
         CompletableFuture.allOf(*futures).join()
         return futures.mapNotNull { it.get() }


plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinPositionManager.kt
```diff
@@ -610,9 +610,9 @@ class KotlinPositionManager(private val debugProcess: DebugProcess) : MultiReque
     private fun getClassesWithInlinedCode(candidatesWithInline: List<String>, line: Int): List<ReferenceType> {
         val candidatesWithInlineInternalNames = candidatesWithInline.map { it.fqnToInternalName() }
         val futures = debugProcess.virtualMachineProxy.allClasses().map { type ->
-            hasInlinedLinesToAsync(type, line, candidatesWithInlineInternalNames).thenApply { hasInlinedLines ->
```
ZetaMaster777, line 613: - [ ]

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinPositionManager.kt
```diff
@@ -610,9 +610,9 @@ class KotlinPositionManager(private val debugProcess: DebugProcess) : MultiReque
     private fun getClassesWithInlinedCode(candidatesWithInline: List<String>, line: Int): List<ReferenceType> {
         val candidatesWithInlineInternalNames = candidatesWithInline.map { it.fqnToInternalName() }
         val futures = debugProcess.virtualMachineProxy.allClasses().map { type ->
```
ZetaMaster777, line 612:
```suggestion
         val futures = debugProcess.virtualMachineProxy.allClasses().map { type ->
```

zuevmaxim:
@nikita-nazarov Could you please attach the error with `ObjectCollectedException`?

zuevmaxim:
@nikita-nazarov Thank you! I can reproduce the issue, but I cannot understand which object is causing it. I see that the exception is thrown during the processing of some `android.*` classes. Maybe we can handle the exception in some explicit way at the place where it is thrown?
Anyway, could you please specify that we expect `ObjectCollectedException`? smth like this
```
                .exceptionally { e ->
                    val exception = DebuggerUtilsAsync.unwrap(e)
                    if (exception is ObjectCollectedException) null else throw e
                }
```

nikita-nazarov:
@zuevmaxim Hi Maksim!
I've applied the requested change. 
Speaking of handling the exception explicitly where it is thrown, the `com.sun.jdi.ReferenceType` documentation says:
> Any method on ReferenceType or which directly or indirectly takes ReferenceType as parameter may throw ObjectCollectedException if the mirrored type has been unloaded.

It means that the exception can be thrown anywhere, for example in `com.jetbrains.jdi`. I think it is totally fine to handle this exception in one place, namely `KotlinPositionManager`.

PR #2875, opened 2024-11-07
Make OrderRootsCacheBridge atomic

csrhodes:
Before this change, there is a window where an unfortunately-timed rootsChanged event would cause a crash, as the atomic references would be nulled out between the check for their existence and their use.  Fix this by returning a cache entry from the method that ensures its presence.  To do this reasonably, unify the cache for Urls and VirtualFiles.

Change-Id: I2264c0a726207a81d716599c2f10088c7ac32770


andrei-kuznetsov:
Merged. Thanks!

PR #2879, opened 2024-11-13
[javafx] IDEA-343577 Make Kotlin DSL the default for new Java/Kotlin …

rzillmann:
…JavaFX Projects

With this commit Kotlin DSL becomes the default for new JavaFX projects that use gradle as the build tool. Groovy projects still use the Groovy DSL

### DIFF
--- a/plugins/javaFX/resources/fileTemplates/j2ee/javafx-build.gradle.ft
+++ b/plugins/javaFX/resources/fileTemplates/j2ee/javafx-build.gradle.ft
@@ -95,8 +95,12 @@ dependencies {
 }
 
 test {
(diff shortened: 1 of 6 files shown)


bartekpacia:
Hi @rzillmann 👋🏻 Thank you so much for your contribution!

I think there's no point in supporting Gradle Groovy DSL, and it's fine to just use Gradle Kotlin DSL in templates – I'll take care of this.

rzillmann:
@bartekpacia FYI: I ditched the "if groovy use grooy" part in the Kotlin DSL script. So if you drop the groovy DSL even for groovy project that part needs be added back in.

bartekpacia:
Yeah, I noticed that and added it back; Thanks for letting me know though!

PR #2885, opened 2024-12-03
Implement CyclePasteAction.

sandipchitale:
I am trying to contribute an action which will supplement the Paste from History... action of JetBrains IDEs. Here is my idea:
My action will do the following:
It will check if the same action was invoked within the second.
  If it was not:
    It will simply behave like the standard pastea action and paste the first item in the hstory. Reset an index counter to 0.
  If it was:
    It will first invoke the undo action to undo the last paste, then increment the index and then paste the next item in the clipboard history.
This basically behaves like a cycle paste from the history without invoking the Clipboard History chooser.


jreznot:
Hi, although the behavior is potentially useful, I think it is far from the expected default behavior for the IDE. I'd suggest implementing it as a plugin instead

PR #2886, opened 2024-12-03
Fix wrong intention when referencing parameters in static fields

SirYwell:
Local classes (and interfaces) can reference parameters from the enclosing method, however not in static contexts.
However, IntelliJ currently incorrectly suggests making the parameter static. This should be avoided.

As `PsiParameter` also covers variables in for-each loops, pattern variables, and catch variables, those were affected too.

Example:
```java
public class StaticReference {

    void m(int p) {
        int lv;
        if (null instanceof Object tp) {
            for (Object ef : new Object[0]) {
                class C {
                    static int LV = lv; // works for local variables correctly, everything else gets the wrong intention
                    static int P = p;
                    static Object TP = tp;
                    static Object EF = ef;
                }
            }
        }
    }
}
```

### DIFF
--- a/java/java-analysis-impl/src/com/intellij/codeInsight/daemon/impl/analysis/HighlightFixUtil.java
+++ b/java/java-analysis-impl/src/com/intellij/codeInsight/daemon/impl/analysis/HighlightFixUtil.java
@@ -199,7 +199,7 @@ public final class HighlightFixUtil {
     if (place instanceof PsiReferenceExpression && place.getParent() instanceof PsiMethodCallExpression) {
       ReplaceGetClassWithClassLiteralFix.registerFix((PsiMethodCallExpression)place.getParent(), info);
     }
-    if (refElement instanceof PsiJvmModifiersOwner) {
+    if (refElement instanceof PsiJvmModifiersOwner && !(refElement instanceof PsiParameter)) {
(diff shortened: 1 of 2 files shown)


BartvHelvert:
@SirYwell thanks for the PR, could you maybe create an issue for it in our [issue tracker](https://youtrack.jetbrains.com/issues/IDEA), also could you add a test?

SirYwell:
> @SirYwell thanks for the PR, could you maybe create an issue for it in our [issue tracker](https://youtrack.jetbrains.com/issues/IDEA), also could you add a test?

Thanks, I created <https://youtrack.jetbrains.com/issue/IDEA-366531/>. I didn't find any related tests, where would they be placed best?

BasLeijdekkers:
@SirYwell take a look at the `com.intellij.codeInsight.daemon.impl.quickfix.ModifierTest` class.

SirYwell:
> @SirYwell take a look at the `com.intellij.codeInsight.daemon.impl.quickfix.ModifierTest` class.

Thanks. I added a test case there. I combined the changes in one commit as that makes things easier for you as far as I know. Let me know if there's anything else to do from my side.

BartvHelvert:
Looks good, thank you, the change was just merged with some minor adjustments to the test data.

PR #2889, opened 2024-12-08
IJPL-43725: New UI: allow showing expanded menu and other toolbar wid…

sandipchitale:
…gets in window header

The idea is to add an opt-in feature controlled by a property:

```
ide.main.menu.expand.horizontal.sticky=false
```

to fix the issue IJPL-43725

By default the property value is false and thus the existing behavior is maintained. If set to true the menu bar is always shown and does not hide.

Obviously this works best on wide screen monitors and no actions have been added to the `left` section of the toolbar.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/wm/impl/customFrameDecorations/header/toolbar/ExpandableMenu.kt
+++ b/platform/platform-impl/src/com/intellij/openapi/wm/impl/customFrameDecorations/header/toolbar/ExpandableMenu.kt
@@ -70,10 +70,19 @@ internal class ExpandableMenu(
         updateBounds()
       }
     })
+    if (isSticky()) {
+      SwingUtilities.invokeLater {
+        switchState();
+      }
+    }
   }
 
   fun isEnabled(): Boolean {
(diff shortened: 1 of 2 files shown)


platform/platform-impl/src/com/intellij/openapi/wm/impl/customFrameDecorations/header/toolbar/ExpandableMenu.kt
```diff
@@ -70,12 +70,21 @@ internal class ExpandableMenu(
...
+      }
+    }
   }
 
   fun isEnabled(): Boolean {
     return !SystemInfoRt.isMac && Registry.`is`("ide.main.menu.expand.horizontal")
   }
 
+  fun isSticky(): Boolean {
+    return isEnabled() && !SystemInfoRt.isMac && Registry.`is`("ide.main.menu.expand.horizontal.sticky")
```
vldF, line 85: It would be useful for macOS. There is a registry keys to force IntelliJ one menu on this OS (`apple.laf.useScreenMenuBar` or `jbScreenMenuBar.enabled` I can't remember which one). It is useful for automated UI tests

PR #2893, opened 2024-12-12
IJPL 173473: fix ArrayIndexOutOfBoundsException when using ContainerUtil.concat()

dieterstueken:
### Context

The method [ContainerUtil.concat(list1, list2)](https://github.com/JetBrains/intellij-community/blob/7b91998a9b35422db4ed6aeb85669ae1ee0dd22e/platform/util/src/com/intellij/util/containers/ContainerUtil.java#L1450) works fine with immutable lists. However, if the lists are modified concurrently, iteration may fail. see: [IJPL-173473](https://youtrack.jetbrains.com/issue/IJPL-173473)

This fix contains a unit test `ContainerUtilTest.testConcatenatedDynamicListsAreIterable` to disclose the problem.
Surprisingly it does not indicate any problem, if you start joining two empty lists. This is, because the size of the composed list is calculated in advance and never change. So, while an underlying `List` is growing, the composed `List` just does nothing. But I think this is also not the expected behavior. But if any `List` shrinks, the concatenated `List` throws an `IndexOutOfBoundsException`.

To fix this, I introduced an `iterator()` method for the concatenated `List` to get the actual elements from both lists.

Other utility methods here, using a straightforward `new AbstractList<T>`, may suffer from the same problem but are not fixed so far.

### DIFF
--- a/platform/platform-tests/testSrc/com/intellij/util/containers/ContainerUtilTest.java
+++ b/platform/platform-tests/testSrc/com/intellij/util/containers/ContainerUtilTest.java
@@ -18,6 +18,7 @@ import java.util.*;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.IntStream;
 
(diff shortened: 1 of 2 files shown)


cdracm:
pushed, thanks

PR #2894, opened 2024-12-18
Remove dependency on LFS64 and use standard functions

krystalgamer:
LFS64 is a legacy non-standard interface.

On version 1.2.4 musl removed support for LFS64 - https://musl.libc.org/releases.html - which broke the usage of JetBrains IDEs on musl-based systems.


trespasserw:
Complete removal of that code was overdue. Thanks for pushing in right direction!

PR #2897, opened 2025-01-08
Allow import of run configuration folderName from Gradle / External Project Model

shartte:
This enables https://github.com/JetBrains/gradle-idea-ext-plugin/issues/87 to be solved, by importing the `folderName` model property for run configurations.

Empty strings are converted to `null` here, since the IDE expects `null` for "no folder" and doesn't react nicely to empty strings being set.

Tested using an adapted idea-ext to work (these run configs were imported from Gradle, including the folders):
image

No tests were adapted since I couldn't find any tests that actually tested this external model import.

---

Example on how to make use of this with idea-ext, without having to build a custom plugin:

```
var appRun = new Application("name", project) {
    @SuppressWarnings("unchecked")
    @Override
    public Map<String, ?> toMap() {
        var result = (Map<String, Object>) super.toMap();
        result.put("folderName", "RUN SUBFOLDER");
        return result;
    }
};
runConfigurations.add(appRun); // add as normal to the container
```

### DIFF
--- a/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/project/settings/RunConfigurationHandler.kt
+++ b/platform/external-system-impl/src/com/intellij/openapi/externalSystem/service/project/settings/RunConfigurationHandler.kt
@@ -49,6 +49,7 @@ internal class RunConfigurationHandler : ConfigurationHandler {
       try {
         importer.process(project, runnerAndConfigurationSettings.configuration, cfg as Map<String, *>, modelsProvider)
         if (!isDefaults) {
+          runnerAndConfigurationSettings.folderName = cfg["folderName"] as? String ?: ""
           runManager.addConfiguration(runnerAndConfigurationSettings)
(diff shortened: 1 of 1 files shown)


nskvortsov:
> Thank you!! If you want I can do a corresponding PR to idea ext when this release lands.

That would be very welcome! You do not have to wait until the release date. As soon as the first EAP version is released, you will be able to experiment with it.

ov7a:
@nskvortsov It seems like that this update combined with issues like [that](https://youtrack.jetbrains.com/issue/IJPL-14171/Run-Configurations-Folders-inconsistent-UI-behaviors) one may cause issues in 2025.1 EAP:
image
Can you take a look, please?

ov7a:
> Hmmm, this change only applies to the external model import from Gradle, so it should have no effect.

Yep, these configurations were created by the gradle-idea-ext-plugin plugin.

PR #2898, opened 2025-01-09
[IJPL-174685] Do not continue processing if kotlin data node is null

karimai-gg:
the search for the dataNode uses a name comparison, and for the case of Android for example, the names won't match and we will end up finding nothing. The function should not continue processing data if the dataNode is null because it will just use fallbacks instead of configuring the right way. For the NAdroid case, we even have our own data service that actually passes the right name in this case.


CherepanovAleksei:
Merged, see [KTIJ-32864](https://youtrack.jetbrains.com/issue/KTIJ-32864) configureFacetByGradleModule should not try to guess information if it can't find the dataNode

PR #2899, opened 2025-01-09
Introduce JdiHelperClassLoader extension point

nikita-nazarov:
Currently some functionality of Intellij Kotlin and Java debuggers rely on helper classes, that are located in the `com.intellij.rt.debugger` module. The debugger loads these classes to the debuggee VM and then invokes corresponding helper methods to implement some of its functionality.

Unfortunately, when debugging Android applications these classes are not loaded. The reason is that you simply can't load a .class file on ART, because it only understands DEX files. Because of that Android Studio users are missing out on the  functionality provided by these helper classes.

For example, the `com.intellij.rt.debugger.coroutines.CoroutinesDebugHelper` class significantly speeds up async stack trace fetching time. The time difference is even more noticeable when debugging Android applications, because JDI communication with a phone is very expensive: 

Async stack trace fetching time between different implementations and devices

This pull request introduces an extension point that would allow the Android Studio debugger to implement its own helper class loading functionality.

@zuevmaxim 
@gorrus

### DIFF
--- a/java/debugger/impl/src/com/intellij/debugger/impl/JdiHelperClassLoader.java
+++ b/java/debugger/impl/src/com/intellij/debugger/impl/JdiHelperClassLoader.java
@@ -1,0 +1,16 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.debugger.impl;
+
+import com.intellij.debugger.engine.evaluation.EvaluateException;
+import com.intellij.debugger.engine.evaluation.EvaluationContextImpl;
+import com.intellij.openapi.extensions.ExtensionPointName;
+import com.sun.jdi.ClassType;
+import org.jetbrains.annotations.Nullable;
+
+public interface JdiHelperClassLoader {
+  ExtensionPointName<JdiHelperClassLoader> EP_NAME =
+    ExtensionPointName.create("com.intellij.debugger.jdiClassLoader");
+
+  @Nullable ClassType getHelperClass(Class<?> cls, EvaluationContextImpl evaluationContext,
+                                     String... additionalClassesToLoad) throws EvaluateException;
+}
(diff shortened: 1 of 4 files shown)


java/debugger/impl/src/com/intellij/debugger/impl/JdiHelperClassLoaderImpl.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Objects;
+
+public class JdiHelperClassLoaderImpl implements JdiHelperClassLoader {
+  @Override
+  public @Nullable ClassType getHelperClass(Class<?> cls, EvaluationContextImpl evaluationContext,
```
nikita-nazarov, line 23: This is just a copy of what used to be in `java/debugger/impl/src/com/intellij/debugger/impl/ClassLoadingUtils.java`

java/debugger/impl/src/com/intellij/debugger/impl/JdiHelperClassLoader.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.intellij.debugger.engine.evaluation.EvaluateException;
+import com.intellij.debugger.engine.evaluation.EvaluationContextImpl;
+import com.intellij.openapi.extensions.ExtensionPointName;
+import com.sun.jdi.ClassType;
+import org.jetbrains.annotations.Nullable;
+
+public interface JdiHelperClassLoader {
+  ExtensionPointName<JdiHelperClassLoader> EP_NAME =
+    ExtensionPointName.create("com.intellij.debugger.jdiClassLoader");
```
zuevmaxim, line 12: The name here is incorrect, but I'll fix it

zuevmaxim (approved):

PR #2900, opened 2025-01-13
Fix incorrect index assignment in toggleKeySelection by adding .key comparison

Moriafly:
Missing .key caused the index check to always be -1, resulting in lastActiveItemIndex being set to -1 after Ctrl + left mouse click selecting an item. This led to an array out-of-bounds error in the onExtendSelectionToKey method when using Shift to select items.

- Corrected the logic in toggleKeySelection to properly compare keys.
- Changed `selectableLazyListState.lastActiveItemIndex = allKeys.indexOfFirst { it == key }` to `selectableLazyListState.lastActiveItemIndex = allKeys.indexOfFirst { it.key == key }`.


rock3r (approved):
Thanks! Well spotted :)

PR #2906, opened 2025-01-14
[JEWEL-711] Add inline banner

hamen:
Reference https://github.com/JetBrains/jewel/pull/711


rock3r (approved):
Already approved on the Jewel repo

PR #2907, opened 2025-01-14
[JEWEL-720] Add slot APIs to GroupHeader

hamen:
Reference https://github.com/JetBrains/jewel/pull/720


rock3r (approved):
Already approved on the Jewel repo

PR #2817, opened 2024-07-20
IJPL-156693 detect local git-bash installation

Den-dp:
Detects Git for Windows installed in user scope via `winget install Git.Git --scope user`, allowing "Git Bash" action to show in the list of Terminal Predefined Actions.

More details: https://youtrack.jetbrains.com/issue/IJPL-156693/

PR #2917, opened 2025-01-17
IJPL-156693 list local or global git-bash in `Tools | Terminal` settings

Den-dp:
Detects Git for Windows installed in either `machine` or `user` scope (using `winget install Git.Git --scope user` for non-admin accounts), enabling the "Git Bash" shell to appear in the `Shell Paths` list under `Tools | Terminal` settings.

More details: https://youtrack.jetbrains.com/issue/IJPL-156693#focus=Comments-27-11392437.0-0

Refs: #2817


Den-dp:
@segrey sounds great, thanks for the update

PR #2918, opened 2025-01-19
Optimize the url filter by using non-capturing regex groups

borissmidt:
It seems especially the URL_PATTERN couldn't be optimized by the regex compiler because all groups are capturing.

I see this all over the cod ebase

### DIFF
--- a/platform/util/src/com/intellij/util/io/URLUtil.java
+++ b/platform/util/src/com/intellij/util/io/URLUtil.java
@@ -29,10 +29,10 @@ public final class URLUtil {
   public static final String JRT_PROTOCOL = "jrt";
   public static final String JAR_SEPARATOR = "!/";
 
-  public static final Pattern DATA_URI_PATTERN = Pattern.compile("data:([^,;]+/[^,;]+)(;charset[=:][^,;]+)?(;base64)?,(.+)");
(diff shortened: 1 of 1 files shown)


borissmidt:
I've created a ticket:
https://youtrack.jetbrains.com/issue/GO-18089/improve-terminal-speed-for-GoConsoleFilter

> Unfortunately, those are public fields with both internal and external (i.e., by 3rd-party plugins) usages relying on the group numbers, so we can't just change the patterns.

As an alternative could we deprecate the existing ones and create a '_V2' version which should be used instead?

Or to limit the scope,  these new regexes could be used just in the PatternBasedFileHyperlinkRawDataFinder.

KonstantinHudyakov:
@borissmidt hello! Thank for investigation and change proposal!
I've merged your change and applied some changes on top to preserve the compatibility with existing usages.
Optimised regexes are now used in UrlFilter: [IJPL-198475](https://youtrack.jetbrains.com/issue/IJPL-198475) and GoConsoleFilter: [IJPL-184427](https://youtrack.jetbrains.com/issue/IJPL-184427)

Sorry for such a long review.

borissmidt:
Thank you, i'm happy that it got merged. I hope the intellij console will feel more snappy after this.
:+1:

borissmidt:
@KonstantinHudyakov I still have one more here related to the capturing regexes:
https://github.com/JetBrains/jediterm/pull/296

PR #2924, opened 2025-01-21
[JEWEL-TBD] Load markdown images using Coil3

obask:
It supports every image as an inline node


obask:
Screenshot 2025-01-24 at 1 19 26 PM
Screenshot 2025-01-24 at 1 19 31 PM
currently there are 2 differences with a JCEF render: JCEF renders SVG as a giant image while we keep it small(I guess it just depends on a default size parameter), and the text in `build|failing` is not shown for some reason...

rock3r:
They look like entirely different images... Are you sure it's the same image it's being displayed, and that it's loading correctly in the first case? That to me looks like a "broken image" placeholder (which also is way too big)

SVG files have an intrinsic size declared in their root node, and any renderer should adhere to that

obask:
images I posted earlier are examples of how JCEF renders my "README.md" which this is how compose shows the same things:
Screenshot 2025-01-24 at 3 27 15 PM
[README.md](https://github.com/user-attachments/files/18538232/README.md)

rock3r:
I see. I had misinterpreted the original message, sorry 🙏 

The first image (the broken square) has a declared size of 800x800 px, encoded as `width="800px" height="800px"`, which browsers seem to accept but maybe the Compose renderer doesn't like, and falls back to the viewBox size of `16x16`. That would explain why it's larger on JCEF.

As for the latter, I think it's because the text for whatever reason just doesn't get rendered:

```xml
<g text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
    <use x="5.5" y="2" fill="#010101" fill-opacity=".3" href="#logo" />
    <use x="5.5" y="1" fill="#fff" href="#logo" />
  <text x="40.5" y="15" fill="#010101" fill-opacity=".3">build</text>
  <text x="40.5" y="14" fill="#fff">build</text>
  <text x="83.5" y="15" fill="#010101" fill-opacity=".3">passing</text>
  <text x="83.5" y="14" fill="#fff">passing</text>
</g>
```

Again, possibly some unsupported value in there that Skia ignores.

That's sort of weird since Skia can interpret image sizes with units, and render text — it does in Chrome! But maybe there is some setup that is missing, or Chrome does some magic stuff

obask:
This fixes both:
* https://github.com/JetBrains/jewel/issues/472
* https://github.com/JetBrains/jewel/issues/746

obask:
> The first image (the broken square) has a declared size of 800x800 px, encoded as `width="800px" height="800px"`, which browsers seem to accept but maybe the Compose renderer doesn't like, and falls back to the viewBox size of `16x16`. That would explain why it's larger on JCEF.
> 
> As for the latter, I think it's because the text for whatever reason just doesn't get rendered:
> 
> Again, possibly some unsupported value in there that Skia ignores.

I'm going to create a separate issue for SVG regressions. Should it go to youtrack?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -439,6 +472,46 @@ public open class DefaultMarkdownBlockRenderer(
...
+                        // make sure image doesn't get downscaled to the placeholder size
+                        .size(Size.ORIGINAL)
+                        .build(),
+                contentDescription = image.title,
+                onSuccess = { state ->
+                    if (knownSize.value == null) {
+                        knownSize.value = state.result.image.let { ImageSize(it.width, it.height) }
+                    }
+                },
+                // TODO: log onError as warning
```
obask, line 509: Is there an easy way to log errors inside of compose rendering?

rock3r:
> Should it go to youtrack?

Please, no. Stick with GH until further notice. You'll know together with everyone else when YT Is usable. We're still waiting.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
```diff
@@ -76,6 +80,7 @@ public fun Markdown(
     markdownStyling: MarkdownStyling = JewelTheme.markdownStyling,
     blockRenderer: MarkdownBlockRenderer = DefaultMarkdownBlockRenderer(markdownStyling),
 ) {
+    setSingletonImageLoaderFactory(::createImageLoader)
```
rock3r, line 83: Does this belong here? It feels like it's init code, so having it once only (e.g., in `ProvideMarkdownStyling`?) would make more sense to me

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
```diff
@@ -131,3 +137,17 @@ public fun LazyMarkdown(
...
 }
+
+private const val IMAGES_MEMORY_CACHE_SIZE = 24L * 1024 * 1024 // 24mb
+
+/**
+ * This method sets up an image loader with a memory cache but disables the disk cache. Disabling the disk cache is
+ * necessary because Coil crashes when attempting to use the file system cache with IDEA platform.
+ *
+ * Otherwise, Coil3 will throw java.lang.NoSuchMethodError: kotlinx.coroutines.CoroutineDispatcher.limitedParallelism
+ */
```
rock3r, line 148: This feels like it should be tracked in an issue for a follow-up

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -236,7 +236,7 @@ public class MarkdownProcessor(
     }
 
     private fun Node.tryProcessMarkdownBlock(): MarkdownBlock? =
-        // Non-Block children are ignored
+        // Nodes that are not blocks or unsupported types are ignored
```
rock3r, line 239: What does "Nodes that are not [...] unsupported types are ignored" mean? I don't think this is clearer than what was here before

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
```diff
@@ -77,17 +77,7 @@ public open class DefaultInlineMarkdownRenderer(private val rendererExtensions:
...
-                        buildString {
-                            appendLine(child.source)
-                            append(child.alt)
-                            if (!child.title.isNullOrBlank()) {
-                                appendLine()
-                                append(child.title)
-                            }
-                        },
-                    )
+                    appendInlineContent(child.source, "![${child.title}](...)")
```
rock3r, line 80:
```suggestion
                    appendInlineContent(child.source, "![${child.title}](child.source)")
```

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -124,9 +137,29 @@ public open class DefaultMarkdownBlockRenderer(
...
+                            if (item.source.isNotBlank()) add(item)
+                        }
+                        is WithInlineMarkdown -> {
+                            collectImagesRecursively(item.inlineContent)
+                        }
+
+                        else -> {}
+                    }
+                }
+            }
```
rock3r, line 159: This doesn't need to be a local function, does it?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -124,9 +137,29 @@ public open class DefaultMarkdownBlockRenderer(
                     .clickable(interactionSource = interactionSource, indication = null, onClick = onTextClick),
             text = renderedContent,
             style = mergedStyle,
+            inlineContent = renderedImages(block.inlineContent),
         )
     }
 
+    private fun getImages(input: List<InlineMarkdown>) =
```
rock3r, line 144: Nit: the argument could be a `WithInlineMarkdown` instead of the list

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
rock3r: You can use JewelLogger

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -439,6 +472,46 @@ public open class DefaultMarkdownBlockRenderer(
...
+            AsyncImage(
+                model =
+                    ImageRequest.Builder(LocalPlatformContext.current)
+                        .data(image.source)
+                        // make sure image doesn't get downscaled to the placeholder size
+                        .size(Size.ORIGINAL)
+                        .build(),
+                contentDescription = image.title,
+                onSuccess = { state ->
+                    if (knownSize.value == null) {
```
rock3r, line 505: Is the check needed? I'd imagine whatever we get in onSuccess we only get once anyway...

rock3r:
Out of curiosity, have you filed the issues for SVG rendering?

I think the size one is caused by https://github.com/coil-kt/coil/blob/main/coil-svg/src/commonMain/kotlin/coil3/svg/SvgDecoder.kt#L39 this parameter being true by default. Ideally, users should be able to override the Coil loader with one they want, but we can keep it for a follow up (it'd be nice to have an issue tracking this in our YouTrack, if you can file it).

The text one is the more baffling one. If you can look into it, maybe you can figure out what is the cause, or at least where it should happen, and file an issue for that in the right place.

rock3r:
@obask ping

obask:
> Out of curiosity, have you filed the issues for SVG rendering?

Thanks for doing initial investigation. I filed a tracking issue, but not the one in a coil repository yet.
https://youtrack.jetbrains.com/issue/JEWEL-762/Coil-image-rendering-issues-with-SVG

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
obask: `ProvideMarkdownStyling` is in a different module and it's not guaranteed that a user will wrap Markdown into a style provider, so it may be confusing for debugging.
Will it attempt to invoke this code on each rerender or on each markdown usage in the app? I don't mind moving it somewhere, I just haven't figured out how exactly.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
obask: this can probably look a bit overwhelming for inline images with long URLs, but I added a source too

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
obask: thanks! This is very helpful for debugging and later.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
rock3r: > Will it attempt to invoke this code on each rerender or on each markdown usage in the app?

It won't attempt to invoke, it will invoke it. That's why I'm saying it'd make more sense to do it elsewhere. I think we should even consider not using a singleton factory since users may want to use different ones; can we create one that we store in a composition local, and set it up in ProvideMarkdownStyling?

We can instruct users to use ProvideMarkdownStyling in the error message for when the composition local is not set up, like we do for other mandatory composition locals.

obask: In the message, I assume you're referring to invocation on each markdown usage?

The code itself is checking if an image loader already set before using  compareAndSet  to invoke a factory. I've set a debugger to see what's going on there. I assume one defererence on each markdown frame is not bad and I still don't completely follow the `ProvideMarkdownStyling` approach.

Does this sound reasonable or should I move this call out?
Screenshot 2025-02-18 at 4 13 37 PM

rock3r: > I assume one defererence on each markdown frame is not bad

Yeah, that's ok, thanks for clarifying. But it's not my main concern. What happens if this is used in a context where the user has set their own loader? Seems to me like setting up a singleton image loader sets some expectations on our part that could be broken if the user had already done the same, with a different loader.

It also feels like we're leaking the loader outside Jewel, which is not very good either. We might be causing problems for our users because they expect to be able to set their own singleton IL, and it doesn't do anything if any Markdown composable just happened to be composed before then.

Is there no way to use composition locals to have a Jewel-only image loader that does not leak externally, and that we know we can trust? Of course, users can override it, but it's fine if they do it knowingly. It's the magical, undocumented side effect that is not acceptable.

obask: As discussed offline at one of the syncs simply bumping a version of IDEA upstream dependency on `kotlinx-coroutines-core-jvm-1_10_1` resolve the class not found issue. So we wouldn't this workaround for prod intellij anymore.
Not sure if this is source of truth these days, but everything works starting from version 1.10:
https://github.com/JetBrains/intellij-community/commit/bf01f2c9b7611ac914baaaa43e3f7bc41b92a3dd#diff-bd2bdeac4092ed45e92ce19a14ce674daee91721e05388a7bb8b668422c32aa6L3984-L3998
Screenshot 2025-03-21 at 2 54 23 PM

PS: I still need to test a standalone behavior, waiting for a build...

obask:
ok, so IDEA dev(build from master) and the standalone sample works, but IDE sample uses an older version of the plarform:
```
2025-03-21 15:34:50,150 [ 148214]   WARN - #Jewel - AsyncImage loading: java.lang.NoSuchMethodError: 'kotlinx.coroutines.CoroutineDispatcher kotlinx.coroutines.CoroutineDispatcher.limitedParallelism$default(kotlinx.coroutines.CoroutineDispatcher, int, java.lang.String, int, java.lang.Object)'
```
So if we want to make it work in the sample, we have to add some small workaround there without affecting other modes.

obask:
> ok, so IDEA dev(build from master) and the standalone sample works, but IDE sample uses an older version of the platform:

@rock3r Let me know what do you think is the right thing here given that we don't want to break 24.3 compatibility. The current master is pointing to 2025.2 EAP, so simply waiting may not be sufficient to get this working in older version of IDEA without workarounds. As I mentioned in the thread, I think that previous implementation was ok, as it will only change images rendering in markdown if user specifies their own image loader. But if their loader works correctly, it will probably be able to render markdown images too as we don't really overwrite an image loader provider if it exists.

platform/jewel/gradle/libs.versions.toml
```diff
@@ -17,6 +18,11 @@ kotlinxBinaryCompat = "0.17.0"
 ktfmtGradlePlugin = "0.22.0"
 
 [libraries]
+coil-compose = { module = "io.coil-kt.coil3:coil-compose", version.ref = "coil" }
```
faogustavo: Should this be `coil-compose-core` instead?

https://coil-kt.github.io/coil/getting_started/#configuring-the-singleton-imageloader

image

platform/jewel/markdown/extensions/images/README.md
faogustavo, line 1: I think this one was copied from somewhere else 😅

platform/jewel/markdown/extensions/images/api-dump-unreviewed.txt
faogustavo, line 1: Same for this one

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImagesRendererExtension.kt
```diff
@@ -0,0 +1,13 @@
+package org.jetbrains.jewel.markdown.extensions.images
+
+import org.jetbrains.jewel.markdown.extensions.ImageRendererExtension
+import org.jetbrains.jewel.markdown.extensions.MarkdownRendererExtension
+
+/**
+ * A [MarkdownRendererExtension] that supports rendering
+ * [org.jetbrains.jewel.markdown.InlineMarkdown.CustomDelimitedNode]s into [androidx.compose.ui.text.AnnotatedString]s.
+ */
+public object Coil3ImagesRendererExtension : MarkdownRendererExtension {
```
faogustavo: Please add the image loader instance as argument to use in the `AsyncImage` composable. Ideally we should not depend in the singleton as mentioned above

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
```diff
@@ -421,6 +421,8 @@ private fun MarkdownExample(project: Project) {
                 | * Tables
                 | * And more — I am running out of random things to say 😆
                 |
+                |![logo](https://avatars.githubusercontent.com/u/878437?s=48)
```
AlexVanGogen: I think `Coil3ImagesRendererExtension0` should be passed to `ProvideMarkdownStyling` as part of `blockRenderer`. Something like:
```kotlin
val styling = remember(JewelTheme.name) { MarkdownStyling.create() }
ProvideMarkdownStyling(project,
                       markdownStyling = styling,
                       markdownBlockRenderer = DefaultMarkdownBlockRenderer(styling, listOf(Coil3ImagesRendererExtension))) {
  // Markdown(...)
}
```

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImagesRendererExtension.kt
```diff
@@ -0,0 +1,13 @@
+package org.jetbrains.jewel.markdown.extensions.images
+
+import org.jetbrains.jewel.markdown.extensions.ImageRendererExtension
+import org.jetbrains.jewel.markdown.extensions.MarkdownRendererExtension
+
+/**
+ * A [MarkdownRendererExtension] that supports rendering
+ * [org.jetbrains.jewel.markdown.InlineMarkdown.CustomDelimitedNode]s into [androidx.compose.ui.text.AnnotatedString]s.
+ */
```
AlexVanGogen: Looks more like a description to `MarkdownDelimitedInlineRendererExtension`

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImagesRendererExtensionImpl.kt
```diff
@@ -0,0 +1,70 @@
...
+import coil3.size.Size
+import org.jetbrains.jewel.foundation.util.JewelLogger
+import org.jetbrains.jewel.markdown.InlineMarkdown
+import org.jetbrains.jewel.markdown.extensions.ImageRendererExtension
+import org.jetbrains.jewel.markdown.extensions.MarkdownRendererExtension
+
+/**
+ * A [MarkdownRendererExtension] that supports rendering [InlineMarkdown.CustomDelimitedNode]s into
+ * [androidx.compose.ui.text.AnnotatedString]s.
+ */
```
AlexVanGogen: And here, too

platform/jewel/gradle/libs.versions.toml
obask: Do you have suggestions where it should be initialized? Should we use the same settings on-disk cache as the public one?

obask: PTAL: https://github.com/JetBrains/intellij-community/compare/721728eb0cd5f610ee70552f247d6b22adba01ed..303a7ab0bf10e5b0d93f38e30d763385c0c75875

platform/jewel/markdown/extensions/images/api-dump-unreviewed.txt
obask: I deleted this file for now since I don't know how to regenerate it

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
AlexVanGogen: Nevermind, I'm washed, we can't use it yet as the plugin has to be built against the EAP build (and the module introduced here is obviously not present there)

obask: since IDE-plugin depends on older platform versions, I'm not adding extensions there yet

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/MarkdownRendererExtension.kt
```diff
@@ -24,4 +24,14 @@ public interface MarkdownRendererExtension {
...
     public val delimitedInlineRenderer: MarkdownDelimitedInlineRendererExtension?
         get() = null
+
+    /**
+     * An extension for [`InlineMarkdownRenderer`][org.jetbrains.jewel.markdown.rendering.InlineMarkdownRenderer] that
+     * will render supported [org.jetbrains.jewel.markdown.InlineMarkdown.CustomDelimitedNode]s into an annotated
+     * string.
+     *
+     * Can be null if this extension doesn't support rendering delimited inline nodes.
+     */
```
AlexVanGogen: Isn't that the doc to `delimitedInlineRenderer`?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/ImageRendererExtension.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import androidx.compose.foundation.text.InlineTextContent
+import androidx.compose.runtime.Composable
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.markdown.InlineMarkdown
+
+/** An extension for the Jewel Markdown rendering engine. */
+@ExperimentalJewelApi
+public interface ImageRendererExtension {
+    @Composable public fun imageContent(image: InlineMarkdown.Image): InlineTextContent
```
AlexVanGogen: Suggestion. Can we change it to `renderImage`? (See the comment to `Coil3ImagesRendererExtensionImpl`)

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageLoaderProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+import coil3.ImageLoader
+import coil3.PlatformContext
+
+internal class Coil3ImageLoaderProvider {
+    @Volatile private var instance: ImageLoader? = null
+
+    fun get(context: PlatformContext): ImageLoader {
+        // using the double-checked locking
+        return instance
+            ?: synchronized(this) { instance ?: ImageLoader.Builder(context).build().also { instance = it } }
```
AlexVanGogen: Let's specify the cache size here? As far as I see, it amounts to around 80MB by default (15% of 512MB). For Markdown we probably don't need this much

AlexVanGogen: Another idea is to pass building steps as a lambda to the method `get`. Something like,
```kotlin
fun getOrCreate(context: PlatformContext, doBuild: Builder.() -> Unit): ImageLoader
```

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImagesRendererExtensionImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+ * This implementation uses [AsyncImage] to load and display images. It reserves space for the image with a
+ * [Placeholder] and updates its size upon successful loading to prevent layout shifts.
+ *
+ * @param imageLoader A custom [ImageLoader] for Coil3 image requests.
+ */
+public class Coil3ImagesRendererExtensionImpl(private val imageLoader: ImageLoader) : ImageRendererExtension {
+    @Composable
+    public override fun imageContent(image: InlineMarkdown.Image): InlineTextContent {
+        val knownSize = remember(image.source) { mutableStateOf<ImageSize?>(null) }
+        return InlineTextContent(
```
AlexVanGogen, line 36: Follow-up to the comment to `ImageRendererExtension`. I feel like this part will be the same for any `ImageRendererExtension`; what will be different is the composable `children` (in this case, it's `AsyncImage` from coil3). So maybe creation of `InlineTextContext` could be moved out of implementation to `org.jetbrains.jewel.markdown.rendering.DefaultMarkdownBlockRenderer#renderedImages`

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
AlexVanGogen: Let's maybe remove this line then? A bit sad to keep it there as a showcase but not be able to render it properly :(

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImagesRendererExtensionImpl.kt
obask: `InlineTextContent` depends on `knownSize` to set the real size of images after they are loaded. Are you suggesting to pass around the known size mutable state or a callback to set actual image sizes back to the render code in markdown core?

obask: discussed offline that we'll keep the extension as is for now

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageLoaderProvider.kt
obask: What should be the right size then? I've set it to 20MB assuming that it shouldn't flicker on each key stroke if someone decides to render a photo gallery.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/ImageRendererExtension.kt
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.jewel.markdown.extensions
+
+import androidx.compose.foundation.text.InlineTextContent
+import androidx.compose.runtime.Composable
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.markdown.InlineMarkdown
+
+/** An extension for the Jewel Markdown rendering engine. */
```
AlexVanGogen: `... image rendering engine`?

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImagesRendererExtension.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.jetbrains.jewel.markdown.extensions.MarkdownRendererExtension
+
+/** A [MarkdownRendererExtension] that supports rendering images and allows setting a custom image loader. */
+public class Coil3ImagesRendererExtension(private val imageLoader: ImageLoader) : MarkdownRendererExtension {
+    override val imageRendererExtension: ImageRendererExtension
+        get() = Coil3ImagesRendererExtensionImpl(imageLoader)
+
+    public companion object {
+        @Composable
+        public fun withDefaultLoader(): Coil3ImagesRendererExtension =
```
AlexVanGogen, line 33: Let's add a doc here? With a reference to `Coil3ImageLoaderProvider.get` and/or storage specifics (20MB of memory cache)

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageLoaderProvider.kt
AlexVanGogen: I'd say it okay, even though a gallery should be a rare use-case. The good thing is that the specifics of the cache are not stated in the codebase (so users can have an easier understanding of them)

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageLoaderProvider.kt
faogustavo, line 1: Nit: Does it makes sense to keep this class? We are creating a new instance for every call to `withDefaultLoader`, so the cache will never happen.

obask: Thanks! I agree that the current implementation was a bit confusing, so I created the image loader directly in the getDefault method.

rock3r:
Ready to merge

AlexVanGogen:
[Merged manually](https://github.com/JetBrains/intellij-community/commit/ce6318e5ac2543755317efc74894a2ce488d15e1)

PR #2921, opened 2025-01-21
[JEWEL-742] Extract component showcase module

hamen:
Based on the work done on
* https://github.com/JetBrains/jewel/pull/742
* https://github.com/JetBrains/jewel/pull/754

PR #2926, opened 2025-01-21
[JEWEL-751] Add Explicit API mode to JPS build

rock3r:
Enables Kotlin's Explicit API mode for the Jewel JPS builds (excluding the samples and the ui-test modules), to align with the settings in the Gradle build.

> [!CAUTION]
> This requires #2913 to be merged first

> [!IMPORTANT]
> This same change is needed for other pending PRs that add new modules (e.g., #2921 — FYI @hamen)


hamen (approved):

jakub-senohrabek:
Please fix the tests (like PainterHintTest), now they don't have the public explicitly stated which introduces an error. Also the commit ebdd52ed01fd3b61be2fa7b2d6d48bb351b3cac2  seems to be wrong. There is duplicite gfmTables entry in .idea/modules.xml, it breaks the path to compose compiler plugin, also make sure there are no empty lines at ends of .iml or .xmls

(also make sure you have added those new modules after rebasing if it is necessary)

PR #2927, opened 2025-01-22
[JEWEL-725] Upgrade Compose to 1.8.0-alpha02

rock3r:
This updates both the JPS and the Gradle builds.

Includes bumping Skiko to 0.8.19, and switching from the combined runtime-all dependency to the separate per-platform dependencies.

> [!CAUTION]
> Requires #2913 to be merged first


hamen (approved):

jakub-senohrabek:
Do I understand correctly that we are instead waiting for non broken release of compose?

rock3r:
I just checked why this is red — it's only because the CI was interrupted. It will still be red if you re-run it, because `master` is red. The fixes are in #2941 and #2915, I won't go and fix them again in this branch, too.

rock3r:
After talking to the CMP team, it looks like we might want to wait for beta01, that should be out in a couple weeks.

rock3r:
False alarm, we're ok with alpha02. The potential binary incompatibility affects other CMP targets, not desktop.

rock3r:
Not sure why an explicit merge was needed, since there were no actual conflicts, but at least GitHub won't be complaining about it now

rock3r:
Merged master to trigger the CI again as it had timed out

jakub-senohrabek:
I have this ready to be merged, but the ScrollingSyncrhonizerTest hangs after update to newer compose. Investigating this with Sasha.

PR #2932, opened 2025-01-27
Use correct mise directories for all platforms

TobiX:
This tries to follow mise's logic for all platforms (See
https://github.com/jdx/mise/blob/31bd9ba54cf887f7e2fd4d9e7c18f498181d1e13/src/env.rs#L58-L66),
so mise-installed JDKs are found on Windows.

This also fixes the XDG variable name (`XDG_DATA_DIR` does not exist, it's
called `XDG_DATA_HOME`)


yopox:
Hi! Thank you for noticing the issue and providing a patch.

PR #2933, opened 2025-01-28
[JEWEL-734] Menu appearance does not match the Swing one

hamen:
Please review and merge after https://github.com/JetBrains/intellij-community/pull/2921

JEWEL
image image

SWING
image
image

Reference https://github.com/JetBrains/jewel/issues/734


rock3r:
This PR only seems to address standalone; can you please confirm they already look correct in bridge?

hamen:
> This PR only seems to address standalone; can you please confirm they already look correct in bridge?

I also added the bits for the Bridge now.

rock3r (approved):

hamen:
I went for merge for merge because rebase was driving me crazy. These PRs are too many commits behind, and rebasing forces me to fix the same conflict over and over and over. It's not sustainable. Merge works fine, and considering that we are squashing when we go into master, it should be invisible anyway.

rock3r:
Having a thousand commits per PR surely doesn't help with rebasing :) It should be fine. Hopefully Kuba manages to get the other pending PRs merged so this becomes less of an issue

rock3r:
I squashed the relevant changes for this PR into a single commit, off the top of master, and removed all unrelated changes.

PR #2934, opened 2025-01-31
IDEA-366795: Set context when creating new declaration

andymagee:
When running in Android Studio against a non-physical type (eg, a data-binding or `R` class) that uses a scope enlarger, this inspection is producing a new variable declaration that uses a fully-qualified type name when in fact it should be a shortened name.

Creating the declaration with an appropriate context element will allow the correct form of the type name to be generated.


BartvHelvert:
Hi, thank you for the PR, could you please add a test for such cases?

andymagee:
I took a quick look at creating a test, but I won't have bandwidth to do that. I couldn't find any existing tests that execute the quick fix (as opposed to just checking for highlighting), and even if there's prior art there the setup for this case would be quite complex (requiring multiple extension point implementations to provide appropriate `LightElement`s, mimic the `ResolveScopeEnlarger` in use, etc).

But it's a moot point now, since your recent commit [9e54c001](https://github.com/JetBrains/intellij-community/commit/9e54c0017f928465d4eb582738e3b4fe017ff209) actually made an equivalent change, so this has already been fixed. I'm closing this PR accordingly.

PR #2940, opened 2025-02-10
[IDEA-367321] Only cache successful update requests

jrlogsdon:
Otherwise, failed update requests i.e. intermittent network failures or cancelled requests will be cached making it impossible to successfully retrived update requests.

Change-Id: I06ebbb941d03e3553b9c221fe73fcae3a8d19c2f


trespasserw:
Thanks! I have further simplified the code (there is no need to use `Result` if we're only interested in the happy path).

PR #2944, opened 2025-02-14
Replace empty folder names in imported run configurations to null

shartte:
So after some additional investigation, while "empty folder name" in the *UI* means no folder, the code still uses `null` to signify this.

With this fix, a folder name only containing whitespace is fine (this is also accepted by the UI)
image
image

A folder name that is set to an empty string will be converted to null silently:
image
image

And null or no folderName set at all will also not set a folder:
image
image

image
image


nskvortsov:
Thank you for the fix. Running through the CI pipeline.

nskvortsov:
Thank you for your contribution. It will be available in the next 2025.1 EAP

PR #2948, opened 2025-02-18
[kotlin][debugger] Disable JdiHelperClassLoadingImpl for android debug processes

nikita-nazarov:
In the best case when used in an android debug process, JdiHelperClassLoader impl either returns null or throws an exception. In the worst case scenario it hangs the variables view rendering for a while.

@zuevmaxim


java/debugger/impl/src/com/intellij/debugger/impl/JdiHelperClassLoaderImpl.kt
```diff
@@ -17,6 +17,10 @@ private class JdiHelperClassLoaderImpl : JdiHelperClassLoader {
     vararg additionalClassesToLoad: String,
   ): ClassType? {
     val vmProxy = evaluationContext.virtualMachineProxy
+    if (DexDebugFacility.isDex(vmProxy.virtualMachine)) {
```
zuevmaxim, line 20: Can we introduce a separate method `isApplicable(evaluationContext: EvaluationContextImpl)` for that?

zuevmaxim (approved):

PR #2951, opened 2025-02-20
[kotlin][debugger] Introduce Kotin class metadata helper

nikita-nazarov:
Related issue: https://youtrack.jetbrains.com/issue/IDEA-357870

This pull requests introduces `MetadataDebugHelper` - a helper class that allows fetching Kotlin class metadata in the debugger. 

Also, "evaluatable getters" filtering is reimplemented using the mentioned helper class. Unfortunately the new implementation of filtering is slower than the previous one that relied on bytecode inspections:

Kotlin class metadata implementation vs  Bytecode implementation

However, at the same time the new implementation allows the debugger to filter out getters like:
```
fun getX()
fun isX()
```
Which were previously shown. And show getters like:
```
val x: Int get() = 1
```
Which were previously filtered out.

### DIFF
--- a/java/java-runtime/src/com/intellij/rt/debugger/MetadataDebugHelper.java
+++ b/java/java-runtime/src/com/intellij/rt/debugger/MetadataDebugHelper.java
@@ -7,6 +7,8 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
 public class MetadataDebugHelper {
+  public static final String METADATA_SEPARATOR = "\n";
+
   private static final String METADATA_CLASS_NAME = "kotlin.Metadata";
 
   private static final String KIND = "kind";
@@ -36,8 +38,14 @@ public class MetadataDebugHelper {
    * the debugger side. Any JDWP communication can be very expensive when debugging remotely,
    * especially when debugging Android applications on a phone.
    */
-  @SuppressWarnings("unchecked")
   public static String getDebugMetadataAsJson(Class<?> cls) {
+    StringBuilder sb = new StringBuilder();
+    appendDebugMetadataAsJson(sb, cls);
+    return sb.toString();
+  }
+
+  @SuppressWarnings("unchecked")
+  private static void appendDebugMetadataAsJson(StringBuilder sb, Class<?> cls) {
     try {
       Class<?> metadataClass = Class.forName(METADATA_CLASS_NAME, false, cls.getClassLoader());
       Method kindMethod = metadataClass.getDeclaredMethod(KIND_METHOD_NAME);
@@ -50,7 +58,6 @@ public class MetadataDebugHelper {
 
       Object metadata = cls.getAnnotation((Class<Annotation>)metadataClass);
 
-      StringBuilder sb = new StringBuilder();
       sb.append("{");
 
       int kind = (int)kindMethod.invoke(metadata);
@@ -75,11 +82,26 @@ public class MetadataDebugHelper {
       appendAsJsonValueNoComma(sb, EXTRA_INT, extraInt);
 
       sb.append("}");
-
-      return sb.toString();
-    } catch (Exception ex) {
-      return null;
+    } catch (Exception ignored) {
     }
+  }
+
+  /*
+   * This function is used similarly to `getDebugMetadataAsJson`, when there is a need to
+   * fetch metadata for multiple classes.
+   *
+   * The return value is a concatenation of metadata JSON representations of given classes
+   * separated by MetadataDebugHelper.METADATA_SEPARATOR.
+   */
+  public static String getDebugMetadataListAsJson(Class<?>... classes) {
+    StringBuilder sb = new StringBuilder();
+    for (Class<?> cls : classes) {
+      if (sb.length() != 0) {
+        sb.append(METADATA_SEPARATOR);
+      }
+      appendDebugMetadataAsJson(sb, cls);
+    }
+    return sb.toString();
   }
 
   private static void appendAsJsonValue(StringBuilder sb, String name, Object value) {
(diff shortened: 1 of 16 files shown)


plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinMetadataDebuggerCacheService.kt
```diff
@@ -0,0 +1,168 @@
...
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.kotlin.idea.debugger.base.util.wrapEvaluateException
+import org.jetbrains.kotlin.idea.debugger.base.util.wrapIllegalArgumentException
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
+import kotlin.metadata.jvm.KotlinClassMetadata
+
+@ApiStatus.Internal
+@Service(Service.Level.PROJECT)
+class KotlinMetadataDebuggerCacheService private constructor(project: Project) {
```
nikita-nazarov, line 29: Is it an ok implementation of a cache? Or should I use like something like this instead:
```
  val cache = vmProxy.getOrCreateUserData(METADATA_CACHE_KEY) {
    KotlinMetadataCache(debugProcess)
  }
```

nikita-nazarov:
@zuevmaxim

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/render/KotlinClassRenderer.kt
```diff
@@ -78,6 +87,42 @@ class KotlinClassRenderer : ClassRenderer() {
...
+        context: EvaluationContext,
+        methods: List<Method>
+    ): List<Method>? {
+        val gettersToShow = calculateGettersToShow(methods, context) ?: return null
+        return methods
+            .filter { it.name() in gettersToShow }
+            .distinctBy { it.name() }
+    }
+
+    private fun calculateGettersToShow(methods: List<Method>, context: EvaluationContext): Set<String>? {
```
zuevmaxim, line 100: Could you please use `UsingMetadata` suffix here also?

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinMetadataDebuggerCacheService.kt
zuevmaxim: Caching in `vmProxy` is better

plugins/kotlin/jvm-debugger/core/src/org/jetbrains/kotlin/idea/debugger/core/KotlinMetadataDebuggerCacheService.kt
```diff
@@ -0,0 +1,168 @@
...
+                methodName,
+                args,
+                true
+            )
+            (value as? StringReference)?.value()
+        }
+    }
+}
+
+private fun <T> wrapJsonSyntaxException(block: () -> T): T? {
```
zuevmaxim, line 162: `inline`? Just because we can

java/java-runtime/src/com/intellij/rt/debugger/MetadataDebugHelper.java
```diff
@@ -0,0 +1,157 @@
...
+
+  private static void appendAsJsonValue(StringBuilder sb, String name, Object value) {
+    appendAsJsonValueNoComma(sb, name, value).append(',');
+  }
+
+  private static StringBuilder appendAsJsonValueNoComma(StringBuilder sb, String name, Object value) {
+    return sb.append(toJson(name)).append(':').append(value);
+  }
+
+  private static String toJson(String str) {
```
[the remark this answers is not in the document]
zuevmaxim: Make sure you load `JsonUtils` if you rely on it

zuevmaxim (changes requested):
I'll dig into details on Monday. In general, it looks great! Thank you!

java/java-runtime/src/com/intellij/rt/debugger/MetadataDebugHelper.java
```diff
@@ -0,0 +1,146 @@
...
+   * This function is used similarly to `getDebugMetadataAsJson`, when there is a need to
+   * fetch metadata for multiple classes.
+   *
+   * The return value is a concatenation of metadata JSON representations of given classes
+   * separated by MetadataDebugHelper.METADATA_SEPARATOR.
+   */
+  public static String getDebugMetadataListAsJson(Class<?>... classes) {
+    StringBuilder sb = new StringBuilder();
+    for (Class<?> cls : classes) {
+      String metadataAsJson = getDebugMetadataAsJson(cls);
```
zuevmaxim, line 97: It could be minor, but what do you think about creating an additional private method, where we can pass StringBuilder?
We can reuse it then, instead of creating a new string every time

plugins/kotlin/jvm-debugger/test/testData/stepping/packets/evaluatableGetters.out
```diff
@@ -0,0 +1,2348 @@
...
+        JavaValue[getter] x8: int (evaluatableGetters.kt:11)
+        JavaValue[getter] x7: int (evaluatableGetters.kt:10)
+        JavaValue[getter] x6: int (evaluatableGetters.kt:9)
+        JavaValue[getter] x5: int (evaluatableGetters.kt:8)
+        JavaValue[getter] x4: int (evaluatableGetters.kt:7)
+        JavaValue[getter] x3: int (evaluatableGetters.kt:6)
+        JavaValue[getter] x2: int (evaluatableGetters.kt:5)
+        JavaValue[getter] x1: int (evaluatableGetters.kt:4)
+evaluatableGetters.kt:117
+Packets sent: 108, methods called: 8
```
zuevmaxim, line 59: After the changes, the number of method calls is increased, which is expected. Could you please add a test that will show that there are no additional method calls if the metadata is cached?

zuevmaxim (approved):

PR #2935, opened 2025-02-03
[JEWEL-583] Add KDOC to public API

hamen:
- Default Banner
- Inline Banner
- *Checkbox and *CheckboxRow
- SegmentedControl
- Link
- Menu
- RadioButton
- TabStrip
- *ListComboBox
- Progressbar
- Scrollbars
- Slider
- TextArea
- TextField
- Tooltip

PR #2955, opened 2025-02-21
[JEWEL-774] Introduce new stateless *ListComboBox

hamen:
We decided to deprecate the current API in favor of a more-Compose stateless one.
KDOC will be reworked once https://github.com/JetBrains/intellij-community/pull/2935 is merged.

Reference https://youtrack.jetbrains.com/issue/JEWEL-774/Introduce-new-stateless-ListComboBox

### DIFF
--- a/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ListComboBoxUiTest.kt
+++ b/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ListComboBoxUiTest.kt
@@ -531,19 +531,19 @@ class ListComboBoxUiTest {
 
     @Test
     fun `when editable ListComboBox selectedIndex changes then text field updates`() {
-        var selectedIdx by mutableStateOf(0)
+        var selectedIndex by mutableStateOf(0)
 
         composeRule.setContent {
-            val textState = rememberTextFieldState(comboBoxItems[selectedIdx])
+            val textState = rememberTextFieldState(comboBoxItems[selectedIndex])
 
             // Update text state when selectedIndex changes
-            LaunchedEffect(selectedIdx) { textState.edit { replace(0, length, comboBoxItems[selectedIdx]) } }
+            LaunchedEffect(selectedIndex) { textState.edit { replace(0, length, comboBoxItems[selectedIndex]) } }
 
             IntUiTheme {
                 EditableListComboBox(
                     items = comboBoxItems,
-                    selectedIndex = selectedIdx,
-                    onItemSelected = { index, _ -> selectedIdx = index },
+                    selectedIndex = selectedIndex,
+                    onItemSelected = { index, _ -> selectedIndex = index },
                     textFieldState = textState, // Pass the explicitly managed text state
                     modifier = Modifier.testTag("ComboBox").width(200.dp),
                     itemContent = { item, isSelected, isActive ->
@@ -560,7 +560,7 @@ class ListComboBoxUiTest {
         }
 
         textField.assertTextEquals("Item 1")
-        selectedIdx = 3
+        selectedIndex = 3
         composeRule.waitForIdle()
 
         textField.assertTextEquals("Book")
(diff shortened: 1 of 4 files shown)


rock3r:
[waiting for Ivan to write some tests for this and #2952 before I review]

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -39,6 +42,250 @@ import org.jetbrains.jewel.ui.Outline
 import org.jetbrains.jewel.ui.component.styling.ComboBoxStyle
 import org.jetbrains.jewel.ui.theme.comboBoxStyle
 
+@Composable
+public fun ListComboBox(
+    items: List<String>,
```
lamba92, line 47: I would not even provide the items here. I would go for `itemsCount: Int` and then it is going to be up to the user to get the model in the `itemContent` lambda.

lamba92 (changes requested):
I would change the APIs of  `ListComboBox`, `EditableComboBox` and `ComboBox` to not force the use to use strings as items. A counter would be more idiomatic and result in a simpler API

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -39,6 +42,250 @@ import org.jetbrains.jewel.ui.Outline
...
+    onItemSelected: (Int, String) -> Unit,
+    modifier: Modifier = Modifier,
+    isEnabled: Boolean = true,
+    outline: Outline = Outline.None,
+    maxPopupHeight: Dp = Dp.Unspecified,
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    style: ComboBoxStyle = JewelTheme.comboBoxStyle,
+    textStyle: TextStyle = JewelTheme.defaultTextStyle,
+    onPopupVisibleChange: (visible: Boolean) -> Unit = {},
+    itemContent: @Composable (text: String, isSelected: Boolean, isActive: Boolean) -> Unit,
```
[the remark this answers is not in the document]
rock3r: Agreed on this, when the user controls the items, we don't need this

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
rock3r: I disagree - the whole point of this API layer is to provide a convenient way to show strings. You can go one layer down if you want to control the list more

rock3r: Although, I see where you're coming from... Let's see what Ivan says

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ListComboBoxUiTest.kt
```diff
@@ -491,6 +492,83 @@ class ListComboBoxUiTest {
         composeRule.onNodeWithTag("Book").assertIsSelected()
     }
 
+    @Test
+    fun `when editable ListComboBox text is edited then selectedIndex remains unchanged`() {
+        var selectedIdx = 1
```
rock3r, line 497: Can we spell out `index`? I had to stop and think "what does `idx` stand for?"

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
hamen: We had a private call, and we came to the conclusion that the current API matches with Swing offers, and that's the mail goal at the moment. We can reiterate on the API to make it more generic in the future.

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ListComboBoxUiTest.kt
hamen: Sorry. A bit of brain fog and a bit of “It's a-me AI. Me help!”.

hamen:
> Still missing tests for #2952

Done in https://github.com/rock3r/intellij-community/commit/25d3efef6c1174e6638729d25678e45a97297e4e 👍🏻

rock3r (approved):

PR #2957, opened 2025-02-21
Avoid expensive operations if the module build script classpath is th…

akerimsenol:
…e same object

This results in a performance boost the module build classpaths are already interned in the cases where they are the same.

Also caching some jar related operations becuase a file lookup is being done.


nskvortsov:
LGTM, running the CI job.

PR #2958, opened 2025-02-21
 Avoid compiler argument serialization in KotlinFacetSettingsWorkspac…

akerimsenol:
…eModel.

 This gets rid of using the entity as the underlying data storage for this, and instead updates the entity in the commit time instead.

### DIFF
--- a/plugins/kotlin/base/facet/src/org/jetbrains/kotlin/idea/serialization/KotlinFacetSettingsWorkspaceModel.kt
+++ b/plugins/kotlin/base/facet/src/org/jetbrains/kotlin/idea/serialization/KotlinFacetSettingsWorkspaceModel.kt
@@ -49,29 +49,11 @@ class KotlinFacetSettingsWorkspaceModel(val entity: KotlinSettingsEntity.Builder
         }
 
     private var _compilerArguments: CommonCompilerArguments? = null
-    private var _lastKnownCompilerArguments: String? = null
     override var compilerArguments: CommonCompilerArguments?
-        get() {
-            val currentSerializedArguments = entity.compilerArguments
-            if (_compilerArguments != null && currentSerializedArguments == _lastKnownCompilerArguments) {
-                // Cache is valid, return the cached value
-                return _compilerArguments
-            }
-
-            // Deserialize and update the cache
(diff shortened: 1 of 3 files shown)


CherepanovAleksei:
>hi :wave: I have some concerns if it will work for other build systems. E.g. Amper can modify entity directly without calling Facet. So this change can potentially broke the logic with other build systems

e.g. AbstractKotlinMavenImporterTest - importer is updating entity directly, not via Facet interface, so when you'll try to get some info from it, cached values will be old, not updated
but the good news is that we are moving to the direction, where serializing will not be needed - we will replace it with WSM entities

plugins/kotlin/base/facet/src/org/jetbrains/kotlin/idea/serialization/KotlinFacetSettingsWorkspaceModel.kt
```diff
@@ -48,30 +48,12 @@ class KotlinFacetSettingsWorkspaceModel(val entity: KotlinSettingsEntity.Builder
             _additionalVisibleModuleNames = value
         }
 
-    private var _compilerArguments: CommonCompilerArguments? = null
-    private var _lastKnownCompilerArguments: String? = null
+    private var _compilerArguments: CommonCompilerArguments? = CompilerArgumentsSerializer.deserializeFromString(entity.compilerArguments)
```
akerimsenol, line 51: This was missing from the initial version of the commit, causing the tests to fail.

CherepanovAleksei:
I've taken a closer look at the updated PR, and discovered there were two problems initially that I've counted as a single one. The first one was with the initial state after the first import, and you've fixed it by filling `_compilerArguments`. But the second one is still here:
If the build system doesn't use KotlinFacet as an entry point and operate with an entity in WSM directly, then some plugin code that fully relies on facets will not get updates after the second reimport. e.g. Amper operates like this
So here is the case, when the updated code will fail:
1) import (build system updated WSM entity)
2) resolve in Kotlin plugin reads facet, facet reads entity, result cached
3) configuration changed(e.g. optIns in compiler args), reimport updates entity diractly in WSM, bypassing the KotlinFacetBridge logic
4) resolve still uses old cached values in facet instead of the updated ones

PR #2961, opened 2025-02-24
Remove invalid function 'getargspec' references in hg4idea library

vijay-chandar:
function 'getargspec' has been deprecated in python 3 and has been removed in python 3.11. Since invalid functions are used here, methods like 'hg pull, hg push and hg update' fails from Intellij UI.

Reference Link: https://docs.python.org/3.0/library/inspect.html?highlight=getargspec#inspect.getargspec


dmitry-zhuravlev:
Thank you for your PR. 
We fixed this issue in the similar way but preserved backward compatibility with older versions of Python.

PR #2964, opened 2025-02-24
IJPL-175783 Revert enforcement of AboutDialog EP contract

rock3r:
Turns out that the lack of enforcement and documentation of the contract of this EP is relied upon by some plugins, so we need to avoid enforcing it. The EP KDoc has also been updated to match the actual behaviour.


rock3r:
CC @goto1134

PR #2967, opened 2025-02-26
Improve performance of JrePathEditor.setPathOrName

asilichenko:
Calling the `findOrAddCustomJre` method can be very resource-intensive, so in this pull request I suggest not calling it in the `JrePathEditor.setPathOrName` unless necessary, embedding it inside the `if` statement where its result is used.


yopox (approved):

yopox:
Thank you for the fix!

PR #2974, opened 2025-02-28
[xml] Update schema to 1.2beta to include spaces as pubIdChars

Vampire:
Whitespace characters are **not** normalized by XML parsing,
not within attribute values where this regex is relevant.

And even if the whitespaces would be normalized, that would probably not mean that they are removed completely but multiple replaced by a single space character.

With the current regex, it looks like this as the whitespaces are missing:
image

### DIFF
--- a/xml/xml-psi-impl/resources/standardSchemas/catalog.xsd
+++ b/xml/xml-psi-impl/resources/standardSchemas/catalog.xsd
@@ -9,11 +9,8 @@
   <xs:import namespace="http://www.w3.org/XML/1998/namespace"/>
 
   <xs:simpleType name="pubIdChars">
-    <!-- A string of the characters defined as pubIdChar in production 13
-         of the Second Edition of the XML 1.0 Recommendation. Does not include
-         the whitespace characters because they're normalized by XML parsing. -->
     <xs:restriction base="xs:string">
-      <xs:pattern value="[a-zA-Z0-9\-'\(\)+,./:=?;!*#@$_%]*"/>
+      <xs:pattern value="[a-zA-Z0-9\-'\(\)+,./:=?;!*#@$_% ]*"/>
     </xs:restriction>
   </xs:simpleType>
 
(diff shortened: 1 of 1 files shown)


turansky:
Catalog `1.2beta` [already contains space](https://xmlcatalogs.org/schema/1.2beta/catalog.xsd)

turansky:
@Vampire could you please update catalog to version [`1.2beta`](https://xmlcatalogs.org/schema/1.2beta/catalog.xsd)?
`1.2beta` update contains only bugfix for `pubIdChars ` ;)

Vampire:
1/3 of a fix, CR and LF which according to spec are also valid pubid characters are still missing

Vampire:
But I incorporated the other changes now

turansky:
@Vampire do you have real use cases for `\r` and `\n`? Multiline "ids"?

xml/xml-psi-impl/resources/standardSchemas/catalog.xsd
```diff
@@ -9,11 +9,8 @@
   <xs:import namespace="http://www.w3.org/XML/1998/namespace"/>
 
   <xs:simpleType name="pubIdChars">
-    <!-- A string of the characters defined as pubIdChar in production 13
-         of the Second Edition of the XML 1.0 Recommendation. Does not include
-         the whitespace characters because they're normalized by XML parsing. -->
     <xs:restriction base="xs:string">
-      <xs:pattern value="[a-zA-Z0-9\-'\(\)+,./:=?;!*#@$_%]*"/>
+      <xs:pattern value="[a-zA-Z0-9\-'\(\)+,./:=?;!*#@$_% \r\n]*"/>
```
turansky, line 13: Let's start from clean update to version `1.2beta` (without `\r\n` in pattern)

turansky (changes requested):

Vampire:
I'm sorry, was occupied with other stuff.
> @Vampire do you have real use cases for \r and \n?

I don't see how me having a use-case is actually relevant.
The folks making up the specification obviously had some use-cases in mind and added it.
It is part of the specification, so why should an implementation of the specification not adhere to it? :-/

I actually do not have a concrete use-case currently, but imho better to follow the spec.
And it is also not that it hurts in any way significantly, especially now that they are the last in the pattern.

turansky (approved):

turansky:
@Vampire thank you for contribution.

Changes merged (b80d8c9, 608e515).

PR #2979, opened 2025-03-05
When moving symbol declared in __all__ in path hierarchy, add import

chbndrhnns:
Fixes https://youtrack.jetbrains.com/issue/PY-54168/Moving-class-from-init.py-into-own-module-removes-it-from-all-attribute


chbndrhnns:
Rebased onto 251.
@avokin2 Do you need anything else from me?

avokin2:
@chbndrhnns, thank you, I need some time to check the change.

avokin2:
Merged. Fix will be available starting from 2025.1.2

PR #2981, opened 2025-03-06
Make "extract method" work on multi-line function calls

chbndrhnns:
Fixes https://youtrack.jetbrains.com/issue/PY-53711/Cannot-extract-method-if-arguments-are-on-separate-line

### DIFF
--- a/python/python-psi-impl/src/com/jetbrains/python/refactoring/PyRefactoringUtil.java
+++ b/python/python-psi-impl/src/com/jetbrains/python/refactoring/PyRefactoringUtil.java
@@ -75,6 +75,15 @@ public final class PyRefactoringUtil {
     if (parent == null) {
       return null;
     }
+
+    // Allow function call where arguments are on multiple lines
+    if (parent instanceof PyArgumentList) {
+      PyCallExpression callExpression = PsiTreeUtil.getParentOfType(parent, PyCallExpression.class);
+      if (callExpression != null) {
+        return callExpression;
+      }
+    }
+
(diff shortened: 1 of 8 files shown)


python/python-psi-impl/src/com/jetbrains/python/refactoring/PyRefactoringUtil.java
```diff
@@ -75,6 +75,15 @@ public void visitElement(final @NotNull PsiElement element) {
     if (parent == null) {
       return null;
     }
+
+    // Allow function call where arguments are on multiple lines
+    if (parent instanceof PyArgumentList) {
```
KalininDL, line 80: Hi! Thank you for your PR!
The only small thing I noticed is that if we place the caret at the end of the last argument, see the code below:
   
```
result = long_function_name(
        first_argument="value1",
        second_argument="value2",
        third_argument="value3"<caret>
    )
```
 the both elements end up being the children of the PyKeywordArgument, so the method will be extracted incorrectly. Adding `|| parent instanceof PyKeywordArgument` will fix it I believe.

chbndrhnns: I added the or statement as suggested.

chbndrhnns:
@KalininDL I also updated the logic to select the whole expression that is going to be extracted in the preview.

KalininDL:
@chbndrhnns Thank you very much for the collaboration! Your changes were approved by QA and merged to the master branch.
They will appear here as soon as the commits are synchronized with this GitHub repository.

PR #2982, opened 2025-03-07
Only process maven repository once if it's the same

akerimsenol:
This should improve import performance for large projects

### DIFF
--- a/plugins/gradle/java/src/service/project/MavenRepositoriesProjectResolver.kt
+++ b/plugins/gradle/java/src/service/project/MavenRepositoriesProjectResolver.kt
@@ -15,6 +15,9 @@ import java.util.*
 
 
 class MavenRepositoriesProjectResolver: AbstractProjectResolverExtension() {
+
+  val processedRepositoryModels = IdentityHashMap<RepositoryModels, Unit>()
+
   override fun populateProjectExtraModels(gradleProject: IdeaProject, ideProject: DataNode<ProjectData>) {
     val repositories = resolverCtx.getRootModel(RepositoryModels::class.java)
     addRepositoriesToProject(ideProject, repositories)
@@ -36,7 +39,7 @@ class MavenRepositoriesProjectResolver: AbstractProjectResolverExtension() {
 
   private fun addRepositoriesToProject(ideProject: DataNode<ProjectData>,
                                        repositories: RepositoryModels?) {
-    if (repositories != null) {
+    if (repositories != null && !processedRepositoryModels.containsKey(repositories)) {
       val knownRepositories = ExternalSystemApiUtil.getChildren(ideProject, MavenRepositoryData.KEY)
         .asSequence()
         .map { it.data }
@@ -47,6 +50,7 @@ class MavenRepositoriesProjectResolver: AbstractProjectResolverExtension() {
         .filter { !knownRepositories.contains(it) }
         .distinct()
         .forEach { ideProject.addChild(DataNode(MavenRepositoryData.KEY, it, ideProject)) }
+      processedRepositoryModels[repositories] = Unit
     }
   }
 }


plugins/gradle/java/src/service/project/MavenRepositoriesProjectResolver.kt
```diff
@@ -15,6 +15,9 @@ import java.util.*
 
 
 class MavenRepositoriesProjectResolver: AbstractProjectResolverExtension() {
+
+  val processedRepositoryModels = IdentityHashMap<RepositoryModels, Unit>()
```
nskvortsov, line 19: Does the `IdentityHashMap` work as expected here?

 At first glance, different instances of `RepositoryModels` can contain duplicates - different  instances of `MavenRepositoryModel` that are logically equal, but have different references.

nskvortsov (changes requested):

plugins/gradle/java/src/service/project/MavenRepositoriesProjectResolver.kt
nskvortsov, line 19: Also storing state here will cause memory leak, as this class is a long-living service. It is not re-created on each sync.

nskvortsov, line 19: Does the IdentityHashMap work as expected here?

At first glance, different instances of RepositoryModels can contain duplicates - different instances of MavenRepositoryModel that are logically equal, but have different references.

nskvortsov (changes requested):

plugins/gradle/java/src/service/project/MavenRepositoriesProjectResolver.kt
akerimsenol: Oh, I didn't realize that. Thanks! What would be the alternative? I've seen a pattern where this sort of cache was stored inside DataNode, is that what I should do?

akerimsenol: Clearing the map in an import finished callback should be fine, too, right?

akerimsenol: For reference, I ran the debugger with a sample project with multiple modules logging the object ids and contents each time it runs and here it is showing they are all the same.
[identity-debugger-output.txt](https://github.com/user-attachments/files/19130740/identity-debugger-output.txt)

akerimsenol: I looked into this in detail, and it looks like it's a feature/quirk of the serialization system, specifically caused by ObjectCollector avoiding processing the "same" (i.e. equal) objects multiple times.

akerimsenol: I just updated this and the other change to clear the map on resolve finished, let me know if that's not a very good solution either.

nskvortsov: What if we change to classic hashmap? Will it become more "future proof" ?

nskvortsov: This looks better now. Thank you!

akerimsenol: That should be fine. Initial reason I did it this way was to avoid long hash code and equals operations for large list. We can instead optimize the hashCode/equals implementation for large lists (more of an issue for the classpath change, I guess). I don't imagine having many many repositories, but one can have many entries in a classpath. Let me update both changes that way.

akerimsenol:
While, I'm happy to see them merged, sorry this has gone stale. I can send follow-up changes to eliminate the use of IdentiyHashMap to make it more future proof.

nskvortsov:
I will gladly accept the followup patches in a pull request!

PR #2983, opened 2025-03-07
Only process each buildscript once in JavaGradleProjectResolver

akerimsenol:
This should improve import performance for large projects

### DIFF
--- a/plugins/gradle/java/src/service/project/JavaGradleProjectResolver.java
+++ b/plugins/gradle/java/src/service/project/JavaGradleProjectResolver.java
@@ -50,6 +50,14 @@ import static com.intellij.openapi.externalSystem.service.execution.ExternalSyst
 @Order(ExternalSystemConstants.UNORDERED)
 public final class JavaGradleProjectResolver extends AbstractProjectResolverExtension {
 
+  private final IdentityHashMap<GradleBuildScriptClasspathModel, List<BuildScriptClasspathData.ClasspathEntry>> buildScriptEntriesMap =
+    new IdentityHashMap<>();
+
+  @Override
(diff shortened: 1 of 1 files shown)


plugins/gradle/java/src/service/project/JavaGradleProjectResolver.java
```diff
@@ -50,6 +50,9 @@
 @Order(ExternalSystemConstants.UNORDERED)
 public final class JavaGradleProjectResolver extends AbstractProjectResolverExtension {
 
+  private final IdentityHashMap<GradleBuildScriptClasspathModel, List<BuildScriptClasspathData.ClasspathEntry>> buildScriptEntriesMap =
```
nskvortsov, line 53: I am confused by the explicit usage of IdentityHashMap here. Are the instances of `GradleBuildScriptClasspathModel` internalized?

nskvortsov, line 53: Storing state here will cause memory leak, as this class is a long-living service. It is not re-created on each sync.

nskvortsov (changes requested):

plugins/gradle/java/src/service/project/JavaGradleProjectResolver.java
akerimsenol: Similarly, it was my experience that they were always the same, but I don't fully understand why. I can dig deeper to understand.

nskvortsov: Looks better now, thank you!

PR #2984, opened 2025-03-07
Remove DummyVersionedEntityStorage, causing "cached values" not to be cached

akerimsenol:
The storage field in StorageSnapshotCache is unused anyway and is deleted. As far as I understand, this should solve the memory concerns that caused this class to be introduced.


kornilova203:
Thank you for the PR!
I'm afraid that this will lead to concurrency problems that might be quite difficult to debug
This MR leads to values being calculated on a modifiable storage instead of storage snapshot (the `StorageSnapshotCache.storage` field)

`VersionedCache` in `ModuleRootComponentBridge` also calculates `ModuleEntity` on a modifiable storage but this is just a single lookup and `ModuleEntity` should already exist inside the storage so it's safe

But if the computation of `CachedValue` inside the `VersionedEntityStorageOnBuilder` does something more difficult that a single lookup in the storage then there'll be race if the storage is modified concurrently

So unfortunately I don't think that this is safe to do
Although it would be quite good if we could find the solution and cache values without using too much memory

PR #2985, opened 2025-03-07
Batch import dependencies in data services

akerimsenol:
This results in a performance boost for large projects


nskvortsov:
Please rebase, if the issue is still relevant

PR #2986, opened 2025-03-07
Skip processing implements if no usage in Kotlin project resolver extension

akerimsenol:
This is going to improve performance because it's eliminating a useless graph traversal


Tapchicoma:
@akerimsenol Could you explain a little more why such optimization helps for Kotlin only resolver? And not, for example, in general?

akerimsenol:
It helps the general case, which is why I even discovered this as an issue. These methods below are expensive and they are trying to do something with no effect (i.e. when there are no "implements" dependencies, they do a whole dependency graph traversal for nothing)



```
     addTransitiveDependenciesOnImplementedModules(gradleModule, ideModule, ideProject)
     addImplementedModuleNames(gradleModule, ideModule, ideProject, gradleModel)
```

I'm not at all familiar with it but, from what I understand this "implements" configuration is a legacy KMP concept and is not applicable in most modern scenarios. (Please note the distinction between "implements" and "implementation", which threw me off track initially)

PR #2989, opened 2025-03-13
[JEWEL-695] Fix ScrollableContainer layout, allow aligning content

rock3r:
This fixes the ScrollableContainer to allow it to "fit content" on both the main axis, and the cross axis. This means that it won't try to take as much space as possible even when not set to fillMax*(), but rather it will only use as much space as it needs.

It also makes the scrollable container's content have a BoxScope as its receiver, so that users can align the content when it's smaller than the scrollable container itself (e.g., center it vertically if shorter than the viewport size).

The Scrollbars sample in the showcase has been updated accordingly.

Demo of the new behaviour

> [!NOTE]
> This PR contains some spurious changes, but as usual, that's fixing some cruft from the main branch that wouldn't allow the static analysis to be green otherwise.


hamen (approved):
Very fancy

rock3r:
Much needed

jakub-senohrabek:
Real commit: ee51c75

PR #2990, opened 2025-03-14
[JEWEL] IJPL-175720 Improve scrolling sync so it handles identical markdown blocks correctly

AlexVanGogen:
I finally did it. After somewhat intensive manual testing, I found that it only works well with [Oleg's changes](https://github.com/JetBrains/intellij-community/pull/2922#discussion_r1945005135). It means there's a bug somewhere in my current spans update, but I decided not to look for it and just merge this part after the mentioned PR :)

### DIFF
--- a/platform/jewel/ui/api/ui.api
+++ b/platform/jewel/ui/api/ui.api
@@ -384,6 +384,7 @@ public final class org/jetbrains/jewel/ui/component/DropdownState$Companion {
 
 public final class org/jetbrains/jewel/ui/component/EditableComboBoxKt {
 	public static final fun EditableComboBox (Landroidx/compose/foundation/text/input/TextFieldState;Landroidx/compose/ui/Modifier;Landroidx/compose/ui/Modifier;ZLorg/jetbrains/jewel/ui/Outline;Landroidx/compose/foundation/interaction/MutableInteractionSource;Lorg/jetbrains/jewel/ui/component/styling/ComboBoxStyle;Landroidx/compose/ui/text/TextStyle;Lkotlin/jvm/functions/Function0;Lkotlin/jvm/functions/Function0;Lkotlin/jvm/functions/Function0;Lorg/jetbrains/jewel/ui/component/PopupManager;Lkotlin/jvm/functions/Function2;Landroidx/compose/runtime/Composer;III)V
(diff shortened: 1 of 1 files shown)


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
```diff
@@ -124,15 +129,24 @@ public abstract class ScrollingSynchronizer {
             }
     }
 
+    @GenerateDataFunctions
```
rock3r, line 132: I think we don't use this annotation any more, you need to generate the equals/hashcode

rock3r (approved):
LGTM; would love for Oleg to also take a look

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
```diff
@@ -176,24 +190,86 @@ public abstract class ScrollingSynchronizer {
             // acceptBlockSpans works on ALL the nodes, including those unchanged,
             // so it will be fully rebuilt during processing anyway
             lines2Blocks.clear()
-            blocks2LineRanges.clear()
         }
 
         override fun afterProcessing() {
```
obask, line 195: please add details about this function

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
```diff
@@ -13,8 +13,10 @@ import androidx.compose.ui.layout.positionInRoot
 import androidx.compose.ui.text.TextLayoutResult
 import java.util.TreeMap
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.foundation.GenerateDataFunctions
 import org.jetbrains.jewel.foundation.util.myLogger
 import org.jetbrains.jewel.markdown.MarkdownBlock
+import org.jetbrains.jewel.markdown.WithChildrenMarkdown
 import org.jetbrains.jewel.markdown.processing.MarkdownProcessor
 
 /**
```
obask, line 22: nit: maybe this docstring should be updated to reflect new changes?

obask (approved):
Please check if some parts of these changes could be cleared and add some extra details into docstring and the commit message(even if it will be auto-generated)

PR #2991, opened 2025-03-14
[launcher ]Check CARGO_CFG_TARGET_OS instead of 'target_os

eugen-yakovets:
`#[cfg]` refers to the environment in which the build.rs's executable will run, which will be the host (compiler) operating system. It is not the actual compilation target system.


eugen-yakovets:
@trespasserw

trespasserw:
Merged, thanks!

PR #2992, opened 2025-03-16
[JEWEL-782] Update Gradle to 8.13, and other dependencies

rock3r:
This is the output of running `./gradlew wrapper --gradle-version latest` at the Jewel root (plus, reverting to using the 'all' version to get access to Gradle sources in the IDE).

* Added the distribution type to the root build.gradle.kts file to avoid having to manually tweak the distribution type in the future.
* Replace deprecated exec calls in Gradle with injected ExecOperations
* Bump 243 target to 243.4.1 (243.25659.59)
* Upgrade Kotlin to 2.0.21 stable (was -RC) as per the table at https://plugins.jetbrains.com/docs/intellij/using-kotlin.html
* Upgrade IJP Gradle plugin to 2.3.0, cleanup deprecations


hamen (approved):
Standalone and IDE samples build on Windows as well 🚢

rock3r:
Thanks for checking @hamen <3

Kuba, this is also ready to merge

PR #2993, opened 2025-03-17
[JEWEL] Add modifiers to Markdown block renderer APIs

rock3r:
This allows implementers of custom renderers to pass in a modifier to the default renderer. This is particularly convenient when the customization solely consists in aspects expressible by a modifier.

Also in this, we fix the Markdown code blocks not looking disabled when the rest of the rendered Markdown does (JEWEL-465), and tweaks the default editor text style line height to better match the IDE's.

Screen Recording 2025-03-16 at 19 44 25


rock3r:
Ready to merge

PR #2996, opened 2025-03-19
[JEWEL] Fix log paths in run configs

rock3r:
This commit updates the log paths in run configs to be independent of the target IJP version. This way when you run the IDE sample, the logs will show up in the host IDE regardless of the target IJ version used.


hamen (approved):

rock3r:
Ready for merge

PR #3000, opened 2025-03-24
[PY-77792] Fix `No Conda environment selected` error

tl-hbk:
https://youtrack.jetbrains.com/issue/PY-77792/No-Conda-enviroment-selected

Tried to get the input validation to re-run when the selection changes.  Not familiar with kotlin but this change worked when I tested locally.

When I stepped through with the debugger the issue seemed to be that the input validation would run right after the conda exe was set so the selection was still null but after the `conda env list` was run in the background to populate the selection, changing what's selected didn't dismiss the error.


throwable-one (changes requested):

throwable-one:
merged https://github.com/JetBrains/intellij-community/commit/5d875ea52fa856b5b41139526d513795006cab39

vladimir-koshelev:
Merged

PR #3001, opened 2025-03-25
IDEA-369696 Fix update DaemonJvmCriteria with invalid or undefined ProjectSDK

vmadalin:
### Context
The `Modify Daemon JVM` option end up invoking the `updateDaemonJvm` task to update `gradle-daemon-jvm.properties` file since with `8.13` that file also contain download URLs for the toolchains being resolved based on the applied project [resolver toolchain plugins](https://docs.gradle.org/current/userguide/toolchain_plugins.html#toolchain_plugins), meaning that cannot be modified directly by the IDE and must be delegated to Gradle. To archive running a Gradle task bypassing a potential JVM criteria not matching with any locally installed toolchain the `Project SDK` is used instead. However, in your case the `project-jdk-name` under `.idea/misc.xml`  isn't defined or points to invalid `jdk.table` entry. 

For sure the `Project SDK` might be updated after a successful sync to point to the Daemon used `java.home` to run the sync (this was already raised on [IDEA-367680](https://youtrack.jetbrains.com/issue/IDEA-367680/The-module-SDK-isnt-always-specified) after discussing [#2947](https://github.com/JetBrains/intellij-community/pull/2947) draft PR). However, there's no guarantee that users don't manually modified it or  end up in this case given not having previously synced project (i.e. import project). For this reason we should consider fallback to `bundled JBR` in those cases using `SystemProperties.getJavaHome()`. All of that is possible given that toolchain version used by `updateDaemonJvm` is less important given that will be used just to evaluate project Gradle settings but not actually compile anything.

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleDaemonJvmHelper.kt
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleDaemonJvmHelper.kt
@@ -13,6 +13,7 @@ import com.intellij.openapi.util.Key
 import com.intellij.openapi.util.UserDataHolderBase
 import com.intellij.openapi.util.registry.Registry
 import com.intellij.openapi.util.text.StringUtil
+import com.intellij.util.SystemProperties
 import org.gradle.internal.buildconfiguration.DaemonJvmPropertiesConfigurator
 import org.gradle.util.GradleVersion
 import org.jetbrains.annotations.ApiStatus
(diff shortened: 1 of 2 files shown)


HackerMadCat:
Hi! Thank you for the pull request!

We fixed this issue in the other way. Could you please take a look at the https://youtrack.jetbrains.com/issue/IDEA-376289?
We created the [org.jetbrains.plugins.gradle.service.execution.GradleExecutionHelper#getJavaHomeForOperation`](https://github.com/JetBrains/intellij-community/blob/24dd57fb64fe23ec3fd0d19956c30881943404ac/plugins/gradle/src/org/jetbrains/plugins/gradle/service/execution/GradleExecutionHelper.java#L371) function that searches the disk for the proper JVM for the Gradle daemon. It is used for running Gradle `wrapper` and `updateDaemonJvm` tasks.

Could you please reopen and adjust your pull request if you still face this issue?

PR #3004, opened 2025-03-26
[JEWEL-775, JEWEL-784] Align Gradle build with IJP 251

rock3r:
This updates all dependencies to IJP 251, and cleans up the Gradle build as required by the now-deprecated/changed APIs we used to rely on.

> [!IMPORTANT]
> The contents of this PR **MUST NOT** be backported to 243. Only to the 251 branch.

1. Dokka 2.0.0 has a new Gradle plugin, which needs a few changes (including a property to make it stop nagging on every build)
2. Removed the accessors used in the Studio releases Gradle plugin, as their imports change with every Kotlin and Gradle update.
3. Sdded a workaround for a current issue with the AS releases JSON, where N Studio releases have a null platformVersion, which crashes the plugin's codegen as it's supposed to be non-null. It also now uses the correct visibility for the generated class (internal).
4. Updating the binary compat plugin means the API files get a few changes due to bugfixes in the plugin itself
5. ktfmt's update re-formatted a bunch of build files
6. Added a few missing dependencies for the Sarif task (which tbh we aren't really using right now)
7. Cleaned up the samples' Gradle files
8. Bump Jewel version to 0.28 preemptively
9. Target JDK 21
10. Remove unneeded GitHub Actions (stale issues management, IJP updates pings

Note: Dokka multi-module setup is causing some issues, and it has been split off to a separate issue, JEWEL-789, as it is not a regression afaik


hamen (approved):

rock3r:
I'll rebase to fix conflicts :)

rock3r:
Ready to merge

PR #3005, opened 2025-03-26
Teach GradleAutoImportAware about .gradle/config.properties

csrhodes:
Handle it similarly to other Gradle-related properties files, in case the user is using the #GRADLE_LOCAL_JAVA_HOME macro.


HackerMadCat:
The pull request is merged by commit 1acf675e0edad4df2f1acf09dfc7adcdb366c4e4

PR #3006, opened 2025-03-27
Android studio wrap floaty elephant

csrhodes:
I think these are the minimal changes that allow Android Studio to wrap the ProjectRefresh floating toolbar to disable it for Android Gradle projects (but continue to allow it for non-Android projects), and to display its own UI (an editor notification banner) for out-of-date Android Gradle projects


HackerMadCat:
The pull request functionality is merged by commits:
f052df9114a9f9dbe1ba2c234054917eb7c94dcd
608e447f687ed79ae40b33a3acedbf4e49a6f1d2

PR #3009, opened 2025-03-28
[JEWEL-781] Remove deadlock workaround for 251+

rock3r:
Until IJP 251 there was a bad deadlock bug that would cause the IDE to freeze if an internal dependency of Compose was first initialised in a modal context (e.g., a dialog) — see IJPL-166436.

The issue was fixed in 2025.1 EAP 7 and the workaround is no longer needed since then, so this removes it.

> [!IMPORTANT]
> This requires #3004 to be merged first, and **MUST NOT** be backported to 243


rock3r:
Ready to merge

PR #3011, opened 2025-03-31
[JEWEL-796] [JEWEL-645] Fix wrong registry key for TooltipMetrics delay

hamen:
One of our users noticed an inconsistency with tooltip delays. This PR addresses the first and most immediate to fix one.
With about 3x faster delay, the tooltip feels indeed _snappier_.

Tested in IDE:

20250331-1508-07 3874485-ezgif com-video-to-gif-converter

Tested in Standalone:

20250331-1511-06 6036174-ezgif com-video-to-gif-converter

*UPDATE*

*Some context from the original ticket*
> After looking into this, it is clear that the tooltip delay should be set to the value of the ide.tooltip.initialReshowDelay registry key (can we recompose when that changes? Do we use rememberUpdatedState for this?), which defaults to 500 ms. This is what HelpTooltip does at a basic level. The Bridge uses ide.tooltip.initialDelay, which is the wrong value. The same check needs to be done in the standalone style, aligning on 500 ms.

Reference on YouTrack [JEWEL-796](https://youtrack.jetbrains.com/issue/JEWEL-796)

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeTooltip.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeTooltip.kt
@@ -17,7 +17,7 @@ internal fun readTooltipStyle(): TooltipStyle {
         metrics =
             TooltipMetrics.defaults(
                 contentPadding = JBUI.CurrentTheme.HelpTooltip.smallTextBorderInsets().toPaddingValues(),
-                showDelay = Registry.intValue("ide.tooltip.initialDelay").milliseconds,
+                showDelay = Registry.intValue("ide.tooltip.initialReshowDelay").milliseconds,
                 cornerSize = CornerSize(JBUI.CurrentTheme.Tooltip.CORNER_RADIUS.dp),
             ),
         colors =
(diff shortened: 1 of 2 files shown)


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeTooltip.kt
```diff
@@ -17,7 +17,7 @@ internal fun readTooltipStyle(): TooltipStyle {
         metrics =
             TooltipMetrics.defaults(
                 contentPadding = JBUI.CurrentTheme.HelpTooltip.smallTextBorderInsets().toPaddingValues(),
-                showDelay = Registry.intValue("ide.tooltip.initialDelay").milliseconds,
+                showDelay = Registry.intValue("ide.tooltip.initialReshowDelay").milliseconds,
```
rock3r, line 20: I'm sure the previous key was correct — I spent quite a lot of time investigating which was the correct key at the time. Can you explain why you needed to change this? And why is this the correct value?

See usages in `com.intellij.ide.IdeTooltipManager` and `com.intellij.ide.HelpTooltip`

rock3r (approved):
Forgot to click the yolo button :)

rock3r:
Ready to merge

PR #3014, opened 2025-04-02
[JEWEL-636] Remove context receivers from KeyActions

hamen:
`KeyActions` was triggering a warning during build time:

> Experimental context receivers are deprecated and will be superseded by context parameters.
Please don't use context receivers. You can either pass parameters explicitly or use members with extensions.

I investigated it a bit, and it looks like we are luckily not using Context Receivers heavily. We should be OK with a couple of passed-in parameters and a couple of `with(...) {...}`.

Tested building the Standalone sample using Gradle and building the whole IJC with JPS.

### DIFF
--- a/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/BasicLazyTree.kt
+++ b/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/BasicLazyTree.kt
@@ -184,7 +184,6 @@ public fun <T> BasicLazyTree(
                         }
                         .clickable(interactionSource = remember { MutableInteractionSource() }, indication = null) {
                             val item = flattenedTree[index] as Tree.Element<T>
-                            JewelLogger.getInstance("BasicLazyTree").warn("Clicky clicky: ${item.data}")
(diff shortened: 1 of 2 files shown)


platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/BasicLazyTree.kt
```diff
@@ -184,7 +184,6 @@ public fun <T> BasicLazyTree(
                         }
                         .clickable(interactionSource = remember { MutableInteractionSource() }, indication = null) {
                             val item = flattenedTree[index] as Tree.Element<T>
-                            JewelLogger.getInstance("BasicLazyTree").warn("Clicky clicky: ${item.data}")
```
hamen, line 187: I apologize for this OT change, but I saw the log and I couldn't in good conscience leave it in 🙏🏻

rock3r: Yeah, definitely needed to be removed.

rock3r (approved):
ready to merge, thanks!

PR #3016, opened 2025-04-03
[JEWEL-766] More work on Tabs a11y and keyboard navigation

hamen:
This PR is a partial work on different tasks on Tabs. It introduces:

* Navigation through tabs using Left and Right keys

### DIFF
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
@@ -18,12 +18,21 @@ import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.selection.selectableGroup
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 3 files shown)


rock3r:
Not sure why the CI is red on this PR, I checked locally and the API dump is correct.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
```diff
@@ -77,8 +117,13 @@ public fun TabStrip(tabs: List<TabData>, style: TabStyle, modifier: Modifier = M
                         interactionSource = remember { MutableInteractionSource() },
                     )
                     .selectableGroup()
+                    .semantics { collectionInfo = CollectionInfo(rowCount = 1, columnCount = tabs.size) }
         ) {
-            tabs.forEach { tabData -> TabImpl(isActive = tabStripState.isActive, tabData = tabData, tabStyle = style) }
+            tabs.forEachIndexed { index, tab ->
+                TabImpl(isActive = tabStripState.isActive, tabData = tab, tabStyle = style, tabIndex = index)
+
+                LaunchedEffect(tabs) { selectedIndex = tabs.indexOfFirst { it.selected } }
```
rock3r, line 125: This being in the for loop is very sus. Are you sure this is correct?

PR #3017, opened 2025-04-07
[IDEA-370503] Disable start page promotions

jrlogsdon:
This is preferred for non-JetBrains IDE (i.e. Android Studio).

Change-Id: I9f726b35ca458b9d43ef6b96848ff508eb914a38


jreznot:
In fact we should have implemented it as a private module which is not added to the platform itself so any non-JetBrains builds would not have it.

But what is wrong with `OnboardingLessonPromoter` ?

jreznot:
I mean all kind of experiments should already be disabled in Android Studio, no A/B experiments expected on your side.

jrlogsdon:
Yes you are right `OnboardingLessonPromoter` is not an issue since this module is not bundled in Android Studio.

Initially, I was just looking at all code adding these promotions that didn't have checks to make sure it isn't Android Studio.

I'll update the PR.

jreznot:
Thank you, we will cherry-pick it right away and think how to isolate the code for the future

PR #3019, opened 2025-04-11
[JEWEL-558] Add support to contrast scrollbars

hamen:
The IDE has an `“Appearance & Behaviour → Appearance → contrast scrollbars”` option that boosts the opacity of scrollbars.

This is implemented in `com.intellij.ui.components.ScrollBarPainter#getColor()` and in `updateTransparency()`.

IDE uses
```kotlin
  private static final List<ColorKey> CONTRAST_ELEMENTS_KEYS = Arrays.asList(
    THUMB_OPAQUE_FOREGROUND,
    THUMB_OPAQUE_BACKGROUND,
    THUMB_OPAQUE_HOVERED_FOREGROUND,
    THUMB_OPAQUE_HOVERED_BACKGROUND,
    THUMB_FOREGROUND,
    THUMB_BACKGROUND,
    THUMB_HOVERED_FOREGROUND,
    THUMB_HOVERED_BACKGROUND
  );
```
to identify which colour to alter with the boosted contrast

The colour alpha is set to the value of the registry key `contrast.scrollbars.alpha.level` or `LIGHT_ALPHA` or `DARK_ALPHA` as fallback.

`SwingBridgeService` now observers the changes in `UISettings` and triggers a theme reload, similarly to when LaF changes.

Recording2025-04-15180848-ezgif com-video-to-gif-converter

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeService.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeService.kt
@@ -4,8 +4,13 @@ import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.unit.sp
 import com.intellij.ide.ui.LafFlowService
+import com.intellij.ide.ui.UISettings
+import com.intellij.ide.ui.UISettingsListener
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.util.messages.impl.subscribeAsFlow
 import kotlin.time.Duration.Companion.milliseconds
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.combine
 import org.jetbrains.jewel.bridge.theme.createBridgeComponentStyling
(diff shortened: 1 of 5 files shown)


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeService.kt
```diff
@@ -15,15 +22,31 @@ import org.jetbrains.jewel.ui.ComponentStyling
 import org.jetbrains.jewel.ui.component.copyWithSize
 
 @Suppress("UnstableApiUsage")
-internal class SwingBridgeService {
+internal class SwingBridgeService : Disposable {
+    // TODO Pick the proper scope
+    private val coroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
hamen, line 27: @rock3r I need your input on this with all your knowledge about running coroutines in IJP 🙏

rock3r: As per the [IJP docs](https://plugins.jetbrains.com/docs/intellij/coroutine-scopes.html#using-a-correct-scope), this is the easiest way:

```kotlin
internal class SwingBridgeService(scope: CoroutineScope) {
```

No need for `Disposable`, either; the platform should take care of it.

rock3r:
Ready to merge

rock3r:
@hamen can you solve the conflict?

PR #3020, opened 2025-04-11
Validate `PermanentUserId` is not blank

fkorotkov:
We've noticed that two users of our plugin on Windows have blank installation id. It seems this is the only way it could happen so adding an extra validation.


trespasserw:
I have updated the class to add more thorough validation; the commit should reach GH soon.

PR #3025, opened 2025-04-15
Skip sync contributors in two scenarios

akerimsenol:
1. When buildSrc is resolved separately, i.e. in Gradle 8.0
2. When the import is a partial sync


HackerMadCat:
Could you please explain why we need to skip sync contributors in case of a secondary call for the buildSrc model fetching?

PR #3028, opened 2025-04-19
IDEA-328157 Allow gradle installations to be read-only

JFronny:
Currently, the Gradle tool window permits any directory containing the correct content to be configured as the Gradle installation home:
https://github.com/JetBrains/intellij-community/blob/0e71313b43b3880db8a7334ab9f29cb8db4864ed/plugins/gradle/src/org/jetbrains/plugins/gradle/service/settings/IdeaGradleProjectSettingsControlBuilder.java#L419-L421

The new project wizard and settings component on the other hand [use `CHECK_DIRECTORY`](https://github.com/JetBrains/intellij-community/blob/0e71313b43b3880db8a7334ab9f29cb8db4864ed/plugins/gradle/java/src/service/project/wizard/GradleNewProjectWizardStep.kt#L196-L199) in addition, which also checks whether the directory is writable, even though this is never needed here:
https://github.com/JetBrains/intellij-community/blob/0e71313b43b3880db8a7334ab9f29cb8db4864ed/platform/platform-impl/src/com/intellij/openapi/ui/validation/validations.kt#L39-L46

As a consequence, users that use a package manager first have to create their projects with gradlew, then manually delete the gradlew files and change the installation home in the tool window instead of just creating their project.

This PR fixes this by providing a variant of `CHECK_DIRECTORY` that does not check for writeability.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/ui/validation/validations.kt
+++ b/platform/platform-impl/src/com/intellij/openapi/ui/validation/validations.kt
@@ -45,6 +45,14 @@ val CHECK_DIRECTORY: DialogValidation.WithParameter<() -> String> = validationPa
   }
 }
 
+val CHECK_SYSTEM_DIRECTORY: DialogValidation.WithParameter<() -> String> = validationPathErrorFor { path ->
+  when {
+    !path.exists() -> null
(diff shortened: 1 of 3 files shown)


nskvortsov:
I've slightly adjusted the validator and is running the CI now

nskvortsov:
Merged, thank you for your contribution!

PR #3029, opened 2025-04-21
[JEWEL-797] Tooltip automatic disappearance improvements

hamen:
# Prologue

> In IJP, tooltips also disappear automatically after a duration from the registry
The default duration is 10s or 30s, depending on whether the text content is single- or multi-line (defined not by the actual layout but by counting if there are any tags)
You can also set tooltips to never hide explicitly in their builder.
Tooltips attached to help buttons (the round ones with the ? symbol) also never hide by default.
We should separately track this behaviour and implement it at some point.

Reference https://youtrack.jetbrains.com/issue/JEWEL-645/Tooltip-default-delay-is-too-long

# Overview

This PR introduces two new values to the `TooltipMetrics`:

* `regularDisappearDelay` based on the Registry key `ide.helptooltip.regular.dismissDelay`
* `fullDisappearDelay` based on the Registry key `ide.helptooltip.full.dismissDelay`

These new values can be used with the new `autoHideDelay` parameter on `Tooltip`.

We also have a new `neverHide` parameter.

### DIFF
--- a/platform/jewel/ui/api/ui.api
+++ b/platform/jewel/ui/api/ui.api
@@ -105,6 +105,15 @@ public final class org/jetbrains/jewel/ui/component/ActionButtonKt {
 	public static final fun ActionButton (Lkotlin/jvm/functions/Function0;Lkotlin/jvm/functions/Function2;Landroidx/compose/ui/Modifier;ZZLorg/jetbrains/jewel/ui/component/styling/IconButtonStyle;Landroidx/compose/foundation/layout/PaddingValues;Lorg/jetbrains/jewel/ui/component/styling/TooltipStyle;Landroidx/compose/foundation/TooltipPlacement;Landroidx/compose/foundation/interaction/MutableInteractionSource;Lkotlin/jvm/functions/Function2;Landroidx/compose/runtime/Composer;III)V
 }
 
+public final class org/jetbrains/jewel/ui/component/AutoHideBehavior : java/lang/Enum {
+	public static final field Long Lorg/jetbrains/jewel/ui/component/AutoHideBehavior;
+	public static final field Never Lorg/jetbrains/jewel/ui/component/AutoHideBehavior;
+	public static final field Normal Lorg/jetbrains/jewel/ui/component/AutoHideBehavior;
+	public static fun getEntries ()Lkotlin/enums/EnumEntries;
+	public static fun valueOf (Ljava/lang/String;)Lorg/jetbrains/jewel/ui/component/AutoHideBehavior;
+	public static fun values ()[Lorg/jetbrains/jewel/ui/component/AutoHideBehavior;
+}
+
 public final class org/jetbrains/jewel/ui/component/BannerKt {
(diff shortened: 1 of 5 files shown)


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tooltip.kt
```diff
@@ -60,10 +72,32 @@ public fun Tooltip(
     enabled: Boolean = true,
     style: TooltipStyle = JewelTheme.tooltipStyle,
     tooltipPlacement: TooltipPlacement = style.metrics.placement,
+    neverHide: Boolean = true,
+    autoHideDelay: Duration = style.metrics.regularDisappearDelay,
```
rock3r, line 76: IMO it would make more sense to have an `AutoHideBehavior` enum here, which has three values: `Never` (default), `Normal`, `Long`.

platform/jewel/ui/api/ui.api
```diff
@@ -1095,7 +1096,7 @@ public final class org/jetbrains/jewel/ui/component/ToggleableIconButtonState$Co
 }
 
 public final class org/jetbrains/jewel/ui/component/TooltipKt {
-	public static final fun Tooltip (Lkotlin/jvm/functions/Function2;Landroidx/compose/ui/Modifier;ZLorg/jetbrains/jewel/ui/component/styling/TooltipStyle;Landroidx/compose/foundation/TooltipPlacement;Lkotlin/jvm/functions/Function2;Landroidx/compose/runtime/Composer;II)V
+	public static final fun Tooltip-k8zXyTc (Lkotlin/jvm/functions/Function2;Landroidx/compose/ui/Modifier;ZLorg/jetbrains/jewel/ui/component/styling/TooltipStyle;Landroidx/compose/foundation/TooltipPlacement;ZJLkotlin/jvm/functions/Function2;Landroidx/compose/runtime/Composer;II)V
```
rock3r, line 1099: Now that we are in IJP stable, we can't make binary-incompatible changes. You need to add new overloads whenever you add, change, or remove parameters, properties, etc.

This goes for both this and the changes to the styling.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/TooltipStyling.kt
```diff
@@ -127,18 +133,32 @@ public class TooltipMetrics(
             "borderWidth=$borderWidth, " +
             "shadowSize=$shadowSize, " +
             "placement=$placement" +
+            "regularDisappearDelay=$regularDisappearDelay, " +
+            "fullDisappearDelay=$fullDisappearDelay" +
             ")"
     }
 
     public companion object {
         public fun defaults(
```
rock3r, line 142: I'm not really sure why this was put here back in the day — it should be in the standalone module!

rock3r (approved):

rock3r:
Ready for merge

rock3r:
Rebased to re-trigger CI

PR #3030, opened 2025-04-22
Remove custom code for default Project View pane

srmurguia:
... in Android Studio.

There is no need to do some custom checks for Android Studio when deciding the default pane ID for the project view now that isDefaultPane is available. Android Studio now checks for studio.projectview property in AndroidProjectViewPane#isDefaultPane.


gharrma:
Friendly ping, this would help us remove a patch that we have on the Android Studio side at Google. Thanks!

stachenov:
Oh, thank you for the ping. I'll have a look tomorrow. It's easy to forget about GitHub PRs at work because GitHub accounts are private and not connected to the corporate e-mail.

stachenov (approved):

PR #3032, opened 2025-04-25
Fix py 54482: Show True/False/None in decorator argument list completions

chbndrhnns:
Fixes https://youtrack.jetbrains.com/issue/PY-54482/Booleans-missing-in-auto-complete-in-pytest-parametrization


chbndrhnns:
@BartvHelvert This has been untouched since three months. May I ask for an update?

EliseevEgor:
I merged your changes to the main branch, thank you for contributing!

PR #3033, opened 2025-04-25
[xml] Fix pubIdChars

Vampire:
This PR adds the missing line terminator characters that are according to the
specification valid pub id characters but are missing from the schema definition.


turansky:
For now I don't see real use cases for this change.
Without them it's preferrable to stay alligned with [`1.2beta`](https://www.xmlcatalogs.org/schema/1.2beta/catalog.xsd).

PR can be reopened if real use cases will be provided or original `catalog.xsd` will include described changes (preferrable option).

PR #3034, opened 2025-04-26
[JEWEL-809] Fix typo in macOS decorated window property

rock3r:
In the decorated window module, on macOS, we are setting the `apple.awt.newFullScreeControls` system property (note the missing 'n'). It used to be spelled this way in IJP.

However, it was fixed in IJP in early 2024 to
`apple.awt.newFullScreenControls`. This PR updates the system property, fixing the window controls in fullscreen.

This also updates `ui.api`, as it seems to be out of date.

#### Before
Screenshot showing "compat" fullscreen controls on macOS

#### After
Screenshot showing the "compat" controls are gone after the fix

### DIFF
--- a/platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.MacOS.kt
+++ b/platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.MacOS.kt
@@ -65,15 +65,15 @@ internal fun DecoratedWindowScope.TitleBarOnMacOs(
         }
 
     if (newFullscreenControls) {
-        System.setProperty("apple.awt.newFullScreeControls", true.toString())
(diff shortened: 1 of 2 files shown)


rock3r:
@alexander-lobas can you please double-check we're using the system properties correctly? As I understand it, the logic is taken from `MacFullScreenControlsManager`, but there is no documentation and I'd like to be sure :) Thanks!

rock3r:
Ready to merge

alexander-lobas:
> @alexander-lobas can you please double-check we're using the system properties correctly? As I understand it, the logic is taken from `MacFullScreenControlsManager`, but there is no documentation and I'd like to be sure :) Thanks!

Yes, `apple.awt.newFullScree**n**Controls` is correct.

alexander-lobas (approved):

PR #3036, opened 2025-04-30
K2 UAST: make sure LC is indeed constructor

jsjeon:
After KT-72722, source PSI of data class copy is the primary constructor, not the enclosing class. Its corresponding SLC method should be converted to a normal UMethod, not KotlinConstructorUMethod.

^KTIJ-33999 fixed


jsjeon:
cc @BartvHelvert @mglukhikh

BartvHelvert:
Thanks, merged.

PR #3037, opened 2025-04-30
[JEWEL-811] Add Jewel release notes

rock3r:
Since the Jewel repo is now archived, the release notes for versions released from the intellij-community repo will be collected in this new file.


rock3r:
Ready to merge as-is, it's just docs (we will probably need to change the date once the 0.28 release actually happens)

PR #3039, opened 2025-05-01
Markdown Preview enhancements and bug fixes

kshetline:
_As discussed with [Yuriy Artamonov](https://platform.jetbrains.com/u/yuriy.artamonov) in the [IntelliJ Platform](https://platform.jetbrains.com/t/poststartupactivity-is-working-in-development-environment-but-not-after-deployment/1515/7) forum._

Enhancements:

- GitHub style options for a very close match to how Markdown will appear on GitHub. Included are both light and dark themes which can be optionally light or dark independently from the user's current IDE theme. _(These changes go beyond what is possible simply by using Custom CSS.)_
- Optional GitHub-like syntax coloring inside code fences.
- Greatly improved synchronized scrolling.
- Ability to double-click inside Preview to scroll the Markdown source to a matching position.

Bug fixes:

- Using **Tools/Markdown/Export Markdown File To...** can crash and hang IDE if a user chooses to overwrite an existing file. A fix for this is included.
- Fixed attempted dereferences of undefined values in `processLinks.js`.

Known issue:

- Even without my changes, there already existed a bug where sporadically the Markdown Preview comes up blank and must be toggled off and on to refresh it. I hoped to find a solution for this, but couldn't recreate the issue reliably enough to figure out the cause.

image

image

I've written and published [an IDEA plugin](https://plugins.jetbrains.com/plugin/14736-ligatures-limited) before, but this is my first try at contributing to the IDEA project itself.

I suspect that you'll have some feedback for me first for suggested code changes before this can be accepted. I've included a copy of the markdown.jar file with my code changes.

[markdown.jar.zip](https://github.com/user-attachments/files/19998356/markdown.jar.zip)


plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt
```diff
@@ -0,0 +1,279 @@
...
+
+      started = true
+
+      var latestJS: String? = null
+      var fallbackJS: String? = null
+      val scriptLatch = CountDownLatch(2)
+
+      Thread {
+        try {
+          latestJS = URL("https://unpkg.com/@highlightjs/cdn-assets/highlight.min.js").readText()
```
kshetline, line 68: I'd definitely understand if you would prefer to always use a fixed version of highlight.js instead of downloading the latest version like this.

plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt
```diff
@@ -0,0 +1,279 @@
...
+    @Volatile private var started = false
+    @Volatile private var startupCompleted = false
+    @Volatile private var available = true
+    @Volatile private var startupLatch: CountDownLatch? = CountDownLatch(1)
+    @Volatile private var jsExecLatch: CountDownLatch? = null
+    @Volatile private var jsResultLatch: CountDownLatch? = null
+    @Volatile private var jsResult: String? = null
+    @Volatile private var jsError = false
+
+    private val browser = object : JBCefBrowser() {
```
jreznot, line 36: Please never use global state and static values. You need application service at least and better project service.

plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt
```diff
@@ -0,0 +1,279 @@
...
+      Thread {
+        try {
+          fallbackJS = ::CodeFenceLanguageParsingSupport.javaClass
+            .getResourceAsStream("highlighter.js")?.bufferedReader().use { it?.readText() }
+        }
+        catch (_: Exception) {}
+        scriptLatch.countDown()
+      }.start()
+
+      scriptLatch.await(5, TimeUnit.SECONDS)
```
jreznot, line 83: Hardcoded timeouts do not look good

plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    private data class TagInfo(val line: Int, val offset: Int)
+
+    private fun addSourceRange(tag: String, start: Int, end: Int): String {
+      return tag.replace(Regex("^(<\\w+)")) { match -> "${match.value} md-src-pos=\"$start..$end\"" }
+    }
+
+    private fun convertToRangedSpans(html: String, startOffset: Int): String {
+      val chunks = html.split(Regex("""(?<=(>|\r\n|\r|\n))|(?=(<|\r\n|\r|\n))""")).toTypedArray()
```
jreznot, line 154: Compiling regex is an expensive operation

kshetline: Doesn't the compiler optimize this so the regexes only get compiled once? Or is it necessary to define these regexes separately somewhere else so to insure the regexes don't get recompiled each time they're used?

plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt
kshetline: This is just a safety measure so, if something unexpected happens while executing the highlight.js code, execution doesn't get stuck. Anything different you'd recommend?

plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt
kshetline: As an experiment, I rewrote `escapeForJs()` like this, benchmarking 1000 repetitions of calling the function with assorted strings:

```kotlin
    private val basicEscapes = Regex("""([\\'"\n\r])""")
    private val controlEscapes = Regex("""([\x00-\x1F])""")

    private fun escapeForJs(text: String): String {
      return text.replace(basicEscapes) { match -> escapees[match.groupValues[1][0]].orEmpty() }
        .replace(controlEscapes) { match -> "\\x" + match.groupValues[1][0].code.toString(16).padStart(2, '0') }
    }
 ```
 
This did not perform any faster than the original version, indicating the Regexes are likely only compiled once, not repeatedly for each call.

Comment (134130) [reply plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt:154]: AFAIK, C# performs optimization, but the JVM does not.

Could I please see the documentation that explains this difference? I can't find it; stackoverflow only has outdated material.

File: plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt (hunk 2 above)
Comment (kshetline) [reply plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt]: Done. Now it's a proper application service. (I decided against a project service instead because it seemed like there would be nothing gained by initializing extra copies of the service for each opened project.)

File: plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt (hunk 3 above)
Comment (kshetline) [reply plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt]: I've made constants for all of these.

File: plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt (hunk 4 above)
Comment (kshetline) [reply plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt]: Even though I didn't see a measurable speed difference, you're definitely right that these Regexes would get recompiled each time, so I've fixed that problem. I've gotten spoiled by JavaScript regex literals.

Comment (134130) [reply plugins/markdown/core/src/org/intellij/plugins/markdown/ui/preview/jcef/CodeFenceLanguageParsingSupport.kt]: Simple iterations are good for the compiler to optimize. In this particular case, 1000 times is too few parameters. At least 100,000 iterations are recommended.

**Comment by kshetline** [comment]
Updated .jar with suggested code changes.

[markdown.jar.zip](https://github.com/user-attachments/files/20020587/markdown.jar.zip)

File: plugins/markdown/compose/src/main/kotlin/com/intellij/markdown/compose/preview/MarkdownComposePanel.kt [hunk 5]
```diff
@@ -190,10 +189,6 @@ internal class MarkdownComposePanel(
   override fun removeScrollListener(listener: MarkdownHtmlPanel.ScrollListener) {
   }
 
-  override suspend fun scrollTo(editor: Editor, line: Int) {
```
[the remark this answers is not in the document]
kshetline: The logic that I'm using for scroll sync is this:

- If the caret position moves, make sure that the corresponding place in the preview is scrolled into view. That's enough to make sure that where you're editing your Markdown source will be visible in the preview panel. The previous `scrollTo` method I deleted above simply wasn't doing a very good job of that.
- If the edit panel is scrolled up, make sure that whatever is at the top of the edit panel is in view at the top of the preview.
- If the edit panel is scrolled down, make sure that whatever is at the bottom of the edit panel is in view at the bottom of the preview.

As for `ensureMarkdownSrcOffsetIsVisible` being non-`suspend`... well, `suspend` is a Kotlin thing, and `ensureMarkdownSrcOffsetIsVisible` is running in the browser in JavaScript, so two very different threading/scheduling models are involved anyway.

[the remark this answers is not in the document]
kshetline: I didn't realize there were two implementations of Markdown Preview. How do I access the other one that I haven't accounted for?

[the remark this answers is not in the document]
kshetline: I've looked at the original code again and the only use of this `scrollTo` method that previously existed traces back to `visibleAreaChanged` in `MarkdownEditorWithPreview`. That one usage isn't needed any more because of the updates I made to `visibleAreaChanged`.

If I'm missing something, please let me know. I'm fairly confident that there's no other implementation to worry about. I made these changes precisely because syncing by line offsets didn't work very well.

Maybe I need to worry if some other plugin developer has created code that expects this interface to exist?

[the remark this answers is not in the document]
kshetline: > registry key markdown.experimental.use.compose.for.preview set to true

Do you mean the Windows registry? If so, what's the macOS equivalent? If not, what other registry do you mean?

At any rate, now I understand why I hadn't run into this issue before!

> > I made these changes precisely because syncing by line offsets didn't work very well.
>
> Could you please elaborate why?

The scrolling of the Markdown editor pane and the preview pane were frequently not well lined up after the line-oriented `scrollTo` method was called in an attempt to sync scrolling. If you look at the README in [this project of mine](https://github.com/kshetline/tubular_time) using IDEA's Markdown Preview you'll see how bad the attempt at syncing would often be.

PR #3043, opened 2025-05-05
KT UAST: fix UField modeling of annotation that belongs to $annotations

jsjeon:
Only `FIELD` or `PROPERTY_DELEGATE_FIELD` use-site targets are applicable
to fields (except for `@JvmField`).

^KTIJ-33663 fixed


jsjeon:
cc @BartvHelvert

BartvHelvert (changes requested):
Some tests start failing, could you take a look at the following tests:
* KotlinUastNonVisitorConversionsTest
* SimpleKotlinRenderLogTest
* KotlinIDERenderLogTest
* K1MissingDeprecatedAnnotationOnScheduledForRemovalApiInspectionTest
* K1MustAlreadyBeRemovedApiInspectionTest
* K1NonNlsUastUtilTest.testNonNlsStringLiterals
* K1SourceToSinkFlowInspectionTest

jsjeon:
> * KotlinUastNonVisitorConversionsTest
> * SimpleKotlinRenderLogTest
> * KotlinIDERenderLogTest

Somehow, many of K1 UAST tests are failing in my local repo... so these are my best guess-work. Let me know if things are still broken.

> * K1MissingDeprecatedAnnotationOnScheduledForRemovalApiInspectionTest
> * K1MustAlreadyBeRemovedApiInspectionTest
> * K1NonNlsUastUtilTest.testNonNlsStringLiterals
> * K1SourceToSinkFlowInspectionTest

For these ones, I rather updated test inputs to properly use `@field:` annotation use-site.

BartvHelvert:
>  I rather updated test inputs to properly use @field: annotation use-site.

Why? I think for these inspections it should work with specifying the use-site target?

jsjeon:
> > I rather updated test inputs to properly use @field: annotation use-site.
> 
> Why? I think for these inspections it should work without specifying the use-site target?

I think they incorrectly depended on annotations on field. For example, `K1MissingDeprecatedAnnotationOnScheduledForRemovalApiInspectionTest` whose inspection is `MissingDeprecatedAnnotationOnScheduledForRemovalApiInspection`, it visited `UClass`/`UMethod`/`UField` and see if a certain annotation is there: https://github.com/JetBrains/intellij-community/blob/master/jvm/jvm-analysis-impl/src/com/intellij/codeInspection/MissingDeprecatedAnnotationOnScheduledForRemovalApiInspection.kt#L38-L51

For field test, here are Java / Kotlin input:
* https://github.com/JetBrains/intellij-community/blob/master/jvm/jvm-analysis-java-tests/testData/codeInspection/missingDeprecatedAnnotationOnScheduledForRemovalApi/missingDeprecatedAnnotations.java#L8-L9
* https://github.com/JetBrains/intellij-community/blob/master/jvm/jvm-analysis-kotlin-tests-shared/testData/codeInspection/missingDeprecatedAnnotationOnScheduledForRemovalApi/missingDeprecatedAnnotations.kt#L8-L9

and I believe the intention was actually putting the annotation on field.

I checked all those "JVM" inspections and they share the same misconception: inspections visit `UField` and check an annotation, assuming `@field:` use-site annotation is associated there.

BartvHelvert:
But the annotation is on the field, take the following example:
```Java
  @Documented
  @Retention(RetentionPolicy.CLASS)
  @Target({
    ElementType.TYPE, ElementType.ANNOTATION_TYPE, ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.PACKAGE
  })
  public @interface ScheduledForRemoval {
    String inVersion() default "";
  }
```

```Kotlin
class Test {
  @ApiStatus.ScheduledForRemoval
  val x: Int = 42
}
```

In bytecode it is represented as:
```
public final class Test {
  ...

  // access flags 0x12
  private final I x
  @Lorg/jetbrains/annotations/ApiStatus$ScheduledForRemoval;() // invisible
  
  ...
}
```

So the annotation is on the field in the bytecode, because the Java annotation has a `ElementType.FIELD` target. So if we check `uField.findAnnotation(SCHEDULED_FOR_REMOVAL_FQN)` on the backing field of `x` it should return the annotation, which is not the case after your change?

plugins/kotlin/uast/uast-kotlin/tests/testData/ClassAnnotation.render.txt
```diff
@@ -7,7 +7,7 @@ public abstract annotation MyAnnotation {
 }
 
 public final class B {
-    @null public static final var Companion: B.Companion
+    @null @MyAnnotation(text = "companion") public static final var Companion: B.Companion
```
jsjeon, line 10: So, these two (below too) are a bug in ULC / K1, resulting in some test data split (e.g., `plugins/kotlin/uast/uast-kotlin-fir/tests/testData/legacyValues/ClassAnnotation.\S+.fe10.txt`).

Also, their anchor parent chains (from that annotation) become wrong. Unlike K2 UAST tests where we can control expected(?) exceptions, K1 UAST tests don't. Since they're technically duplicates, I removed some of K1 UAST tests.

plugins/kotlin/uast/uast-kotlin-fir/tests/testData/legacyRenderLog/ReceiverFun.render.txt
```diff
@@ -1,8 +1,8 @@
 public final class ReceiverFunKt {
-    public static final fun foo(@MyReceiverAnnotation @org.jetbrains.annotations.NotNull $this$foo: java.lang.String) : int {
+    public static final fun foo(@org.jetbrains.annotations.NotNull @MyReceiverAnnotation $this$foo: java.lang.String) : int {
```
jsjeon, line 2: `KotlinReceiverUParameter` had its own annotation creation which put the annotation on the receiver ahead of nullity, while everything else uses the same logic---computing nullity annotation first. Now, with the same logic, which is based on LC's use-site computation, the order is changed, but aligned with other kinds of variables.

plugins/kotlin/uast/uast-kotlin-base/src/org/jetbrains/uast/kotlin/declarations/AbstractKotlinUVariable.kt
```diff
@@ -71,22 +70,19 @@ abstract class AbstractKotlinUVariable(
...
         }
-        if (sourcePsi is KtModifierListOwner) {
-            sourcePsi.annotationEntries
-                .filter { acceptsAnnotationTarget(it.useSiteTarget?.getAnnotationUseSiteTarget()) }
-                .mapTo(annotations) { baseResolveProviderService.baseKotlinConverter.convertAnnotation(it, this) }
-        }
+        // NB: we can't use sourcePsi.annotationEntries directly due to annotation use-site targets.
+        baseResolveProviderService.getPsiAnnotations(javaPsi).asSequence()
+            .mapNotNull { (it as? KtLightElement<*, *>)?.kotlinOrigin as? KtAnnotationEntry }
+            .mapTo(annotations) { baseResolveProviderService.baseKotlinConverter.convertAnnotation(it, this) }
```
jsjeon, line 82: Now I better understand why we did this: https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/uast/uast-kotlin-base/src/org/jetbrains/uast/kotlin/declarations/KotlinUMethod.kt#L74-L80

jsjeon:
Thank you for continuous cross-check! You're right, they should be there in field. Now I understand why I needed a one-off handling for `@JvmField` because it's only applicable to field, skipping property or anything else in default ruling.

Thinking about this further, I concluded that it's an overkill to re-implement retrieving meta-annotation and seeing their annotation applicable targets. Then, maybe we should use LC's use-site computation, same as `UMethod` hierarchy does.

The caveat is... K1 / ULC has a bug---putting an annotation on (`companion`) `object`. I sorted that out by de-duping some tests (see inlined comments). Also, `javax.annotation.Tainted` and `...Untainted` really need `@field:` use-site as they don't have any specified targets.

BartvHelvert:
I think the changes make sense, but I think for us to be able to merge it we need to fix the inspections somehow. The behavior after this change for `@Deprecated` and `@Untainted`/`@Tainted` inspections is undesirable. I suspect this will break a lot of android lint detectors as well?

One possible way to solve this is to introduce something like `UProperty`, it would be a huge change and I am not sure what other problems it could cause. Maybe it makes sense to have it though because Java also supports properties (in records).

jsjeon:
> I suspect this will break a lot of android lint detectors as well?

Surprisingly, no, because Lint already has utils to detect annotation with default use-site and propagate them during annotation handling. For example, https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/libs/lint-api/src/main/java/com/android/tools/lint/detector/api/UastLintUtils.kt;l=555-613;drc=9e119d57c7969048cb2c9e34ebd572e616be6d18

Actually, inspired by that, instead of `UProperty`, how about `UField.defaultUseSiteAnnotations`? I intentionally split it into its own commit to show what is introduced; how to use that in existing inspections; and how test inputs can go back to their original form.

uast/uast-common/src/org/jetbrains/uast/declarations/UVariable.kt
```diff
@@ -95,6 +95,21 @@ interface UField : UVariable, PsiField {
...
+   * If annotations are annotated with no specific use-site, it follows default rules
+   * according to https://kotlinlang.org/docs/annotations.html#annotation-use-site-targets
+   * where `param`, `property`, and `field` are considered in order. Therefore, it is likely
+   * that the annotation is applied to property (before field) and belongs to $annotations
+   * in JVM bytecode. This might not be a user's intention, and some static analyses and/or
+   * inspections still depend on annotation without a use-site on property.
+   * To make [UVariable]'s annotation modeling accurate while mitigating existing analyses'
+   * behavior, this returns annotations with default use-site.
+   */
+  val defaultUseSiteAnnotations: List<UAnnotation>?
```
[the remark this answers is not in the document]
jsjeon: Yeah, `sourceAnnotations` sounds better. Then I dropped the use-site checks at all, and made those inspections to use `sourceAnnotations` if available, instead of weird combination.

BartvHelvert:
Merged, thanks!

jsjeon:
Thank you so much for the long review and merge!

PR #3044, opened 2025-05-08
Fix IDEA-372346 - local classes don't work with the bytecode viewer

Earthcomputer:
Fixed by adding a `JavaLocalClassesHelper` to go along with the existing `JavaAnonymousClassesHelper`. I reviewed the usages of `JavaAnonymousClassesHelper.getName`, and considered most of them could benefit from local classes too, so I extracted out the duplicated code into a new function in `ClassUtil`.

Additionally, I presume that most of the time you want the JVM class name, you actually want to support anonymous and local classes. I reviewed some usages of `ClassUtil.getJVMClassName` and it looked like they could benefit from using the new function, however I lack the domain-specific knowledge to be sure. Because it's a behaviour change, I deprecated getJVMClassName and created a new function rather than changing its behaviour, with the expectation that people will review the documentation and consciously migrate. This is possibly a controversial decision, so let me know and I can always reverse that commit.

### DIFF
--- a/java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
+++ b/java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
@@ -1,0 +1,67 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.ide.util;
+
+import com.intellij.openapi.util.Key;
+import com.intellij.psi.*;
+import com.intellij.psi.util.*;
+import com.intellij.util.containers.ObjectIntHashMap;
+import com.intellij.util.containers.ObjectIntMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class JavaLocalClassesHelper {
+  private static final Key<ParameterizedCachedValue<Map<PsiClass, String>, PsiClass>>
+    LOCAL_CLASS_NAME = Key.create("LOCAL_CLASS_NAME");
+  private static final LocalClassProvider LOCAL_CLASS_PROVIDER = new LocalClassProvider();
+
+  /**
+   * Returns the part of the class name suitable for being appended to the containing class' binary name to produce the local class' binary
+   * name, see {@link ClassUtil#getBinaryClassName} and JLS 13.1.
+   *
+   * <p>For example, if the binary name of a local class is {@code com.example.Foo$1Local}, then this method will return the {@code $1Local}
+   * part.
+   */
+  public static @Nullable String getName(@NotNull PsiClass cls) {
+    if (!PsiUtil.isLocalClass(cls)) {
+      throw new IllegalArgumentException("class " + cls + " must be a local class");
+    }
+    final PsiClass upper = PsiTreeUtil.getParentOfType(cls, PsiClass.class);
+    if (upper == null) {
+      return null;
+    }
+    ParameterizedCachedValue<Map<PsiClass, String>, PsiClass> value = upper.getUserData(LOCAL_CLASS_NAME);
+    if (value == null) {
+      value = CachedValuesManager.getManager(upper.getProject()).createParameterizedCachedValue(LOCAL_CLASS_PROVIDER, false);
+      upper.putUserData(LOCAL_CLASS_NAME, value);
+    }
+    return value.getValue(upper).get(cls);
+  }
+
+  private static final class LocalClassProvider implements ParameterizedCachedValueProvider<Map<PsiClass, String>, PsiClass> {
+    @Override
+    public CachedValueProvider.Result<Map<PsiClass, String>> compute(final PsiClass upper) {
+      final Map<PsiClass, String> map = new HashMap<>();
+      upper.accept(new JavaRecursiveElementWalkingVisitor() {
+        final ObjectIntMap<String> indexByName = new ObjectIntHashMap<>();
+
+        @Override
+        public void visitClass(@NotNull PsiClass aClass) {
+          if (aClass == upper) {
+            super.visitClass(aClass);
+          } else if (PsiUtil.isLocalClass(aClass)) {
+            String name = aClass.getName();
+            if (name != null) {
+              int index = indexByName.getOrDefault(name, 0) + 1;
+              indexByName.put(name, index);
+              map.put(aClass, "$" + index + name);
+            }
+          }
+        }
+      });
+      return CachedValueProvider.Result.create(map, upper);
+    }
+  }
+}
--- a/java/java-psi-api/src/com/intellij/psi/util/ClassUtil.java
+++ b/java/java-psi-api/src/com/intellij/psi/util/ClassUtil.java
@@ -215,6 +215,12 @@ public final class ClassUtil {
     return Character.isDigit(name.charAt(0));
   }
 
+  /**
+   * Returns the binary class name for top-level and nested classes.
+   *
+   * @deprecated Does not work for anonymous classes and local classes. Use {@link #getBinaryClassName} instead.
+   */
+  @Deprecated
   public static @Nullable @NlsSafe String getJVMClassName(@NotNull PsiClass aClass) {
     final PsiClass containingClass = aClass.getContainingClass();
     if (containingClass != null) {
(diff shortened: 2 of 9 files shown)


java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
```diff
@@ -0,0 +1,60 @@
...
+import com.intellij.util.containers.ObjectIntHashMap;
+import com.intellij.util.containers.ObjectIntMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Joseph Burton
```
BartvHelvert, line 16: Could you please remove the author tag? We don't really use them anymore because git blame is more accurate.

BartvHelvert:
Could you also change to commit message to use YouTrack VSC commands? https://www.jetbrains.com/help/youtrack/server/apply-commands-in-vcs-commits.html

So instead of mentioning the YouTrack issue in the subject line, you can need to add `#IDEA-372346 fixed` to the commit body. This will make sure all YouTrack fields will be set automatically, thanks!

java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
Earthcomputer: Sure thing

Earthcomputer:
> Could you also change to commit message to use YouTrack VSC commands?

How do I do that given there are multiple commits?

BartvHelvert:
> How do I do that given there are multiple commits?

`#IDEA-372346 fixed` will change the state of the issue to Fixed, just writing `#IDEA-372346` will attach the commit message to the ticket without any state changes.

Earthcomputer:
I marked the first commit to fix the issue, and the one which deprecates the method to reference it.

java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public final class JavaLocalClassesHelper {
+  private static final Key<ParameterizedCachedValue<Map<PsiClass, String>, PsiClass>>
+    LOCAL_CLASS_NAME = Key.create("LOCAL_CLASS_NAME");
+  private static final LocalClassProvider LOCAL_CLASS_PROVIDER = new LocalClassProvider();
+
+  public static @Nullable String getName(@NotNull PsiClass cls) {
+    final PsiClass upper = PsiTreeUtil.getParentOfType(cls, PsiClass.class);
```
bartekpacia, line 21: Maybe we can add an assertion here in the first line of this method, e.g.:

```java
if (!com.intellij.psi.util.PsiUtil.isLocalClass(cls)) {
  throw new IllegalArgumentException("class " + psiClass + " must be a local class");
}
```

java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class JavaLocalClassesHelper {
+  private static final Key<ParameterizedCachedValue<Map<PsiClass, String>, PsiClass>>
+    LOCAL_CLASS_NAME = Key.create("LOCAL_CLASS_NAME");
+  private static final LocalClassProvider LOCAL_CLASS_PROVIDER = new LocalClassProvider();
+
+  public static @Nullable String getName(@NotNull PsiClass cls) {
```
bartekpacia, line 20: I'd appreciate a small Javadoc here, for example:

---

Returns the binary name of a local Java class or interface. See [JLS 13.1](https://docs.oracle.com/javase/specs/jls/se24/html/jls-14.html#jls-13.1).

The binary name of a local class or interface consists of the binary name of its immediately enclosing class or interface, followed by $, followed by a non-empty sequence of digits, followed by the simple name of the local class.

### Example

Consider this Java snippet:

```java
package com.example.localclass;

public class Main {
    class Local {
      void a() {
        System.out.println("Hi");
      }
    }
  }
}
```

Calling this method with the local class `Local` as argument returns `com.example.localclass.Main$1Local`.

bartekpacia: I know documentation of the IJ codebase is not the best, but with small steps like this we can make the situation a bit better, I hope :)

java/java-psi-api/src/com/intellij/psi/util/ClassUtil.java
```diff
@@ -225,6 +233,31 @@ private static boolean isIndexed(String name) {
     return aClass.getQualifiedName();
   }
 
+  /**
+   * Returns the binary class name, i.e. the string that would be returned by {@link Class#getName}.
+   *
+   * <p>Inner classes ({@code Foo$Bar}), anonymous ({@code Foo$1}), and local ({@code Foo$1Bar}) classes are taken into account.
+   */
+  public static @Nullable @NlsSafe String getBinaryClassName(@NotNull PsiClass aClass) {
```
bartekpacia, line 241: Huge thanks for writing the Javadoc. Minor suggestion:

I know `Class#getName` refers to JLS too but I think it's useful to surface this info early.

```suggestion
  /**
   * Returns the binary class name, i.e. the string that would be returned by {@link Class#getName}. See JLS 13.1.
   *
   * <p>Inner ({@code Foo$Bar}), anonymous ({@code Foo$1}), and local ({@code Foo$1Bar}) classes are taken into account.
   */
  public static @Nullable @NlsSafe String getBinaryClassName(@NotNull PsiClass aClass) {
```

bartekpacia:
> I reviewed some usages of `ClassUtil.getJVMClassName` and it looked like they could benefit from using the new function, however I lack the domain-specific knowledge to be sure. Because it's a behaviour change, I deprecated `getJVMClassName` and created a new function rather than changing its behaviour, with the expectation that people will review the documentation and consciously migrate. This is possibly a controversial decision, so let me know and I can always reverse that commit.

Thanks for doing this!

Some additional context:
- there are 70 usages of the now-deprecated `com.intellij.psi.util.ClassUtil#getJVMClassName(PsiClass)` in the whole IJ monorepo. I also lack knowledge whether it is safe to replace them with `com.intellij.psi.util#getBinaryClassName(PsiClass)`
- there is also a very-similar looking method `com.intellij.lang.jvm.util.JvmClassUtil#getJvmClassName(JvmClass)`, which has 2 usages: on in our DevKit plugin and one in the Android plugin.

bartekpacia (approved):
Overall LGTM, but I'd like a review from someone else too.

java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
```diff
@@ -0,0 +1,67 @@
...
+  private static final LocalClassProvider LOCAL_CLASS_PROVIDER = new LocalClassProvider();
+
+  /**
+   * Returns the part of the class name suitable for being appended to the containing class' binary name to produce the local class' binary
+   * name, see {@link ClassUtil#getBinaryClassName} and JLS 13.1.
+   *
+   * <p>For example, if the binary name of a local class is {@code com.example.Foo$1Local}, then this method will return the {@code $1Local}
+   * part.
+   */
+  public static @Nullable String getName(@NotNull PsiClass cls) {
```
BartvHelvert, line 27: Could we maybe have unit tests for this?

BartvHelvert (changes requested):

java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
BartvHelvert: There is already an example of such tests `JavaAnonymousClassesHelperTest`

bartekpacia:
I really appreciate the changes you made to use the new util method in `java/debugger` and `java/execution`, but let's keep this PR as small as possible, and make those other changes in a subsequent PR. In case something breaks, it'll be easier to bisect and revert.

I hope it's not too much of a burden for you.

java/java-psi-api/src/com/intellij/ide/util/JavaLocalClassesHelper.java
bartekpacia: Thanks, the new Javadoc looks great!

Earthcomputer:
All done

bartekpacia:
Thanks, I scheduled a merge. 
Once it lands (soon), this PR should be closed automatically.

PR #3045, opened 2025-05-09
Fix IDEA-372415 - Convert anonymous to inner refactoring doesn't copy generics when non-static

Earthcomputer:
In `AnonymousToInnerHandler.calculateTypeParametersToCreate`, I believe the faulty check is `CommonJavaRefactoringUtil.isInStaticContext`, this doesn't make much sense to me. Instead, when the new class is non-static, you should be checking if the type parameter would still be in the lexical scope. In this case, this is equivalent to checking if the type parameter owner is a (non-strict) ancestor of the target class (the class into which we are adding the inner class).


Earthcomputer:
Forced pushed to add YouTrack VCS command

fivemtebex:
Very clean code with good test coverage. This is a great example of best practices.

fivemeyestore:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

fivemhostingserver:
Very clean code with good test coverage. This is a great example of best practices.

scriptforfivem:
Excellent contribution to open source! The community will benefit greatly from this.

bartekpacia (approved):
Hi Joseph, thank you very much for your contribution.

LGTM – we will merge it shortly.

PR #3054, opened 2025-05-20
[JEWEL-788] Allow customising language recognition in MarkdownProcessor

rock3r:
This adds a way to provide a `languageRecognizer` parameter to the `MarkdownProcessor` such that custom languages can be recognised and converted to `MimeType`.

By default, this uses the existing `MimeType.Known.fromMarkdownLanguageName()` as it was doing before.

This also adds `PATCH` and `DIFF` to the known mime types.


rock3r:
Ready to merge

PR #3056, opened 2025-05-27
[JEWEL-821] Fix Jewel publishing

rock3r:
This adds a `patchDependencies` capability to `MavenArtifactsBuilder`, which is used by `JewelMavenArtifacts` to change the dependencies that are added to the POM files. 

This way, we set up the correct scopes and exclusions, cleaning up unnecessary dependencies that only exist in the JPS build to accommodate IJP quirks. This makes the artefacts usable for standalone builds.

In particular, the patching:
 * Drops unneeded/undesirable IJP-only dependencies (e.g., `skiko-awt-runtime-all` and the forked coroutines-core)
 * Omits dependencies that are brought in transitively, letting Maven/Gradle do it
 * Ensures proper scoping for all dependencies (mostly "compile"; only Markdown extensions in the Markdown styling module are supposed to be "runtime")
 * Removes exclusions that only make sense in the JPS build, restoring proper transitivity

### DIFF
--- a/build/src/org/jetbrains/intellij/build/IdeaCommunityProperties.kt
+++ b/build/src/org/jetbrains/intellij/build/IdeaCommunityProperties.kt
@@ -78,17 +78,23 @@ open class IdeaCommunityProperties(private val communityHomeDir: Path) : BaseIde
     ))
     mavenArtifacts.patchCoordinates = { module, coordinates ->
       when {
-        JewelMavenArtifacts.isJewel(module) -> JewelMavenArtifacts.patchCoordinates(module, coordinates)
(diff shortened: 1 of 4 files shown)


DmPanov (approved):

rock3r:
Ready to merge

rock3r:
Is it because the monorepo build is trying to publish the bridge artefacts? It shouldn't. The community build doesn't, and so doesn't fail, but I have no way to test the monorepo build.

rock3r:
@DmPanov pushed an update to ignore non-published modules, should fix the failing test.

rock3r:
Ready to merge (but might require changes in the monorepo)

PR #3057, opened 2025-05-27
[JEWEL-822] Target IJP 252 and add bridge tests to Gradle build

rock3r:
This adds the bridge tests to the Gradle build (were broken due to a missing path segment in the settings.gradle.kts file), and moves the target IJP to 252 EAP to match the JPS build.

Also fix a minor issue in SwingBridgeService where we created the theme definition twice for no reason. Lastly, it bumps Gradle to 8.14 and the IJP plugin to 2.6.0.


rock3r:
I'll split this in two parts since the bridge tests have some issues running in the Gradle build

PR #3059, opened 2025-05-28
[JEWEL-822] Target IJP 252 in Gradle build

rock3r:
This aligns the Gradle build to the latest 252 EAP, bumps Gradle to 8.14 and the IJP plugin to 2.6.0. Kotlin is updated to 2.1.20 to match IJP.


rock3r:
Ready to merge

PR #3060, opened 2025-05-28
[JEWEL-826] Add bridge tests in Gradle build

rock3r:
This aligns the Gradle build to the JPS build by adding the bridge tests so they get run together with the rest.

> [!IMPORTANT]
> Since these tests depend on 252 APIs, this requires #3059 to be merged first. In addition, when backporting to 251 and 243, you’ll need to make sure the new 252 API is not included.


rock3r:
Ready to merge

PR #3062, opened 2025-05-29
[JEWEL-829] Fix the bridge not reading the LaF

rock3r:
In JEWEL-558, a change was made to add a new flow to the combine in `SwingBridgeService`. However, that new flow never emitted, causing the whole `combine` to stall until it did. This in turn causes only the theme defaults to be loaded, until the user changes the "contrast scrollbars" option, making the faulty flow emit and unblocking the `combine`.

The fix is to ensure the flow emits on start, like all other flows in the `combine` do.


jakub-senohrabek:
Ready to merge

PR #3064, opened 2025-05-30
[JEWEL-830] Fix Editor TextStyle line height

rock3r:
Swing is not using the font size when computing the base lineHeight, that is then multiplied by the user-set factor (default 1.2). Instead, they use FontMetrics.getHeight(), which computes it as leading + ascent + descent.

This changes the Jewel line height computation for the editor text style to use the same logic. It requires a Graphics2D instance, which is obtained from a Swing BufferedImage on demand, to get the same font metrics as the Swing UI uses.

This gets us a much closer result, but is still not 100% matching. It looks like this introduces a ~10% extra line height in Compose compared to Swing:

image

I will make a follow-up change when I figure out where in Skia’s line height usage there is a divergence from Swing’s, but for now this is already much better than what we did before.

This change also fixes the Jewel sample plugin.xml so it doesn’t crash at startup when run in the IJP build.


rock3r:
Reopening — this works fine for the bridge, but we'll need to tweak standalone too at some point. I decided to split it out to another issue since it's growing in scope quite rapidly

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
```diff
@@ -6,27 +6,69 @@ import androidx.compose.ui.text.platform.asComposeFontFamily
...
+ * Retrieves the default text style from the Swing LaF, applying a line height multiplier.
+ *
+ * This is the font set in _Settings | Appearance & Behavior | Appearance_, and defaults to 13px Inter.
+ */
+@ExperimentalJewelApi
+public fun retrieveDefaultTextStyle(lineHeightMultiplier: Float): TextStyle {
+    val lafFont = UIManager.getFont(KEY_LABEL_FONT) ?: keyNotFound(KEY_LABEL_FONT, "Font")
+    val font = JBFont.create(lafFont, false)
+    val baseLineHeight = computeBaseLineHeightFor(font)
+    return retrieveDefaultTextStyle().copy(lineHeight = (baseLineHeight * lineHeightMultiplier).sp)
```
faogustavo, line 41: Can't this be simplified by using `lineHeightMultiplier.em` instead of calculating the size and then multiplying it?

rock3r: Not sure I follow. Why would I need to treat the multiplier as em? And, the point of this change is that we are replacing skia's calculations with swing's, so if I get rid of the swing calculations, I don't have much left at all, do I?

I am not sure I understand what you're asking 😅

faogustavo: My point is that we may not need to compute the line height to make the multiplication. The `em` size already handles that directly while drawing the text. So instead of doing 

```
public fun retrieveDefaultTextStyle(lineHeightMultiplier: Float): TextStyle {
    val lafFont = UIManager.getFont(KEY_LABEL_FONT) ?: keyNotFound(KEY_LABEL_FONT, "Font")
    val font = JBFont.create(lafFont, false)
    val baseLineHeight = computeBaseLineHeightFor(font)
    return retrieveDefaultTextStyle().copy(lineHeight = (baseLineHeight * lineHeightMultiplier).sp)
}
```

we could simplify with something similar to this:

```
public fun retrieveDefaultTextStyle(lineHeightMultiplier: Float): TextStyle =
    retrieveDefaultTextStyle().copy(lineHeight = lineHeightMultiplier.em)
```

faogustavo: > the point of this change is that we are replacing skia's calculations with swing's

In this case using EM would not work, nevermind 😅

faogustavo (approved):

rock3r:
This builds in the JPS build, but fails in the Gradle build, since the Gradle build doesn't use Jewel 0.28 yet. I'll revert the change that causes the problem asap.

DanielSouzaBertoldi (approved):

rock3r:
Ready to merge

PR #3067, opened 2025-05-30
Update "Hello and welcome!" templates to use println

cargilltay:
The default `Main.java` file that is generated for a new project uses `System.out.printf()`. The IDE immediately underlines this and suggests that it is a "Redundant call to 'printf()'". 
warning

In addition, the output for this example is not ideal. Using `println()` will make the output more readable.

Before:
```
Hello and welcome!i = 1
i = 2
i = 3
i = 4
i = 5
```

After:
```
Hello and welcome!
i = 1
i = 2
i = 3
i = 4
i = 5
```


BartvHelvert:
@cargilltay thank you for the PR! I don't think we can merge it though, the warning is actually put there on purpose to teach newcomers how `alt+enter` works. If we just remove it, the comment above the statement doesn't make sense anymore. If you still think the current behavior is wrong, please open an issue in our [issue tracker](https://youtrack.jetbrains.com) so we can discuss it.

PR #3072, opened 2025-06-03
[JEWEL-818] Make LaF Parsing More Defensive Against Bad Themes 

DanielSouzaBertoldi:
# Context

Third party themes can customize spacings/dimensions/font sizes and whatnot with values that can cause Jewel to crash. This PR goes through all code that fetch information from a theme and coerces the value to be at minimum 0.

There's no evidence this time since it's basically just a matter of adding `coerceAtLeast` to all theme calls. I also took this opportunity to do some pretty minor refactor on some classes that I stumbled upon.

Once again, feel free to suggest as many changes as you'd want! 😎


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/BridgeGlobalMetrics.kt
```diff
@@ -4,11 +4,12 @@ import androidx.compose.ui.unit.dp
...
 import com.intellij.util.ui.JBUI
+import org.jetbrains.jewel.bridge.safeValue
 import org.jetbrains.jewel.foundation.GlobalMetrics
 
 public fun GlobalMetrics.Companion.readFromLaF(): GlobalMetrics =
     GlobalMetrics(
-        outlineWidth = DarculaUIUtil.BW.unscaled.dp,
+        outlineWidth = DarculaUIUtil.BW.unscaled.dp.safeValue(),
         // The rowHeight() function returns a scaled value, but we need the base value
         rowHeight = (JBUI.CurrentTheme.List.rowHeight() / JBUIScale.scale(1f)).dp,
```
DanielSouzaBertoldi: There's no need to call `.safeValue()` in `rowHeight()` since it actually returns a default value:

```java
static int rowHeight() {
    int defaultHeight = JBUIScale.scale(24);
    int result = JBUI.getInt("List.rowHeight", defaultHeight);
    return result <= 0 ? defaultHeight : result;
}
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeComboBox.kt
```diff
@@ -44,7 +45,7 @@ internal fun readDefaultComboBoxStyle(): ComboBoxStyle {
         )
 
     val minimumSize = JBUI.CurrentTheme.ComboBox.minimumSize().toDpSize()
```
DanielSouzaBertoldi: Also, we are already coercing the `minimumSize()` result to `0.dp` in `BridgeUtils.kt`, so we don't need to do it again here.

This is also the case for calls like `retrieveIntAsDpOrUnspecified` and any other that retrieves/parses/handles values that come from themes in `BridgeUtils.kt`

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
```diff
@@ -16,7 +17,7 @@ public fun retrieveDefaultTextStyle(): TextStyle = retrieveTextStyle("Label.font
 public fun retrieveEditorTextStyle(): TextStyle {
     val editorColorScheme = retrieveEditorColorScheme()
 
-    val fontSize = editorColorScheme.editorFontSize.sp
+    val fontSize = editorColorScheme.editorFontSize.safeValue().sp
```
DanielSouzaBertoldi: This actually returns a `0.sp` value. Not sure if we should treat font size as 0 or 1, or `Text.Unspecified.value.sp`. Tell me your thoughts about it 😄

platform/jewel/ide-laf-bridge/api/ide-laf-bridge.api
```diff
@@ -41,6 +41,10 @@ public final class org/jetbrains/jewel/bridge/BridgeUtilsKt {
 	public static synthetic fun retrieveTextStyle-WdJyH8Q$default (Ljava/lang/String;JJZIJILjava/lang/Object;)Landroidx/compose/ui/text/TextStyle;
 	public static final fun retrieveTextStyle-tD9LlGs (Ljava/lang/String;Ljava/lang/String;JZIJ)Landroidx/compose/ui/text/TextStyle;
 	public static synthetic fun retrieveTextStyle-tD9LlGs$default (Ljava/lang/String;Ljava/lang/String;JZIJILjava/lang/Object;)Landroidx/compose/ui/text/TextStyle;
+	public static final fun safeValue (II)I
```
rock3r: Can you also please run `ApiCheckTest` and apply the diffs it finds to the .txt dumps?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -52,6 +53,10 @@ public fun Color.toAwtColor(): java.awt.Color = java.awt.Color(red, green, blue,
 
 public fun java.awt.Color?.toComposeColorOrUnspecified(): Color = this?.toComposeColor() ?: Color.Unspecified
 
+public fun Dp.safeValue(minimumValue: Dp = 0.dp) = this.coerceAtLeast(minimumValue)
```
rock3r: Please keep these internal, we should not be exposing them.

rock3r: Would also be nice to add a tiny KDoc explaining why we need them, considering we use them everywhere now :)

Lastly, nit: try keeping top-down ordering, putting them after their last usage in the file

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -189,14 +209,17 @@ public fun retrieveTextStyle(
             .let { if (bold) it.asBold() else it.asPlain() }
             .let { if (fontStyle == FontStyle.Italic) it.asItalic() else it }
 
+    val safeFontSize = if (derivedFont.size > 0) derivedFont.size.sp else TextUnit.Unspecified
+    val safeLineHeight = if (lineHeight.value > 0) lineHeight else TextUnit.Unspecified
+
     return TextStyle(
         color = color,
-        fontSize = size.takeOrElse { derivedFont.size.sp / UISettingsUtils.getInstance().currentIdeScale },
+        fontSize = size.takeOrElse { safeFontSize / UISettingsUtils.getInstance().currentIdeScale },
```
rock3r: You’re not constraining the size we get passed in, is that intended?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeService.kt
```diff
@@ -57,7 +57,7 @@ internal class SwingBridgeService {
     private fun readThemeData(): BridgeThemeData {
         val themeDefinition = createBridgeThemeDefinition()
         return BridgeThemeData(
-            themeDefinition = createBridgeThemeDefinition(),
+            themeDefinition = themeDefinition,
```
rock3r: I think I have the same fix elsewhere, but better to have it twice than never :D

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/BridgeGlobalMetrics.kt
rock3r: That is not necessarily true — if the LaF defines `List.rowHeight`, that value will be used (the built-in themes don’t, but nothing stops another theme from doing it).

Don't confuse fallbacks for hardcoded values. Any time there is a LaF key lookup, it means the theme can provide that value, and thus we need to sanitise the value we get from it.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeBanner.kt
```diff
@@ -20,7 +21,7 @@ internal fun readDefaultBannerStyle(): DefaultBannerStyles =
                         background = JBUI.CurrentTheme.Banner.INFO_BACKGROUND.toComposeColor(),
                         border = JBUI.CurrentTheme.Banner.INFO_BORDER_COLOR.toComposeColor(),
                     ),
-                metrics = BannerMetrics(borderWidth = DarculaUIUtil.LW.dp),
+                metrics = BannerMetrics(borderWidth = DarculaUIUtil.LW.dp.safeValue()),
```
rock3r: Nit: might be a good opportunity to extract an `internal val borderWidth` and put it next to `componentArc` in `IntUiBridge.kt`

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeMenu.kt
```diff
@@ -69,15 +71,15 @@ internal fun readMenuStyle(): MenuStyle {
...
         metrics =
             MenuMetrics(
-                cornerSize = CornerSize(IdeaPopupMenuUI.CORNER_RADIUS.dp),
+                cornerSize = CornerSize(IdeaPopupMenuUI.CORNER_RADIUS.dp.safeValue()),
                 menuMargin = PaddingValues(vertical = 6.dp),
                 contentPadding = PaddingValues(vertical = 7.dp, horizontal = 2.dp),
                 offset = DpOffset(0.dp, 2.dp),
                 shadowSize = 12.dp,
-                borderWidth = retrieveIntAsDpOrUnspecified("Popup.borderWidth").takeOrElse { 1.dp },
+                borderWidth = retrieveIntAsDpOrUnspecified("Popup.borderWidth").takeOrElse { 1.dp }.coerceAtLeast(1.dp),
```
rock3r: Isn’t this the same as `safeValue(1.dp)`?

I'm also not entirely sure why we need the border to be at least 1.dp; a theme could want it to be 0.dp.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeMenu.kt
```diff
@@ -89,10 +91,20 @@ internal fun readMenuStyle(): MenuStyle {
                                     retrieveIntAsDpOrUnspecified("PopupMenuSeparator.stripeIndent").takeOrElse { 1.dp },
                             ),
                         separatorThickness =
-                            retrieveIntAsDpOrUnspecified("PopupMenuSeparator.stripeWidth").takeOrElse { 1.dp },
-                        separatorHeight = retrieveIntAsDpOrUnspecified("PopupMenuSeparator.height").takeOrElse { 3.dp },
+                            retrieveIntAsDpOrUnspecified("PopupMenuSeparator.stripeWidth")
+                                .takeOrElse { 1.dp }
+                                .coerceAtLeast(1.dp),
```
rock3r: Ditto

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeMenu.kt
```diff
@@ -89,10 +91,20 @@ internal fun readMenuStyle(): MenuStyle {
...
+                        separatorHeight =
+                            retrieveIntAsDpOrUnspecified("PopupMenuSeparator.height")
+                                .takeOrElse { 3.dp }
+                                .coerceAtLeast(3.dp),
```
rock3r: Ditto, if the theme wants this to be 0.dp, why are you preventing it?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeMenu.kt
```diff
@@ -89,10 +91,20 @@ internal fun readMenuStyle(): MenuStyle {
...
                         iconSize = 16.dp,
-                        minHeight = if (isNewUiTheme()) JBUI.CurrentTheme.List.rowHeight().dp else Dp.Unspecified,
+                        minHeight =
+                            if (isNewUiTheme()) {
+                                JBUI.CurrentTheme.List.rowHeight().dp
```
rock3r: Missing `safeValue()`

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -85,8 +90,8 @@ public fun List<Color>.createVerticalBrush(
 
 public fun retrieveIntAsDp(key: String, default: Dp? = null): Dp {
```
rock3r: I wouldn’t change the behaviour of this function, which is a public API and can have negative effects on users. You should create a new parallel `retrieveIntAsNonNegativeDp` (mark it as experimental, or make it internal).

Same for `Insets.toPaddingValues`, `Dimension.toDpSize`, `JBDimension.toDpSize`, etc.

Note that as far as I can tell at a quick glance, only `PaddingValues` hard-requires values to be non-negative. Others will accept negative values and eventually error out on usage, depending on the usage, so we should probably apply `safeValue()` on a case-by-case manner, instead of at this level, impacting everything.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
rock3r: Hmmm it really should not return zero — there is something wrong here. On my machine this returns the actual editor font size in pixel in the IDE. In fact, looking at the IJ code (`FontPreferencesImpl.getSize`) it can never return zero, only 1 or larger:

```java
public static final int DEFAULT_FONT_SIZE = SystemInfo.isWindows || JETBRAINS_MONO.equalsIgnoreCase(DEFAULT_FONT_NAME) ? 13 : 12;

// ==== OTHER FILE ====
@Override
public int getSize(@NotNull String fontFamily) {
  return (int)(getSize2D(fontFamily) + 0.5);
}
@Override
public float getSize2D(@NotNull String fontFamily) {
  float result = myFontSizes.getFloat(fontFamily);
  if (result <= 0) {
    result = myTemplateFontSize; // Defaults to DEFAULT_FONT_SIZE, never changes IRL
  }
  return result > 0 ? result : DEFAULT_FONT_SIZE;
}
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/BridgeGlobalMetrics.kt
DanielSouzaBertoldi: Okay! Since we are calling the `rowHeight()` function I thought we'd always get the `defaultHeight` in case the result was <= 0. I'll coerce the function result too then.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
DanielSouzaBertoldi: True. What I tried to say is that by having the `safeValue()` call, we could return a 0.sp font size (if the IJ code didn't check it, that is). Thanks for sharing the code, totally missed that 🙏🏻

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeMenu.kt
DanielSouzaBertoldi: I thought we'd like to reinforce the fallback value in the `takeOrElse`, but you've a point, I'll call `safeValue()` here and every other place like it 😄

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
DanielSouzaBertoldi: Missed that, sorry :( I was going to say that this property isn't being defined by a LaF property, but that's also the case for `lineHeight` which I am constraining, so it makes sense to constraint `size` too.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
rock3r: Yeah for font size I'd force a minimum of 1.sp, probably

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeMenu.kt
rock3r: The fallback value is not the minimum valid value tho, it's only the default to use when the theme doesn't give us a value :)

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -157,6 +196,18 @@ public fun retrieveArcAsCornerSizeWithFallbacks(vararg keys: String): CornerSize
     keysNotFound(keys.toList(), "Int")
 }
 
+/**
+ * To avoid having Jewel crash due to negative values that third-party themes can define, we need to constraint values
```
rock3r:
```suggestion
 * To avoid having Jewel crash due to negative values that third-party themes can define, we need to constrain values
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -157,6 +196,18 @@ public fun retrieveArcAsCornerSizeWithFallbacks(vararg keys: String): CornerSize
...
+ * that are read from LaF.
+ */
+internal fun Dp.safeValue(minimumValue: Dp = 0.dp) = this.coerceAtLeast(minimumValue)
+
+/**
+ * To avoid having Jewel crash due to negative values that third-party themes can define we need to constraint values
```
rock3r:
```suggestion
 * To avoid having Jewel crash due to negative values that third-party themes can define we need to constrain values
```

rock3r (approved):
LGTM

@fscarponi can you please also take a look?

fscarponi:
> LGTM, couple small typos to fix
> 
> @fscarponi can you please also take a look?

I think that all known cases are covered! Thanks for ping! 
LGTM! (and thank you @DanielSouzaBertoldi )

rock3r:
Ready to merge

rock3r:
Ready to merge

PR #3074, opened 2025-06-03
[JEWEL] Exclude samples:ide-plugin from GitHub CI

rock3r:
This is needed because the ide-plugin currently only can access a published IJ build, which might not contain necessary new APIs.

For samples:ide-plugin we only run ktfmtCheck, as that does not require the project to compile.

See JEWEL-825 for further details.


jakub-senohrabek (approved):

rock3r:
Ready to merge

PR #3076, opened 2025-06-04
Update `MayBeConstant` inspection

martinbonnin:
I have once again been bitten by this. While I agree `const` is a reasonnable default for most app authors out there, the situation is a lot different for library authors.

I don't think it's worth disabling the inspection all together (but maybe it is? curious what you think?). This PR is adding some explicit wording for library authors to at least raise the potential tradeoffs.


vladimirdolzhenko:
LGTM

PR #3078, opened 2025-06-04
[JEWEL-840] Fix styling not always updating in bridge

rock3r:
This is especially visible in Markdown, which doesn’t update its styling when the editor or LaF change, due to a combination of badly keyed remember calls, and lack of recomposition when the editor colour scheme changes.

This makes three changes:
1. Adds the editorSchemeChangeFlow to the SwingBridgeReader (renamed from SwingBridgeService as it’s not a service anymore), so now editor colour scheme changes will also cause a theme re-read, and a recomposition
2. Adds an experimental `LocalInstanceUuid`, which is a UUID that’s re-computed every time the theme definition changes. It can be used as a better proxy for "the theme was updated" than the theme name, since it will change whenever anything in the theme definition is changed for whatever reason.
3. Fix the `remember` keys in `ProvideMarkdownStyling` to use the new `instanceUuid` (or whatever’s appropriate), fixing issues where a font change was not reflected in the Markdown until the composition was completely stopped and restarted.

There is also a minor amount of code clean-up, mostly removing unused private/internal code, and a couple of other minor fixes.

### Before

https://github.com/user-attachments/assets/f89e025f-b18e-4fc3-ac99-01c24780fe54

### After

https://github.com/user-attachments/assets/3888c885-d6fa-4d13-b352-d8f18e201031


rock3r:
Ready to merge

faogustavo (approved):

PR #3079, opened 2025-06-04
[JEWEL] Remove dupe sample plugin manifest

rock3r:
We have both `/intellij.platform.jewel.samples.idePlugin.xml` and `/META-INF/plugin.xml` in the `samples/ide-plugin` resources.

Having two causes the plugin to fail to load if both are "fully filled in", and with some experimentation I determined that the `META-INF/plugin.xml` is the right file to keep. Keeping the one we’re deleting causes the plugin not to work at all, probably because it’s in the wrong folder in the resources.


rock3r:
Ready to merge

jakub-senohrabek:
Let's wait for the dry run to finish to see if this breaks the community/ultimate build somehow.

jakub-senohrabek (approved):

PR #3080, opened 2025-06-05
KT UAST: add a test about property accessor from binary

jsjeon:
^KT-78076


jsjeon:
cc @BartvHelvert / @dimonchik0036 

K2 variant is supposed to be failed without https://github.com/JetBrains/kotlin/pull/5459

PR #3084, opened 2025-06-06
[JEWEL-839] Delete old .github folder

faogustavo:
Github only uses the root dir configuration which already has a Jewel configuration


rock3r (approved):

rock3r:
Ready to merge

PR #3085, opened 2025-06-06
[JEWEL-798] Zero-Delay + Never-Hide tooltip

faogustavo:
- Added TooltipAutoHideBehavior to the tooltip theme
- Updated instance from AutoHideBehavior to TypeAlias with deprecated annotation
- Updated all Jewel components to use non-deprecated APIs por Tooltip


platform/jewel/int-ui/int-ui-standalone/api-dump-unreviewed.txt
```diff
@@ -551,15 +551,6 @@ f:org.jetbrains.jewel.intui.standalone.styling.IntUiTextFieldStylingKt
...
-- sf:dark--hDc7uY(org.jetbrains.jewel.ui.component.styling.TooltipColors$Companion,J,J,J,J):org.jetbrains.jewel.ui.component.styling.TooltipColors
-- bs:dark--hDc7uY$default(org.jetbrains.jewel.ui.component.styling.TooltipColors$Companion,J,J,J,J,I,java.lang.Object):org.jetbrains.jewel.ui.component.styling.TooltipColors
-- sf:light(org.jetbrains.jewel.ui.component.styling.TooltipStyle$Companion,org.jetbrains.jewel.ui.component.styling.TooltipColors,org.jetbrains.jewel.ui.component.styling.TooltipMetrics):org.jetbrains.jewel.ui.component.styling.TooltipStyle
-- bs:light$default(org.jetbrains.jewel.ui.component.styling.TooltipStyle$Companion,org.jetbrains.jewel.ui.component.styling.TooltipColors,org.jetbrains.jewel.ui.component.styling.TooltipMetrics,I,java.lang.Object):org.jetbrains.jewel.ui.component.styling.TooltipStyle
-- sf:light--hDc7uY(org.jetbrains.jewel.ui.component.styling.TooltipColors$Companion,J,J,J,J):org.jetbrains.jewel.ui.component.styling.TooltipColors
-- bs:light--hDc7uY$default(org.jetbrains.jewel.ui.component.styling.TooltipColors$Companion,J,J,J,J,I,java.lang.Object):org.jetbrains.jewel.ui.component.styling.TooltipColors
```
faogustavo: Moved those to the "reviewed" file

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tooltip.kt
```diff
@@ -245,6 +321,12 @@ public fun rememberPopupPositionProviderAtFixedPosition(
         }
     }
 
+@Deprecated(
+    "Replace with TooltipAutoHideBehavior",
+    replaceWith =
+        ReplaceWith("TooltipAutoHideBehavior", "org.jetbrains.jewel.ui.component.styling.TooltipAutoHideBehavior"),
+    level = DeprecationLevel.WARNING,
+)
 public enum class AutoHideBehavior {
```
rock3r, line 265: Nit: I'd add a todo here for whoever will come and delete this API, telling them to move the enum class definition to `TooltipAutoHideBehavior`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/theme/TooltipStyle.kt
```diff
@@ -0,0 +1,25 @@
+package org.jetbrains.jewel.ui.theme
+
+import kotlin.time.Duration.Companion.milliseconds
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.ui.component.styling.TooltipAutoHideBehavior
+import org.jetbrains.jewel.ui.component.styling.TooltipMetrics
+import org.jetbrains.jewel.ui.component.styling.TooltipStyle
+
+@ExperimentalJewelApi
+public fun TooltipStyle.zeroDelayNeverHide(): TooltipStyle =
```
rock3r: We don’t usually do this; is there a specific reason why we're adding this to the API?

If it's only for convenience in the showcase, I'd move it there. Folks can reimplement this easily enough, or copy it from there.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tooltip.kt
```diff
@@ -65,6 +66,81 @@ import org.jetbrains.jewel.ui.util.isDark
  * @see com.intellij.ide.HelpTooltip
  */
 @Composable
+public fun Tooltip(
+    tooltip: @Composable () -> Unit,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    style: TooltipStyle = JewelTheme.tooltipStyle,
+    tooltipPlacement: TooltipPlacement = style.metrics.placement,
+    autoHideBehavior: TooltipAutoHideBehavior = TooltipAutoHideBehavior.Normal,
```
rock3r: I don’t love that now we need to specify the `autoHideBehaviour` everywhere. 

Considering that you’ve added the behaviour to the style, why not remove the parameter from here and just obtain the value from the style instead?

```suggestion
```

That'll also fix the issue that the behaviour in the style is completely unused :)

faogustavo: This one is messed up. It should be getting from the styles. I'll update it.

From using it everywhere, I agree. But that's caused by both this and the deprecated functions (linked below) having the "same signature" (so we need to identify which one we want to call).

About removing the parameter, I'd go that route, but we have one deprecated function that would conflict with it.

```kotlin
@ScheduledForRemoval(inVersion = "2025.2")
@Deprecated("Please, use the overload with [AutoHideBehavior].")
@Composable
public fun Tooltip(
    tooltip: @Composable () -> Unit,
    enabled: Boolean = true,
    style: TooltipStyle = JewelTheme.tooltipStyle,
    tooltipPlacement: TooltipPlacement = style.metrics.placement,
)
```

If we are okay with changing the behavior from this old deprecated function and "un-deprecate it", we could do the change. But I think it would be a behavior change (which I'm not sure if we consider a breaking change).

rock3r: AFAICT, the old one should default to the "normal" behaviour. That should maintain the existing behaviour anyway if folks don't go and modify the style. I think we can un-deprecate it :)

faogustavo: > AFAICT, the old one should default to the "normal" behaviour. 

The one that fallback to default behavior is the one with the `AutoHideBehavior`. The one without argument is basically a "never hide" as it does not implement the hiding logic

rock3r: Ah, I got them confused. We've had that deprecated for a while tho, I think we can do this minor change to behaviour (to make it better aligned with Swing)

faogustavo: Great. I'll change it then

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tooltip.kt
```diff
@@ -245,6 +272,13 @@ public fun rememberPopupPositionProviderAtFixedPosition(
         }
     }
 
+// TODO: When remove it, please replace usages with [TooltipAutoHideBehavior]
```
DanielSouzaBertoldi: Tiny typo!

```suggestion
// TODO: When removing it, please replace usages with [TooltipAutoHideBehavior]
```

faogustavo: Updated. Thanks

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tooltip.kt
faogustavo: Updated the old one with the new behavior

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tooltip.kt
```diff
@@ -145,7 +144,35 @@ public fun Tooltip(
     content: @Composable () -> Unit,
 ) {
     TooltipArea(
-        tooltip = { if (enabled) TooltipImpl(style, tooltip) else Box {} },
+        tooltip = {
+            if (enabled) {
+                var isVisible by remember { mutableStateOf(true) }
```
rock3r: Nit: any chance we can extract this as a private function?

faogustavo: Moved the logic to the `TooltipImpl`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Tooltip.kt
```diff
@@ -154,7 +138,33 @@ public fun Tooltip(
...
+            TooltipAutoHideBehavior.Long -> {
+                delay(style.metrics.fullDisappearDelay)
+                isVisible = false
+            }
+
+            TooltipAutoHideBehavior.Never -> {}
+        }
+    }
+
+    if (!isVisible) return
```
rock3r, line 166: We had this comment for the `!isVisible` case:

```kotlin
// Render an empty box when not visible to allow TooltipArea to dismiss properly
```

Are you sure removing the empty Box does not break anything?

faogustavo: I was not able to find any issues when removing it. Do you remember why exactly it was added? Maybe I'm missing something that is not visible.

# MacOS

## With Box

https://github.com/user-attachments/assets/850d628c-95b7-4416-8848-9ad87cbcee1c


## Without Box

https://github.com/user-attachments/assets/f9df428b-2bb8-4319-8484-3e1864068269

# Windows

## With Box

https://github.com/user-attachments/assets/c1814c6f-7f3f-4330-aa03-6114c6df7b01


## Without Box

https://github.com/user-attachments/assets/cb2f43e8-a246-4641-ba94-3cda3733e74f

rock3r: Ok, seems fine. If we find issues we can restore it :)

rock3r:
Ready to merge

PR #3087, opened 2025-06-06
Jewel - Fix RTL Title-Bar Button Order on Windows

kdroidFilter:
When Compose Desktop runs in RTL (LayoutDirection.Rtl) on Windows, the caption text automatically moves to the right, but the native caption buttons (Minimize / Maximize / Close) remain on the far right. The result is a visually incorrect ordering (Title → Buttons).

This PR ports JetBrains’ upstream patch  into jewel‑decorated‑window and fully restores correct RTL behaviour.

Before : 
צילום מסך 2025-06-06 131359

After : 
Capture d'écran 2025-06-06 184600 s

As a workaround, I simply hide the native controls when RTL is active and reuse the Linux-style window control components, which naturally support RTL layout and behave correctly.

### DIFF
--- a/platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.Linux.kt
+++ b/platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.Linux.kt
@@ -1,10 +1,7 @@
 package org.jetbrains.jewel.window
 
-import androidx.compose.foundation.focusable
 import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.input.pointer.PointerButton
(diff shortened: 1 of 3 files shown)


faogustavo:
Hey there, short update in this one. While I was reviewing, I've found a few issues, like the close color doesn't matching windows (prints below). This is probably because your implementation is using Linux colors.

In any case, I'm checking the possibility to handle this issue directly in the JetBrainsRuntime. I'll try to keep you updated on that.

| RTL | LTR |
| --- | --- |
| image | image

kdroidFilter:
Thank you very much, I also noticed that the same problem exists on Mac but I have not been able to find a solution.
For Windows, I managed to force RTL via JNA in this way : https://github.com/kdroidFilter/Platform-Tools/tree/master/platformtools/rtlwindows/src/jvmMain/kotlin/io/github/kdroidfilter/platformtools/rtlwindows
But when I use it with jewel the result produced is strange, the buttons are positioned at the far left but when I select the cross it selects the button to minimize

https://github.com/user-attachments/assets/a18432f7-6032-4b31-b8c8-e18728c9b40b

Otherwise, this implementation does not respect the colors of Linux either ... in any case neither gnome nor kde

rock3r:
Hey, thanks a lot for the PR. Sorry for the radio silence — we've been looking into this behind the scenes and it looks like the proper way to fix this is in JetBrains Runtime, as this is more of a workaround. @faogustavo is working on it with the JBR team. We'll keep this around until the issue is resolved.

kdroidFilter:
Wow thanks a lot 🤩

faogustavo:
@kdroidFilter the fix on the JBR was merged. Now we just need to wait for the next release.

For now, you can update your pr to set the `controls.rtl` flag in the Windows Titlebar implementation from Jewel, and when the flag gets released, we will have this feature.

```kotlin
// Something like this
titleBar.putProperty("controls.rtl", layoutDirection == LayoutDirection.Rtl)
```

kdroidFilter:
@faogustavo Great, thanks a lot! If I understand correctly, I have to revert my pr, and just add this:

```
// Something like this
titleBar.putProperty("controls.rtl", layoutDirection == LayoutDirection.Rtl)
```

Just, I don't understand why you said "like this", isn't that exactly it?

rock3r:
I will close this PR since the fix was shipped in https://github.com/JetBrains/JetBrainsRuntime/pull/540 and will become available in a JBR build soon. Thanks for reporting the issue and proposing a fix, anyway!

kdroidFilter:
Thanks a lot JB !

rock3r:
We don't work for JetBrains, but we're still happy to help :)

PR #3089, opened 2025-06-07
Improve the handling of Gradle DSL Elements

marchermans:
## TLDR:
- Handle gradle extensions natively, both in navigational actions, as well as during typing previews
- Introduce support for navigating to all kinds of elements of the gradle project classpath (including plugins and gradle itself, as well as provide native support for sources if those exist)

## Changes:
- Support for properly handling a configure block in typing suggestions
- Properly order the extensions and the configure block
- Abstract away how individual extension providers provider their Psi meta information of what extensions are available
- Provide an extension contributor which can provide Psi meta information of extensions which are not on the root, but at least one level in
- Provide support for browsing and navigating in `NamedDomainObjectCollection`, however there are some restrictions when it comes to property mode vs configuration mode, see the comment in `GradleNamedDomainCollectionContributor`
- Rework the project extension provider to use the new abstract extension system
- Expose the required information from the Gradle plugin.

## TODO:
- [ ] Potentially work on tests, I tried writing several but the current testing setup does not suit it very much, and this area of the code seems poorly to not covered.
- [ ] Further research UX improvements that can be made to provide a better experience.


marchermans:
Hello, anything I need to do to get this reviewed?

nskvortsov:
I apologize for the delay. 
Can you give some references to the issues that this pull request intends to fix?

nskvortsov:
Without the tests and references to the fixed issue, it is impossible to evaluate the contribution quality.

PR #3090, opened 2025-06-09
K2 UAST: handle synthetic parameters from Compose compiler plugin

jsjeon:
^KTIJ-34499 fixed


jsjeon:
cc @BartvHelvert

jsjeon:
Like the original addition (https://kotlin.jetbrains.space/p/kotlin/reviews/1249), it's not feasible (AFAIK) to test binary generated with a certain compiler plugin...? Anyhow, the counterpart in Lint CLI is at least tested with the same fix.

jsjeon:
> the counterpart in Lint CLI is at least tested with the same fix.

For the record: https://cs.android.com/android-studio/platform/tools/base/+/81995fcbf6a8b172c00755ee393ccf042d12fd19

BartvHelvert:
Merged, thanks! I'll check whether it is possible for us to add such tests now or maybe make it possible in the future.

PR #3091, opened 2025-06-09
[JEWEL-776] Improvements to Context Menu

DanielSouzaBertoldi:
# Context

Currently our [Context Menu](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-desktop-context-menus.html) has the correct styles, but it's missing the ability to display an icon and shortcut hints as well as handling said shortcuts.

Main changes:

- Overriding `TextContextMenu` so that we can display shortcut hints and handle shortcuts and also shows an icon.
- Created a `ContextMenuItemOption` data class that holds both the context menu item icon and shortcut type.
- Created a `MenuController` interface which will substitute our current `MenuManager` class, which got deprecated. With this interface, we can test the shortcut handling logic.
- Created a sealed class `ContextMenuItemOptionAction` that's used to map to a shortcut hint or the keystroke for the shortcut.
- Created a `selectableItemWithActionType` variant for `MenuScope`. This variation doesn't rely on a `keybinding: Set<String>` but instead a `ContextMenuItemOptionAction`.
- Turned `MenuSelectableItem` into a class. It's public now but only visible for testing.
- Put all related composable logic from MenuItem into a `MenuItemBase`. Added a new variation of `MenuItem` that accepts a `ContextMenuItemOptionAction` instead of a `keybinding: Set<String>`.
- Updated `ContextMenu` code to use our new `MenuController` and delegate key presses to it. If the menu controller can't find a menu shortcut associated with the keys being pressed, then we fallback to the more general purpose `handlePopupMenuOnKeyEvent`.

# Evidences

## Standalone
| Context | Evidence |
|--------|--------|
| `SelectionContainer` light theme |  |
| `SelectionContainer` dark theme |  |
| `SelectionContainer` copy shortcut (clicking on the option/using shortcut/selecting by enter key) | <video src="https://github.com/user-attachments/assets/837fab09-0dfc-437b-89ae-f8f860b511dc" width="400" /> |
| `BasicTextField` light theme |  |
| `BasicTextField` dark theme |  |
| Copy shortcut (directly clicking on the context menu option) | <video src="https://github.com/user-attachments/assets/1423226e-03bb-4a75-9c69-2339ee7f8b0d" width="400" /> |
| Copy shortcut (typing the shortcut without selecting option in context menu) | <video src="https://github.com/user-attachments/assets/14e8b481-9a6c-4671-b571-58e9d3269532" width="400" /> |
| Copy shortcut (selecting option with arrow + enter keys) | <video src="https://github.com/user-attachments/assets/b67c45d2-8fac-4aa4-9902-1c4a5d0fdeba" width="400" /> |
| Paste shortcut | <video src="https://github.com/user-attachments/assets/84e809ed-9d1f-437d-8899-903794b1485a" width="400" /> |
| Cut shortcut | <video src="https://github.com/user-attachments/assets/0332a5cf-fa95-46d2-8e74-0b25d632cc1a" width="400" /> |
| Select All shortcut | <video src="https://github.com/user-attachments/assets/b9004271-1d42-4bfd-b463-a5ad92423f1d" width="400" /> |

## IJP
| Context | Evidence |
|--------|--------|
| `SelectionContainer` light theme |  |
| `SelectionContainer` dark theme |  |
| `SelectionContainer` copy shortcut (clicking on the option/using shortcut/selecting by enter key) | <video src="https://github.com/user-attachments/assets/437fa4ca-8a0b-4185-b1a8-f62da4183faf" width="400" /> |
| `BasicTextField` light theme |  |
| `BasicTextField` dark theme |  |
| Copy shortcut (directly clicking on the context menu option) | <video src="https://github.com/user-attachments/assets/08094b9d-5a99-4ec5-8611-4a1d180c601b" width="400" /> |
| Copy shortcut (typing the shortcut without selecting option in context menu) | <video src="https://github.com/user-attachments/assets/2ed2e4a1-2695-4f1e-8880-453e57059865" width="400" /> |
| Copy shortcut (selecting option with enter key) | <video src="https://github.com/user-attachments/assets/26104f69-d635-4eba-a468-f0ed134827fc" width="400" /> |
| Paste shortcut | <video src="https://github.com/user-attachments/assets/282699c5-1fab-4562-b916-16e77f8bdee7" width="400" /> |
| Cut shortcut | <video src="https://github.com/user-attachments/assets/0083d081-e56b-442f-a812-4fc157407fb5" width="400" /> |
| Select All shortcut | <video src="https://github.com/user-attachments/assets/1f6c1b68-c040-4106-b897-601729668068" width="400" /> |

### New features
 * Created a `ContextMenuItemOption`. You should use this class to provide icons and shortcut data (hint and what keystroke corresponds to said shortcut).

### Deprecated API
 * `MenuManager` is now deprecated. You should use `MenuController` or `BaseMenuController`. This new interface now handles shortcut key presses.

### DIFF
--- a/platform/jewel/samples/standalone/api-dump-unreviewed.txt
+++ b/platform/jewel/samples/standalone/api-dump-unreviewed.txt
@@ -6,6 +6,5 @@ f:org.jetbrains.jewel.samples.standalone.markdown.ComposableSingletons$MarkdownE
 f:org.jetbrains.jewel.samples.standalone.markdown.JewelReadmeKt
 - sf:getJewelReadme():java.lang.String
 f:org.jetbrains.jewel.samples.standalone.markdown.MarkdownEditorKt
-- sf:MarkdownEditor(androidx.compose.foundation.text.input.TextFieldState,androidx.compose.ui.Modifier,androidx.compose.runtime.Composer,I,I):V
 f:org.jetbrains.jewel.samples.standalone.markdown.MarkdownPreviewKt
 - sf:MarkdownPreview(androidx.compose.ui.Modifier,java.lang.CharSequence,androidx.compose.runtime.Composer,I,I):V
--- a/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/MenuContentTest.kt
+++ b/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/MenuContentTest.kt
@@ -1,0 +1,393 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.component
+
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.ui.input.InputMode
+import androidx.compose.ui.test.junit4.createComposeRule
+import javax.swing.KeyStroke
+import org.jetbrains.jewel.intui.standalone.menuShortcut.StandaloneShortcutProvider
+import org.jetbrains.jewel.intui.standalone.theme.IntUiTheme
+import org.jetbrains.jewel.ui.LocalMenuItemShortcutProvider
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+
+// A fake MenuController that records interactions for assertions.
+class FakeMenuController : MenuController {
+    val registeredActions = mutableMapOf<KeyStroke, () -> Unit>()
+    var clearCount = 0
+        private set
+
+    override val onDismissRequest: (InputMode) -> Boolean = { true }
+
+    override fun onHoveredChange(hovered: Boolean) {
+        // DO NOTHING
+    }
+
+    override fun closeAll(mode: InputMode, force: Boolean) {
+        // DO NOTHING
+    }
+
+    override fun close(mode: InputMode): Boolean = true
+
+    override fun isRootMenu(): Boolean = true
+
+    override fun isSubmenu(): Boolean = false
+
+    override fun submenuController(onDismissRequest: (InputMode) -> Boolean): MenuController = this
+
+    override fun registerShortcutAction(keyStroke: KeyStroke, action: () -> Unit) {
+        registeredActions[keyStroke] = action
+    }
+
+    override fun clearShortcutActions() {
+        clearCount++
+        registeredActions.clear()
+    }
+
+    override fun findAndExecuteShortcut(keyStroke: KeyStroke?): Boolean? = null
+}
+
+class MenuContentShortcutRegistrationTest {
+    @get:Rule val rule = createComposeRule()
+
+    private lateinit var fakeMenuController: FakeMenuController
+
+    private val copyAction = ContextMenuItemOptionAction.CopyMenuItemOptionAction
+    private val pasteAction = ContextMenuItemOptionAction.PasteMenuItemOptionAction
+
+    private val copyKeyStroke = StandaloneShortcutProvider.getShortcutKeyStroke(copyAction)!!
+    private val pasteKeyStroke = StandaloneShortcutProvider.getShortcutKeyStroke(pasteAction)!!
+
+    private var triggeredActionType: String? = null
+
+    @Before
+    fun setUp() {
+        fakeMenuController = FakeMenuController()
+        triggeredActionType = null
+    }
+
+    // Helper to create the content lambda for MenuContent
+    private fun createMenuContentLambda(items: List<MenuSelectableItem>): MenuScope.() -> Unit = {
+        items.forEach { item ->
+            selectableItemWithActionType(
+                selected = false,
+                onClick = item.onClick,
+                enabled = item.isEnabled,
+                actionType = item.itemOptionAction,
+                content = item.content,
+            )
+        }
+    }
+
+    @Test
+    fun `reordering items should not mix up shortcut actions`() {
+        val itemsState =
+            mutableStateOf(
+                listOf(
+                    MenuSelectableItem(
+                        itemOptionAction = copyAction,
+                        onClick = { triggeredActionType = "COPY" },
+                        isSelected = true,
+                        isEnabled = true,
+                        iconKey = null,
+                        content = {},
+                    ),
+                    MenuSelectableItem(
+                        itemOptionAction = pasteAction,
+                        onClick = { triggeredActionType = "PASTE" },
+                        isSelected = true,
+                        isEnabled = true,
+                        iconKey = null,
+                        content = {},
+                    ),
+                )
+            )
+
+        rule.setContent {
+            IntUiTheme {
+                CompositionLocalProvider(
+                    LocalMenuController provides fakeMenuController,
+                    LocalMenuItemShortcutProvider provides StandaloneShortcutProvider,
+                ) {
+                    MenuContent(content = createMenuContentLambda(itemsState.value))
+                }
+            }
+        }
+
+        itemsState.value = itemsState.value.reversed()
+        rule.waitForIdle()
+
+        assertTrue("Clear should have been called when items reordered", fakeMenuController.clearCount >= 1)
+        assertEquals(2, fakeMenuController.registeredActions.size)
+
+        fakeMenuController.registeredActions[copyKeyStroke]?.invoke()
+        assertEquals("Copy KeyStroke must still trigger the copy action", "COPY", triggeredActionType)
+
+        fakeMenuController.registeredActions[pasteKeyStroke]?.invoke()
+        assertEquals("Paste KeyStroke must still trigger the paste action", "PASTE", triggeredActionType)
+    }
+
+    @Test
+    fun `initial items should register their shortcuts`() {
+        val initialItems =
+            listOf(
+                MenuSelectableItem(
+                    itemOptionAction = copyAction,
+                    onClick = { triggeredActionType = "COPY" },
+                    isSelected = true,
+                    isEnabled = true,
+                    iconKey = null,
+                    content = {},
+                ),
+                MenuSelectableItem(
+                    itemOptionAction = pasteAction,
+                    onClick = { triggeredActionType = "PASTE" },
+                    isSelected = true,
+                    isEnabled = true,
+                    iconKey = null,
+                    content = {},
+                ),
+            )
+
+        rule.setContent {
+            IntUiTheme {
+                CompositionLocalProvider(
+                    LocalMenuController provides fakeMenuController,
+                    LocalMenuItemShortcutProvider provides StandaloneShortcutProvider,
+                ) {
+                    MenuContent(content = createMenuContentLambda(initialItems))
+                }
+            }
+        }
+
+        assertEquals("Clear should NOT be called on first composition", 0, fakeMenuController.clearCount)
+        assertEquals(2, fakeMenuController.registeredActions.size)
+        assertNotNull(fakeMenuController.registeredActions[copyKeyStroke])
+
+        fakeMenuController.registeredActions[copyKeyStroke]?.invoke()
+        assertEquals("COPY", triggeredActionType)
+    }
+
+    @Test
+    fun `disabled items should not have their shortcuts registered`() {
+        val items =
+            listOf(
+                MenuSelectableItem(
+                    itemOptionAction = copyAction,
+                    onClick = {},
+                    isSelected = true,
+                    isEnabled = true,
+                    iconKey = null,
+                    content = {},
+                ),
+                MenuSelectableItem(
+                    itemOptionAction = pasteAction,
+                    onClick = {},
+                    isEnabled = false, // This one is disabled
+                    isSelected = true,
+                    iconKey = null,
+                    content = {},
+                ),
+            )
+
+        rule.setContent {
+            IntUiTheme {
+                CompositionLocalProvider(
+                    LocalMenuController provides fakeMenuController,
+                    LocalMenuItemShortcutProvider provides StandaloneShortcutProvider,
+                ) {
+                    MenuContent(content = createMenuContentLambda(items))
+                }
+            }
+        }
+
+        assertEquals(
+            "Only enabled items should have shortcuts registered",
+            1,
+            fakeMenuController.registeredActions.size,
+        )
+        assertNotNull("Enabled item's shortcut should be present", fakeMenuController.registeredActions[copyKeyStroke])
+        assertNull("Disabled item's shortcut should be absent", fakeMenuController.registeredActions[pasteKeyStroke])
+    }
+
+    @Test
+    fun `items without an actionType should not have shortcuts registered`() {
+        val items =
+            listOf(
+                MenuSelectableItem(
+                    itemOptionAction = copyAction,
+                    onClick = {},
+                    isSelected = true,
+                    isEnabled = true,
+                    iconKey = null,
+                    content = {},
+                ),
+                MenuSelectableItem(
+                    itemOptionAction = null, // No action type
+                    onClick = {},
+                    isSelected = true,
+                    isEnabled = true,
+                    iconKey = null,
+                    content = {},
+                ),
+            )
+
+        rule.setContent {
+            IntUiTheme {
+                CompositionLocalProvider(
+                    LocalMenuController provides fakeMenuController,
+                    LocalMenuItemShortcutProvider provides StandaloneShortcutProvider,
+                ) {
+                    MenuContent(content = createMenuContentLambda(items))
+                }
+            }
+        }
+
+        assertEquals("Only items with actionType should have shortcuts", 1, fakeMenuController.registeredActions.size)
+        assertNotNull(fakeMenuController.registeredActions[copyKeyStroke])
+        assertEquals(false, fakeMenuController.registeredActions.containsKey(pasteKeyStroke)) // Just to be explicit
+    }
+
+    @Test
+    fun `empty list of items should result in no registered shortcuts`() {
+        val itemsState =
+            mutableStateOf(
+                listOf(
+                    MenuSelectableItem(
+                        itemOptionAction = copyAction,
+                        onClick = {},
+                        isSelected = true,
+                        isEnabled = true,
+                        iconKey = null,
+                        content = {},
+                    )
+                )
+            )
+        rule.setContent {
+            IntUiTheme {
+                CompositionLocalProvider(
+                    LocalMenuController provides fakeMenuController,
+                    LocalMenuItemShortcutProvider provides StandaloneShortcutProvider,
+                ) {
+                    MenuContent(content = createMenuContentLambda(itemsState.value))
+                }
+            }
+        }
+        assertEquals("Should start with one registered shortcut", 1, fakeMenuController.registeredActions.size)
+
+        itemsState.value = emptyList()
+        rule.waitForIdle()
+
+        assertTrue("Clear should have been called when list became empty", fakeMenuController.clearCount == 1)
+        assertTrue(
+            "No shortcuts should be registered for an empty list",
+            fakeMenuController.registeredActions.isEmpty(),
+        )
+    }
+
+    @Test
+    fun `toggling an item's enabled state should re-register shortcuts`() {
+        val itemsState =
+            mutableStateOf(
+                listOf(
+                    MenuSelectableItem(
+                        itemOptionAction = copyAction,
+                        onClick = {},
+                        isSelected = true,
+                        isEnabled = true,
+                        iconKey = null,
+                        content = {},
+                    )
+                )
+            )
+        rule.setContent {
+            IntUiTheme {
+                CompositionLocalProvider(
+                    LocalMenuController provides fakeMenuController,
+                    LocalMenuItemShortcutProvider provides StandaloneShortcutProvider,
+                ) {
+                    MenuContent(content = createMenuContentLambda(itemsState.value))
+                }
+            }
+        }
+        assertEquals("Enabled item should be registered initially", 1, fakeMenuController.registeredActions.size)
+
+        itemsState.value =
+            listOf(
+                MenuSelectableItem(
+                    itemOptionAction = copyAction,
+                    onClick = {},
+                    isSelected = true,
+                    isEnabled = false, // Toggled state
+                    iconKey = null,
+                    content = {},
+                )
+            )
+        rule.waitForIdle()
+
+        assertTrue("Clear should be called again when item is disabled", fakeMenuController.clearCount == 1)
+        assertTrue("Disabled item should not be registered", fakeMenuController.registeredActions.isEmpty())
+
+        itemsState.value =
+            listOf(
+                MenuSelectableItem(
+                    itemOptionAction = copyAction,
+                    onClick = {},
+                    isSelected = true,
+                    isEnabled = true, // Re-enabled state
+                    iconKey = null,
+                    content = {},
+                )
+            )
+        rule.waitForIdle()
+
+        assertTrue("Clear should be called when item is re-enabled", fakeMenuController.clearCount == 2)
+        assertEquals("Re-enabled item should be registered again", 1, fakeMenuController.registeredActions.size)
+        assertNotNull(fakeMenuController.registeredActions[copyKeyStroke])
+    }
+
+    @Test
+    fun `changing the menu manager should re-register shortcuts`() {
+        val items =
+            listOf(
+                MenuSelectableItem(
+                    itemOptionAction = copyAction,
+                    onClick = {},
+                    isSelected = true,
+                    isEnabled = true,
+                    iconKey = null,
+                    content = {},
+                )
+            )
+        val currentMenuController = mutableStateOf<MenuController>(fakeMenuController)
+
+        rule.setContent {
+            IntUiTheme {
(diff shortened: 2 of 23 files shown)


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
```diff
@@ -61,7 +70,10 @@ internal fun ContextMenu(
         onKeyEvent = {
             val currentFocusManager = checkNotNull(focusManager) { "FocusManager must not be null" }
             val currentInputModeManager = checkNotNull(inputModeManager) { "InputModeManager must not be null" }
-            handlePopupMenuOnKeyEvent(it, currentFocusManager, currentInputModeManager, menuManager)
+            val swingKeyStroke = composeKeyEventToSwingKeyStroke(it)
+
+            menuManager.findAndExecuteShortcut(swingKeyStroke)
+                ?: handlePopupMenuOnKeyEvent(it, currentFocusManager, currentInputModeManager, menuManager)
```
DanielSouzaBertoldi: So, the idea here is that context menus have its own especial set of shortcuts, not just the basic ones handled by `handlePopupMenuOnKeyEvent`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/MenuItemShortcutProvider.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Gets the KeyStroke for a given action identifier.
+     *
+     * @param contextMenuItemActionType A type of action to be performed (e.g., ContextMenuItemActionType.COPY).
+     */
+    public fun getShortcutKeyStroke(contextMenuItemActionType: ContextMenuItemActionType): KeyStroke?
+}
+
+public object CommonShortcutHintProvider : MenuItemShortcutProvider {
```
DanielSouzaBertoldi: This is a common implementation that for now will be provided to both IJP and standalone versions [since currently we can't override the default actions](https://issuetracker.google.com/issues/422966784) such as copy/paste/cut from Compose. That means we can't actually tell compose to use the same shortcut in the user's keymap for these actions.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -160,13 +162,30 @@ internal fun MenuContent(
...
+                localMenuItemShortcutProvider.getShortcutKeyStroke(item.actionType)?.let { keyStroke ->
+                    localMenuManager.registerShortcutAction(keyStroke) {
+                        item.onClick()
+                        localMenuManager.closeAll(localInputModeManager.inputMode, true)
+                    }
+                }
+            }
+        }
+        onDispose { localMenuManager.clearShortcutActions() }
+    }
```
DanielSouzaBertoldi: My line of thought for this effect was as follows:

1. The `Menu` composable know about all its items.
2. We need to run a specific shortcut item when a `ContextMenu` is being displayed;

So, we just go through all the current items in this `Menu`, and find the Swing KeyStroke representation of the shortcut. If found, we add to a list of shortcuts in the `MenuManager`. If eiter the items, menu manager or the provider changes, we clear the shortcut actions and add the new ones.

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/MarkdownEditor.kt
```diff
@@ -31,14 +31,18 @@ import org.jetbrains.jewel.ui.component.Text
 import org.jetbrains.jewel.ui.component.TextArea
 
 @Composable
-public fun MarkdownEditor(state: TextFieldState, modifier: Modifier = Modifier) {
+public fun MarkdownEditor(
+    state: TextFieldValue,
+    modifier: Modifier = Modifier,
+    onValueChanged: (TextFieldValue) -> Unit,
```
DanielSouzaBertoldi: We have to go back to using the value-based version of `BasicTextField` since [it's the only one that displays a Context Menu currently](https://youtrack.jetbrains.com/issue/CMP-8332/State-based-BasicTextField-is-missing-context-menu-on-desktop).

However this kinda goes against what we state in the Kdoc of our value-based version of `TextArea`:

```
* A multi-line text input component that follows the standard visual styling.
* > **Note:** you should prefer the state-based variant whenever possible. When the `TextFieldState`-based variant will
* > reach feature parity with this variant, we'll deprecate this variant and, after some time, remove it.
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -506,11 +531,16 @@ internal fun MenuItem(
...
+                        if (actionType != null) {
+                            val shortcutProvider = LocalMenuItemShortcut.current
+                            shortcutProvider.getShortcutHint(actionType)
+                        } else {
+                            remember(keybinding) {
+                                if (hostOs.isMacOS) {
+                                    keybinding?.joinToString("") { it }.orEmpty()
+                                } else {
+                                    keybinding?.joinToString("+") { it }.orEmpty()
+                                }
```
DanielSouzaBertoldi: Mapping an enum to a specific Swing KeyStroke representation is way easier than having to "juggle" with a `Set<String>`.

However, now we have two parameters that have the same intent: displaying a shortcut hint. Is important to note that if a user uses the `keybinding` parameter, then the actual shortcut handling won't happen with these code changes. It'll be visual only.

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
```diff
@@ -151,9 +151,10 @@ private fun RowScope.ColumnOne() {
         }
 
         Row(horizontalArrangement = Arrangement.spacedBy(16.dp)) {
-            val state = rememberTextFieldState("")
+            var state by remember { mutableStateOf(TextFieldValue("")) }
```
rock3r: Please revert this change — we should still prefer state-based BTFs even if their context menu does not currently work. It's up to CMP to fix this :)

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/MarkdownEditor.kt
rock3r: As stated above, please revert this. We should indeed prefer State-based BTF. It was ok for a temporary tweak to test this change, but should not be committed.

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/MarkdownEditor.kt
```diff
@@ -83,10 +87,11 @@ private fun ControlsRow(modifier: Modifier = Modifier, onLoadMarkdown: (String)
 }
 
 @Composable
-private fun Editor(state: TextFieldState, modifier: Modifier = Modifier) {
+private fun Editor(state: TextFieldValue, modifier: Modifier = Modifier, onValueChanged: (TextFieldValue) -> Unit) {
```
rock3r: Ditto

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/MarkdownView.kt
```diff
@@ -21,8 +25,8 @@ import org.jetbrains.jewel.ui.component.Divider
 internal fun MarkdownDemo() {
     Row(Modifier.trackActivation().fillMaxSize().background(JewelTheme.globalColors.panelBackground)) {
         WithMarkdownMode(MarkdownMode.EditorPreview(scrollingSynchronizer = null)) {
-            val editorState = rememberTextFieldState(JewelReadme)
-            MarkdownEditor(state = editorState, modifier = Modifier.fillMaxHeight().weight(1f))
+            var editorState by remember { mutableStateOf(TextFieldValue(JewelReadme)) }
```
rock3r: Ditto

platform/jewel/ui/api-dump.txt
```diff
@@ -13,14 +24,35 @@ e:org.jetbrains.jewel.ui.component.AutoHideBehavior
...
+- sf:$stable:I
+- <init>(org.jetbrains.jewel.ui.icon.IconKey,org.jetbrains.jewel.ui.component.ContextMenuItemActionType,java.lang.String,kotlin.jvm.functions.Function0):V
+- b:<init>(org.jetbrains.jewel.ui.icon.IconKey,org.jetbrains.jewel.ui.component.ContextMenuItemActionType,java.lang.String,kotlin.jvm.functions.Function0,I,kotlin.jvm.internal.DefaultConstructorMarker):V
+- f:getActionType():org.jetbrains.jewel.ui.component.ContextMenuItemActionType
+- f:getIcon():org.jetbrains.jewel.ui.icon.IconKey
 org.jetbrains.jewel.ui.component.MenuScope
 - a:passiveItem(kotlin.jvm.functions.Function2):V
-- a:selectableItem(Z,org.jetbrains.jewel.ui.icon.IconKey,java.util.Set,kotlin.jvm.functions.Function0,Z,kotlin.jvm.functions.Function2):V
-- bs:selectableItem$default(org.jetbrains.jewel.ui.component.MenuScope,Z,org.jetbrains.jewel.ui.icon.IconKey,java.util.Set,kotlin.jvm.functions.Function0,Z,kotlin.jvm.functions.Function2,I,java.lang.Object):V
+- a:selectableItem(Z,org.jetbrains.jewel.ui.icon.IconKey,org.jetbrains.jewel.ui.component.ContextMenuItemActionType,java.util.Set,kotlin.jvm.functions.Function0,Z,kotlin.jvm.functions.Function2):V
```
rock3r: Can you make sure this is not a breaking change? E.g. by adding a new overload and deprecating the existing one

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/MenuItemShortcutProvider.kt
```diff
@@ -0,0 +1,86 @@
...
+/** Provides a way to also get the actual KeyStroke of a shortcut besides its hint */
+public interface MenuItemShortcutProvider {
+    /**
+     * Gets the formatted shortcut string for the given action identifier.
+     *
+     * @param contextMenuItemActionType The action type (e.g., "Copy" or "Paste").
+     * @return The human-readable shortcut string (e.g., "⌘C", "Ctrl + S"), or empty if no shortcut is defined or it
+     *   shouldn't be displayed.
+     */
+    public fun getShortcutHint(contextMenuItemActionType: ContextMenuItemActionType): String
```
rock3r: This should be done differently in Standalone and in Bridge. In the latter, you should only delegate to `com.intellij.openapi.keymap.KeymapUtil#getShortcutText`, ensuring we render shortcuts the same way as the IDE. In Standalone, you can use the logic you have below.

I would also split away this responsibility (rendering a shortcut to text) to a separate component, that is a composition local, and is used by the menu component itself. This way we keep a clear distinction between the shortcut itself, and its text representation (and the logic converting the former to the latter).

Something like this would live in `ui`:

```kotlin
public interface ShortcutRenderer {
  fun render(keyStroke: KeyStroke)
}

public val LocalShortcutRenderer = ...
```

And then have distinct implementations in bridge and standalone. This way we can reuse this mechanism for other use cases.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/MenuItemShortcutProvider.kt
```diff
@@ -0,0 +1,86 @@
...
+        return getKeyStrokeText(keyStroke)
+    }
+
+    override fun getShortcutKeyStroke(contextMenuItemActionType: ContextMenuItemActionType): KeyStroke? =
+        labelToShortcutMap[contextMenuItemActionType]
+}
+
+public val LocalMenuItemShortcut: ProvidableCompositionLocal<MenuItemShortcutProvider> = staticCompositionLocalOf {
+    error("No LocalMenuItemShortcut provided. Have you forgotten the theme?")
+}
```
rock3r:
```suggestion
public val LocalMenuItemShortcutProvider: ProvidableCompositionLocal<MenuItemShortcutProvider> = staticCompositionLocalOf {
    error("No LocalMenuItemShortcutProvider provided. Have you forgotten the theme?")
}
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
rock3r: Have you tested adding/removing menu items at runtime to ensure this works correctly?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
rock3r: This sounds like a problem waiting to happen. Can we have a separate menu item API with the new actionType and deprecate the old one, keeping the old one's functionality as much as possible?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
```diff
@@ -93,3 +116,29 @@ public object ContextMenuDivider : ContextMenuItem("---", {})
...
     ContextMenuItem(label, {})
+
+public class ContextMenuOption(
+    public val icon: IconKey? = null,
+    public val actionType: ContextMenuItemActionType,
+    label: String,
+    action: () -> Unit,
+) : ContextMenuItem(label, action)
+
+public enum class ContextMenuItemActionType {
```
rock3r: I am a bit worried that this is putting us in a corner. What if in the future we want to add something "Copy as plain text" or "Paste as plain text"? Would this be possible to do without breaking changes?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextContextMenu.kt
```diff
@@ -0,0 +1,57 @@
...
+                    ContextMenuOption(
+                        actionType = ContextMenuItemActionType.SELECT_ALL,
+                        label = localization.selectAll,
+                        action = it,
+                    )
+                },
+            )
+        }
+
+        TextContextMenuArea(textManager = textManager, items = items, state = state, content = content)
```
rock3r: Nit: DRY

```suggestion
        TextContextMenuArea(textManager, items, state, content)
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextContextMenu.kt
```diff
@@ -0,0 +1,57 @@
...
+
+public object TextContextMenu : TextContextMenu {
+    @Composable
+    override fun Area(
+        textManager: TextContextMenu.TextManager,
+        state: ContextMenuState,
+        content: @Composable (() -> Unit),
+    ) {
+        val localization = LocalLocalization.current
+        val items = {
```
rock3r: Is this meant to be recreated at every recomposition?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: Great idea! I'll test this scenario and make necessary changes if needed!

DanielSouzaBertoldi: Adding/removing works as normal :)

| Standalone | Plugin |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/f9e845e4-06b9-4f34-8b13-d17bd5a904c8" width="400" /> | <video src="https://github.com/user-attachments/assets/7d984c73-70ac-4890-bd75-d270e49185a6" width="400" /> |

rock3r: I meant that their actions do not get mixed up :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextContextMenu.kt
DanielSouzaBertoldi: Great observation! I don't think it's necessary to keep recreating the list, thinking about it now. TextManager and ContextMenuState won't be changing frequently, we can remember them :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
DanielSouzaBertoldi: Fair enough. I changed to a sealed class, this way we get more leverage if we need to add and extend current actions without breaking changes and ensures we handle all cases properly :) let me know your thoughts!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -449,7 +450,7 @@ private interface MenuItem {
     val content: @Composable () -> Unit
 }
 
-private data class MenuSelectableItem(
+public data class MenuSelectableItem(
```
rock3r: We can't have public data classes. Please switch to `@GenerateDataFunctions` :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuManagerImpl.kt
```diff
@@ -5,10 +5,33 @@ import androidx.compose.runtime.staticCompositionLocalOf
...
+
+    // Methods for internal shortcut handling
+    public fun registerShortcutAction(keyStroke: KeyStroke, action: () -> Unit)
+
+    public fun clearShortcutActions()
+
+    public fun findAndExecuteShortcut(keyStroke: KeyStroke?): Boolean?
+}
+
+public class MenuManagerImpl(
```
rock3r: Shouldn't this be at least `@InternalJewelApi`? I'd also call it something like `DefaultMenuManager` to avoid the `Impl` suffix

DanielSouzaBertoldi: Yeah. Actually I was wondering if I should leave the current MenuManager class and mark it as @Deprecated since it's a public class or if it's ok to just use this new version instead.

I'm still not sure when it's necessary to add Deprecated calls or when it's ok to not mark as deprecated and just change stuff.

In other words: is it a breaking change or not?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: Added ui tests just for that. Thanks for the tip!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuManagerImpl.kt
rock3r: Removing any public API or changing its signature is considered a breaking change and should be avoided. If you see any removed line from the api dump, that's a sign of a breaking change, unless it's only appending an optional parameter to an existing function.

So:

>I was wondering if I should leave the current MenuManager class and mark it as @deprecated since it's a public class

Yes, you should :)

platform/jewel/samples/standalone/api-dump-unreviewed.txt
```diff
@@ -6,6 +6,5 @@ f:org.jetbrains.jewel.samples.standalone.markdown.ComposableSingletons$MarkdownE
 f:org.jetbrains.jewel.samples.standalone.markdown.JewelReadmeKt
 - sf:getJewelReadme():java.lang.String
 f:org.jetbrains.jewel.samples.standalone.markdown.MarkdownEditorKt
-- sf:MarkdownEditor(androidx.compose.foundation.text.input.TextFieldState,androidx.compose.ui.Modifier,androidx.compose.runtime.Composer,I,I):V
```
DanielSouzaBertoldi, line 9: Not sure why this was moved to `api-dump` but ok

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/shortcut/StandaloneShortcutProvider.kt
```diff
@@ -0,0 +1,34 @@
...
+import java.awt.event.KeyEvent
+import javax.swing.KeyStroke
+import org.jetbrains.jewel.ui.MenuItemShortcutProvider
+import org.jetbrains.jewel.ui.component.ContextMenuItemOptionAction
+import org.jetbrains.jewel.ui.component.ContextMenuItemOptionAction.CopyMenuItemOptionAction
+import org.jetbrains.jewel.ui.component.ContextMenuItemOptionAction.PasteMenuItemOptionAction
+import org.jetbrains.jewel.ui.component.ContextMenuItemOptionAction.SelectAllMenuItemOptionAction
+import org.jetbrains.skiko.hostOs
+
+public object StandaloneShortcutProvider : MenuItemShortcutProvider {
```
rock3r: I'd make this internal, no reason to expose new APIs for the impls :) The interface is enough for users.

```suggestion
internal object StandaloneMenuItemShortcutProvider : MenuItemShortcutProvider {
```

Same in bridge

platform/jewel/samples/standalone/api-dump-unreviewed.txt
rock3r: The ways of the IJP API dumps are mysterious!

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/MenuContentTest.kt
```diff
@@ -0,0 +1,393 @@
...
+
+    private lateinit var fakeMenuController: FakeMenuController
+
+    private val copyAction = ContextMenuItemOptionAction.CopyMenuItemOptionAction
+    private val pasteAction = ContextMenuItemOptionAction.PasteMenuItemOptionAction
+
+    private val copyKeyStroke = StandaloneShortcutProvider.getShortcutKeyStroke(copyAction)!!
+    private val pasteKeyStroke = StandaloneShortcutProvider.getShortcutKeyStroke(pasteAction)!!
+
+    private var triggeredActionType: String? = null
```
faogustavo, line 67: Nit: Any reason to have this as a class attribute instead of in the function you need it?

faogustavo (approved):
Looking good. Just a minor question.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuManager.kt
```diff
@@ -3,7 +3,16 @@ package org.jetbrains.jewel.ui.component
...
 import androidx.compose.ui.input.InputMode
+import javax.swing.KeyStroke
+import org.jetbrains.jewel.foundation.InternalJewelApi
 
+@Deprecated(
+    message =
+        "The MenuManager class has been superseded by the MenuController interface, " +
+            "which offers improved abstraction and new functionalities like shortcut management. " +
+            "Depend on the DefaultMenuController class for all menu operations.",
+    replaceWith = ReplaceWith("DefaultMenuController"),
```
faogustavo: Does this "replace with" work? I thought it needed the full package + class to work

DanielSouzaBertoldi: You're right, I had to add the full package and class.

I tried declaring this as:

```
replaceWith = ReplaceWith(
    expression = "DefaultMenuController",
    imports = ["org.jetbrains.jewel.ui.component.DefaultMenuController"],
)
```

But it didn't seem to work too. Weird.

https://github.com/user-attachments/assets/959dc3e1-d8af-41ed-ad82-50d32df3f533

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/MenuContentTest.kt
DanielSouzaBertoldi: I mean, besides it serving as the primary mechanism for each test to observe which action's `onClick` lambda was invoked, not really. I'm clearing the stored value in the `setUp()` function too.

I could change it so that I would declare this variable in each and every single test, but this way it's less boilerplate-y in my opinion. What do you think?

faogustavo: I'd say to keep this in the scope of each test. But no problem with keeping here for now

jakub-senohrabek:
Ready to merge

PR #3094, opened 2025-06-12
[JEWEL-843] Make disabled variant of icons look darker in dark mode

DanielSouzaBertoldi:
# Context

Currently, our disabled icons are really bright compared to Swing's counterpart. That's because we are always adding an arbitrary value to emulate the brightness that Swing applies to icons through its non-linear calculations. So each and every icon will have the same brightness. This PR fixes this so that we don't apply any brightness to an icon if the current theme is set to dark.

## Code changed
- Changed `disabledColorMatrixGammaEncoded` to a `fun` so we can set a specific brightness value given the current theme.
- Created a new `.disabledThemeAware()` Composable function that checks if the current theme
- Added `@Deprecated` annotation to current `disabled()` function since that one doesn't take into account the user's theme.
- Added new neat swing comparisons examples!

### Investigation
<details><summary>If you'd like to learn more about how IJP (IntelliJ Platform) actually greys out their icons to look disabled, check this collapsed section!</summary>
<p>
Whenever a plugin needs to fetch a disabled version of an icon, they typically call the `getDisabledIcon(icon: Icon)` function from `IconLoader`. This code down the line fetch the greyed out icon from this variable:

```kotlin
private val iconToDisabledIcon: LoadingCache<Icon, Icon> = Caffeine.newBuilder()
  .maximumSize(1024)
  .executor(Dispatchers.Default.asExecutor())
  .expireAfterAccess(10.minutes.toJavaDuration())
  .build<Icon, Icon>(CacheLoader { icon ->
    if (icon is CachedImageIcon) {
      icon.createWithFilter(standardDisablingFilter)
    }
    else {
      FilteredIcon(baseIcon = icon, filterSupplier = standardDisablingFilter)
    }
  })
  .also {
    registerIconCacheCleaner(it::invalidateAll)
  }
```

That, as you can see, uses the `standardDisablingFilter` and applies the filter to the specific icon. This variable calls the filter from `UIUtil.getGrayFilter()`:

```java
public static @NotNull RGBImageFilter getGrayFilter() {
  return GrayFilter.namedFilter("grayFilter", new GrayFilter(33, -35, 100));
}
```

Passing the values 33 as `brightness`, -35 as `contrast` and 100 as `opacity`.

The `GrayFilter` class has a function `filterRGB` that, given the `brightness`, `contrast` and `opacity` passed to it, does some non-linear calculations:

```java
public int filterRGB(int x, int y, int rgb) {
  // Use NTSC conversion formula.
  int gray = (int)(0.30 * (rgb >> 16 & 0xff) +
                   0.59 * (rgb >> 8 & 0xff) +
                   0.11 * (rgb & 0xff));

  if (brightness >= 0) {
    gray = (int)((gray + brightness * 255) / (1 + brightness));
  }
  else {
    gray = (int)(gray / (1 - brightness));
  }

  if (contrast >= 0) {
    if (gray >= 127) {
      gray = (int)(gray + (255 - gray) * contrast);
    }
    else {
      gray = (int)(gray - gray * contrast);
    }
  }
  else {
    gray = (int)(127 + (gray - 127) * (contrast + 1));
  }

  int a = ((rgb >> 24) & 0xff) * alpha / 100;

  return (a << 24) | (gray << 16) | (gray << 8) | gray;
}
```

Unfortunately, `ColorMatrix` from Compose doesn't support non-linear calculations yet. In other words, we can't faithfully recreate this code right now. The best we can do is get an approximation, but as you can see from the evidences below, some icons (even though they can have the exact same hex color) may appear a little brighter or darker.

So, for now, this will have to suffice.

</p>
</details> 

# Evidences

## Dark Mode
| Before | After |
|--------|--------|
| image | image |

## Light Mode
| Before | After |
|--------|--------|
| image | image |

## Swing Comparison Tab
| Light Mode | Dark Mode |
|--------|--------|
| image | image |
> Honestly, it's really hard to horizontally align content from two different rows in Swing

## Dynamically switching between themes
| IDE | Standalone| 
|---|---|
| <video src="https://github.com/user-attachments/assets/dcbeb4f0-b290-49c5-8693-a6ba8bcb2333" width="400" /> | <video src="https://github.com/user-attachments/assets/9058a941-31ad-47a3-a65f-532fc8ae52d7" width="400" /> |

### DIFF
--- a/platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
+++ b/platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
@@ -1,42 +1,27 @@
 package org.jetbrains.jewel.samples.ideplugin
 
 import androidx.compose.foundation.border
-import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.calculateEndPadding
-import androidx.compose.foundation.layout.calculateStartPadding
-import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.text.input.rememberTextFieldState
-import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableIntStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
+import androidx.compose.runtime.*
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.unit.dp
+import com.intellij.icons.AllIcons
 import com.intellij.ide.ui.laf.darcula.ui.DarculaButtonUI
 import com.intellij.openapi.editor.colors.EditorFontType
 import com.intellij.openapi.ui.ComboBox
+import com.intellij.openapi.util.IconLoader
 import com.intellij.ui.JBColor
+import com.intellij.ui.components.BrowserLink
 import com.intellij.ui.components.JBLabel
 import com.intellij.ui.components.JBScrollPane
-import com.intellij.ui.dsl.builder.AlignY
-import com.intellij.ui.dsl.builder.COLUMNS_SHORT
-import com.intellij.ui.dsl.builder.Panel
-import com.intellij.ui.dsl.builder.Row
-import com.intellij.ui.dsl.builder.RowLayout
-import com.intellij.ui.dsl.builder.panel
+import com.intellij.ui.dsl.builder.*
 import com.intellij.util.ui.JBFont
 import com.intellij.util.ui.JBUI
 import com.intellij.util.ui.components.BorderLayoutPanel
@@ -50,15 +35,9 @@ import org.jetbrains.jewel.bridge.JewelComposePanel
 import org.jetbrains.jewel.bridge.medium
 import org.jetbrains.jewel.bridge.retrieveEditorColorScheme
 import org.jetbrains.jewel.foundation.theme.JewelTheme
-import org.jetbrains.jewel.ui.component.DefaultButton
-import org.jetbrains.jewel.ui.component.EditableListComboBox
-import org.jetbrains.jewel.ui.component.Icon
-import org.jetbrains.jewel.ui.component.ListComboBox
-import org.jetbrains.jewel.ui.component.OutlinedButton
-import org.jetbrains.jewel.ui.component.Text
-import org.jetbrains.jewel.ui.component.TextArea
-import org.jetbrains.jewel.ui.component.TextField
-import org.jetbrains.jewel.ui.component.Typography
+import org.jetbrains.jewel.ui.component.*
+import org.jetbrains.jewel.ui.disabledThemeAware
+import org.jetbrains.jewel.ui.icons.AllIconsKeys
 import org.jetbrains.jewel.ui.theme.textAreaStyle
 
 internal class SwingComparisonTabPanel : BorderLayoutPanel() {
@@ -71,6 +50,8 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
                 separator()
                 iconsRow()
                 separator()
+                linksRow()
+                separator()
                 textFieldsRow()
                 separator()
                 textAreasRow()
@@ -82,6 +63,28 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
                 border = JBUI.Borders.empty(0, 10)
                 isOpaque = false
             }
+
+    private fun Panel.linksRow() {
+        row("Links:") {
+                browserLink("Enabled link", "")
+
+                compose { ExternalLink(text = "Enabled link", enabled = true, onClick = {}) }
+
+                cell(
+                        component =
+                            BrowserLink(
+                                icon = IconLoader.getDisabledIcon(AllIcons.Ide.External_link_arrow),
+                                text = "Disabled link",
+                                tooltip = null,
+                                url = "",
+                            )
+                    )
+                    .applyToComponent { foreground = JBUI.CurrentTheme.Link.Foreground.DISABLED }
+
+                compose { ExternalLink(text = "Disabled Link", enabled = false, onClick = {}) }
+            }
+            .layout(RowLayout.PARENT_GRID)
+    }
 
     private val scrollingContainer =
         JBScrollPane(
@@ -204,6 +207,59 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
                         contentDescription = "Jewel Tool Window Icon",
                         modifier = Modifier.border(1.dp, Color.Red),
                     )
+                }
+
+                panel {
+                    row {
+                        label("Swing").widthGroup("swing disabled")
+                        label("Compose").widthGroup("compose disabled")
+                        label("Swing").widthGroup("swing enabled")
+                        label("Compose").widthGroup("compose enabled")
+
+                        label("Swing").widthGroup("swing disabled")
+                        label("Compose").widthGroup("compose disabled")
+                        label("Swing").widthGroup("swing enabled")
+                        label("Compose").widthGroup("compose enabled")
+                    }
+                    row {
+                        icon(icon = IconLoader.getDisabledIcon(AllIcons.Actions.CheckOut)).widthGroup("swing disabled")
+                        compose {
+                                Icon(
+                                    key = AllIconsKeys.Actions.CheckOut,
+                                    contentDescription = null,
+                                    colorFilter = ColorFilter.disabledThemeAware(),
+                                )
+                            }
+                            .widthGroup("compose disabled")
+                        icon(icon = AllIcons.Actions.CheckOut).widthGroup("swing enabled")
+                        compose {
+                                Icon(
+                                    key = AllIconsKeys.Actions.CheckOut,
+                                    contentDescription = null,
+                                    colorFilter = null, // Not disabled
+                                )
+                            }
+                            .widthGroup("compose enabled")
+
+                        icon(icon = IconLoader.getDisabledIcon(AllIcons.Actions.Close)).widthGroup("swing disabled")
+                        compose {
+                                Icon(
+                                    key = AllIconsKeys.Actions.Close,
+                                    contentDescription = null,
+                                    colorFilter = ColorFilter.disabledThemeAware(),
+                                )
+                            }
+                            .widthGroup("compose disabled")
+                        icon(icon = AllIcons.Actions.Close).widthGroup("swing enabled")
+                        compose {
+                                Icon(
+                                    key = AllIconsKeys.Actions.Close,
+                                    contentDescription = null,
+                                    colorFilter = null, // Not disabled
+                                )
+                            }
+                            .widthGroup("compose enabled")
+                    }
                 }
             }
             .layout(RowLayout.PARENT_GRID)
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
@@ -1,20 +1,23 @@
 package org.jetbrains.jewel.ui
 
+import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.graphics.ColorMatrix
+import org.jetbrains.jewel.foundation.theme.JewelTheme
 
 // Implements javax.swing.GrayFilter's behaviour with percent = 50, brighter = true
 // to match the GrayFilter#createDisabledImage behavior, used by Swing.
-private val disabledColorMatrixGammaEncoded =
+private fun disabledColorMatrixGammaEncoded(isSystemInDarkMode: Boolean = false) =
     ColorMatrix().apply {
         val saturation = .5f
+        val brightness = if (isSystemInDarkMode) .125f else .25f
 
         // We use NTSC luminance weights like Swing does as it's gamma-encoded RGB,
         // and add some brightness to emulate Swing's "brighter" approach, which is
         // not representable with a ColorMatrix alone as it's a non-linear op.
-        val redFactor = .299f * saturation + .25f
-        val greenFactor = .587f * saturation + .25f
-        val blueFactor = .114f * saturation + .25f
+        val redFactor = .299f * saturation + brightness
+        val greenFactor = .587f * saturation + brightness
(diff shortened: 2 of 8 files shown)


platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
```diff
@@ -83,6 +64,28 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
...
+                cell(
+                        component =
+                            BrowserLink(
+                                icon = IconLoader.getDisabledIcon(AllIcons.Ide.External_link_arrow),
+                                text = "Disabled link",
+                                tooltip = null,
+                                url = "",
+                            )
+                    )
+                    .applyToComponent { foreground = JBUI.CurrentTheme.Link.Foreground.DISABLED }
```
DanielSouzaBertoldi: This only changes the foreground color but the hover behaviour remains. If I try to add `action = null` or `enabled = false` Swing will straight up just make the component invisible, which sucks.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
```diff
@@ -26,4 +29,9 @@ private val disabledColorMatrixGammaEncoded =
         this[2, 2] = blueFactor
     }
 
-public fun ColorFilter.Companion.disabled(): ColorFilter = colorMatrix(disabledColorMatrixGammaEncoded)
+@Deprecated("Use disabledThemeAware instead. This way you get nice colors on dark theme too!")
+public fun ColorFilter.Companion.disabled(): ColorFilter = colorMatrix(disabledColorMatrixGammaEncoded())
+
+@Composable
+public fun ColorFilter.Companion.disabledThemeAware(): ColorFilter =
+    colorMatrix(disabledColorMatrixGammaEncoded(JewelTheme.isDark))
```
DanielSouzaBertoldi: Name suggestions are greatly appreaciated 🙏🏻

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
```diff
@@ -1,42 +1,27 @@
...
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.calculateEndPadding
-import androidx.compose.foundation.layout.calculateStartPadding
-import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.*
```
rock3r: Please don't use star imports. I recommend installing the Detekt IDE plugin and configuring it as such:

image

Just make sure you don't commit `.idea/detekt.xml`

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
rock3r: You can avoid that behaviour by setting `autoHideOnDisable = false` (it is true by default)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
```diff
@@ -1,20 +1,23 @@
...
 
+import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.graphics.ColorMatrix
+import org.jetbrains.jewel.foundation.theme.JewelTheme
 
 // Implements javax.swing.GrayFilter's behaviour with percent = 50, brighter = true
 // to match the GrayFilter#createDisabledImage behavior, used by Swing.
-private val disabledColorMatrixGammaEncoded =
+private fun disabledColorMatrixGammaEncoded(isSystemInDarkMode: Boolean = false) =
```
rock3r: Nit: `isDark` for consistency with the rest of Jewel

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
rock3r:
```suggestion
public fun ColorFilter.Companion.disabledIcon(isDark: Boolean = JewelTheme.isDark): ColorFilter =
    colorMatrix(disabledColorMatrixGammaEncoded(isDark))
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
```diff
@@ -26,4 +29,9 @@ private val disabledColorMatrixGammaEncoded =
         this[2, 2] = blueFactor
     }
 
-public fun ColorFilter.Companion.disabled(): ColorFilter = colorMatrix(disabledColorMatrixGammaEncoded)
+@Deprecated("Use disabledThemeAware instead. This way you get nice colors on dark theme too!")
```
rock3r:
```suggestion
@Deprecated(
  message = "Use disabledIcon() instead to get correct behaviour in dark themes too.",
  replaceWith = ReplaceWith("ColorFilter.Companion.disabledIcon()")
)
```

(please test that the automatic replacement works)

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
DanielSouzaBertoldi: Oof, it slipped right past me.

I also changed the IDE to avoid optimizing imports with wildcard whenever I hit ctl+alt+O

image

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
DanielSouzaBertoldi: Had to change the ReplaceWith string with `ColorFilter.disabledIcon()`, it's working as expected:


https://github.com/user-attachments/assets/90a56eff-7b69-4adf-a3a6-66e9e001ce17

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
DanielSouzaBertoldi: Awesome! Thanks, didn't know about this flag

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
rock3r: You'd need to change the Kotlin rules tho, not the Java ones :)

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/BridgeGrayFilterValues.kt
```diff
@@ -0,0 +1,18 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge.theme
+
+import com.intellij.util.ui.GrayFilter
+import com.intellij.util.ui.UIUtil
+import org.jetbrains.jewel.foundation.GrayFilterValues
+
+public fun GrayFilterValues.Companion.readFromLaF(isDark: Boolean): GrayFilterValues {
```
rock3r: In the bridge you should not need to bring `isDark` with you. You can directly read the value from `org.jetbrains.jewel.bridge.theme.IntUiBridgeKt#isDark`

PS: Add KDoc please

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/GrayFilterValues.kt
```diff
@@ -0,0 +1,40 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.foundation
+
+import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+
+@Immutable
+@GenerateDataFunctions
+public class GrayFilterValues(public val brightness: Int, public val contrast: Int, public val alpha: Int) {
```
rock3r: Could you add KDocs and bound checks in `init{}` to this class?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/theme/IntUiGrayFilterValues.kt
```diff
@@ -0,0 +1,16 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.intui.standalone.theme
+
+import org.jetbrains.jewel.foundation.GrayFilterValues
+
+public fun GrayFilterValues.Companion.light(
```
rock3r: Can you add KDoc to these two?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
```diff
@@ -26,4 +40,89 @@ private val disabledColorMatrixGammaEncoded =
         this[2, 2] = blueFactor
     }
 
+@Deprecated("Use Modifier.grayFilter() instead to get correct behaviour in dark themes too.")
 public fun ColorFilter.Companion.disabled(): ColorFilter = colorMatrix(disabledColorMatrixGammaEncoded)
+
+@Composable
+public fun Modifier.grayFilter(): Modifier {
```
rock3r: I'd move this to its own file (same package) and add KDoc

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
```diff
@@ -26,4 +40,89 @@ private val disabledColorMatrixGammaEncoded =
...
+    val grayFilterValues = LocalGrayFilterValues.current
+
+    val (brightness, contrast, alpha) =
+        arrayOf(grayFilterValues.brightness, grayFilterValues.contrast, grayFilterValues.alpha)
```
rock3r: I'd change these vals to parameters and read the defaults directly in there

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
```diff
@@ -26,4 +40,89 @@ private val disabledColorMatrixGammaEncoded =
...
+
+    require(brightness in -100..100) { "The brightness must be in [-100, 100], but was $brightness" }
+    require(contrast in -100..100) { "The contrast must be in [-100, 100], but was $contrast" }
+    require(alpha in 0..100) { "The alpha must be in [0, 100], but was $alpha" }
+
+    val effect = remember { RuntimeEffect.makeForShader(GRAY_FILTER_SKSL_FOR_IMAGEFILTER) }
+    val paint = remember { SkiaPaint() }
+
+    return this.drawWithContent {
```
rock3r: Nit:

```suggestion
    return drawWithContent {
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
```diff
@@ -138,12 +138,10 @@ public fun Dropdown(
                 contentAlignment = Alignment.Center,
             ) {
                 val alpha = if (dropdownState.isEnabled) 1f else 0.5f
-                val colorFilter = if (dropdownState.isEnabled) null else ColorFilter.disabled()
                 Icon(
-                    modifier = Modifier.alpha(alpha),
+                    modifier = Modifier.alpha(alpha).thenIf(dropdownState.isEnabled) { grayFilter() },
```
rock3r: Shouldn't this be `!dropdownState.isEnabled`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -331,11 +330,11 @@ public fun ListComboBox(
             itemKeys = itemKeys,
             itemContent = { item, isSelected, isActive ->
                 SimpleListItem(
+                    modifier = Modifier.thenIf(enabled) { grayFilter() },
```
rock3r: Shouldn't this be `!enabled`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DisabledColorFilter.kt
rock3r: I'm also pondering if `grayFilter` is the best name. What about something like:

```kotlin
fun Modifier.disabledAppearance(...) : Modifier {
  //...
}
```

We need to make it clear that this is only a visual treatment and doesn’t _really_ disable anything, in the KDoc. Users still have to pass the `enabled` flag around to the relevant places. I prefer this name as it both makes it clear it's about _looking disabled_ and not a generic grayness filter. We should add a simple usage example in the KDoc. Something like:

```kotlin
@Composable
fun SendButton(enabled: Boolean, modifier: Modifier = Modifier) {
  IconActionButton(
    key = MyIconKeys.Actions.Send,
    contentDescription = "Send",
    onClick = { ... },
    enabled = enabled,
    iconModifier = Modifier.thenIf(!enabled) { disabledAppearance() }
  )
}
```

DanielSouzaBertoldi: Yeah, `disabledAppearance`sounds fine to me :)

rock3r: Gemini's suggestion :P

rock3r:
Some styling to fix:

```
ExpressionBodySyntax - [Functions with exact one statement, the return statement, can be rewritten with (...)] at /home/runner/work/intellij-community/intellij-community/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeDefinition.kt:64:9
MaxLineLength - [Line detected, which is longer than the defined maximum line length in the code (...)] at /home/runner/work/intellij-community/intellij-community/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeDefinition.kt:64:1
ExpressionBodySyntax - [Functions with exact one statement, the return statement, can be rewritten with (...)] at /home/runner/work/intellij-community/intellij-community/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/DisabledAppearanceValues.kt:55:9
```

You can run the `./gradlew check` task locally to see if there are any errors

rock3r (approved):

rock3r:
Ready to merge

PR #3095, opened 2025-06-17
KT UAST: resolve annotation on setter parameter

jsjeon:
To do so, for parameter without source PSI (i.e., synthetic),
we keep nullable annotations from java PSI,
while restoring user-given annotations on source.

^KTIJ-34365 fixed


jsjeon:
cc @BartvHelvert

BartvHelvert:
Thanks! Just merged.

PR #3097, opened 2025-06-18
[JEWEL-464] Removing checkmark icons from int-ui-standalone

DanielSouzaBertoldi:
# Context

Since the creation of `IconKey` API, there's no need for us to keep IJ icons in the project to use them in the standalone version, except both checkbox and radio button since there's not an easy way currently to get them from IntelliJ Platform.

Apart from checkboxes and radio button icons, there was only a checkmark icon leftover in the project. It's not even referenced in both `IntUiLightTheme` and `IntUiDarkTheme`. The selected checkbox icon variant already has a checkmark, so yeah.


rock3r:
@DanielSouzaBertoldi I think you can remove the release notes entirely — this is an under-the-hood change and an implementation detail. You only need to add release notes when there are user-visible changes.

faogustavo (approved):

rock3r:
Ready to merge

PR #3099, opened 2025-06-19
[JEWEL] IJPL-190270 Rename Jewel -> Compose

jakub-senohrabek:
Jewel word was removed from public api, I kept backward compatibility by either a typealias, or wrapper function.

Not sure if renaming `BasicJewelUiTest` was needed.

Have I missed something?

### DIFF
--- a/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/JewelTheme.kt
+++ b/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/JewelTheme.kt
@@ -16,6 +16,8 @@ import org.jetbrains.jewel.foundation.LocalDisabledAppearanceValues
 import org.jetbrains.jewel.foundation.LocalGlobalColors
 import org.jetbrains.jewel.foundation.LocalGlobalMetrics
 
+public typealias ComposeTheme = JewelTheme
+
 public interface JewelTheme {
     public companion object {
         public val name: String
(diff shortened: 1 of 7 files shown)


platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/JewelTheme.kt
```diff
@@ -52,11 +54,21 @@ public interface JewelTheme {
     }
 }
 
+@Composable
+public fun ComposeTheme(theme: ThemeDefinition, swingCompatMode: Boolean, content: @Composable () -> Unit) {
```
rock3r: Considering these are lower-level themes that users will not be using directly, I do not think we need to alias these. Users will only ever use `BridgeTheme` and `IntUiTheme`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/theme/JewelTheme.kt
```diff
@@ -173,11 +173,28 @@ public val JewelTheme.Companion.iconButtonStyle: IconButtonStyle
 public val JewelTheme.Companion.sliderStyle: SliderStyle
     @Composable @ReadOnlyComposable get() = LocalSliderStyle.current
 
+// Alias for BaseJewelTheme
+@Composable
```
rock3r: Ditto here — users will not be using any `*JewelTheme()` composables directly; I do not think we need to alias these, they're internal implementation details. You can annotate them as such if you think that's necessary, but I would not introduce an incredibly confusing and generic `ComposeTheme` that people don't need to use anyway.

platform/jewel/ui/src/test/kotlin/org/jetbrains/jewel/BasicJewelUiTest.kt
```diff
@@ -6,6 +6,8 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import kotlinx.coroutines.runBlocking
 import org.junit.Rule
 
+public typealias BasicComposeUiTest = BasicJewelUiTest
```
rock3r: This is not needed; tests do not get published to other modules.

rock3r:
Please add the release notes section to the PR description as described on Slack :)

rock3r:
Adding Gustavo in my stead while I am off

rock3r:
@jakub-senohrabek why was this shipped with no approval?

PR #3100, opened 2025-06-21
Use XDG Desktop Portal to sync light/dark mode on Linux

iso2013:
This uses the XDG Desktop Portal spec for system settings, now standardized by freedesktop: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html#org-freedesktop-portal-settings-read Specifically, it uses the `org.freedesktop.appearance` `color-scheme` setting to detect whether the user's OS is set to prefer dark mode.

To do this, it uses the `dbus-send` and `dbus-monitor` commands, part of the `dbus-bin` package. I'm sure there are Java-DBus interop APIs that could do this more elegantly, but I figured it wasn't worth adding a new dependency for what ends up being a pretty simple check.

(I also do not typically work in Kotlin; please let me know if there are ways I can improve my formatting or execution here, and I will gladly do so.)

### DIFF
--- a/platform/platform-impl/src/com/intellij/ide/ui/laf/SystemDarkThemeDetector.kt
+++ b/platform/platform-impl/src/com/intellij/ide/ui/laf/SystemDarkThemeDetector.kt
@@ -15,6 +15,8 @@ import com.sun.jna.platform.win32.WinReg
 import org.jetbrains.annotations.NonNls
 import java.awt.Toolkit
 import java.beans.PropertyChangeEvent
+import java.io.BufferedReader
+import java.io.InputStreamReader
 import java.util.Locale
 import java.util.function.BiConsumer
 import java.util.function.Consumer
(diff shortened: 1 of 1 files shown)


kilab:
@aydarin can You check this PR?

louwers:
Does this also enable the toggle in the Settings?

pporvatov:
Hi @iso2013 

Thank you very much for your fix. I reviewed it and found several problems like:
* Deprecated API is used
* The exec command can be invoked from the EDT thread
* e.printStackTrace is used
* syncFunction.accept can be invoked twice while one theme switch

I reworked the fix and pushed it into the repo:
https://github.com/JetBrains/intellij-community/commit/bc8dab7a56f35f2ec3d88a3e4b71186a4fea8e6c

Regards, Pavel

pporvatov:
The fix is committed and tested

PR #3101, opened 2025-06-23
[JEWEL-833] Make IntelliJThemeGenerator use theme from IJP

DanielSouzaBertoldi:
# Context

Now that Jewel is part of Intellij Community repo, we don't need to fetch the values from both `expUI_light` and `expUI_dark` files remotely. This PR modifies the code so that we can fetch these files contents directly, using a relative path.

### Bonus!

I also took this opportunity to make the `intelliJThemeGenerator` task cacheable.

Main changes:

- Changed the `themeFile` property to a `RegularFileProperty` instead of a `String`
- Updated both `register` calls to use the correct relative path to both files.
- Changed the `IntelliJThemeGeneratorTask` class to be an `abstract class` instead of `open`. Also changed the class properties to `abstract`, otherwise we wouldn't be able to take advantage of cacheable tasks.
   - This change was made according to Gradle docs, mainly [implementing custom tasks](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:implementing_tasks) and [Declaring Inputs and Outputs via Annotations](https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_output_annotations)

# Evidences

| Running Gradle Task | Generated File Header |
|--------|--------|
| image | image |

> Notice that the "Source" comment now points to the entire relative path!


platform/jewel/buildSrc/src/main/kotlin/intellij-theme-generator.gradle.kts
```diff
@@ -28,7 +28,7 @@ extension.all {
             outputFile = targetDir.file(paths)
             themeClassName = this@all.themeClassName
             ideaVersion = this@all.ideaVersion
-            themeFile = this@all.themeFile
+            themeFile = project.layout.file(this@all.themeFile.map { rootProject.file(it) })
```
DanielSouzaBertoldi: This code runs during the configuration phase, so it's safe to use project.layout and map the relative `themeFile` path to a `RegularFileProperty`. If we attempted to resolve the file path inside the `IntelliJThemeGeneratorTask` (i.e., during the execution phase), Gradle would emit warnings and eventually fail (starting with Gradle 10), since accessing project state at execution time breaks configuration cache compatibility. For cacheable tasks to work properly, all inputs must be fully declared and resolved during configuration.

platform/jewel/int-ui/int-ui-standalone/build.gradle.kts
```diff
@@ -13,12 +13,12 @@ intelliJThemeGenerator {
...
         themeClassName = "org.jetbrains.jewel.intui.core.theme.IntUiLightTheme"
-        themeFile = "platform/platform-resources/src/themes/expUI/expUI_light.theme.json"
+        themeFile = "../../platform/platform-resources/src/themes/expUI/expUI_light.theme.json"
         ideaVersion = targetIdeaVersion
     }
     register("intUiDark") {
         themeClassName = "org.jetbrains.jewel.intui.core.theme.IntUiDarkTheme"
-        themeFile = "platform/platform-resources/src/themes/expUI/expUI_dark.theme.json"
+        themeFile = "../../platform/platform-resources/src/themes/expUI/expUI_dark.theme.json"
         ideaVersion = targetIdeaVersion
```
faogustavo: Nit: If we keep the idea version, we should also get it automatically from the project. The var is only used to create a header comment

image

faogustavo: We could probably remove the source and version. As we get from the local build path, we don't need that info anymore

DanielSouzaBertoldi: Good point! Since Jewel is now part of the intellij-community project, I think it's not really relevant to keep the IDEA version. However, keeping the source comment just makes it easier to figure out if the file we are using to generate the class is stale or not when debugging. I kept it for now, but changed it so that it displays the same relative path declared by both `register` tasks!

What do you think?

faogustavo: Looking good 👌

faogustavo (approved):

jakub-senohrabek:
Ready to merge

PR #3104, opened 2025-06-25
PY-82262: remove usage of noexcept due to no jetbrains ide support

lattwood:
This PR closes PY-82262 by removing the usage of the `noexcept` Cython keyword which is presently unsupported by JetBrains products. (See: https://youtrack.jetbrains.com/issue/PY-59249/Latest-Cython-3.0.0b1-except-noexcept-legal-syntax-red-lined for an explanation why)

This probably creates a performance regression, but the code can now be edited by a JetBrains product.


vladimir-koshelev:
The issue about noexpect support is fixed, so these changes are no longer needed.

PR #3105, opened 2025-06-25
[JEWEL-97] Render popups using the JBPopup component on IDE

faogustavo:
- Created `JewelConfigs` to enable/disable this new feature
- Created `PopupRender` interface to support different ways to handle popups
- Created `Popup` component to use the render based on the flag
- Updated our sample code to use the new popup component instead of the compose version -

# Evidences

### Compose Popup

https://github.com/user-attachments/assets/81f2eaa9-6c3a-4c9e-9543-2ff64a43acff

### JBPopup

https://github.com/user-attachments/assets/7ee63d6d-9089-4852-90f1-3c583e019404

### IDE Zoom

https://github.com/user-attachments/assets/5fbdfe1f-8ee8-497e-bd38-090c9da5ca21

### Zero-delay

https://github.com/user-attachments/assets/1950e747-3e21-4fa7-a205-1c26b4feeeae

### Secondary screen

https://github.com/user-attachments/assets/2ab36345-5d37-45a8-84ec-d2a6bc03da1f

### Secondary screen + IDE Zoom

https://github.com/user-attachments/assets/825068dd-4b75-46bb-a953-10238efbef0b

### ⚠️ Important Changes
 * **JEWEL-97** - Created JewelFlags object that controls how a few components should behave
 * **JEWEL-97** - Created 'useCustomPopupRenderer' flag to allow using external popup implementations

### New features
 * **JEWEL-97** - Added support to use JBPopup API for Popups, Tooltips and Menus. This allows popups to grow over the composable area if needed

### DIFF
--- a/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelFlags.kt
+++ b/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelFlags.kt
@@ -1,0 +1,27 @@
+package org.jetbrains.jewel.foundation
+
+/**
+ * JewelFlags is an object that holds configuration flags used in the Jewel library.
+ *
+ * These flags can control specific behaviors or enable experimental features within Jewel.
+ */
+public object JewelFlags {
+    /**
+     * Enable custom popups handling in Jewel. The default value is `false`.
+     *
+     * If enabled, the Jewel library will use a custom popup renderer, using separate windows instead of being drawn
+     * onto the same layer.
+     *
+     * This is an experimental feature and may not be fully stable. When enabled, Compose's popup settings are ignored
+     * when using Jewel popups and tooltips. This means that setting `compose.layers.type` will have no effect on Jewel
+     * popups and tooltips.
+     *
+     * To set this flag, you can also set the system property `jewel.customPopupRender` to `true`/`false`, or pass the
+     * `-Djewel.customPopupRender=[true|false]` argument when running your application.
+     *
+     * Note that this flag affects popups, menus and tooltips rendering from Jewel Components. It does not affect
+     * `Dialog`s.
+     */
+    @ExperimentalJewelApi
+    public var useCustomPopupRenderer: Boolean = System.getProperty("jewel.customPopupRender", "false").toBoolean()
+}
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
@@ -9,14 +9,17 @@ import com.intellij.openapi.actionSystem.DataSink
 import com.intellij.openapi.actionSystem.UiDataProvider
 import com.intellij.openapi.application.PathManager
 import java.awt.BorderLayout
+import java.awt.Component
 import java.io.File
 import javax.swing.JComponent
 import javax.swing.JPanel
 import org.jetbrains.jewel.bridge.actionSystem.ComponentDataProviderBridge
+import org.jetbrains.jewel.bridge.component.JBPopupRenderer
 import org.jetbrains.jewel.bridge.theme.SwingBridgeTheme
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.foundation.InternalJewelApi
 import org.jetbrains.jewel.foundation.util.JewelLogger
+import org.jetbrains.jewel.ui.component.LocalPopupRenderer
 
 public fun compose(config: ComposePanel.() -> Unit = {}, content: @Composable () -> Unit): JComponent =
     JewelComposePanel(config, content)
@@ -27,7 +30,10 @@ public fun JewelComposePanel(config: ComposePanel.() -> Unit = {}, content: @Com
         config()
         setContent {
             SwingBridgeTheme {
-                CompositionLocalProvider(LocalComponent provides this@createJewelComposePanel) {
+                CompositionLocalProvider(
+                    LocalComponent provides this@createJewelComposePanel,
+                    LocalPopupRenderer provides JBPopupRenderer,
+                ) {
                     ComponentDataProviderBridge(jewelPanel, content = content)
                 }
             }
@@ -43,7 +49,10 @@ public fun JewelToolWindowComposePanel(
     config()
     setContent {
         SwingBridgeTheme {
-            CompositionLocalProvider(LocalComponent provides this@createJewelComposePanel) {
+            CompositionLocalProvider(
+                LocalComponent provides this@createJewelComposePanel,
+                LocalPopupRenderer provides JBPopupRenderer,
+            ) {
                 ComponentDataProviderBridge(jewelPanel, content = content)
             }
         }
@@ -61,7 +70,10 @@ public fun JewelComposeNoThemePanel(config: ComposePanel.() -> Unit = {}, conten
     createJewelComposePanel { jewelPanel ->
         config()
         setContent {
-            CompositionLocalProvider(LocalComponent provides this@createJewelComposePanel) {
+            CompositionLocalProvider(
+                LocalComponent provides this@createJewelComposePanel,
+                LocalPopupRenderer provides JBPopupRenderer,
+            ) {
                 ComponentDataProviderBridge(jewelPanel, content = content)
             }
         }
@@ -82,7 +94,10 @@ public fun JewelToolWindowNoThemeComposePanel(
 ): JComponent = createJewelComposePanel { jewelPanel ->
     config()
     setContent {
-        CompositionLocalProvider(LocalComponent provides this@createJewelComposePanel) {
+        CompositionLocalProvider(
+            LocalComponent provides this@createJewelComposePanel,
+            LocalPopupRenderer provides JBPopupRenderer,
+        ) {
             ComponentDataProviderBridge(jewelPanel, content = content)
         }
     }
@@ -108,6 +123,23 @@ private fun createJewelComposePanel(config: ComposePanel.(JewelComposePanelWrapp
 
 internal class JewelComposePanelWrapper : JPanel(), UiDataProvider {
     internal var targetProvider: UiDataProvider? = null
+
+    val composePanel: ComposePanel
+        get() =
+            components.singleOrNull() as? ComposePanel
+                ?: error("JewelComposePanelWrapper was not initialized with a ComposePanel")
+
+    override fun addImpl(comp: Component, constraints: Any?, index: Int) {
+        require(components.isEmpty()) {
+            "JewelComposePanelWrapper can only contain a single ComposePanel, attempt to add another component"
+        }
+
+        require(comp is ComposePanel) {
+            "JewelComposePanelWrapper can only contain ComposePanel, attempt to add ${comp::class.java.name}"
+        }
+
+        super.addImpl(comp, constraints, index)
+    }
 
     override fun uiDataSnapshot(sink: DataSink) {
         targetProvider?.uiDataSnapshot(sink)
--- a/platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindowFactory.kt
+++ b/platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindowFactory.kt
@@ -14,12 +14,16 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import org.jetbrains.jewel.bridge.addComposeTab
+import org.jetbrains.jewel.foundation.JewelFlags
 import org.jetbrains.jewel.samples.ideplugin.releasessample.ReleasesSampleCompose
 
 @Suppress("unused")
 @ExperimentalCoroutinesApi
 internal class JewelDemoToolWindowFactory : ToolWindowFactory, DumbAware {
     override fun createToolWindowContent(project: Project, toolWindow: ToolWindow) {
+        // Enable custom popup rendering to use JBPopup instead of default Compose implementation
+        JewelFlags.useCustomPopupRenderer = true
+
         toolWindow.addComposeTab("Components") { ComponentShowcaseTab(project) }
 
         toolWindow.addComposeTab("Releases Demo") { ReleasesSampleCompose(project) }
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
@@ -1,7 +1,10 @@
 package org.jetbrains.jewel.ui.component
 
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.focus.FocusDirection
 import androidx.compose.ui.focus.FocusManager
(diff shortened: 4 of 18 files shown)


platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelConfigs.kt
```diff
@@ -0,0 +1,7 @@
+package org.jetbrains.jewel.foundation
+
+@InternalJewelApi
+public object JewelConfigs {
+    public val useCustomPopupRender: Boolean
+        get() = System.getProperty("jewel.customPopupRender", "false").toBoolean()
```
faogustavo: After chatting with Seb last week, we considered adding "flags" for experimental features. This idea is the same thing as what Compose does. But we can change to something else that makes more sense. I'm open to ideas.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRender.kt
```diff
@@ -0,0 +1,181 @@
...
+
+internal object JBPopupRender : PopupRender {
+    @Composable
+    override fun Popup(arguments: Arguments, content: @Composable () -> Unit) {
+        JBPopupLayout(arguments, content)
+    }
+}
+
+@Composable
+private fun JBPopupLayout(arguments: Arguments, content: @Composable () -> Unit) {
```
faogustavo: This composable was heavily inspired by the current Popup component. Unfortunatelly, we don't have access to the same methods (as most of them are internal), so I'm using the ComposePannel for it.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRender.kt
```diff
@@ -0,0 +1,181 @@
...
+            },
+        measurePolicy = { _, _ -> layout(0, 0) {} },
+    )
+
+    val (component, popup) =
+        remember {
+            val composePanel =
+                ComposePanel().apply {
+                    // Setting initial dimensions as 1x1 to calculate the size
+                    preferredSize = Dimension(1, 1)
```
faogustavo: I'm not sure if there is a better API to measure the compose ahead of time for rendering correctly. Please let me know if there are any alternatives

faogustavo:
If we go this route, we'll probably also need to copy the "Tooltip Area" to allow it to be used in tooltips

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRender.kt
```diff
@@ -0,0 +1,181 @@
...
+import androidx.compose.runtime.*
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.awt.ComposePanel
+import androidx.compose.ui.input.key.KeyEvent
+import androidx.compose.ui.layout.*
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.LocalWindowInfo
+import androidx.compose.ui.semantics.popup
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.unit.*
```
DanielSouzaBertoldi: Please avoid star imports 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -19,6 +22,67 @@ import androidx.compose.ui.unit.IntRect
...
+    )
+
+    public companion object
+}
+
+private object DefaultPopupRender : PopupRender {
+    @Composable
+    override fun Popup(arguments: Arguments, content: @Composable () -> Unit) {
+        with(arguments) {
+            androidx.compose.ui.window.Popup(
```
DanielSouzaBertoldi: What do you think of using typealias instead of referencing this class with the package name?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -19,6 +22,67 @@ import androidx.compose.ui.unit.IntRect
...
+    )
+}
+
+@InternalJewelApi
+public val LocalPopupRender: ProvidableCompositionLocal<PopupRender> = staticCompositionLocalOf { DefaultPopupRender }
+
+public interface PopupRender {
+    @Composable public fun Popup(arguments: Arguments, content: @Composable () -> Unit)
+
+    public class Arguments(
```
DanielSouzaBertoldi:
```suggestion
    public class PopupData(
```
Maybe? Arguments is kinda too generic for me 🤔

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
```diff
@@ -13,18 +13,23 @@ import java.io.File
...
 @Suppress("ktlint:standard:function-naming", "FunctionName") // Swing to Compose bridge API
 public fun JewelComposePanel(config: ComposePanel.() -> Unit = {}, content: @Composable () -> Unit): JComponent =
     createJewelComposePanel { jewelPanel ->
         config()
         setContent {
             SwingBridgeTheme {
-                CompositionLocalProvider(LocalComponent provides this@createJewelComposePanel) {
+                CompositionLocalProvider(
+                    LocalComponent provides this@createJewelComposePanel,
+                    LocalPopupRender provides JBPopupRender,
```
DanielSouzaBertoldi: We could add this line to `SwingBridgeTheme`, since it doesn't need context from `createJewelComposePanel`

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRender.kt
DanielSouzaBertoldi: Couldn't find anything "better" myself either... Seems like this is the only way 😞

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
faogustavo: Putting in the `SwingBridgeTheme` would not have effect on the `JewelComposeNoThemePanel` and `JewelToolWindowNoThemeComposePanel`. As it's also not theme-related, but to the behavior, I'd prefer to keep it in this function.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TooltipArea.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.jetbrains.jewel.foundation.JewelConfigs
+
+@ExperimentalJewelApi
+@Composable
+public fun TooltipArea(
+    tooltip: @Composable () -> Unit,
+    modifier: Modifier = Modifier,
+    delayMillis: Int = 500,
+    tooltipPlacement: TooltipPlacement = TooltipPlacement.CursorPoint(offset = DpOffset(0.dp, 16.dp)),
+    content: @Composable () -> Unit,
```
DanielSouzaBertoldi, line 55: Could you please add KDocs to this and the composable below?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,69 @@ import androidx.compose.ui.unit.IntOffset
...
+
+@Composable
+@InternalJewelApi
+public fun Popup(
+    popupPositionProvider: PopupPositionProvider,
+    onDismissRequest: (() -> Unit)? = null,
+    properties: PopupProperties = PopupProperties(),
+    onPreviewKeyEvent: ((KeyEvent) -> Boolean)? = null,
+    onKeyEvent: ((KeyEvent) -> Boolean)? = null,
+    content: @Composable () -> Unit,
```
DanielSouzaBertoldi, line 58: Could you please add Kdoc to this Composable, please?

DanielSouzaBertoldi (approved):
Awesome work! 😎

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
faogustavo: updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TooltipArea.kt
faogustavo: Updated

rock3r:
@faogustavo please add release notes :)

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelConfigs.kt
```diff
@@ -0,0 +1,7 @@
+package org.jetbrains.jewel.foundation
+
+@InternalJewelApi
+public object JewelConfigs {
```
rock3r: I'd call this either `JewelConfig` or, maybe better, `JewelFlags`. I also think we do not need to make it internal (users are still supposed to be able to use this, even if with caution). Then we want to annotate single flags as experimental (we may have non-experimental APIs over time).

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelConfigs.kt
```diff
@@ -0,0 +1,7 @@
...
+    public val useCustomPopupRender: Boolean
```
rock3r: Please add KDoc to this property explaining what it does, and how to enable it from the command line too.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+                onDismissRequest = onDismissRequest,
+                properties = properties,
+                onPreviewKeyEvent = onPreviewKeyEvent,
+                onKeyEvent = onKeyEvent,
+            ),
+        content = content,
+    )
+}
+
+public interface PopupRender {
```
rock3r: `PopupRenderer`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+    @Composable public fun Popup(arguments: PopupData, content: @Composable () -> Unit)
```
rock3r:
```suggestion
    @Composable public fun Popup(popupData: PopupData, content: @Composable () -> Unit)
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+
+    public class PopupData(
```
rock3r: How come we put them in a class and not as parameters on `Popup`? Also, this needs the generate data class functions annotation, and equals/hashcode/tostring

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+        public val properties: PopupProperties = PopupProperties(),
+        public val onPreviewKeyEvent: ((KeyEvent) -> Boolean)? = null,
+        public val onKeyEvent: ((KeyEvent) -> Boolean)? = null,
+    )
+
+    public companion object
+}
+
+@InternalJewelApi
+public val LocalPopupRender: ProvidableCompositionLocal<PopupRender> = staticCompositionLocalOf { DefaultPopupRender }
```
rock3r: `LocalPopupRenderer`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+
+private object DefaultPopupRender : PopupRender {
```
rock3r: `DefaultPopupRenderer`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+@InternalJewelApi
+public fun Popup(
+    popupPositionProvider: PopupPositionProvider,
+    onDismissRequest: (() -> Unit)? = null,
+    properties: PopupProperties = PopupProperties(),
+    onPreviewKeyEvent: ((KeyEvent) -> Boolean)? = null,
+    onKeyEvent: ((KeyEvent) -> Boolean)? = null,
+    content: @Composable () -> Unit,
+) {
+    val popupRender = if (JewelConfigs.useCustomPopupRender) LocalPopupRender.current else DefaultPopupRender
```
rock3r: `popupRender`

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelConfigs.kt
rock3r:
```suggestion
    @ExperimentalJewelApi
    public val useCustomPopupRenderer: Boolean
        get() = System.getProperty("jewel.customPopupRenderer", "false").toBoolean()
```

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindowFactory.kt
```diff
@@ -20,6 +20,8 @@ import org.jetbrains.jewel.samples.ideplugin.releasessample.ReleasesSampleCompos
 @ExperimentalCoroutinesApi
 internal class JewelDemoToolWindowFactory : ToolWindowFactory, DumbAware {
     override fun createToolWindowContent(project: Project, toolWindow: ToolWindow) {
+        System.setProperty("jewel.customPopupRender", "true")
```
rock3r: Why not add a setter to the flags API? So we can avoid the stringly-typed API typical of system props.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
rock3r: Also, please add KDocs :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+    )
+}
+
+public interface PopupRender {
+    @Composable public fun Popup(arguments: PopupData, content: @Composable () -> Unit)
+
+    public class PopupData(
+        public val popupPositionProvider: PopupPositionProvider,
+        public val onDismissRequest: (() -> Unit)? = null,
+        public val properties: PopupProperties = PopupProperties(),
```
rock3r: Nit: put all lambdas together

```suggestion
        public val properties: PopupProperties = PopupProperties(),
        public val onDismissRequest: (() -> Unit)? = null,
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
rock3r: Also, KDoc plz :D

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,89 @@ import androidx.compose.ui.unit.IntOffset
...
+
+/**
+ * Displays a popup with the provided content at a position determined by the given [PopupPositionProvider].
+ *
+ * This function behavior is influenced by the 'jewel.customPopupRender' system property. If set to `true`, it allows
+ * using a custom popup rendering implementation; otherwise, it defaults to the standard Compose popup.
+ *
+ * If running on the IntelliJ Platform and setting the 'jewel.customPopupRender' property to `true`, the plugin will use
+ * the JBPopup implementation for rendering popups. This is useful if your composable content is small and you need to
+ * display a popup that is bigger than the component size.
```
rock3r: I'd refer to the flags API here, instead of the raw property. Once we add KDocs to the flags API, people can refer to the one place where it's written.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TooltipArea.kt
```diff
@@ -0,0 +1,154 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.foundation.JewelConfigs
+
+/**
+ * A composable function that provides a tooltip area to display additional information associated with a UI element.
+ * The tooltip appears when the user hovers over the content with their cursor, after an optional delay.
+ *
+ * This function behavior is influenced by the 'jewel.customPopupRender' system property. If set to true, it will use
```
rock3r: Same here as for Popup — I'd simply point to the flag API.

rock3r (commented):
Great job! I have a few notes, but nothing major.

Re: standalone, I assume that'll happen in a separate PR?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelConfigs.kt
faogustavo: Makes sense

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
faogustavo: So, TBH I don't know exactly why this was happening, but having more than one composable callback was causing issues with the interface. Here is the message:

> java.lang.AbstractMethodError: Receiver class org.jetbrains.jewel.bridge.component.JBPopupRenderer does not define or inherit an implementation of the resolved method 'abstract void Popup(androidx.compose.ui.window.PopupPositionProvider, kotlin.jvm.functions.Function0, androidx.compose.ui.window.PopupProperties, kotlin.jvm.functions.Function1, kotlin.jvm.functions.Function1, kotlin.jvm.functions.Function2, androidx.compose.runtime.Composer, int)' of interface org.jetbrains.jewel.ui.component.PopupRenderer.
>	at org.jetbrains.jewel.ui.component.PopupRenderer$ComposeDefaultImpls.Popup$default(Popup.kt:87)

Have you seen this before? If you have any other solutions/ideas I'm all open to change

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/JewelDemoToolWindowFactory.kt
faogustavo: Added a setter, but kept the default flag value to the system property

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
faogustavo: Updated the order from this class, but I'd prefer to keep it in the composables as they can be a "drop-in replacement" for external users if they want those features.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
rock3r: Hmmm not sure I have seen this. Many composables have multiple slots and they work fine. Have you tried if changing the argument order changes something?

PS: this seems somewhat related https://slack-chats.kotlinlang.org/t/506169/hi-all-has-anyone-had-trouble-with-composable-interface-meth

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelFlags.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * If true, the Jewel library will use a custom popup renderer. This is an experimental feature and may not be fully
+     * stable.
+     *
+     * To enable this feature, you can either set the system property `jewel.customPopupRender` to `true`, or pass the
+     * `-Djewel.customPopupRender=true` argument when running your application.
+     *
+     * Note that this flag affects both popups, menus and tooltips rendering from Jewel Components.
+     */
```
rock3r: I reckon we also need to explain what this does and how it interacts with the Compose layerType property:

```suggestion
    /**
     * Enable custom popups handling in Jewel.
     * The default value is `false`.
     *
     * If enabled, the Jewel library will use a custom popup renderer, using separate 
     * windows instead of being drawn onto the same layer. 
     * 
     * This is an experimental feature and may not be fully stable. When enabled, Compose's popup
     * settings are ignored when using Jewel popups and tooltips. This means that setting
     * `compose.layers.type` will have no effect on Jewel popups and tooltips.
     *
     * To set this flag, you can also set the system property `jewel.customPopupRender` to `true`/`false`,
     * or pass the `-Djewel.customPopupRender=[true|false]` argument when
     * running your application.
     *
     * Note that this flag affects popups, menus and tooltips rendering from Jewel Components.
     * It does not affect `Dialog`s.
     */
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,102 @@ import androidx.compose.ui.unit.IntOffset
...
+}
+
+/**
+ * Interface responsible for rendering a popup with composable content. This can be used for overriding the default
+ * popup rendering behavior in Compose, within the context of the Jewel UI library.
+ *
+ * This interface implementation must then be provided via the [LocalPopupRenderer] composition local, and enable the
+ * [JewelFlags.useCustomPopupRenderer] flag to use it.
+ */
+public interface PopupRenderer {
```
rock3r, line 79: Please make this experimental

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,102 @@ import androidx.compose.ui.unit.IntOffset
...
+ * @param properties Configuration parameters for the popup, such as whether it should consume touch events or focusable
+ *   behavior.
+ * @param onPreviewKeyEvent Callback invoked for key events before they are dispatched to children. Return `true` to
+ *   consume the event.
+ * @param onKeyEvent Callback invoked for key events after they are dispatched to children. Return `true` to consume the
+ *   event.
+ * @param content The composable content to be displayed inside the popup.
+ */
+@Composable
+@InternalJewelApi
```
rock3r: I don't think this should be internal, but rather, it should be experimental

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -18,7 +21,102 @@ import androidx.compose.ui.unit.IntOffset
...
+
+    public companion object
+}
+
+/**
+ * Provides a custom [PopupRenderer] implementation for rendering popups in the Jewel UI library.
+ *
+ * Note that the value will only be used if the [JewelFlags.useCustomPopupRenderer] flag is set to `true`.
+ */
+public val LocalPopupRenderer: ProvidableCompositionLocal<PopupRenderer> = staticCompositionLocalOf {
```
rock3r, line 99: Mark as experimental

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
faogustavo: @rock3r should we also open the `JBPopup` to allow users to use it directly when implementing plugins?

rock3r: I don't think so. It's an implementation detail. If they use Popup and the flag is on they get the behaviour which is the important part. If we find out a use case for it to be opened up we can do it later on.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
faogustavo: updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
faogustavo: updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/JewelFlags.kt
faogustavo: updated

rock3r (approved):

rock3r:
Ready to merge

PR #3106, opened 2025-06-25
[JEWEL-852] Fixing ComboBoxes crash with Keyboard navigation

faogustavo:
- When the combo box does not have an initially selected item, the child focus manager may be null
- This only happens the second time you open a combo box using the keyboard, and the event received is the key up from the trigger to open the menu

# Evidences

| Before | After |
| --- | --- |
| Gravação de Tela 2025-06-25 às 14 53 02 | Gravação de Tela 2025-06-25 às 14 52 02 |


DanielSouzaBertoldi (approved):

jakub-senohrabek:
Ready to merge

rock3r:
@faogustavo can you add release notes for this bugfix? Thanks :)

PR #3107, opened 2025-06-27
[JEWEL-845] Add Context Menu to External Links

DanielSouzaBertoldi:
# Context

> [!IMPORTANT]
> This PR will stay marked as Draft as long as #3091 is not merged. It'd be great to have #3105 merged before too, but is not a prerequesite.

Currently, our ExternalLink implementation doesn't show a context menu with the default action group from IntelliJ `BrowserLink` component (see screenshot below). This PR makes the necessary changes to show such context menu **for external links only**.

image

## Main Changes
- Created a `LabelProvider` interface which both IDE and standalone versions will implement to return the action action label.
- Created a `BridgeUriHandler` class which implements `UriHandler` since in IDE we need to call IntelliJ's `BrowserUtil.browse()` function.
- Created a new `ExternalLink` overload where clicking the element automatically opens the given link and also displays a `ContextMenu`. It also has a new parameter `link` which users can use to specify the actual link from the text.
- This PR also fixes a weird visual bug in the Links row in `SwingComparisonPanel`.

# Evidences

## Right clicking external link
| IDE | Stand-alone |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/45fdbacf-f898-44e6-a0d5-d2996859e500" width="400" /> | <video src="https://github.com/user-attachments/assets/c00ab635-bffc-445d-8d18-1532f8457570" width="400" /> |

## Using a valid browser link
| IDE | Stand-alone |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/9d9afa08-bfb7-4e36-a504-ceb60ac0748d" width="400" /> | <video src="https://github.com/user-attachments/assets/f452c03c-1415-449a-814f-95baadbc1fe8" width="400" /> |

## Using an empty string
| IDE | Stand-alone |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/30f8beb3-de87-4286-81a6-f5ea084a8d5e" width="400" /> | <video src="https://github.com/user-attachments/assets/810edbf6-df4d-4f9f-a59e-bcca4cee8a51" width="400" /> |

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeLabelProvider.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeLabelProvider.kt
@@ -1,0 +1,9 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge
+
+import com.intellij.ide.IdeBundle
+import org.jetbrains.jewel.ui.util.LabelProvider
+
+public class BridgeLabelProvider : LabelProvider {
+    override fun fetchLabelFromKey(key: String): String = IdeBundle.messagePointer(key).get()
+}
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUriHandler.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUriHandler.kt
@@ -1,0 +1,11 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge
+
+import androidx.compose.ui.platform.UriHandler
+import com.intellij.ide.BrowserUtil
+
+public class BridgeUriHandler : UriHandler {
+    override fun openUri(uri: String) {
+        BrowserUtil.browse(uri)
+    }
+}
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/SwingBridgeTheme.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/SwingBridgeTheme.kt
@@ -7,7 +7,10 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.LocalClipboard
 import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.LocalUriHandler
+import org.jetbrains.jewel.bridge.BridgeLabelProvider
 import org.jetbrains.jewel.bridge.BridgePainterHintsProvider
+import org.jetbrains.jewel.bridge.BridgeUriHandler
 import org.jetbrains.jewel.bridge.SwingBridgeReader
 import org.jetbrains.jewel.bridge.clipboard.JewelBridgeClipboard
 import org.jetbrains.jewel.bridge.icon.BridgeNewUiChecker
@@ -17,6 +20,7 @@ import org.jetbrains.jewel.ui.ComponentStyling
 import org.jetbrains.jewel.ui.icon.LocalNewUiChecker
 import org.jetbrains.jewel.ui.painter.LocalPainterHintsProvider
 import org.jetbrains.jewel.ui.theme.BaseJewelTheme
+import org.jetbrains.jewel.ui.util.LocalLabelProvider
 
 private val bridgeThemeReader by lazy { SwingBridgeReader() }
 
@@ -35,6 +39,8 @@ public fun SwingBridgeTheme(content: @Composable () -> Unit) {
             LocalNewUiChecker provides BridgeNewUiChecker,
             LocalDensity provides scaleDensityWithIdeScale(LocalDensity.current),
             LocalClipboard provides remember { JewelBridgeClipboard() },
+            LocalUriHandler provides BridgeUriHandler(),
+            LocalLabelProvider provides BridgeLabelProvider(),
         ) {
             content()
         }
--- a/platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
+++ b/platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
@@ -66,6 +66,7 @@ import org.jetbrains.jewel.ui.component.Typography
 import org.jetbrains.jewel.ui.disabledAppearance
 import org.jetbrains.jewel.ui.icons.AllIconsKeys
 import org.jetbrains.jewel.ui.theme.textAreaStyle
+import org.jetbrains.skia.impl.Stats.enabled
 
 internal class SwingComparisonTabPanel : BorderLayoutPanel() {
     private val mainContent =
@@ -92,24 +93,33 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
             }
 
     private fun Panel.linksRow() {
-        row("Links:") {
-                browserLink("Enabled link", "")
+        val jewelReadmeLink = "https://github.com/JetBrains/intellij-community/tree/master/platform/jewel/#readme"
 
-                compose { ExternalLink(text = "Enabled link", enabled = true, onClick = {}) }
+        row("Links:") {
+                cell(
+                    component =
+                        BrowserLink(
+                                icon = AllIcons.Ide.External_link_arrow,
+                                text = "Enabled link",
+                                tooltip = null,
+                                url = "",
+                            )
+                            .apply { enabled(true) }
+                )
+
+                compose { ExternalLink(text = "Enabled link", link = "", enabled = true, onLinkOpened = {}) }
 
                 cell(
                         component =
                             BrowserLink(
-                                icon = IconLoader.getDisabledIcon(AllIcons.Ide.External_link_arrow),
-                                text = "Disabled link",
-                                tooltip = null,
-                                url = "",
-                            )
+                                    icon = IconLoader.getDisabledIcon(AllIcons.Ide.External_link_arrow),
+                                    text = "Disabled link",
+                                    tooltip = null,
+                                    url = jewelReadmeLink,
+                                )
+                                .apply { isEnabled = false }
                     )
-                    .applyToComponent {
-                        enabled(false)
-                        autoHideOnDisable = false
-                    }
+                    .applyToComponent { autoHideOnDisable = false }
 
                 compose { ExternalLink(text = "Disabled link", enabled = false, onClick = {}) }
             }
--- a/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Links.kt
+++ b/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Links.kt
@@ -27,12 +27,13 @@ import org.jetbrains.jewel.ui.theme.linkStyle
 @Composable
 public fun Links() {
     val alwaysUnderline = JewelTheme.linkStyle.copy(underlineBehavior = LinkUnderlineBehavior.ShowAlways)
+    val jewelReadMeLink = "https://github.com/JetBrains/intellij-community/tree/master/platform/jewel/#readme"
     Row(horizontalArrangement = Arrangement.spacedBy(10.dp), verticalAlignment = Alignment.CenterVertically) {
         Link(text = "Link", onClick = {})
 
         Link(text = "Always underlined", onClick = {}, style = alwaysUnderline)
 
-        ExternalLink("ExternalLink", {})
+        ExternalLink("ExternalLink", link = jewelReadMeLink, { println(it) })
 
         val items = remember { listOf("Light", "Dark", "---", "High Contrast", "Darcula", "IntelliJ Light") }
         var selected by remember { mutableStateOf(items.first()) }
@@ -51,7 +52,7 @@ public fun Links() {
 
         Link(text = "Always underlined", onClick = {}, style = alwaysUnderline, enabled = false)
 
-        ExternalLink(text = "ExternalLink", onClick = {}, enabled = false)
+        ExternalLink(text = "ExternalLink", link = jewelReadMeLink, onLinkOpened = {}, enabled = false)
 
         DropdownLink(text = "DropdownLink", enabled = false) {}
     }
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
@@ -1,5 +1,7 @@
 package org.jetbrains.jewel.ui.component
 
+import androidx.compose.foundation.ContextMenuArea
+import androidx.compose.foundation.ContextMenuItem
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.interaction.FocusInteraction
 import androidx.compose.foundation.interaction.HoverInteraction
@@ -17,18 +19,25 @@ import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.input.InputMode
 import androidx.compose.ui.input.pointer.PointerIcon
 import androidx.compose.ui.input.pointer.pointerHoverIcon
+import androidx.compose.ui.platform.ClipEntry
+import androidx.compose.ui.platform.LocalClipboard
 import androidx.compose.ui.platform.LocalInputModeManager
+import androidx.compose.ui.platform.LocalUriHandler
+import androidx.compose.ui.platform.UriHandler
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.text.style.TextOverflow
 import java.awt.Cursor
+import java.awt.datatransfer.StringSelection
+import kotlinx.coroutines.launch
 import org.jetbrains.jewel.foundation.modifier.onHover
 import org.jetbrains.jewel.foundation.modifier.thenIf
 import org.jetbrains.jewel.foundation.state.CommonStateBitMask
@@ -50,6 +59,8 @@ import org.jetbrains.jewel.ui.disabledAppearance
 import org.jetbrains.jewel.ui.focusOutline
 import org.jetbrains.jewel.ui.icon.IconKey
 import org.jetbrains.jewel.ui.painter.hints.Stateful
+import org.jetbrains.jewel.ui.util.LocalLabelProvider
+import java.io.IOException
 
 /**
  * A clickable text link that follows the standard visual styling with customizable appearance.
@@ -147,6 +158,102 @@ public fun ExternalLink(
         textStyle = textStyle,
         icon = style.icons.externalLink,
     )
+}
+
+/**
+ * An external link that follows the standard visual styling, including an external link icon.
+ *
+ * Please be aware that this ExternalLink will automatically open a link on click, unlike the other.
+ * 
+ * Provides a text link with an external link icon that indicates the link leads to external content. The link supports
+ * various states including enabled/disabled, focused, and hovered, with optional underline behavior based on the style
+ * configuration.
+ *
+ * **Guidelines:** [on IJP SDK webhelp](https://plugins.jetbrains.com/docs/intellij/link.html#external-link-icon)
+ *
+ * **Usage example:**
+ * [`Links.kt`](https://github.com/JetBrains/intellij-community/blob/master/platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/Links.kt)
+ *
+ * **Swing equivalent:** [`JLabel`](https://docs.oracle.com/javase/tutorial/uiswing/components/label.html) with HTML
+ * link styling and external link icon
+ *
+ * @param text The text to be displayed as a link
+ * @param link The actual link that will be used to open in the browser. Defaults to the text parameter value.
+ * @param onLinkOpened Called after the link was opened. Returns whether the opening was successful or not.
+ * @param modifier Modifier to be applied to the link
+ * @param enabled Controls whether the link can be interacted with
+ * @param textStyle The typography style to be applied to the link text
+ * @param overflow How the text should handle overflow
+ * @param interactionSource Source of interactions for this link
+ * @param style The visual styling configuration for the link
+ * @see javax.swing.JLabel
+ */
+@Composable
+public fun ExternalLink(
+    text: String,
+    link: String = text,
+    onLinkOpened: (success: Boolean) -> Unit,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    textStyle: TextStyle = JewelTheme.defaultTextStyle,
+    overflow: TextOverflow = TextOverflow.Clip,
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    style: LinkStyle = LocalLinkStyle.current,
+) {
+    val clipboard = LocalClipboard.current
+    val uriHandler = LocalUriHandler.current
+    val labelProvider = LocalLabelProvider.current
+    val scope = rememberCoroutineScope()
+
+    ContextMenuArea(
+        items = {
+            listOf(
+                ContextMenuItem(
+                    label = labelProvider.fetchLabelFromKey("action.text.open.link.in.browser"),
+                    onClick = {
+                        val success = openLink(uriHandler, link)
+                        
+                        onLinkOpened(success)
+                    },
+                ),
+                ContextMenuItem(
+                    label = labelProvider.fetchLabelFromKey("action.text.copy.link.address"),
+                    onClick = { scope.launch { clipboard.setClipEntry(ClipEntry(StringSelection(link))) } },
+                ),
+            )
+        },
+        enabled = enabled,
+        content = {
+            LinkImpl(
+                text = text,
+                onClick = {
+                    val success = openLink(uriHandler, link)
+                    
+                    onLinkOpened(success)
+                },
+                modifier = modifier,
+                enabled = enabled,
+                overflow = overflow,
+                interactionSource = interactionSource,
+                style = style,
+                textStyle = textStyle,
+                icon = style.icons.externalLink,
+            )
+        }
+    )
+}
+
+private fun openLink(uriHandler: UriHandler, link: String): Boolean {
+    return try {
+        uriHandler.openUri(link)
+        true
+    } catch(e: IllegalArgumentException) {
+        println("Oh no!!!! $e")
+        false
+    } catch(e: IOException) {
+        println("Oh DAMN! $e")
+        false
+    }
 }
 
 /**
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
@@ -37,6 +37,7 @@ import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.text.style.TextOverflow
 import java.awt.Cursor
 import java.awt.datatransfer.StringSelection
+import java.io.IOException
 import kotlinx.coroutines.launch
 import org.jetbrains.jewel.foundation.modifier.onHover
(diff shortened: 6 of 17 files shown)


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeLabelProvider.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge
+
+import com.intellij.ide.IdeBundle
+import org.jetbrains.jewel.ui.util.LabelProvider
+
+public class BridgeLabelProvider : LabelProvider {
+    override fun fetchLabelFromKey(key: String): String = IdeBundle.messagePointer(key).get()
```
DanielSouzaBertoldi: This `IdeBundle` call is the same used by `BrowerLink`:

```kotlin
private class OpenLinkInBrowser(val url: String)
  : DumbAwareAction(messagePointer("action.text.open.link.in.browser"), AllIcons.Nodes.PpWeb) {
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUriHandler.kt
```diff
@@ -0,0 +1,11 @@
...
+package org.jetbrains.jewel.bridge
+
+import androidx.compose.ui.platform.UriHandler
+import com.intellij.ide.BrowserUtil
+
+public class BridgeUriHandler : UriHandler {
+    override fun openUri(uri: String) {
+        BrowserUtil.browse(uri)
+    }
+}
```
DanielSouzaBertoldi: The default implementation for `UriHandler` can't really handle invalid URI's. `BrowserUtil` on the other hand will fallback to calling the command `open "<link>"`. That's why, for the IDE version of Jewel to be 1:1 with IntelliJ, we also need to use `BrowserUtil.browse()`.

Not really sure if running the `open` command would be necessary in the standalone version. For now, if the URI is invalid we just ignore.

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
```diff
@@ -92,24 +93,33 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
...
+                                    tooltip = null,
+                                    url = jewelReadmeLink,
+                                )
+                                .apply { isEnabled = false }
                     )
-                    .applyToComponent {
-                        enabled(false)
-                        autoHideOnDisable = false
-                    }
+                    .applyToComponent { autoHideOnDisable = false }
```
DanielSouzaBertoldi, line 118: Had to rearrange some stuff here because for some reason the comparison panel just started bugging out:

image

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -149,6 +160,102 @@ public fun ExternalLink(
...
+ * @param overflow How the text should handle overflow
+ * @param interactionSource Source of interactions for this link
+ * @param style The visual styling configuration for the link
+ * @see javax.swing.JLabel
+ */
+@Composable
+public fun ExternalLink(
+    text: String,
+    link: String = text,
+    onLinkOpened: (success: Boolean) -> Unit,
```
DanielSouzaBertoldi: Since now for external links we are opening the link by default on click instead of leaving it up to the user to handle it, I thought of having this callback that will tell the user if opening up the link was successful or not.

Maybe we should have a `beforeLinkOpened()` callback too? Not sure how far we should take this.

However, keeping the default action matches `BrowserLink` default behaviour.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/SwingBridgeTheme.kt
```diff
@@ -35,6 +39,8 @@ public fun SwingBridgeTheme(content: @Composable () -> Unit) {
             LocalNewUiChecker provides BridgeNewUiChecker,
             LocalDensity provides scaleDensityWithIdeScale(LocalDensity.current),
             LocalClipboard provides remember { JewelBridgeClipboard() },
+            LocalUriHandler provides BridgeUriHandler(),
+            LocalLabelProvider provides BridgeLabelProvider(),
```
faogustavo: Nit: IMO we should movi this to the `JewelComposePanel` functions. Some of them are without theme, but we still want to handle the links properly in the IDE

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeLabelProvider.kt
faogustavo: Any reason to use the lazy version (`messagePointer`) instead of the default `IdeBundle.message()`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/LabelProvider.kt
```diff
@@ -0,0 +1,15 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.util
+
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import org.jetbrains.jewel.foundation.InternalJewelApi
+
+@InternalJewelApi
+public interface LabelProvider {
```
faogustavo: I'm not sure about the `LabelProvider` name. Label usually represents the UI element on screen to me, not the text.

How about `TextProvider` or `StringProvider`? `StringProvider` would be a good match in the future if we add compose resources and localize the messages properly.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -149,6 +160,102 @@ public fun ExternalLink(
...
+        },
+    )
+}
+
+private fun openLink(uriHandler: UriHandler, link: String): Boolean {
+    return try {
+        uriHandler.openUri(link)
+        true
+    } catch (e: IllegalArgumentException) {
+        // Log something?
```
faogustavo: Maybe this: `JewelLogger.getInstance("ExternalLink").warn(e)`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
faogustavo: Nit: the `onLinkOpened` gives me the idea of being called only on success. Maybe we can keep the `onClick` naming.

For the `beforeLinkOpened`, it would be good to have if the user wants to trigger some analytics, etc

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Links.kt
```diff
@@ -27,12 +27,13 @@ import org.jetbrains.jewel.ui.theme.linkStyle
...
 public fun Links() {
     val alwaysUnderline = JewelTheme.linkStyle.copy(underlineBehavior = LinkUnderlineBehavior.ShowAlways)
+    val jewelReadMeLink = "https://github.com/JetBrains/intellij-community/tree/master/platform/jewel/#readme"
     Row(horizontalArrangement = Arrangement.spacedBy(10.dp), verticalAlignment = Alignment.CenterVertically) {
         Link(text = "Link", onClick = {})
 
         Link(text = "Always underlined", onClick = {}, style = alwaysUnderline)
 
-        ExternalLink("ExternalLink", {})
+        ExternalLink("ExternalLink", link = jewelReadMeLink, { println(it) })
```
faogustavo: Nit: Other places use the jewel logger.

```kotlin
JewelLogger.getInstance("ChipsAndTree").info("Click on $index")
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeLabelProvider.kt
DanielSouzaBertoldi: My idea was to replicate the exact function call that the platform calls, but yeah, there's no harm in using `.message()` 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/LabelProvider.kt
DanielSouzaBertoldi: Fine by me!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
```diff
@@ -119,7 +119,7 @@ public class ContextSubmenu(label: String, public val submenu: () -> List<Contex
 
 public class ContextMenuItemOption(
     public val icon: IconKey? = null,
-    public val actionType: ContextMenuItemOptionAction,
+    public val actionType: ContextMenuItemOptionAction? = null,
```
DanielSouzaBertoldi, line 121: The `actionType` parameter (which is used to map to a shortcut hint) should've been optional, my bad. You can have an item with an icon but no shortcut or vice versa.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -167,7 +167,9 @@ public fun MenuContent(
     val selectableItems = remember(items) { items.filterIsInstance<MenuSelectableItem>() }
 
     val anyItemHasIcon = remember { selectableItems.any { it.iconKey != null } }
-    val anyItemHasKeybinding = remember { selectableItems.any { it.keybinding != null || it.itemOptionAction != null } }
+    val anyItemHasKeybinding = remember {
+        selectableItems.any { it.keybinding?.isNotEmpty() == true || it.itemOptionAction != null }
```
DanielSouzaBertoldi, line 170: Another minor mistake from the previous PR: the `keybinding` parameter will always be an empty set. So we shouldn't check if it's null, but if it's not empty.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/IntUiStringProvider.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.jetbrains.jewel.ui.util.StringProvider
+
+public class IntUiStringProvider : StringProvider {
+    override fun fetchStringFromKey(key: String): String =
+        when (key) {
+            "action.text.open.link.in.browser" -> "Open Link in Browser"
+            "action.text.copy.link.address" -> "Copy Link Address"
+            else -> ""
+        }
+}
```
DanielSouzaBertoldi: Ideally we should have a more robust architecture to map an label key to its actual string representation, but that's outside the scope of this PR.

If this approach is fine for everybody, I'll create an issue in our board about improving this.

rock3r:
@DanielSouzaBertoldi please add release notes... and check if there are conflicts — GitHub seems very confused :D

image

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUriHandler.kt
```diff
@@ -0,0 +1,11 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge
+
+import androidx.compose.ui.platform.UriHandler
+import com.intellij.ide.BrowserUtil
+
+public class BridgeUriHandler : UriHandler {
```
rock3r: Please add KDoc; also, this can be an object

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/StringProvider.kt
```diff
@@ -0,0 +1,15 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.util
+
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import org.jetbrains.jewel.foundation.InternalJewelApi
+
+@InternalJewelApi
+public interface StringProvider {
```
rock3r: Please add KDoc. Also, I'd name this something like `MessageResourceResolver` as it seems to be what it does in the IDE version, but am open to alternatives.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/StringProvider.kt
```diff
@@ -0,0 +1,15 @@
...
+    public fun fetchStringFromKey(key: String): String
+}
+
+public val LocalStringProvider: ProvidableCompositionLocal<StringProvider> = staticCompositionLocalOf {
```
rock3r: Ditto with the renaming

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/IntUiStringProvider.kt
```diff
@@ -0,0 +1,13 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.intui.standalone
+
+import org.jetbrains.jewel.ui.util.StringProvider
+
+public class IntUiStringProvider : StringProvider {
```
rock3r: Please ensure this has KDoc. It should also be internal.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeStringProvider.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge
+
+import com.intellij.ide.IdeBundle
+import org.jetbrains.jewel.ui.util.StringProvider
+
+public class BridgeStringProvider : StringProvider {
```
rock3r: Please make this internal and add KDoc

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/StringProvider.kt
```diff
@@ -0,0 +1,15 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.util
+
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import org.jetbrains.jewel.foundation.InternalJewelApi
+
+@InternalJewelApi
+public interface StringProvider {
+    public fun fetchStringFromKey(key: String): String
```
rock3r: I'd call this something like `resolveIdeBundleMessage` — I don't love it leaks the `IdeBundle` concept into standalone, but if we need to fetch from other bundles in the future, at least we have space for maneuvering.

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/SwingComparisonTabPanel.kt
rock3r: Mysteries of Swing

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -149,6 +161,110 @@ public fun ExternalLink(
...
+ * @param textStyle The typography style to be applied to the link text
+ * @param overflow How the text should handle overflow
+ * @param interactionSource Source of interactions for this link
+ * @param style The visual styling configuration for the link
+ * @see javax.swing.JLabel
+ */
+@Composable
+public fun ExternalLink(
+    text: String,
+    link: String = text,
```
rock3r: I'd call this parameter `uri`. Also not sure about the default value, but can be persuaded :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -149,6 +161,110 @@ public fun ExternalLink(
     )
 }
 
+/**
+ * An external link that follows the standard visual styling, including an external link icon.
+ *
+ * Please be aware that this ExternalLink will automatically open a link on click, unlike the other.
```
rock3r:
```suggestion
 * Please be aware that this `ExternalLink` will automatically open a link on click, unlike the other overloads.
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -149,6 +161,110 @@ public fun ExternalLink(
...
+ * Provides a text link with an external link icon that indicates the link leads to external content. The link supports
+ * various states including enabled/disabled, focused, and hovered, with optional underline behavior based on the style
+ * configuration.
+ *
+ * **Guidelines:** [on IJP SDK webhelp](https://plugins.jetbrains.com/docs/intellij/link.html#external-link-icon)
+ *
+ * **Usage example:**
+ * [`Links.kt`](https://github.com/JetBrains/intellij-community/blob/master/platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/Links.kt)
+ *
+ * **Swing equivalent:** [`JLabel`](https://docs.oracle.com/javase/tutorial/uiswing/components/label.html) with HTML
```
rock3r: Not true, it's `com.intellij.ui.components.BrowserLink` — if the other overloads say the same, please fix them too (only for `ExternalLink`! other link types are `ActionLink` and `DropDownLink`)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -149,6 +161,110 @@ public fun ExternalLink(
...
+ * @param interactionSource Source of interactions for this link
+ * @param style The visual styling configuration for the link
+ * @see javax.swing.JLabel
+ */
+@Composable
+public fun ExternalLink(
+    text: String,
+    link: String = text,
+    onClick: (linkOpened: Boolean) -> Unit,
+    beforeOpeningLink: () -> Unit = {},
```
rock3r: I don't think we should have these. If people care about what happens when the link is clicked, they can use the other overloads, which give them control. This is supposed to be the "easy, works out of the box" version that covers the majority of simple cases.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -149,6 +161,110 @@ public fun ExternalLink(
...
+    overflow: TextOverflow = TextOverflow.Clip,
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    style: LinkStyle = LocalLinkStyle.current,
+) {
+    val clipboard = LocalClipboard.current
+    val uriHandler = LocalUriHandler.current
+    val stringProvider = LocalStringProvider.current
+    val scope = rememberCoroutineScope()
+
+    ContextMenuArea(
```
rock3r, line 234: Looks like you haven't added the menu to the other `ExternalLink` overload, is that intentional? Is it because you don't really know what to do with Open in Browser when the user doesn't provide you with a link? IMO it should call `onClick` in the other case

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
DanielSouzaBertoldi: Well, I was thinking of two cases:

1. When the user wants to have something like: [link text](https://www.github.com), in this case they'd declare both the text and the link that should be opened on click
2. When the user just wants to pass down a link: https://www.github.com. In this case, they wouldn't need to declare the link twice (one for `text`, another for `uri`)

That's why the default value for `uri` would be the text provided 😎

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
DanielSouzaBertoldi: Exactly the second case! Because the other overloads only expect a text, nothing more. So what if the user just passes down a string that's not a proper link? Or what if the user doesn't really want to open anything in the browser? We'd have a context menu with an option that when clicked wouldn't reflect what it actually does.

I mean, the name `ExternalLink` is pretty straightforward, but it doesn't necessarily mean the link should be opened in the browser.

Or maybe I'm overthinking it

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
rock3r: Yeah, I figured that was the case, but tbh the cases where people show full URLs in links are extremely rare. I would optimise for the case where you have a link that's distinct from the text; if a user has a lot of links that show a full URL, they can easily write a wrapper that sets both in one go:

```kotlin
@Composable
fun MyExternalLink(
  uri: String,
  ...
) {
  ExternalLink(uri, uri, ...)
}
```

But I don't want to commit to such an API in Jewel itself, given it's a niche need.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
rock3r: Ahah yeah maybe overthinking it a bit. I think ExternalLinks should always show the menu, and the "open in browser" action always calls the user-controlled `onClick`.

I mean, the Swing equivalent also uses the "open in browser" verb very loosely, since it can actually open a local file/path instead. We'd just be doing the same thing as Swing does.

DanielSouzaBertoldi: Yeah you got a point. I'll make it so that all `ExternalLink`s will display a context menu with the two default options 😎

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
DanielSouzaBertoldi: Yeah, that makes sense. To me it's okay to not set the default value as the text, just thought it'd be more "clean" in such rare cases lol

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/StringProvider.kt
DanielSouzaBertoldi: Honestly I'm kinda bad in coming up with names but I like your suggestion, it's way more specific and direct. @faogustavo what do you think?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -112,8 +124,7 @@ public fun Link(
  * **Usage example:**
  * [`Links.kt`](https://github.com/JetBrains/intellij-community/blob/master/platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/Links.kt)
  *
- * **Swing equivalent:** [`JLabel`](https://docs.oracle.com/javase/tutorial/uiswing/components/label.html) with HTML
- * link styling and external link icon
+ * **Swing equivalent:** com.intellij.ui.components.BrowserLink with HTML link styling and external link icon
```
rock3r:
```suggestion
 * **Swing equivalent:** [`BrowserLink`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-api/src/com/intellij/ui/components/BrowserLink.kt)
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -145,10 +156,118 @@ public fun ExternalLink(
...
+ * Provides a text link with an external link icon that indicates the link leads to external content. The link supports
+ * various states including enabled/disabled, focused, and hovered, with optional underline behavior based on the style
+ * configuration.
+ *
+ * **Guidelines:** [on IJP SDK webhelp](https://plugins.jetbrains.com/docs/intellij/link.html#external-link-icon)
+ *
+ * **Usage example:**
+ * [`Links.kt`](https://github.com/JetBrains/intellij-community/blob/master/platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/Links.kt)
+ *
+ * **Swing equivalent:** com.intellij.ui.components.BrowserLink with HTML link styling and external link icon
```
rock3r:
```suggestion
 * **Swing equivalent:** [`BrowserLink`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-api/src/com/intellij/ui/components/BrowserLink.kt)
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -145,10 +156,118 @@ public fun ExternalLink(
...
+ *
+ * @param text The text to be displayed as a link
+ * @param uri The actual link that will be used to open in the browser. Defaults to the text parameter value.
```
rock3r: Probably worth mentioning that errors happening when opening the URL are suppressed and logged.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
```diff
@@ -145,10 +156,118 @@ public fun ExternalLink(
...
+        style = style,
+        textStyle = textStyle,
+    )
+}
+
+@Composable
+private fun ExternalLinkImpl(
+    text: String,
+    onClick: () -> Unit,
+    uri: String = "",
```
rock3r: Does this need a default?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/StringProvider.kt
faogustavo: The term "Message" is a bit unusual to me at first. However, since the IDE uses this naming convention, we can also use it to maintain consistency.

rock3r: I'm also not terribly a fan of "leaking" the IDE naming, which is indeed a bit odd. But "resource" means something else already, especially in Compose land. Any better naming suggestion is welcome.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
DanielSouzaBertoldi: Agreed!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Link.kt
DanielSouzaBertoldi: I added this default so that we wouldn't need to add the `uri` parameter for `ExternalLinkImpl` in the `ExternalLink` overload that only accepts a text, since it's kinda "useless" in this case

rock3r: Yeah, but we've been bitten in the past by putting defaults in private impls, only to then forgetting passing the actual value, so I'd prefer to avoid that

DanielSouzaBertoldi: No problem, I understand! 😎 I removed the default value

rock3r (approved):
Lgtm!

faogustavo (approved):

rock3r:
Ready to merge

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/StringProvider.kt
DanielSouzaBertoldi: Well, seems like we are sticking with MessageResource for now. I'll resolve this conversation and also create an issue to create a better architecture to fetch string resources instead of relying on a really basic `when` clause 😎

PR #3112, opened 2025-07-02
[JEWEL-869] Fix Releases Demo tab in Jewel Plugin

DanielSouzaBertoldi:
# Context

The releases demo isn't functioning as expected. Here's a list of its bugs:
1. When typing in the search bar (e.g., "apple"), the typed text appears but the releases list doesn’t update accordingly
2. Clicking on an item list tag (release/canary/etc) doesn't show the tag name in the search bar and also doesn't update the releases list.
3. Changing the releases type from Android Studio to Android versions doesn't update the releases list.
4. Additionally, we are missing some Android Studio and Android releases, so this PR also includes updates for those.

Current state of the releases demo showing such bugs:

https://github.com/user-attachments/assets/4935d8e7-a314-48f7-a3d1-d6bb49ee9204

Also, I poked a bit in the code and changed the hover animation used in the Android Studio release image to face the user's cursor position. If the user clicks anywhere in the image box, the previous animation runs. This was entirely optional but I thought it'd be fun to update the interactive portion of the image nonetheless. Let me know what you think!

## Main Changes

// TODO!

# Evidences

## Main Fixes
| Context | Evidence |
|--------|--------|
| Typing in Search Bar | <video src="https://github.com/user-attachments/assets/a2d36418-98d9-4a36-b872-e88667baf42e" width="400" /> |
| Clicking Item List Tag | <video src="https://github.com/user-attachments/assets/b0886ea3-edf9-40f5-8aa7-e91d44e94f5a" width="400" /> |
| Changing List Type | <video src="https://github.com/user-attachments/assets/e938dc30-b69d-4a3e-aad0-596e7d01d9bb" width="400" /> |

## Bonus Stuff (?)
| Context | Evidence |
|--------|--------|
| Hovering AS release image | <video src="https://github.com/user-attachments/assets/909e069c-9a41-4619-81cb-8eb0a58b1da8" width="400" /> |
| Clicking on AS release image | <video src="https://github.com/user-attachments/assets/44d2d988-af9c-4d51-9ac3-2d8c585f19b6" width="400" /> |


platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/demodata/AndroidStudioReleasesReader.kt
```diff
@@ -118,7 +118,7 @@ internal object AndroidStudioReleasesReader {
     private fun translateDate(rawDate: String): String {
         val month = rawDate.substringBefore(" ").trimStart('0')
         val year = rawDate.substringAfterLast(" ".trimStart('0'))
-        val day = rawDate.substring(month.length + 1, rawDate.length - year.length - 1).trimStart('0')
+        val day = rawDate.substring(month.length + 1, rawDate.length - year.length - 2).trimStart('0')
```
DanielSouzaBertoldi: Had to update the releases gradle task because the [link where we fetch the metadata](https://jb.gg/android-studio-releases-list.json) returns a date with the format: "June 26, 2025"

This means that decrementing 1 from the end index will result in a substring with value `26,`. Comma included, which we don't need.

This could be achieved by keeping decrementing 1 but changing `year` to:

```kotlin
val year = date.substringAfterLast(",".trimStart('0')).trim()
```

Either way works, let me know what you think!

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ContentSource.kt
```diff
@@ -312,6 +312,14 @@ internal object AndroidReleases : ContentSource<ContentItem.AndroidRelease>() {
                 apiLevel = 35,
                 releaseDate = LocalDate(2024, 2, 17),
             ),
+            ContentItem.AndroidRelease(
+                displayText = "Android 16 (preview)",
+                imagePath = "/android-releases/16.svg",
+                versionName = "16",
+                codename = "Baklava",
```
DanielSouzaBertoldi, line 319: Not a big deal, but it kinda grinds my gears that the codename went from Vanilla Ice Cream to Baklava between Android 15 and 16. Wth?! 😭

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
```diff
@@ -131,34 +125,36 @@ private fun LeftColumn(project: Project, modifier: Modifier = Modifier, onSelect
...
-                items(
-                    items = currentContentSource.items,
-                    key = { it.key },
-                    contentType = {
-                        when (it) {
-                            is ContentItem.AndroidRelease -> ItemType.AndroidRelease
-                            is ContentItem.AndroidStudio -> ItemType.AndroidStudio
-                        }
+        key(currentContentSource) {
+            val listState = rememberSelectableLazyListState()
```
DanielSouzaBertoldi: This means that the scroll state will be reset with each filter. I'm not sure if that's ideal, but to me it makes sense since the list size can change depending on the filter result.

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
```diff
@@ -383,38 +385,48 @@ private fun RightColumn(selectedItem: ContentItem?, modifier: Modifier) {
...
+
+                            if (newIsHovered != controller.isHovered) {
+                                controller.onHoverStateChanged(newIsHovered)
+                            }
+
+                            if (newIsHovered) {
+                                val offset =
+                                    Offset(
+                                        x = (event.changes.first().position.x / size.width.toFloat()) * 2f - 1f,
+                                        y = (event.changes.first().position.y / size.height.toFloat()) * 2f - 1f,
```
DanielSouzaBertoldi, line 392: We need to convert the pixel coordinates (e.g., (75, 120)) into a generic, size-independent format. `(x / width) * 2 - 1` maps the range [0, width] to [-1, 1].

This means the center of the image is always (0, 0), and the edges are at -1 or 1.

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
```diff
@@ -383,38 +385,48 @@ private fun RightColumn(selectedItem: ContentItem?, modifier: Modifier) {
...
+                            }
+
+                            if (event.type == PointerEventType.Press) {
+                                controller.onClick()
+                            }
+                        }
                     }
                 }
-                .thenIf(applyModifier) { holoFoil(offset, intensity) },
+                .thenIf(intensity > 0.01f) {
```
DanielSouzaBertoldi, line 403: A small threshold (0.01f in this case) prevents the `holoFoil` modifier from being applied on the very last frame of the exit animation

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/FoilModifier.kt
```diff
@@ -88,9 +103,113 @@ internal fun Modifier.holoFoil(offset: Float, intensity: Float = 1f) = graphicsL
...
+    // We negate it so that moving the mouse down tilts the card up.
+    rotationX = -normalizedOffset.y * 10f * intensity
+
+    rotationZ = -normalizedOffset.x * 3f * intensity
     scaleX = 1f - .1f * intensity
     scaleY = 1f - .1f * intensity
 }
+
+@Composable
+internal fun rememberFoilInteractionController(): FoilInteractionController {
```
DanielSouzaBertoldi, line 119: Since all new code is tighly related to the holo foil effect, I opted to adding them in this same file.

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/demodata/AndroidStudioReleasesReader.kt
faogustavo: Just an idea for the future. We could also use the date parse API:

```kotlin
// Parsing and formatting LocalDate values using a custom format
val customFormat = LocalDate.Format {
    monthName(MonthNames.ENGLISH_FULL); char(' '); day(); chars(", "); year()
}
val date = customFormat.parse("Apr 16, 2024")
check(date == LocalDate(2024, Month.APRIL, 16))
val formatted = date.format(customFormat)
check(formatted == "Apr 16, 2024") 
```

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/FoilModifier.kt
```diff
@@ -88,9 +103,113 @@ internal fun Modifier.holoFoil(offset: Float, intensity: Float = 1f) = graphicsL
...
+    var mode by mutableStateOf(FoilShaderMode.Tracking)
+        private set
+
+    var isHovered by mutableStateOf(false)
+        private set
+
+    private var currentCursorPosition by mutableStateOf(Offset.Zero)
+    private var animationJob: Job? = null
+
+    fun onHoverStateChanged(newIsHovered: Boolean) {
```
faogustavo: nit: newIsHovered sounds a bit weird. Just 'isHovered' should work and you can use 'this.isHovered = isHovered' to set the value

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
faogustavo: Using the `key(currentContentSource)` to fix seems like a code smell. It appears that something is not working as expected to update the content, and this just recreates the whole composable.

You can verify that by checking adding logs and changing the content:

```
key(currentContentSource) {
  // Current code

  DisposableEffect(Unit) {
    println("Created")
    onDispose { println("Disposed") }
  }
}
```

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/demodata/AndroidStudioReleasesReader.kt
rock3r: Feel free to do it — I have stopped maintaining this sample a long time ago, as you could tell by the fact it wasn't really working anymore 💃  but I am happy to see it come back to life.

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ContentSource.kt
rock3r: tl;dr — https://www.androidauthority.com/android-trunk-stable-explained-3495640/ because of this :)

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/AndroidStudioReleases.kt
```diff
@@ -10,13 +10,299 @@ import org.jetbrains.jewel.samples.ideplugin.releasessample.ContentItem.AndroidS
 internal object AndroidStudioReleases : ContentSource<ContentItem.AndroidStudio>() {
     override val items: List<ContentItem.AndroidStudio> =
         listOf(
+            AndroidStudio(
+                displayText = "Android Studio Narwhal Feature Drop | 2025.1.2 Canary 7",
+                imagePath = null,
```
rock3r: The best place I know of to get the splash screens is... from the sources :)
1. Pick the tag you need on https://android.googlesource.com/platform/tools/idea/+refs (you want the ones starting with `studio-*` — e.g., `studio-2025.1.2-alpha07`)
2. Navigate to `adt-branding/resources/artwork/` and get the splash screens from there

DanielSouzaBertoldi: Dang it, I tried looking it up but wasn't successful. Thank you very much! Already saved this link in case I need something from AS builds later on 😎

rock3r: Figuring out where stuff is in Studio sources is not easy :D Having a local checkout made it a lot easier!

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/releasessample/ReleasesSampleCompose.kt
DanielSouzaBertoldi: You're totally right. I fixed this in my latest commit!

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -55,7 +58,11 @@ public fun SelectableLazyColumn(
     content: SelectableLazyListScope.() -> Unit,
 ) {
     val scope = rememberCoroutineScope()
-    val container = remember(content) { SelectableLazyListScopeContainer().apply(content) }
+    val latestContent = rememberUpdatedState(content)
+    val containerState = remember {
+        derivedStateOf(referentialEqualityPolicy()) { SelectableLazyListScopeContainer().apply(latestContent.value) }
+    }
+    val container = containerState.value
```
DanielSouzaBertoldi: The bug we had was really subtle, but here's an explanation of why it was happening and how these new lines fix it.

# The Bug 🐛 

Only remembering the `content` like so:

```kotlin
val container = remember(content) { SelectableLazyListScopeContainer().apply(content) }
```

Makes Compose discard every new `content` changes. It'll only run the first time, but if we filter the items from the list, the `content` lambda is recreated with a new `items()` content, but technically, to Compose, the key never really changed.  If the "shape" and captures of a lambda don't change in a meaningful way from its perspective, it will often pass the exact same lambda instance across recompositions to avoid creating new objects.

From Compose's point of view, both times the content has changed, it'll look like `{ items(someList, ...) }`.

The UI never updates because we failed to signal to remember that the _underlying data_ used by the lambda had changed.

# The fix!

1. By passing our content lambda to `rememberUpdatedState`, it'll update its internal `.value` to point to the newly created content lambda from the latest recomposition.
2. Then, we can derive a state from our updated state using the `referentialEqualityPolicy` function. This is the same policy that `LazyColumn` uses under the hood (this solution was based on how `LazyColumn` handles content updates!). This policy tells the derived state to check for changes using referential equality (===), in other words, if the new result is the exact same object instance in memory as the old one.

So, when we filter items from the list, the parent recomposes and a new content lambda is created. The `rememberUpdatedState` object's internal `.value` is updated to this new lambda. This change automatically triggers our `derivedStateOf` block to re-run its calculation.

Inside the `derivedStateOf`, we create a new `SelectableLazyListScopeContainer()` and apply the new content lambda to it. Since this new container is always a new object instance, the `referentialEqualityPolicy` confirms it's a change, and the new state is published. This finally provides the `LazyColumn` with the updated list data.

If you've read this far, here's a cookie: 🍪 
Also, hope the explanation made sense!

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -55,7 +58,11 @@ public fun SelectableLazyColumn(
     content: SelectableLazyListScope.() -> Unit,
 ) {
     val scope = rememberCoroutineScope()
-    val container = remember(content) { SelectableLazyListScopeContainer().apply(content) }
+    val latestContent = rememberUpdatedState(content)
+    val containerState = remember {
```
faogustavo: nit: can't we use the delegation (`by`) here?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
rock3r: This is a _brilliant_ explanation and analysis, thank you a ton for taking the time to write it down. It helps us understand things better and hopefully learn valuable lessons 🙏

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
rock3r: If you use delegation you can't do the `.value` below

rock3r (approved):

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
DanielSouzaBertoldi: No problem, I changed it to use delegation 😎

rock3r: Ah lol I totally looked at the wrong place 💀

rock3r:
Ready to merge

PR #3116, opened 2025-07-08
[JEWEL-868] Fix parameter value leak in Modifiers

rock3r:
`Modifier.onHover`, `Modifier.onMove`, and `Modifier.customTitleBarMouseEventHandler` are using a captured parameter in `pointerInput`, without using it as a key.

This means that if the parameter value changes, the old value is kept around in the `pointerInput` and can cause leaks and bugs where the old lambda gets invoked instead of the new one.

### Bug fixes
 * Fix parameter value leak in `Modifier.onHover` and `Modifier.onMove`


faogustavo (approved):

rock3r:
Ready to merge

PR #3117, opened 2025-07-08
[JEWEL-872] Remove unnecessary JBPanel API

rock3r:
This is a new experimental API added by #3099, a PR that was merged without approval. It can be safely removed.

We already have the new `compose { }` and the existing `JewelComposePanel { }` APIs, we do not need a `JBPanel { }` API too — especially since it does NOT create a `JBPanel`.


faogustavo (approved):

rock3r:
Ready to merge

PR #3119, opened 2025-07-09
[JEWEL-838] Remove API Dumps from Gradle Builds

DanielSouzaBertoldi:
# Context

We don't need to run `apiDump` now that we have `ApiCheckTest` run config from IntelliJ Platform (IJP). This PR removes a bunch of stuff but also repurposes some code to still validate if the api dump contains data classes that shouldn't be excluded from validation.

## Investigation
The API dump generated by `ApiCheckTest` is way more "packed" than the version generated by `binaryCompatValidator`. Here's a snippet of an dump:

```
org.jetbrains.jewel.foundation.lazy.SelectableLazyListScope
- a:item(java.lang.Object,java.lang.Object,Z,kotlin.jvm.functions.Function3):V
- bs:item$default(org.jetbrains.jewel.foundation.lazy.SelectableLazyListScope,java.lang.Object,java.lang.Object,Z,kotlin.jvm.functions.Function3,I,java.lang.Object):V
- a:items(I,kotlin.jvm.functions.Function1,kotlin.jvm.functions.Function1,kotlin.jvm.functions.Function1,kotlin.jvm.functions.Function4):V
- bs:items$default(org.jetbrains.jewel.foundation.lazy.SelectableLazyListScope,I,kotlin.jvm.functions.Function1,kotlin.jvm.functions.Function1,kotlin.jvm.functions.Function1,kotlin.jvm.functions.Function4,I,java.lang.Object):V
- a:stickyHeader(java.lang.Object,java.lang.Object,Z,kotlin.jvm.functions.Function3):V
- bs:stickyHeader$default(org.jetbrains.jewel.foundation.lazy.SelectableLazyListScope,java.lang.Object,java.lang.Object,Z,kotlin.jvm.functions.Function3,I,java.lang.Object):V
org.jetbrains.jewel.foundation.theme.JewelTheme
- sf:Companion:org.jetbrains.jewel.foundation.theme.JewelTheme$Companion
f:org.jetbrains.jewel.foundation.theme.JewelTheme$Companion
- f:getInstanceUuid(androidx.compose.runtime.Composer,I):java.util.UUID
f:org.jetbrains.jewel.foundation.theme.JewelThemeKt
- sf:getLocalThemeInstanceUuid():androidx.compose.runtime.ProvidableCompositionLocal
f:org.jetbrains.jewel.foundation.theme.ThemeDefinition
- f:getDisabledAppearanceValues():org.jetbrains.jewel.foundation.DisabledAppearanceValues
```

From this snippet we can infer some valuable info that'll help us to identify data classes.

First, the file follows a clear hierarchical structure: lines indented with a - prefix are members (methods or fields) of the preceding, non-indented class declaration.

Second, the declarations are prefixed with modifier specifiers that indicate their behavior, mirroring Kotlin keywords:

| Prefix | Meaning |
|--------|--------|
| f | **Final** - The class or member cannot be overridden. |
| a | **Abstract** - The class or member must be implemented by a subclass. |
| sf | **Static Final** - A static and final member, often from a companion object. |
| bs | Bridge Static - A [compiler-generated bridge method](https://docs.oracle.com/javase/tutorial/java/generics/bridgeMethods.html) that is also static. Seen with copy$default. |

If a class declaration has no prefix, it is open for extension (such as an open class or an interface).

Cool! How does that help us? Well, the fun thing about data classes is that:

This information is crucial because of two non-negotiable rules for Kotlin data classes:

1. They are always implicitly final.

2. The compiler automatically generates copy() and componentN() methods.

So, we can go line by line and find classes marked as _final_ that contain `copy/copy$default` and `component*` properties!

Just to make it a bit more clear, here's an example of how a data class will be presented in the dump file:

```
f:org.jetbrains.jewel.foundation.theme.DataClass
- sf:$stable:I
- <init>(java.lang.String,java.lang.String,java.lang.String):V
- f:component1():java.lang.String // <--- component method for each parameter in the data class
- f:component2():java.lang.String
- f:component3():java.lang.String
- f:copy(java.lang.String,java.lang.String,java.lang.String):org.jetbrains.jewel.foundation.theme.DataClass
- bs:copy$default(org.jetbrains.jewel.foundation.theme.DataClass,java.lang.String,java.lang.String,java.lang.String,I,java.lang.Object):org.jetbrains.jewel.foundation.theme.DataClass //  <--- copy methods!
- equals(java.lang.Object):Z
- f:getFirstParam():java.lang.String
- f:getSecondParam():java.lang.String
- f:getThirdParam():java.lang.String
- hashCode():I
```

And here's the data class declaration:

```kotlin
public data class DataClass(
    public val firstParam: String, 
    public val secondParam: String, 
    public val thirdParam: String,
)
```

## Main Changes
- Removed `binaryCompatValidator` dependency + all code related to it since now we don't need to generate an API dump anymore.
- Updated `Pre_push.xml` to run `validatePublicApi` instead of `apiDump`.
- Refactored `ValidatePublicApiTask` to look for data classes in the new dump format.
- Deleted all `.api` files + `api/` directory

# Evidences
| Context | Evidence |
|--------|--------|
| Data Class Violation Found |  |
| Excluded Classes |  |
| No violations |  |


platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
```diff
@@ -1,16 +1,23 @@
...
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskAction
+import java.io.File
+import java.util.regex.PatternSyntaxException
 
 @CacheableTask
-open class ValidatePublicApiTask : SourceTask() {
+abstract class ValidatePublicApiTask : DefaultTask() {
```
DanielSouzaBertoldi: Honestly I'm not sure if there are any advantages in inheriting from `DefaultTask` instead of `SourceTask`, but since all examples in the [writing a task](https://docs.gradle.org/current/userguide/writing_tasks.html#header) doc use `DefaultTask` I opted to follow the documentation per se.

I also asked Gemini about the difference and the answer was a basic "SourceTask is kinda old, DefaultTask is more robust" and the links it provided had nothing to do with their claim, so yeah. Googling about it didn't seem to bring any meaningful discussion about it too.

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
```diff
@@ -64,6 +64,13 @@ open class ValidatePublicApiTask : SourceTask() {
                     }
                     appendLine()
                 }
+
+                appendLine("Avoid using data classes. Turn it into a class and add the @GenerateDataFunctions annotation.")
+                appendLine(
+                    "For specific cases, you can exclude a data class from the validation. " +
+                        "For this, just add it to the 'apiValidation.excludedClassRegexes' block in " +
+                        "your build.gradle.kts."
+                )
```
DanielSouzaBertoldi: Thought it'd be nice to give more context on what should be done if this error occurs.

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
```diff
@@ -72,57 +79,63 @@ open class ValidatePublicApiTask : SourceTask() {
...
 
-            if (line.contains("}")) {
-                currentClassStack.pop()
-                continue
-            }
+                // We are inside a class, so check for data class methods.
+                val info = dataClasses.getOrPut(currentClassFqn) { DataClassInfo() }
+
+                // Detect copy() method and the $default version.
+                if ("copy(" in line || "copy\$default(" in line) {
```
DanielSouzaBertoldi, line 97: The `copy$default` is a [synthetic method](https://www.baeldung.com/java-synthetic) generated by the Kotlin compiler to handle default parameters in the copy function (e.g., user.copy(name = "New Name")). If you have a data class without default parameters, then the dump would generate the copy method as a plain `copy()`.

platform/jewel/.idea/runConfigurations/Pre_push.xml
```diff
@@ -22,4 +22,4 @@
     <RunAsTest>false</RunAsTest>
     <method v="2" />
   </configuration>
-</component>
\ No newline at end of file
+</component>
```
rock3r: Make sure not to commit empty lines at EOF in IDE settings xml files — the IDE will try to remove them all the time. Can you revert the added empty line in this file, and in the reformat run config?

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
```diff
@@ -64,6 +64,13 @@ open class ValidatePublicApiTask : SourceTask() {
                     }
                     appendLine()
                 }
+
+                appendLine("Avoid using data classes. Turn it into a class and add the @GenerateDataFunctions annotation.")
```
rock3r:
```suggestion
                appendLine("Avoid using data classes in public API. Turn them into regular classes, implement equals()/hashCode()/toString(), and add the @GenerateDataFunctions annotation.")
```

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
```diff
@@ -54,7 +54,7 @@ open class ValidatePublicApiTask : SourceTask() {
 
         if (violations.isNotEmpty()) {
             val message = buildString {
-                appendLine("Data classes found in public API.")
+                appendLine("Data classes are not allowed. Found violations:")
```
rock3r:
```suggestion
                appendLine("Data classes are not allowed in public API. Found violations:")
```

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
```diff
@@ -64,6 +64,13 @@ open class ValidatePublicApiTask : SourceTask() {
                     }
                     appendLine()
                 }
+
+                appendLine("Avoid using data classes. Turn it into a class and add the @GenerateDataFunctions annotation.")
+                appendLine(
+                    "For specific cases, you can exclude a data class from the validation. " +
+                        "For this, just add it to the 'apiValidation.excludedClassRegexes' block in " +
+                        "your build.gradle.kts."
```
rock3r:
```suggestion
                        "the module's build.gradle.kts."
```

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
```diff
@@ -72,57 +79,63 @@ open class ValidatePublicApiTask : SourceTask() {
...
+            isExcluded
+        }.toSet()
     }
 }
 
 @Suppress("DataClassShouldBeImmutable") // Only used in a loop, saves memory and is faster
 private data class DataClassInfo(
-    val fqn: String,
     var hasCopyMethod: Boolean = false,
-    var isLikelyValueClass: Boolean = false,
```
rock3r, line 127: Can you confirm this doesn't match value classes? I remember getting false positives with those back in the day; just want to double check.

DanielSouzaBertoldi: Just tested, the dump file doesn't add `copy` nor `componentN` members, so we're safe 😎 

Value class:

```kotlin
@JvmInline
public value class ValueClass(public val firstParam: String)
```

api-dump.txt result:

```
f:org.jetbrains.jewel.foundation.theme.ValueClass
- bsf:box-impl(java.lang.String):org.jetbrains.jewel.foundation.theme.ValueClass
- s:constructor-impl(java.lang.String):java.lang.String
- equals(java.lang.Object):Z
- s:equals-impl(java.lang.String,java.lang.Object):Z
- sf:equals-impl0(java.lang.String,java.lang.String):Z
- f:getFirstParam():java.lang.String
- hashCode():I
- s:hashCode-impl(java.lang.String):I
- s:toString-impl(java.lang.String):java.lang.String
- bf:unbox-impl():java.lang.String
```

`./gradlew validatePublicApi` result:

image

platform/jewel/.idea/runConfigurations/Pre_push.xml
DanielSouzaBertoldi: Will do! Thanks for pointing that out 🙏🏻

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
rock3r: Usually static analysis tools use `SourceTask` as base. The docs are pretty vague, but still hint at why you'd want to use a more specialised task such as `SourceTask`:

> A SourceTask performs some operation on source files.

Most samples use `DefaultTask` because it's the most generic type, but in this case, we're doing source-based operations, so it's appropriate to use `SourceTask`.

Gemini was completely wrong about this :) There is no "old" factor involved; only using the specialised base type for a task that reads source files.

I also asked 2.5 Pro and the answer it gave me is correct! https://g.co/gemini/share/13383e2ee816

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
rock3r: Perfect, thanks for checking :)

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
DanielSouzaBertoldi: Yeah that makes sense. I reverted these changes to keep using `SourceTask`, thanks for your input 🙏🏻

platform/jewel/buildSrc/src/main/kotlin/ValidatePublicApiTask.kt
DanielSouzaBertoldi: Also, we could link to this article by Jake Wharton about why it's a bad idea do use data classes in public APIs in Kotlin: https://jakewharton.com/public-api-challenges-in-kotlin/

What do you guys think? @faogustavo @rock3r feel free to suggest other articles too if you want 🙏🏻

rock3r: +1 for mentioning that article (it's the main one I can think of, too).

DanielSouzaBertoldi: Added article link 😎 + updated evidence image for violations found

faogustavo (approved):

rock3r:
@DanielSouzaBertoldi of course, more conflicts. Until we merge this and remove the dumps, any merged PR will cause them :P

DanielSouzaBertoldi:
@rock3r  No problem, we can merge it this friday when things settle down a bit 😎 until then I'll just ignore the merge conflicts for now

rock3r:
I'll add the "ready" label on this as it formally is, just will need a rebase at the last minute before being merged :)

PR #3120, opened 2025-07-10
[JEWEL] Improve release notes extractor script

rock3r:
This adds two features to the script:
 * The "since" date is now inferred from the latest value in the `RELEASE NOTES.md` file. It can be overridden by passing a date via the `--since` parameter.
 * The script accepts an `--until` date parameter; the default behaviour if that's not specified is "until now"


DanielSouzaBertoldi (approved):
This script seems like a pretty neat use case to use [Clikt](https://ajalt.github.io/clikt/), but this also works fine, just a thought 🙏🏻

rock3r:
Yeah, I want to migrate it to click, but it's on the backlog :)

rock3r:
Ready to merge

PR #3122, opened 2025-07-14
[JEWEL-891] Annotate detectPressAndCancel as internal API

rock3r:
It was opened up by someone without our knowledge, and it risks leaking into the public API otherwise.

APIs annotated as `@ApiStatus.Internal` do not show up in the API dumps, because as far as IJP is concerned, they should not be used. However, we also have standalone users, and for them that annotation has no meaning.

This will need to be cherry-picked onto 252 asap, too. It is being done manually in the 251 cherry pick.


rock3r:
Ready to merge

PR #3125, opened 2025-07-15
KT UAST: detach fake PSI for annotation from FE1.0

jsjeon:
Now the conversions from source annotation entry to PSI annotation
are implemented on each frontend's resolve provider service.

^KTIJ-34874 fixed


jsjeon:
cc @BartvHelvert

BartvHelvert:
Just merged with some additional changes to drop the fake do/while loop, let me know if you see any issues, thanks!

PR #3126, opened 2025-07-15
[JEWEL-895] Clean up Markdown image extension

rock3r:
This cleans up a number of issues with the Markdown image extension:
1. APIs were not annotated as experimental, but everything under Markdown should be
2. Coil3ImageRendererExtensionImpl was public for no good reason
3. The alt text for inline images was populated incorrectly
4. The dependencies were not set up correctly (partly not exported, both in JPS and Gradle)
5. We were using an outdated Coil version
6. Naming conventions were not fully followed for the image loading extension
7. The extension APIs were missing KDoc and the readme was not very clear
8. In the sample and Markdown preview we were creating a new image loader on every recomposition
9. We only had a composable way to init a new extension, when all we needed was a Coil context
10. Some opportunistic cleanup of the Markdown sample, including updating the Readme with more images
11. Images in Markdown were missing the tooltip with the alt text (if any)

In the release notes below I'll assume this is for a 0.29 release, as we need to fix this public API blunder before 0.29 is release...

> [!IMPORTANT]
> This will need manual cherry-picking to 251 and 252

### ⚠️ Important Changes
 * Coil version updated from 3.1.0 to 3.2.0

### New features
 * Markdown images (experimental) show their alt text in a tooltip when hovered.

### Bug fixes
 * Fixed several issues with Markdown image loading.


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
```diff
@@ -168,7 +168,21 @@ public open class DefaultInlineMarkdownRenderer(rendererExtensions: List<Markdow
         currentTextStyle: TextStyle,
     ) {
         // Each image source corresponds to one rendered image.
-        appendInlineContent(node.source, "![${node.title}](${node.source})")
+        appendInlineContent(
+            node.source,
+            buildString {
+                append(" ![")
```
faogustavo, line 174: Is this extra space in the start intended?

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtension.kt
```diff
@@ -0,0 +1,64 @@
...
+         * This shouldn't be used if there is an app-wide image loader already available; instead, use the constructor
+         * to pass in the already available image loader.
+         *
+         * Note that every invocation creates a new [ImageLoader]; it is not recommended to call this method multiple
+         * times in a process. Instead, create one top-level instance and share it throughout the process if at all
+         * possible.
+         */
+        @ExperimentalJewelApi
+        @Composable
+        public fun withDefaultLoader(): Coil3ImageRendererExtension = withDefaultLoader(LocalPlatformContext.current)
```
faogustavo, line 42: Should we add a 'remember' here to prevent creating an instance on each recomposition? Similar to what you've done in the sample?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
rock3r: Yes otherwise when you select and copy to the clipboard there is no space around this. It's a workaround :)

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtension.kt
rock3r: I'd rather keep the two variants behaving the same tbh

rock3r:
Even though this will require manual merging: ready to merge

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultInlineMarkdownRenderer.kt
rock3r: There is one at the end too for the same reason btw

rock3r:
(amended the commit message to match the PR title/description)

faogustavo (approved):

rock3r:
ready to merge (will require manual merging)

AlexVanGogen:
[Merged manually](https://github.com/JetBrains/intellij-community/commit/6dafece15ce817f38328853e81c5a79969c5ffe9), going to be part of 0.29

PR #3127, opened 2025-07-15
[JEWEL-898] TooltipArea should not be public

rock3r:
Our public tooltip API should remain `Tooltip`, which delegates to `TooltipArea` that is a private impl detail. It should not be in the public API.

To avoid further confusion, the `TooltipArea` composable has also been renamed to `FlagAwareTooltipArea`. This better describes its function and is clearly a private API name.

This also fixes `ComposeShowcase` to use `Tooltip` instead of `TooltipArea`, and sets up the Jewel VCS settings to recognise YouTrack issues. A few issues in the `ComposeShowcase` have also been fixed.

### DIFF
--- a/platform/jewel/.idea/vcs.xml
+++ b/platform/jewel/.idea/vcs.xml
@@ -10,6 +10,16 @@
   <component name="GitSharedSettings">
     <option name="synchronizeBranchProtectionRules" value="false" />
   </component>
+  <component name="IssueNavigationConfiguration">
+    <option name="links">
+      <list>
+        <IssueNavigationLink>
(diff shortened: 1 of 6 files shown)


platform/jewel/.idea/vcs.xml
```diff
@@ -10,6 +10,16 @@
...
+  <component name="IssueNavigationConfiguration">
+    <option name="links">
+      <list>
+        <IssueNavigationLink>
+          <option name="issueRegexp" value="(?x)\b(PRJ_ROAD|CPRTINF|EXPOSED|IDEABKL|MERMAID|SUPPORT|GROOVY|JEWEL|AMPER|BAZEL|IJSDK|IJUCF|JUNIE|RSCPP|RIDER|SKIKO|SPNSR|PSAP|DEXP|AQUA|DCVR|DMRY|DOTP|DTRC|EIJP|GTLB|IDES|IDEA|IJPL|JBAP|JBAI|JPAB|KTIJ|KTLC|KTNB|KTOR|PKGS|PROF|RDCT|RSRP|RSPL|RDOC|RUBY|RUST|TABF|VSAI|VSRS|YTDB|LLM|AIT|AIR|CDC|CPP|CLJ|CWM|RDO|CMP|DBE|DOC|DMU|DPA|EDC|GTW|GRZ|HUB|IFT|VIM|IMP|EDU|JBR|KTI|KMT|MPS|MJP|NIM|PRJ|RCB|BND|SCL|TCC|TBX|WEB|WRS|YTM|OC|AT|CN|DS|FL|GC|GO|IA|OT|KI|KG|KT|KE|LV|MP|MR|WI|PY|QD|RF|RG|RS|TW|XD|JT|R)-\d+\b#YouTrack" />
+          <option name="linkRegexp" value="https://youtrack.jetbrains.com/issue/$0" />
+        </IssueNavigationLink>
+      </list>
+    </option>
+  </component>
```
DanielSouzaBertoldi, line 22: Should we just ignore changes to `.idea` files that are not related to run configurations?

faogustavo (approved):

platform/jewel/.idea/vcs.xml
rock3r: I added this on purpose as it makes the issue IDs clickable across the IDE (commit messages etc); this already works on the JPS project but wasn't enabled in the Gradle one, it's just realigning

rock3r:
Ready to merge

PR #3129, opened 2025-07-16
[JEWEL-286] Support IconButtons Without Background

DanielSouzaBertoldi:
# Context

Our `IconButton` implementation doesn't offer an out-of-the-box way of using `IconButton`s without background, which can be useful in some scenarios, like the clear button in a text field. This PR adds such support.

## Main Changes
- Created a composition local `LocalIconButtonStyleWithoutBackground` that'll provide an `IconButtonStyle` with `IconButtonColors` with `Color.Transparent` colors for parameters related to background.
- Added a new variable to `DefaultComponentStyling` called `iconButtonStyleWithoutBackground`.
   - Both bridge and standalone are providing the required `IconButtonStyle` just using transparent colors for background values.
- Updated the `Buttons` showcase to better structure the examples. Also added an example of background-less `IconButton`.

# Evidences

## `IconButton` without background 
| IDE | Standalone |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/d2147ebc-c02d-444a-b85f-5154d9344cb3" /> | <video src="https://github.com/user-attachments/assets/de77c8de-1a80-41a5-bea0-43ffcb3a3ec4" /> |

## Buttons showcase restructuring
| Before | After |
|--------|--------|
|  |  |

### New features
 * Jewel now has support for `IconButton`s with transparent background. Just use a regular `IconButton` but set `style = JewelTheme.transparentIconButtonStyle`.


platform/jewel/int-ui/int-ui-standalone/api-dump-unreviewed.txt
```diff
@@ -610,11 +610,7 @@ f:org.jetbrains.jewel.intui.standalone.theme.IntUiGlobalMetricsKt
...
...rg.jetbrains.jewel.ui.component.styling.HorizontalProgressBarStyle,org.jetbrains.jewel.ui.component.styling.IconButtonStyle,org.jetbrains.jewel.ui.component.styling.InlineBannerStyles,org.jetbrains.jewel.ui.component.styling.LazyTreeStyle,org.jetbrains.jewel.ui.component.styling.LinkStyle,org.jetbrains.jewel.ui.component.styling.MenuStyle,org.jetbrains.jewel.ui.component.styling.ButtonStyle,org.jetbrains.jewel.ui.component.styling.PopupContainerStyle,org.jetbrains.jewel.ui.component.styling.SplitButtonStyle,org.jetbrains.jewel.ui.component.styling.RadioButtonStyle,org.jetbrains.jewel.ui.component.styling.ScrollbarStyle,org.jetbrains.jewel.ui.component.styling.SegmentedControlButtonStyle,org.jetbrains.jewel.ui.component.styling.SegmentedControlStyle,org.jetbrains.jewel.ui.component.styling.SelectableLazyColumnStyle,org.jetbrains.jewel.ui.component.styling.SliderStyle,org.jetbrains.jewel.ui.component.styling.SimpleListItemStyle,org.jetbrains.jewel.ui.component.styling.TextAreaStyle,org.jetbrains.jewel.ui.component.styling.TextFieldStyle,org.jetbrains.jewel.ui.component.styling.TooltipStyle,org.jetbrains.jewel.ui.component.styling.DropdownStyle):org.jetbrains.jewel.ui.ComponentStyling
```
DanielSouzaBertoldi: Don't worry, all these dumps were moved to `api-dump.txt` 🙏🏻

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
```diff
@@ -22,6 +22,16 @@ public fun IconButtonStyle.Companion.dark(
     metrics: IconButtonMetrics = IconButtonMetrics.defaults(),
 ): IconButtonStyle = IconButtonStyle(colors, metrics)
 
+public fun IconButtonStyle.Companion.lightWithoutBackground(
+    colors: IconButtonColors = IconButtonColors.lightWithoutBackground(),
+    metrics: IconButtonMetrics = IconButtonMetrics.defaults(),
+): IconButtonStyle = IconButtonStyle(colors, metrics)
```
DanielSouzaBertoldi, line 32: Yeah, I could've created an `IconButtonStyleWithoutBackground` along with `IconButtonColorsWithoutBackground` and just create new functions like:

```kotlin
public fun IconButtonStyleWithoutBackground.Companion.light(
    colors: IconButtonColorsWithoutBackground = IconButtonColorsWithoutBackground.light(),
    metrics: IconButtonMetrics = IconButtonMetrics.defaults(),
): IconButtonStyleWithoutBackground = IconButtonStyleWithoutBackground(colors, metrics)
```

But to me it seems pretty "boilerplate-y". We can reuse what we currently have, the downside is breaking the name convention for this functions. I'm eager to read your thoughts about this. Is it better to create new variations for everything or is it ok to reuse them like this?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
```diff
@@ -57,6 +67,34 @@ public fun IconButtonColors.Companion.light(
...
 
+public fun IconButtonColors.Companion.lightWithoutBackground(
+    foregroundSelectedActivated: Color = IntUiLightTheme.colors.gray(14),
+    border: Color = Color.Transparent,
+    borderDisabled: Color = Color.Transparent,
+    borderSelected: Color = Color.Transparent,
+    borderSelectedActivated: Color = Color.Transparent,
+    borderFocused: Color = Color.Transparent,
+    borderPressed: Color = Color.Transparent,
+    borderHovered: Color = Color.Transparent,
```
DanielSouzaBertoldi: So, the borders for both bridge and the light/dark variations in standalone that do have background just reference the colors from background related parameters.

I'm not sure if in this case all borders should reference the default colors passed as background to `IconButton` like the code above this.

In other words: it's ok to use color transparent, or for borders we'd need to keep the default colors from the `IconButton` with background?

```kotlin
    border: Color = Color.Unspecified,
    borderDisabled: Color = background,
    borderSelected: Color = IntUiLightTheme.colors.gray(11),
    borderSelectedActivated: Color = IntUiLightTheme.colors.blue(4),
    borderPressed: Color = IntUiLightTheme.colors.gray(11),
    borderHovered: Color = IntUiLightTheme.colors.gray(12),
backgroundFocused: Color = backgroundHovered,
```

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
```diff
@@ -92,6 +130,34 @@ public fun IconButtonColors.Companion.dark(
         borderHovered = borderHovered,
     )
 
+public fun IconButtonColors.Companion.darkWithoutBackground(
```
DanielSouzaBertoldi: Same questioning from previous comment

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/IconButtonStyling.kt
```diff
@@ -273,3 +273,8 @@ public class IconButtonMetrics(
 public val LocalIconButtonStyle: ProvidableCompositionLocal<IconButtonStyle> = staticCompositionLocalOf {
     error("No IconButtonStyle provided. Have you forgotten the theme?")
 }
+
+public val LocalIconButtonStyleWithoutBackground: ProvidableCompositionLocal<IconButtonStyle> =
```
DanielSouzaBertoldi: By the way, I was divided between `IconButtonStyleWithoutBackground` or `BackgroundlessIconButtonStyle`.

Feel free to suggest other names if you think of something better!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
rock3r: No, no, this is the right approach — no point in creating all new stuff for just some colours that change. It's just a stylistic variant, like DefaultButton/OutlinedButton!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
```diff
@@ -57,6 +67,34 @@ public fun IconButtonColors.Companion.light(
         borderHovered = borderHovered,
     )
 
+public fun IconButtonColors.Companion.lightWithoutBackground(
+    foregroundSelectedActivated: Color = IntUiLightTheme.colors.gray(14),
+    border: Color = Color.Transparent,
```
rock3r: I'd use `Color.Unspecified` instead :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
rock3r: No borders either, this is fine. I'm debating whether we should still expose the bg colours, just making them all Unspecified by default, just so that if folks want to customise somehow they can? Or maybe remove the borders as well, and just make them hardcoded to `Unspecified` like we do for backgrounds. WDYT?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
rock3r: Same answer :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/IconButton.kt
```diff
@@ -96,6 +97,27 @@ public fun IconButton(
     )
 }
 
+@Composable
+public fun BackgroundlessIconButton(
```
rock3r: I don't think we need this tbh, users can just pass a different style to the existing composable and the rest doesn't change

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/IconButton.kt
```diff
@@ -96,6 +97,27 @@ public fun IconButton(
...
+        enabled = enabled,
+        focusable = focusable,
+        style = style,
+        interactionSource = interactionSource,
+        content = content,
+    )
+}
+
 @Composable
 public fun SelectableIconButton(
```
rock3r: I'm wondering whether we should have selectable and toggleable background-less variants, but most likely, no. You still need a bg to show when they're selected/checked, and they do not have a bg otherwise...

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/IconButtonStyling.kt
rock3r: Perhaps we could call this `TransparentIconButtonStyle` or `NoBackgroundIconButtonStyle`? They're a bit of a mouthful, but we don't have better alternatives that I can think of.

rock3r: Oh, also, mark all the new API as experimental — we may decide we don't actually want these styles at some point, and until people use them consistently, I don't want to take on the maintenance burden.

rock3r (commented):
Thanks! Few notes

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
DanielSouzaBertoldi: I mean, `Color.Transparent` is more in line with the purpose of these functions, no? Unspecified to me kinda reads like "no color provided" instead of "we want transparent border"

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
DanielSouzaBertoldi: Hmmmm exposing background colors would go against the purpose of the function as well to me :( Like, the name is specifically called `WithoutBackground`, so I would assume the API wouldn't let me change the background anyways

Hardcoding borders is more ok to me. If the default implementation is for borders to mimick the background color, then it would make sense to get rid of the borders as well for this case 😈

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/IconButton.kt
DanielSouzaBertoldi: True that!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/IconButtonStyling.kt
DanielSouzaBertoldi: +1 for `TransparentIconButtonStyle`

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
rock3r: Not a big deal, we can always change it later if we find out that transparent is more expensive than unspecified (e.g., we miss `takeOrElse` behaviour, or there are optimisations that skip unspecified but not transparent)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiIconButtonStyling.kt
rock3r: Then let's hardcode all the things, YOLO

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/IconButton.kt
DanielSouzaBertoldi: Sorry, forgot to respond to this comment yesterday. But yeah, those should have some sort of visual indication. I think it's safe to just have icon buttons without bg for now

faogustavo (approved):

rock3r:
Ready to merge

PR #3131, opened 2025-07-16
[JEWEL-901] Fixing popup destruction on clicking outside

faogustavo:
- When clicking outside of the popup, the JBPopup was getting destroyed, but the composable node was still present;
- As the composable node was still there, it was requiring an extra click to properly destroy it and be able to show again;

| When | Recording |
| --- | --- |
| Before | Screen Recording 2025-07-16 at 19 15 50
| After | Screen Recording 2025-07-16 at 19 25 15

### Bug fixes
 * Fixed an issue in the JBPopup API where the popup was getting destroyed, but the node wasn't. It was resulting in the need for two clicks to open the popups again

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
@@ -49,7 +49,6 @@ import java.awt.event.KeyEvent.KEY_LOCATION_STANDARD
 import java.awt.event.KeyEvent.KEY_LOCATION_UNKNOWN
 import java.awt.event.KeyEvent.KEY_PRESSED
 import java.awt.event.KeyEvent.KEY_RELEASED
-import java.awt.event.MouseEvent
 import org.jetbrains.jewel.bridge.JewelComposePanelWrapper
 import org.jetbrains.jewel.bridge.LocalComponent
 import org.jetbrains.jewel.bridge.compose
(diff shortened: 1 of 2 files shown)


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
```diff
@@ -223,7 +223,7 @@ private fun Dimension.withSystemDensity(owner: Component): Dimension {
  * This function converts the point relative to the component's current monitor screen coordinates.
  *
  * @return An `IntRect` representing the rectangle in the component's current monitor screen coordinates to be used in
- *   the `popupPositionProvider.calculatePosition` call.
+ *    the `popupPositionProvider.calculatePosition` call.
```
rock3r: Nit: was this needed?

rock3r (approved):

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
faogustavo: I think it was the pre-push job, I haven't changed that LOL. Let me revert it

faogustavo: Actually it was the IDE formatting (cmd + shift + L) when open in the root project 🤔 

Reverted it anyway

rock3r:
Ready to merge

PR #3135, opened 2025-07-20
Map GitHub issue labels to Task types for Task Tracking.

jaxzin:
### Summary

This change extends GitHub issue tracking integration to better infer `TaskType` based on default GitHub labels. Previously, **all GitHub issues were treated as `TaskType.BUG`**, represented with a 🐞 icon in the IDE's Task Tracking UI.

### What's Changed

Tasks are now categorized based on their GitHub labels:

```text
Github Label  → IntelliJ TaskType
----------------------------------
"bug"         → TaskType.BUG
"enhancement" → TaskType.FEATURE
(any other)   → TaskType.OTHER
```

* The label-matching logic is case-sensitive.
* Only default GitHub labels are supported.
* Custom mappings (e.g., for `TaskType.EXCEPTION`) are not currently implemented.

#### Additional Changes
Unit tests were added to the github-tracker module for the first time, so I've included the project and Bazel build configurations associated with adding the unit tests.

### Rationale

Improves issue-type accuracy in the task UI without requiring any custom configuration from the user.

### Testing

New unit tests are included to validate behavior across:

* No label
* Single mapped label
* Single unmapped label
* Multiple labels with mixed mapping
* Multiple mapped labels (first match wins)

All tests pass successfully on macOS ARM64.

Example test run:

```bash
❯ ./bazel.cmd test //plugins/github/github-tracker:vcs-github-tracker_test \
  --test_output=all --nocache_test_results
...
Test plan finished with 6 tests.
...
PASSED in 1.8s
Executed 1 out of 1 test: 1 test passes.
```

### DIFF
--- a/plugins/github/github-tracker/BUILD.bazel
+++ b/plugins/github/github-tracker/BUILD.bazel
@@ -28,3 +28,36 @@ jvm_library(
   runtime_deps = [":vcs-github-tracker_resources"]
 )
 ### auto-generated section `build intellij.vcs.github.tracker` end
+
+### auto-generated section `test intellij.vcs.github.tracker` start
+load("@community//build:tests-options.bzl", "jps_test")
+
+jvm_library(
+  name = "vcs-github-tracker_test_lib",
+  srcs = glob(
+    [
+      "src/**/*.kt",
+      "src/**/*.java",
+      "test/**/*.kt",
+      "test/**/*.java",
+    ],
+    allow_empty = True,
+  ),
+  associates = [":vcs-github-tracker"],
+  visibility = ["//visibility:private"],
+  deps = [
+    "//platform/testFramework",
+    "//platform/testFramework:testFramework_test_lib",
+    "@lib//:assert_j",
+    "@lib//:junit5",
+  ],
+)
+
+jps_test(
+  name = "vcs-github-tracker_test",
+  runtime_deps = [
+    ":vcs-github-tracker",
+    ":vcs-github-tracker_test_lib",
+  ],
+)
+### auto-generated section `test intellij.vcs.github.tracker` end
--- a/plugins/github/github-tracker/src/org/jetbrains/plugins/github/tasks/GithubRepository.java
+++ b/plugins/github/github-tracker/src/org/jetbrains/plugins/github/tasks/GithubRepository.java
@@ -23,6 +23,7 @@ import org.jetbrains.plugins.github.api.GithubServerPath;
 import org.jetbrains.plugins.github.api.data.GithubIssue;
 import org.jetbrains.plugins.github.api.data.GithubIssueBase;
 import org.jetbrains.plugins.github.api.data.GithubIssueCommentWithHtml;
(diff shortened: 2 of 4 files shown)


plugins/github/github-tracker/BUILD.bazel
```diff
@@ -27,4 +27,37 @@ jvm_library(
...
+
+### auto-generated section `test intellij.vcs.github.tracker` start
+load("@community//build:tests-options.bzl", "jps_test")
+
+jvm_library(
+  name = "vcs-github-tracker_test_lib",
+  srcs = glob(
+    [
+      "src/**/*.kt",
+      "src/**/*.java",
```
jaxzin, line 40: GithubRepository is package-private. I'm brand new to Bazel and the only way I found to test a package-private class is to include it's compilation directly into the *_test_lib module instead of depending on the `vcs-github-tracker` module where GithubRepository resides. I didn't want to include any additional refactor of the existing code to improve testability beyond implementing the new feature, like converting it to Kotlin or increasing the class or field visibility.

plugins/github/github-tracker/test/org/jetbrains/plugins/github/tasks/GithubRepositoryTest.java
```diff
@@ -0,0 +1,129 @@
...
+   * @throws InvocationTargetException if the method invocation fails
+   * @throws IllegalAccessException    if access to the method is denied
+   */
+  private static Task createTask(GithubRepository repo, GithubIssueBase issue, List<GithubIssueCommentWithHtml> comments)
+    throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+    Method createTask = GithubRepository.class.getDeclaredMethod("createTask", GithubIssueBase.class, List.class);
+    createTask.setAccessible(true);
+    Task task = (Task) createTask.invoke(repo, issue, comments);
+    return task;
+  }
```
jaxzin, line 127: Did some relatively hacky things in the name of unit testing to instantiate and access the `org.jetbrains.plugins.github.tasks.*` classes and methods, such as creating anonymous subclasses and using reflection to circumvent method accessibility. This was intentional to avoid refactoring the code more than was needed for the intent of this pull request, while still providing unit testing of the changed logic.

plugins/github/github-tracker/src/org/jetbrains/plugins/github/tasks/GithubRepository.java
```diff
@@ -203,7 +204,21 @@ public String getDescription() {
 
       @Override
       public @NotNull TaskType getType() {
-        return TaskType.BUG;
+        List<GithubIssueLabel> labels = issue.getLabels();
+        if (labels == null || labels.isEmpty()) return TaskType.OTHER;
+
+        // Map GitHub's default labels to TaskType
+        for (GithubIssueLabel label : labels) {
+          String name = label.getName();
```
clemairej, line 212: Probably a good idea to lowercase this, to include slightly more variations of the label.

PR #3136, opened 2025-07-21
[JEWEL-892] Add Detekt rule to check API annotations

rock3r:
This introduces a new Detekt rule, `MissingApiStatusAnnotationRule`, that enforces the presence of matching ApiStatus annotations for Jewel's API visibility annotations.

The rule ensures that:
* `@InternalJewelApi` is always paired with `@ApiStatus.Internal`
* `@ExperimentalJewelApi` is always paired with `@ApiStatus.Experimental`

This makes sure we do not mistakenly leak APIs to either the platform (by only using Jewel annotations) or standalone (by only using the `ApiStatus` annotations).

This PR also adds a few missing test cases to `EqualityMembersRuleSpec` (coverage is now 100%, except for "bad PSI scenarios" handling code, which is hard to test), and bumps Gradle's IJP target to `252.23892.248` (252 Beta 1).

Note: this also includes addressing all the issues that the new rule flags (which is noisy), and adapting all the API dumps (which is uber noisy) — sorry!

### ⚠️ Important Changes
 * All Jewel internal and experimental APIs are now also annotated with the corresponding `ApiStatus` annotation. This means they'll be correctly identified as such by all JetBrains tooling, including DevKit and Marketplace.


platform/jewel/detekt-plugin/api-dump-unreviewed.txt
```diff
@@ -1,10 +0,0 @@
-f:org.jetbrains.jewel.detekt.JewelRuleSetProvider
```
rock3r, line 1: These were moved to `api-dump.txt`

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/EqualityMembersRule.kt
```diff
@@ -90,10 +91,11 @@ class EqualityMembersRule(config: Config) : Rule(config) {
         if (missingFunctionNames.isNotEmpty()) {
             val functionNames = missingFunctionNames.joinToString(", ")
             report(
-                CodeSmell(
+                CorrectableCodeSmell(
```
rock3r, line 95: For what I understand, if we use this subclass, the IDE plugin can offer fixes in the editor directly.

platform/jewel/ui/api/ui.api
```diff
@@ -3644,6 +3644,7 @@ public final class org/jetbrains/jewel/ui/icons/AllIconsKeys$FileTypes {
 	public final fun getHprof ()Lorg/jetbrains/jewel/ui/icon/IntelliJIconKey;
```
rock3r: Changes to this file are due to the bump in IJP version (but this file is going away anyway)

platform/jewel/detekt-plugin/src/test/kotlin/org/jetbrains/jewel/detekt/EqualityMembersRuleSpec.kt
```diff
@@ -427,10 +539,4 @@ class EqualityMembersRuleSpec {
             assertThat(result).isEqualTo(code)
         }
     }
-
-    private fun EqualityMembersRule.lintAndFix(@Language("kotlin") code: String): Pair<List<Finding>, String> {
```
rock3r, line 431: Extracted to `TestUtil.kt` so it can be used in the new rule's tests, too.

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/MissingApiStatusAnnotationRule.kt
```diff
@@ -0,0 +1,146 @@
...
+    override val issue: Issue =
+        Issue(
+            javaClass.simpleName,
+            Severity.Defect,
+            "Reports Jewel API annotations that are not paired with a corresponding ApiStatus annotation.",
+            Debt.FIVE_MINS,
+        )
+
+    private val annotationPairs =
+        listOf("InternalJewelApi" to "ApiStatus.Internal", "ExperimentalJewelApi" to "ApiStatus.Experimental")
```
faogustavo, line 43: Is there a reason to start with a list rather than a map directly? It will probably not make a significant difference, but it would prevent the '.toMap' operation inside a loop

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/EqualityMembersRule.kt
```diff
@@ -30,7 +31,7 @@ import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
  * This rule also supports auto-correction. It can generate the missing or incorrect functions, ensuring they are
  * correctly implemented based on the class's properties.
  */
-class EqualityMembersRule(config: Config) : Rule(config) {
+class EqualityMembersRule(private val config: Config) : Rule(config) {
```
faogustavo: Was the addition of the `private val` intended? We are not using it in this scope

faogustavo (approved):
No blockers. All good to go

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/MissingApiStatusAnnotationRule.kt
rock3r: It's mostly because the "main" usage in detection uses pairs rather than key-value, so semantically speaking this seemed more correct to me.

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/EqualityMembersRule.kt
rock3r: Ah, I had used it at some point to access the auto correct flag, only to realise I didn't need to... I'll revert

rock3r:
I just pushed a _large_ change as I realised the rule wasn't actually running. Then I found an issue in the autofix code that tests had not caught, so I also had to update all the tests.

All the changes to API dumps can be safely ignored as they were autogenerated; there is no real change, except that now internal APIs are excluded from dumps, and experimental APIs are in a separate dump file. This alone cuts in ~half the changes in this PR.

Then, of course, there are all the places where annotations were missing and have been added now. You can safely ignore those as well, since the changes are mechanical and validated by the rule.

rock3r:
Oh and another thing — for whatever reason, the Detekt auto-correct doesn't seem to do anything in the IDE. Since in tests it works, I will chalk it up to a bug in the Detekt IJ Plugin...

And if you want to get the rules to show in the IDE too, you can add the rules JAR to the aforementioned plugin:

image

(ignore the Compose rules JAR, those will come in a separate PR as JEWEL-920, I'm just trialling locally)

DanielSouzaBertoldi (approved):
Great job! 🚀 Didn't know you could make Detekt "automatically" fix the code, pretty neat

rock3r:
It'd be even better if it worked lol

The command appears but doesn't do anything. I'll have to file a bug at some point

rock3r:
Ready to merge

PR #3138, opened 2025-07-21
[JEWEL-906] Mandate valid JEWEL issue ID in single commit messages

rock3r:
This adds a check to GitHub Actions that validates:
 1. The PR has only one commit
 2. The commit message starts with "[JEWEL-xxx] ", where xxx is a positive number

These checks only run if the PR contains changes in the Jewel folder other than `AllIconsKeys`. That gets autogenerated by a script when any changes are made to IJP icons, and is high-traffic. Any IJP dev can run the script, and it should not block PRs added elsewhere in the IJP that update the icons.

Additionally, if there is a `JEWEL_YT_TOKEN` secret defined, the check also validates that the referenced issue is valid — meaning, it exists, is not resolved, and not a draft.

### DIFF
--- a/.github/workflows/jewel_checks.yml
+++ b/.github/workflows/jewel_checks.yml
@@ -11,16 +11,18 @@ defaults:
 
 jobs:
   checks:
+    name: CI code checks
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v4
+        name: Check out repository
 
-      - name: Set up JDK 21
+      - name: Set up JBR 21
         uses: actions/setup-java@v4
         with:
           java-version: 21
-          distribution: zulu
+          distribution: jetbrains
           cache: gradle
 
       - name: Grant execute permission for gradlew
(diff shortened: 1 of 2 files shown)


.github/workflows/jewel_checks.yml
```diff
@@ -37,3 +39,54 @@ jobs:
...
+      - uses: actions/checkout@v4
+        name: Check out repository
+
+      - uses: dorny/paths-filter@v3
+        name: Check which files are changed
+        id: filter
+        with:
+          filters: |
+            run_formalities:
+              - '!all(platform/jewel/ui/generated/org/jetbrains/jewel/ui/icons/AllIconsKeys.java)'
```
faogustavo, line 61: Should we add some condition to ensure we only run this when the user changes jewel files? Otherwise, it would run for all PRs, right?


```suggestion
      - uses: dorny/paths-filter@v3
        name: Check which files are changed
        id: filter
        with:
          predicate-quantifier: 'every'
          filters: |
            run_formalities:
              - 'platform/jewel/**'
              - '!all(platform/jewel/ui/generated/org/jetbrains/jewel/ui/icons/AllIconsKeys.java)'
```

faogustavo: Does it make sense to simplify to skip any generated folders inside jewel?

```
- '!all(platform/jewel/**/generated/**)'
```

rock3r: > Should we add some condition to ensure we only run this when the user changes jewel files? Otherwise, it would run for all PRs, right?

That filtering has always been applied — at the [workflow level](https://github.com/rock3r/intellij-community/blob/2b0c02555af38d976bb06e1360646fdeef6b1413/.github/workflows/jewel_checks.yml#L6) tho.

> Does it make sense to simplify to skip any generated folders inside jewel?

We don't have any others, and I'd rather err on the side of caution than find out someone has added some new autogenerated file we don't know about :)

faogustavo: Missed the filter in the top level lol

rock3r:
Ready to merge

.github/workflows/jewel_checks.yml
```diff
@@ -37,3 +39,54 @@ jobs:
...
+            exit 1
+          fi
+
+      - name: Grant execute permission for validation script
+        run: chmod +x validate-commit-message.sh
+
+      - name: Validate commit message
+        env:
+          JEWEL_YT_TOKEN: ${{ secrets.JEWEL_YT_TOKEN }}
+        run: ./validate-commit-message.sh "${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }}"
```
DanielSouzaBertoldi, line 92: Hmmm this got me thinking: why not check if the PR description contains the `## Release notes` string too? I'm not familiar at all with GitHub actions, but judging by a quick search it's viable to do so.

I know this is outside the scope of this PR since it's focused on commit messages and we would need to update the release notes extractor to account for PRs that don't need to have release notes.

Maybe have the extractor ignore any:

```
## Release notes
!NONE
```

Or something. But anyways, just food for thought!

DanielSouzaBertoldi (approved):

.github/workflows/jewel_checks.yml
rock3r: It's because not all PRs require release notes :)

rock3r: It's just easier to check when reviewing if they're missing

PR #3139, opened 2025-07-21
[JEWEL-899] Remove old Gradle plugins and tasks

rock3r:
This removes a number of now-unused Gradle plugins and tasks:
 * `ide-version-checker` (not used anymore)
 * `jewel-publish`, `PublishConfiguration`, and related tasks (publishing is handled by the JPS build)
 * `MergeSarifTask` (not used anymore) and related tasks
 * Related run configs

Note: `icon-keys-generator` has not been removed yet, even though the actual generator runs as part of the IJP's generation, since the Gradle build would remain without the generated file otherwise. Since we're planning to eventually sunset the Gradle build, this is only a temporary necessity.


faogustavo (approved):

rock3r:
Random CI failure due to an HTTP 502, rebased & pushed again

rock3r:
Ready to merge

PR #3141, opened 2025-07-22
[JEWEL-910] Fix GitHub CI to use the JBR

rock3r:
The CI stopped working seemingly overnight due to Gradle needing the JetBrains Runtime, and the agents not having it. This makes the agents use the JBR runtime instead of `zulu`, and updates the foojay plugin to the latest version (as the plugin should be able to download the JBR if missing anyway, so maybe that stopped working?).

This also bumps the Kotlin version in the Gradle build to match IJP 252, and replaces the deprecated Gradle Enterprise plugin with the new Develocity one.

### DIFF
--- a/.github/workflows/jewel_checks.yml
+++ b/.github/workflows/jewel_checks.yml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           java-version: 21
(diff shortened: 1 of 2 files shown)


rock3r:
Ready to merge

PR #3142, opened 2025-07-22
[JEWEL-916] Fix crash in UiInspectorInfo

rock3r:
On master, invoking the UI inspector when some (but not all!) Jewel UI is attached can make the UI inspector crash. This is because our own `UiInspectorInfo` is assuming all a11y related API return non-null values, which clearly was a valid assumption until recently.

However, now some of those APIs do return null values, so we need to treat them accordingly. This fixes the issue, and cleans up/simplifies the code a bit in the process.

### Bug fixes
 * Fixed a crash in UI Inspector caused by unexpected null values in `AccessibleContext`


faogustavo (approved):

rock3r:
Ready to merge

DanielSouzaBertoldi (approved):

PR #3144, opened 2025-07-22
[JEWEL-911] Fix infoForeground colour in Darcula theme

rock3r:
In the Darcula theme, for some reason, `infoForeground` is only defined in `Component`, instead of being in the wildcard "*" node. This results in the info text in this theme being displayed as normal text.

This fixes the issue by moving all `BridgeGlobalColors` to use various IJP known colour APIs instead of trying to read specific keys. These APIs are "guaranteed" to work — or at least to break in the same ways as Swing would.

It also adds two new colours to TextColors: disabledFocused and warning.


faogustavo (approved):

rock3r:
Ready to merge

PR #3145, opened 2025-07-22
[JEWEL-897] Support Rendering Resource Images in Markdown

DanielSouzaBertoldi:
# Context

Our Markdown Image renderer only renders URLs, but it should also render images from the resources folder. This PR adds support for that + adds unit tests and refactors the coil renderer code a bit.

## Main Changes
- Added `coil-test` to `libs.versions.toml` and `platform/jewel/markdown/extensions/images/build.gradle.kts`
- Created a `resolveImageSource` function in `Coil3ImagesRendererExtensionImpl.kt` that tries to convert the raw destination string provided from `CMImage` to a Coil `URI`. If unsuccessful, we try to fetch the image from the resources folder root.
- Refactored `Coil3ImagesRendererExtensionImpl` to not have the AsyncImage inside the placeholder tell the placeholder size, but actually delegate the calculation to `rememberAsyncImagePainter`. Also now we provide placeholders according to the `imageResult` size. This also made it easier to test the code.
- Added unit tests that test if the Placeholder size is the one expected given the scenario (image was loaded successfully/some error occurred/is loading then loads successfully)
- Added unit tests for the `resolveImageSource` function

# Evidences

## Using an URL
| Sample | Before | After |
|-------|--------|--------|
| Standalone |  |  |
| Plugin |  |  |

## Using local image inside the `resource` directory
| Sample | Before | After |
|--------|--------|--------|
| Standalone |  |  |
| Plugin |  |  |

## Live editing image paths

https://github.com/user-attachments/assets/4c98bbd7-a2e5-4d53-a0e9-edb89d233926

### ⚠️ Important Changes
 * Experimental API `renderImagesContent` renamed to `renderImageContent` (singular).

### New features
 * You can now include local images in your Markdown content. Simply add your image files (like PNGs, JPGs, or SVGs) to your `src/main/resources` folder and reference them directly. For example, to display `my-logo.png` located in `src/main/resources/images/`, you would write: `My Logo`


platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -36,48 +37,70 @@ import org.jetbrains.jewel.ui.component.Tooltip
...
+            JewelLogger.getInstance("Jewel")
+                .warn(
+                    "Markdown image '$rawDestination' expected at classpath '$rawDestination' but not found. " +
+                        "Please ensure it's in your 'src/main/resources/' folder."
                 )
-            }
+            return rawDestination // This will cause Coil to fail and not render anything.
         }
+
+        return resourceUrl.toExternalForm()
```
DanielSouzaBertoldi: This will return a jar:file://<path> string, which Coil3 can handle using its [JarFileFetcher](https://github.com/coil-kt/coil/blob/main/coil-core/src/jvmCommonMain/kotlin/coil3/fetch/JarFileFetcher.kt)

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -36,48 +37,70 @@ import org.jetbrains.jewel.ui.component.Tooltip
...
+
+        if (uri.scheme != null) return rawDestination
+
+        val resourceUrl = this@Coil3ImageRendererExtensionImpl::class.java.classLoader.getResource(rawDestination)
+
+        if (resourceUrl == null) {
+            JewelLogger.getInstance("Jewel")
+                .warn(
+                    "Markdown image '$rawDestination' expected at classpath '$rawDestination' but not found. " +
+                        "Please ensure it's in your 'src/main/resources/' folder."
```
DanielSouzaBertoldi: This will be logged for each key press when the user is editing a local image path that's not valid, so it may pollute logs. Maybe it's not worth it to log it?

platform/jewel/markdown/extensions/images/build.gradle.kts
```diff
@@ -14,6 +14,8 @@ dependencies {
     implementation(projects.markdown.core)
     runtimeOnly(libs.ktor.client.java)
     testImplementation(compose.desktop.uiTestJUnit4)
+    testImplementation(libs.coil.test)
```
rock3r, line 16: Make sure this is also added in the JPS build if you still need it!

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
rock3r: You could move the check to the composable and wrap the log call in a remember/launched effect keyed on `resourceUrl` and `rawDestination`, should cut down on the spam I think...

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -36,48 +37,70 @@ import org.jetbrains.jewel.ui.component.Tooltip
...
-                            ?: Modifier) // Otherwise, we'll do it when we have it
-                            .pointerHoverIcon(PointerIcon.Default),
+
+        return InlineTextContent(placeholder) {
+            Image(painter = painter, contentDescription = image.title, modifier = Modifier.fillMaxSize())
+        }
+    }
+
+    @VisibleForTesting
+    internal fun resolveImageSource(rawDestination: String): String {
```
rock3r: What about extracting it to a collaborator? In the future we may even consider making it an interface with a default impl that behaves like the current impl and lets users swap it (e.g. via a composition local?)

DanielSouzaBertoldi: The future is now 😎 [is this what you meant](https://github.com/JetBrains/intellij-community/pull/3145/commits/26718980e6e3b27a1e4e0bb627a3e7983ec7b27d)?

btw, just realized I also have to add the `@ApiStatus.Experimental` annotation. Will do that in the next commit!

.idea/libraries/io_coil_kt_coil3_test.xml
```diff
@@ -0,0 +1,56 @@
+<component name="libraryTable">
```
DanielSouzaBertoldi: So, I couldn't find the coil-test library so I had to add it. Since it's outside the jewel project scope, this should be added in a separate PR?

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/JewelReadme.kt
```diff
@@ -9,7 +9,7 @@ internal val JewelReadme: String =
 
 # Jewel: a Compose for Desktop theme
 
-![Jewel logo"]("platform/jewel/art/jewel-logo.svg)
+![Jewel logo](readme/jewel-logo.svg)
```
faogustavo, line 12: Nit: Do we need to keep this old asset?

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/ImageSourceResolver.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        return resourceUrl.toExternalForm()
+    }
+}
+
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public val LocalMarkdownImageSourceResolver: ProvidableCompositionLocal<ImageSourceResolver> =
+    staticCompositionLocalOf {
+        error("No LocalMarkdownImageSourceResolver provided. Have you forgotten the theme?")
```
faogustavo, line 60: How about providing an instance of `DefaultImageSourceResolver` as the default instead of throwing an error?

.idea/libraries/io_coil_kt_coil3_test.xml
rock3r: Please remove it and re-add it as a module-only dependency:

Image

platform/jewel/.idea/runConfigurations/Stand_alone_sample.xml
```diff
@@ -20,4 +20,4 @@
     <DebugAllEnabled>false</DebugAllEnabled>
     <method v="2" />
   </configuration>
-</component>
\ No newline at end of file
+</component>
```
rock3r: Please revert

platform/jewel/markdown/core/api-dump-experimental.txt
```diff
@@ -207,7 +207,7 @@ f:org.jetbrains.jewel.markdown.SemanticsKt
 *:org.jetbrains.jewel.markdown.WithTextContent
 - a:getContent():java.lang.String
 *:org.jetbrains.jewel.markdown.extensions.ImageRendererExtension
-- a:renderImagesContent(org.jetbrains.jewel.markdown.InlineMarkdown$Image,androidx.compose.runtime.Composer,I):androidx.compose.foundation.text.InlineTextContent
+- a:renderImageContent(org.jetbrains.jewel.markdown.InlineMarkdown$Image,androidx.compose.runtime.Composer,I):androidx.compose.foundation.text.InlineTextContent
```
rock3r: Make sure to mention the experimental API change in the important changes section of the PR release notes

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -1,83 +1,95 @@
...
+ * An [ImageRendererExtension] that uses Coil 3 to render images from Markdown.
  *
- * This implementation uses [AsyncImage] to load and display images. It reserves space for the image with a
- * [Placeholder] and updates its size upon successful loading to prevent layout shifts.
+ * This implementation is designed to be robust, handling image loading asynchronously and resizing the layout correctly
+ * upon success. It uses a callback-driven approach to update its state, which avoids common race conditions and
+ * lifecycle issues in Compose.
  *
- * @param imageLoader A custom [ImageLoader] for Coil3 image requests.
+ * @param imageLoader The Coil [ImageLoader] instance to be used for all image requests.
```
rock3r: Nit:
```suggestion
 * @param imageLoader The Coil [ImageLoader] instance used for all image requests.
```

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -1,83 +1,95 @@
...
+                imageLoader = imageLoader,
+                onSuccess = { successState ->
+                    // onSuccess should only be called once, but adding additional protection from
+                    // unnecessary rerender
+                    if (imageResult == null) {
+                        imageResult = successState.result
+                    }
+                },
+                onError = { error ->
+                    JewelLogger.getInstance("Jewel").warn("AsyncImage error: ${error.result.throwable}")
```
rock3r:
```suggestion
                    JewelLogger.getInstance(Coil3ImageRendererExtensionImpl.javaClass).warn("AsyncImage error: ${error.result.throwable}")
```

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/ImageSourceResolver.kt
rock3r: I think it's ok to expect the theme to provide these as it may be different in standalone and bridge

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
```diff
@@ -0,0 +1,137 @@
...
+import coil3.test.FakeImageLoaderEngine
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.test.StandardTestDispatcher
+import org.jetbrains.jewel.markdown.InlineMarkdown
+import org.junit.Rule
+import org.junit.Test
+
+@OptIn(ExperimentalCoilApi::class)
+public class Coil3ImageRendererExtensionImplTest {
+    @get:Rule public val composeTestRule: ComposeContentTestRule = createComposeRule()
```
rock3r, line 35: Nit:

```suggestion
    @get:Rule public val composeTestRule = createComposeRule()
```

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        val imageMarkdown =
+            InlineMarkdown.Image(source = imageUrl, alt = "Alt text", title = "Image loaded successfully")
+
+        setContent(extension, imageMarkdown)
+
+        val imageNode = composeTestRule.onNodeWithContentDescription("Image loaded successfully")
+        imageNode.assertExists()
+
+        imageNode.assertWidthIsEqualTo(fakeImageWidth.dp)
+        imageNode.assertHeightIsEqualTo(fakeImageHeight.dp)
```
rock3r: Nit — in other places too:


```suggestion
        composeTestRule.onNodeWithContentDescription("Image loaded successfully")
            .assertExists()
            .assertWidthIsEqualTo(fakeImageWidth.dp)
            .assertHeightIsEqualTo(fakeImageHeight.dp)
```

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/DefaultImageSourceResolverTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public fun `resolveImageSource resolves existing classpath resource`() {
+        // This test requires a file named `test-image.svg` to exist in `src/test/resources/`.
+        val defaultImageSourceResolver = DefaultImageSourceResolver()
+        val resourceName = "test-image.svg"
+
+        val result = defaultImageSourceResolver.resolve(resourceName)
+
+        assert(result.startsWith("file:/"))
```
rock3r, line 26:
```suggestion
        assert(result.startsWith("file://"))
```

It should check for the file schema prefix, which ends with the second forward slash. Same below

platform/jewel/markdown/extensions/images/intellij.platform.jewel.markdown.extensions.images.iml
```diff
@@ -209,6 +209,7 @@
         </SOURCES>
       </library>
     </orderEntry>
+    <orderEntry type="library" exported="" name="io.coil.kt.coil3.test" level="project" />
```
rock3r: You need to mark it as a TEST dependency, or it'll end up in the production classpath:

Image

platform/jewel/markdown/int-ui-standalone-styling/intellij.platform.jewel.markdown.intUiStandaloneStyling.iml
```diff
@@ -40,6 +38,7 @@
     <orderEntry type="module" module-name="intellij.platform.jewel.intUi.standalone" />
     <orderEntry type="module" module-name="intellij.platform.jewel.markdown.extensions.gfmAlerts" />
     <orderEntry type="module" module-name="intellij.platform.jewel.markdown.extensions.gfmTables" />
+    <orderEntry type="module" module-name="intellij.platform.jewel.markdown.extensions.images" />
```
rock3r: I just realised that these should be marked as provided dependencies, not as compiled (see [this](https://www.jetbrains.com/help/idea/working-with-module-dependencies.html#z5dxrlw_51) for info). We should not add all the dependencies as the equivalent of `implementation` for Gradle, they should be `compileOnly`.

Same in the bridge...

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/JewelReadme.kt
rock3r: It's to test loading from resources I think

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
rock3r: Even better:

```kotlin
JewelLogger.getInstance(this.javaClass)
    .warn("AsyncImage loading failed.", error.result.throwable)
```

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/JewelReadme.kt
DanielSouzaBertoldi: Exactly. In my mind it serves as an example that you can use images from resources :)

platform/jewel/.idea/runConfigurations/Stand_alone_sample.xml
DanielSouzaBertoldi: This happens EVERY time dammnnnnn 😭

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
DanielSouzaBertoldi: Can't remove the type due to error `Return type must be specified in explicit API mode.` :(

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/JewelReadme.kt
faogustavo: My bad, I saw "readme" and thought it was the readme.md file 🤦

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
DanielSouzaBertoldi: The thing is that everytime the user types/deletes a character, the `renderImageContent` will be called again, which in turn means `rawDestination` (now `image.source`) will always trigger the remember/launched effect also :( but I guess this "flood" of logs should be expected when adding/changing the path/url of an image

platform/jewel/markdown/int-ui-standalone-styling/intellij.platform.jewel.markdown.intUiStandaloneStyling.iml
DanielSouzaBertoldi: Changed them to be provided instead of compiled ✅

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
rock3r: That's enabled in tests, too? Uh, ok

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/DefaultImageSourceResolverTest.kt
DanielSouzaBertoldi: Actually, Java's URL class returned by `getResource()`during test runs returns only contains the scheme `file:` and the path (that starts with `/`). There is no "authority" (`//`) since there's no server hostname. I added some print statements to demonstrate this:

```
# getResource call result
resourceUrl: file:/Users/me/intellij-community/out/test/intellij.platform.jewel.markdown.extensions.images/test-image.svg

# .toExternalForm() result
external form: file:/Users/me/intellij-community/out/test/intellij.platform.jewel.markdown.extensions.images/test-image.svg
```

rock3r: Oh! TIL. I stand corrected then :)

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
DanielSouzaBertoldi: It's enabled in the `images` module, according to the BUILD.bazel file. The `ui-tests` module doesn't have this line: `x_explicit_api_mode = "strict"` so we don't have this sort of problem there

faogustavo (approved):

rock3r:
Ready to merge

platform/jewel/markdown/extensions/images/src/main/resources/readme/jewel-logo.svg
rock3r, line 1: You need to move this to the standalone sample resources, or the build will fail

jakub-senohrabek:
If I am not mistaken you should not bring your own ktor and instead use the one from intellij.libraries.ktor.client

DanielSouzaBertoldi:
@jakub-senohrabek thanks for pointing that out, didn't know that! I changed so that we depend on `intellij.libraries.ktor.client` and everything works as normal :)

PR #3148, opened 2025-07-23
[JEWEL-918] Fix checkbox and radio button icons in Darcula

rock3r:
The Darcula theme is now considered "new UI". Our check to adopt the "old themes" metrics (that consider that the Darcula SVGs are borked) thus doesn't work any more. The fix is to explicitly adopt the "old theme" metrics when the current theme is Darcula.

 Before | After
 --- | ---
 before | after

The fix is not 100% perfect, but our old metrics are "good enough" for now. Swing is faring [a lot worse than this](https://youtrack.jetbrains.com/issue/JEWEL-918/Checkboxes-and-radio-buttons-in-the-Darcula-theme-look-wrong) anyway, and we should go back to polish only after the Swing implementation is fixed.

This only impacts the bridge, since we do not support Darcula in the standalone mode.

### Bug fixes
 * Fixed the checkbox and radio button appearance in the IDE when using the Darcula theme

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
@@ -311,6 +311,8 @@ internal fun lafName(): String {
     return lafInfo.name
 }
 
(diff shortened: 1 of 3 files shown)


rock3r:
Filed JEWEL-919 to track

rock3r:
Ready to merge

PR #3149, opened 2025-07-24
Updated KeyMap for Visual Studio - Added shorts for cloning caret

kentkost:
Added shortcuts that are already present in Visual Studio today.
I believe these shortcuts were added in the 2022 release of Visual Studio. Meaning previous didn't have them and that includes macOS

Cloning carets up or down in Visual Studio is possible via:

`ALT+Shift+ARROWUP` and `ALT+Shift+ARROWDOWN`


jreznot:
Hi,
Unfortunately, we cannot merge it as is, as Rider has a special support for those key binds already.

If we want to improve the experience of users who prefer Visual Studio keymap in other IDEs, we need to move the `CloneCaretAboveWithVirtualSpace` and `CloneCaretBelowWithVirtualSpace` implementatations to the platform.

Let us think how to proceed here

PR #3152, opened 2025-07-28
[JEWEL-935] Make GitHub checks post comments

rock3r:
When there is some issue in a PR found by the GitHub actions workflow, a comment is posted to the PR describing the issue found. This also cleans up existing scripts and collects them in `platform/jewel/scripts` so they do not pollute the rest of the codebase.

Lastly, this uses the GitHub CLI to count the commits in a PR as the
previous method had issues with rebases. If the token has permissions and there is an issue, the actions add a comment to the PR. However, this does not work for fork-based PRs (I may look into a better solution in the future)


DanielSouzaBertoldi (approved):

rock3r:
Ready to merge

PR #3153, opened 2025-07-29
[JEWEL-873] Creating custom popup renderer for stand alone apps

faogustavo:
- Created new Renderer to use JDialog
- Enabled compose interop blending to fix background fill from other windows
- Properly handling keyboard and windows focus events

| Case | Video |
| --- | --- |
| Normal conditions |Screen Recording 2025-07-29 at 07 34 50|
| Render outside window | Screen Recording 2025-07-29 at 07 35 15
| Preventing draw outside of the screen | Screen Recording 2025-07-29 at 07 36 01

### New features
 * Added support for Popups using a native window
   * This feature allows your popups to draw outside of your Compose Panel/Window
   * The implementation is based on JDialog and requires one of the following conditions to work:
     * Use the JetBrains Runtime;
     * Enabled the 'compose.interop.blending' system property;
     * Set the LaF flag 'Panel.background' to a transparent value;
   * If none of the requirements are met, we will fallback to the compose implementation to avoid UI glitches


platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -29,12 +32,18 @@ import org.jetbrains.jewel.samples.standalone.view.TitleBarView
...
 import org.jetbrains.jewel.window.styling.TitleBarStyle
 
 @ExperimentalLayoutApi
 public fun main() {
     JewelLogger.getInstance("StandaloneSample").info("Starting Jewel Standalone sample")
+    JewelFlags.useCustomPopupRenderer = true
+
+    // If we don't enable the blending, the popup will be rendered with a white background
+    // This must be called before the first compose block is executed
+    System.setProperty("compose.interop.blending", "true")
```
faogustavo: This one will need to be documented on our side. Without this flag, the popups will render with a white background

| With Flag | Without flag |
| --- | --- |
| Image | Image

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -77,8 +86,10 @@ public fun main() {
                     processKeyShortcuts(keyEvent = keyEvent, onNavigateTo = MainViewModel::onNavigateTo)
                 },
                 content = {
-                    TitleBarView()
-                    ProvideMarkdownStyling { currentView.content() }
+                    CompositionLocalProvider(LocalPopupRenderer provides JDialogRenderer(window)) {
```
faogustavo: I'm open to ideas on how to provide the window. Compose already has access to it, so they don't need to provide. But I haven't found any local provider for it

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Buttons.kt
```diff
@@ -232,7 +232,7 @@ private fun SplitButtons() {
             content = { Text("Split button") },
             menuContent = {
                 items(
-                    items = listOf("Item 1", "Item 2", "Item 3"),
+                    items = List(10) { "Item ${it + 1}" },
```
faogustavo: Added more items to test overflow

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
faogustavo: Something I was thinking is about adding a check in the `JDialogRenderer` to check if this flag is enabled; otherwise, fall back to the default renderer.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
```diff
@@ -0,0 +1,305 @@
...
+ * hierarchies.
+ *
+ * Note that this renderer requires you to enable the 'compose.interop.blending' system property. Otherwise, the popup
+ * will be rendered with a white background. For more information, check the official documentation available at
+ * https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-desktop-swing-interoperability.html#experimental-interop-blending
+ *
+ * @param window The parent window that will own the popup dialog
+ */
+public class JDialogRenderer(window: Window) : PopupRenderer {
+    private val currentWindow = WeakReference(window)
```
faogustavo, line 70: I'm not sure if the weak ref here is really needed. I still need to review. Probably can be removed as the instance should be released on the compose dispose

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
rock3r: Sounds like a good idea. If the popups flag is on, and we're in standalone (this doesn't seem to be needed in bridge?), and this flag is off, fall back and log a warning.

PS: any pointers as to what `compose.interop.blending` does? I found it was added in https://github.com/JetBrains/compose-multiplatform-core/pull/915 but there may still be some open issues with it
* https://youtrack.jetbrains.com/issue/CMP-4941/Make-SwingPanel-interop-blending-by-default
* https://youtrack.jetbrains.com/issue/CMP-7645/Support-compose.interop.blending-on-OpenGL-Linux

rock3r: Just read the KDoc about the flag — the documentation still doesn't explain everything since it only talks about popups in the abstract, and about `SwingPanel`, which we don't use...

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
rock3r: We should probably do the same we do in the bridge with `JewelComposePanelWrapper.kt`

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
```diff
@@ -0,0 +1,362 @@
...
+ * This renderer creates a JDialog that hosts Compose content, allowing for integration between Swing and Compose UI
+ * hierarchies.
+ *
+ * Note that this renderer requires you to enable the 'compose.interop.blending' system property. Otherwise, the popup
+ * will be rendered with a white background. For more information, check the official documentation available at
+ * https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-desktop-swing-interoperability.html#experimental-interop-blending
+ *
+ * @param window The parent window that will own the popup dialog
+ */
+public class JDialogRenderer(private val window: Window) : PopupRenderer {
```
rock3r: This should be internal like the bridge counterpart is (see the other comment on how to inject this)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
```diff
@@ -0,0 +1,362 @@
...
+            layout = null
+            isOpaque = false
+            background = java.awt.Color(0, 0, 0, 0)
+
+            setContent {
+                ProvideValuesFromOtherContext(compositionLocalContext) {
+                    val positionProvider = currentPopupPositionProvider.value
+                    val parentBounds = parentBoundsInRoot.value ?: return@ProvideValuesFromOtherContext
+
+                    println("parentBounds: $parentBounds")
```
rock3r: I guess this should be removed :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
```diff
@@ -0,0 +1,362 @@
...
+                                }
+                            },
+                    )
+                }
+            }
+        }
+    }
+
+    val dialog = remember {
+        JDialog(window).apply {
```
rock3r: Any reason why we use JDialog and not JPopup? Does JDialog work with "light" popups like tooltips?

rock3r: Mostly meaning, a dialog usually comes with different assumptions than a popup, so I'd like to be sure it's the right choice

faogustavo: So, I haven't found any JPopup; just the JPopupMenu and the Popup from Swing.

* The JPopupMenu is used to build a list of options in a menu format. We can't properly add a compose panel as content to render it (at least I haven't found a way for it);
* The Swing Popup version only exposes methods to show/hide the content. So we would need to already have the position and size of it (which does not happen until the first render);

Also, when I checked the [CfD implementation of 'native windows'](https://github.com/JetBrains/compose-multiplatform-core/blob/2d77505b41717cd8474a5b2751de880f3827104c/compose/ui/ui/src/desktopMain/kotlin/androidx/compose/ui/scene/WindowComposeSceneLayer.desktop.kt#L65-L78), they are using the JDialog class.

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
faogustavo: > this doesn't seem to be needed in bridge?

Yep. I don't know what type of witchcraft the JBPopup is doing, but it just works there 🤷 

> any pointers as to what compose.interop.blending does?

Yes, this is documented for the SwingPanel, but it works with views in interop in general. AFAIK, the interop with compose works by "creating a hole in the rendered content", and putting the compose below it (That's why we see corners - they are the window background). 

But when the blending flag is enabled, we also change the order we place stuff, and put the compose view [above other UI elements](https://github.com/JetBrains/compose-multiplatform-core/blob/2d77505b41717cd8474a5b2751de880f3827104c/compose/ui/ui/src/desktopMain/kotlin/androidx/compose/ui/scene/ComposeSceneMediator.desktop.kt#L162-L166). So with that, render the content above and don't have the corners

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
rock3r: Sorry, I namedropped a JPopup that I had hallucinated, probably thinking of JBPopup :D Cool, was worth checking there was no alternative we hadn't considered — you've done your homework, as usual :)

faogustavo: Just to keep it documented here too, I also tested using the `DialogWindow` composable, but it has the same issues as the official implementation of native windows (using the 'compose.layers.type' flag).

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
faogustavo: ✅

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
faogustavo: ✅

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/component/JDialogRenderer.kt
```diff
@@ -0,0 +1,444 @@
...
+import org.jetbrains.jewel.ui.component.DefaultPopupRenderer
+import org.jetbrains.jewel.ui.component.LocalPopupRenderer
+import org.jetbrains.jewel.ui.component.PopupRenderer
+
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+@Composable
+public fun ProvideNativeWindowPopupRenderer(window: Window, content: @Composable () -> Unit) {
+    CompositionLocalProvider(LocalPopupRenderer provides JDialogRenderer(window)) { content() }
+}
```
faogustavo: As there is no way to get the current window from composition locals (like the LocalActivity or LocalViewController), I had to create this wrapper. If you have other possible solutions, let me know

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
```diff
@@ -37,25 +37,27 @@ public fun PopupContainer(
         popupPositionProvider = popupPositionProvider,
         onDismissRequest = onDismissRequest,
         properties = popupProperties,
-    ) {
-        OverrideDarkMode(style.isDark) {
-            val colors = style.colors
-            val popupShape = RoundedCornerShape(style.metrics.cornerSize)
+        cornerSize = style.metrics.cornerSize,
+        content = {
```
faogustavo: why has this moved to inside? 🤔  let me revert it

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/component/LocalComponent.kt
```diff
@@ -0,0 +1,15 @@
...
+import javax.swing.JComponent
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+
+@Suppress("CompositionLocalAllowlist")
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public val LocalComponent: ProvidableCompositionLocal<JComponent> = staticCompositionLocalOf {
+    error("No Component provided")
+}
```
faogustavo: Should we provide a function that helps with this for the stand-alone users of the library (thinking when people don't use a decorated window)?

```kotlin
fun main() {
    application {
        IntUiTheme {
            DecoratedWindow {
                Text("Foo")
            }
        }
    }
}
```

We could provide a helper function like this:

```kotlin
@Composable
@ApiStatus.Experimental
@ExperimentalJewelApi
public fun ProvideLocalComponentFromWindow(window: ComposeWindow, content: @Composable () -> Unit) {
    CompositionLocalProvider(
        LocalComponent provides window.contentPane.components.filterIsInstance<JComponent>().first(),
        content = content,
    )
}
```

rock3r: As we discussed multiple times, we need the same thing in standalone that we have in bridge (`JewelComposePanelWrapper` functions)

faogustavo: Created the `platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/JewelWindow.kt` for this

platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/DecoratedWindow.kt
```diff
@@ -152,7 +154,10 @@ public fun DecoratedWindow(
                 Modifier
             }
 
-        CompositionLocalProvider(LocalTitleBarInfo provides TitleBarInfo(title, icon)) {
+        CompositionLocalProvider(
+            LocalComponent provides window.contentPane.components.filterIsInstance<JComponent>().first(),
```
rock3r: Nit: this should be fast, but is it worth remembering on `window`?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/component/LocalComponent.kt
```diff
@@ -0,0 +1,15 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.foundation.component
+
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import javax.swing.JComponent
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+
+@Suppress("CompositionLocalAllowlist")
```
rock3r: FYI this should not be needed once JEWEL-920 lands (I disabled that rule as it makes no sense for Jewel)

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/component/LocalComponent.kt
```diff
@@ -0,0 +1,15 @@
...
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public val LocalComponent: ProvidableCompositionLocal<JComponent> = staticCompositionLocalOf {
```
rock3r: Can we have a bit of KDoc?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
```diff
@@ -62,7 +63,8 @@ internal object JBPopupRenderer : PopupRenderer {
         onDismissRequest: (() -> Unit)?,
         onPreviewKeyEvent: ((KeyEvent) -> Boolean)?,
         onKeyEvent: ((KeyEvent) -> Boolean)?,
-        content: @Composable () -> Unit,
+        cornerSize: CornerSize,
+        content: @Composable (() -> Unit),
```
rock3r: Nit:
```suggestion
        content: @Composable () -> Unit,
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
```diff
@@ -50,7 +49,7 @@ public fun JewelComposePanel(config: ComposePanel.() -> Unit = {}, content: @Com
         setContent {
             SwingBridgeTheme {
                 CompositionLocalProvider(
-                    LocalComponent provides this@createJewelComposePanel,
+                    org.jetbrains.jewel.foundation.component.LocalComponent provides this@createJewelComposePanel,
```
rock3r: Import alias?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/component/LocalComponent.kt
```diff
@@ -0,0 +1,15 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.foundation.component
```
rock3r: This should go one level up I reckon, the `component` package is for our composables :)


```suggestion
package org.jetbrains.jewel.foundation
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
```diff
@@ -151,6 +150,9 @@ internal class JewelComposePanelWrapper : JPanel(), UiDataProvider {
 @Suppress("CompositionLocalAllowlist")
 @ApiStatus.Experimental
 @ExperimentalJewelApi
-public val LocalComponent: ProvidableCompositionLocal<JComponent> = staticCompositionLocalOf {
-    error("CompositionLocal LocalComponent not provided")
-}
+@Deprecated(
+    "Use the LocalComponent from the foundation API",
+    replaceWith = ReplaceWith("org.jetbrains.jewel.foundation.component.LocalComponent"),
```
rock3r: If you add the import separately, it should result in a nicer replacement UX:

```suggestion
    replaceWith = ReplaceWith("LocalComponent", "org.jetbrains.jewel.foundation.LocalComponent"),
```

(assuming it's moved to `org.jetbrains.jewel.foundation`)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/JewelWindow.kt
```diff
@@ -0,0 +1,222 @@
...
+import androidx.compose.ui.window.singleWindowApplication as composeSingleWindowApplication
+import javax.swing.JComponent
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.foundation.component.LocalComponent
+
+/**
+ * Creates a window with the specified parameters and provide the
+ * [org.jetbrains.jewel.foundation.component.LocalComponent]. For more details in the parameters, check the official
+ * documentation of [androidx.compose.ui.window.Window].
```
rock3r:
```suggestion
 * Creates a window with the specified parameters and provide the
 * [org.jetbrains.jewel.foundation.component.LocalComponent]. For more details on the parameters, check the
 * [androidx.compose.ui.window.Window] documentation.
 * 
 * @see androidx.compose.ui.window.Window
```

But even, to be honest, we could copy the parameters KDoc directly here... It'll be better for users

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/JewelWindow.kt
rock3r, line 1: Nit: this could be named just `Window` to match the composable

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/JewelWindow.kt
```diff
@@ -0,0 +1,222 @@
...
+ */
+@Composable
+@ExperimentalJewelApi
+@ApiStatus.Experimental
```
rock3r: Given that we are copying the `Window` API and only inject a composition local, I think we can avoid starting with experimental

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/JewelWindow.kt
rock3r: Same for the other public API here

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/JewelWindow.kt
rock3r: Same for the other public API here

platform/jewel/int-ui/int-ui-standalone/intellij.platform.jewel.intUi.standalone.iml
```diff
@@ -14,9 +14,7 @@
           </stringArguments>
           <arrayArguments>
             <arrayArg name="pluginClasspaths">
-              <args>
-                <arg>$MAVEN_REPOSITORY$/org/jetbrains/kotlin/kotlin-compose-compiler-plugin/2.2.0/kotlin-compose-compiler-plugin-2.2.0.jar</arg>
-              </args>
+              <args>$MAVEN_REPOSITORY$/org/jetbrains/kotlin/kotlin-compose-compiler-plugin/2.2.0/kotlin-compose-compiler-plugin-2.2.0.jar</args>
```
rock3r: Please revert

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
```diff
@@ -35,6 +36,7 @@ import org.jetbrains.jewel.window.styling.TitleBarStyle
 @ExperimentalLayoutApi
 public fun main() {
     JewelLogger.getInstance("StandaloneSample").info("Starting Jewel Standalone sample")
+    JewelFlags.useCustomPopupRenderer = true
```
rock3r: Maybe we can put a checkbox in the welcome screen?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+import kotlin.math.floor
+import org.jetbrains.jewel.foundation.component.LocalComponent
+import org.jetbrains.jewel.foundation.modifier.thenIf
+import org.jetbrains.jewel.foundation.util.JewelLogger
+import org.jetbrains.jewel.ui.component.PopupRenderer
+
+/**
+ * A popup renderer implementation that uses JDialog to display popups in a Compose UI.
+ *
+ * This renderer creates a JDialog that hosts Compose content, allowing for integration between Swing and Compose UI
```
rock3r: Nit:

```suggestion
 * A popup renderer implementation that uses [JDialog] to display popups in a Compose UI.
 *
 * This renderer creates a [JDialog] that hosts Compose content, allowing for integration between Swing and Compose UI
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+ * hierarchies.
+ *
+ * Note that this renderer requires you to met at least one of the following conditions:
```
rock3r:
```suggestion
 * Note that this renderer requires you to meet at least one of the following conditions:
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+ * - Be running on JBR (JetBrains Runtime);
+ * - Have set the 'compose.interop.blending' system property;
```
rock3r:
```suggestion
 * - Having set the 'compose.interop.blending' system property;
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.component.popup
```
rock3r: How come this is not in standalone? I don't think we want people to use this in bridge

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+        if (!isJBREnvironment && !supportBlending) {
+            LaunchedEffect(Unit) {
+                JewelLogger.getInstance("PopupRenderer")
+                    .warn(
+                        "JDialogRenderer is enabled, but the requirements to use it are not met. " +
+                            "Falling back to the default renderer. To use it, make sure to match one of the " +
+                            "following rules:\n" +
+                            "- Run the project on a JBR-based environment.\n" +
+                            "- Enable the 'compose.interop.blending' system property.\n" +
+                            "- Set the 'Panel.background' Swing Property color to fully transparent."
```
rock3r: I'd generally just point people to the KDoc for the class so if we tweak it we don't risk forgetting to update here too

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+        val supportBlending = remember {
+            System.getProperty("compose.interop.blending", "false").toBoolean() ||
+                UIManager.getColor("Panel.background").alpha == 0
+        }
+
+        val currentComponent = LocalComponent.currentOrNull // Avoiding break the app if LocalComponent is not provided
+        val window = remember(currentComponent) { SwingUtilities.getWindowAncestor(currentComponent) }
+
+        if (!isJBREnvironment && !supportBlending) {
+            LaunchedEffect(Unit) {
```
rock3r:
```suggestion
        LaunchedEffect(isJBREnvironment, supportBlending) {
            if (!isJBREnvironment && !supportBlending) {
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+                            "- Enable the 'compose.interop.blending' system property.\n" +
+                            "- Set the 'Panel.background' Swing Property color to fully transparent."
+                    )
+            }
+        }
+
+        if (window == null) {
+            JewelLogger.getInstance("PopupRenderer")
+                .warn("LocalComponent was not provided. Falling back to the default renderer.")
+        }
```
rock3r:
```suggestion
        LaunchedEffect(window) {
            if (window == null) {
                JewelLogger.getInstance("PopupRenderer")
                    .warn("LocalComponent was not provided. Falling back to the default renderer.")
            }
        }
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+            ComposePopup(
+                popupPositionProvider = popupPositionProvider,
+                properties = properties,
+                onDismissRequest = onDismissRequest,
+                onPreviewKeyEvent = onPreviewKeyEvent,
+                onKeyEvent = onKeyEvent,
+                content = content,
+            )
+
+            return
```
rock3r: Else?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+    val popupDensity = LocalDensity.current
+    val component = LocalComponent.current
+
+    val currentContent = rememberUpdatedState(content)
+    val currentPopupPositionProvider = rememberUpdatedState(popupPositionProvider)
+
+    val compositionLocalContext = currentCompositionLocalContext
+
+    val parentBoundsInRoot = remember { mutableStateOf<IntRect?>(null) }
+    val popupRectangle = remember { mutableStateOf<Rectangle?>(null) }
```
rock3r: Nit: `by`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+                    true
+                } else {
+                    consumed
+                }
+            }
+        } else {
+            onKeyEvent
+        }
+
+    val dialog = remember {
```
rock3r: Feels like we may need to update the values read from properties on the dialog when they change — right now this may not cause the JDialog's focusability to change if the property changes. Same for the parent window.

Maybe they should be keys for the remember? It looks like you already have the machinery to move the composepanel down below, so this might be an oversight

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+            focusableWindowState = properties.focusable
+            isUndecorated = true
+            rootPane.isOpaque = false
+            background = Color(0, 0, 0, 0)
+            contentPane.background = Color(0, 0, 0, 0)
+            rootPane.putClientProperty("Window.shadow", true)
+        }
+    }
+
+    val composePanel = remember {
```
rock3r: This remember computation also accesses a lot of things that may change, do they need to be keys? I'm not sure how it works with nested compositions

(I realise a lot of this may also affect the JBPopup implementation... I did not pick them up at the time)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
```diff
@@ -0,0 +1,452 @@
...
+        onDispose { Toolkit.getDefaultToolkit().removeAWTEventListener(listener) }
+    }
+
+    DisposableEffect(dialog) {
+        dialog.contentPane = composePanel
+
+        dialog.isVisible = true
+        dialog.size = composePanel.preferredSize
+
+        onDispose {
```
rock3r: Do we need to detach the composePanel too?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -57,17 +62,29 @@ public fun Popup(
     properties: PopupProperties = PopupProperties(),
     onPreviewKeyEvent: ((KeyEvent) -> Boolean)? = null,
     onKeyEvent: ((KeyEvent) -> Boolean)? = null,
+    cornerSize: CornerSize = ZeroCornerSize,
```
rock3r: I know this is experimental, but can we create an overload to avoid breaking changes?

rock3r (commented):
As usual, excellent job — I have a few questions tho :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: I was using the early return to avoid adding another indentation level. Just personal preference. Else also works.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: That's something I was thinking, but since all the access I do is to "state" types, we shouldn't need to recreate the compose panel; instead, they should "recompose" as expected.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: To avoid recreating the whole popup instance, Ill just move the access from the 'properties' to a LaunchedEffect

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: So, I kept this one intentionally. As I access and modify from the "outer context" (the other compose panel), I was unsure if using delegation could cause any issues.

I was unable to reproduce any, but I haven't thoroughly tested it.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: I thought disposing of the parent (dialog) would call the children (compose panel). But it should not be a problem to call it

platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/DecoratedWindow.kt
faogustavo: Updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/component/LocalComponent.kt
faogustavo: Updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: ~~I can move it to stand alone~~

I moved it here as the decorated-window module does not depend on the int-ui-standalone. In that case, the only solutions I was able to think was:

1. Having two copies of the implementation
2. Moving the 'DecoratedWindow' to the `int-ui-decorated-window'
3. Reviewing the dependency tree to make this visible in the decorated-window module

The best solution was to keep it here. As this object is internal and we override it in the bridge module, it would not affect them.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: Added

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/Main.kt
faogustavo: Added

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
rock3r: Why does this need to be in `decorated-window`? It should be in standalone, and it can still be used from `intellij.platform.jewel.intUi.decoratedWindow`, which does depend on standalone.

The `decorated-window` module really should not be used directly as it does not have a theme layer. That's the same as trying to use the `ui` module directly instead of through standalone or bridge — it won't work unless you provide _all_ the theming and composition locals by yourself. That's by design.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
rock3r: I think it should be fine... there is no magic behind it:

```kotlin
inline operator fun <T> State<T>.getValue(thisObj: Any?, property: KProperty<*>): T = value

inline operator fun <T> MutableState<T>.setValue(thisObj: Any?, property: KProperty<*>, value: T) {
    this.value = value
}
```

We can always revert it if needed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
rock3r: Yeah, we should not need to recreate it, but I don't know if the inner composition does pick these up if they change. Any way you could test it? And add a comment to document this too, just in case

faogustavo:
Converted to draft until we merge #3175

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: It seems to work just fine. I'll push the update to use it

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/popup/JDialogRenderer.kt
faogustavo: Everything is working as expected. Updated the tooltip sample to add checkboxes to randomly change the position provider and content every second.


https://github.com/user-attachments/assets/fc8526ae-08db-4de4-a015-d013bdb16808

rock3r (approved):
Had forgotten to approve 💃

rock3r:
Ready to merge

PR #3155, opened 2025-07-29
[JEWEL-354] Use instanceUuid instead of isDark as remember key

DanielSouzaBertoldi:
# Context

It's not technically correct to use the `isDark` variable from `JewelTheme` to recalculate something once the user changes any LaF value. For instance, changing from Darcula to Dark, they are both dark themes but are different themes.

This PR fixes the few cases where we used `isDark` as a remember key.

### Implementation
- Changed variations of `remember(isDark)` to `remember(JewelTheme.instanceUuid)`
   - Markdown was the module with the most changes
- Also updated some README files to show how to properly check for theme changes.

### DIFF
--- a/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/theme/IntUiTheme.kt
+++ b/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/theme/IntUiTheme.kt
@@ -146,7 +146,7 @@ public fun JewelTheme.Companion.darkThemeDefinition(
 @Composable
 public fun ComponentStyling.default(): ComponentStyling = with {
     val isDark = JewelTheme.isDark
-    remember(isDark) { if (isDark) dark() else light() }
+    remember(JewelTheme.instanceUuid) { if (isDark) dark() else light() }
 }
 
 public fun ComponentStyling.dark(
@@ -298,7 +298,7 @@ public fun ComponentStyling.light(
 @Composable
 public fun IntUiTheme(isDark: Boolean = false, swingCompatMode: Boolean = false, content: @Composable () -> Unit) {
     val themeDefinition =
-        remember(isDark) {
+        remember(JewelTheme.instanceUuid) {
             if (isDark) {
                 JewelTheme.darkThemeDefinition()
             } else {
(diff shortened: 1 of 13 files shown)


platform/jewel/markdown/int-ui-standalone-styling/src/main/kotlin/org/jetbrains/jewel/intui/markdown/standalone/IntUiProvideMarkdownStyling.kt
```diff
@@ -33,7 +33,7 @@ import org.jetbrains.jewel.markdown.rendering.MarkdownStyling
 public fun ProvideMarkdownStyling(
     isDark: Boolean = JewelTheme.isDark,
     markdownStyling: MarkdownStyling =
-        remember(isDark, JewelTheme.instanceUuid) {
```
DanielSouzaBertoldi, line 36: Kinda redundant to check for both `isDark` and `JewelTheme.instanceUuid` in my opinion since changing the theme will also change the instance uuid

DanielSouzaBertoldi: Also, the more I think about it, the more I realize it's not worth it to check for the instance ID in standalone. Instance IDs are only making sense in my head for bridge, not standalone 🤔

rock3r: Technically, users may provide a standalone theme that has the same `isDark` value (and even name!) but change one of the colours or something else... That's why I think it's still worth doing this, even tho it's unlikely.

rock3r (commented):
Waiting for the replacement of `JewelTheme.name` as key, too

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/theme/IntUiTheme.kt
```diff
@@ -292,7 +292,7 @@ public fun ComponentStyling.light(
 @Composable
 public fun IntUiTheme(isDark: Boolean = false, swingCompatMode: Boolean = false, content: @Composable () -> Unit) {
     val themeDefinition =
-        remember(isDark) {
+        remember(JewelTheme.instanceUuid) {
```
DanielSouzaBertoldi, line 301: This broke tests in `:ui-test` since we use this composable and now it's actively checking for the `JewelTheme.instanceUuid` which is only provided by `JewelTheme`. I created an `IntUiTestTheme` that you can check out below that provides a random UUID and updated all tests to use this test theme instead 😄

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/IntUiTestTheme.kt
```diff
@@ -0,0 +1,19 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.component
```
rock3r: Nit:

```suggestion
package org.jetbrains.jewel
```

faogustavo (approved):

rock3r:
Ready to merge

PR #3158, opened 2025-07-30
[JEWEL-936] Fix scrolling layout when AlwaysVisible but not enough content to be scrollable

rock3r:
In scrollable containers, when the `ScrollbarVisibility` is `AlwaysVisible`, but the container is large enough that the content doesn't scroll, we were still leaving empty space for the scrollbars, which were invisible.

This fixes the issue by checking that the layout only reserves space for the scrollbars when they can actually scroll, in `AlwaysVisible`. This does not affect the `WhenScrolling` visibility setting, as that does not reserve any space for the scrollbars anyway (they're overlaid over the content).

This also improves the scrollbars showcase to have more representative examples, and more comprehensive than before.

It also adds tests, a lot of tests! I even found and fixed a bug thanks to the tests.

### Bug fixes
 * Fixed scrollable containers with `AlwaysVisible` reserving space for the scrollbar even when the content is smaller than the viewport
 * Fixed a small bug in height calculation for horizontal scrollable containers in edge conditions


rock3r:
FYI since the tests are failing but not on my machine, I added some fancy failure reporting to the workflow run:

image

You can even download the test reports: 
image

.github/workflows/jewel-checks.yml
```diff
@@ -32,6 +32,21 @@ jobs:
         # Run checks for all modules except the IDE plugin sample, as that is bound to have missing APIs issues
         run: ./gradlew check -x :samples:ide-plugin:check --continue --no-daemon
 
+      - name: Annotate JUnit test failures
+        uses: mikepenz/action-junit-report@v5
+        if: failure()
+        with:
+          report_paths: '**/build/test-results/test/TEST-*.xml'
+          annotate_only: true # A forked repo cannot write comments, so just do annotations
```
faogustavo, line 40: I'm not sure if this issue also happens for forks, but I had some issues with annotations in the past in another project: https://github.com/mikepenz/action-junit-report/issues/40

faogustavo (approved):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainer.kt
```diff
@@ -642,8 +661,11 @@ private fun computeContentConstraints(
 @Composable
 public fun scrollbarContentSafePadding(style: ScrollbarStyle = JewelTheme.scrollbarStyle): Dp =
     when {
-        hostOs != OS.MacOS -> style.scrollbarVisibility.trackThicknessExpanded
+        hostOs != OS.MacOS -> style.scrollbarVisibility.trackThicknessExpanded + 1.dp
```
DanielSouzaBertoldi, line 664: Honestly, I don't quite get this. If the scrollbars overlay the content, why do we need to add the extra 1.dp? 🤔

.github/workflows/jewel-checks.yml
rock3r: Seemed to work fine when I tested — the issue seems to be when multiple workflows are involved, which for us should be fine since there is only one. If issues come up we'll figure something else out.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainer.kt
rock3r: Ah yeah should have added a note — this is to leave 1 dp between the scrollbar and the actual content, so they do not "touch". It's just a minor visual tweak that imo makes it look better in most cases.

rock3r:
Ready to merge

PR #3160, opened 2025-07-31
[JEWEL-946] Fix SimpleListItem styling

rock3r:
In the bridge, we were reading the `SimpleListItem` colours from the wrong keys. This was not obvious in the light theme, but in the dark theme this resulted in weird background colours. Additionally, we were not correctly applying the "active" status to items in the popup, and the inactive selection background colour was incorrect in standalone, too.

This fixes both issues :) Note that in the light theme you didn't really see any issue, so don't be surprised if you can't tell below!

| |Standalone|
|---|---|
|After (light)|image|
|After (dark)|image|

| |Bridge|
|---|---|
|Before (light)|image|
|After (light)|image|
|Before (light)|image|
|After (dark)|image|

### Bug fixes
 * Fixed `SimpleListItem` colours in standalone and bridge
 * Fixed incorrect application of the "active" state in list combobox items


faogustavo (approved):

DanielSouzaBertoldi (approved):

rock3r:
Ready to merge

PR #3161, opened 2025-08-01
[JEWEL-920] Add Detekt Compose rules to the project

rock3r:
We now apply additional Compose rules to our Detekt checks. The rules are documented at https://mrmans0n.github.io/compose-rules/rules/

A couple of rules have been disabled as they do not apply to Jewel as a library. The rules JAR was added to the repository so that it can also be used by the IntelliJ Detekt plugin, and show issues in the IJ editor.

This applies trivial fixes from the rules, leaving challenging ones for a later time (suppressed and filed tech debt issues to fix them).

> ### 🚨 IMPORTANT REQUEST
> I'd appreciate it if when reviewing this PR y'all did some smoke testing, too. I did it, and it's fine, but just in case...

Notable changes from applying the rules:
 * Fixed a bug in `BasicLazyTree` where a derived state was not remembered
 * Fixed a bug in `ButtonImpl` where the `onStateChange`, accessed in a `LaunchedEffect`, was not used as a key to the effect
   * This effectively was not causing issues as it’s a private impl and we control the lambdas, but was a mistake nonetheless
   * We have recently fixed a similar bug in JEWEL-868
 * Fixed the same kind of bug in `ListComboBox`
 * Fixed the same kind of bug in `CircularProgressIndicator`
 * Fixed several missing remember keys in `PopupMenu`'s `MenuContent`
 * Fixed a bug in `ContextMenu` where the focus manager and input mode manager were not remembered
 * Deprecated several mistakenly public APIs in `Menu.kt`, and made them to delegate to a now-private implementation, so the public one can be safely removed in the future
 * Made sure there is no indication set by default, instead of the stock one provided by Compose
 * Adjusted the signature of many internal composables
 * Improved the SLC and tree samples in the showcase
 * Added a simpler variant to `Modifier.outline()`
 * Removed duplication in `ComboBox` implementations by making the string-based variant
   delegate to slot-based one

This also adds an artefact configuration, so that we get a JAR for our own Detekt rules plugin when the IDE is built from JPS.

### ⚠️ Important Changes
 * The default `Indication` has been set to a no-op implementation in both standalone and bridge, instead of the previous default implementation we inherited from Compose. We handle visual states separately from the `Indication` API, and as such this would only cause visual issues when using certain modifiers (e.g., `selectable`)
 * The experimental slot-based `ComboBox` overload has changed in a **breaking** way by reordering its parameters

### New features
 * Added a simpler variant to `Modifier.outline()`

### Bug fixes
 * Fixed a bug in `BasicLazyTree` where the item background state was not properly remembered
 * Fixed a bug in `ListComboBox` where changing the `itemKeys` parameter value would not be picked up by the component until it exited and re-entered the composition
 * Fixed a bug in `CircularProgressIndicator` where changing the `frameRetriever` parameter value would not be picked up by the component until it exited and re-entered the composition
 * Fixed a bug where the `PopupMenu` was over-remembering some internal state

### Deprecated API
 * Deprecated several APIs in `Menu.kt` that were left public by mistake so they can be made private as they should, in the future


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/RadioButton.kt
```diff
@@ -145,16 +145,15 @@ public fun RadioButtonRow(
...
         enabled = enabled,
         outline = outline,
         interactionSource = interactionSource,
         style = style,
         textStyle = textStyle,
         verticalAlignment = verticalAlignment,
-    ) {
-        Text(text)
-    }
+        { Text(text) },
```
faogustavo: Nit: naming this param?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainer.kt
```diff
@@ -101,18 +101,22 @@ public fun VerticallyScrollableContainer(
             )
         },
         horizontalScrollbar = null,
-        modifier = modifier.withKeepVisible(style.scrollbarVisibility.lingerDuration, scope) { keepVisible = it },
         scrollbarStyle = style,
-    ) {
-        Box(Modifier.layoutId(ID_CONTENT).verticalScroll(scrollState, reverseScrolling = reverseLayout)) { content() }
-    }
+        modifier = modifier.withKeepVisible(style.scrollbarVisibility.lingerDuration, scope) { keepVisible = it },
+        {
```
faogustavo: Nit: Should we move this lambda out or name it?

platform/jewel/ui/api-dump-experimental.txt
```diff
@@ -1,5 +1,5 @@
 f:org.jetbrains.jewel.ui.component.ComboBoxKt
-- *sf:ComboBox-05tvjtU(androidx.compose.ui.Modifier,androidx.compose.ui.Modifier,Z,org.jetbrains.jewel.ui.Outline,F,androidx.compose.foundation.interaction.MutableInteractionSource,org.jetbrains.jewel.ui.component.styling.ComboBoxStyle,kotlin.jvm.functions.Function0,kotlin.jvm.functions.Function0,org.jetbrains.jewel.ui.component.PopupManager,kotlin.jvm.functions.Function2,kotlin.jvm.functions.Function2,androidx.compose.runtime.Composer,I,I,I):V
+- *sf:ComboBox-8u0NR3k(kotlin.jvm.functions.Function2,kotlin.jvm.functions.Function2,androidx.compose.ui.Modifier,androidx.compose.ui.Modifier,Z,org.jetbrains.jewel.ui.Outline,F,androidx.compose.foundation.interaction.MutableInteractionSource,org.jetbrains.jewel.ui.component.styling.ComboBoxStyle,kotlin.jvm.functions.Function0,kotlin.jvm.functions.Function0,org.jetbrains.jewel.ui.component.PopupManager,androidx.compose.runtime.Composer,I,I,I):V
```
DanielSouzaBertoldi, line 2: Even though it's an experimental api, shouldn't we mention this breaking change in the release notes?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
```diff
@@ -53,7 +52,7 @@ import org.jetbrains.jewel.ui.outline
 import org.jetbrains.jewel.ui.painter.hints.Stateful
 import org.jetbrains.jewel.ui.theme.dropdownStyle
 
-@ScheduledForRemoval(inVersion = "2025.2")
+@Suppress("ComposableParamOrder")
```
DanielSouzaBertoldi, line 55: I see that there are some suppress annotations for ComposableParamOrder that are commented with "// To fix in JEWEL-929" while others are not.

Should we just fix those that are actively followed by the comment or should we fix all of these Composables being supressed?

platform/jewel/ui/api-dump.txt
```diff
@@ -993,7 +981,7 @@ org.jetbrains.jewel.ui.component.banner.BannerLinkActionScope
 f:org.jetbrains.jewel.ui.component.banner.ComposableSingletons$BannerLinkActionScopeKt
 - sf:INSTANCE:org.jetbrains.jewel.ui.component.banner.ComposableSingletons$BannerLinkActionScopeKt
 - <init>():V
-- f:getLambda$-1363944258$intellij_platform_jewel_ui():kotlin.jvm.functions.Function3
+- f:getLambda$-483619984$intellij_platform_jewel_ui():kotlin.jvm.functions.Function3
```
DanielSouzaBertoldi: This is a breaking change, as far as I know (please correct me if I'm wrong). Even though the `RowScope.BannerActionsRow` (now just `BannerActionsRow`) is an internal function, it's still part of the ABI (Application Binary Interface).

JVM doesn't understand `internal` declarations so the [Kotlin compiler compiles it as public in the JVM bytecode](https://kotlinlang.org/docs/java-to-kotlin-interop.html#visibility).

In other words, this breaks binary compatibility. If a module B depends on on our UI module and is using the current `RowScope.BannerActionsRow`, it'll try to use the old signature. If module B doesn't recompile our module it'll crash at runtime since it won't be able to find the expected signature.

However, I'm not sure if the problem is the removal of the `RowScope.` or the changes inside the function or both 🤔 

In this case I think the best approach is to deprecate the current `RowScope.BannerActionsRow` and any a new one with the new changes

rock3r: Yeah, this is _technically_ breaking, but as far as I understand, this only impacts non-Kotlin calling code. The Kotlin compiler would recognise `internal` and prevent external usages from Kotlin code.

Now, since this change is on composable functions, which can't be used from Java or any other non-Kotlin JVM language, I reckon this change is actually not "really" a change.

I cannot think of a single scenario where this change would _really_ be binary or source incompatible in practice. Happy to be proven wrong tho, if you have any in mind...

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
rock3r: In this case I haven't added one because this is a deprecated API we'll drop sooner or later, so it's not worth the effort of creating a copy that is _also_ deprecated :)

platform/jewel/ui/api-dump.txt
DanielSouzaBertoldi: Hmmmm I'm by no means an expert in API breaking changes (yet! 😎), I'm questioning based on what I looked up last week while figuring out what could be potential breaking changes.

Like, suppose we created an IJ IDEA plugin and have Jewel as a dependency. Our plugin has a Composable with like something this:

```
Row {
    // This call is valid with Jewel v0.29 because BannerActionsRow
    // is an extension on RowScope.
    BannerActionsRow(spaceBetweenItems = 8.dp) {
        // cool code
    }
}
```

We compile our plugin against v0.29, and the resulting `.jar` file from our plugin (`CoolPlugin.jar`) now contains bytecode with a hardcoded reference to the method with `RowScope.BannerActionsRow`. We publish the plugin and everything's fine.

Then, suppose we ship this change in v0.30. We update our ✨ cool plugin ✨ build gradle to use the new v0.30 release of Jewel but we do not recompile our cool plugin source code. We push this update to JetBrains plugins marketplace.

When the end-user installs or updates our cool plugin, it'll have: `CoolPlugin.jar` still compiled agains v0.29, and the updated dependency v0.30. When the time comes to render our `Row`, it'll try to call `RowScope.BannerActionsRow` but since this signature doesn't exist anymore, the JVM will throw an `NoSuchMethorError` and crash.

Besides, I think it's safer to just trust the API dump and be reaaally careful with any minor change. Unless we are removing deprecated code, maybe we should just play it safe and avoid any changes to existing methods/classes/whatnot

Do you think my line of thought makes sense? I'm still not really familiar with plugin bundling, shipping and etc so maybe I'm overthinking this (again lol)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
DanielSouzaBertoldi: Hadn't checked if the composable was deprecated, my bad 😭

rock3r: No sweat :)

platform/jewel/ui/api-dump.txt
rock3r: I'm still convinced this is not a breaking change because of what I wrote earlier — you can only invoke this from Kotlin, and Kotlin respects `internal` — calling this from any other language will result in compiler errors anyway. However, while it's _technically_ not an issue, I agree with the principle of caution you're expressing, so I'm going with the deprecation and copy approach.

FWIW I am seriously thinking we should look into using [Metalava](https://android.googlesource.com/platform/tools/metalava/) as it feels a lot more competent than IJP's dumps... But I also don't want to end up _again_ into having two separate dumps :D

rock3r: Also for the record, this is in the `ComposableSingletons` generated class, which is invisible to users :)

DanielSouzaBertoldi: Yeah Metalava seems like a good alternative too, we could test it down the line :) Also, totally overlooked the fact this is from a generated class, but thanks for the discussion anyways 🙏🏻

rock3r: So, I checked with the Compose team:

>`ComposableSingletons` is supposed to be internal, so any changes to it do not impact binary compat

So we should make sure it's omitted from the dumps :)

rock3r:
Ready to merge

PR #3162, opened 2025-08-04
[JEWEL-949] Fix interaction of links and selection in Markdown

rock3r:
This fixes the issue that is present when some selectable `Markdown` (i.e., when the content is wrapped in a `SelectionContainer`) also has one or more links. In that case, the `clickable` added to the `Text` composables used to render paragraphs steals the drag events and makes it impossible to start selecting text from said `Text` components.

 Before | After
 --- | ---
 Before | After

This in turn has a wide-ranging impact in that it makes the `onTextClick` parameter unused throughout the Markdown APIs. Turns out that the original use case (making clickable Markdown-based components such as checkboxes) would be better served by a custom block renderer. So in this change I removed the `clickable` modifier and `onTextClick` from all Markdown rendering APIs.

Given this is a large change, even if this is an experimental API, I decided to still keep the old APIs around for backwards compatibility. I also took the opportunity to rename the `render` APIs to names that are more descriptive and respect Compose naming conventions.

Then, I fixed a subtle bug where disabled fenced code blocks would be more "transparent" than indented ones (we applied `alpha` twice!).

 Before | After
 --- | ---
 Before | image

Lastly, I have added a bunch of KDoc around, and cleaned up a lot of duplication in the `Markdown`/`LazyMarkdown` code, also making sure a change in `selectable` doesn't lose state.

### ⚠️ Important Changes
 * All the experimental `*.render` APIs have been renamed and have lost the `onTextClick` parameter (non-breaking change). They now all have a default implementation that delegates to the new counterparts, ignoring `onTextClick`
 * The experimental `Markdown` and `LazyMarkdown` composables have lost the `onTextClick` parameter (non-breaking change)
 * The experimental API `ImageRendererExtension.renderImagesContent` was renamed to `renderImageContent` (singular "image") in a **breaking** manner
 * The experimental `GitHubTableBlockRenderer` has been made private

### Bug fixes
 * Fixed a bug that made it impossible to select text in a Markdown paragraph that contains one or more links
 * Fixed a bug where disabled fenced code blocks in Markdown would look "lighter" than indented code blocks

### Deprecated API
 * All `MarkdownBlockRenderer.render` APIs have been deprecated in favour of the new APIs with better naming, and no `onTextClick` parameter
 * `Markdown` and `LazyMarkdown` overloads with the `onTextClick` parameter have been deprecated

### DIFF
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/ImageRendererExtension.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/ImageRendererExtension.kt
@@ -6,9 +6,22 @@ import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.markdown.InlineMarkdown
 
-/** An extension for the Jewel images rendering engine. */
+/**
+ * Extension for rendering images in Jewel Markdown.
+ *
+ * Implement this interface to provide a custom renderer for images. This is useful for handling image loading from
+ * different sources (e.g., network, local files) or for applying custom visual treatments to images.
+ *
+ * The [renderImageContent] function will be called for each image found in the Markdown content.
+ */
 @ApiStatus.Experimental
 @ExperimentalJewelApi
 public interface ImageRendererExtension {
-    @Composable public fun renderImagesContent(image: InlineMarkdown.Image): InlineTextContent
+    /**
+     * Renders an image from a Markdown document.
+     *
+     * @param image The image data, containing information like the image URL and alt text.
+     * @return An [InlineTextContent] that will be embedded in the text flow, which will be used to display the image.
+     */
+    @Composable public fun renderImageContent(image: InlineMarkdown.Image): InlineTextContent
 }
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/MarkdownBlockRendererExtension.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/MarkdownBlockRendererExtension.kt
@@ -9,12 +9,27 @@ import org.jetbrains.jewel.markdown.MarkdownBlock.CustomBlock
 import org.jetbrains.jewel.markdown.rendering.InlineMarkdownRenderer
 import org.jetbrains.jewel.markdown.rendering.MarkdownBlockRenderer
 
-/** An extension for [MarkdownBlockRenderer] that can render one or more [MarkdownBlock.CustomBlock]s. */
+/**
+ * An extension for [MarkdownBlockRenderer] that can render one or more [MarkdownBlock.CustomBlock]s.
+ *
+ * This is the primary way to handle custom Markdown elements that are not part of the standard Markdown syntax.
+ * Implement this interface to define how a specific type of [CustomBlock] should be rendered.
+ *
+ * If you want to customize how a standard block is rendered, you can create a new [MarkdownBlockRenderer] with the
+ * required custom behavior. Extending [org.jetbrains.jewel.markdown.rendering.DefaultMarkdownBlockRenderer] is the
+ * simplest choice in that case.
+ *
+ * Implementations are responsible for both checking if they can render a given block, through the [canRender] method,
+ * and for providing the Composable content for it, through the [RenderCustomBlock] method.
+ */
 @ApiStatus.Experimental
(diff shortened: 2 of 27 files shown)


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/ImageRendererExtension.kt
```diff
@@ -6,9 +6,22 @@ import org.jetbrains.annotations.ApiStatus
...
+ *
+ * Implement this interface to provide a custom renderer for images. This is useful for handling image loading from
+ * different sources (e.g., network, local files) or for applying custom visual treatments to images.
+ *
+ * The [RenderImage] function will be called for each image found in the Markdown content.
+ */
 @ApiStatus.Experimental
 @ExperimentalJewelApi
 public interface ImageRendererExtension {
-    @Composable public fun renderImagesContent(image: InlineMarkdown.Image): InlineTextContent
```
rock3r, line 13: Breaking change, noted in release notes.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -166,15 +183,26 @@ public open class DefaultMarkdownBlockRenderer(
...
-            5 -> render(block, styling.h5, enabled, onUrlClick, onTextClick, modifier)
-            6 -> render(block, styling.h6, enabled, onUrlClick, onTextClick, modifier)
-            else -> error("Heading level ${block.level} not supported:\n$block")
+            1 -> RenderHeading(block, styling.h1, enabled, onUrlClick, modifier)
+            2 -> RenderHeading(block, styling.h2, enabled, onUrlClick, modifier)
+            3 -> RenderHeading(block, styling.h3, enabled, onUrlClick, modifier)
+            4 -> RenderHeading(block, styling.h4, enabled, onUrlClick, modifier)
+            5 -> RenderHeading(block, styling.h5, enabled, onUrlClick, modifier)
+            6 -> RenderHeading(block, styling.h6, enabled, onUrlClick, modifier)
+            else -> JewelLogger.getInstance(javaClass).error("Heading level ${block.level} not supported:\n$block")
```
rock3r, line 206: Decided that logging was better than crashing

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/MarkdownBlockRendererExtension.kt
```diff
@@ -42,5 +61,30 @@ public interface MarkdownBlockRendererExtension {
...
         onUrlClick: (String) -> Unit,
         onTextClick: () -> Unit,
+    ) {
+        RenderCustomBlock(block, blockRenderer, inlineRenderer, enabled, modifier, onUrlClick)
+    }
+
+    /**
+     * Renders a [MarkdownBlock.CustomBlock] as a native Composable.
+     *
+     * Note that if [canRender] returns `false` for a given [block], the implementation of this method might throw.
```
DanielSouzaBertoldi: Really small tiny nit:
```suggestion
     * Note that if [canRender] returns `false` for a given [block], the implementation of this method may throw an exception.
```

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollSyncMarkdownBlockRenderer.kt
```diff
@@ -47,72 +54,64 @@ public open class ScrollSyncMarkdownBlockRenderer(
...
     @Composable
-    override fun render(
-        block: MarkdownBlock,
-        enabled: Boolean,
-        onUrlClick: (String) -> Unit,
-        onTextClick: () -> Unit,
-        modifier: Modifier,
-    ) {
-        if (block is ScrollingSynchronizer.LocatableMarkdownBlock) {
+    override fun RenderBlock(block: MarkdownBlock, enabled: Boolean, onUrlClick: (String) -> Unit, modifier: Modifier) {
```
DanielSouzaBertoldi, line 58: All the `Render*` composable functions here don't have KDocs. Is this intentional?

rock3r: They should inherit the interface's.

rock3r:
Ready to merge

PR #3163, opened 2025-08-04
[JEWEL-842] Fix Ligatures Not Mirroring IDE Settings

DanielSouzaBertoldi:
# Bug

On bridge, if the user changes the IDE current font's ligature settings, the markdown styling kinda just ignores it. This PR fixes this.

# Evidences

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/e6b29b0a-d4d3-4aef-b777-f22fc5169469" width="400" /> | <video src="https://github.com/user-attachments/assets/8b515a2e-672c-4572-9487-3bfb040c3904" width="400" /> |

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
@@ -70,7 +70,7 @@ public fun retrieveEditorTextStyle(): TextStyle {
             fontFamily = font.asComposeFontFamily(),
             fontSize = editorColorScheme.editorFontSize.sp,
             lineHeight = (computedLineHeight * EDITOR_LINE_HEIGHT_FACTOR).coerceAtLeast(1f).sp,
-            fontFeatureSettings = if (!editorColorScheme.isUseLigatures) "liga 0" else "liga 1",
+            fontFeatureSettings = if (!editorColorScheme.isUseLigatures) "liga 0, calt 0" else "liga 1, calt 1",
         )
 }
 


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
```diff
@@ -70,7 +70,7 @@ public fun retrieveEditorTextStyle(): TextStyle {
             fontFamily = font.asComposeFontFamily(),
             fontSize = editorColorScheme.editorFontSize.sp,
             lineHeight = (computedLineHeight * EDITOR_LINE_HEIGHT_FACTOR).coerceAtLeast(1f).sp,
-            fontFeatureSettings = if (!editorColorScheme.isUseLigatures) "liga 0" else "liga 1",
+            fontFeatureSettings = if (!editorColorScheme.isUseLigatures) "liga 0, calt 0" else "liga 1, calt 1",
```
DanielSouzaBertoldi, line 73: So, according to [W3C docs](https://www.w3.org/TR/css-fonts-3/#propdef-font-variant-ligatures), there's also `dlig` and `hlig`:

image

The problem is finding fonts with such OpenType features. `Tratattello` is one such case where no matter the combinations I set for `fontFeatureSettings`, an capital Q still shows up as if the ligatures were on 🤔 

Screenshot 2025-08-04 at 15 32 28

Personally, we could toggle all ligature types, but such cases as described above could still happen and further investigation would be necessary, which I don't think it's worth the effort right now. What do you think?

rock3r: Can you try and understand how the IDE/JBR defines "ligature" and adopt the same behaviour?

rock3r: I'm going to answer myself — JBR uses `liga` and `calt`, so we're a-ok.

https://github.com/JetBrains/JetBrainsRuntime/blob/3e5f4bb719f01d8a64935d9974541ab8643f269b/src/java.desktop/share/classes/java/awt/Font.java#L285-L286

DanielSouzaBertoldi: Beat me to it. Sorry for the late response :( And thanks for finding out how JBR deals with ligatures 😄

rock3r: Ahah I was just curious. Was easy enough to search for `calt` in the JBR repo anyway!

faogustavo (approved):

rock3r:
Ready to merge — should not need a rebase.

PR #3165, opened 2025-08-06
Keep async keyword when pulling up async method

chbndrhnns:
Fixes https://youtrack.jetbrains.com/issue/PY-56793/Pull-members-up-converts-async-method-to-sync-method
Fixes https://youtrack.jetbrains.com/issue/PY-53643/Missing-async-statement-when-pulling-method-up-as-abstract


avokin2:
Merged

PR #3166, opened 2025-08-06
[JEWEL-948] Add Scrollable containers for staggered grids

rock3r:
This adds a new overload to ScrollableContainer APIs that takes in a `ScrollableState`. This new API allows it to be used with any scroll-based container, both lazy and non-lazy.

This then deprecates the older lazy-based overloads as they can be expressed trivially in terms of the new ScrollableState-based ones.

Lastly, this adds clarification to the various KDocs.

### New features
 * Added a new overload for scrollable containers that takes a more general `ScrollableState` parameter, that can be used with all lazy containers, as well as non-lazy containers that want to own their scroll modifier.

### Deprecated API
 * Deprecated `LazyListState`- and `LazyGridState`-based APIs for scrollable containers as they can be trivially migrated to the new `ScrollableState`-based APIs.


rock3r:
Ready to merge

PR #3169, opened 2025-08-06
[JEWEL-944] Annotate breaking changes in API dumps

rock3r:
This adds a script that's run by the GitHub CI to validate that the PR is not introducing binary changes, and if it finds any, it adds annotations that show on the PR code. For stable API breaking changes, this will also fail the check.

This script should be able to be used locally as well, as when there is no PR_NUMBER env set, it just checks against HEAD~1 (assumes our changes are always at most one commit — we have other checks for that).

With this script, we get annotations in the PR for breakages.

* Dumps with issues get highlighted in the file tree:
  
  <br/>
* Stable breaking API changes get an error annotation:
  
  <br/>
* Experimental breaking API changes get an error annotation:
  
  <br/>
* It also appends a summary to the Actions run:
  image

 

### Notes
 * This only leaves annotations for experimental API changes, it does not fail the build. Its goal is to call out such changes so that the reviewers can evaluate them properly
 * The checks are naive; if an API is moved in the same file, it will still be considered broken. This should not be a big issue, since it is only posting annotations
 * We only get one annotation per removal "block" — i.e., if multiple consecutive rows are deleted, we only get one annotation. The log will contain all of them, however
 * The check explicitly skips the samples folder as those are not published modules and do not have any guarantees whatsoever. They should not have dumps to begin with...

### DIFF
--- a/platform/jewel/scripts/utils.main.kts
+++ b/platform/jewel/scripts/utils.main.kts
@@ -1,0 +1,91 @@
+@file:DependsOn("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.1")
+@file:DependsOn("com.github.pgreze:kotlin-process:1.5.1")
+@file:Suppress("RAW_RUN_BLOCKING")
+
+import com.github.pgreze.process.Redirect
+import com.github.pgreze.process.process
+import kotlin.system.exitProcess
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+import org.intellij.lang.annotations.Language
+import java.io.File
+
+fun checkGhTool() = runBlocking { runCommand(command = "which gh", workingDir = null, exitOnError = false).isSuccess }
+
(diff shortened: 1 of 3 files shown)


platform/jewel/scripts/utils.main.kts
```diff
@@ -0,0 +1,91 @@
...
+import com.github.pgreze.process.process
+import kotlin.system.exitProcess
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+import org.intellij.lang.annotations.Language
+import java.io.File
+
+fun checkGhTool() = runBlocking { runCommand(command = "which gh", workingDir = null, exitOnError = false).isSuccess }
```
faogustavo, line 15: nit: Should we make this a val? I don't think this is something that can change on runtime

rock3r: I don't love having vals that calls `runBlocking` or suspend functions, that's mainly why it's a function

DanielSouzaBertoldi (approved):

rock3r:
Ready to merge

PR #3170, opened 2025-08-06
Set skiko.library.path jvm arg at launch

jrlogsdon:
Currently, this gets set in JewelComposePanelWrapper at runtime. This has the potential for causing issues if the UI uses `ComposePanel` without wrapping it in a  `JewelComposePanelWrapper` before the JVM property gets set - causing failures when initiliaizing skiko. While this is not fully supported it shouldn't break the IDE.

This became an issue in 2025.2, when the skiko OS level binaries began being shipped separately for each OS platform outside of the module inside the lib/skiko-awt-runtime-all directory instead of being bundled in the intellij.libraries.skiko.jar in commit 74527c8.

Change-Id: I44e7ed265181254070e7bf645de4d68b7b42654c

### DIFF
--- a/platform/build-scripts/src/org/jetbrains/intellij/build/impl/BuildContextImpl.kt
+++ b/platform/build-scripts/src/org/jetbrains/intellij/build/impl/BuildContextImpl.kt
@@ -373,6 +373,7 @@ class BuildContextImpl internal constructor(
     jvmArgs += "-Djna.noclasspath=true"
     jvmArgs += "-Dpty4j.preferred.native.folder=${macroName}/lib/pty4j".let { if (isScript) '"' + it + '"' else it }
     jvmArgs += "-Dio.netty.allocator.type=pooled"
+    jvmArgs += "-Dskiko.library.path=${macroName}/lib/skiko-awt-runtime-all"
 
     if (useModularLoader || generateRuntimeModuleRepository) {
       jvmArgs += "-Dintellij.platform.runtime.repository.path=${macroName}/${MODULE_DESCRIPTORS_COMPACT_PATH}".let { if (isScript) '"' + it + '"' else it }


jakub-senohrabek:
I have already done this in the past and it 1) breaks support for multiple jewel/skiko versions 2) breaks other stuff that uses different skiko versions. We can extract the logic that sets the natives path, but we cannot do it globally for whole ide.

jrlogsdon:
Wouldn't these issues still occur after the first `JewelComposePanel` is used and this system property is set (it's just a matter of the sequencing) since it isn't resetting this system property?

rock3r:
Maybe this should not be a system property, but initialised per-classloader? This way different plugins can still use their own Skiko build, as they have separate classloaders. But I understand this would be a CMP change.

> if the UI uses ComposePanel without wrapping it in a JewelComposePanelWrapper

@jrlogsdon that's a classic case of "holding it wrong" tho, you're never supposed to use a raw ComposePanel in Jewel/IJ

jrlogsdon:
Could I get some more context here -  was the main reason to go away from bundling the binaries inside jars the size consideration or was there a deeper issue? (I noticed the original message mentioned `pre-signing` but I hadn't noticed any issue related to this in `2025.1`).

From what I could tell, there isn't existing build infra to bundle native libs inside jars per OS but perhaps we could add this functionality and it could be useful for other cases e.g. sql lite if the main concern was size.

> per-classloader
I have to think about this more for feasibility, but it sounds safer than system properties.

jrlogsdon:
re: Per-classloader
This is essentially going back to the old implementation (doing a lookup inside the jar). I think this is the safest bet, but we should only bundle the native library in each jar by OS to keep size down (as skiko binaries are quite large).

Once the build infra supports this other jars could use this functionality to reduce their size e.g. debugger-memory-agent.jar and sqlite.jar if wanted as well.

jrlogsdon:
For now, we migrated existing usages of `ComposePanel` and will match IntelliJ behavior but I still have concerns as outlined here.

I've filed https://youtrack.jetbrains.com/issue/IDEA-377632/Support-bundling-native-libraries-inside-module-jars-per-OS as a feature request for a long term fix.

gharrma:
We still run into the original issue, sometimes in places where it doesn't make sense to wrap with `JewelComposePanel`. For example, an early call to `toComposeImageBitmap()` will trigger the same problem:
```
java.lang.NullPointerException: 
  java.base/java.util.Objects.requireNonNull
  java.base/java.nio.file.Files.copy
  org.jetbrains.skiko.Library.unpackIfNeeded
  org.jetbrains.skiko.Library.findAndLoad
  org.jetbrains.skiko.Library.load
  org.jetbrains.skia.impl.Library$Companion.staticLoad
  org.jetbrains.skia.Bitmap.<clinit>
  androidx.compose.ui.graphics.DesktopImageConverters_desktopKt.toComposeImageBitmap
  ...
```

To fix this in Android Studio, we are resorting to setting this system property early on during app startup.

---

> I have already done this in the past and it 1) breaks support for multiple jewel/skiko versions 2) breaks other stuff that uses different skiko versions. We can extract the logic that sets the natives path, but we cannot do it globally for whole ide.

@jakub-senohrabek Given that `JewelComposePanel` already sets `skiko.library.path` globally for the whole ide currently, is it fair to say that this pull request does not make the situation worse?

jakub-senohrabek:
Seems like the only other usage that doesn't use libraries/skiko is android-adt-ui, we can merge this, but I recommend testing if the android plugin still works.

PR #3171, opened 2025-08-07
[JEWEL-741] Use TextMate as Fallback for Languages Without Parsers

DanielSouzaBertoldi:
## DRAFT PR!

I opted to open this PR as draft so we can discuss the best approach for this problem before I continue implementing it, since there are quite a few options.

First of all, you can ignore most of the changed code for now (tons of breaking changes and dubious code currently 😎) Focus on `LexerBasedCodeHighlighter`. I'll start explaining the problem, then the path I'm more inclined to take, then I'll talk briefly about other options. Bear with me here, please.

## Problem

Our syntax highlighting code for Markdown code blocks currently only highlights code that the IDE has a plugin that support it. So, if you open PyCharm for instance, it'll render python code blocks but not Kotlin. The idea here is to fallback to using [TextMate](https://www.jetbrains.com/help/idea/textmate.html) to highlight code in such cases.

## Current solution

We have a `MimeType` class that maps to a given programming language declared in the `Known` companion object:

```kt
public val KOTLIN: MimeType = MimeType("text/kotlin")
public val JAVA: MimeType = MimeType("text/java")
public val XML: MimeType = MimeType("text/xml")
public val SQL: MimeType = MimeType("text/x-sql")
```

### DIFF
--- a/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
+++ b/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
@@ -4,17 +4,22 @@ import org.jetbrains.jewel.foundation.code.MimeType.Known.AGSL
 import org.jetbrains.jewel.foundation.code.MimeType.Known.AIDL
 import org.jetbrains.jewel.foundation.code.MimeType.Known.C
 import org.jetbrains.jewel.foundation.code.MimeType.Known.CPP
+import org.jetbrains.jewel.foundation.code.MimeType.Known.CSS
 import org.jetbrains.jewel.foundation.code.MimeType.Known.DART
 import org.jetbrains.jewel.foundation.code.MimeType.Known.DIFF
 import org.jetbrains.jewel.foundation.code.MimeType.Known.GO
 import org.jetbrains.jewel.foundation.code.MimeType.Known.GRADLE
 import org.jetbrains.jewel.foundation.code.MimeType.Known.GRADLE_KTS
 import org.jetbrains.jewel.foundation.code.MimeType.Known.GROOVY
+import org.jetbrains.jewel.foundation.code.MimeType.Known.HTML
 import org.jetbrains.jewel.foundation.code.MimeType.Known.JAVA
 import org.jetbrains.jewel.foundation.code.MimeType.Known.JAVASCRIPT
 import org.jetbrains.jewel.foundation.code.MimeType.Known.JSON
+import org.jetbrains.jewel.foundation.code.MimeType.Known.JSON5
+import org.jetbrains.jewel.foundation.code.MimeType.Known.JSON_LINES
 import org.jetbrains.jewel.foundation.code.MimeType.Known.KOTLIN
 import org.jetbrains.jewel.foundation.code.MimeType.Known.MANIFEST
+import org.jetbrains.jewel.foundation.code.MimeType.Known.OBJECTIVE_C
 import org.jetbrains.jewel.foundation.code.MimeType.Known.PATCH
 import org.jetbrains.jewel.foundation.code.MimeType.Known.PROGUARD
 import org.jetbrains.jewel.foundation.code.MimeType.Known.PROPERTIES
@@ -22,6 +27,7 @@ import org.jetbrains.jewel.foundation.code.MimeType.Known.PROTO
 import org.jetbrains.jewel.foundation.code.MimeType.Known.PYTHON
 import org.jetbrains.jewel.foundation.code.MimeType.Known.REGEX
 import org.jetbrains.jewel.foundation.code.MimeType.Known.RESOURCE
+import org.jetbrains.jewel.foundation.code.MimeType.Known.RUBY
 import org.jetbrains.jewel.foundation.code.MimeType.Known.RUST
 import org.jetbrains.jewel.foundation.code.MimeType.Known.SHELL
 import org.jetbrains.jewel.foundation.code.MimeType.Known.SQL
@@ -31,8 +37,46 @@ import org.jetbrains.jewel.foundation.code.MimeType.Known.TOML
 import org.jetbrains.jewel.foundation.code.MimeType.Known.TYPESCRIPT
 import org.jetbrains.jewel.foundation.code.MimeType.Known.UNKNOWN
 import org.jetbrains.jewel.foundation.code.MimeType.Known.VERSION_CATALOG
+import org.jetbrains.jewel.foundation.code.MimeType.Known.XHTML
 import org.jetbrains.jewel.foundation.code.MimeType.Known.XML
 import org.jetbrains.jewel.foundation.code.MimeType.Known.YAML
+
+private val ALREADY_NORMALIZED_BUILTIN_TYPES =
+    listOf(
+        AGSL,
+        AIDL,
+        C,
+        CPP,
+        DART,
+        DIFF,
+        GO,
+        GRADLE,
+        GRADLE_KTS,
+        GROOVY,
+        JAVA,
+        JAVASCRIPT,
+        JSON,
+        KOTLIN,
+        MANIFEST,
+        PATCH,
+        PROGUARD,
+        PROPERTIES,
+        PROTO,
+        PYTHON,
+        REGEX,
+        RESOURCE,
+        RUST,
+        SHELL,
+        SQL,
+        SVG,
+        TEXT,
+        TOML,
+        TYPESCRIPT,
+        UNKNOWN,
+        VERSION_CATALOG,
+        XML,
+        YAML,
+    )
 
 /**
  * Represents the language and dialect of a source snippet, as an RFC 2046 mime type.
@@ -56,7 +100,7 @@ import org.jetbrains.jewel.foundation.code.MimeType.Known.YAML
 @JvmInline
 public value class MimeType(private val mimeType: String) {
     public fun displayName(): String =
-        when (normalizeString()) {
+        when (getBaseMimeType()) {
             KOTLIN.mimeType -> if (isGradle()) "Gradle DSL" else "Kotlin"
             JAVA.mimeType -> "Java"
             XML.mimeType -> {
@@ -67,12 +111,14 @@ public value class MimeType(private val mimeType: String) {
                         val folderType = getAttribute(ATTR_FOLDER_TYPE)
                         folderType?.capitalizeAsciiOnly() ?: "XML"
                     }
-
                     else -> "XML"
                 }
             }
-
+            HTML.mimeType -> "HTML"
+            XHTML.mimeType -> "XHTML"
             JSON.mimeType -> "JSON"
+            JSON5.mimeType -> "JSON5"
+            JSON_LINES.mimeType -> "JSON Lines"
             TEXT.mimeType -> "Text"
             REGEX.mimeType -> "Regular Expression"
             GROOVY.mimeType -> if (isGradle()) "Gradle" else "Groovy"
@@ -96,92 +142,98 @@ public value class MimeType(private val mimeType: String) {
             else -> mimeType
         }
 
+    public fun toFileExtensionIfKnown(): String? =
+        when (mimeType) {
+            AGSL.mimeType -> "agsl"
+            AIDL.mimeType -> "aidl"
+            C.mimeType -> "c"
+            CPP.mimeType -> "cpp"
+            CSS.mimeType -> "css"
+            DART.mimeType -> "dart"
+            GO.mimeType -> "go"
+            GRADLE.mimeType -> "gradle"
+            GRADLE_KTS.mimeType -> "kts"
+            GROOVY.mimeType -> "groovy"
+            JAVA.mimeType -> "java"
+            JAVASCRIPT.mimeType -> "js"
+            JSON.mimeType -> "json"
+            JSON5.mimeType -> "json5"
+            JSON_LINES.mimeType -> "jsonl"
+            KOTLIN.mimeType -> "kt"
+            MANIFEST.mimeType -> "xml" // Manifest files
+            PROGUARD.mimeType -> "pro"
+            PROPERTIES.mimeType -> "properties"
+            PROTO.mimeType -> "proto"
+            PYTHON.mimeType -> "py"
+            REGEX.mimeType -> "regex"
+            RESOURCE.mimeType -> "xml" // Resource files
+            RUST.mimeType -> "rs"
+            SHELL.mimeType -> "sh"
+            SQL.mimeType -> "sql"
+            SVG.mimeType -> "svg"
+            TEXT.mimeType -> "txt"
+            TOML.mimeType -> "toml"
+            TYPESCRIPT.mimeType -> "ts"
+            VERSION_CATALOG.mimeType -> "toml"
+            XML.mimeType -> "xml"
+            YAML.mimeType -> "yaml"
+            RUBY.mimeType -> "rb"
+            OBJECTIVE_C.mimeType -> "m"
+            "text/x-erlang" -> "erl"
+            else -> null
+        }
+
+    private fun getBaseMimeType(): String {
+        val baseMimeType = base().toString()
+        // Built-ins are already normalized, don't do string and sorting work
+        if (this in ALREADY_NORMALIZED_BUILTIN_TYPES) return baseMimeType
+
+        return when (baseMimeType) {
+            "text/x-java-source",
+            "application/x-java",
+            "text/x-java" -> JAVA.mimeType
+
+            "application/kotlin-source",
+            "text/x-kotlin",
+            "text/x-kotlin-source" -> KOTLIN.mimeType
+
+            "application/xml" -> XML.mimeType
+            "application/json",
+            "application/vnd.api+json",
+            "application/hal+json",
+            "application/ld+json" -> JSON.mimeType
+
+            "image/svg+xml" -> XML.mimeType
+            "text/x-python",
+            "application/x-python-script" -> PYTHON.mimeType
+
+            "text/dart",
+            "text/x-dart",
+            "application/dart",
+            "application/x-dart" -> DART.mimeType
+
+            "application/javascript",
+            "application/x-javascript",
+            "text/ecmascript",
+            "application/ecmascript",
+            "application/x-ecmascript" -> JAVASCRIPT.mimeType
+
+            "application/typescript" + "application/x-typescript" -> TYPESCRIPT.mimeType
+            "text/x-rust",
+            "application/x-rust" -> RUST.mimeType
+
+            "text/x-sksl" -> AGSL.mimeType
+            "application/yaml",
+            "text/x-yaml",
+            "application/x-yaml" -> YAML.mimeType
+            "application/x-patch" -> PATCH.mimeType
+
+            else -> baseMimeType
+        }
+    }
+
     private fun normalizeString(): String {
-        when (this) {
-            // Built-ins are already normalized, don't do string and sorting work
-            AGSL,
-            AIDL,
-            C,
-            CPP,
-            DART,
-            DIFF,
-            GO,
-            GRADLE,
-            GRADLE_KTS,
-            GROOVY,
-            JAVA,
-            JAVASCRIPT,
-            JSON,
-            KOTLIN,
-            MANIFEST,
-            PATCH,
-            PROGUARD,
-            PROPERTIES,
-            PROTO,
-            PYTHON,
-            REGEX,
-            RESOURCE,
-            RUST,
-            SHELL,
-            SQL,
-            SVG,
-            TEXT,
-            TOML,
-            TYPESCRIPT,
-            UNKNOWN,
-            VERSION_CATALOG,
-            XML,
-            YAML -> return this.mimeType
-        }
-
-        val baseEnd = mimeType.indexOf(';')
-        val normalizedBase =
-            when (val base = if (baseEnd == -1) mimeType else mimeType.substring(0, baseEnd)) {
-                "text/x-java-source",
-                "application/x-java",
-                "text/x-java" -> JAVA.mimeType
-
-                "application/kotlin-source",
-                "text/x-kotlin",
-                "text/x-kotlin-source" -> KOTLIN.mimeType
-
-                "application/xml" -> XML.mimeType
-                "application/json",
-                "application/vnd.api+json",
-                "application/hal+json",
-                "application/ld+json" -> JSON.mimeType
-
-                "image/svg+xml" -> XML.mimeType
-                "text/x-python",
-                "application/x-python-script" -> PYTHON.mimeType
-
-                "text/dart",
-                "text/x-dart",
-                "application/dart",
-                "application/x-dart" -> DART.mimeType
-
-                "application/javascript",
-                "application/x-javascript",
-                "text/ecmascript",
-                "application/ecmascript",
-                "application/x-ecmascript" -> JAVASCRIPT.mimeType
-
-                "application/typescript" + "application/x-typescript" -> TYPESCRIPT.mimeType
-                "text/x-rust",
-                "application/x-rust" -> RUST.mimeType
-
-                "text/x-sksl" -> AGSL.mimeType
-                "application/yaml",
-                "text/x-yaml",
-                "application/x-yaml" -> YAML.mimeType
-                "application/x-patch" -> PATCH.mimeType
-
-                else -> base
-            }
-
-        if (baseEnd == -1) {
-            return normalizedBase
-        }
+        val normalizedBase = getBaseMimeType()
 
         val attributes =
             mimeType
@@ -297,7 +349,12 @@ public value class MimeType(private val mimeType: String) {
         public val XML: MimeType = MimeType("text/xml")
         public val PROPERTIES: MimeType = MimeType("text/properties")
         public val TOML: MimeType = MimeType("text/toml")
+        public val HTML: MimeType = MimeType("text/html")
+        public val XHTML: MimeType = MimeType("application/xhtml+xml")
+        public val CSS: MimeType = MimeType("text/css")
         public val JSON: MimeType = MimeType("text/json")
+        public val JSON5: MimeType = MimeType("text/json5")
+        public val JSON_LINES: MimeType = MimeType("application/x-ndjson")
         public val REGEX: MimeType = MimeType("text/x-regex-source")
         public val GROOVY: MimeType = MimeType("text/groovy")
         public val C: MimeType = MimeType("text/c")
@@ -316,6 +373,9 @@ public value class MimeType(private val mimeType: String) {
         public val SHELL: MimeType = MimeType("application/x-sh")
         public val YAML: MimeType = MimeType("text/yaml")
         public val GO: MimeType = MimeType("text/go")
+        public val RUBY: MimeType = MimeType("text/x-ruby")
+        public val CLOJURE: MimeType = MimeType("text/x-clojure")
+        public val OBJECTIVE_C: MimeType = MimeType("text/x-objectivec")
 
         /** Note that most resource files will also have a folder type, so don't use equality on this mime type */
         public val RESOURCE: MimeType = MimeType("$XML; $ATTR_ROLE=resource")
@@ -326,21 +386,27 @@ public value class MimeType(private val mimeType: String) {
         public val DIFF: MimeType = MimeType("text/x-diff")
         public val PATCH: MimeType = MimeType("text/x-patch")
 
-        /** Maps from a markdown language [name] back to a mime type. */
+        /** Maps from a Markdown language [name] back to a mime type. */
         public fun fromMarkdownLanguageName(name: String): MimeType? =
             when (name) {
                 "kotlin",
                 "kt",
                 "kts" -> KOTLIN
-
+                "clj",
+                "cljs",
+                "cljr",
+                "cljc",
+                "cljd" -> CLOJURE
                 "java" -> JAVA
-                "xml" -> XML
-                "json",
-                "json5" -> JSON
-
+                "html" -> HTML
+                "xhtml" -> XHTML
+                "css" -> CSS
+                "json" -> JSON
+                "json5" -> JSON5
+                "json lines",
+                "jsonl" -> JSON_LINES
                 "regex",
                 "regexp" -> REGEX
-
                 "groovy" -> GROOVY
                 "toml" -> TOML
                 "c" -> C
@@ -354,28 +420,23 @@ public value class MimeType(private val mimeType: String) {
                 "python3",
                 "py",
                 "python" -> PYTHON
-
                 "dart" -> DART
                 "rust" -> RUST
                 "js",
                 "javascript" -> JAVASCRIPT
-
+                "ts",
                 "typescript" -> TYPESCRIPT
                 "sksl" -> AGSL
                 "sh",
                 "bash",
                 "zsh",
                 "shell" -> SHELL
-
                 "yaml",
                 "yml" -> YAML
-
                 "go",
-                "golang" -> YAML
-
+                "golang" -> GO
                 "diff" -> DIFF
                 "patch" -> PATCH
-
                 else -> null
             }
     }
--- a/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/highlighting/CodeHighlighter.kt
+++ b/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/highlighting/CodeHighlighter.kt
@@ -6,7 +6,6 @@ import androidx.compose.ui.text.AnnotatedString
 import kotlinx.coroutines.flow.Flow
 import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
-import org.jetbrains.jewel.foundation.code.MimeType
 
 @ApiStatus.Experimental
 @ExperimentalJewelApi
@@ -26,7 +25,7 @@ public interface CodeHighlighter {
      *
      * @see [NoOpCodeHighlighter]
      */
-    public fun highlight(code: String, mimeType: MimeType?): Flow<AnnotatedString>
+    public fun highlight(code: String, langName: String?): Flow<AnnotatedString>
 }
 
 public val LocalCodeHighlighter: ProvidableCompositionLocal<CodeHighlighter> = staticCompositionLocalOf {
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
@@ -9,6 +9,7 @@ import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.text.withStyle
 import com.intellij.lang.Language
 import com.intellij.lang.LanguageUtil
+import com.intellij.lexer.EmptyLexer
 import com.intellij.openapi.editor.colors.EditorColorsManager
 import com.intellij.openapi.editor.colors.EditorColorsScheme
 import com.intellij.openapi.editor.markup.EffectType
@@ -29,19 +30,45 @@ import org.jetbrains.jewel.bridge.toComposeColorOrUnspecified
 import org.jetbrains.jewel.foundation.code.MimeType
 import org.jetbrains.jewel.foundation.code.highlighting.CodeHighlighter
 
+private val langAliases =
+    mapOf(
+        "js" to "javascript",
+        "py" to "python",
+        "kt" to "kotlin",
+        "kts" to "kotlin",
+        "yml" to "yaml",
+        "sh" to "shell",
+        "bash" to "shell",
+        "zsh" to "shell",
+        // ... other languages
+    )
+
 internal class LexerBasedCodeHighlighter(
     private val project: Project,
     private val reHighlightingRequests: Flow<Unit>,
     private val highlightDispatcher: CoroutineDispatcher = Dispatchers.Default,
 ) : CodeHighlighter {
-    override fun highlight(code: String, mimeType: MimeType?): Flow<AnnotatedString> {
-        val language = mimeType?.toLanguageOrNull() ?: return flowOf(AnnotatedString(code))
-        val fileExtension = language.associatedFileType?.defaultExtension ?: return flowOf(AnnotatedString(code))
+    override fun highlight(code: String, langName: String?): Flow<AnnotatedString> {
+        langName ?: return flowOf(AnnotatedString(code))
+
+        val canonicalName = langAliases[langName.lowercase()] ?: langName.lowercase()
+        val nativeLanguage = LanguageUtil.findRegisteredLanguage(canonicalName)
+
+        val (language, fileExtension) =
+            if (nativeLanguage != null) {
+                val extension = nativeLanguage.associatedFileType?.defaultExtension ?: canonicalName
+                nativeLanguage to extension
+            } else {
+                val textmateLanguage = Language.findLanguageByID("textmate") ?: return flowOf(AnnotatedString(code))
+                textmateLanguage to langName
+            }
+
         val virtualFile = LightVirtualFile("markdown_code_block_${code.hashCode()}.$fileExtension", language, code)
         val colorScheme = EditorColorsManager.getInstance().globalScheme
         val highlighter =
-            SyntaxHighlighterFactory.getSyntaxHighlighter(language, project, virtualFile)
-                ?: return flowOf(AnnotatedString(code))
+            SyntaxHighlighterFactory.getSyntaxHighlighter(language, project, virtualFile)?.takeIf {
+                it.highlightingLexer !is EmptyLexer
+            } ?: return flowOf(AnnotatedString(code))
 
         return flow {
             highlightAndEmit(highlighter, code, colorScheme)
@@ -76,7 +103,11 @@ internal class LexerBasedCodeHighlighter(
         }
     }
 
-    private fun MimeType.toLanguageOrNull(): Language? = LanguageUtil.findRegisteredLanguage(displayName().lowercase())
+    private fun MimeType.toLanguageOrNull(): Language? =
+        when (this) {
+            MimeType.Known.REGEX -> LanguageUtil.findRegisteredLanguage("RegExp")
+            else -> LanguageUtil.findRegisteredLanguage(displayName().lowercase())
+        }
 
     private fun AnnotatedString.Builder.withTextAttributes(
         textAttributes: TextAttributes?,
@@ -101,4 +132,29 @@ internal class LexerBasedCodeHighlighter(
                     else -> null
                 },
         )
+
+    private fun getHighlightingParams(code: String, mimeType: MimeType): HighlightingParams? {
+        val existingLanguage = mimeType.toLanguageOrNull()
+        println("mimeType: $mimeType")
+        val language = existingLanguage ?: Language.findLanguageByID("textmate") ?: return null
+
+        val fileExtension =
+            if (existingLanguage == null) {
+                mimeType.toFileExtensionIfKnown() ?: mimeType.toString().removePrefix("text/x-")
+            } else {
+                language.associatedFileType?.defaultExtension
+            } ?: return null
+
+        val virtualFile = LightVirtualFile("markdown_code_block_${code.hashCode()}.$fileExtension", language, code)
+        val colorScheme = EditorColorsManager.getInstance().globalScheme
+
+        val highlighter =
+            SyntaxHighlighterFactory.getSyntaxHighlighter(language, project, virtualFile)?.takeIf {
+                it.highlightingLexer !is EmptyLexer
+            } ?: return null
+
+        return HighlightingParams(highlighter, colorScheme)
+    }
+
+    private data class HighlightingParams(val highlighter: SyntaxHighlighter, val colorScheme: EditorColorsScheme)
 }
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
@@ -3,7 +3,6 @@ package org.jetbrains.jewel.markdown
 import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.foundation.GenerateDataFunctions
-import org.jetbrains.jewel.foundation.code.MimeType
 
 @ApiStatus.Experimental
 @ExperimentalJewelApi
@@ -54,7 +53,7 @@ public sealed interface MarkdownBlock {
         @ApiStatus.Experimental
         @ExperimentalJewelApi
         @GenerateDataFunctions
-        public class FencedCodeBlock(override val content: String, public val mimeType: MimeType?) : CodeBlock {
+        public class FencedCodeBlock(override val content: String, public val info: String?) : CodeBlock {
             override fun equals(other: Any?): Boolean {
                 if (this === other) return true
                 if (javaClass != other?.javaClass) return false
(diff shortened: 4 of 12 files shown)


platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
```diff
@@ -96,128 +133,55 @@ public value class MimeType(private val mimeType: String) {
             else -> mimeType
         }
 
-    private fun normalizeString(): String {
```
DanielSouzaBertoldi, line 99: Deleted unused code + isRelevantAttribute()

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/CodeAnnotator.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.ui.text.style.TextDecoration
+import androidx.compose.ui.text.withStyle
+import com.intellij.openapi.editor.colors.EditorColorsScheme
+import com.intellij.openapi.editor.markup.EffectType
+import com.intellij.openapi.editor.markup.TextAttributes
+import com.intellij.openapi.fileTypes.SyntaxHighlighter
+import java.awt.Font
+import org.jetbrains.jewel.bridge.toComposeColorOrUnspecified
+
+internal class CodeAnnotator {
```
DanielSouzaBertoldi, line 18: The code here was extracted from `LexerBasedCodeHighlighter` so that we can test it out more easily

platform/jewel/ide-laf-bridge/src/main/resources/intellij.platform.jewel.ideLafBridge.xml
```diff
@@ -3,5 +3,6 @@
     <module name="intellij.libraries.skiko" />
     <module name="intellij.platform.jewel.foundation" />
     <module name="intellij.platform.jewel.ui" />
+    <module name="intellij.markdown" />
```
DanielSouzaBertoldi: Needed to call `CodeFenceLanguageGuesser.guessLanguageForInjection()`

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighterTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+internal class LexerBasedCodeHighlighterTest : LightPlatformTestCase() {
+    private lateinit var highlighter: LexerBasedCodeHighlighter
+
+    override fun setUp() {
+        super.setUp()
+        highlighter = LexerBasedCodeHighlighter(project, emptyFlow())
+    }
+
+    fun testShouldHighlightJavaCodeUsingNativeHighlighter() = runTest {
```
DanielSouzaBertoldi: Yeah `LightPlatformTestCase` extends JUnit **3**. We are forced to prepend `test` to each test case name here :(

platform/jewel/ide-laf-bridge/intellij.platform.jewel.ideLafBridge.iml
```diff
@@ -14,9 +14,7 @@
           </stringArguments>
           <arrayArguments>
             <arrayArg name="pluginClasspaths">
-              <args>
-                <arg>$MAVEN_REPOSITORY$/org/jetbrains/kotlin/kotlin-compose-compiler-plugin/2.2.0/kotlin-compose-compiler-plugin-2.2.0.jar</arg>
-              </args>
+              <args>$MAVEN_REPOSITORY$/org/jetbrains/kotlin/kotlin-compose-compiler-plugin/2.2.0/kotlin-compose-compiler-plugin-2.2.0.jar</args>
```
DanielSouzaBertoldi: No idea why this change happened

platform/jewel/ide-laf-bridge/intellij.platform.jewel.ideLafBridge.iml
```diff
@@ -49,5 +48,10 @@
     <orderEntry type="module" module-name="intellij.libraries.compose.foundation.desktop" />
     <orderEntry type="module" module-name="intellij.platform.editor.ex" />
     <orderEntry type="module" module-name="intellij.libraries.compose.foundation.desktop.junit" scope="TEST" />
+    <orderEntry type="module" module-name="intellij.markdown" />
+    <orderEntry type="module" module-name="intellij.platform.testFramework" scope="TEST" />
+    <orderEntry type="module" module-name="intellij.java.testFramework" scope="TEST" />
```
DanielSouzaBertoldi, line 55: Without this test dependency our `testShouldHighlightJavaCodeUsingNativeHighlighter` test fails since `CodeFenceLanguageGuesser.guessLanguageForInjection()` won't be able to find the Java plugin

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
```diff
@@ -416,8 +416,79 @@ private fun MarkdownExample(project: Project) {
                 |    * But I'm not!
                 |       * Have fun indenting your lists as your heart pleases!
                 |
+                |```kt
```
DanielSouzaBertoldi: Maybe I exaggerated a bit with the amount of examples here 😬

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -569,6 +572,53 @@ public open class DefaultMarkdownBlockRenderer(
...
+            modifier
+                .background(styling.background, styling.shape)
+                .border(styling.borderWidth, styling.borderColor, styling.shape)
+                .thenIf(styling.fillWidth) { fillMaxWidth() }
+                .thenIf(!enabled) { alpha(.5f) },
+        ) {
+            Column(Modifier.padding(styling.padding)) {
+                if (styling.infoPosition.verticalAlignment == Alignment.Top) {
+                    FencedBlockInfo(
+                        block.info,
```
DanielSouzaBertoldi: Another thing to note here:

The `render` that uses a `MimeType` will pass down the language name here instead of what's in the code block (that is, if it knows about the language).

So,

```
\```kt
println("hello!")
\```
```

Will pass down `Kotlin`, while this new implementation will pass `kt`.

rock3r:
> usually code blocks use the file extension instead of the programming language name

I always use the language name personally lol... maybe we should still keep a lang name -> extension that, while it won't cover all cases, at least should cover the ones the IDE supports out of the box?

platform/jewel/ide-laf-bridge/intellij.platform.jewel.ideLafBridge.iml
rock3r: You can revert it

platform/jewel/ide-laf-bridge/intellij.platform.jewel.ideLafBridge.iml
```diff
@@ -49,5 +48,10 @@
     <orderEntry type="module" module-name="intellij.libraries.compose.foundation.desktop" />
     <orderEntry type="module" module-name="intellij.platform.editor.ex" />
     <orderEntry type="module" module-name="intellij.libraries.compose.foundation.desktop.junit" scope="TEST" />
+    <orderEntry type="module" module-name="intellij.markdown" />
```
rock3r: We **really** should not depend on the _entire_ Markdown plugin in the Bridge. Can we get the parts we need extracted (or copied for now, and maybe extracted in the future?)

platform/jewel/ide-laf-bridge/intellij.platform.jewel.ideLafBridge.iml
rock3r: Yeah that's to be expected

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/highlighting/CodeHighlighter.kt
```diff
@@ -26,7 +26,40 @@ public interface CodeHighlighter {
...
+        message =
+            "This method is not scalable as it relies on a pre-resolved MimeType object. " +
+                "This prevents automatic support for languages not explicitly defined in the MimeType system" +
+                "(e.g., from TextMate bundles). Use the overload that accepts the raw " +
+                "`languageFileExtension` string instead.",
+        replaceWith = ReplaceWith("highlight(code, languageFileExtension = \"\")"),
+    )
     public fun highlight(code: String, mimeType: MimeType?): Flow<AnnotatedString>
+
+    /**
```
rock3r, line 39: This paragraph is very bridge-centric for being in the foundation module. You should keep closer to the previous KDoc here, and only be as specific about implementation details in the implementation KDoc.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/highlighting/CodeHighlighter.kt
```diff
@@ -26,7 +26,40 @@ public interface CodeHighlighter {
...
+     *
+     * @param code The source code to be highlighted.
+     * @param languageFileExtension The file extension or Markdown info string identifying the programming language
+     *   (e.g., "kt", "py", "js", "bat"). This is case-insensitive. Common aliases (like "js" for "javascript") are
+     *   supported if there is a native IntelliJ Platform plugin installed that supports the language extension.
+     * @return A [Flow] that emits an `AnnotatedString` with syntax highlighting applied. The flow may emit new values
+     *   if the underlying color scheme changes.
+     * @see [NoOpCodeHighlighter]
+     */
+    public fun highlight(code: String, languageFileExtension: String = ""): Flow<AnnotatedString>
```
rock3r: I do not think we should focus on the _extension_ here; the interface should just give you the raw language name as specified in the Markdown, and it's up to implementations to decide how to deal with that. The fact that it ultimately needs to boil down to an extension is an implementation detail for the textmate-based highlighter in the IDE, but it doesn't mean that other highlighters will need that information.

Say we implement this in the standalone module; in that case, we could simply have a map string -> bundle, with the keys being all possible values for a given language (e.g., `kotlin`, `kt`, `kts`, `kotlinscript`, `gradle.kts`, `gradle-kts`, ...).

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
```diff
@@ -96,128 +133,55 @@ public value class MimeType(private val mimeType: String) {
...
-                .joinToString("; ")
-
-        return if (attributes.isNotBlank()) {
-            "$normalizedBase; $attributes"
-        } else {
-            normalizedBase
+    private fun getBaseMimeType(): String {
+        val baseMimeType = base().toString()
+        // Built-ins are already normalized, don't do string and sorting work
+        if (this in ALREADY_NORMALIZED_BUILTIN_TYPES) return baseMimeType
```
rock3r, line 153: `ALREADY_NORMALIZED_BUILTIN_TYPES` could be a set to make this faster

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
```diff
@@ -371,7 +333,7 @@ public value class MimeType(private val mimeType: String) {
                 "yml" -> YAML
 
                 "go",
-                "golang" -> YAML
+                "golang" -> GO
```
rock3r, line 630: Ooops :) I think we fixed this in Studio (where most of this code is from) but it wasn't brought here too

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
```diff
@@ -96,128 +133,55 @@ public value class MimeType(private val mimeType: String) {
...
+            "application/dart",
+            "application/x-dart" -> DART.mimeType
+
+            "application/javascript",
+            "application/x-javascript",
+            "text/ecmascript",
+            "application/ecmascript",
+            "application/x-ecmascript" -> JAVASCRIPT.mimeType
+
+            "application/typescript" + "application/x-typescript" -> TYPESCRIPT.mimeType
```
rock3r: I don't think that should be a `+`

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
rock3r, line 1: I was thinking, we could rename this to `TextMateCodeHighlighter` for clarity. WDYT?

DanielSouzaBertoldi:
@rock3r I mean, if that's the case then I think we wouldn't even need to call `CodeFenceLanguageGuesser` then, we could always delegate to TextMate since we will be mapping the info block to a given file extension 🤔

rock3r:
That wouldn't work, as we would lose semantic highlighting if we only relied on textmate bundles. I'll finish reviewing asap and will add more context :)

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
```diff
@@ -49,56 +50,61 @@ internal class LexerBasedCodeHighlighter(
...
-
-            while (tokenType != null) {
-                val attributes: TextAttributes? = run {
-                    val attrKey = highlighter.getTokenHighlights(tokenType).lastOrNull() ?: return@run null
-                    colorScheme.getAttributes(attrKey) ?: attrKey.defaultAttributes
-                }
-                withTextAttributes(attributes) { append(tokenText) }
-                advance()
-            }
+    private fun findNativeHighlighter(languageFileExtension: String): SyntaxHighlighter? {
```
rock3r: This valued that gets passed in is not an extension, since we pass it to `guessLanguageForInjection`, which is designed to receive the raw language from the fenced code block. In the Markdown plugin's `CodeFenceInjector`:

```
protected open fun findLangForInjection(element: MarkdownCodeFence): Language? {
  val name = element.fenceLanguage ?: return null
  return CodeFenceLanguageGuesser.guessLanguageForInjection(name).takeIf {
    MarkdownSettings.getInstance(element.project).areInjectionsEnabled
  }
}
```

For example, for Kotlin, this would accept `Kotlin`, `kotlin`, `kt`, and `kts` — see `CodeFenceLanguageAliases`.

platform/jewel/ide-laf-bridge/src/main/resources/intellij.platform.jewel.ideLafBridge.xml
rock3r: We really should not depend on other plugins. This means if the user disables the Markdown plugin, then they disable ALL plugins that use Jewel.

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighterTest.kt
rock3r: You can use `LightPlatform4TestCase` for JUnit 4 ;)

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
```diff
@@ -75,6 +82,31 @@ public sealed interface MarkdownBlock {
 
             override fun toString(): String = "FencedCodeBlock(content='$content', mimeType=$mimeType)"
         }
+
+        @ApiStatus.Experimental
+        @ExperimentalJewelApi
+        @GenerateDataFunctions
+        public class FencedCodeBlockInfo(override val content: String, public val info: String) : CodeBlock {
```
rock3r: Please don't create a parallel class; add new constructors to the existing class:

```kotlin
public class FencedCodeBlock internal constructor(
  override val content: String,
  public val mimeType: MimeType?,
  public val language: String?,
) {
  @Deprecated(...)
  constructor(content: String, mimeType: MimeType?) : this(content, mimeType, mimeType?.language)

  constructor(content: String, language: String?) : this(content, language?.let { MimeType.Known.fromMarkdownLanguageName(it) }, language)

  //...
}
```

This should maintain both binary compatibility and behavioural compatibility.

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
rock3r: Just maybe XD Consider adding a separate tab for the Markdown examples?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
rock3r: Actually, this is not even relying on textmate bundles alone; we should probably call it `IntelliJCodeHighlighter`

platform/jewel/ide-laf-bridge/src/main/resources/intellij.platform.jewel.ideLafBridge.xml
DanielSouzaBertoldi: Ooof, true. Haven't thought of that :(

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighterTest.kt
DanielSouzaBertoldi: Neat, thanks! I read through [this doc](https://plugins.jetbrains.com/docs/intellij/light-and-heavy-tests.html#lightprojectdescriptor) and they didn't mention the junit 4 alternative 😭

rock3r: Ah yeah the doc is probably just old :) I usually look at other similar pieces of code in the platform to see how they do things. You find out all sorts of things that way

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/highlighting/CodeHighlighter.kt
DanielSouzaBertoldi: Great points. Agreed!

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
DanielSouzaBertoldi: ooops!

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/LexerBasedCodeHighlighter.kt
DanielSouzaBertoldi: IntelliJCodeHighlighter to me sounds fine :)

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
DanielSouzaBertoldi: For this to work I'll have to turn MimeType into a class instead of `value class`, since to the compiler both constructors would be identical (both String, String at the end of the day)

rock3r: You can avoid that by swapping `content` (`String`) and `language` (`String?`) maybe?

DanielSouzaBertoldi: Not this time :(

To the compiler, it'll still be the same signature, `String; String` 😭

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageAliases.kt
```diff
@@ -0,0 +1,111 @@
...
+            "python" to "py",
+            "restructuredText" to "rst",
+            "ruby" to "rb",
+            "rust" to "rs",
+            "search result" to "code-search",
+            "shaderlab" to "shader",
+            "shell script" to "sh",
+            "shellscript" to "sh",
+            "terraform" to "tf",
+            "html (twig)" to "html.twig",
```
rock3r, line 84: Can we add `twig` as well?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageAliases.kt
```diff
@@ -0,0 +1,111 @@
...
+            "powershell" to "ps1",
+            "ps" to "ps1",
+            "pwsh" to "ps1",
+            "python" to "py",
+            "restructuredText" to "rst",
+            "ruby" to "rb",
+            "rust" to "rs",
+            "search result" to "code-search",
+            "shaderlab" to "shader",
+            "shell script" to "sh",
```
rock3r, line 78: Can we add `shell`, `bash`, and `zsh` too?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageAliases.kt
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Jewel specific code. This is a way to turn a language name to its main extension.
+     *
+     * So, `python` -> `py`
+     *
+     * This way, TextMate is able to correctly highlight the syntax of the code block.
+     */
+    fun findExtensionGivenName(value: String): String {
```
rock3r:
```suggestion
    fun findExtensionGivenAlias(alias: String): String {
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageGuesser.kt
```diff
@@ -0,0 +1,42 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge.code.highlighting.codefence
+
+import com.intellij.lang.Language
+import com.intellij.lang.LanguageUtil
+import org.jetbrains.annotations.ApiStatus
+
+// Copied from org.intellij.plugins.markdown.injection.aliases.CodeFenceLanguageGuesser
+@ApiStatus.Internal
```
rock3r: We don't need the annotation if it's already internal (will also trigger Detekt if you don't pair it with Jewel's annotation)

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageAliases.kt
```diff
@@ -0,0 +1,111 @@
...
+            Entry("Kotlin", "kotlin", setOf("kt", "kts")),
+            Entry("HCL-Terraform", "terraform", setOf("hcl-terraform", "tf")),
+            Entry("C#", "csharp", setOf("cs", "c#")),
+            Entry("F#", "fsharp", setOf("fs", "f#")),
+            Entry("Shell Script", "shell", setOf("shell script", "bash", "zsh", "sh")),
+        )
+
+    // Jewel specific map
+    private val aliasToMainExtension =
+        mapOf(
```
DanielSouzaBertoldi, line 34: This map was created based on the `package.json` of each language inside `TextMate` bundle.

I omitted extensions that are the same as the language name, such as `.go`, `bicep`, `dart`, `diff` etc.

For programming languages that can map to more than one file extension, the "main" one was used.

So, `kotlin` maps to `kt`, not `kts`, for instance

platform/jewel/samples/ide-plugin/src/main/kotlin/org/jetbrains/jewel/samples/ideplugin/ComponentShowcaseTab.kt
DanielSouzaBertoldi: Done, for now it's just a showcase for markdown blocks. Not sure if it'd be better to just port the markdown example from the first tab to this new one, let me know what you think!

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
DanielSouzaBertoldi: If we want to keep the same class, unfortunately I don't think there's an easy way out of this one... We are forced to push such a breaking change :/

That is, turn `MimeType` into a class... Trying to think of some other way but no ideas yet

rock3r: Let's turn MimeType into a class then :(

faogustavo (approved):

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageAliases.kt
DanielSouzaBertoldi: Sure! `shell` is not listed as a valid alias for `sh` in TextMate, but I guess it wouldn't hurt. `bash` and `zsh` are also valid extensions according to the `package.json`, but I think it's also ok to map it to `sh` anyways 😎

rock3r: Yolo 🚀

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageAliases.kt
rock3r: Nit: maybe worth adding this to the comment on L32? For future reference

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/codefence/CodeFenceLanguageGuesser.kt
```diff
@@ -0,0 +1,39 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.bridge.code.highlighting.codefence
+
+import com.intellij.lang.Language
+import com.intellij.lang.LanguageUtil
+
+// Copied from org.intellij.plugins.markdown.injection.aliases.CodeFenceLanguageGuesser
```
rock3r: Nit: we should note that we temporarily removed support for `CodeFenceLanguageProvider` and `AdditionalFenceLanguageSuggester` extension points. I have an idea on how to fix it in the future tho!

You can reference JEWEL-973 in the comment.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/IntelliJCodeHighlighter.kt
```diff
@@ -0,0 +1,134 @@
...
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.withContext
+import org.jetbrains.jewel.bridge.code.highlighting.codefence.CodeFenceLanguageAliases
+import org.jetbrains.jewel.bridge.code.highlighting.codefence.CodeFenceLanguageGuesser
+import org.jetbrains.jewel.foundation.code.MimeType
+import org.jetbrains.jewel.foundation.code.highlighting.CodeHighlighter
+import org.jetbrains.jewel.foundation.code.highlighting.NoOpCodeHighlighter
+
+internal class IntelliJCodeHighlighter(
```
rock3r, line 38: Nit: KDoc?

platform/jewel/ide-laf-bridge/intellij.platform.jewel.ideLafBridge.iml
```diff
@@ -49,5 +50,9 @@
     <orderEntry type="module" module-name="intellij.libraries.compose.foundation.desktop" />
     <orderEntry type="module" module-name="intellij.platform.editor.ex" />
     <orderEntry type="module" module-name="intellij.libraries.compose.foundation.desktop.junit" scope="TEST" />
+    <orderEntry type="module" module-name="intellij.platform.testFramework" scope="TEST" />
+    <orderEntry type="module" module-name="intellij.java.testFramework" scope="TEST" />
+    <orderEntry type="module" module-name="intellij.textmate" scope="TEST" />
+    <orderEntry type="library" scope="TEST" name="io.mockk.jvm" level="project" />
```
rock3r, line 58: Can we use mockito-kotlin? It's already in the platform

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/IntelliJCodeHighlighter.kt
DanielSouzaBertoldi: Oops, forgot about this one!

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/IntelliJCodeHighlighter.kt
```diff
@@ -0,0 +1,134 @@
...
+    private val highlightDispatcher: CoroutineDispatcher = Dispatchers.Default,
+) : CodeHighlighter {
+    private val codeAnnotator = CodeAnnotator()
+
+    @Deprecated(
+        "This method is not scalable as it relies on a pre-resolved MimeType object." +
+            "This prevents automatic support for languages not explicitly defined in the MimeType" +
+            "system(e.g., from TextMate bundles)." +
+            "Use the overload that accepts the raw `languageFileExtension` string instead.",
+        replaceWith = ReplaceWith("highlight(code, languageFileExtension = null)"),
```
rock3r:
```suggestion
        replaceWith = ReplaceWith("highlight(code, language)"),
```
It's not an extension anymore (maybe other ReplaceWith need the same fix?) and it shouldn't be null, just a value to fill in

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -67,8 +67,16 @@ public class MarkdownProcessor(
     private val markdownMode: MarkdownMode = MarkdownMode.Standalone,
     private val commonMarkParser: Parser =
         MarkdownParserFactory.create(optimizeEdits = markdownMode is MarkdownMode.EditorPreview, extensions),
-    private val languageRecognizer: (String) -> MimeType? = { MimeType.Known.fromMarkdownLanguageName(it) },
 ) {
+    @Deprecated("`languageRecognizer` is not necessary anymore. Use the constructor without it.")
```
rock3r: Nit: `ReplaceWith`?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -569,16 +574,45 @@ public open class DefaultMarkdownBlockRenderer(
         }
     }
 
+    @Deprecated(
+        message =
+            "This class function is not scalable as it relies on a pre-resolved MimeType object. " +
+                "This prevents automatic support for languages not explicitly defined in the MimeType system" +
+                "(e.g., from TextMate bundles). Use FencedCodeBlockInfo instead."
```
rock3r: Is `FencedCodeBlockInfo` still relevant?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -569,16 +574,45 @@ public open class DefaultMarkdownBlockRenderer(
...
+    )
     @Composable
     internal open fun RenderCodeWithMimeType(
         block: FencedCodeBlock,
         mimeType: MimeType,
         styling: MarkdownStyling.Code.Fenced,
         enabled: Boolean,
     ) {
+        JewelLogger.getInstance("DefaultMarkdownBlockRenderer")
```
rock3r: Can you put the log in [`onFirstVisible`](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier#%28androidx.compose.ui.Modifier%29.onFirstVisible%28kotlin.Long,kotlin.Float,androidx.compose.ui.layout.LayoutBoundsHolder,kotlin.Function0%29) so it only prints once?

rock3r: Also, you can use `[this@DefaultMarkdownBlockRenderer.]myLogger()`

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -277,7 +285,7 @@ public class MarkdownProcessor(
     }
 
     private fun FencedCodeBlock.toMarkdownCodeBlockOrNull(): CodeBlock.FencedCodeBlock =
-        CodeBlock.FencedCodeBlock(content = literal.removeSuffix("\n"), mimeType = languageRecognizer(info))
+        CodeBlock.FencedCodeBlock(content = literal.removeSuffix("\n"), mimeType = info)
```
rock3r: I think we should be using the other ctor here :)

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollSyncMarkdownBlockRenderer.kt
```diff
@@ -111,6 +111,13 @@ public open class ScrollSyncMarkdownBlockRenderer(
         }
     }
 
+    @Deprecated(
+        message =
+            "This class function is not scalable as it relies on a pre-resolved MimeType object. " +
+                "This prevents automatic support for languages not explicitly defined in the MimeType system" +
+                "(e.g., from TextMate bundles). Use renderCodeWithBlockInfo instead.",
+        replaceWith = ReplaceWith("renderCodeWithBlockInfo(block, styling, enabled)"),
```
rock3r:
```suggestion
                "(e.g., from TextMate bundles). Use RenderCodeWithLanguage instead.",
        replaceWith = ReplaceWith("RenderCodeWithLanguage(block, styling, enabled)"),
```

platform/jewel/markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/MarkdownProcessorDocumentParsingTest.kt
```diff
@@ -426,7 +425,7 @@ public class MarkdownProcessorDocumentParsingTest {
          * <pre><code class="language-foo+bar">foo
          * </code></pre>
          */
-        parsed.assertEquals(fencedCodeBlock("foo", MimeType.Known.fromMarkdownLanguageName("foo\\+bar")))
+        parsed.assertEquals(fencedCodeBlock("foo", "foo+bar"))
```
rock3r, line 428: We should update `fencedCodeBlock` to use the new ctor

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
```diff
@@ -54,7 +53,8 @@ public sealed interface MarkdownBlock {
         @ApiStatus.Experimental
         @ExperimentalJewelApi
         @GenerateDataFunctions
-        public class FencedCodeBlock(override val content: String, public val mimeType: MimeType?) : CodeBlock {
+        public class FencedCodeBlock internal constructor(override val content: String, public val mimeType: String?) :
```
rock3r:
```suggestion
        public class FencedCodeBlock internal constructor(override val content: String, public val language: String?) :
```

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
```diff
@@ -73,7 +73,9 @@ public sealed interface MarkdownBlock {
                 return result
             }
 
-            override fun toString(): String = "FencedCodeBlock(content='$content', mimeType=$mimeType)"
+            override fun toString(): String {
```
rock3r: Can you revert to the previous style?

platform/jewel/markdown/core/api-dump-experimental.txt
```diff
@@ -101,7 +101,7 @@
 - sf:$stable:I
 - equals(java.lang.Object):Z
 - getContent():java.lang.String
-- f:getMimeType-VwCp3SY():java.lang.String
```
rock3r, line 104: I reckon we can add a deprecated `val mimeType = MimeType.Known.fromBlahBlah(language)` to the fenced code block body (not ctor!) to keep binary compat?

platform/jewel/ide-laf-bridge/intellij.platform.jewel.ideLafBridge.iml
DanielSouzaBertoldi: Luckly for us, `mockk.jvm` is also in the platform 🤯 

I'm a mockK fan, what can I say mwhahah

rock3r: I'm more of a [fakes](https://blog.sebastiano.dev/test-doubles-mocks-stubs-fakes/) person myself :) But if it's already in the platform, mockK it is!

platform/jewel/markdown/core/api-dump-experimental.txt
DanielSouzaBertoldi: That actually worked 🤯

rock3r: image

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
```diff
@@ -197,6 +203,13 @@ public value class MimeType(private val mimeType: String) {
      * For example, for `text/kotlin; role=gradle` this will return `text/kotlin`. For `text/plain; charset=us-ascii`
      * this returns `text/plain`
      */
+    @Deprecated(
```
DanielSouzaBertoldi, line 206: The deprecated message above is reused for any other occasion below, don't bother reading them again and again and again and again

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -549,23 +549,26 @@ public open class DefaultMarkdownBlockRenderer(
...
                         styling.infoPosition.horizontalAlignment
                             ?: error("No horizontal alignment for position ${styling.infoPosition.name}"),
                         styling.infoTextStyle,
                         Modifier.fillMaxWidth().padding(styling.infoPadding),
                     )
                 }
 
-                if (mimeType != null) {
+                if (MIME_TYPE_REGEX.containsMatchIn(language)) {
+                    val languageFromMimeType = language.substringAfter('/').substringAfter("x-")
```
DanielSouzaBertoldi: Since the language was flagged by the regex, our best bet to extract the language name to pass to `fromMarkdownLanguageName()` is to fetch the substring after `/` and/or `x-`, since most of them should follow the same structure:

```
"application/kotlin-source",
"text/x-kotlin",
"text/x-kotlin-source"
```

This solution is not bullet proof though, and actually mapping a mimetype string to a MimeType would require a new map, which we want to avoid.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -549,23 +549,26 @@ public open class DefaultMarkdownBlockRenderer(
...
+                        language,
                         styling.infoPosition.horizontalAlignment
                             ?: error("No horizontal alignment for position ${styling.infoPosition.name}"),
                         styling.infoTextStyle,
                         Modifier.fillMaxWidth().padding(styling.infoPadding),
                     )
                 }
 
-                if (mimeType != null) {
+                if (MIME_TYPE_REGEX.containsMatchIn(language)) {
```
rock3r: I think this should be "matches" — we want the entirety of the string to look like a MIME type, not just part of it

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -68,6 +68,7 @@ import org.jetbrains.jewel.ui.component.HorizontallyScrollableContainer
 import org.jetbrains.jewel.ui.component.Text
 
 private const val DISABLED_CODE_ALPHA = .5f
+private val MIME_TYPE_REGEX = "(\\w+/.+)".toRegex()
```
rock3r:
```suggestion
private val MIME_TYPE_REGEX = "^\\w+/.+$".toRegex()
```

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
rock3r: Why not using `MimeType.Known.find { it.value == language }?.name?.lowercase`? We already have the actual MIME types

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/code/MimeType.kt
```diff
@@ -634,6 +634,59 @@ public value class MimeType(private val mimeType: String) {
...
+
+        /** Maps from a mime type string [mimeType] to a MimeType. */
+        @Deprecated(
+            message =
+                "The MimeType class is deprecated in favor of using the code block info strings (e.g., \"kt\", \"python\"). " +
+                    "This class creates an unnecessary layer of abstraction and requires manual maintenance " +
+                    "to support new languages. Use the new `highlight(code, language)` function " +
+                    "to handle language resolution automatically."
+        )
+        public fun fromMimeTypeString(mimeType: String): MimeType =
```
rock3r: Why not internal?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
DanielSouzaBertoldi: `Known` is an object, not an `enum` class 😭 However, I added the new internal function that can map from a string directly to a MimeType as we discussed 😎

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
DanielSouzaBertoldi: Glad it didn't come to that!

rock3r (approved):
🚢

rock3r:
We'll punt this to 0.31, since we have to figure out why the test fails.

rock3r:
The gods of CI are with us, let's gooooo! Ready to merge

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/code/highlighting/IntelliJCodeHighlighter.kt
```diff
@@ -0,0 +1,176 @@
...
+     * @param code The source code to highlight.
+     * @return A [Flow] that emits styled strings.
+     */
+    private fun createHighlightingFlow(highlighter: SyntaxHighlighter, code: String): Flow<AnnotatedString> {
+        val colorScheme = EditorColorsManager.getInstance().globalScheme
+        return flow {
+            highlightAndEmit(highlighter, code, colorScheme)
+            reHighlightingRequests.collect { highlightAndEmit(highlighter, code, colorScheme) }
+        }
+    }
```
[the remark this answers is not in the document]
rock3r: @faogustavo this is probably worth looking into in a follow-up, can you track it?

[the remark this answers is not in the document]
faogustavo: [JEWEL-1187](https://youtrack.jetbrains.com/issue/JEWEL-1187) Fix the color scheme fetch in the IntelliJCodeHighlighter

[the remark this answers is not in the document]
rock3r: @faogustavo I see that ticket is ready to merge but can't find the PR, is it done in here? I do not see any references here either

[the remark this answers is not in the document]
faogustavo: Yeah, it was done here. I forgot to update it. Sorry 😅 

---

The fix was just changing the color scheme to become a getter. This way we always get the latest version

```
    private val colorScheme: EditorColorsScheme
        get() = EditorColorsManager.getInstance().globalScheme
```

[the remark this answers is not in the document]
rock3r: @DanielSouzaBertoldi could you add JEWEL-1187 to the commit message and PR title too?

faogustavo (approved):
You can remove the disclaimer now. The 253 is already out :)

rock3r:
Did we finally get this to work?!?

PR #3174, opened 2025-08-08
Fix py 82962

chbndrhnns:
When you assign a name to the result of a factory function that produces a class/callable, and then use that name as a callee, navigation tends to follow the factory function (often in stubs) instead of the user’s alias assignment.

This PR fixes this behaviour for:
- NewType
- namedtuple
- Enum
- TypedDict


PetrGolubev:
https://github.com/JetBrains/intellij-community/commit/31edbaae5c293c30ad35724108fabdcf40a3cd4b
https://github.com/JetBrains/intellij-community/commit/3ef9d6f4a43812cface8397698d9a5cabf57cf85

PR #3181, opened 2025-08-13
Keep annotations as-is when renaming a fixture from its declaration site

chbndrhnns:
Fix for https://youtrack.jetbrains.com/issue/PY-83215


KotlinIsland:
this has been merged

PR #3184, opened 2025-08-14
[JEWEL-976] Fix ListComboBox ArrayIndexOutOfBoundsException Error

DanielSouzaBertoldi:
This PR fixes the mentioned exception that's caused by `ListComboBox` when opening the plugin sample

Thanks to @rock3r for actually coming up with the fix 🚀 

# Evidences
| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/24b42f48-bd91-4836-ace7-da95ba3c3f67" /> | <video src="https://github.com/user-attachments/assets/535de2a2-3ea6-4ff8-9c4b-e245e2b4db71" /> |


rock3r (approved):
I may be biased, but ayyyye :Fonzie:

rock3r:
Ready to merge

PR #3186, opened 2025-08-15
Fix ParallelTreeManager cannot find nodeid when using xdist + pytest parameters + teamcity

YueChen-C:
When I use `xdist`+`pytest parameters`+`teamcity` to execute testcases, some testcase data cannot be recorded. I found that `ParallelTreeManager` cannot find `nodeids` when the testcase uses `@pytest.mark.parametrize` :

```python
class Test1:
    @pytest.mark.parametrize("data", [1, 2, 3, 4])
    def test_1(self, data):
        LOG.info(f'test_1->>{data}')
        assert True
```

https://github.com/JetBrains/teamcity-messages/blob/fe6d835390d37d56e967688098eabb5a067614f4/teamcity/messages.py#L27
image

The teamcity class directly uses the `message` method to write data, causing the `ParallelTreeManager` class `get_node_ids` method to fail to obtain the testcase's nodeid.

image


EliseevEgor:
I merge your changes into the master branch!

PR #3187, opened 2025-08-15
[pycharm] ENH: More robust type renderer match routine, and fix for pandas upcoming impactfull changes

dangreb:
While experimenting with a Python environment based on releases provided in python's [scientific nightly wheels conda reepository](https://anaconda.org/scientific-python-nightly-wheels/), i found an imminent conflict between PyCharm's Type Renderers and a pandas enhancement soon to be released.

I opened an [issue at pandas repository](https://github.com/pandas-dev/pandas/issues/62121), with details on the matter. The [mentioned enhancement](https://github.com/pandas-dev/pandas/issues/55178) aims to change how some pandas principal classes are publicly exposed.

I'm opening this request, however, because in my analysis over PyCharm's Type Renderers throughout this case i believe that  we could achieve an additional match functionality based on resolving names into runtime imports, allowing for a flexible correspondence criteria, that could mitigate eventual future divergences between canonical/qualified paths and the ones provided by UI/stubs that are used to save the qualified names in configuration. This could be important considering novel best pratices advising the release of stubs in sparated packages.

The implementation will fisrt attempt to use the utility provided by in _pydevd_bundle.pydevd_import_class to resolve the comparing names, aiming for integration with the IDE toolset. It will fallback to an implementation around importlib in case of error. The match criteria for now will require exact **type** correspondance.

Finally, i employed LRU Cache as to mitigate some overhead the current code may impose. My tests show a very satisfactory overall performance, but the implementation can still be perfected as to eliminate avoidable repetition. Also, the solution seems to suggest that an optional routine for permissive matching criteria based on subclass correspondance could be considered. I intend on investing following efforts around both this remarks next.

I kindly ask you to proceed with review, as well as evaluation over the merits of the fix and the chosen implementation, i'm eager for your feedback! Thanks.

### DIFF
--- a/python/helpers/pydev/_pydevd_bundle/pydevd_user_type_renderers_utils.py
+++ b/python/helpers/pydev/_pydevd_bundle/pydevd_user_type_renderers_utils.py
@@ -1,7 +1,40 @@
 #  Copyright 2000-2021 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
 import inspect
 
+from functools import lru_cache
+import importlib
+
 from _pydevd_bundle import pydevd_utils
+
+try:
+    # Prefer the IDE's own importer if available (handles namespace packages, zip, etc.)
+    from _pydevd_bundle import pydevd_import_class as _pydevd_import_class
+except Exception:
+    _pydevd_import_class = None
+
+@lru_cache(maxsize=256)
+def _resolve_type(path: str):
+    """Resolve 'pkg.mod.Class' -> class object, or None."""
+    if not path:
+        return None
+    # Try pydevd importer first (if present)
+    if _pydevd_import_class is not None:
+        for attr in ('import_name', 'ImportName'):
+            fn = getattr(_pydevd_import_class, attr, None)
+            if callable(fn):
+                try:
+                    return fn(path)
+                except Exception:
+                    pass
+    # Fallback to standard importlib
+    try:
+        mod, _, name = path.rpartition('.')
+        if not mod or not name:
+            return None
+        m = importlib.import_module(mod)
+        return getattr(m, name, None)
+    except Exception:
+        return None
 
 
 class TypeRenderersConstants:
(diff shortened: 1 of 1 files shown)


EliseevEgor:
Since our debugger still supports Python 2.7, we need to adjust the changes a bit

python/helpers/pydev/_pydevd_bundle/pydevd_user_type_renderers_utils.py
```diff
@@ -1,8 +1,41 @@
 #  Copyright 2000-2021 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
 import inspect
 
+from functools import lru_cache
```
EliseevEgor: `functools.lru_cache` is unavailable in Python 2.7

python/helpers/pydev/_pydevd_bundle/pydevd_user_type_renderers_utils.py
```diff
@@ -1,8 +1,41 @@
...
 from _pydevd_bundle import pydevd_utils
 
+try:
+    # Prefer the IDE's own importer if available (handles namespace packages, zip, etc.)
+    from _pydevd_bundle import pydevd_import_class as _pydevd_import_class
+except Exception:
+    _pydevd_import_class = None
+
+@lru_cache(maxsize=256)
+def _resolve_type(path: str):
```
EliseevEgor: Type hints like `path: str` are not supported in Python 2.7

dangreb:
> Thank you for the contribution! Since our debugger still supports Python 2.7, we need to adjust the changes a bit

Thank you for the review, also for noticing the incompatibility! I apologise for the long hiatus, I was on a holiday.

I implemented adjustments, effectively removing both the cache functionality and the type hint!

dangreb:
I took some time over the weekend for deeper tests and some improvements reegarding the general structure of the code. I ran through several scenarios, including Python2.7.18. 

I also assessed it solves the potential issues related to pandas 3.0 ([62121](https://github.com/pandas-dev/pandas/issues/62121), [55178](https://github.com/pandas-dev/pandas/issues/55178)), via tests performed over nightly wheels. 

I believe it should be ready for further reviewing, if of course you wish to consider it's implementation.

Thanks and Regards

EliseevEgor:
I will merge your changes into the master branch!

PR #3189, opened 2025-08-15
Add Goose MCP client detection and configuration support

Kvadratni:
## Summary
This PR adds support for detecting and configuring the Goose AI assistant as an MCP client in IntelliJ IDEA.

## Changes
- Added `GooseClient` class to handle Goose's YAML-based configuration format
- Implemented detection of Goose config at:
  - macOS/Linux: `~/.config/goose/config.yaml`
  - Windows: `%APPDATA%/goose/config.yaml`
- Added YAML parsing logic to check for `jetbrains2` extension presence and enabled status
- Implemented `configure()` method to add/update `jetbrains2` extension in YAML config
- Added Goose to `MCPClientNames` enum and detection flow

## Technical Details
Unlike other MCP clients that use JSON configuration, Goose uses YAML format with an `extensions` section. The implementation:
- Reads and writes YAML configuration without requiring additional dependencies
- Checks for the `jetbrains2` extension in the extensions section
- Adds or updates the extension with proper STDIO settings when configuring
- Returns empty map for `readMcpServers()` since Goose doesn't use JSON mcpServers structure

## Testing
- Verified detection works when Goose config directory exists
- Tested configuration adds/updates the `jetbrains2` extension correctly
- Ensured no impact on existing MCP client detection

## Related
- Similar to existing support for Cursor, VSCode, Claude, etc.
- Follows the same pattern but adapted for YAML configuration format

### DIFF
--- a/plugins/mcp-server/src/com/intellij/mcpserver/clientConfiguration/clients.kt
+++ b/plugins/mcp-server/src/com/intellij/mcpserver/clientConfiguration/clients.kt
@@ -20,6 +20,8 @@ import kotlin.io.path.createParentDirectories
 import kotlin.io.path.exists
 import kotlin.io.path.inputStream
 import kotlin.io.path.outputStream
+import kotlin.io.path.readText
+import kotlin.io.path.writeText
 import kotlin.reflect.jvm.javaMethod
 
 enum class MCPClientNames(val displayName: String){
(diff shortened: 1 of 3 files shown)


nerzhulart:
Hi, thanks a lot for your contribution and for taking the time to add Goose support!
At the moment, there is a product decision to keep the list of supported MCP clients limited to those already included, and we don’t plan to expand it in IntelliJ IDEA itself.
Sorry that you’ve already spent time on this - if you’d like, you could instead open a PR to introduce an extension point and implement Goose support as a separate plugin.

PR #3191, opened 2025-08-18
[JEWEL-980] Add missing internal annotations

rock3r:
We had forgotten to annotate `LocalMessageResourceResolverProvider` as internal and it was leaking the (correctly annotated as internal) `MessageResourceResolver` interface onto the public API.

I found this by pure chance while testing metalava locally.


rock3r:
Gustavo rightly pointed out that I should update the API dumps :D

rock3r:
Ready to merge

PR #3196, opened 2025-08-22
[JEWEL-987] Fix Detekt issues in ide-plugin module

rock3r:
The checks were not being run so a few slipped through. This fixes them and makes sure some basic Detekt form is run on the GitHub CI as well, to avoid repeats.

This happened because we can't run detektMain/detektTest as they require the module to compile, but it can't. IDE plugins need to reference released builds of the platform, which are bound not to have the stuff we haven't released yet.


rock3r:
Ready to merge

PR #3199, opened 2025-08-26
[JEWEL-981, JEWEL-979] Add Metalava signature generator

rock3r:
This adds the Metalava Gradle plugin to the Gradle build. I have also generated the 0.29.0 and 0.30.0 dumps.

> [!NOTE]
> The 0.29.0 sources have been lightly edited to allow Metalava to build those. The changes are [here](https://github.com/JetBrains/intellij-community/compare/32f60827...rock3r:sebp/0.29.0-for-metalava).
> The 0.29.0 dumps have also been lightly manually edited to account for things like APIs being moved across modules and dependencies (moving to use JBR APIs in `decorated-window` and having code moved from `int-ui-decorated-window` to `decorated-window`). Look for `[MANUAL CHANGE]` in those modules' 0.29.0 dumps.

I had originally used an existing Gradle Metalava plugin, but that didn't seem to work well for us. It especially failed at passing custom args to Metalava, which we need to do.

Instead, I forked the setup from https://github.com/autonomousapps/dependency-analysis-gradle-plugin and customised it to better fit our needs:
* Use the latest Metalava
* Fix all the issues I found in it (no idea how it even compiles for them, even though we're on the same Gradle version)
* Add a distinction between stable and experimental dumps, and support handing both at the same time

There is also a new script in `scripts/` which can be used to easily validate or update dumps:

```
Usage: kotlin metalava-signatures.main.kts <command> [options]
Commands:
  update     Generate and update API signature dumps.
  validate   Validate the current signatures against previously generated ones.

Options:
  --release <version>   Tells Metalava to create a versioned release archival dump (or validate against one).
                        If omitted, it will use the current Jewel API version from gradle.properties.
  --module <path>       Only checks/generates the dump for one module (e.g., ':ui').
  --stable-only         Only run tasks for the stable API surface.
  --experimental-only   Only run tasks for the experimental API surface.
  --force               Forces a clean build before running the tasks.
  --help                Display this help and exit.
```

This PR also:
* Updates the IJ icon generator to emit a properly annotated AllIcons keys file
* Runs Metalava on the GitHub Actions CI
* Updates JUnit 5 to the latest version, 5.13.4 [Gradle]
* Fixes the JUnit setup for the detekt-plugin module [Gradle]
* Bumps the IJP target to the latest stable (252.23892.409) [Gradle]
* Removes the old Gradle-based icon generator (JEWEL-979) [Gradle]

Thanks to this setup, I have been able to validate that we do not have breaking API changes between 0.29.0 and 0.30.0 — although as mentioned, in this one-off case it involved a bit of a painful manual process to account for past mistakes.

In the future I'd also like to enable Metalava's API linter, but I have not done it in this PR to keep the scope as small as possible.

> [!NOTE]
> You probably don't want to review the Metalava dumps themselves :)


rock3r:
A million thanks to @liutikas for the help setting up Metalava 🙌

faogustavo (approved):

rock3r:
Ready to merge

PR #3200, opened 2025-08-27
[JEWEL-881] Write 0.30 release notes

rock3r:
This adds the release notes for the 0.30 release.

It also improves the release extractor plugin structure to make it more difficult to invoke it incorrectly and get partial notes (which may or may not have happened to me 🙃).


DanielSouzaBertoldi (approved):
Release notes are all good IMO!

However, I had an idea that probably you already had too but Imma shoot my shot anyways:

What if we could have specific sections for breaking and non-breaking changes instead of just marking changes with (non-breaking) or \*\*breaking\*\*?

Something like

⚠️ Important Changes

Breaking changes
- Change 1
- Change 2

Non-breaking changes
- Change 3
- Change 4

Yes, this would require changing our current template to have a section with some checkboxes (source breaking change/binary breaking change/no breaking change/etc) where the script could look for and it would increase the complexity of the script, buuuuuuuutttttttt it would make it even easier for folks to know what might break if they only care about that.

Probably not worth the effort, but hey, no harm in asking right? 😎

rock3r:
That's not a bad idea, but I would hope we'll stop doing breaking changes :D even for experimental APIs wherever possible

DanielSouzaBertoldi:
That's a good point too! I mean, it's not 100% avoidable depending of what needs to be done (**cough cough** syntax highlighting stuff **cough cough**) but since breaking changes are supposed to be rare, that's one more reason to not really make the effort anyways

rock3r:
Yes, we'll still have to have breaking changes in experimental APIs from time to time, but I reckon they should be the exception overall

rock3r:
Ready to merge

PR #3201, opened 2025-08-27
[JEWEL-952] Ensuring only one submenu is active at the same time

faogustavo:
The problem:

- The submenus' opening was controlled by each "submenu item"
- The "activation" was based on hover, but the "deactivation" was based on dismissing the new submenu popup

The fix:
- Hoisting the selected submenu state to the MenuContent
- With that, we can ensure that only one submenu from that menu is active at the same time

> [!WARNING]
> This fix only works in the custom popup implementations. Using the compose popups has other issues that will be reported to the CMP team

## Evidences

| Case | Video | 
| --- | --- |
| Stand-alone Before | Screen Recording 2025-08-27 at 09 27 51 
| Stand-alone After | Screen Recording 2025-08-27 at 09 24 36
| IDE Before | Screen Recording 2025-08-27 at 09 28 56 |
| IDE After | Screen Recording 2025-08-27 at 09 26 13

### Bug fixes
 * Fixed an issue with multiple sub-menus being presented at the same time

### DIFF
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
@@ -35,6 +35,7 @@ import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -214,7 +215,16 @@ public fun MenuContent(
                 .onHover { localMenuController.onHoveredChange(it) }
     ) {
         Column(Modifier.verticalScroll(scrollState).padding(style.metrics.contentPadding)) {
-            items.forEach { MenuItem(it, anyItemHasIcon, anyItemHasKeybinding) }
+            var selectedSubMenu by remember { mutableStateOf<SubmenuItem?>(null) }
+            items.forEach { item ->
+                MenuItem(
+                    item = item,
+                    showIcons = anyItemHasIcon,
+                    showKeybindings = anyItemHasKeybinding,
+                    selectedSubMenu = selectedSubMenu,
+                    setSelectedSubMenu = { selectedSubMenu = it },
+                )
+            }
         }
 
         Box(modifier = Modifier.matchParentSize()) {
@@ -227,7 +237,23 @@ public fun MenuContent(
 }
 
 @Composable
-private fun MenuItem(item: MenuItem, showIcons: Boolean, showKeybindings: Boolean) {
+private fun MenuItem(
+    item: MenuItem,
+    showIcons: Boolean,
+    showKeybindings: Boolean,
+    selectedSubMenu: SubmenuItem?,
+    setSelectedSubMenu: (SubmenuItem?) -> Unit,
+) {
+    val currentSetSelectedSubMenu by rememberUpdatedState(setSelectedSubMenu)
+
+    fun deselectSubmenu() {
+        currentSetSelectedSubMenu(null)
+    }
+
+    fun selectSubmenu(item: SubmenuItem) {
+        currentSetSelectedSubMenu(item)
+    }
+
     when (item) {
         is MenuSelectableItem ->
             if (item.itemOptionAction != null) {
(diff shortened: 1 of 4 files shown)


rock3r:
Re: the release notes, can you specify that the fix is only for our custom popups?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -239,7 +256,10 @@ private fun MenuItem(item: MenuItem, showIcons: Boolean, showKeybindings: Boolea
                     canShowIcon = showIcons,
                     canShowKeybinding = showKeybindings,
                     enabled = item.isEnabled,
-                    content = item.content,
+                    content = {
+                        LaunchedEffect(it.isHovered) { if (it.isHovered) currentSetSelectedSubMenu(null) }
```
rock3r: I don't love (ab)using state this way. Maybe a local function on L247 could also work instead?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -731,19 +758,20 @@ public fun MenuSubmenuItem(
     submenu: MenuScope.() -> Unit,
     content: @Composable () -> Unit,
 ) {
-    MenuSubmenuItem(canShowIcon, submenu, modifier, enabled, iconKey, interactionSource, style, content)
+    MenuSubmenuItem(canShowIcon, false, submenu, modifier, enabled, iconKey, interactionSource, style) { content() }
```
rock3r: Nit:


```suggestion
    MenuSubmenuItem(canShowIcon, isSelected = false, submenu, modifier, enabled, iconKey, interactionSource, style) { content() }
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -731,19 +758,20 @@ public fun MenuSubmenuItem(
...
 }
 
 @Composable
 internal fun MenuSubmenuItem(
     showIcon: Boolean,
+    isSelected: Boolean,
```
rock3r: Can we call this `selected` instead for consistency?

rock3r: Also, should this value not be passed to the itemState initial creation?

```kotlin
var itemState by
        remember(interactionSource) {
            @Suppress("DEPRECATION") mutableStateOf(MenuItemState.of(selected = selected, enabled = enabled))
        }
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -824,7 +852,7 @@ internal fun MenuSubmenuItem(
             }
         }
 
-        if (itemState.isSelected) {
+        if (isSelected) {
```
rock3r: This change looks sus. Looks like we have two different "isSelected" values now, and only one is used

faogustavo: I'll try to merge the parameter with the itemState

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
faogustavo: Added `deselectSubmenu` and changed to run inside a remember

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
faogustavo: updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
faogustavo: Updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
faogustavo: Merged the itemState with the parameter

faogustavo:
> Re: the release notes, can you specify that the fix is only for our custom popups?

In theory, this fix is also needed for the standard popups. The problem is that there is another bug in the compose API, so we can't really test this

rock3r:
Ready to merge

PR #3202, opened 2025-08-27
[JEWEL-964] Adding PopupMenu with position provider parameter

faogustavo:
### New features
 * Created a PopupMenu variant that takes the PopupPositionProvider as parameter


rock3r:
Ready to merge

PR #3204, opened 2025-08-28
Inline code with correct indentation

chbndrhnns:
Formatting was applied to a detached block and then per-statement after insertion, which broke Python suite indentation and disturbed nearby comments.

Now:
  - Perform one context-aware indentation adjustment (adjustLineIndent) over the contiguous inserted range after insertion; commit/unblock the document first.
  - Detect and delete redundant self-assignment (e.g., `x = x`) from single-return inlining.

Fixes https://youtrack.jetbrains.com/issue/PY-83746/Inline-refactoring-deletes-code
Fixes https://youtrack.jetbrains.com/issue/PY-83747/Inline-refactoring-creates-syntax-error-with-wrong-indentation


KotlinIsland:
this has been merged

PR #3205, opened 2025-08-28
Changed maven search from `search.maven.org` to `central.sonatype.com`

mlytvyn:
Since February 2023, it is advisable to use central.sonatype.com for Maven-related API requests.

> See: https://central.sonatype.org/faq/what-happened-to-search-maven-org/#is-it-possible-to-continue-using-searchmavenorg

`search.maven.org` experiences periodic downtime, which can prevent or even interrupt the downloading of dependencies. With an uptime of around 92%, it is considered unreliable and unstable. This can be confirmed on the following page: https://status.maven.org.

To ensure developers can download Maven dependencies without interruptions, the new official search domain should be used.

> See: https://central.sonatype.org/search/

New API is backward compatible.

### DIFF
--- a/java/java-impl/src/com/intellij/jarFinder/MavenCentralSourceSearcher.java
+++ b/java/java-impl/src/com/intellij/jarFinder/MavenCentralSourceSearcher.java
@@ -22,11 +22,11 @@ public class MavenCentralSourceSearcher extends SourceSearcher {
                               @NotNull String version,
                               @NotNull VirtualFile classesJar) throws SourceSearchException {
     try {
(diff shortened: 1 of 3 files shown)


CLOVIS-AI:
Since this morning, the old site seems to be even less reliable than previously. I've tried multiple times today, but I can't download any dependencies. JetBrains please prioritize!

I created a related ticket: [IDEA-378426](https://youtrack.jetbrains.com/issue/IDEA-378426) Migrate "download sources" from search.maven.org to central.sonatype.org

PR #3206, opened 2025-08-29
Parenthesize multi-line expression when using extract method

chbndrhnns:
Fixes https://youtrack.jetbrains.com/issue/PY-54512/Extract-method-omits-backslashes-when-extracting-from-a-parenthesized-expression-spanning-multiple-lines

### DIFF
--- a/python/python-psi-impl/src/com/jetbrains/python/refactoring/extractmethod/PyExtractMethodUtil.java
+++ b/python/python-psi-impl/src/com/jetbrains/python/refactoring/extractmethod/PyExtractMethodUtil.java
@@ -576,16 +576,37 @@ public final class PyExtractMethodUtil {
       builder.makeAsync();
     }
     final String originalText = expression.getText();
-
-    boolean needsParens =
-      (expression instanceof PyYieldExpression) ||
-      ((originalText.contains("\n") || originalText.contains("\r")) &&
-       ((expression instanceof PyGeneratorExpression) || !isParenthesizedExpression((PyExpression)expression)));
-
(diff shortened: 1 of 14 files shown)


PetrGolubev:
Thank you for your contribution.
I've found a couple of cases when it adds redundant parentheses.

1. When `PyGeneratorExpression` contain parentheses. E.g. for
```
_ = <selection>(print(x) for x in [1, 2])</selection>
```
the method body will be
```
return ((print(x) for x in [1, 2]))
```

2. New lines in brackets/parentheses. E.g. for
```
_ = <selection>[1,
     2,
     3] + [4]</selection>
```
the method body will be
```
return ([1,
         2,
         3] + [4])
```

chbndrhnns:
@PetrGolubev Thanks for these examples. I added them as tests and fixed the code.

PetrGolubev:
This approach has some issues.

Generator expressions.
In the following case the necessary parentheses are not added:
```
print(<selection>i for i in range(1, 10)</selection>)
```
In this case, redundant parentheses are added:
```
_ = <selection>(print(i) for i in range(1,
                             10))</selection>
```


Line breaks. In this example the required parentheses will not be added:
```
def accept(_):
    ...

accept(<selection>[1] +
       [2, 3]</selection>)
```

Explicit line continuation (with a backslash) is not taken into account (redundant parenthesis are added in this case).

The [solution](https://github.com/JetBrains/intellij-community/commit/51ad67e9f7ac67dacc03369de4cfbc84f7d2053a) I used.

PR #3208, opened 2025-08-29
[JEWEL-953] Improving keyboard navigation with custom popups 

faogustavo:
> [!NOTE]
> This PR needs to be merged after #3201 as the only change from it is the second commit

- Updated JDialogRenderer to support dismissOnClickOutside
- Updated JDialogRenderer to handle Key Events using AWT APIs
- Updated JDialogRenderer focus to be cyclic inside of it
- Removed custom focus logic from JDialogRenderer and JBPopupRenderer to prevent losing focus
- Updated JDialogRenderer and JBPopupRenderer to match calls to dismiss callback

## Evidences 

| | Video |
| --- | --- |
| Stand Alone | Screen Recording 2025-08-29 at 11 05 08
| IDE | Screen Recording 2025-08-29 at 11 07 17

### Bug fixes
 * Fixed Keyboard navigation on both JDialogRenderer and JBPopupRenderer
   * Now, clicking the left arrow only returns one level
   * Fixes incorrect focus force on JBPopup
 * Fixed JDialogRenderer keyboard focus to be cyclic, similar to Compose and JBPopupRenderer implementations
 * Added support for `dismissOnClickOutside=false` to JDialogRenderer


rock3r:
Ready to merge

PR #3203, opened 2025-08-28
[JEWEL-994] Add branch comparison script

rock3r:
This PR adds a script to compare branches for differences in Jewel commits. The script can be used to get a list of commits to cherry-pick for a new release, or to sanity check before publishing that all the commits that need to be in a branch are there (master vs release branch, or release branch vs release branch).

> [!NOTE]
> ✅ ~~This requires #3200 (0.30 release notes) to be merged first. Only the second commit remains in this PR.~~ MERGED

PR #3209, opened 2025-08-29
[JEWEL-997] Write Maven artefacts validation script

rock3r:
This adds a new script, validate-maven-artifacts.main.kts, that ensures that the Jewel artefacts publishing is set up correctly. The script will flag:
 * Differences in `JewelMavenArtifacts` across branches
 * Artefacts not being published in one or more branches
 * Artefact dependencies being different across branches

Usage: `validate-maven-artifacts [<options>] <branches>...`

> [!IMPORTANT]
> ✅ ~~This PR requires #3203 to be merged first.~~ MERGED

This also fixes the usage help and readme for jps-bootstrap to mention that you need to pass the project root (the ultimate or community root path) to the script. There was no mention of it, and I was having issues with the script, so after checking with @chashnikov, we agreed that should be updated. I also asked @shalupov, and he also confirmed.

Example run:


platform/jps-bootstrap/src/main/java/org/jetbrains/jpsBootstrap/JpsBootstrapMain.kt
```diff
@@ -88,6 +88,7 @@ class JpsBootstrapMain(args: Array<String>?) {
       jarFileTarget = null
     }
     else {
+      println("!!! ARGS: $freeArgs")
```
shalupov, line 91: pls remove :)

rock3r:
Random failure, will rebase to re-trigger the CI

platform/jewel/scripts/validate-maven-artifacts.main.kts
```diff
@@ -0,0 +1,623 @@
...
+        if (!noBuild) {
+            val gitStatus = runCommand("git status --porcelain", communityRoot, exitOnError = true).output.trim()
+            if (gitStatus.isNotEmpty()) {
+                println()
+                exitWithError(
+                    "Local changes detected in the git repository at ${communityRoot.canonicalPath}. " +
+                        "Please commit or stash them."
+                )
+            }
+        }
```
faogustavo, line 108: Nit: Can you extract the conditions before this line to an external function? Just to reduce the number of lines from this. Something like `checkPreConditions` or a similar function name

platform/jewel/scripts/utils.main.kts
```diff
@@ -145,6 +154,13 @@ fun String.asError() = if (stylingSupported) "$CODE_ERR$this$CODE_CLEAR" else th
  */
 fun String.asBold(): String = if (stylingSupported) "$CODE_BOLD$this$CODE_CLEAR" else this
 
+/**
+ * Formats a string as a success (green) text with ANSI escape codes if the terminal supports it.
+ *
+ * @return The formatted text, or the text itself if the terminal does not support styling.
+ */
+fun String.asSuccess(): String = if (stylingSupported) "$CODE_SUCCESS$this$CODE_CLEAR" else this
```
faogustavo, line 162: Nit: The `stylingSupported` does not work in my environment (Using TMux and Ghostty - which support colors/theme). We probably need to fix the variable condition to something like this:

```kotlin
private fun doesTerminalSupportStyling(): Boolean =
    System.getenv("TERM")?.contains("color") == true || System.getenv("COLORTERM") == "truecolor"
```

Note that as you will be changing this `utils` file and not the original, you need to delete the caches to make it work (took me some time to discover [this](https://youtrack.jetbrains.com/projects/KT/issues/KT-42101/Scripts-fileImport-in-kotlin-main-kts-uses-a-stale-cache) 🤦‍♂️ )

platform/jewel/scripts/validate-maven-artifacts.main.kts
```diff
@@ -0,0 +1,623 @@
...
+                println(diffResult.output)
+            } else {
+                printlnErr("Could not compute diff for $filePath between branches '$referenceBranch' and '$branch'.")
+            }
+        }
+
+        print("Do you want to proceed with the build anyway? (y/n): ")
+        if (readlnOrNull()?.trim()?.lowercase() != "y") {
+            exitWithError("Build cancelled by user due to file inconsistencies.")
+        }
```
faogustavo: Maybe use the [prompt](https://github.com/ajalt/clikt/blob/e3133d41e26978ab0b9239082317ebe75f9d7e4c/samples/copy/src/main/kotlin/com/github/ajalt/clikt/samples/copy/main.kt#L31)?

```kotlin
if (YesNoPrompt("Do you want to proceed with the build anyway?", terminal, default = true).ask() == false) {
  exitWithError("Build cancelled by user due to file inconsistencies.")
}
```

faogustavo (approved):

platform/jewel/scripts/utils.main.kts
rock3r: >Note that as you will be changing this utils file and not the original, you need to delete the caches to make it work (took me some time to discover this 🤦‍♂️ )

Yikes. Thanks for the heads up! I'll need to change the main script anyway for the other comments, but this would have been a pain to figure out otherwise

platform/jewel/scripts/validate-maven-artifacts.main.kts
rock3r: The reason I have not done it is that this not only checks preconditions, but also determines the IJ root dir, which is then used later. I feel a function that does both things would be weird. I also would avoid changing the order too much... but I suppose if I move the git checks above the IJ root search I can at least extract the rest

rock3r: Ah no nevermind, the order can't be changed as those also require the root dir.

rock3r: I don't see a sensible way to do it in a way that doesn't break SRP. It's 30 lines in a 60 lines function, I think it's still reasonable

platform/jewel/scripts/validate-maven-artifacts.main.kts
rock3r: TIL!

rock3r:
Ready to merge

PR #3210, opened 2025-08-30
When extracting a method into a class, only validate against the clas…

chbndrhnns:
…s’ own namespace

The name-clash validator in the Python Extract Method utility used to check all enclosing scopes outward up to the module and treated any declaration it finds as a conflict, including module-level declarations.

Fixes https://youtrack.jetbrains.com/issue/PY-61591/Refactor-Extract-method-for-instance-class-complains-about-existing-symbol-in-module


PetrGolubev:
The change introduces another issue:
```
class A:
    def foo(self):
        def bar():
            print('bar')

```
Invoke "Extract method..." at line 4 and enter name "bar".
No name clash alert will be shown.

chbndrhnns:
@PetrGolubev Thanks for the feedback. I added checks for the case you reported.

PetrGolubev:
I merged your first [commit](https://github.com/JetBrains/intellij-community/commit/eeadf185e8a1665150c9bcb5ff464562d0fc5ac9) and [fixed](https://github.com/JetBrains/intellij-community/commit/36257a1ee1ab4f462cc1d7a62ce5746fdf3b7c49) the case with nested code blocks.

PR #3211, opened 2025-09-02
[IJPL-190726] Fix(macOS): Handle corrupted PWD from 'open -na' command

ImTotem:
### Description
This PR fixes a long-standing issue on macOS where launching the IDE via the Toolbox script fails if the current path contains non-ASCII characters.

### The Problem
The `open -na` command causes a failure when it launches a native executable under the following two conditions:

1. The executable is launched via `open -na`.
2. The `PWD` environment variable passed to the `open` command contains non-ASCII characters (e.g. `한글`Korean).

### **Related Issue:**
Fixes [IJPL-190726](https://youtrack.jetbrains.com/issue/IJPL-190726/Command-line-interface-Korean-UTF-8-paths-appear-incorrectly-ISO-8859-1-encoded)
Duplicate with [IJPL-148652](https://youtrack.jetbrains.com/issue/IJPL-148652)

### DIFF
--- a/native/XPlatLauncher/src/lib.rs
+++ b/native/XPlatLauncher/src/lib.rs
@@ -30,6 +30,8 @@ variant_size_differences
 )]
 
 use std::env;
+use std::ffi::OsStr;
+use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, bail, Context, Result};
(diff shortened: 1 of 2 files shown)


trespasserw:
Thanks! I'll try to take a look later this week.

ImTotem:
​Hi, just checking in for any updates on this PR.

​As a side note, Rust isn't my primary language, so I'd appreciate your feedback on the implementation details.

​I considered decoding from the system locale, but that would require adding the encoding_rs dependency, and I was cautious about adding a new dependency without discussion. So, I decided that a simple UTF-8 decoding would be a reasonable and safe approach, given that many systems use it for their locale.

​Thanks!

trespasserw:
Sorry, still didn't find time :( I just find the situation strange and didn't manage to reproduce the problem yet.

ImTotem:
Thanks for the update! No worries at all, please take your time.

ImTotem:
Duplicate with [IJPL-148652](https://youtrack.jetbrains.com/issue/IJPL-148652)

trespasserw:
Sorry for it taking so long.

I still didn't manage to find out why does macOS behave the way it does, even with all the help from great models we have nowadays. Applied is a somewhat modified version of your fix, which transforms the path only when a directory does not exist.

Thanks!

PR #3212, opened 2025-09-02
[JEWEL-146] Speed Search styling

faogustavo:
> [!NOTE] 
> - This is PR **1/3**
> - The release notes will be set in the PR that adds the component the user can "interact with"

- Created new classes to hold the Speed Search theme configuration
- Created IntUI and Bridge methods to fill the theme

## Evidences

| Case | Light | Dark |
| --- | --- | --- |
| Found Entry | image | image|
| Missing Entry | image | image


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
```diff
@@ -106,7 +110,82 @@ public class DefaultComponentStyling(
     public val textFieldStyle: TextFieldStyle,
     public val tooltipStyle: TooltipStyle,
     public val undecoratedDropdownStyle: DropdownStyle,
+    public val speedSearchStyle: SpeedSearchStyle?,
+    public val searchMatchStyle: SearchMatchStyle?,
```
faogustavo: I was not sure how to provide defaults for those. So I'm setting them as nullable and filtering out in the "styles". I'm open to ideas on how to handle this better, as I was not able to find one 😓

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchMatchStyle.kt
```diff
@@ -0,0 +1,63 @@
...
+        result = 31 * result + endBackground.hashCode()
+        result = 31 * result + foreground.hashCode()
+        return result
+    }
+
+    override fun toString(): String =
+        "SearchMatchColors(startBackground=$startBackground, endBackground=$endBackground, foreground=$foreground)"
+
+    public companion object {
+        public fun defaultForegroundColor(isDark: Boolean): Color = if (isDark) Color(0xFF000000) else Color(0xFF323232)
```
rock3r: I really don't like this being here. I know it duplicates things a bit, but I'd rather inline this constant in the bridge and standalone style creation. This way if the IDE ever changes how they handle it we don't need to have deprecations here in the `ui` module.

rock3r:
Would be good to have screenshots/videos given this is about styling :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
rock3r: They should be done exactly like the others — non-nullable. Can we have a private default fallback for these styling classes that make it look hideous and forces people to use the right constructor? :D

It's already at level ERROR, so we only need to maintain binary compat here; people that don't update this usage are also likely not setting up speed search in their codebase, so it should be invisible to them anyway.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchMatchStyle.kt
faogustavo: Makes sense. Let me change that

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
faogustavo: Updated. Added a "default" method that is internal to this module and replicates the light theme

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchMatchStyle.kt
faogustavo: ✅

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchMatchStyle.kt
```diff
@@ -0,0 +1,71 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.component.styling
+
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.ui.graphics.Color
+import org.jetbrains.jewel.foundation.GenerateDataFunctions
+
+@GenerateDataFunctions
+public class SearchMatchStyle(public val colors: SearchMatchColors) {
```
rock3r: We don't seem to be accounting for it yet, but it seems like we should include the metrics too, since we should if not now at least in a future release draw the highlight by ourselves. According to `com.intellij.util.ui.UIUtil#drawSearchMatch(java.awt.Graphics2D, float, float, int, java.awt.Color, java.awt.Color)`, there are:
 * A padding compared to the line height: `PaddingValues(vertical = 2.dp)`
 * A corner radius of `2.5.dp`

We need to check that using these values in Jewel produces the expected results compared to the Swing alternative, and in case tweak them accordingly, but in any case we should have them already to spare us the deprecation cycle in the (near?) future.

faogustavo: Added the metrics. I'll update the component implementation to start including it

rock3r:
Ready to merge

PR #3213, opened 2025-09-02
[JEWEL-146] Speed Search string matcher

faogustavo:
> [!NOTE] 
> - This is PR **2/3**
> - The release notes will be set in the PR that adds the component the user can "interact with"

- To support matching entries in a list, we must implement the classes/methods available in the IC modules
- Created SpeedSearchMatcher to be the interface used for matching
- Created SubstringSpeedSearchMatcher to perform text search for the whole search as a substring
- Created PatternSpeedSearchMatcher to perform text search similar to the IntelliJ search


platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,135 @@
...
+ * Examples:
+ * ```kotlin
+ * val matcher = SpeedSearchMatcher.substringMatcher("foo")
+ * matcher.matches("foobar") // [0..3]
+ * matcher.matches("foboar") // null
+ * ```
+ *
+ * If you need a more flexible way to match text, use [patternMatcher].
+ */
+public fun SpeedSearchMatcher.Companion.substringMatcher(
```
rock3r: Any specific reason these are extension functions instead of being members of the companion object?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,135 @@
...
+public fun interface SpeedSearchMatcher {
+    public fun matches(text: String?): List<IntRange>?
+
+    public companion object
+}
+
+/**
+ * Returns a [SpeedSearchMatcher] that searches for the given [pattern] in the text.
+ *
+ * Note that this method only matches if the [pattern] is a substring of the text.
```
rock3r: What happens if the substring is present multiple times in a string?
I.e., in "an eye for an eye", if I match "eye" it should return two matches, right? And if I match for "oo" in "goooo" does it return overlapping matches too (i.e, `[1,2],[2,3],[3,4]`), or only two (`[1,2],[3,4]`)?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,135 @@
...
+    pattern: String,
+    ignoreCase: Boolean = true,
+): SpeedSearchMatcher = SubstringSpeedSearchMatcher(pattern, ignoreCase)
+
+/**
+ * Returns a [SpeedSearchMatcher] that searches for the given [pattern] in the text. This pattern can match multiple
+ * parts of the string submitted in the [SpeedSearchMatcher.matches] function, not needing the whole string to match as
+ * one substring.
+ *
+ * Examples:
```
rock3r: Same question about multiple partial matches here, too. I think in Swing they only ever use this matcher, and it seems like they don't do multiple matches, so it should be fine not to, but it should be documented.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Case sensitivity options for matching.
+ *
+ * **Swing equivalent:**
+ * [NameUtil.MatchingCaseSensitivity](https://github.com/JetBrains/intellij-community/blob/master/platform/util/text-matching/src/com/intellij/psi/codeStyle/NameUtil.java)
+ */
+public enum class MatchingCaseSensitivity {
+    NONE,
+    FIRST_LETTER,
+    ALL,
```
rock3r:
```suggestion
    None,
    FirstLetter,
    All,
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * Split the input into words based on case changes, digits, and special characters, and join them with the wildcard
+ * ('*') character.
+ *
+ * **Swing equivalent:**:
+ * [SpeedSearchComparator.obtainMatcher](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/ui/SpeedSearchComparator.java)
+ */
+internal fun String.convertToPattern(matchFromBeginning: Boolean): String {
```
rock3r, line 163: No tests for this?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+    private val ignoreCase = options != MatchingCaseSensitivity.ALL
+
+    private val hasHumps: Boolean
+    private val hasSeparators: Boolean
+    private val hasDots: Boolean
+    private val meaningfulCharacters: CharArray
+    private val minNameLength: Int
+
+    init {
+        val meaningful = StringBuilder()
```
rock3r: Nit: use `buildString`?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+            patternLowerCase[k] = c.lowercaseChar()
+
+            if (!isWildcard(k)) {
+                meaningful.append(patternLowerCase[k])
+                meaningful.append(patternUpperCase[k])
+            }
+        }
+        var start = 0
+        while (isWildcard(start)) start++
+        hasHumps = hasFlag(start + 1, isUpperCase) && hasFlag(start, isLowerCase)
```
rock3r: Nit for readability:

```suggestion
        hasHumps = hasFlag(start, isLowerCase) hasFlag(start + 1, isUpperCase)
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+        hasSeparators = hasFlag(start, isWordSeparator)
+        hasDots = (start until pattern.size).any { pattern[it] == '.' }
+        meaningfulCharacters = meaningful.toString().toCharArray()
+        minNameLength = meaningfulCharacters.size / 2
```
rock3r: Not entirely sure how this works, can you explain it?

Also, there are a few mentions of "name" in this file, but I don't know what it is, if you could clarify that too. It looks like it's the text we're try to match against, but then if it is, calling it "name" is super confusing :D

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+    private fun String.matchingFragments(): List<IntRange>? {
+        if (length < minNameLength) {
+            return null
+        }
+
+        if (pattern.size > MAX_CAMEL_HUMP_MATCHING_LENGTH) {
+            return containsMatcher.matches(this)
+        }
+
+        val length = length
```
rock3r: Can be inlined?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+            return matchFragment(patternIndex, nameIndex, isAscii)
+        }
+
+        var currentPatternIndex = patternIndex
+        do {
+            currentPatternIndex++
+        } while (isWildcard(currentPatternIndex))
+
+        if (currentPatternIndex == pattern.size) {
+            // Suppressing condition to keep the structure as similar as the Java/Swing files
```
rock3r: Nit:
```suggestion
            // Suppressing condition to keep the structure as similar as the Java/Swing files as possible
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+                    return listOf(IntRange.from(spaceIndex, 1))
+                }
+            }
+            return null
+        }
+
+        return matchSkippingWords(
+            currentPatternIndex,
+            findNextPatternCharOccurrence(nameIndex, currentPatternIndex, isAscii),
+            true,
```
rock3r: Nit: avoid unnamed bool parameters


```suggestion
            allowSpecialChars = true,
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+
+    private fun String.findNextPatternCharOccurrence(startAt: Int, patternIndex: Int, isAscii: Boolean): Int =
+        if (!'*'.isPatternChar(patternIndex - 1) && !isWordSeparator[patternIndex]) {
+            indexOfWordStart(patternIndex, startAt, isAscii)
+        } else {
+            indexOfIgnoreCase(startAt, pattern[patternIndex], patternIndex, isAscii)
+        }
+
+    private fun String.checkForSpecialChars(start: Int, end: Int, patternIndex: Int): Int {
+        if (end < 0) return -1
```
rock3r: Not sure if possible, but hey, wouldn't hurt I guess?

```suggestion
        if (end < 0 || end < start) return -1
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,530 @@
...
+            listOf(IntRange.from(from, length)) + ranges
+        }
+    }
+}
+
+private fun IntRange.Companion.from(startOffset: Int, length: Int) = startOffset..startOffset + length
+
+private fun Char.isUpperCaseOrDigit(): Boolean = isUpperCase() || isDigit()
+
+private fun Char.hasCase(): Boolean = isUpperCase() || isLowerCase()
```
rock3r, line 433: Isn't this the same as `isLetter`? Or is there a semantic difference between the two (e.g., letters that are neither uppercase nor lowercase)?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/SubstringSpeedSearchMatcher.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import org.jetbrains.jewel.foundation.search.SpeedSearchMatcher
+
+/**
+ * Tells whether a string matches a specific substring. Allows for lowercase substring matching.
+ *
+ * **Swing equivalent:**
+ * [MinusculeMatcherImpl.matchBySubstring](https://github.com/JetBrains/intellij-community/blob/master/platform/util/text-matching/src/com/intellij/psi/codeStyle/MinusculeMatcherImpl.java)
+ */
+internal class SubstringSpeedSearchMatcher(private val pattern: String, private val ignoreCase: Boolean = true) :
```
rock3r: Should we call this `ExactSubstringSpeedSearchMatcher` to better highlight the difference in behaviour?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: None that I recall. I'll move to the companion to avoid extra import

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: If I recall correctly, it will only return the first occurrence in that string. I'll add to the documentation and some tests.

faogustavo: Same happens in the IJ APIs. Only the "first" occurrence

image

rock3r: Yes, but theirs is the pattern matcher. They don't really do exact substring matching. Anyway, probably better to keep consistency between the two. Just make sure the KDoc mentions this only matches the first occurrence (same for pattern)

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: So, this is used to check if the size of the string I'm testing is at least the same size as the pattern's valid characters (excluding spaces and `*`).

Note that this is dividing by 2 because we are adding the upper and lower case versions of the character from the pattern in this string.

---

For the "name" part, some I missed updating for a more meaningful variable name. When I was moving the implementation from Java (which uses `name` for everything), I copied the same variables with the old name by mistake. I'll take another look to rename to something more meaningful

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: Something I can do is hide this option. I just exposed in case final users wanted to use a "light-weight" version matcher. This does the exactly same thing from the MinusculeMatcherImpl.matchBySubstring

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: 🤦‍♂️ probably some refactor I forgot to clean up LOL

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: Going down the rabbit hole, I think they do the same thing. I'll change and test it.

Just used the same logic from java when added it:

```java
  private static boolean hasCase(char patternChar) {
    return Character.isUpperCase(patternChar) || Character.isLowerCase(patternChar);
  }
```

faogustavo: So, according to Junie and Gemini, they are different. Using the `isLetter` has more supported languages/codes (see table below), but that would be a different behavior from the IDE current implementation. 

| Character | Code Point | isLetter() | isUpperCase() \|\| isLowerCase() | Reason for Mismatch |
| -------- | --------- | ----------- | ------------------------------- | ------------------ |
| A         | \u0041   | true | true | Match (Uppercase)   |
| a         | \u0061   | true | true | Match (Lowercase)   |
| ǅ         | \u01C5   | true | false | Titlecase Letter    |
| 你         | \u4F60   | true | false | Other Letter        |
| 5         | \u0035   | false | false | Match (Not a letter)|

What are your thoughts on this? IMO we can stay with upper/lower validation for now.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: Updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: Added in the examples and updated matcher name

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: Updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: 👀 Added

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: Updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: Makes sense. Updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: Updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: Updated

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: Added it

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
faogustavo: Added it. I don't think it would happen, but it doesn't hurt

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/SubstringSpeedSearchMatcher.kt
faogustavo: Renamed

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/PatternSpeedSearchMatcher.kt
rock3r: Yeah, was mostly curiosity :) The current impl is fine.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
rock3r: I think we are ok with having it, as long as we're clear this is not the default behavior in the IDE, which uses the pattern matcher instead.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,141 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.foundation.search
+
+import org.jetbrains.jewel.foundation.search.impl.ExactSubstringSpeedSearchMatcher
+import org.jetbrains.jewel.foundation.search.impl.PatternSpeedSearchMatcher
+
+public fun interface SpeedSearchMatcher {
+    public fun matches(text: String?): List<IntRange>?
+
+    public companion object {
```
rock3r, line 15: Can you add the parameters KDocs for these two functions?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,141 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.foundation.search
+
+import org.jetbrains.jewel.foundation.search.impl.ExactSubstringSpeedSearchMatcher
+import org.jetbrains.jewel.foundation.search.impl.PatternSpeedSearchMatcher
+
+public fun interface SpeedSearchMatcher {
+    public fun matches(text: String?): List<IntRange>?
```
rock3r: Missing KDocs :)

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Case sensitivity options for matching.
+ *
+ * **Swing equivalent:**
+ * [NameUtil.MatchingCaseSensitivity](https://github.com/JetBrains/intellij-community/blob/master/platform/util/text-matching/src/com/intellij/psi/codeStyle/NameUtil.java)
+ */
+public enum class MatchingCaseSensitivity {
+    None,
+    FirstLetter,
+    All,
```
rock3r: Can you add a bit of KDoc, with examples for `FirstLetter`?

platform/jewel/foundation/src/test/kotlin/org/jetbrains/jewel/foundation/search/ConvertToPatternTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    @Test
+    public fun `should split digits from separators`() {
+        assertEquals("*-*foo*_*bar", "-foo_bar".convertToPattern(false))
+        assertEquals("-*foo*_*bar", "-foo_bar".convertToPattern(true))
+    }
+
+    @Test
+    public fun `should keep blank input unchanged`() {
+        assertEquals("", "".convertToPattern(false))
+        assertEquals(" ", " ".convertToPattern(false))
```
rock3r, line 43: Nit: missing the `true` cases

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
rock3r: Also, I was just thinking, is there a better way to return a result than a nullable list? E.g.,

```kotlin
sealed interface MatchResult {
  object NoMatch: MatchResult

  @GenerateDataFunctions
  class Match(public val ranges: List<IntRange>): MatchResult { ... }
}
```

Not a huge fan of using nullable returns when we can do sealed results

platform/jewel/foundation/src/test/kotlin/org/jetbrains/jewel/foundation/search/ConvertToPatternTest.kt
faogustavo: Good catch, thanks. Added those.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: Added it. Also updated the test cases to add a few more :)

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: Added the KDoc.

-- 

For the return null, I used it as an "optimization". Initially, I had a sealed class, but I noticed a significant number of memory allocations and increased RAM usage.

I know that our sample in the "ChipsAndTree" is not the most accurate scenario; but, in that case (where we have 1 million entries), it was almost getting to 1GB. I also don't enjoy it, but I decided to make it this way intentionally. 

However, I'll retest it and gather more data to ensure it won't affect anything. I'm not sure if I was using an object for "NoMatch" or returning an empty list.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
faogustavo: added

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
rock3r: If there is a significant memory pressure difference, let me know. We can keep the nullable return. I was thinking we could further reduce allocations by using a value class for `Match`, by the way.

faogustavo: Pushed the refactor to a sealed type. I'll rebase the other PR and retest it.

faogustavo: This solution has almost no usage impact; we can keep it. I think the problem was when I returned the empty list 🤦‍♂️ 

Thanks for the suggestion :)

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -0,0 +1,219 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.foundation.search
+
+import org.jetbrains.jewel.foundation.GenerateDataFunctions
+import org.jetbrains.jewel.foundation.search.impl.ExactSubstringSpeedSearchMatcher
+import org.jetbrains.jewel.foundation.search.impl.PatternSpeedSearchMatcher
+
+public fun interface SpeedSearchMatcher {
+    /** Returns a list of ranges from the [text] where the pattern matches, or `null` if the pattern does not match. */
```
rock3r: This KDoc needs updating :P

faogustavo: Updated

rock3r:
Ready to merge

PR #3214, opened 2025-09-02
[JEWEL-146] Speed Search Area 

faogustavo:
> [!NOTE] 
> - This is PR **3/3**
> - Other PRs are #3212 and #3213

- Creates the SpeedSearchArea component to control the speed search popup
- Created `SpeedSearchableLazyColumn` component as an extension from `SpeedSearchScope`
- Updated SimpleListItem to support AnnotatedString
- Updated SelectableLazyColumn to receive interactionSource
- Updated SelectableColumnOnKeyEvent to support "filtering" which items can be accessed
- Updated Text Match ranges to return open-ended ranges
  - This was needed to improve the render from the highlight

## Evidences

| Speed Search in selectable lazy column |
| --- | 
| Screen Recording 2025-09-02 at 13 49 02 |

### New features
 * Created `SpeedSearchSelectableLazyColumn` component that allows quickly finding the entry by its text
   * To allow users a more flexible solution, we are exposing the `SpeedSearchMatcher` so you can provide your own implementation
   * Out of the box, we support `SpeedSearchMatcher.substringMatcher()` and `SpeedSearchMatcher.patternMatcher()`
 * 

### Bug fixes
 * Fixed issue in the `LazyListState.visibleItemsRange` and  `SelectableLazyListState.visibleItemsRange` methods that were returning one item over the amount of visible items

### Deprecated API
 * Deprecated `SelectableLazyColumn` component that does not support an `interactionSource` as parameter


platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -55,8 +89,11 @@ public fun SelectableLazyColumn(
     flingBehavior: FlingBehavior = ScrollableDefaults.flingBehavior(),
     keyActions: KeyActions = DefaultSelectableLazyColumnKeyActions,
     pointerEventActions: PointerEventActions = DefaultSelectableLazyColumnEventAction(),
+    interactionSource: MutableInteractionSource? = null,
     content: SelectableLazyListScope.() -> Unit,
 ) {
+    val finalInteractionSource = interactionSource ?: remember { MutableInteractionSource() }
```
rock3r: _Final_ seems an odd choice. `actual`? Or do like `clickable` does:

```kotlin
val intSource = interactionSource ?: remember { MutableInteractionSource() }
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/system/SystemInfo.kt
```diff
@@ -0,0 +1,21 @@
...
+
+@InternalJewelApi
+@ApiStatus.Internal
+/**
+ * System info singleton for platform-specific checks.
+ *
+ * **IC Version:**
+ * [SystemInfoRt](https://github.com/JetBrains/intellij-community/blob/master/platform/util-rt/src/com/intellij/openapi/util/SystemInfoRt.java)
+ */
+public object SystemInfo {
```
rock3r: We don't need this, use `org.jetbrains.skiko.hostOs` directly

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SimpleListItem.kt
```diff
@@ -80,6 +81,61 @@ public fun SimpleListItem(
...
+ * @param iconModifier Optional [Modifier] to apply to specifically to the icon.
+ * @param icon Optional [IconKey] representing the icon displayed in the list item.
+ * @param iconContentDescription Optional content description [String] for accessibility purposes for the icon.
+ * @param style Optional [SimpleListItemStyle] for defining the appearance of the list item; default is based on the
+ *   Jewel theme.
+ * @param height The height of the list item; default is based on the Jewel theme's global metrics.
+ * @param colorFilter Optional [ColorFilter] to apply to the icon, if any.
+ * @param painterHints Optional [PainterHint]s to apply to the icon, if any.
+ */
+@Suppress("ComposableParamOrder") // To fix in JEWEL-926
```
rock3r: You're adding a new overload, please fix it for the new one :) I think it's only complaining because the vararg is at the end instead of after the state.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SimpleListItem.kt
```diff
@@ -131,6 +187,57 @@ public fun SimpleListItem(
...
+ * @param textModifier Optional [Modifier] to apply to specifically to the text.
+ * @param iconModifier Optional [Modifier] to apply to specifically to the icon.
+ * @param icon Optional [IconKey] representing the icon displayed on the start side of the list item.
+ * @param iconContentDescription Optional content description [String] for the icon for accessibility purposes.
+ * @param style The [SimpleListItemStyle] defining the appearance of the list item; default is based on the Jewel theme.
+ * @param height The height of the list item; default is based on the Jewel theme's global metrics.
+ * @param colorFilter Optional [ColorFilter] to apply to the icon, if any.
+ * @param painterHints Optional [PainterHint]s to apply to the icon, if any.
+ */
+@Suppress("ComposableParamOrder") // To fix in JEWEL-926
```
rock3r: Ditto

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -0,0 +1,302 @@
...
+                            )
+                        }
+
+                        innerTextField()
+                    }
+                }
+            },
+        )
+
+        LaunchedEffect(Unit) { focusRequester.requestFocus() }
```
rock3r: We can use [`onFirstVisible`](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier#(androidx.compose.ui.Modifier).onFirstVisible(kotlin.Long,kotlin.Float,androidx.compose.ui.layout.LayoutBoundsHolder,kotlin.Function0)) since we're on 1.9 now :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -0,0 +1,302 @@
...
+            AWTKeyEvent.VK_PAGE_DOWN,
+        )
+
+    override fun processKeyEvent(event: KeyEvent): Boolean {
+        val textFieldState = speedSearchState.textFieldState
+
+        return when {
+            event.type != KeyEventType.KeyDown -> false // Only handle key down events
+            event.isMetaPressed -> false // Ignore meta key events
+            (event.isShiftPressed || event.isAltPressed) && event.key.nativeKeyCode in navigationKeys -> false
```
rock3r: Not sure if this covers it, but shift-left or shift-right dismiss the speed search in Swing

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -0,0 +1,302 @@
...
+                    if (selection.start != selection.end) {
+                        replace(selection.min, selection.max, char.toString())
+                    } else {
+                        append(char.toString())
+                    }
+                }
+
+                if (!speedSearchState.isVisible && textFieldState.text.isNotEmpty()) {
+                    speedSearchState.isVisible = true
+                }
```
rock3r: Nit: extract method (maybe for the other multi-line cases, too)

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/system/SystemInfo.kt
faogustavo: TIL

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
faogustavo: updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
faogustavo: It was not covering. Updated it

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
faogustavo: TIL

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SimpleListItem.kt
faogustavo: It kinda feels odd having a vararg in the middle. So I've just added a default value to an empty array. Any thoughts?

rock3r: Vararg in the middle please :) Let's keep consistency with other APIs

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/ExactSubstringSpeedSearchMatcher.kt
```diff
@@ -15,7 +15,7 @@ internal class ExactSubstringSpeedSearchMatcher(private val pattern: String, pri
         if (pattern.isBlank() || text.isNullOrBlank()) return null
         val matchIndex = text.indexOf(pattern, ignoreCase = ignoreCase)
         return if (matchIndex >= 0) {
-            listOf(matchIndex..(matchIndex + pattern.length))
+            listOf(matchIndex until (matchIndex + pattern.length))
```
rock3r: This change seems significant, but it is not mentioned in the PR description (same for the other matcher)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SimpleListItem.kt
rock3r: Or rather — err, if there is no downside to putting a default, I'm actually ok with the vararg. I had not thought of using a default for it.

The thing I am worried about is that by having the default, the signature will be ambiguous if you don't specify `painterHints = ...`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -0,0 +1,333 @@
...
+    styling: SpeedSearchStyle = JewelTheme.speedSearchStyle,
+    textFieldStyle: TextFieldStyle = JewelTheme.textFieldStyle,
+    textStyle: TextStyle = JewelTheme.defaultTextStyle,
+    searchMatchStyle: SearchMatchStyle = JewelTheme.searchMatchStyle,
+    interactionSource: MutableInteractionSource? = null,
+    content: @Composable SpeedSearchScope.() -> Unit,
+) {
+    val finalInteractionSource = interactionSource ?: remember { MutableInteractionSource() }
+    Box(modifier = modifier) {
+        val scope = remember(this, state) { SpeedSearchScopeImpl(this, state, finalInteractionSource) }
```
rock3r: Don't we need to set the interaction scope as key, too?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -0,0 +1,333 @@
...
+}
+
+internal interface SpeedSearchScope : BoxScope {
+    val speedSearchState: SpeedSearchState
+    val interactionSource: MutableInteractionSource
+
+    fun processKeyEvent(event: KeyEvent): Boolean
+}
+
+public interface SpeedSearchItemScope {
```
rock3r: This requires having custom composables, otherwise the `item` API will not have the scope associated with it.

Is there any other way we can do this without having to have e.g., `SpeedSearchLazyColumn`, `SpeedSearchLazyRow`, `SpeedSearchLazyVerticalGrid`, `SpeedSearchLazyHorizontalGrid`, `SpeedSearchLazyVerticalStaggeredGrid`, `SpeedSearchLazyHorizontalStaggeredGrid`, plus all the non-lazy ones (column, row, flex)?

Can we use composition or modifier locals to provide a `HighlightMatchProvider?` that items can access? That would have a `getMatchInfo(item: T)` and a `getMatchInfo(index: Int)` that you can use to get highlighting info for a given item.

In the future we could even consider adding highlighting automatically to `SimpleListItem` when the item text matches items with a highlight in the provider... but not necessary today :) I think a "some effort needed" approach for now is better than going down the rabbit hole of having to create custom layouts for all possible variants. As long as we have samples, it should be ok.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/impl/ExactSubstringSpeedSearchMatcher.kt
faogustavo: This was required when I implemented the render to the highlight. As the feature was still in development, I skipped the comment in the description, but added it now.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
faogustavo: Yes, I think we can. I only discovered the existence of `ModifierLocalProvider` yesterday. I'll try to assemble something :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
faogustavo: Added

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SimpleListItem.kt
faogustavo: I've tested it with this code block, and the signature looks fine IMO:

```
SimpleListItem(
    item.highlightTextSearch(),
    isSelected,
    Modifier.fillMaxWidth().selectable(isSelected) {
        JewelLogger.getInstance("ChipsAndTree").info("Click on $item")
    },
    Modifier.highlightTextSearchArea(textLayoutResult),
    Modifier,
    isActive,
    null,
    null,
    JewelTheme.simpleListItemStyle,
    JewelTheme.globalMetrics.rowHeight,
    null,
    { textLayoutResult = it },
    PainterHint.None,
    PainterHint.None,
    PainterHint.None,
    PainterHint.None,
)
```

Can you elaborate on your concern?

rock3r: It was more a concern that if there are no painter hints specified, there could be a clash between signatures in case there was a variant with a single PainterHint. However, there isn't one :) So all is well

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
```diff
@@ -67,7 +67,15 @@ public interface SelectableColumnOnKeyEvent {
     /** Select Previous Node inherited from Up. */
     public fun onSelectPreviousItem(keys: List<SelectableLazyListKey>, state: SelectableLazyListState) {
         val initialIndex = state.lastActiveItemIndex ?: return
-        for (index in initialIndex - 1 downTo 0) {
+        onSelectPreviousItem(keys, state, initialIndex - 1 downTo 0)
+    }
+
+    public fun onSelectPreviousItem(
```
rock3r, line 77: Can you annotate these two new APIs as internal?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/Highlight.kt
```diff
@@ -0,0 +1,81 @@
...
+    for (match in matchResult.ranges) {
+        // We need to add 1 to the last index to fill the last letter
+        addStyle(SpanStyle(color = matchState.style.colors.foreground), match.first, match.last + 1)
+    }
+}
+
+@Composable
+@ExperimentalJewelApi
+@ApiStatus.Experimental
+public fun Modifier.highlightTextSearchArea(
```
rock3r: The naming seems a bit off, why not `highlightSpeedSearchMatches`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SelectableLazyColumn.kt
```diff
@@ -0,0 +1,306 @@
...
+private fun SelectableLazyListSpeedSearchScrollSideEffect(
+    selectableLazyListState: SelectableLazyListState,
+    speedSearchState: SpeedSearchState,
+    keys: List<Any?>,
+) {
+    LaunchedEffect(Unit) {
+        snapshotFlow { speedSearchState.matchingIndexes }
+            .filter { it.isNotEmpty() }
+            .distinctUntilChanged()
+            .flowOn(Dispatchers.Default)
```
rock3r: This will be problematic for tests... never hardcode dispatchers. It should be provided as a parameter (via the component, with a default) like we do in other places

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableColumnOnKeyEvent.kt
faogustavo: Added

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/Highlight.kt
faogustavo: I'm not sure why I used that name 😅 . Updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SelectableLazyColumn.kt
faogustavo: Added

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
```diff
@@ -0,0 +1,320 @@
...
+
+    LaunchedEffect(state, dispatcher) {
+        val entriesFlow = snapshotFlow { currentStateToList.value.first }
+        speedSearchState.attach(entriesFlow, dispatcher)
+    }
+}
+
+@ExperimentalJewelApi
+@ApiStatus.Experimental
+public class SpeedSearchSelectableLazyColumnKeyActions(
```
rock3r: Forgot to rename this: `SpeedSearchableLazyColumnKeyActions`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
```diff
@@ -0,0 +1,320 @@
...
+import kotlinx.coroutines.flow.onEach
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.foundation.lazy.DefaultMacOsSelectableColumnKeybindings.Companion.isSelectAll
+import org.jetbrains.jewel.foundation.lazy.DefaultMacOsSelectableColumnKeybindings.Companion.isSelectNextItem
+import org.jetbrains.jewel.foundation.lazy.DefaultMacOsSelectableColumnKeybindings.Companion.isSelectPreviousItem
+import org.jetbrains.jewel.foundation.lazy.DefaultSelectableOnKeyEvent.Companion.onSelectAll
+import org.jetbrains.jewel.foundation.lazy.DefaultSelectableOnKeyEvent.Companion.onSelectNextItem
+import org.jetbrains.jewel.foundation.lazy.DefaultSelectableOnKeyEvent.Companion.onSelectPreviousItem
+import org.jetbrains.jewel.foundation.lazy.SelectableLazyColumn as FoundationSelectableLazyColumn
```
rock3r: The alias should not be needed any longer

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+
+        delegate.handleOnKeyEvent(event, keys, state, selectionMode).invoke(this)
+    }
+}
+
+@ExperimentalJewelApi
+@ApiStatus.Experimental
+public interface SpeedSearchSelectableListScope {
```
rock3r: `SpeedSearchableLazyColumnScope`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
```diff
@@ -0,0 +1,320 @@
...
+    public fun stickyHeader(
+        key: Any,
+        textContent: () -> String?,
+        contentType: Any? = null,
+        selectable: Boolean = false,
+        content: @Composable SelectableLazyItemScope.() -> Unit,
+    )
+}
+
+internal class SpeedSearchSelectableListScopeImpl(
```
rock3r: `SpeedSearchableLazyColumnScopeImpl`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
```diff
@@ -0,0 +1,320 @@
...
+        contentType: Any?,
+        selectable: Boolean,
+        content: @Composable (SelectableLazyItemScope.() -> Unit),
+    ) {
+        delegate.stickyHeader(key, contentType, selectable, content)
+    }
+}
+
+@Composable
+private fun SelectableLazyListSpeedSearchScrollSideEffect(
```
rock3r: `SpeedSearchableLazyColumnScrollEffect`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
faogustavo: Updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
faogustavo: Removed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
faogustavo: Updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
faogustavo: Updated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
faogustavo: Updated

rock3r (approved):

rock3r:
Ready to merge

PR #3215, opened 2025-09-03
[JEWEL-1000] Update standalone theme defs, fix generator

rock3r:
This updates the `ThemeDefinition` dumps for standalone for the main branch, fixes the Gradle generator to generate them in the correct folder, makes sure they're properly formatted.

It also adds a step to the release guide to make sure we update them.

Bonus changes:
* Removed unused `ijp.target` property and relative logic from the Gradle build and docs.
* Bumped target IJP to 252.1 in Gradle build
* Bump IJP Gradle Plugin to 2.8.0

> [!NOTE]
> No need to review the generated theme definitions :)


platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntUiThemeDescriptorReader.kt
```diff
@@ -7,22 +7,24 @@ import com.squareup.kotlinpoet.KModifier
...
+        sourcePath: String,
+        ijpMajor: String,
+    ) =
         FileSpec.builder(className)
             .apply {
                 indent("    ")
-                val todayUtc = ZonedDateTime.now().withZoneSameInstant(ZoneOffset.UTC).toLocalDate()
-                addFileComment("Generated by the Jewel Int UI Palette Generator on $todayUtc\n")
+                addFileComment("Generated by the Jewel Int UI Palette Generator for IJP $ijpMajor\n")
+                addFileComment("*** DO NOT EDIT MANUALLY! Use generateThemes task to update. ***\n")
```
rock3r, line 27: This change was made so that the generated file is more stable and does not require re-committing every day :)

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJThemeGeneratorPlugin.kt
```diff
@@ -59,14 +54,39 @@ abstract class IntelliJThemeGeneratorTask : DefaultTask() {
...
+        //  * 241.SNAPSHOT
+        if (buildNumber.isBlank()) return false
+        if (buildNumber.length < 5) return false
+        if (buildNumber.take(3).toIntOrNull()?.takeIf { it > 240 } == null) return false
+        if (buildNumber[3] != '.') return false
+
+        val afterDot = buildNumber.drop(4)
+        return afterDot == "SNAPSHOT" || afterDot.all { it.isDigit() || it == '.' }
+    }
 }
```
faogustavo, line 90: How about using a regex for this? Something like this:

```regex
^(\\d{3})\\.((\\d+\\.?)+|(SNAPSHOT))$
```

The only part it does not validate is the `it > 240`, but we can do with the match result:

```kotlin
/**
 * You can edit, run, and share this code.
 * play.kotlinlang.org
 */
fun main() {
	val regex = "^(\\d{3})\\.((\\d+\\.?)+|(SNAPSHOT))$".toRegex()
    println(regex.find("253.1234.567")?.groups?.get(1)?.value?.toIntOrNull()?.let { it > 240 } == true)
}
```

faogustavo (approved):
No blockers. Happy to see the theme files formatted now xd

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJThemeGeneratorPlugin.kt
rock3r: I thought about it, but it seems less readable to me — and I am a regex lover :D

faogustavo: Yeah, I feel the same way. But we have lots of returns. How about changing to `when` conditions and having only one return statement?

```suggestion
    // Examples:
    //  * 253.1234.567
    //  * 241.SNAPSHOT
    private fun validateBuildNumber(buildNumber: String): Boolean =
        when {
            buildNumber.isBlank() -> false
            buildNumber.length < 5 -> false
            buildNumber.take(3).toIntOrNull()?.takeIf { it > 240 } == null -> false
            buildNumber[3] != '.' -> false
            else -> {
                val afterDot = buildNumber.drop(4)
                afterDot == "SNAPSHOT" || afterDot.all { it.isDigit() || it == '.' }
            }
        }
}
```

rock3r: Using a when this way feels weird, especially the else part... I dunno, I don't love it either 👅 I don't mind having a bunch of return statements if they're all together and very easily readable like in this case. It seems more descriptive to me, if you get what I mean?

rock3r: Anyway, if you feel strongly about this, I'd rather do the regex at that point 👯

faogustavo: All options work for me. We can keep it as it is. Just trying to help xd

rock3r: Given how unimportant this is in the grand scheme of things, I'll keep what's there and cut our losses :D

rock3r:
Ready to merge

PR #3216, opened 2025-09-03
[JEWEL-1002] Revisit usages of instanceUuid early in theme init

rock3r:
We have two places where we're building defaults in standalone, where we use instanceUuid, but that's not great since the composition local is only set up much later. We should use isDark in those cases. We can also remove the test theme.

### Bug fixes
 * Fixed an issue that required explicitly providing a theme `instanceUuid` in tests


DanielSouzaBertoldi (approved):
Quickest fix in the wild west 🤠 may you rest in peace `IntUiTestTheme`

faogustavo (approved):

rock3r:
Ready to merge

PR #3217, opened 2025-09-03
[JEWEL] IJPL-176416 Compose icon infrastructure

jakub-senohrabek:


.idea/runConfigurations/IDEA.xml
```diff
@@ -7,7 +7,7 @@
...
...D --add-opens=java.desktop/javax.swing.text=ALL-UNNAMED --add-opens=java.desktop/javax.swing.text.html=ALL-UNNAMED --add-opens=java.desktop/javax.swing.text.html.parser=ALL-UNNAMED --add-opens=java.desktop/javax.swing=ALL-UNNAMED --add-opens=java.desktop/sun.awt.X11=ALL-UNNAMED --add-opens=java.desktop/sun.awt.datatransfer=ALL-UNNAMED --add-opens=java.desktop/sun.awt.image=ALL-UNNAMED --add-opens=java.desktop/sun.awt.windows=ALL-UNNAMED --add-opens=java.desktop/sun.awt=ALL-UNNAMED --add-opens=java.desktop/sun.font=ALL-UNNAMED --add-opens=java.desktop/sun.java2d=ALL-UNNAMED --add-opens=java.desktop/sun.lwawt.macosx=ALL-UNNAMED --add-opens=java.desktop/sun.lwawt=ALL-UNNAMED --add-opens=java.desktop/sun.swing=ALL-UNNAMED --add-opens=jdk.attach/sun.tools.attach=ALL-UNNAMED --add-opens=jdk.internal.jvmstat/sun.jvmstat.monitor=ALL-UNNAMED --add-opens=jdk.jdi/com.sun.tools.jdi=ALL-UNNAMED --add-opens=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED -Didea.platform.prefix=Idea -Dadditional.modules=&quot;intellij.devkit&quot; -Djava.system.class.loader=com.intellij.util.lang.PathClassLoader -Djavax.xml.parsers.SAXParserFactory=com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl" />
```
rock3r, line 10: Hard to tell what changed here

.idea/markdown.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rock3r, line 1: Did you mean to commit this?

platform/core-ui/src/ui/icons/IconWithOverlay.java
```diff
@@ -17,7 +18,7 @@
  * @author Konstantin Bulenkov
  */
 public abstract class IconWithOverlay extends LayeredIcon {
-  public IconWithOverlay(@NotNull Icon main, @NotNull Icon overlay) {
+  public IconWithOverlay(@NotNull javax.swing.Icon main, @NotNull javax.swing.Icon overlay) {
```
rock3r, line 21: I'd revert changes to this file

platform/core-ui/src/ui/icons/IconWrapperWithToolTipComposite.java
```diff
@@ -6,6 +6,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.Icon;
```
rock3r, line 9: Ditto

platform/core-ui/src/util/IconUtil.kt
```diff
@@ -42,6 +42,7 @@ import java.util.*
 import java.util.concurrent.ConcurrentHashMap
 import java.util.function.Supplier
 import java.util.function.ToIntFunction
+import javax.swing.Icon
```
rock3r, line 45: Ditto

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/DynamicIcon.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+import kotlinx.coroutines.flow.StateFlow
+
+interface DynamicIcon : Icon {
+  val onUpdate: StateFlow<IconState>
+}
```
rock3r, line 8: Nit: missing empty line at EOF — will not repeat for all files, but there are several

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/DynamicIcon.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+import kotlinx.coroutines.flow.StateFlow
+
+interface DynamicIcon : Icon {
+  val onUpdate: StateFlow<IconState>
```
rock3r, line 7: With the current impl of `IconState`, this would likely cause missed redraws due to `StateFlow`'s implicit `distinct()` call. This should be a `Flow<IconState>` imo

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/IconManager.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+interface IconManager {
+  fun loadIcon(id: IconIdentifier, path: String, aClass: Class<*>): Icon?
+  fun registerIcon(id: IconIdentifier, icon: Icon)
```
rock3r, line 6: Nit: for consistency

```suggestion
  fun loadIcon(identifier: IconIdentifier, path: String, aClass: Class<*>): Icon?
  fun registerIcon(identifier: IconIdentifier, icon: Icon)
```

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+import kotlin.reflect.KClass
+
+interface ImageResourceProvider {
+  companion object {
+    @Volatile
```
rock3r, line 8: Do we need this to be volatile? It may become clearer later why we do, but it is suspicious that we need this for now

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+    private var instance: ImageResourceProvider? = null
+
+    @JvmStatic
+    fun getInstance(): ImageResourceProvider = instance ?: DummyImageResourceProvider
+
+    fun activate(provider: ImageResourceProvider) {
```
rock3r, line 14: Should this be `internal`? Doesn't look like something we want anyone to be able to do at any time

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+interface BitmapImageResource : ImageResource {
+  fun getRGB(x: Int, y: Int): Int
+  fun getRGBOrNull(x: Int, y: Int): Int?
+  val width: Int
+  val height: Int
+}
+
+object EmptyBitmapImageResource : BitmapImageResource {
+  override fun getRGB(x: Int, y: Int): Int = 0
+  override fun getRGBOrNull(x: Int, y: Int): Int? = 0
```
rock3r, line 46: Shouldn't this return `null`?

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+  override val height: Int = 0
+}
+
+interface RescalableImageResource : ImageResource {
+  fun scale(scale: ImageScale): BitmapImageResource
+  fun calculateExpectedDimensions(scale: ImageScale): Bounds
+}
+
+sealed interface ImageScale {
+  fun calculateScalingFactorByOriginalDimensions(width: Int, height: Int? = null): Float
```
rock3r, line 57: Not a fan of the nullable `height`. I'd rather pass in a default that's equal to the `width`, or have separate functions to calculate the scaling by height, width, or square (smallest dimension)

```suggestion
  fun calculateScalingFactorByOriginalDimensions(width: Int, height: Int = width): Float
```

I'm also not sure what `ByOriginalDimensions` means

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/PaintingApi.kt
```diff
@@ -0,0 +1,37 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+interface PaintingApi {
+  val bounds: Bounds
+  fun drawImage(image: BitmapImageResource, x: Int, y: Int, width: Int? = null, height: Int? = null)
+  fun drawImage(image: RescalableImageResource, x: Int, y: Int, width: Int? = null, height: Int? = null)
```
rock3r, line 7: Ditto here — can we use non-nullable values for width and height? Using the resource's intrinsic size as defaults looks better to me.

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    fun activate(provider: ImageResourceProvider) {
+      instance = provider
+    }
+  }
+}
+
+object DummyImageResourceProvider : ImageResourceProvider
+
+interface ImageResource {
```
rock3r, line 22: I think this should expose an intrinsic size

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/PaintingApi.kt
```diff
@@ -0,0 +1,37 @@
...
+}
+
+class Bounds(
```
rock3r, line 10: Are we not using an AWT `Dimension` to avoid confusion with IJP's own scaling?

platform/icons-api/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+private val composeVersion
+    get() = ComposeBuildConfig.composeVersion
+
+dependencies {
+    api("org.jetbrains.compose.foundation:foundation-desktop:$composeVersion")
+
+    testImplementation(compose.desktop.currentOs) { exclude(group = "org.jetbrains.compose.material") }
+}
```
rock3r, line 18: Doesn't look like we need these. The iml only uses coroutines and kotlin stdlib (which in Gradle you can probably mark as `compileOnly` as they'll be put on the classpath by Jewel)

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/BridgeJewelIconProvider.kt
```diff
@@ -0,0 +1,35 @@
...
+        val painter = remember(icon, iconState) {
+            BridgeIconPainter(icon)
+        }
+        return painter
+    }
+}
+
+private class BridgeIconPainter(val icon: Icon) : Painter() {
+    override val intrinsicSize: Size
+        get() = Size(32f, 32f)
```
rock3r, line 29: This is definitely wrong, this should come from the `Icon`'s intrinsic size.

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/PaintingApi.kt
```diff
@@ -0,0 +1,37 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+interface PaintingApi {
```
rock3r, line 4: Is this supposed to be an abstraction over `Canvas`? If so, we may need a bunch more primitives; this only makes it possible to draw a resource image, but does not cover things like icons drawn at runtime.

I'd also consider calling this something like `IconPaintingScope` given it's supposed to be used like a DSL

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/ComposeBitmapImageResource.kt
```diff
@@ -0,0 +1,100 @@
...
+            skia
+        )
+        lastBitmap = new
+        return new.composeBitmap
+    }
+}
+
+private fun BitmapImageResource.composeBitmapWithoutCaching(): Pair<Bitmap, ImageBitmap> {
+    val bitmap = Bitmap()
+    bitmap.allocPixels(ImageInfo.makeS32(width, height, ColorAlphaType.UNPREMUL))
```
rock3r, line 75: This is assuming both that the image is 32 bpp, and that it is sRGB. It will be true in most cases, but can cause issues. We should read the colour depth and colour space from the resource itself, not assume it, or we'll end up with wrong colours.

Besides, [this Skia API](https://api.skia.org/structSkImageInfo.html#abdcd4db3946067772663f96b7bd3c310) uses the "native" 32bpp pixel format, which as the documentation says, can lead to inconsistent results across platforms:

Image

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+inline fun <reified TBitmap : Any> CrossApiImageBitmapCache.cachedBitmap(noinline generator: () -> TBitmap): TBitmap =
+  cachedBitmap(TBitmap::class, generator)
+
+interface CrossApiImageBitmapCache {
+  fun <TBitmap : Any> cachedBitmap(bitmapClass: KClass<TBitmap>, generator: () -> TBitmap): TBitmap
+}
+
+interface BitmapImageResource : ImageResource {
+  fun getRGB(x: Int, y: Int): Int
+  fun getRGBOrNull(x: Int, y: Int): Int?
```
rock3r, line 39: Do these return RGB? Or ARGB? Or RGBA? It's unclear from the name. Usages seem to suggest ARGB, but... can you clarify?

Also, why would we need `orNull`? I don't think we should allow anyone to try to read out of the bounds of the image. It makes no sense.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/ComposeBitmapImageResource.kt
```diff
@@ -0,0 +1,100 @@
...
+    bitmap.setPixelsFrom(this)
+    return bitmap to bitmap.asComposeImageBitmap()
+}
+
+private fun Bitmap.setPixelsFrom(image: BitmapImageResource) {
```
rock3r, line 80: This is really expensive, although inevitable — @igordmn mentioned at some point it should be possible to handle [BGRA bitmaps directly](https://api.skia.org/SkColorType_8h.html#a3ea8ab47612dbc4801c18445176c6481) in Skia, so if that is possible, we can spare ourselves the byte manipulation here.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/ComposeBitmapImageResource.kt
```diff
@@ -0,0 +1,100 @@
...
+            dimensions,
+            compose,
+            skia
+        )
+        lastBitmap = new
+        return new.composeBitmap
+    }
+}
+
+private fun BitmapImageResource.composeBitmapWithoutCaching(): Pair<Bitmap, ImageBitmap> {
```
rock3r, line 73: Can we make this return a private data class or something? It's hard to follow when we access `first` and `second`

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+
+interface ImageResourceWithCrossApiCache {
+  val crossApiCache: CrossApiImageBitmapCache
+}
+
+inline fun <reified TBitmap : Any> CrossApiImageBitmapCache.cachedBitmap(noinline generator: () -> TBitmap): TBitmap =
+  cachedBitmap(TBitmap::class, generator)
+
+interface CrossApiImageBitmapCache {
+  fun <TBitmap : Any> cachedBitmap(bitmapClass: KClass<TBitmap>, generator: () -> TBitmap): TBitmap
```
rock3r, line 34: This name has me really confused. Is it a `getOrPut`?

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+}
+
+interface ImageResourceWithCrossApiCache {
+  val crossApiCache: CrossApiImageBitmapCache
+}
+
+inline fun <reified TBitmap : Any> CrossApiImageBitmapCache.cachedBitmap(noinline generator: () -> TBitmap): TBitmap =
+  cachedBitmap(TBitmap::class, generator)
+
+interface CrossApiImageBitmapCache {
```
rock3r, line 33: Considering there are no actual cross-API implementations (there is a Compose one, and there is an AWT one, but neither is cross-API), I'm not sure we actually really need this... It feels like it could be replaced by a `Map<KClass<TBitmap>, TBitmap>`, or even by a `var cache: TBitmap`

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/ImageResource.kt
```diff
@@ -0,0 +1,103 @@
...
+}
+
+object EmptyBitmapImageResource : BitmapImageResource {
+  override fun getRGB(x: Int, y: Int): Int = 0
+  override fun getRGBOrNull(x: Int, y: Int): Int? = 0
+  override val width: Int = 0
+  override val height: Int = 0
+}
+
+interface RescalableImageResource : ImageResource {
```
rock3r, line 51: Does this signify that an image can be scaled losslessly? And that by contrast other `ImageResources` will have lossy scaling?

I'm asking because that's a major gripe I have with the current IJP Swing icon APIs — all images are inherently scalable, the question is whether that's a lossless or lossy scaling. I would prefer being more precise/explicit in this new API if we can.

I would imagine that if my interpretation is correct, then vector-based and runtime-generated icons are inherently losslessly scalable, whereas bitmaps are not.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/ComposePaintingApi.kt
```diff
@@ -0,0 +1,58 @@
...
+      x: Int,
+      y: Int,
+      width: Int?,
+      height: Int?,
+    ) {
+        drawComposeImage(image.composeBitmap(), x, y, width, height)
+    }
+
+    override fun drawImage(image: RescalableImageResource, x: Int, y: Int, width: Int?, height: Int?) {
+        drawComposeImage(image.composeBitmap(FitAreaScale(bounds.width, bounds.height)), x, y, width, height)
```
rock3r, line 36: If we always use the fit scaling, why do we have the fixed scale option, too? Also, in case we wanted to implement something like a composite icon, wouldn't doing this cause all icons to be as large as the canvas, or at least have a needless scale applied?

We're saying, if this is a scalable image, then get the cached version that's as big as the canvas bounds, but then draw it using the width and height provided (which can be null, or larger, or smaller, than the implicit size of the image).

I think this is fundamentally wrong; the intrinsic dimensions should be the default — not nulls! — and the provided width and height must be used for the cache access, too!

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/ComposeBitmapImageResource.kt
```diff
@@ -0,0 +1,100 @@
...
+    val cached = cache?.cachedBitmap<SingleBitmapCache> { SingleBitmapCache() }
+    if (cached != null) {
+        return cached.getOrPut(this, scale)
+    } else {
+        return scale(scale).composeBitmap()
+    }
+}
+
+private class SingleBitmapCache {
+    private var lastBitmap: CachedBitmap? = null
```
rock3r, line 36: If we draw an icon in multiple sizes, this will cache nothing, will it?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/ComposePaintingApi.kt
```diff
@@ -0,0 +1,58 @@
...
+        y: Int,
+        width: Int?,
+        height: Int?,
+    ) {
+        val targetSize = IntSize(width ?: image.width, height ?: image.height)
+        if (targetSize.width == 0 || targetSize.height == 0) return
+        drawScope.drawImage(
+            image,
+            IntOffset(x, y),
+            targetSize,
```
rock3r, line 51: This only works if the targetSize == image size, otherwise you're drawing a subset of the image.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/icon/ComposePaintingApi.kt
```diff
@@ -0,0 +1,58 @@
...
+            dstSize = targetSize,
+            alpha = 1.0f,
+            colorFilter = null,
+            filterQuality = FilterQuality.Low
```
rock3r, line 55: Why low? If this is not a losslessly resizable image, you definitely want high quality filtering

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Icon.kt
```diff
@@ -81,6 +86,18 @@ public fun Icon(
  * @param hint [PainterHint] to be passed to the painter.
  */
 @Suppress("ComposableParamOrder") // To fix in JEWEL-929
```
rock3r, line 88: This annotation should also be left on the existing `Icon`

platform/jewel/ui/intellij.platform.jewel.ui.iml
```diff
@@ -74,6 +74,7 @@
       </library>
     </orderEntry>
     <orderEntry type="module" module-name="intellij.libraries.compose.foundation.desktop" exported="" />
+    <orderEntry type="module" module-name="intellij.platform.icons.api" exported="" />
```
rock3r, line 77: This means that module also needs to be published to Maven Central

platform/util/src/com/intellij/ui/icons/CompositeIcon.java
```diff
@@ -3,6 +3,7 @@
 
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.Icon;
```
rock3r, line 6: Revert?

platform/util/src/com/intellij/ui/icons/DarkIconProvider.java
```diff
@@ -3,6 +3,7 @@
 
 import org.jetbrains.annotations.NotNull;
 
+import javax.swing.Icon;
```
rock3r, line 6: Ditto

platform/util/src/com/intellij/ui/icons/IconReplacer.java
```diff
@@ -3,6 +3,7 @@
 
 import org.jetbrains.annotations.Contract;
 
+import javax.swing.Icon;
```
rock3r, line 6: Ditto

platform/util/src/com/intellij/ui/icons/ReplaceableIcon.java
```diff
@@ -4,7 +4,7 @@
 import com.intellij.openapi.diagnostic.Logger;
 import org.jetbrains.annotations.NotNull;
 
-import javax.swing.*;
+import javax.swing.Icon;
```
rock3r, line 7: Revert?

platform/util/src/com/intellij/ui/icons/RowIcon.java
```diff
@@ -4,7 +4,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Unmodifiable;
 
-import javax.swing.*;
```
rock3r: Ditto

platform/util/ui/src/com/intellij/ui/icons/AwtImageResourceImpl.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.jetbrains.icons.api.CrossApiImageBitmapCache
+import org.jetbrains.icons.api.ImageResourceWithCrossApiCache
+import java.awt.Image
+import java.awt.image.BufferedImage
+
+class AwtImageResourceImpl(override val image: Image) : AwtImageResource, ImageResourceWithCrossApiCache {
+  override fun getRGB(x: Int, y: Int): Int {
+    if (image is BufferedImage) {
+      return image.getRGB(x, y)
+    } else error("Reading RGB values from non buffered image is not supported yet.")
```
rock3r, line 13: Should we just restrict all these APIs from taking `Image` to taking `BufferedImage` instead?

platform/util/ui/src/com/intellij/ui/icons/AwtImageResourceImpl.kt
```diff
@@ -0,0 +1,35 @@
...
+  override val height: Int = image.getHeight(null)
+
+  override val crossApiCache: CrossApiImageBitmapCache = CrossApiImageBitmapCacheImpl()
+
+  init {
+    crossApiCache.cachedBitmap(Image::class) { image }
+  }
+}
+
+class AwtImageResourceProviderImpl : AwtImageResourceProvider {
```
rock3r, line 31: Is this meant to be internal?

platform/util/ui/src/com/intellij/ui/icons/CachedImageIcon.kt
```diff
@@ -58,6 +69,46 @@ internal val pathTransform: AtomicReference<IconTransform> = AtomicReference(
...
+
+  override fun hashCode(): Int {
+    var result = imageFlags
+    result = 31 * result + flags
+    return result
+  }
+}
+
+class CachedImageIconIdentifier(
+  val coords: Pair<String, ClassLoader>?
```
rock3r, line 96: Why not two fields?

platform/util/ui/src/com/intellij/ui/icons/CachedImageIcon.kt
```diff
@@ -115,7 +171,60 @@ open class CachedImageIcon private constructor(
...
+        val rawImage = oldIcon.resolveImage(scaleContext = scaleContext)
+        val awtImage = (rawImage as? JBHiDPIScaledImage)?.delegate ?: rawImage
+        if (awtImage == null) { return EmptyBitmapImageResource }
+        val imageResource = ImageResource.fromAwtImage(awtImage)
+        cache = BitmapCache(imageResource, scale, oldIcon.state)
+      }
+      return cache!!.image
+    }
+
+    private class BitmapCache(
```
rock3r, line 217: No equals/hashcode?

platform/util/ui/src/com/intellij/ui/icons/CopyableIcon.java
```diff
@@ -3,6 +3,7 @@
 
 import org.jetbrains.annotations.NotNull;
 
+import javax.swing.Icon;
```
rock3r, line 6: Revert?

platform/util/ui/src/com/intellij/ui/icons/CrossApiImageBitmapCache.kt
```diff
@@ -0,0 +1,18 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.ui.icons
+
+import org.jetbrains.icons.api.CrossApiImageBitmapCache
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+import kotlin.reflect.KClass
+
+class CrossApiImageBitmapCacheImpl : CrossApiImageBitmapCache {
```
rock3r, line 9: Should this be internal?

platform/util/ui/src/com/intellij/ui/icons/IconWithToolTip.java
```diff
@@ -3,7 +3,7 @@
 
 import com.intellij.openapi.util.NlsSafe;
 import org.jetbrains.annotations.Nullable;
-
+import javax.swing.Icon;
```
rock3r, line 6: Ditto

platform/util/ui/src/com/intellij/ui/icons/IntelliJIconManager.kt
```diff
@@ -0,0 +1,19 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.ui.icons
+
+import org.jetbrains.icons.api.Icon
+import org.jetbrains.icons.api.IconIdentifier
+
+class IntelliJIconManagerImpl: IntelliJIconManager {
```
rock3r, line 7: Is this meant to be empty?

platform/util/ui/src/com/intellij/ui/icons/LazyImageIcon.java
```diff
@@ -9,6 +9,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.Icon;
```
rock3r, line 12: Revert?

platform/util/ui/src/com/intellij/ui/icons/MenuBarIconProvider.java
```diff
@@ -3,6 +3,7 @@
 
 import org.jetbrains.annotations.NotNull;
 
+import javax.swing.Icon;
```
rock3r, line 6: Revert?

platform/util/ui/src/com/intellij/ui/icons/CopyableIcon.java
rock3r, line 6: Revert?

platform/util/ui/src/com/intellij/ui/icons/SwingIconImpl.kt
```diff
@@ -0,0 +1,25 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.ui.icons
+
+import org.jetbrains.icons.api.PaintingApi
+import java.awt.Component
+import java.awt.Graphics
+import javax.swing.Icon
+
+abstract class SwingIconImpl : SwingIcon {
```
rock3r, line 9: Is this meant to be internal?

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/IconManager.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+interface IconManager {
```
rock3r, line 4: This looks like it's only extended in the IJP by another interface, and in turn has an empty implementation. Do we need it?

platform/util/ui/src/com/intellij/ui/icons/SwingPaintingApi.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+  override fun drawImage(image: RescalableImageResource, x: Int, y: Int, width: Int?, height: Int?) {
+    val swingImage = image.scale(FitAreaScale(width ?: bounds.width, height ?: bounds.height))
+    drawImage(swingImage, x, y, width, height)
+  }
+}
+
+fun PaintingApi.swing(): SwingPaintingApi? = this as? SwingPaintingApi
```
rock3r, line 41: This makes the API look very different from the one on the Compose side in usages. Even more confusingly, there is no benefit to doing this case as there are no generics nor extra methods exposed by `SwingPaintingApi` over `PaintingApi`

platform/util/ui/src/com/intellij/ui/icons/UpdatableIcon.java
```diff
@@ -4,6 +4,7 @@
 import com.intellij.openapi.util.ModificationTracker;
 import org.jetbrains.annotations.NotNull;
 
+import javax.swing.Icon;
```
rock3r, line 7: Revert?

platform/util/src/com/intellij/ui/icons/AwtImageResource.kt
```diff
@@ -0,0 +1,19 @@
...
+  val image: java.awt.Image
+}
+
+interface AwtImageResourceProvider : ImageResourceProvider {
+  fun fromAwtImage(image: java.awt.Image): AwtImageResource
+}
+
+fun ImageResource.Companion.fromAwtImage(image: java.awt.Image): AwtImageResource {
+  val provider = ImageResourceProvider.getInstance() as? AwtImageResourceProvider
+  return provider?.fromAwtImage(image) ?: error("Current ImageResource provider doesn't support AWT images: $provider")
```
rock3r, line 18: This could also throw because the provider has not been initialized yet...

platform/util/ui/src/com/intellij/ui/icons/SwingPaintingApi.kt
```diff
@@ -0,0 +1,41 @@
...
+  override val bounds: Bounds get() {
+    if (c == null) return Bounds(0, 0)
+    return Bounds(
+      c.width,
+      c.height
+    )
+  }
+
+  override fun drawImage(image: BitmapImageResource, x: Int, y: Int, width: Int?, height: Int?) {
+    val swingImage = image as? AwtImageResource ?: error("Image resource must be AwtImageResource to be rendered using swing.")
```
rock3r, line 27: This tells me we need a generic type parameter on `PaintingApi`, or a better API overall. This is not good design as users have no idea they're not supposed to pass in whatever `BitmapImageResource`.

In general I am a bit confused as to what the benefit is to this whole approach, since icons do not share any memory between Swing and Compose (mostly understandable), but more pressingly, the pipelines remain completely distinct between the two, to the point where indeed each implementation only accepts the one resource type it likes. It's not much different from the current implementation at that point, I think

platform/util/ui/src/com/intellij/ui/RetrievableIcon.java
```diff
@@ -5,7 +5,7 @@
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
-import javax.swing.*;
+import javax.swing.Icon;
```
rock3r, line 8: Revert?

platform/util/ui/src/com/intellij/ui/SizedIcon.java
```diff
@@ -10,7 +10,7 @@
 import com.intellij.util.ui.JBCachingScalableIcon;
 import org.jetbrains.annotations.NotNull;
 
-import javax.swing.*;
+import javax.swing.Icon;
```
rock3r, line 13: Ditto

platform/util/ui/src/com/intellij/util/ui/EmptyIcon.java
```diff
@@ -158,3 +177,25 @@ public static final class EmptyIconUIResource extends EmptyIcon implements UIRes
     }
   }
 }
+
+class EmptyIconIdentifier implements IconIdentifier {
```
rock3r, line 181: Private?

plugins/devkit/intellij.devkit.compose/src/demo/ComponentShowcaseTab.kt
```diff
@@ -284,7 +288,7 @@ private fun IconsShowcase() {
     }
 
     IconButton(onClick = {}, Modifier.size(24.dp)) {
-      Icon(key = AllIconsKeys.Actions.Close, contentDescription = "Close")
+      Icon(AllIconsKeys.Actions.Close, "Close")
```
rock3r, line 291: Revert?

plugins/devkit/intellij.devkit.compose/src/demo/ComponentShowcaseTab.kt
```diff
@@ -295,9 +299,25 @@ private fun IconsShowcase() {
       hints = arrayOf(Size(24)),
       tooltip = { Text("Hello there") },
     )
+
+    Box {
```
rock3r, line 303: No need for these boxes

plugins/devkit/intellij.devkit.compose/src/demo/ComponentShowcaseTab.kt
```diff
@@ -295,9 +299,25 @@ private fun IconsShowcase() {
...
+      Icon(AllIcons.General.OpenDisk as Icon, "Build Load Changes")
```
rock3r, line 304: This cast looks very weird, why is it needed?

plugins/devkit/intellij.devkit.compose/src/demo/ComponentShowcaseTab.kt
```diff
@@ -295,9 +299,25 @@ private fun IconsShowcase() {
...
+    }
+
+    Box {
+      Icon(deferedIcon as Icon, "Deferred Icon Sample")
+    }
   }
 }
 
+private val deferedIcon = IconDeferrer.getInstance().deferAsync(
```
rock3r, line 313: Typo: `deferredIcon`

plugins/devkit/intellij.devkit.compose/src/showcase/ComposePerformanceDemoAction.kt
```diff
@@ -129,6 +145,111 @@ private class MyDialog(project: Project?, dialogTitle: String) :
...
 }
 
+private fun createIconsComponent(): JComponent {
+  return compose {
+    var minFps by remember { mutableStateOf(Int.MAX_VALUE) }
+    var maxFps by remember { mutableStateOf(Int.MIN_VALUE) }
+    val frameTimes = remember { LinkedList<Long>() }
+    SideEffect {
+      frameTimes.add(System.nanoTime())
+      frameTimes.removeAll { it < System.nanoTime() - 1_000_000_000 }
```
rock3r, line 155: This is potentially quite expensive as it requires looping _all_ `frameTimes`, on the EDT. It would be much cheaper to check the length of the linked list and drop the oldest (earilest) `n` items to fit into the measurements cap (you probably only need ~1-2 seconds' worth of frames)

plugins/devkit/intellij.devkit.compose/src/showcase/ComposePerformanceDemoAction.kt
```diff
@@ -129,6 +145,111 @@ private class MyDialog(project: Project?, dialogTitle: String) :
...
+
+    val transition = rememberInfiniteTransition("coso")
+    val iconSize by transition.animateFloat(
+      30f,
+      50f,
+      infiniteRepeatable(tween(durationMillis = 1000, easing = EaseInOut), repeatMode = RepeatMode.Reverse),
+    )
+
+    Column {
+      val fps = frameTimes.size
```
rock3r, line 166: This fps counter is completely wrong. It's using the `size` of the linked list as fps value. That makes no sense; the fps is the number of frames rendered in the last second — or in this case, _recomposed_ in the last second.

Likewise, the min and max fps values below make no sense. minFps will always be `0` as the linked list starts empty, and maxFps is going to grow on each recomposition.

plugins/devkit/intellij.devkit.compose/src/showcase/ComposeShowcase.kt
```diff
@@ -128,6 +133,27 @@ private fun InfiniteAnimation() {
   Box(Modifier.alpha(animatedAlpha)) {
     Text("Animation!")
   }
+
+  val iconSize by transition.animateFloat(
+    0f,
+    250f,
+    infiniteRepeatable(tween(durationMillis = 1000, easing = EaseInOut), repeatMode = RepeatMode.Reverse),
+  )
+  Box {
```
rock3r, line 142: No need for these boxes

plugins/devkit/intellij.devkit.compose/src/showcase/ComposeShowcase.kt
```diff
@@ -128,6 +133,27 @@ private fun InfiniteAnimation() {
...
+    Icon(AllIcons.General.OpenDisk as Icon, "Build Load Changes", modifier = Modifier.size(iconSize.dp))
+  }
+
+  Box {
+    Icon(deferedIcon as Icon, "Deferred Icon Sample")
+  }
+}
+
+private val deferedIcon = IconDeferrer.getInstance().deferAsync(
```
rock3r, line 151: Same typo: `deferredIcon`

.idea/runConfigurations/IDEA.xml
rock3r: The IDE shows it in the diff 🎉

`-Dadditional.modules=&quot;intellij.devkit&quot;` was added

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/icon/IconPainterProvider.kt
```diff
@@ -0,0 +1,18 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.icon
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.ui.graphics.painter.Painter
+import org.jetbrains.icons.api.Icon
+
+public interface IconPainterProvider {
```
rock3r, line 10: No implementation in Standalone?!?

platform/util/ui/src/com/intellij/ui/icons/FallbackIconIdentifier.kt
```diff
@@ -0,0 +1,21 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.ui.icons
+
+import org.jetbrains.icons.api.IconIdentifier
+
+class FallbackIconIdentifier(
```
rock3r, line 6: Is this fallback or nested?

rock3r (changes requested):
I have a bunch of notes, but at a high level, I don't think this design actually achieves the goals of interoperability and unifying processing between Swing and Compose. It's also missing a Compose standalone implementation, which is where things probably will break down even further.

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/IconManager.kt
```diff
@@ -0,0 +1,9 @@
...
+  fun loadIcon(id: IconIdentifier, path: String, aClass: Class<*>): Icon?
+  fun registerIcon(id: IconIdentifier, icon: Icon)
+}
+
+interface IconIdentifier
```
chashnikov, line 9: One of the things we can solve with new icons API is getting rid of dependency on javax.swing from *.psi modules. It seems that `javax.swing.Icon` is the only class from javax.swing which is used there (`PsiElement` interface extends `Iconable` with `getIcon` method). So if we introduce `IconIdentifier`, we can use it in `PsiElement` instead of `javax.swing.Icon` and eventually get rid of the latter.

However, to do that we probably need to put `IconIdentifier` in a different module, so it can be added to dependencies of *.psi modules without having any actual rendering API there.

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/IconManager.kt
chashnikov, line 6: `registerIcon` method is supposed to be used to register all available icons in advance, right? I think it's better to avoid such a requirement and allow locating icons by demand only.

platform/icons-api/src/main/kotlin/org/jetbrains/icons/api/Icon.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.api
+
+interface Icon {
+  val identifier: IconIdentifier
+  val state: IconState
+
+  fun render(api: PaintingApi)
```
chashnikov, line 8: Do we really need to have fully abstract painting api?
It seems (please correct me if I'm wrong) the vast majority of icons go through the following stages:
1. An element (`PsiElement`, action, tool window) determines how it should be presented. Often it just returns a field from one of *Icons classes which wraps a path to the svg file. For `PsiElement` the presentation may be composed of multiple layers. Sometimes calculating the proper presentation requires some time, so it shouldn't be done synchronously.
2. The presentation is adjusted according to device-specific configuration: light or dark theme, scaling.
3. The resulting image is actually painted.

The problem with the current implementation is that the same interface `javax.swing.Icon` is used during all these stages.
And code which determines presentation during the first stage shouldn't know specify how the icon should be rendered. 
We would need this if we want to have absolute flexibility here and allow elements to generate bitmaps on the fly, for example. But do we really need it? If in 99% of cases the icons are svg files with few transformations applied, painting them all via such a generic API may be not efficient.

PR #3221, opened 2025-09-06
Support fixture overriding with direct test parametrization

chbndrhnns:
fixes https://youtrack.jetbrains.com/issue/PY-30070/pytest-fixture-overriding-triggers-false-positive-following-arguments-are-not-declared-when-used-with-parametrization


EliseevEgor:
Thank you for your PR. Merged to the master branch

PR #3222, opened 2025-09-06
Suggest argument names for parametrized classes

chbndrhnns:
Fixes: https://youtrack.jetbrains.com/issue/PY-56149/On-completion-suggest-argument-names-for-pytest-parametrized-test-classes


avokin2:
Merged. Thank you for your contribution

PR #3223, opened 2025-09-08
ENH: Option to apply type renderers to subclases fo the target type

dangreb:
While tending to an issue regarding Type Renderers on PyCharm, I thought it would be nice if we had an option to apply certain customized renderers to subclasses of the indiccated Target Type. It's been ages since i dabbled with Java like code so i went ahead and did it!

image

I also did some cosmetics for the customizing GUI. The new checkbox labeled "Apply to subclasses" makes so that instances of any subclasses of the indicated Target Type will receive the very same treatment from the IDE, in regards to that singular renderer of course.

I need to test it further still, so i'll open this draft for now, also so we can ponder over its relevance and viability of course!

cheers!


dangreb:
After several tests, adjustments and improvements, it seems reasonable to promote this for reviewing. Please feel free to do so at your convenience! 

Some important facts:

- A rule was included to ignore subclass matching when builtin types are set as the Target Type, even when the option is selected. This should grant safety and discard an arguably unecessary scenario.
- Target Type resolving will take place during construction of UserTypeRenderer instances, for better architecture and availability, as well as potentialy less repetition

Also:

- Fixed validation mechanics for the Target Type input field.
- Added context helper tooltip for further clarity

cheers!

image

PR #3224, opened 2025-09-08
Ignore test-like methods for gutter icon and run config that pytest w…

chbndrhnns:
…ould not collect

Fixes https://youtrack.jetbrains.com/issue/PY-43704/Pycharm-thinks-my-test-method-is-a-PyTest-test-and-wont-run-it-normally...-Its-just-a-method?backToIssues=false


EliseevEgor:
Thank you for your PR. Merged to the master branch

PR #3226, opened 2025-09-10
Allow asyncio.create_task() to be called without await

chbndrhnns:
Fixes https://youtrack.jetbrains.com/issue/PY-84027/False-positive-Coroutine-createtask-is-not-awaited

Depending how the `Task` type actually resolves, Pycharm might see different qualified names.

In CPython, the concrete implementation of Task lives in the C accelerator module _asyncio (a binary module). The pure-Python asyncio.tasks.Task is effectively an alias to that implementation. PyCharm builds “binary skeletons” for extension modules. When type inference resolves the class coming from your interpreter’s standard library, it often prefers those skeletons. As a result, the qualified name for Task is taken from the binary module where it’s actually defined: _asyncio.Task. Typeshed stubs reflect this too (depending on Python version). They define/alias Task so that asyncio.tasks.Task, asyncio.Task, and _asyncio.Task can all refer to the same runtime class. Which one PyCharm reports as typeQName depends on how resolution landed: through the binary skeleton (_asyncio) vs the re-exported names in the asyncio package.

### DIFF
--- a/python/python-psi-impl/src/com/jetbrains/python/inspections/PyAsyncCallInspection.kt
+++ b/python/python-psi-impl/src/com/jetbrains/python/inspections/PyAsyncCallInspection.kt
@@ -67,7 +67,11 @@ class PyAsyncCallInspection : PyInspection() {
       AWAITABLE, COROUTINE
     }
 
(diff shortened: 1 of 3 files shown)


avokin2:
Merged

PR #3227, opened 2025-09-10
[JEWEL-857] LazySelectableList/LazyTree/Tabs/SegmentedButton do not read content from selected item

faogustavo:
- Updated sample app semantics configuration, marking containers as traversal groups
- Adding 'SelectableGroup' semantics to the SelectableLazyColumn
- Setting role for the item in the 'Modifier.selectable' from SelectableLazyColumn to read the state correctly
- Merging children nodes in the 'Modifier.selectable' to ensure the screen reader understands it as one and reads the text properly
- Merging the child nodes in the SimpleListItem to ensure the screen reader understands it as one and reads the text properly
- Tabs and segmented buttons are working properly and have not needed any change

## Evidences

| When | Video |
| --- | --- |
| Tree | tree
| Selectable Lazy Column | selectable lazy column
| Segmented Control | segmented controls
| Tabs | tabs

### Bug fixes
 * 'SimpleListItem' merges the semantic tree, ensuring it will read the content description from the children elements
 * 'SelectableLazyColumn' selectable items now merge the semantics tree, ensuring it will read the content description from the children elements
 * 'SelectableLazyColumn' selectable items now read the selected/unselected state correctly
   * If the list uses multiple selection mode, items will be considered checkboxes
   * If the list uses single selection mode, items will be considered radio buttons
 * Properly setting `onClick` accessibility action to ensure items can be correctly toggled with a screen reader


platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/ChipsAndTree.kt
```diff
@@ -177,27 +175,27 @@ public fun SelectableLazyColumnSample(modifier: Modifier = Modifier) {
...
+
                     SimpleListItem(
                         text = item,
                         selected = isSelected,
                         active = isActive,
-                        modifier =
-                            Modifier.fillMaxWidth().selectable(isSelected) {
-                                JewelLogger.getInstance("ChipsAndTree").info("Click on $item")
-                            },
+                        modifier = Modifier.fillMaxWidth(),
```
faogustavo, line 194: Using the `.selectable`, it does not get triggered when using accessibility services. So I replaced it with a launch effect that logs when the entry gets selected.

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/views/ComponentsView.kt
```diff
@@ -43,7 +45,11 @@ public fun ComponentsView(
     toolbarButtonMetrics: IconButtonMetrics,
     modifier: Modifier = Modifier,
 ) {
-    Row(modifier.trackActivation().fillMaxSize().background(JewelTheme.globalColors.panelBackground)) {
+    Row(
+        modifier.trackActivation().fillMaxSize().background(JewelTheme.globalColors.panelBackground).semantics {
+            isTraversalGroup = true
```
faogustavo, line 50: Added the `isTraversal` to all root content items in our samples so the screen reader does not group it and wait to "enter" the focus

| Before | After |
| --- | --- |
| Image | Image

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -229,9 +236,30 @@ private fun Modifier.selectable(
...
+                }
+
+            onClick {
+                actionHandler.toggleKeySelection(
+                    selectableLazyListState = selectableState,
+                    selectionMode = selectionMode,
+                    allKeys = allKeys,
+                    key = itemKey,
+                )
+                requester?.requestFocus()
```
faogustavo: Calling requestFocus before handing the action was causing some intermittent automatic selection of the first item based on the focus logic:

```kt
.onFocusChanged {
    isFocused = it.hasFocus
    with(state) {
        if (isFocused && lastActiveItemIndex == null && selectedKeys.isEmpty()) {
            keyActions.actions.onSelectFirstItem(keys, this)
        }
    }
}
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -229,9 +236,30 @@ private fun Modifier.selectable(
         }
         // Add semantics for accessibility
         .semantics(mergeDescendants = true) {
-            selected = itemKey in selectableState.selectedKeys
+            val isSelected = itemKey in selectableState.selectedKeys
+
+            selected = isSelected
+            toggleableState = ToggleableState(isSelected)
```
rock3r: I'm not sure this is correct — it's not a checked/unchecked situation here

faogustavo: Yeah, I'm not sure either, but I was not able to make the screen reader "tell" the selection state. 

This was the closest solution I've found. We are using the same for tabs and segmented buttons.

faogustavo: Ok, so I decided to take a step back and check how the IDE works with accessibility in this scenario. Looks like they don't allow any selection (either single or multi-selection) if screen readers are enabled.

https://github.com/user-attachments/assets/ce7580e8-b66f-4adc-b0db-6306cfd7aeff

I was trying to ensure that it announces the text and the element is "interactive", but that's not what it is doing on Swing. I'll revert those changes and only keep the state read.

faogustavo: Reverted to match the Swing implementation

rock3r:
Ready to merge

PR #3228, opened 2025-09-10
[JEWEL-174] Fixed ToolWindow focus

faogustavo:
- The `.trackActivation()` call on the Column modifier was overriding the component activation value
  - With modifier added, we would only track the activation IF the column (or any descendant) get the focus
- To fix the problem, I just removed the track activation to use the ToolWindow activation state
- Additionally, I've added a FocusRequester to focus on the item when we open the ToolWindow
  - This is similar to what is done in the ReleaseDemo
- As all changes are only to the sample, no release notes are required


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
```diff
@@ -114,32 +119,38 @@ private fun createJewelComposePanel(config: ComposePanel.(JewelComposePanelWrapp
...
-    val composePanel = ComposePanel()
-    jewelPanel.add(composePanel, BorderLayout.CENTER)
-    composePanel.config(jewelPanel)
+    jewelPanel.composePanel.config(jewelPanel)
     ComposeUiInspector(jewelPanel)
     return jewelPanel
 }
 
-internal class JewelComposePanelWrapper : JPanel(), UiDataProvider {
+internal class JewelComposePanelWrapper : JPanel(BorderLayout()), UiDataProvider {
```
rock3r:
```suggestion
internal class JewelComposePanelWrapper : BorderLayoutPanel(), UiDataProvider {
```
We can directly use `BorderLayoutPanel` in this case :)

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
```diff
@@ -114,32 +119,38 @@ private fun createJewelComposePanel(config: ComposePanel.(JewelComposePanelWrapp
...
+    }
 
-    val composePanel: ComposePanel
-        get() =
-            components.singleOrNull() as? ComposePanel
-                ?: error("JewelComposePanelWrapper was not initialized with a ComposePanel")
+    val composePanel: ComposePanel = ComposePanel()
+
+    init {
+        super.addImpl(composePanel, BorderLayout.CENTER, -1)
```
rock3r: If `this` is a `BorderLayoutPanel` then this becomes:
```suggestion
        super.addToCenter(composePanel)
```

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
```diff
@@ -114,32 +119,38 @@ private fun createJewelComposePanel(config: ComposePanel.(JewelComposePanelWrapp
...
     return jewelPanel
 }
 
-internal class JewelComposePanelWrapper : JPanel(), UiDataProvider {
+internal class JewelComposePanelWrapper : JPanel(BorderLayout()), UiDataProvider {
     internal var targetProvider: UiDataProvider? = null
+    private val listener = AWTEventListener { event ->
+        if (event !is MouseEvent || event.button == MouseEvent.NOBUTTON) return@AWTEventListener
+        if (!isFocusOwner && SwingUtilities.isDescendingFrom(event.component, this)) {
+            composePanel.requestFocus()
```
rock3r, line 218: I'm wondering if we should make this something that can be disabled with a property? E.g., `var requestFocusOnClick = true`

faogustavo: That's a good idea. We can have a flag for it. I'll add one

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
faogustavo: Updated

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
faogustavo: Updated

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
faogustavo: Added it as a parameter when creating the compose panel

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/JewelComposePanelWrapper.kt
```diff
@@ -27,11 +31,28 @@ import org.jetbrains.jewel.ui.util.LocalMessageResourceResolverProvider
...
+ *
+ * The [content] is wrapped in a [SwingBridgeTheme], which will be derived from the current Swing LaF.
+ *
+ * @param config A lambda to configure the underlying [ComposePanel].
+ * @param content The Composable content to display.
+ */
+@Deprecated("Use the version with 'focusOnClickInside' parameter")
 public fun compose(config: ComposePanel.() -> Unit = {}, content: @Composable () -> Unit): JComponent =
-    JewelComposePanel(config, content)
+    JewelComposePanel(focusOnClickInside = false, config, content)
```
faogustavo, line 55: Keeping "false" in the old APIs to keep the same behavior

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/ToolWindowExtensions.kt
```diff
@@ -17,6 +19,7 @@ public fun ToolWindow.addComposeTab(
     @TabTitle tabDisplayName: String? = null,
     isLockable: Boolean = true,
     isCloseable: Boolean = false,
+    focusOnClickInside: Boolean = true,
```
rock3r, line 22: I reckon the default for toolwindows should be true, to match Swing's behaviour. Wdyt?

faogustavo: That makes sense; I'll update it to have parity with swing.

faogustavo: Updated

rock3r (approved):

rock3r:
Ready to merge

jreznot:
This has rather bad DX consequences in all usages of `compose {}`

Screenshot 2025-09-26 at 16 33 42

PR #3232, opened 2025-09-12
Fix fqname for ByteBuddyDynamicAttach

gharrma:
The reference to ByteBuddyDynamicAttach was originally added in IntelliJ commit c175babc14, but the class was later renamed by commit https://github.com/Kotlin/kotlinx.coroutines/commit/1500c83c62.

Without this change, invocations of DebugProbes.install() in IDE plugin tests can result in an error from bytebuddy:

    "class redefinition failed: attempted to delete a method"

For more background see https://youtrack.jetbrains.com/issue/IJPL-166.

cc @develar @sellmair

### DIFF
--- a/platform/build-scripts/src/org/jetbrains/intellij/build/mergeJars.kt
+++ b/platform/build-scripts/src/org/jetbrains/intellij/build/mergeJars.kt
@@ -338,7 +338,7 @@ private fun getIgnoredNames(): Set<String> {
     }
   }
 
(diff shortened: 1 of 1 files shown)


develar:
Thanks! WIll be merged soon.

develar:
I filed an issue to improve the current solution —https://youtrack.jetbrains.com/issue/IJPL-207588/Get-rid-of-the-hack-to-enable-debug-support-for-coroutines (private, I made it visibile to you ).

gharrma:
Thanks!

PR #3236, opened 2025-09-15
[JEWEL-1001] Fix excessive line height with IDE zoom

rock3r:
This fixes the line calculation for text in the IDE when the IDE zoom feature is enabled. We were not measuring the line height correctly in that case, as it turns out that for LaF fonts we actually need to take the IDE zoom into account and divide all metrics by it to get the right result.

Amusingly, this is not the case for editor scheme fonts — those are not pre-scaled like LaF fonts are, so special accommodations are made for those.

### Bug fixes
 * Fixed line height calculations with IDE zoom on in the bridge


platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeText.kt
```diff
@@ -107,13 +111,26 @@ private val image = ImageUtil.createImage(1, 1, TYPE_INT_ARGB)
 /**
  * Computes the "base" line height with the same logic used by
  * [com.intellij.openapi.editor.impl.view.EditorView.initMetricsIfNeeded].
+ *
+ * @param alreadyUnscaled When true, the font metrics are not "unscaled" (i.e., they do not get compensation for the IDE
+ *   scale). This is useful e.g., for editor scheme fonts, which are not scaled, contrary to LaF fonts.
  */
-internal fun computeBaseLineHeightFor(font: Font): Int {
+internal fun computeBaseLineHeightFor(font: Font, alreadyUnscaled: Boolean): TextUnit {
```
faogustavo: nit: how about simplify to `isUnscaled`, or maybe `isScaled` to not be a "negation"

rock3r: Makes sense 👍

rock3r: Done, renamed to `treatAsUnscaled` and clarified KDoc

daaria-s (approved):

rock3r:
Ready to merge

PR #3238, opened 2025-09-17
[JEWEL-1006] Fix on JDialogRendered for submenus

faogustavo:
- Updated the click outside check the see if the click happened in any of the child components

## Evidences

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/871e9955-9e3b-45e6-92cb-2964d205ece4" /> | <video src="https://github.com/user-attachments/assets/8180d5e4-9318-4fcc-8f16-127803720e87" /> |

### Bug fixes
 * Fixed an issue in the custom popup renderer `JDialogRenderer` that was not checking for clicks on child popups, triggering the click outside callback


rock3r:
Ready to merge

PR #3243, opened 2025-09-19
[JEWEL-962] Showing the selected item when opening the combo box

faogustavo:
- Fixed an issue where the scroll call was happening with indices that do not match the available item index
- Fixed the issue where the first render was not moving to the selected index
- Running the scroll on launch instead of dispose as it may not behave well when dismissing by clicking an option

## Evidences

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/a899b407-eec6-4957-8187-882bec941456" /> | <video src="https://github.com/user-attachments/assets/942671c2-12e9-4e63-a5a6-fbfc64cd3830" /> |

### Bug fixes
 * When the ListComboBox shows the pop-up for the first time, it will automatically scroll to the selected item;


rock3r:
Ready to merge

PR #3245, opened 2025-09-22
[JEWEL-977] Validate IJP API dumps are up-to-date on CI

rock3r:
This change makes sure we validate that the API dumps are up-to-date in the GitHub Actions CI. It also rewrites the `validate-api-dump-changes.main.kts` script to use clikt, like newer ones do.


.github/workflows/jewel-checks.yml
```diff
@@ -100,32 +100,58 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
         with:
+          # We need fetch-depth: '2' so we get HEAD~1 in case this is missing the PR token
           fetch-depth: '2'
         name: Check out repository
 
       - name: Grant execute permission to the validation script
-        run: chmod +x ./scripts/validate-api-dump-changes.main.kts
+        run: chmod +x ./scripts/annotate-api-dump-changes.main.kts
```
faogustavo, line 114: Nit: Can we change to make both files executable in one step?


```suggestion
        run: |
          chmod +x ./scripts/annotate-api-dump-changes.main.kts
          chmod +x platform/jewel/scripts/check-api-dumps.main.kts
```

.github/workflows/jewel-checks.yml
```diff
@@ -100,32 +100,58 @@ jobs:
...
+        working-directory: platform/jewel
 
   metalava:
     name: Check for breaking API changes with Metalava
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: '1'
```
faogustavo, line 128: Was this intended?

platform/jewel/scripts/annotate-api-dump-changes.main.kts
```diff
@@ -0,0 +1,241 @@
...
+            apiDumpFiles
+                .map { file ->
+                    async {
+                        var foundBreakages = false
+                        val log = buildString {
+                            appendLine("\n  Checking ${file.toRelativeString(jewelRoot)}")
+
+                            val isModifiedResult =
+                                runCommand(
+                                    "git diff --quiet $baseCommit -- ${file.absolutePath}",
```
faogustavo, line 86: I know this is old code that was moved, but does this work if, for some reason, we have more than one commit in the branch? 

It is unlikely to happen, as we have other validation checks in place. I'm just thinking about possible interactions with the `fetch-depth: '2'`

.github/workflows/jewel-checks.yml
rock3r: No, because this is a separate job. Doing it here would not affect the check job

.github/workflows/jewel-checks.yml
rock3r: Yeah, 1 is the default :)

platform/jewel/scripts/annotate-api-dump-changes.main.kts
rock3r: As you point out, we have other checks that make the PR red if there is more than one commit, so that's not really a problem. Since a PR with more than one commit is invalid in any case, there's not really much of a point for the added complexity to count the commits in the PR. That would also require adjusting both the fetch-depth and this logic accordingly.

It is more of a theoretical issue than a practical one, so back then I decided not to bother :) I still think it's the right choice

rock3r:
Ready to merge

PR #3247, opened 2025-09-23
[JEWEL-978] Check Bazel build is up-to-date in GitHub CI

rock3r:
This change adds a step to the GitHub CI to run the Bazel build generator script, and then check it has not made changes to any Bazel build files (ignoring the android submodule, which we do not care about). If it has, then the committer needs to run the script locally and commit & push these changes in the PR.

> [!IMPORTANT]
> ~~This PR requires #3245 (JEWEL-977) to be merged first~~ DONE

### DIFF
--- a/.github/workflows/jewel-checks.yml
+++ b/.github/workflows/jewel-checks.yml
@@ -120,7 +120,7 @@ jobs:
         run: ./scripts/annotate-api-dump-changes.main.kts
 
   check_ij_api_dumps:
-    name: Check that IJP API dumps are up-to-date
+    name: Check that the IJP API dumps are up-to-date
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


.github/workflows/jewel-checks.yml
```diff
@@ -142,10 +142,52 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        name: Check out repository
+
+      - uses: actions/checkout@v4
+        name: Checkout JetBrains/android submodule
+        with:
+          repository: JetBrains/android
+          path: android
+          ref: master
```
nebojsa-vuksic, line 165: Why do we need android here?

rock3r: IIRC (it's been a while...) JpsToBazel was failing without it

daaria-s (approved):

.github/workflows/jewel-checks.yml
rock3r: We decided on holding off on this change until @nebojsa-vuksic has a bit of time to see if he can remove the dependency on the Android submodule from the Bazel generator

nebojsa-vuksic: See https://github.com/JetBrains/intellij-community/pull/3247#pullrequestreview-3477642002

rock3r:
Ready to merge

PR #3250, opened 2025-09-24
[JEWEL-1018] Speed Searchable ComboBox

faogustavo:
- Created new component supporting String-based and Generic-based APIs
- Added support to choose SpeedSearch input position to support showing in the opposite size of the drop-down
- Added `position` to SpeedSearchState so we can show it in the oposite side of the dropdown
- Added `index` parameter to the `itemContent` to perform scrolling to the selected item

## Evidences

| Basic flow | Inverted dropdown popup |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/092a1130-4ce8-4e79-8c75-597163aa5abf" /> | <video src="https://github.com/user-attachments/assets/33339416-0d70-4ed3-972a-ed1848146b70" />

### New features
 * Added `SpeedSearchableComboBox` component that supports speed search
   * It is available inside a `SpeedSearchArea` and has a similar syntax to a normal ComboBox
   * It has two overloads, supporting string and generic lists

### DIFF
--- a/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBoxTest.kt
+++ b/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBoxTest.kt
@@ -1,0 +1,227 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.ui.component.search
+
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableIntStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+import androidx.compose.ui.input.key.Key
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.test.SemanticsNodeInteraction
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertIsSelected
+import androidx.compose.ui.test.hasAnyAncestor
+import androidx.compose.ui.test.hasTestTag
+import androidx.compose.ui.test.hasText
+import androidx.compose.ui.test.junit4.ComposeContentTestRule
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.unit.dp
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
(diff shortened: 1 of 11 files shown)


platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBoxTest.kt
```diff
@@ -0,0 +1,208 @@
...
+    private fun runComposeTest(
+        entries: List<String> = List(500) { "Item ${it + 1}" },
+        block: ComposeContentTestRule.() -> Unit,
+    ) {
+        rule.setContent {
+            val focusRequester = remember { FocusRequester() }
+
+            IntUiTheme {
+                var selectedIndex by remember { mutableIntStateOf(0) }
+                SpeedSearchArea(modifier = Modifier.testTag("SpeedSearchArea")) {
```
vprotasov: Why is `SpeedSearchArea` used 2 times here?

jakub-senohrabek:
Would it make sense to instead just add argument "searchable" to the original ListComboBox?

rock3r:
@jakub-senohrabek not really, because it's not something that can be done just with a boolean flag, if you look at the implementation

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
```diff
@@ -0,0 +1,256 @@
...
+    ListComboBoxImpl(
+        items = items,
+        selectedIndex = selectedIndex,
+        onSelectedItemChange = onSelectedItemChange,
+        itemKeys = itemKeys,
+        modifier =
+            modifier.onPreviewKeyEvent { event ->
+                if (!popupVisible) return@onPreviewKeyEvent false
+
+                if (!processKeyEvent(event) && speedSearchState.isVisible && speedSearchState.searchText.isNotEmpty()) {
```
nebojsa-vuksic: nit: `speedSearchState.isVisible && speedSearchState.searchText.isNotEmpty()` can be a single function call on a speedSearchState object. Having calls like this is a feature envy code smell

jakub-senohrabek (approved):

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBoxTest.kt
faogustavo: Probably copy/paste mistake 😬 

I"ll double check this

vprotasov: Thanks, also this test seems to be flaky, failed a few times for me
`fun `on type continue typing, continue selecting first occurrence`() `

faogustavo: Removed. Thanks for catching :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
faogustavo: Makes sense. Extracted to an extension

rock3r:
Ready to merge

PR #3239, opened 2025-09-17
[JEWEL-914] Chips style from IDE

faogustavo:
- Updated chip theme to consume disabled and metrics from LaF
- Updated border to be rounded
- Moved 'focusedBorderWidth' and 'buttonCornerSize to IntUiBridge with existing 'borderWidth'
- Updated the IDE sample to have chips

## Evidences

| Before | After | Meet the new UI |
| --- | --- | --- |
| image | image | image

### Bug fixes
 * Fixed disabled color from Chips on IDE/LaF
 * Fixed corner size from Chips on IDE/LaF

PR #3251, opened 2025-09-25
[JEWEL-914] Fix Chips height to use defaultMinHeight

faogustavo:
- In the #3239 PR, I've fixed the sizes to match the IDE implementation, however, it was breaking on changing screen density on MacOS
- Updated the height to be `defaultMinSize` to ensure it can grow if needed
- Retested the IDE to ensure it still matches the same size from the IDE

## Evidences

| Before | After |
| --- | --- |
| image | image | 

| IDE Jewel | IDE Swing |
| --- | --- |
| image | image


rock3r (approved):
Small fix, ready to merge

PR #3253, opened 2025-09-25
[JEWEL-1013,JEWEL-1016] Using compose drop shadow APIs for popups

faogustavo:
- Disabled native shadow APIs for JBPopupRenderer and JDialogRenderer
- Using the drop shadow API to replicate the same UI style

## Evidences

| Where | Before | Swing Shadow | CfD Drop Shadow |
| --- | --- | --- | --- |
| IDE | - | image | image
| Stand Alone | image | image | image

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
@@ -146,6 +146,7 @@ private fun JBPopup(
             .setCancelOnClickOutside(currentProperties.dismissOnClickOutside)
             .setCancelOnWindowDeactivation(currentProperties.dismissOnClickOutside)
             .setLocateWithinScreenBounds(false)
+            .setShowShadow(false)
             .setKeyEventHandler { event ->
                 val composeEvent = event.toComposeKeyEvent()
                 val consumed =
--- a/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
+++ b/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
@@ -195,7 +195,7 @@ private fun JPopupImpl(
             rootPane.isOpaque = false
             background = Color(0, 0, 0, 0)
             contentPane.background = Color(0, 0, 0, 0)
-            rootPane.putClientProperty("Window.shadow", true)
+            rootPane.putClientProperty("Window.shadow", false)
         }
     }
 
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
@@ -40,7 +40,7 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.drawBehind
-import androidx.compose.ui.draw.shadow
+import androidx.compose.ui.draw.dropShadow
 import androidx.compose.ui.focus.FocusManager
 import androidx.compose.ui.focus.FocusRequester
 import androidx.compose.ui.focus.focusRequester
@@ -48,6 +48,7 @@ import androidx.compose.ui.geometry.CornerRadius
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.geometry.Size
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.shadow.Shadow
 import androidx.compose.ui.input.InputMode
 import androidx.compose.ui.input.InputModeManager
 import androidx.compose.ui.input.key.Key
@@ -59,6 +60,7 @@ import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalFocusManager
 import androidx.compose.ui.platform.LocalInputModeManager
 import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.DpOffset
 import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.window.PopupPositionProvider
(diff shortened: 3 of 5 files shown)


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -236,11 +238,13 @@ public fun MenuContent(
...
-                    elevation = style.metrics.shadowSize,
-                    shape = menuShape,
-                    ambientColor = colors.shadow,
-                    spotColor = colors.shadow,
+                .dropShadow(
+                    RoundedCornerShape(style.metrics.cornerSize),
+                    Shadow(
+                        radius = style.metrics.shadowSize,
+                        color = colors.shadow,
+                        offset = DpOffset(0.dp, style.metrics.shadowSize / 2),
```
faogustavo: On Swing, it was using the same size from the shadow as the offset, to prevent shadow on top. 

On this Compose call, only half of the value is needed to reproduce the desired UI.

| full size | base-line | half size |
| --- | --- | --- |
| Image | Image | Image

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
```diff
@@ -146,6 +146,7 @@ private fun JBPopup(
             .setCancelOnClickOutside(currentProperties.dismissOnClickOutside)
             .setCancelOnWindowDeactivation(currentProperties.dismissOnClickOutside)
             .setLocateWithinScreenBounds(false)
+            .setShowShadow(false)
```
rock3r: I don't think this is the right approach — we should use the same shadow that other popups do (which is drawn by the OS, now Swing).

Us drawing our own shadow is the right choice only when we're not using native popups. It's a fallback.

If the native popups don't show shadows, I think the issue may be in CMP clipping it out (that famous custom property I keep asking you about and that I do not think should be at all necessary for native popups!)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -195,7 +195,7 @@ private fun JPopupImpl(
             rootPane.isOpaque = false
             background = Color(0, 0, 0, 0)
             contentPane.background = Color(0, 0, 0, 0)
-            rootPane.putClientProperty("Window.shadow", true)
+            rootPane.putClientProperty("Window.shadow", false)
```
rock3r: Same here — this should not be the case for native popups. Native windows have a shadow drawn by the OS.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
faogustavo: Updated to make sure we only use it when the native pop-up flag is disabled.

| Swing | Compose |
| --- | --- |
| image | image

rock3r: Thanks :) Does the native popup shadow work now, or do we still need a separate bug for that?

rock3r (approved):

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
faogustavo: It should be good. I'll boot the Windows/Linux VMs to ensure it works on other platforms as well. But all CMP and Native shadows should have the same appearance

rock3r: Thanks for checking

faogustavo: Looks like it's not working because we need to use the 'Undecorated' window type to hide the navigation bar. I'll check if there are other ways to achieve similar results.

The "Window.shadow" property only applies to MacOS :/

daaria-s (approved):

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
rock3r: Before we merge, any updates on this? If not, can you file it as a follow-up for 0.33?

jakub-senohrabek (approved):

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
faogustavo: Yup, I've completed testing, and shadows aren't working on Linux due to the "undecorated" setting [JEWEL-1060](https://youtrack.jetbrains.com/issue/JEWEL-1060/Shadows-do-now-show-up-on-Linux). 

For Windows, there is another issue now that is crashing when showing the custom popup [JEWEL-1059](https://youtrack.jetbrains.com/issue/JEWEL-1059/Custom-Popup-Render-does-not-show-the-popup-on-Windows).

rock3r: Ok, so we should probably wait merging this, right?

faogustavo: We are good to merge this. The problems are not related to the changes in this scope.

The existing issues on Windows are also happening on `master`, and the issues on Linux are more related to the AWT integration/configuration.

rock3r: Ok, let's ship then

rock3r:
Ready to merge

PR #3254, opened 2025-09-26
[JEWEL-965] Make Jewel Detekt rules only apply to Jewel

rock3r:
This change makes sure that the Jewel rules only apply to code in the org.jetbrains.jewel package. This avoids false positives flagged in the rest of the IJP code.


rock3r:
Ready to merge

PR #3255, opened 2025-09-26
[JEWEL-1030] Marking source-compatible deprecated methods with Hidden level

faogustavo:
- Some methods got new params, but kept the source compatible
- Those methods had their visibility changed to Hidden, so they still exist
- But if the users recompile, they will use the new method

### ⚠️ Important Changes
 * A few methods/composable functions that received new parameters, but kept the source compatibility, had their deprecation level changed to `Hidden`
   * Methods that got hidden are:
     * JewelComposeNoThemePanel (method without focusOnClickInside)
     * JewelComposePanel (method without focusOnClickInside)
     * compose (methods without focusOnClickInside)

### DIFF
--- a/platform/jewel/ide-laf-bridge/api-dump-experimental.txt
+++ b/platform/jewel/ide-laf-bridge/api-dump-experimental.txt
@@ -4,11 +4,11 @@ f:org.jetbrains.jewel.bridge.BridgeUtilsKt
 - *sf:toNonNegativeDpSize(com.intellij.util.ui.JBDimension):J
 - *sf:toNonNegativeDpSize(java.awt.Dimension):J
 f:org.jetbrains.jewel.bridge.JewelComposePanelWrapperKt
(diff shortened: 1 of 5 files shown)


faogustavo:
Breaking changes in the dump file are expected as we changed them to deprecated level hidden

faogustavo:
Adding more context on the dump changes.

The `level = Hidden` from the deprecation annotation adds the `synthetic` modifier to hide the function from the users when compiling against a new version of the code.

For older versions, the method will still be there and accessible with the same signature. It will just be a synthetic method now.

In other words, new compilations will point to the non-deprecated method. Old compilations, which now reference the (now) deprecated method, will still function as intended, since the method was not removed and its signature has not changed.

More details on synthetic change can be found in the `JvmSynthetic` annotation:

> Marking target as synthetic is a binary compatible change, already compiled Java code will be able to access such target.
> \- [JvmSynthetic](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-synthetic/)

PR #3256, opened 2025-09-26
Remove vestigial Compose runtime lib

gharrma:
This Compose runtime-desktop lib contains no classes. It appears to be a vestige of the transition from org.jetbrains.compose.runtime to androidx.compose.runtime.

This lib contains META-INF/runtime.kotlin_module which conflicts with the correct metadata inside androidx.compose.runtime:runtime-desktop. This conflict breaks compilation in third party plugins using Compose.

cc @rock3r


jreznot (approved):

gharrma:
Awesome, thanks! Note, a cherry-pick to 252 may be useful as the issue affects IJ 2025.2.2 too.

PR #3258, opened 2025-09-29
[IJPL-149653] Fix Apply button in Run/Debug configs always enabled

sgosk:
**Description**

This PR fixes an issue where 'Run/Debug Configurations' dialog's Apply button is enabled without any change in 'IDEA' run configuration in IDEA project.

**The Problem**

When selecting 'IDEA' run configuration in IDEA project and opening 'Run/Debug Configurations' dialog, the 'Apply' button should be disabled, since there were no changes made to the configuration yet. However, it was enabled without any modifications to the configuration. The core issue was incorrect creation of the configuration snapshot, which was compared to the original configuration - certain elements of the XML were omitted, and therefore the comparison resulted in a false positive.

**Related Issue:**

Fixes [IJPL-149653](https://youtrack.jetbrains.com/issue/IJPL-149653/Run-Debug-Configurations-dialogs-Apply-button-is-enabled-without-any-change)

### DIFF
--- a/platform/execution-impl/src/com/intellij/execution/impl/ConfigurationSettingsEditor.java
+++ b/platform/execution-impl/src/com/intellij/execution/impl/ConfigurationSettingsEditor.java
@@ -2,6 +2,7 @@
 
 package com.intellij.execution.impl;
 
+import com.intellij.configurationStore.SerializableScheme;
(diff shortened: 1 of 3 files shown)


batya239:
Hey, thanks for PR! Could you please add a test?

PR #3259, opened 2025-09-29
[JEWEL-1009] Using the native file picker only on MacOS

faogustavo:
- The native file picker does not fileter the files on Linux nor Windows
- Updating to use the JFileChooser always


rock3r:
Ready to merge

PR #3260, opened 2025-09-29
[JEWEL-1031] Fixed issues closing the popups automatically on ComboBoxes

faogustavo:
- The 'CustomPopupContent' was using the 'PopupContainer', which creates a new popup
- This was causing two popups to be presented together, and it was closing the previous one
- As the combo boxes already use the 'PopupContainer' inside, we can skip the composable
- Note that this bug was only in the Sample App, not in the component itself

## Evidences

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/dd762d53-9e68-4469-8616-6199c5972c4a" /> | <video src="https://github.com/user-attachments/assets/55902613-9141-4d71-8d73-c13395ec608c" /> |


rock3r (approved):

rock3r:
Ready to merge

faogustavo:
When implementing #3262, I copied this change to the other branch but forgot to revert it before committing. The other PR was already merged with this fix, so I'll close this and mark the ticket as resolved

PR #3262, opened 2025-10-02
[JEWEL-1032] Adding popupModifier to all ComboBoxes and SplitButton

faogustavo:
- Added `maxPopupWidth` and `popupModifier` parameters across multiple ComboBox and SplitButton components
- Added tests to check the size constraints

## Notes

- I attempt to only expose the 'popupModifier', but unfortunately, that route does not work well with Lazy components (like the ListComboBox that uses LazyColumn)
- Removing the 'maxPopupHeight' was not a good idea, because if the user wanted to intentionally have a popup larger than the value set on the theme metric, it was not allowing
- Adding the 'maxPopupWidth' was able to give freedom to the pop-up to grow if possible, but allowing the users to limit how much it can
- Using the 'requiredSize' modifiers was able to fix most of the problems; however, it was not calculating the size until the pop-up was shown, breaking the positioning logic
- To prevent broken UI, all ListComboBoxes have a pre-defined size limitation, to ensure it does not occupy all the screen width

### Useful information

- To make sure ListComboBoxes don't take too much space, we are using the 'Dp.Hairline'. This is a 0.dp const value, and the min size constraints will ensure to coerce to be at least the owner width
- We can use the same idea to "enforce" the pop-up to always be the same width as the button

## Evidences

| Before | After | Notes |
| --- | --- | --- |
| image | image | The pop-up was not growing to match the content, and it was limited to button size
| image | image | The pop-up was not growing to match the content, and it was limited to button size
| image | image | The pop-up was not growing to match the content, and it was limited to button size
| image | image | Ensuring the lazy component does not grow to take all screen size

### ⚠️ Important Changes
 * SplitButtons and ComboBoxes pop-ups are now allowed to grow over the owner's width
   * You can still limit the pop-up width to always match the owner width by passing the 'Dp.Hairline' in the 'maxPopupWidth' property
   * Note that this is not true for "ListComboBoxes" as they use a Lazy component, which always occupies the whole width
     * If you want to allow the pop-up to grow, just use the 'maxPopupHeight' parameter with a bigger value

### New features
 * All ComboBoxes API now have parameters to control the pop-up. They are:
   * `maxPopupHeight` - Defines the maximum height for pop-up
   * `maxPopupWidth` - Defines the maximum width for the pop-up
     * This property will always be coerced to have at least the same width as the owner
   * `popupModifier` - Modifier applied to the pop-up container that can be used to control other properties

### Deprecated API
 * Deprecated all SplitButtons and ComboBoxes composables that don't support `maxPopupWidth` or `popupModifier`


platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/ComboBoxes.kt
```diff
@@ -275,21 +295,27 @@ private fun CustomComboBoxes() {
...
-            verticalArrangement = Arrangement.spacedBy(8.dp),
-        ) {
-            InfoText("Your custom content here!")
-            DefaultButton(onClick = onButtonClick) { Text("Pick a random item") }
-        }
+private fun CustomPopupContent(onButtonClick: () -> Unit) {
+    Column(Modifier.padding(horizontal = 16.dp, vertical = 8.dp), verticalArrangement = Arrangement.spacedBy(8.dp)) {
+        Text(
+            "Your custom content here! Generic popup content with a bigger text that will check if the popup can handle it properly.",
+            overflow = TextOverflow.Ellipsis,
```
rock3r: Does this do anything without a `maxLines` set?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
```diff
@@ -279,10 +340,18 @@ public fun ComboBox(
...
+                    Modifier.testTag("Jewel.ComboBox.Popup")
                         .heightIn(max = maxHeight)
-                        .width(comboBoxWidth)
+                        .widthIn(
+                            min = comboBoxWidth,
+                            max =
+                                if (maxPopupWidth.isSpecified) {
+                                    maxPopupWidth.coerceAtLeast(comboBoxWidth)
+                                } else {
+                                    maxPopupWidth
```
rock3r: Doesn’t this resolve to `Dp.Unspecified`?

Besides, I am not a fan of suggesting the `Hairline` trick. If anything, it should be `0.dp`, but neither actually make semantic sense.

I think, according to the usual semantics of `Unspecified`, the behaviour should be:
 * If the `maxPopupWidth` is `Dp.Unspecified`, then the popup max width is the combo box width
 * If the `maxPopupWidth` is specified, then the popup max width is whatever you pass in

I understand that you're trying to make the default behaviour be "actually, there is no max width", but I'm not sold on the API design. Maybe you're trying to cram too much meaning into a Dp; maybe we need a `PopupMaxWidthBehavior` sealed interface that has `Unlimited`, `MatchParent`, and `Custom`? Or maybe we have a `popupMaxWidthProvider: (comboBoxWidth: Dp) -> Dp = { Dp.Unspecified }` lambda? Dunno. Ideas are welcome :D

In any case we should have the max width coerced to be at least `comboBoxWidth`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -194,10 +197,19 @@ public fun EditableComboBox(
                     }
                 },
                 modifier =
-                    popupModifier
-                        .testTag("Jewel.ComboBox.Popup")
+                    Modifier.testTag("Jewel.ComboBox.Popup")
                         .semantics { contentDescription = "Jewel.EditableComboBox.Popup" }
-                        .width(comboBoxWidth)
+                        .heightIn(max = maxPopupHeight)
+                        .widthIn(
```
rock3r: Ditto :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -175,10 +253,12 @@ public fun ListComboBox(
         enabled = enabled,
         outline = outline,
         maxPopupHeight = maxPopupHeight,
+        maxPopupWidth = maxPopupWidth.takeOrElse { Dp.Hairline },
```
rock3r: I don't think it should be done here, but rather, in the impl

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -492,7 +717,6 @@ internal fun <T : Any> ListComboBoxImpl(
                 previewSelectedItemIndex = hoveredItemIndex,
                 currentlySelectedIndex = selectedIndex,
                 listState = listState,
-                popupMaxHeight = popupMaxHeight,
```
rock3r, line 495: Why is this removed? We don't need it anymore?

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ComboBoxTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    @get:Rule val composeRule = createComposeRule()
+
+    private val popupMenu: SemanticsNodeInteraction
+        get() = composeRule.onNodeWithTag("Jewel.ComboBox.Popup")
+
+    private val comboBox: SemanticsNodeInteraction
+        get() = composeRule.onNodeWithTag("ComboBox")
+
+    @Test
+    fun `popup width can be bigger than button width when nothing is set`() {
```
rock3r: Nit: not technically a button :P

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Button.kt
```diff
@@ -451,30 +720,43 @@ private fun SplitButtonImpl(
...
-            if (secondaryContent != null) {
-                PopupContainer(
-                    modifier = Modifier.width(buttonWidth).onClick { popupVisible = false },
-                    onDismissRequest = { popupVisible = false },
-                    horizontalAlignment = Alignment.Start,
-                    content = secondaryContent,
-                )
+            val splitButtonPopupModifier =
+                Modifier.heightIn(max = maxPopupHeight)
+                    .widthIn(
```
rock3r: Ditto

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/ComboBoxes.kt
faogustavo: Nope. I just copy/pasted the code from 'InfoText' and forgot to remove this 🤦

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
faogustavo: Not needed anymore. The size constraint was moved to the parent container

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
faogustavo: I added the if condition to prevent the coercion from happening on Dp.Unspecified, however, the logic of `Comparable<Dp>` already handles this, so I can simplify.

I also agree about trying to take too much from the Dp type, but I was trying to keep the API more similar to what we already have in place. The provider idea is nice, but I was able to work on a new solution.

- Dp.Unspecified will allow the content to grow automatically
- When using a "ListComboBox" type, the max width will fallback to the same size as the component

The hairline trick is not required anymore. But if you still prefer a different solution, I'm okay with implementing the providers.

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/ComboBoxes.kt
faogustavo: Removed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
rock3r: Sounds good :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
```diff
@@ -129,6 +188,7 @@ public fun ComboBox(
  * @param enabled Controls whether the combo box can be interacted with
  * @param outline The outline style to be applied to the combo box
  * @param maxPopupHeight The maximum height of the popup
```
rock3r: Nit: can we describe the effect of Unspecified on this too? (same on the other components/overloads) For consistency

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
```diff
@@ -129,6 +188,7 @@ public fun ComboBox(
...
+ * @param maxPopupWidth The maximum width of the popup. If it's not set, it will allow the content to grow as needed
```
rock3r:
```suggestion
 * @param maxPopupWidth The maximum width of the popup. If it's unspecified, it will allow the content to grow as needed
```

(same on other components/overloads)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
```diff
@@ -114,6 +116,63 @@ public fun ComboBox(
...
+ * @param style The visual styling configuration for the combo box
+ * @param textStyle The typography style to be applied to the text
+ * @param onArrowDownPress Called when the down arrow key is pressed while the popup is visible
+ * @param onArrowUpPress Called when the up arrow key is pressed while the popup is visible
+ * @param popupManager Manager for controlling the popup visibility state
+ * @param popupContent Composable content for the popup
+ */
+@Suppress("UnavailableSymbol") // TODO(JEWEL-983) Address Metalava suppressions
+@Composable
+@Deprecated("Use the option with 'maxPopupWidth'")
```
nebojsa-vuksic: Can we improve our depreciation messages in overall? IMHO, `"Use the option with 'maxPopupWidth'"` doesn't help offloading the burden of manually figuring out a non-deprecated ComboBox function signature of the developer.

Maybe we want to start adding `repaceWith` param to help out with this. Something like: 
```@Deprecated(
    "Use the ComboBox() overload that also specifies 'maxPopupWidth'.",
    replaceWith = ReplaceWith("ComboBox(labelText, popupContent, maxPopupHeight = "Dp.Unspecified",  maxPopupWidth = Dp.Unspecified)")
)```

File: platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt [hunk 11]
```diff
@@ -292,6 +353,62 @@ public fun ComboBox(
...
+ * @param interactionSource Source of interactions for this combo box
+ * @param style The visual styling configuration for the combo box
+ * @param onArrowDownPress Called when the down arrow key is pressed while the popup is visible
+ * @param onArrowUpPress Called when the up arrow key is pressed while the popup is visible
+ * @param popupManager Manager for controlling the popup visibility state
+ */
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+@Composable
+@Deprecated("Use the option with 'maxPopupWidth'")
```
nebojsa-vuksic: Same as a comment [above](https://github.com/JetBrains/intellij-community/pull/3262/files#r2428685673).

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
faogustavo: I was reviewing them, and as we have source compatibility with the new methods, we can directly move to deprecated hidden.

nebojsa-vuksic (approved):

rock3r:
Ready to merge

PR #3384, opened 2026-01-12
[JEWEL-1212] Calculate ComboBox Popup Height Based on maxPopupRowCount

DanielSouzaBertoldi:
# Context
While investigating why the ComboBox's popup height in the standalone version of Jewel wasn't filling the entire available height as is the case in bridge, I noticed something:

_Our popup height was not mirroring Swing's behaviour at all!_

If we wanted to take the quick route to keep parity between Jewel's bridge and standalone implementations, then this fix would be quite simple (just set `popUpMaxHeight` to `Dp.Unspecified` instead of the current `200.dp` value). But if the idea is to faithfully replicate Swing's behaviour, why not fix this for good?

Below you can see how Swing renders the popup list and how we are rendering the list in bridge:

| Swing | Jewel Bridge | Jewel Standalone |
|--------|--------|--------|
| <video src="https://github.com/user-attachments/assets/8aafe06d-0453-47b5-8671-990016b64665" /> | <video src="https://github.com/user-attachments/assets/cd1d629a-1f33-4438-add1-7a2e241d7385" /> | <video src="https://github.com/user-attachments/assets/7ab156b6-3aeb-49bd-b812-f284bbd3c1d2" />

The thing is: by default, Swing will only render at most 10* (check comments) items from the list. Since each row in the list has a fixed size, it can use both informations to calculate the popup's height. So, the bulk of this PR is adding a new property to `ComboBoxMetrics` that users can use to decide how many items they want the popup to display at once. A consequence of this choice is that now all public Composables related to ComboBox don't need the `popUpMaxHeight` parameter anymore, since we are calculating the height based on that new parameter from `ComboBoxMetrics`.

<details><summary>For a deep dive into how IJP's/Java implementation of ComboBox popup height works, click here!</summary>
<p>

In the standard `javax.swing.JComboBox.java` class there's a property called `maximumRowCount`, declared as:

```java
    /**
     * This protected field is implementation specific. Do not access directly
     * or override. Use the accessor methods instead.
     *
     * @see #getMaximumRowCount
     * @see #setMaximumRowCount
     */
    protected int maximumRowCount = 8;
```

In `com.intellij.ide.ui.laf.darcula.ui.DarculaComboBoxUI`, in the `show()` function the popup size is calculated as follows:

```java
popupSize.setSize(popupSize.width - (insets.right + insets.left), getPopupHeightForRowCount(comboBox.getMaximumRowCount()));
```

`getPopupHeightForRowCount` is the key here. The value being passed as the parameter here is 8, as shown previously, from the `JComboBox.java` class.

This function calculates the popup height by:
- Getting the base height from parent UI
   ```java
   int result = super.getPopupHeightForRowCount(maxRowCount); // This will call `BasicComboPopup.java`
   ```
- Adding actual measured row heights via `UIUtil.updateListRowHeight()`
   ```java
   for (int i = 0; i < rowCount; i++) {
       Component c = renderer.getListCellRendererComponent(list, value, i, false, false);
       Dimension preferredSize = c.getPreferredSize();
       result -= preferredSize.height;  // Remove estimated height
       result += UIUtil.updateListRowHeight(preferredSize).height;  // Add actual height (min 24dp)
   }
   ```
- Add popup insets
   ```java
   Insets borderInsets = ExperimentalUI.isNewUI() ? getListVerticalInset() : JBInsets.emptyInsets();
   return result + borderInsets.top + borderInsets.bottom;
   ```

To sum up, the calculation can be broken down into three variables:

1. The number of visible items (`maxPopupRowCount`, defaults to 8)
2. The height of each row in the list (minimum is `24.dp` from `JBUI.CurrentTheme.List.rowHeight()`, but can be taller)
3. The content padding (multiple insets can be chosen depending on the IDE configuration)

Now, IJP's version is a bit more sophisticated than ours because it supports items with different height. Ours doesn't, it only relies on `JewelTheme.globalMetrics.rowHeight`. But for 99% of the cases this solution will do. The alternative to that is facing a chicken-and-egg problem: we'd need to measure items to know their height, but items won't render until the popup is shown, and the popup size constraints affect how items render.

Some IJP ComboBoxes such as `com.intellij.openapi.ui.ComboBoxWithWidePopup` actually define a default value of 10 rows to be shown instead of 8, but the idea here is to stick to the basic implemenatation.

</p>
</details> 

# Changes
- Add `maxPopupRowCount` property to ComboBoxMetrics (default value: 10)
- Extract `ListComboBoxImpl` and `EditableListComboBoxImpl` internal functions
- Created new public `ComboBox`/`ListComboBox`/`SpeedSearchableComboBox` functions without `maxPopupHeight` parameter
- Add `calculatePopupHeight()` helper that calculates the final popup height based on the number of rows + the height of each row + the content padding for the popup.
- *BONUS* Added KDoc to `SpeedSearchableComboBox` Composables

# Evidences
| Composable | Recording |
|--------|--------|
| ListComboBox | <video src="https://github.com/user-attachments/assets/dad886a4-116e-4731-9cb8-ee93ae312db9" /> |
| EditableComboBox | <video src="https://github.com/user-attachments/assets/e1c844f9-409d-457a-b2c1-b887d04521bd" /> |
| SpeedSearchableComboBox | <video src="https://github.com/user-attachments/assets/64c1f9f6-87b1-426b-876d-6199d3f99e04" /> |

| Bridge | Standalone |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/c3db13af-9f17-40c8-8e43-194f5e74fa4a" /> | <video src="https://github.com/user-attachments/assets/cb63afdf-447e-4f7d-8349-683f22c1014c" /> |

|  | Empty Popup |
|--------|--------|
| Swing Component | image |
| Jewel Bridge | image |
| Jewel Standalone | image |

### ⚠️ Important Changes
  * **Behavioral Breaking Change for list ComboBoxes**:
      * The default sizing behavior for `ListComboBox`, `EditableListComboBox`, and `SpeedSearchableComboBox` has changed from pixel-based to row-count-based.
      * New Default: Popups now default to displaying 10 rows.
      * Customization: You can adjust this global default via `ComboBoxMetrics.maxPopupRowCount` or override it per component.
      * Fallback Change: `ComboBoxMetrics.maxPopupHeight` is no longer used as a fallback for these list-based wrappers when their internal `maxPopupHeight` is unspecified. It now only applies to standard `ComboBox` components using custom `popupContent`.
      * Action Required: If your implementation requires a fixed pixel height rather than dynamic row-based sizing, you must explicitly continue passing a value to the `maxPopupHeight` parameter at the component level.


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
```diff
@@ -43,12 +43,137 @@ import org.jetbrains.jewel.ui.component.ProvideSearchMatchState
...
 import org.jetbrains.jewel.ui.disabledAppearance
 import org.jetbrains.jewel.ui.theme.comboBoxStyle
 import org.jetbrains.jewel.ui.theme.popupContainerStyle
 
+/**
+ * A combo box with integrated speed search functionality for string items.
+ *
+ * This composable provides a dropdown selection component with built-in speed search capabilities, allowing users to
+ * quickly filter and select items by typing. Speed search is activated automatically when the user starts typing while
+ * the popup is visible.
```
DanielSouzaBertoldi: Please let me know if this description is insufficient for the purposes of speed search, I tried my best 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -97,6 +96,101 @@ import org.jetbrains.jewel.ui.theme.popupContainerStyle
...
+ * @param listState The State object for the selectable lazy list in the popup
+ * @param itemContent Composable content for rendering each item in the list
+ * @see com.intellij.openapi.ui.ComboBox
+ */
+@Deprecated(
+    "Use the overload without maxPopupHeight parameter. The popup height is now calculated based on maxPopupRowCount in ComboBoxMetrics.",
+    ReplaceWith(
+        "ListComboBox(items, selectedIndex, onSelectedItemChange, itemKeys, modifier, popupModifier, enabled, outline, maxPopupWidth, interactionSource, style, onPopupVisibleChange, listState, itemContent)"
+    ),
+)
```
DanielSouzaBertoldi: It's important to note that this class has **two** `ListComboBox` marked as Deprecated. One of them is marked as deprecated with the following reason:

> "Deprecated in favor of the method with 'popupModifier' and 'maxPopupWidth' parameters"

And now there's this one that's just been deprecated 😎 

Developer experience-wise, this can be a bit frustrating when trying to find what parameter makes the real difference between impls

Ideally one of them should be deleted for good, but I understand that it's not that simple

DanielSouzaBertoldi (commented):
Also, please let me know if I left out/did anything that doesn't conform to the ✨ Jewel's way of coding ✨

Suggestions are not just welcomed, I **DEMAND** them

(no pressure, tho!)

DanielSouzaBertoldi:
Just noticed the size of an empty popup is now smaller than it should be. I will look into this when I have the time (hopefully soon!)

platform/jewel/int-ui/int-ui-standalone/api-dump.txt
```diff
@@ -111,8 +111,8 @@ f:org.jetbrains.jewel.intui.standalone.styling.IntUiComboBoxStylingKt
...
-- bs:undecorated-EfrW7C0$default(org.jetbrains.jewel.ui.component.styling.ComboBoxMetrics$Companion,J,J,androidx.compose.foundation.shape.CornerSize,androidx.compose.foundation.layout.PaddingValues,androidx.compose.foundation.layout.PaddingValues,F,F,I,java.lang.Object):org.jetbrains.jewel.ui.component.styling.ComboBoxMetrics
+- sf:undecorated-xhyNMpg(org.jetbrains.jewel.ui.component.styling.ComboBoxMetrics$Companion,J,J,androidx.compose.foundation.shape.CornerSize,androidx.compose.foundation.layout.PaddingValues,androidx.compose.foundation.layout.PaddingValues,F,F,I):org.jetbrains.jewel.ui.component.styling.ComboBoxMetrics
+- bs:undecorated-xhyNMpg$default(org.jetbrains.jewel.ui.component.styling.ComboBoxMetrics$Companion,J,J,androidx.compose.foundation.shape.CornerSize,androidx.compose.foundation.layout.PaddingValues,androidx.compose.foundation.layout.PaddingValues,F,F,I,I,java.lang.Object):org.jetbrains.jewel.ui.component.styling.ComboBoxMetrics
 f:org.jetbrains.jewel.intui.standalone.styling.IntUiDefaultBannerStylesFactory
```
DanielSouzaBertoldi, line 122: Hmmm not sure if this means I screwed up binary compatibility. I kept the old constructor and added a new one with the new `maxPopupRowCount`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/ComboBoxStyling.kt
```diff
@@ -190,7 +190,28 @@ public class ComboBoxMetrics(
     public val popupContentPadding: PaddingValues,
     public val borderWidth: Dp,
     public val maxPopupHeight: Dp,
+    public val maxPopupRowCount: Int,
```
DanielSouzaBertoldi, line 193: Also, this is not really deprecated since the `maxPopupHeight` parameter still exists in the non-deprecated constructor.

The alternative to that is creating a `ComboBoxMetrics2` (?) without the parameter but idk... I don't have any other ideas. If you have any, please lmk!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
rock3r, line 221: Shouldn't we deprecate-and-hide this?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
```diff
@@ -205,7 +205,7 @@ public fun ComboBoxMetrics.Companion.default(
     contentPadding: PaddingValues = PaddingValues(9.dp, 1.dp, 6.dp, 1.dp),
     popupContentPadding: PaddingValues = PaddingValues(vertical = 8.dp),
     borderWidth: Dp = 1.dp,
-    maxPopupHeight: Dp = 200.dp,
+    maxPopupHeight: Dp = Dp.Unspecified,
```
rock3r, line 228: No maxPopupRowCount here?

platform/jewel/int-ui/int-ui-standalone/api-dump.txt
rock3r: I think this is because you haven't kept a deprecated-and-hidden version of `ComboBoxMetrics.Companion.undecorated` — you changed its API by adding the (optional) row count. You'll need to add the deprecated-but-hidden overloads...

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
```diff
@@ -97,14 +222,132 @@ public fun SpeedSearchScope.SpeedSearchableComboBox(
             selected = isSelected,
             active = isActive,
             iconContentDescription = item,
-            onTextLayout = {
-                @Suppress("AssignedValueIsNeverRead")
```
rock3r: Have they fixed this now? God it was so annoying

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/ComboBoxStyling.kt
```diff
@@ -190,7 +190,28 @@ public class ComboBoxMetrics(
...
 ) {
+    @Deprecated("Use the constructor with the maxPopupRowCount parameter")
```
rock3r: Mark as hidden, too, so it doesn't break source compat

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
DanielSouzaBertoldi: oooopppssiiieee i forgor

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
DanielSouzaBertoldi: Looks like it's finally fixed! 🙏🏻 

Image

Also found [this YT ticket](https://youtrack.jetbrains.com/issue/KT-78881/K2-False-positive-Assigned-value-is-never-read-in-composable-function), looks like it was fixed in Kotlin 2.3.0-Beta1 🤯

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBoxTest.kt
```diff
@@ -40,6 +41,7 @@ import org.junit.Rule
 class SpeedSearchableComboBoxTest {
     @get:Rule val rule = createComposeRule()
 
+    @OptIn(ExperimentalCoroutinesApi::class)
```
DanielSouzaBertoldi: Got a warning telling me to add this annotation so I did :)

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBoxTest.kt
```diff
@@ -138,20 +140,20 @@ class SpeedSearchableComboBoxTest {
     @Test
     fun `on arrow up or down, navigate to the next and previous occurrences`() = runComposeTest {
         comboBox.performClick()
-        comboBox.performKeyPress("Item 9", rule = this)
-        onComboBoxItem("Item 9").assertIsDisplayed().assertIsSelected()
+        comboBox.performKeyPress("1", rule = this)
+        onComboBoxItem("Item 1").assertIsDisplayed().assertIsSelected()
```
DanielSouzaBertoldi, line 144: had to update which items we check if are displayed because now a ListComboBox will only render 8 items at a time, so it's physically impossible for users (or the instrumented test for this matter) to click on Item 9 if when it first opens, it's only rendering up until the 8th item.

I opted to verify items that are closer together and that also have the number declared on line 143

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
faogustavo: Actually we should keep this visible as we are removing one parameter (maxPopupHeight) and adding a new one (maxPopupRowCount), which is not a code compatible change. We need to keep both visible and use the "ReplaceWith" to recommend the most up-to-date version.

faogustavo (commented):
Just a minor note on the deprecation cycle. 

If the intended behavior is to replace a parameter (remove one and put another; or rename; or change type; etc), you need to keep both visible (not use deprecation hidden) and make sure you add the "ReplaceWith" pointing to the up-to-date version. As the changes are not "code compatible", we must keep both alive. 

However, if the original intention was to keep the existing params and add new ones with default values, that would be a "code compatible" change, so we could use the DeprecationLevel.Hidden.

DanielSouzaBertoldi:
@faogustavo thank you for clearing that up! My intention is to remove `maxPopupHeight` and just stick to `maxPopupRowCount`. I've already added the `ReplaceWith()` parameter, so in this case we should not add `DeprecationLevel.Hidden` then.

But that means [binary compatibility will break](https://github.com/JetBrains/intellij-community/pull/3384#discussion_r2683546988), AFAIK 🤔

faogustavo:
So, you will need to keep the existing method with 'maxPopupHeight', and keep support for it for a while to prevent this breaking change. An old compilation should still work (and behave as intended) when using the 'maxPopupHeight'.

To drop that property, just create a new method has the parameter, and does not have the maxPopupHeight. This change would not break neither on code or binary compatibility.

We may need to change the popup behavior to support both for a while, until the deprecation flow is completed and we can safely remove it

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -84,7 +84,6 @@ import org.jetbrains.jewel.ui.theme.popupContainerStyle
  * @param popupModifier Modifier to be applied to the popup of the combo box
  * @param enabled Controls whether the combo box can be interacted with
  * @param outline The outline style to be applied to the combo box
- * @param maxPopupHeight The maximum height of the popup list
```
faogustavo, line 87: Something I was discussing with Daniel in a huddle is that maybe we should keep this parameter existing and support both using the maxHeight or maxItems. 

Let's say that the default implementations is set to render 8 items. However, in one of my dropdowns I have some bigger items that take 800.dp to render. The user should still have the freedom to change that specific popup to be at most 300.dp (example).

In other words, we should keep the 'maxPopupHeight'. If that's specified, use it, otherwise, use the number of items. Thoughts?

rock3r: This is a good point, but IIRC list items are forced to be `List.rowHeight`, no?

rock3r: ...so? :D

DanielSouzaBertoldi: So, during my investigation, like I pointed out in the deep dive section in the PR description, IJP's ComboBox does handle items with varying height:

```java
// code snippet from com.intellij.ide.ui.laf.darcula.ui.DarculaComboBoxUI#getPopupHeightForRowCount

// Recalculate the minimum row height of a list cell, which is calculated in WideSelectionListUI.updateLayoutState
int rowCount = Math.min(maxRowCount, comboBox.getItemCount());
ListCellRenderer<Object> renderer = list.getCellRenderer();
for (int i = 0; i < rowCount; i++) {
  Object value = list.getModel().getElementAt(i);
  Component c = renderer.getListCellRendererComponent(list, value, i, false, false);
  Dimension preferredSize = c.getPreferredSize();
  result -= preferredSize.height;
  // vvvvvv IMPORTANT LOC BELOW!!!!!!
  result += UIUtil.updateListRowHeight(preferredSize).height;
}
```

Don't be fooled by the `updateListRowHeight` name. It actually just returns the biggest value between the item's `preferredSize.height` and `List.rowHeight` (24dp).

So, let's say the first item in the list is a simple "hi" and the second one is something like "🚀 hi folks! 🚀". The first item will add 24dp to the popup's height calculation (the default List.rowHeight). The second one actually has a preferred height of 30dp, so it contributes 30dp. The total popup height would be 54dp (not counting insets), not 48dp as it would be if both items were forced to 24dp.

But that is a problem in our Compose code: for us to measure the row height based on the items list, we'd have to first render the items before calculating the popup height, and we can't calculate the items' size without rendering the popup in the first place (chicken-and-egg problem). So we use the simpler approach of assuming all items are `rowHeight` (24dp) for the popup height calculation, even though items can actually render taller than that in the list. Still, I think that this solution will cover the majority of use cases.

Having said that, given the limitation, I'd be okay with keeping `maxPopupHeight`, although that could be a bit confusing for users when deciding which parameters to use. If they want their popup height to fill the entire available space they'd have to code something like:

```kotlin
val metrics = ComboBoxMetrics(maxPopupRowCount = null)

ListComboBox(
     maxPopupHeight = Dp.Unspecified
)
```

Otherwise, supposing the default behaviour after this PR is merged is to always prioritize the calculation given the number of rows, just doing:

```
ListComboBox(maxPopupHeight = Dp.Unspecified)
```

Would not be enough, since it'd fallback to checking the `maxPopupRowCount` value (default 8) and calculate the height based on that

rock3r: Yeah I agree, thanks for the (re)explanation

DanielSouzaBertoldi: No problem! This time I feel like I did a better job at explaining it, in the deep dive section I kinda just brushed over it 😄 

So just to make sure: do we all agree to keep both parameters then? `maxPopupCount` can be nullable but by  default its value will be 8

rock3r: Let's keep both. Maybe we should rename the new one to default instead of max? If we document that it's up to, I think it's pretty intuitive that if there are fewer then it is shorter.

I'm thinking maybe it could make sense to reorganise the new parameter in some way to make it clearer (for one, putting it next to the old one instead of in a different place). But I don't have great ideas yet

morki: What about not supporting both together? I mean two versions, one with old one (deprecated) and one with new one.

rock3r: I don't know if we should deprecate the old way at that point. If the goal is to have the fixed height items for 99% of cases, and to have free height items for the odd use case that needs it (is there any?) then we should probably have two separate component variants, neither deprecated.

@DanielSouzaBertoldi wdyt?

rock3r: Ping

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ListComboBoxUiTest.kt
```diff
@@ -1005,31 +1005,35 @@ class ListComboBoxUiTest {
...
             }
         }
 
         comboBox.assertIsDisplayed().performClick()
         popupMenu.assertIsDisplayed()
 
-        // The popup should be equal to the 'maxPopupHeight' size
-        popupMenu.assertHeightIsEqualTo(500.dp)
+        // Expected height: (5 rows * 24dp (default list item height)) + 8dp top padding + 8dp bottom padding = 136dp
+        val expectedHeight = (24.dp * 5) + 8.dp + 8.dp
```
nebojsa-vuksic: nit: Can we use variables instead of magic numbers here? In that case there is no need for a comment

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
DanielSouzaBertoldi: Sorry for the late reply!

Yeah, having separate component variants seem like less of a headache in the long run. But I'm inclined to update our examples to use this new approach of calculating the height based on the number of rows instead of using an arbitrary height value and in the release notes + kdocs make it clear that this is the new preferred way of using ListComboBox/SpeedSearchComboBox and etc

rock3r: Cool, let's do it then :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
```diff
@@ -22,13 +22,13 @@ public val ComboBoxStyle.Companion.Default: IntUiDefaultComboBoxStyleFactory
 public object IntUiDefaultComboBoxStyleFactory {
     public fun light(
         colors: ComboBoxColors = ComboBoxColors.Default.light(),
-        metrics: ComboBoxMetrics = ComboBoxMetrics.default(),
+        metrics: ComboBoxMetrics = ComboBoxMetrics.default(maxPopupRowCount = 8),
```
DanielSouzaBertoldi: One more thing: since we have a pattern of naming metrics with `default` and now we have one that uses `maxPopupHeight` and another for `maxPopupRowCount` there's overload ambiguity.

Unless we break this pattern and have a function called `defaultWithPopupRowCount`, this is the only way to solve the ambiguity AFAIK :/

The third option is to just have one `.default()` and just add this new parameter. That works as well.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
DanielSouzaBertoldi: Done 😎

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -84,7 +88,95 @@ import org.jetbrains.jewel.ui.theme.popupContainerStyle
...
+public fun <T : Any> ListComboBox(
+    items: List<T>,
+    selectedIndex: Int,
+    onSelectedItemChange: (Int) -> Unit,
+    itemKeys: (Int, T) -> Any,
+    modifier: Modifier = Modifier,
+    popupModifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    outline: Outline = Outline.None,
+    maxPopupWidth: Dp = Dp.Unspecified,
```
DanielSouzaBertoldi: As [discussed here](https://github.com/JetBrains/intellij-community/pull/3384#discussion_r2691192122), now we have variants that don't have the `maxPopupHeight` parameter so they calculate their height based on the `maxPopupRowCount` property from `ComboBoxMetrics` 😄 Users are encouraged to use this over the other variant though, that we decided to not deprecate

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ListComboBoxUiTest.kt
```diff
@@ -1032,6 +1037,37 @@ class ListComboBoxUiTest {
         popupMenu.assertHeightIsEqualTo(500.dp)
     }
 
+    @Test // TODO: rename this test
```
faogustavo: Can you solve or remove this todo? xd

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
```diff
@@ -43,12 +43,135 @@ import org.jetbrains.jewel.ui.component.ProvideSearchMatchState
...
+ * @param textStyle The text style to apply to the combo box label.
+ * @param onPopupVisibleChange Callback invoked when the popup visibility changes.
+ * @param listState The state object for the selectable lazy list within the popup.
+ * @param dispatcher The coroutine dispatcher for speed search operations.
+ */
+@Composable
+@ExperimentalJewelApi
+@ApiStatus.Experimental
+public fun SpeedSearchScope.SpeedSearchableComboBox(
+    @Nls items: List<String>,
```
faogustavo: Is this check working or it needs to be inside the type like `List<@Nls String>`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/ComboBoxStyling.kt
```diff
@@ -190,6 +190,7 @@ public class ComboBoxMetrics(
     public val popupContentPadding: PaddingValues,
     public val borderWidth: Dp,
     public val maxPopupHeight: Dp,
+    public val maxPopupRowCount: Int,
```
faogustavo, line 193: We'll need a new constructor to this type with the old parameters to ensure binary compatibility. 

I'd recommend making the `maxPopupRowCount` nullable or use it only if the value is `> 0` so you can have a default

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ListComboBoxUiTest.kt
DanielSouzaBertoldi: Honestly I forgot why I added this. I don't have any better names for this test lol

Thanks for noticing this!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
DanielSouzaBertoldi: The `@Nls` annotation was added in this PR: https://github.com/JetBrains/intellij-community/pull/3331

Looking at the [documentation](https://plugins.jetbrains.com/docs/intellij/internationalization.html#property-keys), seems like the annotation should be inside the type like you mentioned:

```kotlin
// Example taken from the docs

fun lazyMessage(
      @PropertyKey(resourceBundle = BUNDLE) key: String,
      vararg params: Any
  ): Supplier<@Nls String>
```

And also judging from usages in another classes:

image

The [annotation itself targets](https://github.com/JetBrains/java-annotations/blob/24.0.0/common/src/main/java/org/jetbrains/annotations/Nls.java) `TYPE`, `TYPE_USE` and `PARAMETER`, so from what I understand, we should probably be adding the annotation to the type, not the variable.

I'm not familiar with the `@Nls` annotation but now that you mentioned it, to me having it scoped to the parameter is kinda like checking the variable (the container) itself, not the items.

So yeah, I think it's more precise if we target the contents of the list (each individual String), not the variable.

Great catch! 🚀

rock3r: We missed this in that PR's review, but yes it should be `List<@Nls String>` :)

rock3r: (it means "this is a UI string and should be treated as such" for IJ tooling)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/ComboBoxStyling.kt
DanielSouzaBertoldi: I don't think we need to make `maxPopupRowCount` nullable, since now we have different component variants. I think it's safe to have the "new-old" constructor adding the default value of 8. Developers that are already using the `maxPopupHeight` won't be affected by these changes 🤔

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeComboBox.kt
```diff
@@ -65,6 +65,7 @@ internal fun readDefaultComboBoxStyle(): ComboBoxStyle {
                     ),
                 borderWidth = borderWidth,
                 maxPopupHeight = Dp.Unspecified,
+                maxPopupRowCount = 8,
```
DanielSouzaBertoldi: Unfortunately `JBUI` doesn't expose the `maximumRowCount` property from `ComboBox`, so we are forced to hardcode it in Bridge :(

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
DanielSouzaBertoldi: I updated all occurrences in all affected ComboBoxes in this PR to `List<@Nls String>`, thanks guys!

rock3r:
I was looking at `DarculaJBPopupComboPopup.java`, `ComboBoxPopup.Context`, etc — they seem to use 10 as max items, overriding the 8 used by `JComboBox`. Could you double check?

rock3r:
Update: I think that applies to "custom popup" combo boxes only — see `com.intellij.openapi.ui.ComboBox#setSwingPopup`. Essentially, when there is a custom popup (off by default) the height is 10, otherwise it's 8.

They use it for cases like filterable/speed searchable combo boxes. A (somewhat uncommon) such case:

https://github.com/user-attachments/assets/ea583118-a225-420d-80cb-a89354d2d9af

But in this case it looks like the default height is not respected, since it shows more than 10 items...

DanielSouzaBertoldi:
@rock3r  Hmmmmm yeah I did take a look at some of them and noticed that sometimes they'd set 10 instead of 8, but I opted to follow the basic `DarculaComboBoxUI` that just uses the default from JComboBox.

Now, I did not tested the speed searchable combo boxes like you've shown in your video 🤔 that certainly is odd, I'll look around and see what's going on with this guy.

For now, I'll mark this PR as draft until I get an idea of what's happening (this way the PR bot won't keep bugging you guys for a review)

rock3r:
The rabbit hole keeps growing 🫠 

We don't need to support those speed searching/filtering thingies to be clear!

DanielSouzaBertoldi:
@rock3r So after spending the past 3 hours trying to debug this I've came to the conclusion that this behaviour does not make any sense 😬 

I could go over all the code that I found that handles the popup creation and list rendering but I think I can sum it up with this:

`ComboBoxPopup.java`

```java
private void configurePopup() {
    setMaxRowCount(myContext.getMaximumRowCount());
    // .... rest of code
}
```

The `maximumRowCount()` for comboboxes that are created by `DarculaJBPopupComboPopup.java` will always be 10 for the font combobox since it does not set another value for `maximumRowCount`. `DarculaJBPopupComboPopup` has a check to always list at least 10 items in a list:

```java
@Override
public int getMaximumRowCount() {
  return Math.max(10, myComboBox.getMaximumRowCount());
}
```

The only combobox that changes the default `maximumRowCount` value is `JComboboxAction` (it expects 12 rows), but the combobox used to set the font is `FontFamilyCombo`.

In other words, the font combobox should not be rendering 12 items. It simply does not make sense to do so 🤯 

I've found that if I comment out the code that adds separator to `FontFamilyCombo.java`:

```java
@Override
public @Nullable ListSeparator separatorFor(MyFontItem value) {
  if (getModel() instanceof MyModel m) {
    if (!m.myItems.isEmpty() && ContainerUtil.find(m.myItems, item -> item.myIsMonospaced) == value)
      return new ListSeparator(ApplicationBundle.message("settings.editor.font.monospaced"));
    if (!m.myItems.isEmpty() && ContainerUtil.find(m.myItems, item -> !item.myIsMonospaced && !(item instanceof MyNoFontItem)) == value)
      return new ListSeparator(ApplicationBundle.message("settings.editor.font.proportional"));
  }
  return null;
}
```

Then the number of rendered items in the list is reduced to 11, not 12, which is really weird because the separator is just another item in the list, the `maximumRowCount` doesn't change if a list has separators or not. I've no idea what might be causing the 11th item to appear.

Everything suggests that this might be a bug on the IJ side, but I highly doubt it.

This really is weird 🤯

rock3r:
I had seen those bits of code and have given up trying to make sense of them. I have no clue how anything works anymore at this point... I'm as confused as you are 🫠 

Wanna try checking git blame and pinging the right Swing people to get an explanation? Yuriy or Dejan can help putting you in touch

DanielSouzaBertoldi:
@rock3r I asked around but unfortunately didn't reach a satisfying conclusion. Combo box code is relatively old and folks that did work on it don't know what's happening either 🤷🏻 so, we can create a compromise for now and just up the default row count to 10 :/

Reopening this PR for now.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
rock3r: I think this new overload is a silent source-compatibility break for positional call sites. Before, a call such as `ListComboBox(..., Outline.None, 200.dp)` bound that `200.dp` to `maxPopupHeight`; with this overload in place it can now bind to `maxPopupWidth` instead and still compile with different behavior. Could we avoid introducing overloads that shift parameters in an existing position? I think the same issue applies to the analogous overloads added below and in `SpeedSearchableComboBox`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -857,6 +1033,134 @@ internal fun <T : Any> ListComboBoxImpl(
...
+    }
+
+    val contentPadding = style.metrics.popupContentPadding
+
+    EditableComboBox(
+        textFieldState = textFieldState,
+        modifier =
+            modifier.onSizeChanged { comboBoxSize = with(density) { DpSize(it.width.toDp(), it.height.toDp()) } },
+        popupModifier = popupModifier,
+        maxPopupHeight = maxPopupHeight,
```
rock3r, line 855: I think the editable path lost the old `style.metrics.maxPopupHeight` fallback during the extraction. On `master`, `EditableListComboBox` first did `maxPopupHeight.takeOrElse { style.metrics.maxPopupHeight }` before calling `EditableComboBox`. Here the old overload forwards raw `maxPopupHeight`, and `EditableComboBox` uses it directly in `heightIn(max = maxPopupHeight)`, so a style-level `metrics.maxPopupHeight` is ignored unless the caller also passes `maxPopupHeight` explicitly. Was that intended? Non-editable still keeps the fallback via `ListComboBoxImpl`.

rock3r (changes requested):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
DanielSouzaBertoldi: Damn. That totally slipped past me :( yes. it is a source breaking change. Will fix it ASAP, thanks!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
DanielSouzaBertoldi: Well, actually is the other way around: (edit: i guess this would be a good place to add a — or –)  I forgot to remove the `.takeOrElse` from the `ListComboBoxImpl`.

See, with my changes to the code, the fallback case when `maxPopupHeight` is `Dp.Unspecified` is to calculate the popup height, mirroring IJP's behavior.

For `ListComboBox` and `EditableListComboBox`, the `metrics.maxPopupHeight` is useless now.

The only place where falling back to `metrics.maxPopupHeight` would make sense now is in our "barebones" `ComboBox` that takes a `popupContent: @Composable (() -> Unit)`. Due to its nature its up to the developer to play around and list whatever it wants to in the popup, there are no guarantees that they'll display a list.

Now, I'm in favor of falling back to `calculatePopupHeight`. If users want to have a specific height to their popup, then they can do so by just passing a value to the `maxPopupHeight` parameter directly.

Hooowweeever, yes, if this PR is merged as is, then we will be introducing a behavioral breaking change. One that, to me, it's okay and should not be that big of a deal.

DanielSouzaBertoldi: Oh, also, I increased the default max popup row count to 10, as a compromise

rock3r: Ok, we can proceed with this as long as the change is noted in the release notes

DanielSouzaBertoldi:
@rock3r Release notes updated 😎

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
rock3r: This looks like a leftover oversight? The bridge style now hardcodes `maxPopupRowCount = 10`, while the standalone default/undecorated style factories still force `8`, so the default popup height differs between bridge and standalone despite our discussion converging on `10` as the compromise.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
```diff
@@ -205,7 +205,27 @@ public fun ComboBoxMetrics.Companion.default(
     contentPadding: PaddingValues = PaddingValues(9.dp, 1.dp, 6.dp, 1.dp),
     popupContentPadding: PaddingValues = PaddingValues(vertical = 8.dp),
     borderWidth: Dp = 1.dp,
```
rock3r, line 207: I think this still creates a source-breaking overload ambiguity for common calls such as `ComboBoxMetrics.default()` / `undecorated()`, because both overloads have all-default parameters.

Since we've already debated whether the old and new styles should coexist, I think this needs to be one of two explicit shapes:

1. keep a single visible `default`/`undecorated` API that includes the new row-count parameter, and keep the old ABI-only overload as `DeprecationLevel.HIDDEN`; or
2. keep the old visible height-based overload as-is, but give the new row-count-based factory a distinct name, then update Jewel's own defaults/usages to prefer that new factory.

Slight personal preference for option 2, since it makes the two sizing modes harder to confuse, but either would avoid the source break. Up to you.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableComboBox.kt
```diff
@@ -68,7 +70,8 @@ import org.jetbrains.jewel.ui.theme.popupContainerStyle
  * @param enabled Whether the combo box is interactive.
  * @param outline The [Outline] to draw around the combo box.
  * @param maxPopupHeight The maximum height of the popup. Defaults to [Dp.Unspecified] (unconstrained).
```
rock3r: Nit: this overload now has two `@param maxPopupHeight` entries. I think the first one is stale and we should restore `maxPopupWidth`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
```diff
@@ -933,3 +998,13 @@ private fun <T : Any> PopupContent(
 }
 
 internal fun Int.takeIfInBoundsOrZero(acceptedIndices: IntRange) = if (this in acceptedIndices) this else 0
+
+@Composable
+internal fun calculatePopupHeight(itemCount: Int, maxPopupRowCount: Int, popupContentPadding: PaddingValues): Dp {
```
rock3r, line 1003: Should we clamp or validate `maxPopupRowCount` before using it here? Since it feeds directly into the height calculation, values like `0` or negative numbers can produce a popup with only padding or a nonsensical height. I think clamping to at least `1` here would be enough, unless you'd rather validate in the metrics/factory path.

rock3r (changes requested):
I left a few comments, mostly around keeping bridge/standalone behavior aligned and avoiding source-compatibility traps.

Could we also move/expand the release note into `Important Changes`? This is not only a new preferred API shape: the default popup sizing behaviour changes, and `ComboBoxMetrics.maxPopupHeight` is no longer the fallback used by these list combo wrappers when their own `maxPopupHeight` is unspecified.

I think the note should explicitly call out the new row-count-based default, the chosen default row count, and that callers who need fixed-height popups should keep passing `maxPopupHeight` at the component level.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ListComboBox.kt
DanielSouzaBertoldi: Validating at the ComboBoxMetrics class seems more correct IMO, thanks!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiComboBoxStyling.kt
DanielSouzaBertoldi: Great catch! Did not realize this ambiguity breaking change.

Given that we decided to let users choose between deciding how many rows the popup should show or explicitly set a popup height, option 2 is more aligned with this decision.

The downside is a quirky name for our API

rock3r: Our library has all the quirks & features

rock3r (approved):
LGTM, but needs updating the PR description with the correct number (10, not 8)

nebojsa-vuksic (approved):

rock3r:
Ready to merge

PR #3386, opened 2026-01-14
Add standard patch format & full commit message options to `Create Patch...`

l0rinc:
When I’m prototyping changes during GitHub code review, I often want to share results as a patch.
The current "Create Patch" output is hard to share externally because it can include IntelliJ-specific headers (Index: + IDEA additional info: blocks), and commit messages are commonly reduced to a single subject line instead of the full multi-line
message.

#### Fix
This change adds two independent, persisted options to the "Patch File Settings" dialog:

##### Standard patch format
* New checkbox: "Use standard patch format (no IDEA additional info)"
* When enabled, patch output omits IntelliJ-specific patch headers (Index:, IDEA additional info:, and the separator) and keeps a standard unified diff format.

##### Include full commit message
* New checkbox: "Include full commit message"
* When enabled (and when a full message is available from VCS log / file history selection), the patch Subject: [PATCH] … header uses the full multi-line commit message instead of only the subject/preset message.

### Validation
* Automated coverage updated for patch headers and commit message header behavior.
* Manually validated all scenarios (both options on/off, full-message present/absent, and patch creation flows), confirming output matches expectations.

I haven't pushed here in at least 6 years, don't remember what the flow is, but feel free to merge and edit if that's simpler than requesting changes.

### DIFF
--- a/platform/vcs-log/impl/src/com/intellij/vcs/log/ui/actions/VcsLogCreatePatchActionProvider.java
+++ b/platform/vcs-log/impl/src/com/intellij/vcs/log/ui/actions/VcsLogCreatePatchActionProvider.java
@@ -4,9 +4,14 @@ package com.intellij.vcs.log.ui.actions;
 import com.intellij.openapi.actionSystem.ActionUpdateThread;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.AnActionExtensionProvider;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectManager;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vcs.VcsDataKeys;
 import com.intellij.openapi.vcs.changes.Change;
+import com.intellij.openapi.vcs.changes.CommitContext;
 import com.intellij.openapi.vcs.changes.actions.CreatePatchFromChangesAction;
+import com.intellij.openapi.vcs.changes.patch.PatchWriter;
 import com.intellij.openapi.vcs.changes.ui.ChangesBrowserBase;
 import com.intellij.vcs.log.VcsLogCommitSelection;
 import com.intellij.vcs.log.VcsLogDataKeys;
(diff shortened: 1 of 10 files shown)


BartvHelvert:
Hi, thanks for the PR, could you create a feature request in our [issue tracker](https://youtrack.jetbrains.com/issues/IJPL) for it?

l0rinc:
Thanks, opened https://youtrack.jetbrains.com/issue/IJPL-230383/Add-standard-patch-format-full-commit-message-options-to-Create-Patch

iasemenov:
Hey @l0rinc 
Thanks a lot for the contribution!
Could you please provide an example of a situation where having some custom patch headers becomes an issue?

platform/vcs-tests/testSrc/com/intellij/openapi/vcs/diff/impl/patch/CreateApplyPatchSymmetryTest.kt
```diff
@@ -95,4 +99,41 @@ class CreateApplyPatchSymmetryTest : HeavyDiffTestCase() {
 
     assertEquals(text2, text3)
   }
+
+  fun testCommitMessageHeader() {
```
iasemenov, line 103: This doesn't seem like a correct place for this test

platform/vcs-log/impl/src/com/intellij/vcs/log/ui/actions/VcsLogCreatePatchActionProvider.java
```diff
@@ -83,7 +88,11 @@ public void actionPerformed(@NotNull AnActionEvent e) {
 
     selection.requestFullDetails(details -> {
       List<Change> changes = VcsLogUtil.collectChanges(details);
-      CreatePatchFromChangesAction.createPatch(e.getProject(), commitMessage, changes, mySilentClipboard);
+      CommitContext commitContext = new CommitContext();
+      commitContext.putUserData(PatchWriter.FULL_COMMIT_MESSAGE_KEY, StringUtil.join(details, it -> it.getFullMessage(), "\n\n"));
+      Project project = e.getProject();
+      if (project == null) project = ProjectManager.getInstance().getDefaultProject();
+      CreatePatchFromChangesAction.createPatch(project, commitMessage, changes, mySilentClipboard, commitContext);
```
iasemenov, line 95: From what I can see, `VcsDataKeys.PRESET_COMMIT_MESSAGE` is always `null` in the case of changes selection (previous `if`), and here we don't really need to use it at all. I would propose dropping this data key altogether and always using either `com.intellij.vcs.log.VcsCommitMetadata.getFullMessage` or `com.intellij.vcs.log.VcsShortCommitDetails.getSubject` to build the commit message.

Also it might be better to just avoid the checkbox configuration altogether and use the following logic:
* if a single commit is selected - use the full message of the commit
* if multiple commits are selected - join the subjects like it is done in `com.intellij.vcs.log.ui.table.VcsLogGraphTable.uiDataSnapshot`

This reduces the variability of the implementation and keeps the UI cleaner.

As a followup, I actually think that we might want to create a "cumulative" patch the way e.g. GitHub does it, where a separate patch is created for each commit, and then they all are concatenated together into a single file.

iasemenov (changes requested):

l0rinc:
> Could you please provide an example of a situation where having some custom patch headers becomes an issue?

When I'm reviewing code in Bitcoin Core I usually apply the suggestions locally, and when I'm done and have a unified direction, I post those commits as patches. I am currently copying the hash and doing `git show <hash> | pbcopy` instead of just right-clicking, since IDEA ignores most of the commit message and adds a spammy IntelliJ advertisement at the beginning.

PR #3387, opened 2026-01-16
[JEWEL-1138] Scale images to fit available width

AlexVanGogen:
## Description

This is achieved by splitting paragraphs and headings by images and rendering them in a FlowRow as separate composables. Paragraphs and headings without images are rendered as a single Text composable, like before.

There are two notable changes:
  - `ScalableImage` is introduced. This is an `Image` composed in a custom `Layout` that takes care of adjusting height according to the current downscaling factor.
  - Text with images is split into separate composables (see above).

Pros:
  - It doesn't increase complexity and doesn't affect performance for previously supported cases: text without images, images without text.
  - It doesn't add additional subcompositions.

Cons:
  - If text surrounds an image from both sides, this text will be copied with a line break in place of the image. It can break

Possible alternatives:
  - Wrap an image in a `BoxWithConstraints`. It would limit Markdown extensions rendering a lot, as it prohibits possible containers to measure themselves by using children's intrinsics. Note that `FlowRow` doesn't cause such a problem because it defines custom intrinsic measurement (see `FlowMeasurePolicy`).
  - Custom `SubcomposeLayout` with a custom intrinsic measurement instead of a `FlowRow`. `FlowRow` downscales an image properly, but it keeps the reserved vertical space, so other content is rendered far enough as if the image were unscaled. While theoretically possible, that would be too challenging to implement, and maybe not immensely required for our case -- images right in-between text shouldn't be a common practice.

## Examples

https://github.com/user-attachments/assets/c8fff00c-2f9b-4e9d-b746-81e2309ade26

https://github.com/user-attachments/assets/a398c508-e8a8-453b-8044-fa59e84ddc49

https://github.com/user-attachments/assets/d35ff9d1-efd0-4cf0-8256-082bcc615cfa

### Bug fixes
 * **JEWEL-1138** Scale images to fit available width and adjust vertical space

---

> [!NOTE]
> **Medium Risk**
> Changes core Compose text/layout measurement to dynamically scale inline image placeholders, which can affect rendering and intrinsic sizing in tables and other constrained containers. Adds/rewires test infrastructure and new UI tests to validate sizing behavior across many edge cases.
> 
> **Overview**
> Adds automatic *downscaling* for inline markdown images when the available width is smaller than the image placeholder, so images fit their container and vertical space matches the scaled size.
> 
> `DefaultMarkdownBlockRenderer` now routes paragraph/heading inline rendering through a new `TextWithScalableInlineContent` path that uses `SubcomposeLayout` plus custom intrinsics (with a fast path for no/zero-width images), and exposes `DOWNSCALED_INLINE_CONTENT_TAG` for testing.
> 
> Build/test wiring is updated to introduce `core` test fixtures (`java-test-fixtures`, Bazel/IML src roots) and share a common `TestTheme`; the images extension adds a dependency on these fixtures and expands its Coil tests to cover downscaling, non-upscaling, intrinsic-width behavior, and edge cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e8e21f24730fd1549e1bd937aed02741fce99e21. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/platform/jewel/markdown/core/BUILD.bazel
+++ b/platform/jewel/markdown/core/BUILD.bazel
@@ -105,7 +105,7 @@ jvm_library(
   name = "jewel-markdown-core-tests_test_lib",
   testonly = True,
   visibility = ["//visibility:public"],
-  srcs = glob(["src/test/kotlin/**/*.kt", "src/test/kotlin/**/*.java", "src/test/kotlin/**/*.form"], allow_empty = True),
+  srcs = glob(["src/test/kotlin/**/*.kt", "src/test/kotlin/**/*.java", "src/test/kotlin/**/*.form", "src/testFixtures/kotlin/**/*.kt", "src/testFixtures/kotlin/**/*.java", "src/testFixtures/kotlin/**/*.form"], allow_empty = True),
   resources = [":jewel-markdown-core-tests_test_resources"],
   kotlinc_opts = ":custom_jewel-markdown-core-tests",
   associates = [
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
@@ -5,13 +5,11 @@ import androidx.compose.foundation.border
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.FlowRow
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.text.InlineTextContent
@@ -36,15 +34,36 @@ import androidx.compose.ui.graphics.isSpecified
 import androidx.compose.ui.graphics.takeOrElse
 import androidx.compose.ui.input.pointer.PointerIcon
 import androidx.compose.ui.input.pointer.pointerHoverIcon
+import androidx.compose.ui.layout.IntrinsicMeasurable
+import androidx.compose.ui.layout.IntrinsicMeasureScope
+import androidx.compose.ui.layout.Layout
+import androidx.compose.ui.layout.Measurable
+import androidx.compose.ui.layout.MeasurePolicy
+import androidx.compose.ui.layout.MeasureResult
+import androidx.compose.ui.layout.MeasureScope
+import androidx.compose.ui.layout.SubcomposeLayout
 import androidx.compose.ui.layout.onFirstVisible
 import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.LocalFontFamilyResolver
+import androidx.compose.ui.platform.LocalLayoutDirection
+import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.ParagraphIntrinsics
+import androidx.compose.ui.text.Placeholder
 import androidx.compose.ui.text.TextLayoutResult
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.font.FontFamily
+import androidx.compose.ui.text.rememberTextMeasurer
+import androidx.compose.ui.text.resolveDefaults
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.Constraints
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.unit.LayoutDirection.Ltr
 import androidx.compose.ui.unit.dp
+import kotlin.math.ceil
 import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.foundation.code.MimeType
@@ -78,6 +97,7 @@ import org.jetbrains.jewel.ui.component.HorizontallyScrollableContainer
 import org.jetbrains.jewel.ui.component.Text
 
 private const val DISABLED_CODE_ALPHA = .5f
+private const val COMPOSE_INLINE_CONTENT_ANNOTATION_TAG = "androidx.compose.foundation.text.inlineContent"
 private val MIME_TYPE_REGEX = "^\\w+/.+$".toRegex()
 
 /**
@@ -145,6 +165,7 @@ public open class DefaultMarkdownBlockRenderer(
             ThematicBreak -> RenderThematicBreak(rootStyling.thematicBreak, enabled, modifier)
             is MarkdownBlock.HtmlBlockWithAttributes ->
                 RenderHtmlBlockWithAttributes(block, enabled, onUrlClick, modifier)
+
             is CustomBlock -> {
                 rendererExtensions
                     .find { it.blockRenderer?.canRender(block) == true }
@@ -206,31 +227,50 @@ public open class DefaultMarkdownBlockRenderer(
         softWrap: Boolean,
         maxLines: Int,
     ) {
-        val onlyImages = remember(block) { block.inlineContent.all { it is InlineMarkdown.Image } }
-        val images = renderedImages(block)
+        RenderBlockWithInlines(
+            block,
+            styling.inlinesStyling,
+            enabled,
+            onUrlClick,
+            onTextLayout,
+            modifier,
+            overflow,
+            softWrap,
+            maxLines,
+        )
+    }
 
-        if (onlyImages) {
-            RenderImages(images, modifier)
-        } else {
-            val renderedContent = rememberRenderedContent(block, styling.inlinesStyling, enabled, onUrlClick)
-            val textColor =
-                styling.inlinesStyling.textStyle.color
-                    .takeOrElse { LocalContentColor.current }
-                    .takeOrElse { styling.inlinesStyling.textStyle.color }
-            val mergedStyle = styling.inlinesStyling.textStyle.merge(TextStyle(color = textColor))
+    @Composable
+    private fun RenderBlockWithInlines(
+        block: WithInlineMarkdown,
+        inlinesStyling: InlinesStyling,
+        enabled: Boolean,
+        onUrlClick: (String) -> Unit,
+        onTextLayout: (TextLayoutResult) -> Unit,
+        modifier: Modifier = Modifier,
+        overflow: TextOverflow = TextOverflow.Clip,
+        softWrap: Boolean = true,
+        maxLines: Int = Int.MAX_VALUE,
+    ) {
+        val originalImages = renderedImages(block)
 
-            Text(
-                modifier = modifier,
-                text = renderedContent,
-                overflow = overflow,
-                softWrap = softWrap,
-                maxLines = maxLines,
-                onTextLayout = onTextLayout,
-                inlineContent = images,
-                style = mergedStyle,
-                textAlign = LocalTextAlignment.current,
-            )
-        }
+        val renderedContent = rememberRenderedContent(block, inlinesStyling, enabled, onUrlClick)
+        val textColor = inlinesStyling.textStyle.color.takeOrElse { LocalContentColor.current }
+        val mergedStyle = inlinesStyling.textStyle.merge(TextStyle(color = textColor))
+        val density = LocalDensity.current
+
+        TextWithScalableInlineContent(
+            text = renderedContent,
+            inlineContent = originalImages,
+            density = density,
+            modifier = modifier,
+            overflow = overflow,
+            softWrap = softWrap,
+            maxLines = maxLines,
+            onTextLayout = onTextLayout,
+            style = mergedStyle,
+            textAlign = LocalTextAlignment.current,
+        )
     }
 
     @Composable
@@ -284,30 +324,15 @@ public open class DefaultMarkdownBlockRenderer(
         onUrlClick: (String) -> Unit,
         modifier: Modifier,
     ) {
-        val onlyImages = remember(block) { block.inlineContent.all { it is InlineMarkdown.Image } }
-        val images = renderedImages(block)
-
-        if (onlyImages && images.isEmpty()) return
-
         Column(modifier = modifier.padding(styling.padding)) {
-            if (onlyImages) {
-                RenderImages(images)
-            } else {
-                val renderedContent = rememberRenderedContent(block, styling.inlinesStyling, enabled, onUrlClick)
-
-                val textColor =
-                    styling.inlinesStyling.textStyle.color.takeOrElse {
-                        LocalContentColor.current.takeOrElse { styling.inlinesStyling.textStyle.color }
-                    }
-                val mergedStyle = styling.inlinesStyling.textStyle.merge(TextStyle(color = textColor))
-                Text(
-                    text = renderedContent,
-                    style = mergedStyle,
-                    modifier = Modifier.focusProperties { this.canFocus = false },
-                    inlineContent = images,
-                    textAlign = LocalTextAlignment.current,
-                )
-            }
+            RenderBlockWithInlines(
+                block,
+                styling.inlinesStyling,
+                enabled,
+                onUrlClick,
+                {},
+                Modifier.focusProperties { this.canFocus = false },
+            )
 
             if (styling.underlineWidth > 0.dp && styling.underlineColor.isSpecified) {
                 Spacer(Modifier.height(styling.underlineGap))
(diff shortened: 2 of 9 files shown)


AlexVanGogen:
I totally need a keen eye of a Compose pro here :)

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -203,31 +205,98 @@ public open class DefaultMarkdownBlockRenderer(
...
-        val images = renderedImages(block)
-
-        if (onlyImages) {
-            RenderImages(images, modifier)
-        } else {
-            val renderedContent = rememberRenderedContent(block, styling.inlinesStyling, enabled, onUrlClick)
-            val textColor =
-                styling.inlinesStyling.textStyle.color
-                    .takeOrElse { LocalContentColor.current }
-                    .takeOrElse { styling.inlinesStyling.textStyle.color }
```
AlexVanGogen, line 219: Note: I removed the last `takeOrElse` during refactoring as it repeats the first picked color and this shouldn't change the result

rock3r:
Doesn't using a `FlowRow` breaks completely when you have longer runs of text in between images? E.g.:

```markdown
This is a paragraph and it contains an image or two, but there is also a lot of text. ![](...) That was the first image! So exciting. It shows a couple of capybaras swimming in a pool. It might be AI generated. ![](...) Ah, this second image shows clearly that there is a Gemini logo in the bottom right corner, so it's definitely a Nano Banana image.
```

AlexVanGogen:
@rock3r True, the text will be placed on the next line. I see no other way to make it work besides providing a custom `SubcomposeLayout` over text with a custom `MeasurePolicy` that calculates intrinsics (otherwise we'd be forcing extensions to refrain from getting layout children intrinsic sizes)...

rock3r:
Note: I suggested (offline) looking into using https://github.com/romainguy/combo-breaker for this

AlexVanGogen:
Came up with a different approach, inspired by combo-breaker. It's not the same thing though AFAIU (combo-breaker accomodates text around fixed-size images, here we need to adjust images so they fit). Yet, I try to compute all the necessary information during measurement phase.

https://github.com/user-attachments/assets/27fed377-8eec-43b3-b30b-97859b9a28b7

AlexVanGogen:
Note: I removed `FlowRow` to render paragraphs that only contain images, as probably we should render them as links instead. That's what GitHub does, for example:

this link does not exist.

I can bring it back if needed. cc @faogustavo 

The issue is due

faogustavo (commented):
I'm not sure if that's related to your changes, but when there are multiple images in the same row, resizing to reduce and then return works. But when there is only one per line, it doesn't.

https://github.com/user-attachments/assets/e0357353-e3c9-4f9f-8940-709e32361d3e

AlexVanGogen:
@faogustavo Thanks for checking it! That's a very interesting behavior, seems like `Modifier.onSizeChanged` is not triggered whenever the horizontal space is expanded, but ONLY when the text is just a single inline. It could some low-level optimization in Compose that just composes this `inlineTextContent.children()` instead of a `Text`. 🤔

AlexVanGogen:
Updating the state right during the measurement helps with the issue. But, there's a caveat: the state is updated during the measurement. Not sure there's another way to scale the placeholder properly...

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,23 +836,149 @@ public open class DefaultMarkdownBlockRenderer(
         }
     }
 
+    /** Scales the inline content placeholders by the given scale factor. */
+    private fun scaleInlineContent(
```
DanielSouzaBertoldi: Pretty minor nit: could we move this function after the `TextWithScalableInlineContent` composable? It's last used there so it "reads" better than being on top

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
```diff
@@ -123,6 +134,175 @@ public class Coil3ImageRendererExtensionImplTest {
...
+            .assertWidthIsEqualTo(containerWidth.dp)
+            .assertHeightIsEqualTo(expectedHeight1.dp)
+
+        composeTestRule
+            .onNodeWithContentDescription("Second image")
+            .assertExists()
+            .assertWidthIsEqualTo(containerWidth.dp)
+            .assertHeightIsEqualTo(expectedHeight2.dp)
+    }
+
```
DanielSouzaBertoldi, line 504: There are some missing test scenarios I can think of, let me know if you agree!

1. Test if `availableWidth >= width` (so, `fakeImageWidth = 100` and `container = 100`) or `if width == 0` (from this if statement in your `scaleInlineContentFunction`: `if (width == 0 || availableWidth >= width)`
2. What if `fakeImage1` can fit but `fakeImage2` can't? Then we need to assert that `fakeImage1` has its initial properties while `fakeImage2` has been downscaled
3. Test the optimization from `placeholderWidths.sum() <= 0.01f`
4. This might be pushing it a bit but what if the `fakeImageWidth` is just barely over the `container` width? Like, 101 to 100. Then we can test that there are no rounding errors, possibly

DanielSouzaBertoldi (commented):
Sorry for taking too long to review your code. I was so focused on the LazyTable PR that I totally forgot about this one, my bad 😭 

Also: can't think of any other way to measure the space each component should fit if not by biting the bullet and using `Subcomposelayout`, even though it has drawbacks such as accessibility issues as you described in your commit description.

The only other alternative would be to ask Compose devs to implement some sort of dynamic placeholder or something that we can use to describe how to compute the size of it, kinda like what `BoxWithConstraints` does

But yeah, that too would be too far-fetched unfortunately

One final note: AWESOME image choice in the examples section 😎

AlexVanGogen:
@DanielSouzaBertoldi Thank you for the review! I don't think dynamic placeholders will ever be supported, as it has a very niche application (and mixing text and composables is IMO not a good practice for mobile UX?) and could be very costly.

As for `SubcomposeLayout`, yes, it's an alternative. I think we could combine the two approaches, and move downscaling to the subcomposition phase, when we have the incoming constraints. It also implies providing all the intrinsics for the text, but the entire process should work in one composition phase. WDYT?

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
AlexVanGogen: Those are good suggestions! I'm not sure about your first case when the image width is the same as the container width, as I don't think the behavior matters here much: even if the composition goes the slow path, I suppose it should rarely happen in practice when the image width is exactly the same as the width of its parent. And the case when `availableWidth > fakeImageWidth` is covered already.

And I'd probably also check that we do take the fast path if the image fits the space naturally.

DanielSouzaBertoldi:
@AlexVanGogen yeah great plan! I think that's a nice compromise given the limitations we have 😄

AlexVanGogen:
Replaced the manual state change with subcomposition + intrinsics calculation. All the previous video examples are still valid!

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
AlexVanGogen: Added a few tests:
- paragraph with a 0 width image goes fast path
- one image fits and the other does not
- image doesn't fit just a tiny bit

Also, added a test tag to verify whether the composition goes fast path (`Text`) or slow path (`SubcomposeLayout`)

DanielSouzaBertoldi (approved):
Really really great work!!! 🚀🚀🚀🚀🚀

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,25 +836,209 @@ public open class DefaultMarkdownBlockRenderer(
...
 
-        val density = LocalDensity.current
-        FlowRow(modifier) {
-            images.map { (text, inlineBlock) ->
-                Box(
-                    modifier =
-                        with(density) {
-                            Modifier.size(inlineBlock.placeholder.width.toDp(), inlineBlock.placeholder.height.toDp())
+        // Fast path: no images or only zero-width images - just render regular Text
+        if (placeholderWidths.sum() <= 0.01f) {
```
rock3r, line 874: The fast path here renders plain `Text(...)` without forwarding `modifier` or `inlineContent`. That seems to cause two regressions: text-only blocks/headings lose caller modifiers (for example the heading `focusProperties`), and zero-width/loading images can fall back to the raw markdown alt text because the inline placeholders are no longer provided. Should this call pass both through?

platform/jewel/markdown/core/BUILD.bazel
```diff
@@ -30,7 +30,7 @@ jvm_library(
   name = "core",
   module_name = "intellij.platform.jewel.markdown.core",
   visibility = ["//visibility:public"],
-  srcs = glob(["src/main/kotlin/**/*.kt", "src/main/kotlin/**/*.java", "src/main/kotlin/**/*.form"], allow_empty = True),
+  srcs = glob(["src/main/kotlin/**/*.kt", "src/main/kotlin/**/*.java", "src/main/kotlin/**/*.form", "src/testFixtures/kotlin/**/*.kt", "src/testFixtures/kotlin/**/*.java", "src/testFixtures/kotlin/**/*.form"], allow_empty = True),
```
rock3r: This now pulls `src/testFixtures` into the main `core` target. Together with the `.iml` / `api-dump` changes, it looks like `TestTheme.kt` becomes part of the production/public API surface. Was that intentional? I would expect the shared test helpers to stay test-only instead of shipping in `markdown.core`.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,25 +836,209 @@ public open class DefaultMarkdownBlockRenderer(
...
+                        val mergedStyle = style.merge(TextStyle(textAlign = textAlign))
+                        val constraints =
+                            if (width == Constraints.Infinity) {
+                                Constraints()
+                            } else {
+                                Constraints(maxWidth = width)
+                            }
+
+                        val measuredText =
+                            textMeasurer.measure(
```
rock3r, line 989: Should the intrinsic-height measurement mirror the actual `Text(...)` params here? Right now `computeHeightForWidth()` ignores `softWrap`, `maxLines`, and `overflow`, so callers like `maxLines = 1` or `softWrap = false` can get an intrinsic height that is taller than the real layout.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,25 +836,209 @@ public open class DefaultMarkdownBlockRenderer(
...
-                ) {
-                    inlineBlock.children(text)
+                    }
+
+                    override fun IntrinsicMeasureScope.maxIntrinsicWidth(
+                        measurables: List<IntrinsicMeasurable>,
+                        height: Int,
+                    ): Int = intrinsicWidth
+
+                    override fun IntrinsicMeasureScope.minIntrinsicWidth(
```
rock3r, line 960: I’m a bit worried about hardcoding `minIntrinsicWidth()` to `0` for mixed text+image content. Images can shrink to zero, but the text still has a non-zero minimum intrinsic width. Could this under-report width to parents that rely on `IntrinsicSize.Min` / `minIntrinsicWidth`?

rock3r (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
AlexVanGogen: Now I'm calculating minIntrinsicWidth based on the longest visible word in a paragraph. Not sure if it's the best approach (and that I'm handling it correctly), but it works nicely. I tested it on top of an improved table layout (the one that uses minIntrinsicWidth) and the images only downscale to the width of the longest word.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
AlexVanGogen: Missed this one, moved the method down

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,21 +845,290 @@ public open class DefaultMarkdownBlockRenderer(
...
+                            fontFamilyResolver = fontFamilyResolver,
+                        )
+                        .minIntrinsicWidth
+                )
+                .toInt()
+        }
+    }
+
+    /** Scales the inline content placeholders by the given scale factor. */
+    private fun scaleInlineContent(
```
rock3r, line 1112: This scaling path seems to assume all image placeholders use `sp` units. `ImageRendererExtension` exposes raw Compose `InlineTextContent`, and `Placeholder.width` / `height` are `TextUnit`s that can also be `em`. `Density.roundToPx()` throws for non-`sp` text units (`Only Sp can convert to Px`), while passing the placeholder directly to `Text` lets Compose resolve `em` placeholders relative to the text style.

The built-in Coil renderer uses `sp`, so it is fine, but a custom image renderer returning e.g. `Placeholder(width = 1.em, height = 1.em, ...)` would now crash once this scaling path runs. Can we either handle `em` placeholders here, or make the `ImageRendererExtension` contract explicitly require `sp` placeholders and guard this path with a clearer failure mode?

platform/jewel/markdown/core/src/testFixtures/kotlin/org/jetbrains/jewel/markdown/TestTheme.kt
```diff
@@ -0,0 +1,183 @@
...
+import org.jetbrains.jewel.foundation.TextColors
+import org.jetbrains.jewel.foundation.theme.ThemeColorPalette
+import org.jetbrains.jewel.foundation.theme.ThemeDefinition
+import org.jetbrains.jewel.foundation.theme.ThemeIconData
+import org.jetbrains.jewel.markdown.rendering.InlinesStyling
+import org.jetbrains.jewel.markdown.rendering.MarkdownStyling
+import org.jetbrains.jewel.markdown.rendering.MarkdownStyling.List.Ordered.NumberFormatStyles.NumberFormatStyle
+import org.jetbrains.jewel.markdown.rendering.MarkdownStyling.List.Unordered.BulletCharStyles
+
+public fun createThemeDefinition(): ThemeDefinition =
```
rock3r, line 29: Could these tests reuse the existing `markdown.testing` helpers instead of adding a new `core` test-fixtures source set? `TestTheme.kt` looks very close to `MarkdownTestTheme.kt` in `platform/jewel/markdown/testing` (`createMarkdownTestThemeDefinition()`, `createMarkdownTestStyling()`, etc.), and the existing testing module is already used by other Markdown tests.

If the images tests can depend on `projects.markdown.testing`, we could avoid duplicating the theme/styling setup and avoid adding extra `java-test-fixtures` / source-root wiring to `markdown.core`. If there is a reason this needs to live in core test fixtures instead, can we keep the fixture surface minimal and/or name it consistently with the existing test helpers?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -78,6 +97,7 @@ import org.jetbrains.jewel.ui.component.HorizontallyScrollableContainer
 import org.jetbrains.jewel.ui.component.Text
 
 private const val DISABLED_CODE_ALPHA = .5f
+private const val COMPOSE_INLINE_CONTENT_ANNOTATION_TAG = "androidx.compose.foundation.text.inlineContent"
```
rock3r, line 100: Since this mirrors Compose's internal `INLINE_CONTENT_TAG`, could we leave a comment here and maybe add a small guard test? I get why we need the exact tag to filter inline-content annotations, but the constant is internal to Compose (`InlineTextContent.kt`), so a future Compose upgrade could change it without a compiler error.

If that happens, regular `Text` rendering may still work, but `getPlaceholders()` / `withoutInlineContent()` would stop seeing inline-content ranges and our custom intrinsic calculations would silently diverge. A comment plus a tiny test around `appendInlineContent("id")` would make this dependency explicit.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,21 +845,290 @@ public open class DefaultMarkdownBlockRenderer(
...
+
+        // Pre-measure text with unscaled inline content to get intrinsic width.
+        // This is used for grid-based containers to preserve proportions, i.e., tables.
+        val textMeasurer = rememberTextMeasurer()
+        val fontFamilyResolver = LocalFontFamilyResolver.current
+        val layoutDirection = LocalLayoutDirection.current
+        val mergedStyle = remember(style, textAlign) { style.merge(TextStyle(textAlign = textAlign)) }
+        val textWithoutInlineContent = remember(text, inlineContentIds) { text.withoutInlineContent(inlineContentIds) }
+
+        val maxIntrinsicWidth =
```
rock3r, line 897: Edge-case question on the custom `minIntrinsicWidth`: the mixed text+image case is covered, but this manually computed value can diverge from regular `Text` intrinsics in a few cases.

* For image-only paragraphs/headings, `textWithoutInlineContent` becomes effectively empty, so `minIntrinsicWidth` returns `0`. A parent using `IntrinsicSize.Min` or otherwise querying `minIntrinsicWidth` could then measure the image at zero width.
* `withoutInlineContent()` rebuilds the text without preserving span annotations, and `minParagraphIntrinsicWidth()` measures each segment with `annotations = emptyList()`. Inline code/emphasis/links/custom inline renderers can change font size/family/weight/etc., so this can under- or over-report compared with what `Text` would use for the same styled content.
* `text.split(WHITESPACE_REGEX)` assumes line-break opportunities are whitespace-delimited. Compose text layout can break some content without whitespace, for example CJK text, so this can also over-report the min width compared with regular `Text`.

Is this approximation an accepted trade-off? If yes — and I don't think it should be — can we add a short comment/test for the intended edge-case behavior? Otherwise, should this preserve more of the original `AnnotatedString`/text-layout behavior?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,21 +845,290 @@ public open class DefaultMarkdownBlockRenderer(
...
+                density,
+            ) {
+                object : MeasurePolicy {
+                    override fun MeasureScope.measure(
+                        measurables: List<Measurable>,
+                        constraints: Constraints,
+                    ): MeasureResult {
+                        val placeable = measurables.firstOrNull()?.measure(constraints)
+
+                        return layout(
```
rock3r, line 947: This wrapper should probably forward the text alignment lines. Before this change the paragraph rendered a `Text` directly, so parents using baseline alignment could read `FirstBaseline` / `LastBaseline`. In the scalable path we now have `Layout` -> `SubcomposeLayout` -> `Text`, but both `layout(...)` calls place the child without passing any `alignmentLines`, so the markdown block no longer exposes the text baselines when this path is used.

Could we forward the measured child’s `FirstBaseline` / `LastBaseline` through both wrapper layers when specified, or add coverage if baseline alignment is meant to remain supported?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -935,3 +1231,8 @@ private fun MimeType.Known.fromMimeTypeString(mimeType: String): MimeType =
 
         else -> UNKNOWN
     }
+
+/** Test tag applied to the scalable inline content layout when SubcomposeLayout path is used. */
+public const val DOWNSCALED_INLINE_CONTENT_TAG: String = "ScalableInlineContent"
```
rock3r, line 1236: Can we avoid exposing this as a regular top-level production API? I get that the images extension tests need cross-module access, but this tag is still an implementation/testing detail.

Could we move it under a `@VisibleForTesting public object JewelMarkdownTags` or similar in `markdown.core`, and use that from both production code and tests? At minimum, this public const should be annotated as `@VisibleForTesting`. If it remains public, the API dumps should be updated too.

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
```diff
@@ -123,6 +143,365 @@ public class Coil3ImageRendererExtensionImplTest {
             .assertHeightIsAtLeast(fakeImageHeight.dp)
     }
 
+    @Test
+    public fun `single image wider than container is downscaled to fit available width`() {
```
rock3r, line 147: Could we add one assertion that covers the reserved vertical space, not only the inline image child size? The PR/release note specifically calls out adjusting vertical space, but the new tests mostly assert the `Image` node dimensions and only check that `DOWNSCALED_INLINE_CONTENT_TAG` exists.

A regression where the child image is measured at the scaled size but the surrounding text/paragraph still reserves the original placeholder height would not be caught. This should be possible without adding production test surface by tagging the test-owned wrapper around `RenderParagraph`, for example:

```kotlin
Box(
    modifier = Modifier
        .width(containerWidthDp.dp)
        .testTag("paragraph-wrapper"),
) {
    blockRenderer.RenderParagraph(...)
}

composeTestRule
    .onNodeWithTag("paragraph-wrapper")
    .assertHeightIsEqualTo(expectedParagraphHeight.dp)
```

Alternatively, the test could place a tagged sibling after this wrapper and assert its position.

rock3r (changes requested):

rock3r:
Could we update the PR description/release notes to match the current implementation? The description still talks about splitting content into a `FlowRow` and introducing `ScalableImage`, but the current code uses the `TextWithScalableInlineContent` / `SubcomposeLayout` approach instead.

Also, the image-only paragraph/heading behavior looks user-visible: we previously called out the switch to `FlowRow` for image-only paragraphs in the release notes, and this PR changes that path again. If that behavior change is intentional, can we mention it in the release notes in user-facing terms?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
DanielSouzaBertoldi: so after going over the code over and over to understand what's happening, I'm pondering what to do about number 1.....

Because scaling an image down to fix `0` is semantically defensible. Technically the content can render at 0 width, even if it's effectively invisible. As a user, what's a useful expected lower bound in this case? Should we return 1? The image width? The downside is that any non-zero value would prevent scaling below that point, which arguably defeats the purpose of this feature :/

rock3r: I would imagine it should report the min intrinsic size based on the inline image's size (which would be ~zero if it's not specified and the image is not loaded yet, that's not something we can improve on)

DanielSouzaBertoldi: Let's take this discussion [over to my PR](https://github.com/JetBrains/intellij-community/pull/3525#discussion_r3341818982) 😄

PR #3365, opened 2025-12-30
[JEWEL-1131] Implement badge component

wellingtoncosta:
## Summary
This PR introduces a new component called `Badge`, which is designed to be used in promotions, like New or Beta features. It isn't implemented in IJ platform, so this implementation is derived from similar small promotional UI components, like a small button.

The API allows to use a brush background with a gradient or even a solid color. It can be clickable if you pass a valid lambda to the `onClick` parameter, or passing `null` to make it non-clickable. Additionally, it has both enabled and disabled states as expected. The content is a slot where you can pass a simple text or even a row with icon + text.

## Evidences

| Light |
|-|
| image | 

| Dark |
|-|
| image |

### New features
 * Introduced the new `Badge` component, which is designed for promotions, like New or Beta.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new UI component and extends `DefaultComponentStyling`/`JewelTheme` with a new `badgeStyle` composition local, which may affect consumers relying on styling constructors or theme setup. Risk is mainly API surface expansion and potential visual regressions in theming across light/dark and IDE-bridge variants.
> 
> **Overview**
> Adds a new `Badge` composable (clickable or static, enabled/disabled, hover/focus/press-aware) backed by a new styling model (`BadgeStyles`/`BadgeStyle`/`BadgeColors`/`BadgeMetrics`) and exposes it via `JewelTheme.badgeStyle`.
> 
> Extends theming and styling plumbing to provide badge styles end-to-end: `DefaultComponentStyling` now carries `badgeStyle` (with hidden deprecated constructors and a `fallbackBadgeStyle()`), Int UI standalone gains `BadgeStyles.light()`/`.dark()` factories, and the IDE LaF bridge reads badge colors/metrics from Swing (`IntUiBridgeBadge.kt`).
> 
> Updates the showcase app with a new Badges screen + icons and adds UI tests validating rendering, semantics (role/click action), and enabled/disabled click behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47860d28d38983b6af208828a6e25a05f1cc13f7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3390, opened 2026-01-21
[JEWEL-1228] Add CornerSize and Padding to BannerMetrics

DanielSouzaBertoldi:
# Context
Our current `InnerBannerImpl` code has some hardcoded values for both `cornerSize` and `padding`.

This PR adds both to `BannerMetrics` so users can freely set whatever value they want to, as is the case with all other Jewel components 😄

# Changes
- Added `InlineBanner` and `Banner` comparisons in `SwingComparisonTabPanel`
- Created the new parameters with their default values
   - CornerSize with `8.dp` radius
      - In the IJP's implementation, they actually set the diameter of the arc instead of just the radius. As can be seen in com.intellij.ui.InlineBannerBase:
        ```
        val cornerRadius = JBUI.scale(16)
        ```
        Luckily for us the radius of a diameter is basically D/2, so `8.dp`! :)
   - 12.dp padding
      - Still in com.intellij.ui.InlineBannerBase, the border is declared as:
         ```
         border = JBUI.Borders.empty(12)
         ```
         I.e, an empty view with borders and a content offset of `12 pixels`. We can directly translate this to Compose as `12.dp` 🕺🏻 (there's more to this than just this one line of code, but let's keep it simple for now, ok?) 
- Created a `defaultBannerMetrics()` function  in Bridge that applies the default values for all parameters in `BannerMetrics`
- Added some code to properly align the banner's icon and content with `iconActions` (oh boy.... You'll know what I mean by this soon enough)

<details><summary>First approach (calculating padding based on IconButton minSize patch)</summary>
<p>

If you want to compare the Subcomposlayout approach with my first one, that calculated the action icon padding based on IconButton's minSize, just apply the patch below:

[0001-JEWEL-1228-Add-CornerSize-and-Padding-to-BannerMetrics.patch](https://github.com/user-attachments/files/25097608/0001-JEWEL-1228-Add-CornerSize-and-Padding-to-BannerMetri.patch)

</p>
</details> 

# Screenshots

## `InlineBanner` Variants
| Variant | Before | After (Subcomposelayout) |
|---------|--------|--------|
| No icon  | image | image |
| Simple | image | image |
| One Action Icon | image | image |
| Two Action Icons | image | image |
| Three Action Icon (Jewel only)¹ | image | image |
| One Link Action | image | image |
| Two Link Actions | image | image |
| Three Link Actions | image | image |
| For or More Link Actions | image | image |
| "More" Link Actions Submenu | <video src="https://github.com/user-attachments/assets/739dc2df-65b5-4671-9a43-2211141f8ff2" /> | <video src="https://github.com/user-attachments/assets/ce7be4a9-9f9c-42c5-9c92-dfb2161d856d" /> |
| Title (Jewel only) | image | image |
| Complete | image | image |
| Complete (w/ Title) | image | image |

¹ While messing around with all the variations, I found out that the text content in IJP's  `InlineBanner` doesn't respect the left most icon padding. This is what I mean:

image

The Figma spec doesn't explicitly defines a padding between the text and the icon too, but if it were to have a padding, it'd be `12.dp` it seems. We could easily add this padding into our calculations BUT my idea is to be as faithful as possible to IJP's implementation.

## Standalone Component Showcase
| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/7fa0fbbf-935a-48ec-a10f-f1739269c775" /> | <video src="https://github.com/user-attachments/assets/a416271c-72f0-468e-bdfb-b3cbf5224a19" />  | 

## Jewel Tool Window Components Tab
| Before | After |
|--------|--------|
| image | image | 

## New Demos
| Wrapping Content Size | RTL |
|---|---|
| image | image |

For other less important notes, check the section below

<details><summary>Details</summary>

1. You can see that our `InlineBanner` is slightly smaller than Swing. At first I thought that discrepancy was because `IconActionButton` has a default of `minSize: DpSize = DpSize(24.dp, 24.dp),` while the default icon size for Swing is `26px`, but setting a `minSize` of 26.dp in our `IconButton` does not match Swing's height still. I went over the Swing code again and again and so far I'm unable to tell exactly what's the culprit here... Ours perfectly match the specs from Figma, though.
2. Also, our `DefaultBanner` is the inverse: it's slightly bigger than Swing, but this PR (for now) is focused solely on `InlineBanner`.

</details> 

### ⚠️ Important Changes
 * `InlineBanner` no longer fills the available width by default; it now wraps its content instead. If you want banners to span the full width, be sure to add `modifier = Modifier.fillMaxWidth()`.

### Bug fixes
 * Now it's possible to customize the corner size and the padding of the `InlineBanner` components

### Deprecated API
* `BannerMetrics` constructor has been deprecated in favor of the new constructor that takes `borderWidth`, `cornerSize` and `paddingValues`
 * `BannerMetrics.Companion.default(borderWidth: Dp)` has been deprecated in favor of `BannerMetrics.Companion.default(borderWidth: Dp, cornerSize: CornerSize, paddingValues: PaddingValues)`

### DIFF
--- a/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUIBannerStyling.kt
+++ b/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUIBannerStyling.kt
@@ -1,5 +1,7 @@
 package org.jetbrains.jewel.intui.standalone.styling
 
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.shape.CornerSize
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
@@ -183,7 +185,18 @@ public object IntUiInlineBannerStylesFactory {
         InlineBannerStyles(information = information, success = success, warning = warning, error = error)
 }
 
-public fun BannerMetrics.Companion.default(borderWidth: Dp = 1.dp): BannerMetrics = BannerMetrics(borderWidth)
+@Deprecated(
+    "Use the `default()` function with `cornerSize` and `paddingValues`",
+    replaceWith = ReplaceWith("default(borderWidth, cornerSize = TODO(), paddingValues = TODO())"),
+)
+public fun BannerMetrics.Companion.default(borderWidth: Dp = 1.dp): BannerMetrics =
+    BannerMetrics(borderWidth, CornerSize(8.dp), PaddingValues(12.dp))
+
+public fun BannerMetrics.Companion.default(
+    borderWidth: Dp = 1.dp,
+    cornerSize: CornerSize = CornerSize(8.dp),
+    paddingValues: PaddingValues = PaddingValues(12.dp),
+): BannerMetrics = BannerMetrics(borderWidth, cornerSize, paddingValues)
 
 // region Inline Information Banner
 public val InlineBannerStyle.Companion.Information: IntUiInlineInformationBannerStyleFactory
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
@@ -16,25 +16,26 @@ import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.RowScope
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.calculateEndPadding
+import androidx.compose.foundation.layout.calculateStartPadding
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.CornerSize
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.layout.SubcomposeLayout
+import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontWeight.Companion.Bold
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import org.jetbrains.annotations.Nls
-import org.jetbrains.jewel.foundation.modifier.thenIf
 import org.jetbrains.jewel.foundation.theme.JewelTheme
 import org.jetbrains.jewel.foundation.theme.LocalContentColor
 import org.jetbrains.jewel.ui.component.banner.BannerActionsRow
@@ -44,6 +45,9 @@ import org.jetbrains.jewel.ui.component.banner.BannerLinkActionScope
 import org.jetbrains.jewel.ui.component.styling.InlineBannerStyle
 import org.jetbrains.jewel.ui.icons.AllIconsKeys
 import org.jetbrains.jewel.ui.theme.inlineBannerStyle
+
+private const val BANNER_ICON_SIZE = 16
+private const val BANNER_CONTENT_SPACING = 8
 
 /**
  * Displays an informational inline banner providing subtle, non-intrusive context or feedback.
@@ -1222,46 +1226,121 @@ private fun InlineBannerImpl(
     content: @Composable (() -> Unit),
 ) {
     val borderColor = style.colors.border
+    val layoutDirection = LocalLayoutDirection.current
+    val originalPadding = style.metrics.padding
+
+    val endPadding = if (actionIcons != null) 0.dp else originalPadding.calculateEndPadding(layoutDirection)
+    val adjustedPadding =
+        PaddingValues(
+            start = originalPadding.calculateStartPadding(layoutDirection),
+            top = originalPadding.calculateTopPadding(),
+            bottom = originalPadding.calculateBottomPadding(),
+            end = endPadding,
+        )
+
     RoundedCornerBox(
         modifier = modifier.testTag("InlineBanner"),
         borderColor = borderColor,
         backgroundColor = style.colors.background,
         contentColor = JewelTheme.contentColor,
-        borderWidth = 1.dp,
-        cornerSize = CornerSize(8.dp),
-        padding = PaddingValues(),
+        borderWidth = style.metrics.borderWidth,
+        cornerSize = style.metrics.cornerSize,
+        padding = adjustedPadding,
     ) {
-        Row(modifier = Modifier.padding(start = 12.dp)) {
-            if (icon != null) {
-                Box(modifier = Modifier.padding(top = 12.dp, bottom = 12.dp).size(16.dp)) { icon() }
-                Spacer(Modifier.width(8.dp))
-            }
+        SubcomposeLayout { constraints ->
+            val spacingPx = BANNER_CONTENT_SPACING.dp.roundToPx()
 
-            Column(
-                modifier =
-                    Modifier.weight(1f)
-                        .padding(top = 12.dp, bottom = 12.dp) // kftmt plz behave
-                        .thenIf(actionIcons == null) { padding(end = 12.dp) }
-            ) {
-                if (title != null) {
-                    Text(text = title, style = textStyle, fontWeight = Bold)
-                    Spacer(Modifier.height(8.dp))
-                }
-                content()
+            val unconstrained = constraints.copy(minWidth = 0)
+            val actionIconsPlaceables =
+                subcompose("actionIcons") {
+                        if (actionIcons != null) {
+                            Row(horizontalArrangement = Arrangement.spacedBy(2.dp)) { actionIcons() }
+                        }
+                    }
+                    .map { it.measure(unconstrained) }
+            val actionIconsWidth = actionIconsPlaceables.maxOfOrNull { it.width } ?: 0
 
-                if (actions != null) {
-                    Spacer(Modifier.height(8.dp))
-                    FlowRow(horizontalArrangement = Arrangement.spacedBy(16.dp)) { actions() }
-                }
-            }
+            val iconPlaceables =
+                subcompose("icon") {
+                        if (icon != null) {
+                            Box(modifier = Modifier.size(BANNER_ICON_SIZE.dp)) { icon() }
+                        }
+                    }
+                    .map { it.measure(unconstrained) }
+            val iconWidth = iconPlaceables.firstOrNull()?.width ?: 0
+            val iconHeight = iconPlaceables.firstOrNull()?.height ?: 0
 
-            if (actionIcons != null) {
-                Spacer(Modifier.width(8.dp))
-                Row(
-                    modifier = Modifier.align(Alignment.Top).padding(top = 8.dp, end = 8.dp, bottom = 8.dp),
-                    horizontalArrangement = Arrangement.spacedBy(2.dp),
-                ) {
-                    actionIcons()
+            // Calculate width available for the main content column
+            // Total - Icon - Spacing
+            val startOffset = if (iconWidth > 0) iconWidth + spacingPx else 0
+            val contentAvailableWidth = (constraints.maxWidth - startOffset).coerceAtLeast(0)
+
+            // Text MUST respect Action Icons (i.e, subtract their width)
+            val textConstraints =
+                constraints.copy(minWidth = 0, maxWidth = (contentAvailableWidth - actionIconsWidth).coerceAtLeast(0))
+            val textPlaceables =
+                subcompose("text") {
+                        Column {
+                            if (title != null) {
+                                Text(text = title, style = textStyle, fontWeight = Bold)
+                                Spacer(Modifier.height(BANNER_CONTENT_SPACING.dp))
+                            }
+                            content()
+                        }
+                    }
+                    .map { it.measure(textConstraints) }
+            val textHeight = textPlaceables.maxOfOrNull { it.height } ?: 0
+            val textWidth = textPlaceables.maxOfOrNull { it.width } ?: 0
+
+            // Link Actions must IGNORE Action Icons (use full available width)
+            // This allows buttons to render underneath the top-right icons
+            val linkConstraints = constraints.copy(minWidth = 0, maxWidth = contentAvailableWidth)
+            val linkPlaceables =
+                subcompose("links") {
+                        if (actions != null) {
+                            FlowRow(horizontalArrangement = Arrangement.spacedBy(16.dp)) { actions() }
+                        }
+                    }
+                    .map { it.measure(linkConstraints) }
+            val linksHeight = linkPlaceables.maxOfOrNull { it.height } ?: 0
+            val linksWidth = linkPlaceables.maxOfOrNull { it.width } ?: 0
+
+            val contentHeight = textHeight + (if (linksHeight > 0) spacingPx else 0) + linksHeight
+            val totalHeight = maxOf(iconHeight, contentHeight)
+            val layoutHeight = totalHeight.coerceIn(constraints.minHeight, constraints.maxHeight)
+            // calculating the width the banner actually wants to occupy, coerced to
+            // fit within the incoming minWidth/maxWidth constraints
+            val naturalWidth =
+                (startOffset + maxOf(textWidth + actionIconsWidth, linksWidth)).coerceIn(
+                    constraints.minWidth,
+                    constraints.maxWidth,
+                )
+
+            layout(naturalWidth, layoutHeight) {
+                iconPlaceables.forEach { it.placeRelative(0, 0) }
+
+                // Starts after icon
+                textPlaceables.forEach { it.placeRelative(startOffset, 0) }
+
+                // Starts after icon, below text
+                val linkY = textHeight + (if (linksHeight > 0) spacingPx else 0)
+                linkPlaceables.forEach { it.placeRelative(startOffset, linkY) }
+
+                if (actionIconsPlaceables.isNotEmpty()) {
+                    // We always offset the action icon to half the padding
+                    val topPaddingPx = adjustedPadding.calculateTopPadding().roundToPx()
+                    val halfPaddingTop = (originalPadding.calculateTopPadding().toPx() / 2).toInt()
+                    val halfPaddingEnd = (originalPadding.calculateEndPadding(layoutDirection).toPx() / 2).toInt()
+
+                    // Offset Logic:
+                    val yPos = halfPaddingTop - topPaddingPx
+
+                    actionIconsPlaceables.forEach { placeable ->
+                        // For the X calculation, naturalWidth is the edge (since we killed end padding)
+                        // Move left by icon width + half original padding
+                        val xPos = naturalWidth - placeable.width - halfPaddingEnd
+                        placeable.placeRelative(x = xPos, y = yPos)
+                    }
                 }
             }
         }
@@ -1286,7 +1365,8 @@ private fun RoundedCornerBox(
                 .border(borderWidth, borderColor, shape)
                 .background(backgroundColor, shape)
                 .clip(shape)
-                .padding(padding)
+                .padding(padding),
+        propagateMinConstraints = true,
     ) {
         CompositionLocalProvider(LocalContentColor provides contentColor) { content() }
     }
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
@@ -1,11 +1,14 @@
 package org.jetbrains.jewel.ui.component.styling
 
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.shape.CornerSize
 import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.ProvidableCompositionLocal
 import androidx.compose.runtime.Stable
 import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
 import org.jetbrains.jewel.foundation.GenerateDataFunctions
 
 @Stable
@@ -170,21 +173,48 @@ public class BannerColors(public val background: Color, public val border: Color
     public companion object
 }
 
+/**
+ * Defines the geometric and spacing properties for a banner component.
+ *
+ * @param borderWidth The width of the banner's border stroke.
+ * @param cornerSize The corner radius applied to the banner's shape.
+ * @param padding The internal padding applied to the banner's content area.
+ */
 @Stable
 @GenerateDataFunctions
-public class BannerMetrics(public val borderWidth: Dp) {
+public class BannerMetrics(
+    public val borderWidth: Dp,
+    public val cornerSize: CornerSize,
+    public val padding: PaddingValues,
+) {
+    @Deprecated(
+        "Use the constructor with `cornerSize`  and `padding` parameters",
+        replaceWith = ReplaceWith("BannerMetrics(borderWidth, cornerSize = TODO(), padding = TODO())"),
+    )
+    public constructor(borderWidth: Dp) : this(borderWidth, CornerSize(8.dp), PaddingValues(12.dp))
+
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
         if (javaClass != other?.javaClass) return false
 
         other as BannerMetrics
 
-        return borderWidth == other.borderWidth
+        if (borderWidth != other.borderWidth) return false
+        if (cornerSize != other.cornerSize) return false
+        if (padding != other.padding) return false
+
+        return true
     }
 
-    override fun hashCode(): Int = borderWidth.hashCode()
+    override fun hashCode(): Int {
+        var result = borderWidth.hashCode()
+        result = 31 * result + cornerSize.hashCode()
+        result = 31 * result + padding.hashCode()
+        return result
+    }
 
-    override fun toString(): String = "BannerMetrics(borderWidth=$borderWidth)"
+    override fun toString(): String =
+        "BannerMetrics(" + "borderWidth=$borderWidth, " + "cornerSize=$cornerSize, " + "padding=$padding" + ")"
 
     public companion object
 }
--- a/plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
+++ b/plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
@@ -31,6 +31,8 @@ import com.intellij.ide.ui.laf.darcula.ui.DarculaButtonUI
 import com.intellij.openapi.editor.colors.EditorFontType
 import com.intellij.openapi.ui.ComboBox
 import com.intellij.openapi.util.IconLoader
+import com.intellij.ui.EditorNotificationPanel
+import com.intellij.ui.InlineBanner
 import com.intellij.ui.JBColor
 import com.intellij.ui.components.BrowserLink
 import com.intellij.ui.components.JBLabel
@@ -48,9 +50,12 @@ import org.jetbrains.jewel.bridge.JewelComposePanel
 import org.jetbrains.jewel.bridge.retrieveEditorColorScheme
 import org.jetbrains.jewel.foundation.theme.JewelTheme
 import org.jetbrains.jewel.ui.component.DefaultButton
+import org.jetbrains.jewel.ui.component.DefaultInformationBanner
 import org.jetbrains.jewel.ui.component.EditableListComboBox
 import org.jetbrains.jewel.ui.component.ExternalLink
 import org.jetbrains.jewel.ui.component.Icon
+import org.jetbrains.jewel.ui.component.InformationDefaultBanner
+import org.jetbrains.jewel.ui.component.InlineInformationBanner
 import org.jetbrains.jewel.ui.component.ListComboBox
 import org.jetbrains.jewel.ui.component.OutlinedButton
 import org.jetbrains.jewel.ui.component.Text
@@ -60,6 +65,7 @@ import org.jetbrains.jewel.ui.disabledAppearance
 import org.jetbrains.jewel.ui.icons.AllIconsKeys
 import org.jetbrains.jewel.ui.theme.textAreaStyle
 import org.jetbrains.jewel.ui.typography
+import java.awt.Dimension
 import javax.swing.BoxLayout
 import javax.swing.DefaultComboBoxModel
 import javax.swing.JLabel
@@ -76,6 +82,10 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
       iconsRow()
       separator()
       linksRow()
+      separator()
+      inlineBannersRow()
+      separator()
+      bannersRow()
       separator()
       textFieldsRow()
       separator()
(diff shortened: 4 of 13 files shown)


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
```diff
@@ -1222,26 +1225,51 @@ private fun InlineBannerImpl(
...
+    val layoutDirection = LocalLayoutDirection.current
+
+    // We use max(0.dp) to prevent crashes if a user sets padding < 4dp
+    val adjustedContainerPadding =
+        PaddingValues(
+            start = originalPadding.calculateStartPadding(layoutDirection),
+            end = originalPadding.calculateEndPadding(layoutDirection),
+            top = (originalPadding.calculateTopPadding() - buttonInternalPadding).coerceAtLeast(0.dp),
+            bottom = bottomPadding,
+        )
```
DanielSouzaBertoldi: Hmmm.... That made a lot of sense, right?
...Right?

Maybe not. In any case, I feel obliged to either apologize for the confusion (I'm sorry 😭) or explain why this "monstrosity" was necessary. Bear with me!

First of all, I want you to see the specs for the Swing component:

image

Now, let's check how our current code does it:

image

Basically, each part of this UI is applying some sort of padding. The `Row` composable that wraps all the content only adds the `12.dp` to the start. Then it's up to each part in the component to handle its own padding. As you've probably seen from the evidences section, this works!

But that's because the `padding` that we are passing to `RoundedCornerBox` (the "root" Composable of this banner) is 0.

Now, since users can add whatever padding they want, we don't get this luxury anymore of letting each part of the UI choose how much padding they need. We have to use whatever is in `padding` and adjust our component accordingly.

That doesn't seem like that much of a headache. I mean, just fetch the info from `style.metrics.padding`, pass it down to `RoundedCornerBox` and we're golden, right?

...Right?

Maybe not. Here, let's recreate Swing's specs faithfully and check the result!

image

Now, the `RoundedCornerBox` applies the overall padding, and we removed most of the specific paddings from sub Composables. Judging solely by the image above, everything seems fine. Let's check it on the IDE!

| Comparing W/ Swing | Colored Background |
|--------|--------|
| image | image |

Yeah, not really. Both the icon and the text are not aligned with the close button icon. Also, now our `InlineBanner` is bigger than Swing!

Reviewer, I've been withholding a minor detail from you. You've probably noticed it, but the thing is, there's a catch in the Swing component:

image

The 12px of padding in the first image is taking into account the 5px from the vertical spacing from the button icon. The actual padding applied on top is 7px (7px + 5px of padding = 12px).

If we apply the padding not accounting for the icon button's padding, we end up with the result in our first try. The padding will be applied on top of the biggest component (the button icon, in this case), that's why our `InlineBanner` is now bigger than Swing. Besides, the text is not aligned with the button because it's aligned with the button's padding.

So. The plan now is to:

1. Get the intended padding from `style.metrics.padding`
2. Fetch the declared `IconButtonStyle.metrics.minSize`
3. Subtract 16 from its `height` (I can't guarantee that all icons have 16px, judging by the icons in `AllIconsKeys.java` they _all probably_ are)
4. Divide it by 2 to finally fetch the number of `dp`s on the top/bottom of the icon button.

Then, we fetch the top padding in `style.metrics.padding` and just subtract the value we got from step 4 above, given us the correct top padding the user expects!

We also offset the `icon` and the "middle content" of the `InlineBanner` with the value from step 4, so that it's aligned with the icon itself NOT its padding. Here's the result:

Now, supposing the `IconButton` has a `minSize` of `24` and the `12.dp` of padding, the padding applied in our Composable looks like this:

image

The `IconButton` will have a vertical padding of `4.dp`, we just apply `8.dp` as padding and we get `12.dp` total! The `4.dp` of internal padding is applied to the top of the icon (grey) and the text (green).

| Comparing W/ Swing | Colored Background |
|--------|--------|
| image | image |

Now we faithfully replicate IJP's version of `InlineBanner` (minus the 2px of difference in height due to different icon button sizes) + the icon and the text is now properly aligned with the close button icon!

DanielSouzaBertoldi: Two main problems with this approach though:

1. We rely solely on `IconButtonStyle.minSize` to figure out the internal padding
2. We suppose that the icon itself will be only `16.dp` in size

What if the icon itself is `30px` in size? After all, users can set an arbitrary icon that they wish. We would still be calculating based on `minSize`, so, supposing a `minSize` of `24.dp`, we would apply a top padding of `8px` instead of `5px` ((30-16)/2 = 7).

Even though we are supposing a bunch of stuff, the code is already overly complex. I think we could do the calculations if we apply `onGloballyPositioned` in the `Row` that wraps the action icons and then update the padding accordingly, but that would "flicker" the component since we'd have to waste at least 1 frame to readjust it...

🤔

Any other better ideas?
~~Help~~

faogustavo: > What if the icon itself is 30px in size?

Have you checked if the behavior is still working as intended if you increase the IDE scale?

faogustavo (approved):
Great work man 👏

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
DanielSouzaBertoldi: It is not, unfortunately. Here, let me show a few more screenshots!

image

Here, the yellow line is the calculated top padding, the blue line is the "added" padding (so the icon reaches its minSize of 24.dp), the red line is the size of the icon itself and the green line where the content on the left side should be rendered. This is the "happy" scenario.

Now, let's render an icon that has 40px in size:

image

The colors follow the same logic as before. There's no blue line because the icon itself is greater than its min size. In this case, we are only adding `8.dp` of padding instead of `12.dp`. The "empty" space in the icon is just that the icon is centralized, but its occupying the whole available space. We are also adding a padding of `4.dp` to where the content starts, although it should be rendered at the start of the outline, since the icon itself doesn't have a "padding".

Here's the screenshot in its entirety:

image

rock3r: giphy

DanielSouzaBertoldi: Resolving this conversation for now since my second commit uses `Subcomposelayout` to manually place the composables!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
```diff
@@ -1222,46 +1226,110 @@ private fun InlineBannerImpl(
...
+
+                // Starts after icon, below text
+                val linkY = textHeight + (if (linksHeight > 0) spacingPx else 0)
+                linkPlaceables.forEach { it.place(startOffset, linkY) }
+
+                if (actionIconsPlaceables.isNotEmpty()) {
+                    // We always offset the action icon to half the padding
+                    val topPaddingPx = adjustedPadding.calculateTopPadding().roundToPx()
+                    val halfPaddingTop = (originalPadding.calculateTopPadding().toPx() / 2).toInt()
+                    val halfPaddingEnd = (originalPadding.calculateEndPadding(layoutDirection).toPx() / 2).toInt()
```
DanielSouzaBertoldi, line 1333: Now, instead of trying to calculate the action icon position given the `minSize` in `IconButtonMetrics`, we just halve the padding set by the user and call it a day.

Yes, this means that for Jewel, in a default scenario (explained in the resolved conversation above) , action icons will be 1 pixel off on the top and end positions, since ideally it should be 7 but not it'll be 6

I am fine with that, though

plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
```diff
@@ -97,6 +133,80 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
...
+      inlineBanner.addAction("Action C") {}
+      inlineBanner.addAction("Action D") {}
+
+      cell(
+        component = inlineBanner
+      )
+        .align(AlignY.CENTER)
+
+      compose {
+        Box(Modifier.width(256.dp)) {
```
DanielSouzaBertoldi: I've no clue as to why, but if we don't restrict the `InlineBanner` to a certain width the constraints passed down to it will be HUGE. This only happens with this component (before and after these changes), which I find very strange. `256` is Swing's width, so I just mirrored it.

I may be tempted to revamp this tab to something more easy to use and with more examples though..... in another PR, of course 😄

wellingtoncosta:
@DanielSouzaBertoldi you have to squash your commits into a single one

DanielSouzaBertoldi:
@wellingtoncosta yeah I know, I mentioned this in my description 😄 

> If you are reviewing this PR before it is merged, you can compare my second solution with the first by checking both commits!

Thanks for the heads up!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
```diff
@@ -172,19 +174,33 @@ public class BannerColors(public val background: Color, public val border: Color
 
 @Stable
 @GenerateDataFunctions
-public class BannerMetrics(public val borderWidth: Dp) {
+public class BannerMetrics(
+    public val borderWidth: Dp,
+    public val cornerSize: CornerSize,
+    public val padding: PaddingValues,
```
nebojsa-vuksic, line 188: This change will break source and binary compatibility for existing consumers. To maintain backward compatibility, could you add a secondary constructor that takes only `borderWidth` and delegates to the primary one?"

daaria-s (commented):
I don't see any changes in api-dump.txt files, but I guess they should appear here. Could you please run api check configuration for that?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
DanielSouzaBertoldi: Damn, sorry, I was so focused on the UI side that I forgot to run ApiCheckTest and verify breaking changes. Just pushed the api changes :)

daaria-s (approved):
looks good to me now

plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
nebojsa-vuksic: There appears to be a specific problem with how Banners interoperate with Swing.
I'm not 100% sure about the root cause of the problem, but this is what I got so far.

During the measuring and layout phase, Compose uses initial constraints set to `Infinity`:
```
private const val Infinity = Int.MAX_VALUE
```
This is defined in `androidx.compose.ui.unit.Constraints.kt` in the `ui-unit-desktop` jar.

During this process, Compose attempts to acquire constraints from its parent (in this case, the Swing component) but it seems these constraints aren't yet available at this point of time. This failure causes Compose to fall back to using the `Infinity` constraint, which ultimately triggers the error during Metal renderer initialization:
```-[MTLTextureDescriptorInternal validateWithDevice:]:1416: failed assertion 
`MTLTextureDescriptor has width (32766) greater than maximum allowed size of 16384.
```

Setting the predefined width works around the problem. The core issue is still present in the Compose implementation and it should be fixed. 

cc @rock3r

DanielSouzaBertoldi: Weird. Just [read the ticket](https://youtrack.jetbrains.com/issue/JEWEL-1247/InlineBanner-crashes-on-macOS-Metal-texture-width-exceeds-16384-pixel-limit) and according to Seb this looks like a CMP/Skiko render problem 🤔

rock3r: It is :)

rock3r: The Metal crash is, to be clear. The constraints thing isn't

rock3r: Did some AI-powered triage, and it seems like there is one issue here: the layout is trying to take all available room even when it doesn't need it. If `InlineBanner`s were content-sized by default (wrapping its text/actions), it would behave just like every other component in the demo that doesn't crash — `Button`, `Text`, `TextField`, etc. all size themselves from content and are fine with infinite constraints.

Users who want fill-width (the common case) would just write:

```kotlin
InlineInformationBanner(
    text = "...",
    modifier = Modifier.fillMaxWidth(),
)
```

And in the Swing interop case, the parent would already need to provide bounded constraints for `fillMaxWidth()` to work sensibly anyway — which means the problem is caught at the right level (the embedding site) rather than hidden inside the component.

The only downside is it's a behavioral breaking change — every existing usage without an explicit `fillMaxWidth()` would suddenly shrink-wrap. But that's arguably the more correct Compose convention (most built-in components like `Button`, `CheckBoxRow`, `TextField` are content-sized by default, and users opt into filling).

@DanielSouzaBertoldi wdyt

DanielSouzaBertoldi: Hmmmm for default banners I can see that we do set a weight of 1f in the content, forcing the component to fill the entire available space:

`Box(Modifier.weight(1f)) { content() }`

But in this case, for inline banners, we are just adjusting the layout based on the constraints provided to the Composable... ~~Nowhere in the code we set fillMaxWidth or add a weight. I'll investigate this a bit further~~

Nvm, I do use constraints.maxWidth to calculate the position of everything on screen. I'll have to update the subcomposelayout code to somehow wrap the content size

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
```diff
@@ -1222,46 +1226,110 @@ private fun InlineBannerImpl(
...
+                            FlowRow(horizontalArrangement = Arrangement.spacedBy(16.dp)) { actions() }
+                        }
+                    }
+                    .map { it.measure(linkConstraints) }
+            val linksHeight = linkPlaceables.maxOfOrNull { it.height } ?: 0
+
+            val contentHeight = textHeight + (if (linksHeight > 0) spacingPx else 0) + linksHeight
+            val totalHeight = maxOf(iconHeight, contentHeight)
+
+            layout(constraints.maxWidth, totalHeight) {
```
rock3r: I think this is the remaining piece from the earlier sizing discussion: the custom layout still hard-codes its width to `constraints.maxWidth`, so the banner expands even when its content is smaller. That seems to keep the Swing interop / unbounded-constraints issue alive. Would it make sense for `InlineBanner` to be content-sized by default and only fill width when callers opt into `Modifier.fillMaxWidth()`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
```diff
@@ -1222,46 +1226,110 @@ private fun InlineBannerImpl(
...
+                iconPlaceables.forEach { it.place(0, 0) }
+
+                // Starts after icon
+                textPlaceables.forEach { it.place(startOffset, 0) }
```
rock3r: This layout is computing start/end padding with `LocalLayoutDirection`, but the actual child placement is done with absolute `place(...)` calls. That looks LTR-only. Could we switch the placements in this layout to `placeRelative(...)` / relative positioning so standalone RTL layouts behave correctly too?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
```diff
@@ -172,19 +175,40 @@ public class BannerColors(public val background: Color, public val border: Color
 
 @Stable
 @GenerateDataFunctions
-public class BannerMetrics(public val borderWidth: Dp) {
+public class BannerMetrics(
+    public val borderWidth: Dp,
+    public val cornerSize: CornerSize,
```
rock3r, line 187: One design question here: `BannerMetrics` is shared by default + inline banners, and this change adds `cornerSize`/`padding` to the shared model. But `DefaultBanner` still hardcodes its own padding/border layout and doesn't consume these fields. Is the intent for the new metrics to be inline-only for now? If so, should the model be split, or should `DefaultBanner` be updated too?

rock3r (changes requested):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
DanielSouzaBertoldi: Yes it does, then we won't be needing to hack our way around displaying it in `SwingComparisonTabPanel` 😎 after reading the issue on YT I thought that it was a Skiko problem lol

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
DanielSouzaBertoldi: Oh right. I think it'd be easier to review if I broke this down into two PRs: one for inline banners (as the issue name makes it clear) and one for DefaultBanner. Then in the other PR I can add more screenshots/screen recordings (you know I'm a big fan of those, c'mon 😛) specifically for it.

I'll duplicate the issue on YT and assign it to me

Edit: [here it is!](https://youtrack.jetbrains.com/issue/JEWEL-1290)!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
DanielSouzaBertoldi: I overlooked this detail entirely 😭 yep, as per the documentation, `place` won't implicitly react to RTL while `placeRelative` will.

I'll see if I can add a screenshot of the standalone app in RTL to make sure this was fixed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
DanielSouzaBertoldi: image

Cool, now it's wrapping the content, not filling all available width!

(the icon is touching the text due to the way IJP's inline banner behaves, as explained in the description)

I'll update the release notes and add the behavioral breaking change under the `Important Changes` section

I'll also add `.fillMaxWidth` to the showcase so that they don't render like this

plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
DanielSouzaBertoldi: Done ;)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
```diff
@@ -1280,14 +1359,16 @@ private fun RoundedCornerBox(
     content: @Composable () -> Unit,
 ) {
     val shape = RoundedCornerShape(cornerSize)
-    Box(
+    Layout(
```
DanielSouzaBertoldi: The change to `Layout` was necessary because, from testing, `Box` seems to always measure its children with `minWidth = 0`, screwing up the inline banner `Subcomposelayout` calculations if users set `Modifier.fillMaxWidth()`.

`Layout` on the other hand does not do none of that stuff, it just passes down the information which is exactly what we need

rock3r: I think Box has a propagateMinConstraints parameter for that (why is that not the default I have no clue)

DanielSouzaBertoldi: Oh right, I totally forgot about this parameter. I'll revert it back to a box, thanks!

daaria-s:
Sorry if I didn't find the answer in the discussion, but why do we keep the banner like that with a settings icon being too close? I thought we could add `modifier = Modifier.fillMaxWidth()` to make it look prettier :slightly_smiling_face:
image

DanielSouzaBertoldi:
Hey @daaria-s thanks for the review!

Yes, I do agree it would look a bit better than the current layout, but Swing actually does the same thing and I just mirrored it 😄 

I talked a bit about it in the PR description:

image

My idea was to faithfully follow the weird quirks of Swing (even though it might be a visual bug in this case, but I doubt it given that this component exists for a good while now)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
```diff
@@ -172,19 +175,40 @@ public class BannerColors(public val background: Color, public val border: Color
 
 @Stable
 @GenerateDataFunctions
-public class BannerMetrics(public val borderWidth: Dp) {
+public class BannerMetrics(
```
rock3r, line 185: Nit: could we add KDoc for these new public metrics? `cornerSize` and `padding` are user-facing styling knobs now, and it would be useful to spell out what they control and what the defaults are based on.

It might also be worth mentioning the current scope here, since `BannerMetrics` is shared by default + inline banner styles but this PR only wires the new values into `InlineBanner`.

platform/jewel/ui/metalava/ui-api-0.36.0.txt
```diff
@@ -806,8 +806,11 @@ package org.jetbrains.jewel.ui.component {
...
+    method @SuppressCompatibility @androidx.compose.runtime.Composable @org.jetbrains.annotations.ApiStatus.Experimental @org.jetbrains.jewel.foundation.ExperimentalJewelApi public static void SpeedSearchArea(optional androidx.compose.ui.Modifier modifier, optional kotlin.jvm.functions.Function1<? super java.lang.String,? extends org.jetbrains.jewel.foundation.search.SpeedSearchMatcher> matcherBuilder, optional org.jetbrains.jewel.ui.component.styling.SpeedSearchStyle styling, optional org.jetbrains.jewel.ui.component.styling.TextFieldStyle textFieldStyle, optional androidx.compose.ui.text.TextStyle textStyle, optional org.jetbrains.jewel.ui.component.styling.SearchMatchStyle searchMatchStyle, optional androidx.compose.foundation.interaction.MutableInteractionSource? interactionSource, optional boolean dismissOnLoseFocus, kotlin.jvm.functions.Function1<? super org.jetbrains.jewel.ui.component.SpeedSearchScope,kotlin.Unit> content);
```
rock3r: Is this `SpeedSearch` part intentional in this PR?

The PR scope is banners, but this Metalava dump also includes `SpeedSearchArea` / `SpeedSearchState` changes (`dismissOnLoseFocus`, `rememberSpeedSearchState`, `currentMatcher`, etc.). I don't see corresponding `SpeedSearchArea.kt` source changes in this PR, so this looks like unrelated generated output unless the dump on the base branch was already stale.

If it's not required here, can we keep the Metalava diff limited to the banner API changes?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
```diff
@@ -172,19 +175,40 @@ public class BannerColors(public val background: Color, public val border: Color
...
+    public val borderWidth: Dp,
+    public val cornerSize: CornerSize,
+    public val padding: PaddingValues,
+) {
+    @Deprecated(
```
rock3r, line 190: Looks like this preserves binary compatibility, but not source compatibility. With `DeprecationLevel.HIDDEN`, already-compiled calls to `BannerMetrics(Dp)` keep working, but source calls like `BannerMetrics(1.dp)` won't see a deprecation warning — they'll stop resolving and need to migrate to the new constructor.

That may be an acceptable trade-off, but then the release note should probably call this out as source-incompatible / migration-relevant rather than only as deprecated API.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
```diff
@@ -1222,46 +1227,120 @@ private fun InlineBannerImpl(
...
+                        if (actions != null) {
+                            FlowRow(horizontalArrangement = Arrangement.spacedBy(16.dp)) { actions() }
+                        }
+                    }
+                    .map { it.measure(linkConstraints) }
+            val linksHeight = linkPlaceables.maxOfOrNull { it.height } ?: 0
+            val linksWidth = linkPlaceables.maxOfOrNull { it.width } ?: 0
+
+            val contentHeight = textHeight + (if (linksHeight > 0) spacingPx else 0) + linksHeight
+            val totalHeight = maxOf(iconHeight, contentHeight)
```
rock3r, line 1309: I think the layout should also coerce the measured height into the incoming constraints. Width is already handled below with `coerceIn(constraints.minWidth, constraints.maxWidth)`, but height is passed to `layout(...)` as the raw content height, so something like `Modifier.heightIn(min = 100.dp)` won't affect the measured banner height.

Could we mirror the width handling while keeping the current top-aligned placement? Something like:

```kotlin
val totalHeight = maxOf(iconHeight, contentHeight)
val layoutHeight = totalHeight.coerceIn(constraints.minHeight, constraints.maxHeight)

layout(naturalWidth, layoutHeight) {
    // existing Y positions unchanged; content stays top-aligned
}
```

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUIBannerStyling.kt
```diff
@@ -183,7 +185,15 @@ public object IntUiInlineBannerStylesFactory {
         InlineBannerStyles(information = information, success = success, warning = warning, error = error)
 }
 
-public fun BannerMetrics.Companion.default(borderWidth: Dp = 1.dp): BannerMetrics = BannerMetrics(borderWidth)
+@Deprecated("Use the `default()` function with `cornerSize` and `paddingValues`", level = DeprecationLevel.HIDDEN)
```
rock3r: Could we make the release note for this one a bit more specific?

This hidden overload is deprecated for binary compatibility, but source calls like `BannerMetrics.default()` / `BannerMetrics.default(borderWidth = 1.dp)` should continue to resolve to the new visible overload, so this isn't source-breaking in the same way as the constructor change.

Maybe the release note can say that the factory now also accepts `cornerSize` and `paddingValues`, and that the old binary overload is preserved/hidden, rather than implying normal source users will see a deprecation for `default(borderWidth)`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
rock3r: Have you forgotten to? Should this be a `Box(propagateMinConstraints = true)` now?

The custom `Layout` here is effectively acting as a one-child box that propagates the incoming constraints. I checked with a small harness locally, and `Box(propagateMinConstraints = true)` gives the same min-constraint behavior:

```kotlin
@Composable
private fun Current(modifier: Modifier, content: @Composable () -> Unit) {
    Layout(modifier = modifier, content = content) { measurables, constraints ->
        val placeable = measurables.first().measure(constraints)
        layout(placeable.width, placeable.height) { placeable.placeRelative(0, 0) }
    }
}

@Composable
private fun Replacement(modifier: Modifier, content: @Composable () -> Unit) {
    Box(modifier = modifier, propagateMinConstraints = true) { content() }
}
```

The standard `Box` version would be a bit easier to read, and less maintenance burden

rock3r (changes requested):

platform/jewel/ui/metalava/ui-api-0.36.0.txt
DanielSouzaBertoldi: This is most likely happening because there's changes to `InlineBanner` which resides in the `ui` module, so the script just checks for any API changes in the entire module and outputs it.

I can delete unrelated API changes from the dump file, but that will break the "Check for breaking API changes with Metalava" job, so keep that in mind.

Maybe a PR dedicated to updating all metalava dumps with the current code state would make these situations more managable, I'll see if I can raise a PR for this.

-- Also, I just realized I'll have to update the metalava dumps anyways because Jewel is at v0.37, not 0.36

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
DanielSouzaBertoldi: Hmm then maybe we could just not hide it. That would solve binary and source, with source raising a warning that this API is now deprecated and users should migrate.

Actually, that makes the most sense because then the `replaceWith` parameter would actually be useful. Wdyt?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUIBannerStyling.kt
DanielSouzaBertoldi: Same point as the last one for the BannerMetrics deprecated constructor. Do not hide it, add a `replaceWith` and that's that.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
DanielSouzaBertoldi: I stashed the work and never committed it...... 🤡 Yeah, a Box is way better than a Layout for this

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
DanielSouzaBertoldi: About scope, we can merge #3466 in the same release, then that's a non issue 😎

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
DanielSouzaBertoldi: Great catch! Fixed 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
DanielSouzaBertoldi: I took the liberty of doing just that but please lmk your thoughts, I can revert the changes 😛

platform/jewel/ui/metalava/ui-api-0.36.0.txt
DanielSouzaBertoldi: Had the same problem as in the [Badge PR](https://github.com/JetBrains/intellij-community/pull/3365#discussion_r3196649193). I decided to remove the API changes on [Horizontally|Vertically]ScrollableContainer this time, tho

rock3r: Coordinate with Nebojsa, just in case he's also doing it

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
rock3r: As long as now we don't have resolution ambiguity because two non-hidden overloads can resolve `BannerMetrics(1.dp)`, I'm ok with that too

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUIBannerStyling.kt
rock3r: Same counterpoint about making sure it doesn't cause resolution ambiguity :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
rock3r: Happens!

rock3r:
Ready to merge

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/BannerStyling.kt
DanielSouzaBertoldi: `BannerMetrics` can't have resolution ambiguity because the new primary constructor requires all three parameters with no defaults

`BannerMetrics.default()` will always resolve to the deprecated method unless users specify one of the new 2 parameters :( but such is life, eventually we will get rid of the deprecated .default()

rock3r (approved):

PR #3392, opened 2026-01-22
[JEWEL-849] Add support to filter items with FilterableLazyColumn

faogustavo:
> [!warning]
> This PR is the continuation of #3361. The only valid commit from this PR is the last one

- The IDE has one pattern for searching on lists that use a TextField on top of the content
- The search text is then used to filter the content from the list below
- To support that feature, we will introduce a few new components to Jewel
- The first one is the SearchTextField
  - The main goal is to render the search text input, with the correct icons and behaviors
  - About behavior, for now, only the history popup and the clear actions are supported
- The second one is the SearchArea
  - The main goal is to render the search input, with a divider and an additional content passed in the slot
  - The idea is to replicate SpeedSearchArea so users could expand the idea as needed, with their own components
- The third and last is the FilterableLazyColumn
  - The goal is to render the selectable lazy list with additional support for filtering
  - It automatically handles callbacks (such as onSelectedIndexesChange) that should return the external index, instead of the filtered one
  - It also handles auto scroll to the selected entry on typing, and if none of the selected items are in the filtered content, select a new one to ensure the user has a valid selection

### SearchTextField

Screen Recording 2026-01-22 at 12 01 53

### FilterableLazyColumn

Screen Recording 2026-01-22 at 12 08 58

### Speed Search

| Before | After |
| --- | --- |
| image | image | 
| Screen Recording 2026-01-22 at 14 05 49 | Screen Recording 2026-01-22 at 14 04 28 | 

### New features
 * Created `SearchTextField` component
   * It's a text field with the "Search" look and feel
   * It replicates the behavior of the Swing [SearchField](https://plugins.jetbrains.com/docs/intellij/search-field.html)
   * The only features supported in the initial release are the "History" pop-up and clearing the content
   * This component is being used in both the SpeedSearchArea and SearchArea components
 * Created `SearchArea` component
   * It's a container for the filter type of interaction on the list
   * The container is intended to be used as a way to implement the behavior on your custom components
   * We strongly encourage you to use this component instead of the `SpeedSearchArea` for places where you need to filter the content as you type the text
 * Created `FilterableLazyColumn` component
   * It's a combination of `SearchArea` and the `SelectableLazyColumn`
   * The goal is to support filtering the entries as you type

### ⚠️ Important Changes
 * `TextField` with `undecorated = true` now renders `leadingIcon` and `trailingIcon`. Previously the undecorated variant ignored both. If you pass icons to an undecorated text field, they will now be visible. Remove them to keep the old appearance.

---

> [!NOTE]
> **Medium Risk**
> Adds new experimental UI APIs (SearchArea, FilterableLazyColumn, SearchTextField styling) and adjusts selection/speed-search matching behavior, which could affect list navigation and filtering semantics across consumers. Changes are covered by new unit/UI tests but touch core interaction flows.
> 
> **Overview**
> Adds **experimental** `SearchArea` and `FilterableLazyColumn`, providing an integrated search input that filters `SelectableLazyColumn` content while remapping selection callbacks back to original (unfiltered) indices and auto-adjusting selection/scrolling when filters change.
> 
> Extends `SpeedSearchMatcher` to support `CharSequence` matching, introduces `EmptySpeedSearchMatcher`, and adds convenience `matches`/`filter` extensions for efficiently filtering iterables; updates matchers’ implementations accordingly.
> 
> Updates selection synchronization in `SelectableLazyColumn` to also react when SpeedSearch changes the active index, and adds theme/styling support for `SearchTextFieldStyle` (IDE bridge + IntUI standalone) plus showcase demos and comprehensive new UI tests for filtering and `dismissOnLoseFocus` behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b3a7bca0cb42543183bd0efeb06098741183df9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextField.kt
```diff
@@ -110,31 +110,20 @@ public fun TextField(
...
-                            Modifier.defaultMinSize(minWidth = minSize.width, minHeight = minSize.height)
-                                .padding(style.metrics.contentPadding),
-                        innerTextField = innerTextField,
-                        textStyle = textStyle,
-                        placeholderTextColor = style.colors.placeholder,
-                        placeholder = if (state.text.isEmpty()) placeholder else null,
-                        leadingIcon = leadingIcon,
-                        trailingIcon = trailingIcon,
-                    )
-                }
```
faogustavo, line 137: I don't know why, but this implementation was using two different decoration boxes, while the others used only the `TextFieldDecorationBox`. I've changed to always use the same to ensure we are rendering the leading/trailing icons correctly.

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/SpeedSearches.kt
```diff
@@ -270,6 +282,33 @@ private fun SpeedSearchListWithFiltering(modifier: Modifier = Modifier) {
...
+ */
+@Composable
+private fun SimpleSearchAreaSample(modifier: Modifier = Modifier) {
+    FilterableLazyColumn(modifier, onSelectedIndexesChange = { println("Selected index changed to $it") }) {
+        items(TEST_LIST, textContent = { it }, key = { it }) { item ->
+            LaunchedEffect(isSelected) {
+                if (isSelected) {
+                    JewelLogger.getInstance("SpeedSearches").info("Item $item got selected")
+                }
+            }
```
rock3r: Do we need this? We should use `onSelectedIndexesChange` for this, and in fact you already have a println there, so maybe just move the log there and remove the println

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/SelectableLazyColumn.kt
```diff
@@ -130,7 +130,10 @@ public fun SelectableLazyColumn(
     LaunchedEffect(container, state.selectedKeys) {
         val selectedKeysSnapshot = state.selectedKeys
         val indices = selectedKeysSnapshot.mapNotNull { key -> container.getKeyIndex(key) }
-        if (indices != lastEmittedIndices) {
+        // The `state.lastActiveItemIndex` can also be changed by the SpeedSearch flow.
+        // If the last active index is not among the selected indices, the SpeedSearch
+        // has selected a different value and an update must be triggered.
```
DanielSouzaBertoldi: I don't get it. How come SpeedSearch can select a different value that's not in the selected indices?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
DanielSouzaBertoldi, line 14:
```suggestion
     * Returns a [MatchResult.Match] with a list of ranges where the pattern matches, or [MatchResult.NoMatch]
```

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -12,6 +14,12 @@ public fun interface SpeedSearchMatcher {
      */
     public fun matches(text: String?): MatchResult
 
+    /**
+     * Returns a [MatchResult.Match] with a list of ranges from the where the pattern matches, or [MatchResult.NoMatch]
```
DanielSouzaBertoldi, line 20:
```suggestion
     * Returns a [MatchResult.Match] with a list of ranges where the pattern matches, or [MatchResult.NoMatch]
```

platform/jewel/foundation/src/test/kotlin/org/jetbrains/jewel/foundation/search/IterableFilterTest.kt
```diff
@@ -0,0 +1,59 @@
...
+            listOf(
+                User("John Doe", "john@example.com"),
+                User("Jane Smith", "jane@example.com"),
+                User("Johnny Cash", "johnny@example.com"),
+            )
+        val matcher = SpeedSearchMatcher.exactSubstringMatcher("Joana")
+        val result = users.filter(matcher) { it.name }
+
+        assertEquals(0, result.size)
+    }
```
DanielSouzaBertoldi, line 58: We have `SubstringSpeedSearchMatcherTest`, what do you think of moving these tests over there?

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridge.kt
```diff
@@ -104,6 +104,7 @@ internal fun createBridgeComponentStyling(theme: ThemeDefinition): ComponentStyl
         undecoratedDropdownStyle = readUndecoratedDropdownStyle(menuStyle),
         speedSearchStyle = readSpeedSearchStyle(),
         searchMatchStyle = readSearchMatchStyle(theme.isDark),
+        searchTextFieldStyle = readSearchTextFieldStyle(),
```
DanielSouzaBertoldi, line 115: No `filterableLazyColumn`? How can users customize it? By adjusting `SearchArea` and `SelectableLazyColumn`?

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/SpeedSearches.kt
```diff
@@ -0,0 +1,361 @@
...
+import org.jetbrains.jewel.ui.theme.colorPalette
+
+@Composable
+@Suppress("Nls")
+internal fun FilterAndSearch(modifier: Modifier = Modifier) {
+    Column(modifier, verticalArrangement = Arrangement.spacedBy(16.dp)) {
+        InlineWarningBanner(
+            text =
+                "One of the samples is using the SpeedSearch feature for filtering the content. " +
+                    "Despite being possible, we strongly recommend using a different component for this behavior."
```
DanielSouzaBertoldi, line 64: > we strongly recommend using a different component for this behavior.

Why? 😯

Would be cool to let the user the reasons why and which component they should use instead. Judging solely by the name seems like _**Speed**_ Search would be the best one. And maybe some reason why this one could be used

DanielSouzaBertoldi (commented):
There's a lot to unpack here, I'll wrap up my review shortly, until then, these are the questions I have so far 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/TextFieldStateExtension.kt
```diff
@@ -0,0 +1,175 @@
...
+import androidx.compose.ui.input.key.nativeKeyCode
+import androidx.compose.ui.input.key.type
+import androidx.compose.ui.input.key.utf16CodePoint
+import androidx.compose.ui.text.TextRange
+import java.awt.event.KeyEvent as AWTKeyEvent
+import org.jetbrains.skiko.hostOs
+
+internal fun TextFieldState.handleKeyEvent(
+    event: KeyEvent,
+    onTextChange: (CharSequence) -> Unit = {},
```
nebojsa-vuksic: nit: Let's move `onTextChange` param to the last place so we can call this extension function with a trailing lambda: 
```handleKeyEvent(...) { }```

File: platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/TextFieldStateExtension.kt [hunk 9]
```diff
@@ -0,0 +1,175 @@
...
+    allowNavigationWithArrowKeys: Boolean = true,
+    allowedSymbols: String? = null,
```
nebojsa-vuksic, line 25: It's a bit strange that allowed symbols is a String. I would expect some kind of an array or collection.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchArea.kt
```diff
@@ -0,0 +1,150 @@
...
+ * @param fallbackKeyEventHandler Fallback handler for unhandled key events from the text field.
+ * @param content The content to display below the search area.
+ */
+@Composable
+@ExperimentalJewelApi
+@ApiStatus.Experimental
+public fun SearchArea(
+    state: SearchAreaState,
+    modifier: Modifier = Modifier,
+    searchModifier: Modifier = Modifier,
```
nebojsa-vuksic: Hm, I feel it's bit risky to have two separate modifiers here. I would expect that modifier dictating size of container should also influence on the size of the `SearchTextField`. SearchTextField should match the container width. The API is designed in such way that there is a lots of area for misuse.

Let's simply pass `Modifier.fillMaxWidth()` to `SearchTextField`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
```diff
@@ -0,0 +1,339 @@
...
+                }
+            }
+            .flowOn(dispatcher)
+            .collect()
+    }
+
+    LaunchedEffect(itemToScrollTo) {
+        val index = itemToScrollTo ?: return@LaunchedEffect
+
+        state.scrollToItem(index + 2, true)
```
nebojsa-vuksic: Why `+ 2`? Do we have to check for possible index out of bounds error here?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
```diff
@@ -114,7 +117,84 @@ public class DefaultComponentStyling(
     public val undecoratedDropdownStyle: DropdownStyle,
     public val speedSearchStyle: SpeedSearchStyle,
     public val searchMatchStyle: SearchMatchStyle,
+    public val searchTextFieldStyle: SearchTextFieldStyle,
 ) : ComponentStyling {
+    @Deprecated("Use the variant with speedSearchStyle.", level = DeprecationLevel.HIDDEN)
```
nebojsa-vuksic: We should also update other deprecated constructor so that their deprecation message points to the constructor with `searchTextFieldStyle`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
```diff
@@ -0,0 +1,339 @@
...
+    reverseLayout: Boolean = false,
+    onSelectedIndexesChange: (List<Int>) -> Unit = {},
+    verticalArrangement: Arrangement.Vertical = if (!reverseLayout) Arrangement.Top else Arrangement.Bottom,
+    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
+    flingBehavior: FlingBehavior = ScrollableDefaults.flingBehavior(),
+    keyActions: KeyActions = DefaultSelectableLazyColumnKeyActions,
+    pointerEventActions: PointerEventActions = DefaultSelectableLazyColumnEventAction(),
+    interactionSource: MutableInteractionSource? = null,
+    dispatcher: CoroutineDispatcher = Dispatchers.Default,
+    content: FilterableLazyColumnScope.() -> Unit,
```
nebojsa-vuksic, line 90: I'm wondering if we should start grouping a common params into a data class? 🤔 
Number of params in our surface API is getting out of hand.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
```diff
@@ -0,0 +1,339 @@
...
+        textContent: (item: T) -> String?,
+        key: (item: T) -> Any,
+        contentType: (item: T) -> Any?,
+        selectable: (item: T) -> Boolean,
+        itemContent: @Composable (SelectableLazyItemScope.(item: T) -> Unit),
+    ) {
+        val filteredList = mutableListOf<T>()
+
+        for (item in items) {
+            if (filterText.isEmpty() || textContent(item)?.contains(filterText, ignoreCase = true) ?: false) {
```
nebojsa-vuksic: This is constant O(n) filtering of items. Maybe we should dispatching to the Default dispatcher?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -438,58 +522,99 @@ internal class SpeedSearchStateImpl(private val matcherBuilderState: State<(Stri
...
+        matcherBuilderFlow: Flow<(String) -> SpeedSearchMatcher>,
+    ) =
         combine(searchTextFlow, matcherBuilderFlow) { text, buildMatcher ->
-                val items = entriesFlow.value
+            val matcher =
+                if (text.isBlank()) {
+                    EmptySpeedSearchMatcher
+                } else {
+                    buildMatcher(text).cached()
+                }
```
nebojsa-vuksic: This seems like a feature envy code smell. Can we delegate initialization of `EmptySpeedSearchMatcher` to `buildMatcher` lambda implementation? I assume implementation comes for `foundation` module. This way we can mark `EmptySpeedSearchMatcher` private to `foundation` module.

nebojsa-vuksic (changes requested):
Waiting for changes

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/theme/IntUiTheme.kt
```diff
@@ -282,6 +283,7 @@ public fun ComponentStyling.dark(
     speedSearchStyle: SpeedSearchStyle = SpeedSearchStyle.dark(),
     searchMatchStyle: SearchMatchStyle = SearchMatchStyle.dark(),
     popupAdStyle: PopupAdStyle = PopupAdStyle.dark(),
```
rock3r, line 306: Same ABI concern here for the public standalone theme factories: adding `searchTextFieldStyle` changes the callable overload set for already-compiled clients. I think we need to preserve the previous public `light(...)` / `dark(...)` signatures as hidden compatibility overloads and delegate internally.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -111,48 +192,162 @@ public fun SpeedSearchArea(
...
+ * This interface provides access to the current search state and methods for controlling the search behavior. Use
+ * [rememberSpeedSearchState] to create an instance of this state.
+ */
 @ExperimentalJewelApi
 @ApiStatus.Experimental
 public interface SpeedSearchState {
+    /** The vertical position of the search input overlay (Top or Bottom). */
     public var position: Alignment.Vertical
+
+    /** The current search text entered by the user. */
```
rock3r, line 261: Experimental API should not break binary compatibility. Technically, this interface gained a new abstract member (`currentMatcher`), which will break implementors. I think this is low risk in this case, but let's be VERY careful.

Another note: `isVisible` changed from `val` to `var`, which adds a setter to the ABI. Any external implementation compiled against the previous version will break here — same issue.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
```diff
@@ -0,0 +1,339 @@
...
+                filteredList.add(item)
+                _entries.add(
+                    Entry(
+                        index = lastIndex++,
```
rock3r: `onSelectedIndexesChange` is documented as returning indices from the original unfiltered list, but `lastIndex` only advances for items that survive filtering, so `Entry.index` is currently tracking filtered position, not original position. The same issue seems to apply to `item(...)` and `stickyHeader(...)` below as well.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
```diff
@@ -0,0 +1,339 @@
...
+
+        combine(selectionFlow, renderableEntriesFlow) { selectedKeys, container ->
+                val shouldRefreshSelectionIfNeeded = selectedKeys.isNotEmpty() || searchState.searchText.isNotEmpty()
+
+                val validEntriesByKey = container.values.associateBy { it.key }
+                val selectedIndexes = selectedKeys.mapNotNull { validEntriesByKey[it]?.filteredIndex }
+                var currentSelectedIndexes = selectedIndexes
+
+                if (shouldRefreshSelectionIfNeeded && selectedIndexes.isEmpty()) {
+                    container[state.firstVisibleItemIndex]?.let {
```
rock3r: This fallback selection logic looks too optimistic: it picks the first visible entry without checking whether that entry is actually selectable. If the first visible item is a non-selectable header, this can put a non-selectable key into `selectedKeys`. I think this needs to choose the first selectable visible/renderable entry instead.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
rock3r: I think the problem here is worse than just `why + 2?` / possible bounds issues. `SelectableLazyListState.scrollToItem()` also updates `lastActiveItemIndex = itemIndex`, so passing `index + 2` here corrupts the logical active index as well. Near the end of the filtered list that can point past the actual selected item, which then affects subsequent keyboard navigation. If the goal is only to scroll with some visual slack, that needs to be separated from the logical active-item update.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
rock3r: I still think this concern is unresolved. The expensive filtering work is still happening inline while building `FilterableLazyColumnScopeImpl` (`contains(...)` over the whole item set on every change), i.e. on the composition path. The `dispatcher` parameter only affects the selection/scroll side-effect later on; it doesn’t move the actual filtering work off the main/UI path. So both the implementation and the KDoc still look misleading here.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUIBridgeSearchInputField.kt
```diff
@@ -0,0 +1,35 @@
...
+
+internal fun readSearchTextFieldStyle() =
+    SearchTextFieldStyle(
+        colors =
+            SearchTextFieldColors(
+                foreground = Color.Unspecified,
+                error = retrieveColor("Label.errorForeground", JBColor.RED.toComposeColor()),
+            ),
+        metrics =
+            SearchTextFieldMetrics(
```
rock3r, line 25: For the bridge side, I think at least the padding/insets should come from existing JBUI/LaF sources instead of being hardcoded here. We already have theme/JBUI support for related search-field metrics (`Editor.SearchField.borderInsets`, `Popup.SearchField.borderInsets`, `Popup.SearchField.inputInsets`). Swing does appear to hardcode the icon gap itself, so I’m not worried about mirroring the `2dp` icon spacing — but the insets/padding part looks like it should be sourced.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -154,6 +162,106 @@ public enum class MatchingCaseSensitivity {
...
+}
+
+/**
+ * Checks whether the given text matches the current [SpeedSearchMatcher] pattern.
+ *
+ * This is a convenience method that simplifies checking for matches by returning a boolean instead of requiring pattern
+ * matching on [SpeedSearchMatcher.MatchResult].
+ *
+ * Example:
+ * ```kotlin
```
rock3r, line 234: The example here still uses `doesMatch(...)`, but the actual API name is `matches(...)`.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchTextFieldStyle.kt
```diff
@@ -0,0 +1,162 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import org.jetbrains.jewel.foundation.GenerateDataFunctions
+import org.jetbrains.jewel.ui.icon.IconKey
+import org.jetbrains.jewel.ui.icons.AllIconsKeys
+
+@Immutable
+@GenerateDataFunctions
+public class SearchTextFieldStyle(
```
rock3r, line 24: This is new public theming API and should have KDoc.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchTextFieldStyle.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    override fun toString(): String = "SearchTextFieldStyle(colors=$colors, metrics=$metrics, icons=$icons)"
+
+    public companion object
+}
+
+@Immutable
+@GenerateDataFunctions
+public class SearchTextFieldColors(public val foreground: Color, public val error: Color) {
```
rock3r, line 63: Same here: public theming API, but undocumented.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchTextFieldStyle.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    override fun toString(): String = "SearchInputFieldColors(foreground=$foreground, error=$error)"
+
+    public companion object
+}
+
+@Immutable
+@GenerateDataFunctions
+public class SearchTextFieldMetrics(
```
rock3r, line 97: Same here: public theming API, but undocumented.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/SearchTextFieldStyle.kt
```diff
@@ -0,0 +1,162 @@
...
+            "popupContentPadding=$popupContentPadding, " +
+            "spaceBetweenIcons=$spaceBetweenIcons" +
+            ")"
+
+    public companion object
+}
+
+@Immutable
+@GenerateDataFunctions
+public class SearchTextFieldIcons(
```
rock3r, line 142: Same here: public theming API, but undocumented.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiSearchInputFieldStyling.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.jewel.intui.core.theme.IntUiDarkTheme
+import org.jetbrains.jewel.intui.core.theme.IntUiLightTheme
+import org.jetbrains.jewel.ui.component.styling.SearchTextFieldColors
+import org.jetbrains.jewel.ui.component.styling.SearchTextFieldIcons
+import org.jetbrains.jewel.ui.component.styling.SearchTextFieldMetrics
+import org.jetbrains.jewel.ui.component.styling.SearchTextFieldStyle
+import org.jetbrains.jewel.ui.icon.IconKey
+import org.jetbrains.jewel.ui.icons.AllIconsKeys
+
+public fun SearchTextFieldStyle.Companion.light(
```
rock3r, line 24: All of the public factory helpers in this file are part of the new public styling surface, but the file is entirely undocumented. We should document these just like the rest of the standalone styling API.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchArea.kt
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * State holder for [SearchArea], managing search visibility and text input.
+ *
+ * Wraps [SearchTextFieldState] with automatic visibility management. The search area shows when text is present and can
+ * be dismissed with Escape.
+ *
+ * Use [rememberSearchAreaState] to create an instance.
+ */
+public class SearchAreaState internal constructor(internal val searchState: SearchTextFieldState) {
```
rock3r, line 91: Is it intentional that `SearchArea` itself is experimental, but `SearchAreaState` and `rememberSearchAreaState()` landed in stable API? That freezes the state API earlier than the widget API, which may make iteration harder if this surface is still evolving. All this to say — I think this should be experimental too :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
rock3r: Yes and no — we can but I'd rather wait for the new Compose styling API to properly land and move over to those :) And for lambdas/callbacks, I don't think we should nest them into objects, it makes the APIs cumbersome to use imho

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
rock3r: Just noticed Bugbot also picked this up lol

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchArea.kt
rock3r: Let's not hardcode `fillMaxWidth()` or `weight()` ever, because they force users to set explicit sizes at usage sites. A better approach is a `defaultMinSize` here, like we do for buttons/text fields

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextField.kt
rock3r: Probably just forgot to update this one LOL Can we make sure we don't cause regressions?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
```diff
@@ -118,8 +121,9 @@ public class DefaultComponentStyling(
     public val speedSearchStyle: SpeedSearchStyle,
     public val searchMatchStyle: SearchMatchStyle,
     public val popupAdStyle: PopupAdStyle,
+    public val searchTextFieldStyle: SearchTextFieldStyle,
```
rock3r, line 185: This changes the stable public constructor shape again. I don’t see a compatibility constructor preserving the previous signature that ended at `popupAdStyle`, so this looks like an ABI break for already-compiled clients (`NoSuchMethodError` territory). Binary compatibility is a hard requirement here — can we add the old constructor shape as hidden secondary ctor and delegate to the new one with a fallback `SearchTextFieldStyle`?

rock3r (changes requested):

rock3r:
Nice direction overall, but I still see a few things that I think need to be addressed before this is safe to merge:

- **ABI compatibility:** adding `searchTextFieldStyle` in stable public styling/theme APIs looks like an ABI break unless the previous public signatures are kept as compatibility overloads/constructors. The `SpeedSearchState` changes also look ABI-breaking for external implementations.
- **`FilterableLazyColumn` logic:** the original-index tracking seems wrong after filtering, the fallback selection logic can pick non-selectable entries, and `scrollToItem(index + 2)` appears to corrupt the logical active index used for subsequent navigation.
- **Filtering/performance/KDoc mismatch:** the expensive filtering work still seems to happen inline on the composition path, so the current `dispatcher`/KDoc wording still looks misleading.
- **Bridge + docs:** for the bridge side, at least the padding/insets should probably come from existing JBUI/LaF sources instead of being hardcoded. Also, the new public styling/theme surface and the `SpeedSearchMatcher` docs still need documentation fixes.
- **Clarifications requested:** the broader undecorated `TextField` behavior change looks regression-risky, and I’m also not sure whether it was intentional for `SearchAreaState` / `rememberSearchAreaState()` to be stable while `SearchArea` itself is still experimental.

Details are in the inline comments.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/util/TextFieldStateExtension.kt
DanielSouzaBertoldi: Why is that? I think it's best to be able to pass all characters we want to be allowed as a String instead of `listOf(Symbol.Hashcode, Symbol.Underline)` etc or `listOf('*', '_', '-')`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
DanielSouzaBertoldi: I mean, moving the `if` block and the `cached()` ext fun over to `foundation` is fine, they are more appropriate there. But we can't make `EmptySpeedSearchMatcher` private because it is also used to initialize `currentMatcher`.

We could expose a companion property instead of `EmptySpeedSearchMatcher` directly but that to me seems more just indirection for the sake of making a object internal.

What do you think?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/theme/IntUiTheme.kt
DanielSouzaBertoldi: Hmmm I'm a bit confused. We added the `searchTextFieldStyle` to the not deprecated `.dark()`. Below there's the variant without `searchTextFieldStyle` in the parameter list

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchArea.kt
DanielSouzaBertoldi: Let's keep it consistent by making both experimental 🙏🏻

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
DanielSouzaBertoldi: I dont know what happened, but the deprecated message in 134L is wrong. It should be 'Use the variant with searchTextFieldStyle'.

This PR did add a deprecated constructor that stops at `popupAdStyle`, is at line 222 🤔 

Looks like there was some sort of problem when merging the PR that adds the slim button styles or somethiing

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchArea.kt
DanielSouzaBertoldi: The thing though, is that the `SearchTextField` Swing version sets the size of the textfield as `setColumns(15)` so the preferred width is 15 characters widths (if I'm understanding the code right). Since it's font-dependent there's no reliable way to define a default min size on the Compose end.

We can get rid of the `searchModifier` and keep `fillMaxWidth()` to `SearchTextField` modifier. Users can then decide on having a fixed size for `SearchArea` or not.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/FilterableLazyColumn.kt
DanielSouzaBertoldi: Fixed by splitting the work in two: the DSL now only collects item descriptors during composition (no filtering), while a background coroutine (flowOn(dispatcher)) uses combine + snapshotFlow to do the O(n) text matching and produce a set of visible keys. Composition then does O(1) key lookups against that set.

[Check it out here](https://github.com/JetBrains/intellij-community/pull/3392/changes#diff-89c6f57ed8054b7dd1f4612d48cff2ff36c790303666d7fc68a59b37088de1feR113)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextField.kt
rock3r: @DanielSouzaBertoldi I think this one is still worth checking. The thread is current, and the state-based `TextField` still always passes `TextFieldDecorationBox` as the decorator even when `undecorated = true`; `InputField` skips the background/border/focus bits, but the decorator still applies the min-size/content-padding/layout path. Could we make sure the undecorated behavior did not regress here?

rock3r:
@DanielSouzaBertoldi One previous ABI comment is now on an outdated thread, but I think the concern still applies: https://github.com/JetBrains/intellij-community/pull/3392#discussion_r2914037141

`SpeedSearchState` is a public experimental interface, and the current version still adds a new abstract `currentMatcher` property and changes `isVisible` from `val` to `var`. That can still break external implementations compiled against the previous API. Could you take another look?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchArea.kt
rock3r: We can use an approx Dp width as defaultMinSize for this, it's probably fine. If you wanna be nitpicky, you can do it the same way Swing does it: measure the width of `m` and multiply it by 15 to get the width :D

rock3r (changes requested):

DanielSouzaBertoldi:
Alright I'm finally back to this PR, let's close this one for good

DanielSouzaBertoldi:
@rock3r soooooo good news and bad news.... :/

Good news is: the `currentMatcher` change is done, now it defaults to `EmptySpeedSearchMatcher`.


The bad news is that the `val` -> `var` change was merged way back in [JEWEL-1024](https://github.com/JetBrains/intellij-community/pull/3361) and shipped in 0.37 😬 Looking at the release notes we did not acknowledge this breaking change

rock3r:
Can we retroactively add it to the release notes? 😁

DanielSouzaBertoldi:
@rock3r Sure thing! You can check it out here: https://github.com/JetBrains/intellij-community/pull/3392/changes#diff-5c746a61978ec8c1460e36461be0e0f2f985adf767edabd5bf9c2fcff06b51f0

rock3r (approved):

nebojsa-vuksic (approved):

DanielSouzaBertoldi:
@rock3r @wellingtoncosta @nebojsa-vuksic it's been too long since this was last reviewed. I resolved conflicts, but better to be safe than sorry. Can you guys take a look at this again pretty please? 🥺

wellingtoncosta:
@DanielSouzaBertoldi are all (or most of) comments addressed? Asking because I see a lot of comments not marked as resolved. Would be nice to keep them up to date.

DanielSouzaBertoldi:
@wellingtoncosta yes all comments have been addressed.

I can't close them because I am not the PR's author

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextField.kt
rock3r: I think this is still the case? Can you check?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchArea.kt
rock3r: Ditto, this still isn't addressed

rock3r:
> You can check it out here: https://github.com/JetBrains/intellij-community/pull/3392/changes#diff-5c746a61978ec8c1460e36461be0e0f2f985adf767edabd5bf9c2fcff06b51f0

@DanielSouzaBertoldi the release notes change seems gone

rock3r (changes requested):

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -116,6 +120,37 @@ public fun interface SpeedSearchMatcher {
...
+ *
+ * This is useful when the same matcher is applied repeatedly over a stable set of items, e.g., when re-filtering a list
+ * on every recomposition.
+ *
+ * @return A caching wrapper around this matcher, or this matcher itself if it is [EmptySpeedSearchMatcher].
+ */
+public fun SpeedSearchMatcher.cached(): SpeedSearchMatcher =
+    this as? EmptySpeedSearchMatcher
+        ?: object : SpeedSearchMatcher {
+            private val cache = LRUCache<CharSequence, SpeedSearchMatcher.MatchResult>(100)
```
wellingtoncosta, line 138: Shouldn't it be keyed to a `String` instead, since `CharSequence` does not implement `equals()`?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchTextField.kt
```diff
@@ -0,0 +1,479 @@
...
+ * Use this overload to share the [TextFieldState] with other components or for greater control over state
+ * initialization.
+ *
+ * @param textFieldState The [TextFieldState] to use. Defaults to a new empty state.
+ * @return A remembered [SearchTextFieldState] instance.
+ */
+@Composable
+public fun rememberSearchTextFieldState(
+    textFieldState: TextFieldState = rememberTextFieldState()
+): SearchTextFieldState = remember { SearchTextFieldState(textFieldState) }
```
wellingtoncosta, line 279: Shouldn't the `remember` be keyed to `textFieldState`?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
```diff
@@ -116,6 +120,37 @@ public fun interface SpeedSearchMatcher {
     }
 }
 
+/**
+ * Wraps this matcher so that match results are reused for texts that have already been matched.
+ *
+ * Results are held in an LRU cache of up to 100 entries, keyed by the matched text. Null or blank text is never cached
+ * and is always delegated to the original matcher. [EmptySpeedSearchMatcher] is returned unwrapped, as it matches
+ * everything and has nothing worth caching.
```
wellingtoncosta, line 128: This says `EmptySpeedSearchMatcher` "matches everything", but it returns `MatchResult.NoMatch` for every input (L225), and its own KDoc at L207 says it "never matches any text". Which semantic is intended?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchTextField.kt
```diff
@@ -0,0 +1,479 @@
...
+    outputTransformation: OutputTransformation? = null,
+    undecorated: Boolean = true,
+    popupStyle: PopupContainerStyle = JewelTheme.popupContainerStyle,
+    history: List<CharSequence> = emptyList(),
+    error: Boolean = false,
+    allowClear: Boolean = true,
+    fallbackKeyEventHandler: (KeyEvent) -> Boolean = { false },
+) {
+    val density = LocalDensity.current
+    state.updateHistoryEntries(history)
```
wellingtoncosta, line 133: Since `updateHistoryEntries()` mutates Compose state, shouldn’t this be performed in a `SideEffect` rather than directly during composition?

wellingtoncosta (commented):
Few questions / nits

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
DanielSouzaBertoldi: Yeah, caching an object like CharSequence did break the cached intention of this function. Deleted the `matches(text: CharSequence?)` function and added 4 unit tests

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/search/SpeedSearchMatcher.kt
DanielSouzaBertoldi: EmptySpeedSearchMatcher matches nothing. Updated KDoc

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SearchTextField.kt
DanielSouzaBertoldi: I noticed we are unwrapping the SpeedSearchState and rewrapping into a SearchTextFieldState when SpeedSearchState already holds a SearchTextFieldState.

Got rid of this "roundabout" and streamlined the code to just use the SpeedSearchState values directly

rock3r:
@DanielSouzaBertoldi my comments are still unanswered, are they addressed?

DanielSouzaBertoldi:
@rock3r this entire PR has a huge pile of comments lol i only read comments after i asked for reviews a ~fourth~ second time last week. Are you asking about the RELEASE NOTES one? If so then yes, I added it back 😛 

any comments prior to my rebase last week have been addressed too back then

Maybe you added comments but didnt publish tem during review? 👀

rock3r:
At this point I don't even know if I was looking at this PR helppppp

rock3r:
I asked my second brain :)

* [Undecorated spacing](https://github.com/JetBrains/intellij-community/pull/3392#discussion_r3990388211) remains: TextField.kt:113–126 applies minimum size and padding unconditionally. InputField forwards that decorator unchanged, even with undecorated = true.
* [Forced width remains](https://github.com/JetBrains/intellij-community/pull/3392#discussion_r3990400881): SearchArea.kt:70 still uses Modifier.fillMaxWidth(), not the default minimum sizing you requested.

DanielSouzaBertoldi:
@rock3r 

1. Got fixed by keeping the unified decorator and gating only the padding ✅ 
2. I'm not exactly sure what you meant by "default min size", so I took a look at how Swing's `com.intellij.ui.SearchTextField` handles this and it does:

```
myTextField.setColumns(15); // default minimum
//....
add(myTextField, BorderLayout.CENTER);
```

So they set a preferred size of 15 columns but they do stretch it to fill the panel.

In fix 1 (`TextField`) we are applying the default minimum size, and in `SearchArea` we set it to fill the max width. If I remove the `fillMaxWidth` modifier, then we get this:

image

Which does not match Swing :/

Or maybe I just misunderstood your second bullet point 😄

rock3r:
Ah, ok that makes sense. I have forgotten what I meant in the meantime lol sorry but maybe I had misunderstood the code or had the wrong assumption on how Swing worked (likely had not checked!). Padding is correct, I see. However, I'm wondering whether the min size makes sense when undecorated — I'd think perhaps not much?

DanielSouzaBertoldi:
@rock3r yeah, thinking about this I agree. If we are not setting paddings then we should not set a default min size too. Besides, it's an easy fix. Updated the code 😎 

Had to also add a `defaultMinSize` modifier to `SearchTextField` otherwise it would've looked too cramped, given that it sets `undecorated = true` by default:

image

rock3r (approved):

DanielSouzaBertoldi:
Here we go once again 🙏🏻 Ready to merge

PR #3395, opened 2026-01-23
[JEWEL] Update Detekt and enable checks for public surface documentation

faogustavo:
- The initial idea of this PR was basically to start using the existing detekt rules to ensure we add public documentation. They are:
  - OutdatedDocumentation
  - UndocumentedPublicClass
  - UndocumentedPublicFunction
  - UndocumentedPublicProperty
- What I was trying to do is to enable a "warning level" report, and reuse that information to comment in the PR in case any of the "issues" are related to the modified code
- But, when one of these conditions matches, the version 1.x from detekt considers it a failure and crashes the run. Some features, like setting the minimum version to "break", are only added in the 2.x version of detekt, making it impossible to implement the feature without the version bump
- After fighting detekt for a while, I was able to migrate it to the new version, ensure all tests from our detekt-plugin work, etc
- However, 2.x is still in alpha and has some bugs:
  - It does not auto-correct the code based on some interactions with the buildUponDefaultConfig (https://github.com/detekt/detekt/issues/8988)
  - The auto-correct does not use the same APIs to write the report, having to make changes to our rules to make it work (https://kotlinlang.slack.com/archives/C88E12QH4/p1768582000454989)
    - This one is probably fixed by now, but needs more testing
    - See `JewelBaseRule` and `JewelRuleSetProvider.init`
  - The auto-correct has some new properties that are formatting the code in a different way
    - This one needs more tests and review our config file comparing to the base version

> [!note]
> TL;DR; The work is like 60~65% done. We need to revisit the documentation for v2 and update our config file to ensure we have everything under control. After wrapping up the detekt, we will need to get the report and combine with the changes in the PR to identify which entries the PR could resolve.


platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/JewelBaseRule.kt
```diff
@@ -0,0 +1,34 @@
...
+abstract class JewelBaseRule(config: Config, description: String, url: URI? = null) : Rule(config, description, url) {
+    private lateinit var workingFile: KtFile
+    private lateinit var originalFile: KtFile
+
+    override fun visit(root: KtFile) {
+        if (autoCorrect) {
+            // Create a writable copy of the file for autocorrect
+            val fileCopy =
+                KtPsiFactory(root.project)
+                    .createPhysicalFile(fileName = root.name, text = root.modifiedText ?: root.text)
```
DanielSouzaBertoldi, line 19: Don't we need to remove this copy after the autocorrect is done running?

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/EqualityMembersRule.kt
```diff
@@ -116,10 +109,10 @@ class EqualityMembersRule(config: Config) : Rule(config) {
                             if (it == null) {
                                 hasAll = false
                                 report(
-                                    CodeSmell(
-                                        issue,
+                                    Finding(
                                         Entity.from(declaredFunction),
                                         "Function ${declaredFunction.name} is missing property $prop.",
+                                        suppressReasons = listOfNotNull("Auto correct".takeIf { autoCorrect }),
```
DanielSouzaBertoldi, line 115: nit: create a const val with the "Auto correct" string

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/EqualityMembersRule.kt
```diff
@@ -26,22 +22,18 @@ import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
...
-            Severity.Defect,
-            "This rule detects missing or incomplete equals/hashCode/toString functions.",
-            Debt.FIVE_MINS,
-        )
+class EqualityMembersRule(config: Config) :
+    JewelBaseRule(config, "This rule detects missing or incomplete equals/hashCode/toString functions.") {
 
     private val functionsToCheck: List<String> by config(defaultValue = listOf("equals", "hashCode", "toString"))
+
+    @Configuration("only check classes annotated with these annotations")
```
DanielSouzaBertoldi, line 36: What's up with this `@Configuration` annotation?

platform/jewel/detekt-plugin/src/main/kotlin/org/jetbrains/jewel/detekt/rules/JewelBaseRule.kt
```diff
@@ -0,0 +1,34 @@
...
+            workingFile = fileCopy
+            originalFile = root
+        } else {
+            workingFile = root
+            originalFile = root
+        }
+
+        super.visit(workingFile)
+
+        if (autoCorrect && workingFile.modificationStamp > 0) {
```
nebojsa-vuksic, line 30: Won't `workingFile.modificationStamp > 0` always evaluate to `true`?

platform/jewel/buildSrc/src/main/kotlin/jewel-linting.gradle.kts
```diff
@@ -1,15 +1,17 @@
...
-    id("io.gitlab.arturbosch.detekt")
+    id("dev.detekt")
     id("org.jmailen.kotlinter")
     id("com.ncorti.ktfmt.gradle")
 }
 
 detekt {
+    buildUponDefaultConfig = true
     autoCorrect = true
+    debug = true
```
nebojsa-vuksic, line 12: What's the idea behind having `debug` enabled? Is this a leftover from the current work?

Having it enabled by default will clutter CI logs with verbose logs from Detekt.

nebojsa-vuksic:
```suggestion
    debug = providers.gradleProperty("detekt.debug").map { it.toBoolean() }.getOrElse(false)
```

DanielSouzaBertoldi:
This PR will be closed in favor of #3553

PR #3399, opened 2026-01-27
IJPL-232015 MCP Server: persist resolved port on enable

codefromthecrypt:


codefromthecrypt:
@nerzhulart is there interest in this? if so I can rebase

PR #3405, opened 2026-02-01
[vcs-log] restore "All" action to clear the filters

juliabeliaeva:
"All" action was removed in 3a774b5000532726e8ee9afbbfedf27c135cfce7, however it made it very inconvenient to use the filters from Quick Actions Popup. I don't see how having "All" action would interfere with anything, and it makes keyboard access significantly easier. So this PR restores the action.

[IJPL-89255](https://youtrack.jetbrains.com/issue/IJPL-89255)


iasemenov:
The desired behavior was implemented in a slightly different way
image

PR #3406, opened 2026-02-04
Added delay in startup.kt to account for restart time, if existing IDE process is exiting

namratapal-google:
IJPL-179219, IJPL-185318: Start failed issues

### DIFF
--- a/platform/platform-impl/bootstrap/src/com/intellij/platform/ide/bootstrap/startup.kt
+++ b/platform/platform-impl/bootstrap/src/com/intellij/platform/ide/bootstrap/startup.kt
@@ -117,6 +117,10 @@ fun startApplication(
   val isHeadless = AppMode.isHeadless()
 
   val lockSystemDirsJob = scope.launch {
+    // Adding startup delay to account for restart process, if existing IDE instance is exiting 
+    // Ref: https://youtrack.jetbrains.com/issue/IJPL-179219, https://youtrack.jetbrains.com/issue/IJPL-185318 
+    delay(2000)
+    
     // the "import-needed" check must be performed strictly before IDE directories are locked
     configImportNeededDeferred.join()
     span("system dirs locking") {


trespasserw:
I've acknowledged the PR, but I don't think artificially slowing down the startup is the best solution. I'll take a deeper look later.

namratapal-google:
More context on this:

1. We are investigating Android Studio startup failure issues raised by our end-users.
2. On investigation, we found that the error directly originates from Intellij, not Android Studio code. Stacktrace is same as IJPL-185318. 
3. Our analysis is that the shutdown process is slow, might hang up due to below dependencies, preventing a restart in IDE restart scenarios: (e.g restart scenario: 'Update' in Jetbrains Toolbox shuts down existing process and tries to start new instance)

Dependencies during IDE shutdown (refer ApplicationImpl.destructApplication): 
a. Wait for user confirmation for shutdown when any process (sync / index / any process in progress in terminal) is in progress.
b. AppLifecycleListener(s) - appClosing(), beforeAppWillBeClosed(), appWillBeClosed() - that perform operations before app can be closed.
c. Stop service preloading with ComponentManagerImpl.stopServicePreloading.
d. Save app-level settings.
e. Save project settings.
f. Dispose application container.
g. Attempt to finally finish closing event.

4. Due to the above 7 or more operations during IDE shutdown process, our analysis is that we have only 3 below options to fix startup failure: 
- Account for each of the above 7 operations and their statuses during startup process, and halt startup until earlier process is fully shutdown. (ApplicationImpl.isExitInProgress can probably be used to halt startup, but whether isExitInProgress accounts for all above 7 operations or not is something to be checked) 
- If existing process is still open during IDE startup, force kill the process and proceed with startup
- Add few seconds hard-coded wait to alleviate the issue 

While I have implemented the 3rd option from above, I think 1st option might be the best one. 
LMK your thoughts, will make changes accordingly.

trespasserw:
I fully acknowledge that the shutdown sequence is a bit overloaded and might be slow, but it is hardly the major source of this problem. The Toolbox app and the restart helper both wait until the OS process finishes. Besides, the problem is relatively rare, but the suggested fix taxes each start.

I'm closing the request in favor of #3498.

PR #3407, opened 2026-02-04
[JEWEL-993] Add MacOS Scrollbar Detection to Standalone

DanielSouzaBertoldi:
# Context
After #3372  was merged, we finally have access to JNA in the `:int-ui-standalone` module. This means we can safely fetch data directly from MacOS settings and use it in our components 🥳 

This PR basically just changes `ScrollbarStyle.Companion.macOsLight()` and `ScrollbarStyle.Companion.macOsDark()` to fetch the user's current option from the new `StandaloneScrollbarHelper` class instead of hardcoding an option.

I also added a minor bonus: now you can change the configurations of all scrollbars in our `Scrollbar` showcase, changing from "manual" mode to "system", check the screenshots below!

# Changes
- Now that both `StandalonePlatformCursorController` and `StandaloneScrollbarHelper` use JNA to fetch style information, I opted to just move all JNA code from `:decorated-window` over to `:int-ui-standalone`, since the code we had there was more robust and `:decorated-window` depends on `:int-ui-standalone` anyway.
- Added new methods to Jewel's `Foundation` class so that we can add observers via JNA.
- Updated `Scrollbars` showcase code to add an option for scrollbars to follow the behavior/visibility of the system.
- Added a new icon called "Terminal"

# Screenshots

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/fc785b33-4289-4996-be4e-46e8ad4ba732" /> | <video src="https://github.com/user-attachments/assets/917f9b0f-723f-4afe-acd5-84c15d5fb572" /> |

## Scrollbars Showcase
| Standalone | Bridge |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/70f560c3-3b49-4fd0-9d00-378beba6a7f0" /> | <video src="https://github.com/user-attachments/assets/47d1fab1-36e6-4e5c-899f-01012d1f3def" /> |

### ⚠️ Important Changes
 * Now, by default, both Scrollbars and Tabs for MacOS respect the user chosen option in their MacOS settings. If you don't want this, please provide a fixed style, either `ScrollbarVisibility.WhenScrolling` or `ScrollbarVisibility.AlwaysVisible`.


platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/macos/Foundation.kt
```diff
@@ -0,0 +1,141 @@
...
+        val selector = createSelector(stringSelector)
+        if (!invoke(cls, "respondsToSelector:", selector).booleanValue()) {
+            error("Missing selector $stringSelector for $stringCls")
+        }
+        return invoke(cls, selector, *args)
+    }
+
+    operator fun invoke(id: ID?, selector: String?, vararg args: Any?): ID = invoke(id, createSelector(selector), *args)
+
+    fun allocateObjcClassPair(superCls: ID, name: String) =
```
DanielSouzaBertoldi, line 178: All code here was taken directly from `com.intellij.ui.mac.foundation.Foundation.java` class and converted to Kotlin.

However, I did not import 100% of the code there

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/StandaloneScrollbarHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+        get() = readScrollbarVisibility()
+
+    init {
+        listenToTrackClickBehaviorChange()
+        listenToScrollbarVisibilityChange()
+    }
+
+    private fun listenToTrackClickBehaviorChange() {
+        callMac {
+            // Copied from MacScrollBarUI
```
DanielSouzaBertoldi: Unfortunately, as is the case with `ScrollbarHelper` from bridge, we can't directly use the code in `MacScrollBarUI` since all its classes are `internal`.

This means this code is duplicated. We have the same lines over in bridge, but such is the case with our structure. The only difference is that bridge uses IJP's version of `Foundation`.

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Scrollbars.kt
```diff
@@ -148,36 +167,89 @@ private fun SettingsRow(
...
     onReducedContentChange: (Boolean) -> Unit,
     onClickBehaviorChange: (TrackClickBehavior) -> Unit,
+    onConfigurationChange: (ScrollbarConfiguration) -> Unit,
 ) {
-    Row(verticalAlignment = Alignment.CenterVertically) {
-        CheckboxRow(checked = alwaysVisible, onCheckedChange = onAlwaysVisibleChange, text = "Always visible")
+    Column {
+        Text("Configuration", style = JewelTheme.typography.h2TextStyle)
+
+        Spacer(Modifier.height(16.dp))
```
DanielSouzaBertoldi, line 179: We can't set the vertical arrangement to `spacedBy(16.dp)` in our parent `Column` because that screws up the animation:

https://github.com/user-attachments/assets/0430a061-f146-4f1f-81cf-ff7571fa223f

See? it kinda jumps at the end. That's because the parent Column has to get rid of the space that was reserved for the settings row after it's done animating :(

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/StandaloneScrollbarHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+import org.jetbrains.jewel.intui.standalone.styling.default
+import org.jetbrains.jewel.intui.standalone.styling.macOs
+import org.jetbrains.jewel.intui.standalone.window.macos.Foundation
+import org.jetbrains.jewel.intui.standalone.window.macos.ID
+import org.jetbrains.jewel.ui.component.styling.ScrollbarVisibility
+import org.jetbrains.jewel.ui.component.styling.TrackClickBehavior
+import org.jetbrains.skiko.OS
+import org.jetbrains.skiko.hostOs
+
+internal interface ScrollbarHelper {
```
nebojsa-vuksic: Please add kdoc

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/StandalonePlatformCursorController.kt
```diff
@@ -12,45 +11,24 @@ import org.jetbrains.skiko.hostOs
...
-                val nsCursorClass = lib.objc_getClass("NSCursor")
-                if (nsCursorClass == null || Pointer.nativeValue(nsCursorClass) == 0L) return
+                val nsCursorClass = Foundation.getObjcClass("NSCursor") ?: return
 
-                val selector = lib.sel_registerName("setHiddenUntilMouseMoves:")
+                val selector = Foundation.createSelector("setHiddenUntilMouseMoves:")
                 if (selector == null || Pointer.nativeValue(selector) == 0L) return
 
-                lib.objc_msgSend(nsCursorClass, selector, true)
+                Foundation.invoke(nsCursorClass, selector, true)
```
nebojsa-vuksic: `Foundation` is a low level API tightly coupled with a MacOS platform. By it's name, this controller should be able to handle `hideCursor` on any platform, right? Let's move the Foundation API usage to MacOS util or some other delegate class. 

We should also add tests for StandalonePlatformCursorController.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/StandaloneScrollbarHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+                createDelegate(
+                    "JewelScrollbarTrackClickBehaviorObserver",
+                    Foundation.createSelector("handleBehaviorChanged:")!!,
+                    this,
+                ),
+                Foundation.createSelector("handleBehaviorChanged:"),
+                Foundation.nsString("AppleNoRedisplayAppearancePreferenceChanged"),
+                ID.NIL,
+                2, // NSNotificationSuspensionBehaviorCoalesce
+            )
```
nebojsa-vuksic: This part of code has a feature envy code smell. Too many invocations of a `Foundation` low level API. 
Maybe we can move it to `MacUtil` or some other delegate component?

p.s. We are missing bunch of tests here :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/StandaloneScrollbarHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+        val delegateClass = Foundation.allocateObjcClassPair(Foundation.getObjcClass("NSObject")!!, name)
+        if (ID.NIL != delegateClass && delegateClass != null) {
+            if (!Foundation.addMethod(delegateClass, pointer, callback, "v@")) {
+                @Suppress("detekt:TooGenericExceptionThrown") // Copied from IJP
+                throw RuntimeException("Cannot add observer method")
+            }
+            Foundation.registerObjcClassPair(delegateClass)
+        }
+        return Foundation.invoke(name, "new")
+    }
```
nebojsa-vuksic: Should this be moved to `Foundation`?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/StandaloneScrollbarHelper.kt
```diff
@@ -0,0 +1,179 @@
...
+        val pool = Foundation.NSAutoreleasePool()
+        try {
+            return producer()
+        } catch (e: Throwable) {
+            logger.warn(e)
+        } finally {
+            pool.drain()
+        }
+        return null
+    }
```
nebojsa-vuksic: Same as a comment above :)

nebojsa-vuksic (changes requested):
We are missing tests :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/StandaloneScrollbarHelper.kt
DanielSouzaBertoldi: Yeah, totally! I'll move it over to `MacUtil`, thanks!

DanielSouzaBertoldi:
@nebojsa-vuksic just pushed the changes you asked + added tests. However, to avoid having to add the JNA dependency in the test module, I had to make a bunch of changes to the code which in turn made it a bit more complex:

1. Created an interface that `MacUtils` implements, so that we can create a fake impl and use it in tests.
2. With the use of callbacks, `StandaloneScrollbarHelper` doesnt inherits from `JNA`

Like, I think this is still pretty okay overall, but let me know if you disagree (in other words, if you think it's a bit overkill)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/macos/MacPlatformServicesDefaultImpl.kt
DanielSouzaBertoldi, line 43: To avoid making `StandaloneScrollbarHelper` inherit from `JNA` and subsequently forcing us to add the JNA dependency in our `int-ui-standalone-tests` module, I had to move references to JNA's `Callback` here.

With the use of callbacks, we can make `StandaloneScrollbarHelper` refresh its state with whatever the callback returns. BUT! We have to do this "hacky" thing to keep the reference from being purged by GC 😭

DanielSouzaBertoldi, line 25: Please feel free to suggest better names if you think of any! 🙏🏻

nebojsa-vuksic (approved):

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/theme/IntUiTheme.kt
rock3r: Should we remember the instance? Otherwise we build a new one every time now that it is a class

DanielSouzaBertoldi: Yep, silly mistake. Thanks for pointing this out!

platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.MacOS.kt
```diff
@@ -12,8 +12,8 @@ import androidx.compose.ui.platform.debugInspectorInfo
 import androidx.compose.ui.unit.dp
 import com.jetbrains.JBR
 import org.jetbrains.jewel.foundation.theme.JewelTheme
+import org.jetbrains.jewel.intui.standalone.window.macos.MacPlatformServicesDefaultImpl
```
rock3r: Nice direction overall. Could we make consumers depend on `MacPlatformServices` rather than `MacPlatformServicesDefaultImpl` directly? If this is meant to be the standalone-side analogue of IJPL platform services, it would be cleaner if the concrete default impl stayed hidden and call sites used the interface/capability boundary instead. That keeps coupling lower and makes substitution/testing easier.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/macos/MacPlatformServices.kt
```diff
@@ -0,0 +1,246 @@
...
+                 *
+                 * @param self The Objective-C receiver object
+                 * @param selector The selector that was invoked
+                 * @param event The notification event
+                 */
+                @Suppress("unused")
+                fun callback(self: ID?, selector: Pointer?, event: ID?) {
+                    action()
+                }
+            }.also { nativeCallbackReference = it }
```
rock3r: `nativeCallbackReference` only retains the most recently registered callback. If `onPreferencesChanged()` is called more than once, earlier callbacks lose their strong reference while still being registered natively. Is the lifecycle here guaranteed singleton-only? If not, we probably want to retain all active callbacks (or return a disposable/registration handle).

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiTabStripScrollbarStyling.kt
```diff
@@ -30,7 +31,7 @@ public fun ScrollbarStyle.Companion.tabStripDark(): ScrollbarStyle =
 public fun ScrollbarStyle.Companion.tabStripMacOsLight(
     colors: ScrollbarColors = ScrollbarColors.macOsLight(),
     metrics: ScrollbarMetrics = ScrollbarMetrics.tabStripMacOs(),
-    trackClickBehavior: TrackClickBehavior = TrackClickBehavior.NextPage,
+    trackClickBehavior: TrackClickBehavior = ScrollbarHelper.trackClickBehavior,
```
rock3r, line 47: Question on parity: standalone tab-strip defaults now read the macOS system track-click behavior here, but the bridge tab-strip path still hardcodes `TrackClickBehavior.JumpToSpot` in `IntUiBridgeScrollbar.readTabStripScrollbarStyle()`. Is this divergence intentional? If yes, it'd be good to document it or track the follow-up.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiTabStripScrollbarStyling.kt
```diff
@@ -30,7 +31,7 @@ public fun ScrollbarStyle.Companion.tabStripDark(): ScrollbarStyle =
 public fun ScrollbarStyle.Companion.tabStripMacOsLight(
```
rock3r, line 44: Could we add KDoc for the macOS tab-strip helpers too? Their default behavior changed in this PR, so it would help to call out that the track-click behavior now follows system preferences by default.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/macos/MacPlatformServicesDefaultImpl.kt
```diff
@@ -0,0 +1,246 @@
...
+
+    public fun readScrollbarVisibility(): ScrollbarVisibility
+
+    public fun onPreferencesChanged(action: () -> Unit)
+}
+
+@ApiStatus.Internal
+@InternalJewelApi
+public object MacPlatformServicesDefaultImpl : MacPlatformServices {
+    private val logger = Logger.getLogger(MacPlatformServicesDefaultImpl::class.java.simpleName)
```
rock3r: Shouldn't we use JewelLogger?

platform/jewel/samples/showcase/src/main/resources/icons/terminal.svg
rock3r, line 1: Aren't these in AllIconsKeys?

rock3r (changes requested):

platform/jewel/samples/showcase/src/main/resources/icons/terminal.svg
DanielSouzaBertoldi: Nope, `AllIcons.java` which is the class we fetch the icons from to put in our `AllIconsKeys` doesn't have a reference to it.

From what I've gathered, it only maps icons in `platform/icons`

This terminal icon I ~blatantly stole~ copied from `plugins/terminal` 😈

platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/TitleBar.MacOS.kt
DanielSouzaBertoldi: Agree!

Just created a

```kt
public val LocalMacPlatformServices: ProvidableCompositionLocal<MacPlatformServices> = staticCompositionLocalOf {
    error("No LocalMacPlatformServicesProvider provided. Have you forgotten the theme?")
}
```

And provided it in `IntUiTheme` like `LocalMacPlatformServices provides MacPlatformServicesDefaultImpl,`

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/macos/MacPlatformServicesDefaultImpl.kt
DanielSouzaBertoldi: Yep! I forgot we had our own logger 😭

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/window/macos/MacPlatformServices.kt
DanielSouzaBertoldi: It is a guaranteed singleton (currently!), because the only place we call it is in `StandaloneScrollbarHelper`, which calls `onPreferenceChanged` only once in its `init` block. The helper itself is initialized lazily only once as well:

`private val scrollbarService by lazy { StandaloneScrollbarHelper() }`

In other words, `nativeCallbackReference` is only ever written once and never overwritten, the `var` is just a GC anchor. Without this hacky bit the `callback` object would be collected immediately after `init` returns since there are no other strong references to it :(

This was the approach I opted to go for because otherwise we'd have to add a dependency to `JNA` and the code would be a real pain in the ass to test. By passing `Callback` as... well, a callback to `Foundation`, anytime the OS sends a new update via the callback we can automatically change the flows that control the scrollbar behavior/visibility

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiTabStripScrollbarStyling.kt
DanielSouzaBertoldi: Not intentional. Fixed!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TabStrip.kt
```diff
@@ -148,30 +148,7 @@ public fun TabStrip(
...
-            exit =
-                fadeOut(
-                    tween(
-                        durationMillis = scrollbarVisibility.thumbColorAnimationDuration.inWholeMilliseconds.toInt(),
-                        delayMillis = scrollbarVisibility.lingerDuration.inWholeMilliseconds.toInt(),
-                        easing = LinearEasing,
-                    )
-                ),
-            modifier = Modifier.semantics { hideFromAccessibility() },
-        ) {
```
DanielSouzaBertoldi, line 172: Had to remove this because it was overriding the logic from `BaseScrollbar` which already handles hiding/showing the scrollbar + some other neat stuff.

https://github.com/user-attachments/assets/719c3c50-d6bd-4a7c-b462-d46b60018697

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiTabStripScrollbarStyling.kt
```diff
@@ -30,34 +33,42 @@ public fun ScrollbarStyle.Companion.tabStripDark(): ScrollbarStyle =
...
     ScrollbarStyle(
         colors = colors,
         metrics = metrics,
         trackClickBehavior = trackClickBehavior,
-        scrollbarVisibility = scrollbarVisibility,
+        scrollbarVisibility =
+            when (scrollbarVisibility) {
+                is WhenScrolling -> WhenScrolling.tabStrip()
+                is AlwaysVisible -> AlwaysVisible.tabStrip()
+            },
```
DanielSouzaBertoldi, line 86: I had to add this `when` here because our `MacScrollbarHelperImpl` only returns either a `ScrollbarVisibility.WhenScrolling.macOs()` or `ScrollbarVisibility.AlwaysVisible.macOs()`. Check it:

Image

The tab strip scrollbar is a bit thinner, so instead of adding a new variable to our helper impl just for tab strip OR passing down a string, I opted to just map it to the correct style.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiTabStripScrollbarStyling.kt
```diff
@@ -98,13 +109,14 @@ public fun ScrollbarMetrics.Companion.tabStripWindowsAndLinux(
             "trackPaddingWithBorder = trackPaddingWithBorder" +
             ")"
     ),
+    level = DeprecationLevel.HIDDEN,
```
DanielSouzaBertoldi, line 140: Without this, calling `.tabStrip()` in this file would reference this deprecated function instead of the newly created one in `IntUiScrollbarStyling`

rock3r (changes requested):
https://github.com/JetBrains/intellij-community/pull/3407/changes#r2917935735 seems still not addressed despite being resolved

DanielSouzaBertoldi:
@rock3r thanks for pointing the missing KDoc, it totally slipped past me.

Also, updated API dumps to 0.39 😎 let'sss goooooo

rock3r (approved):

rock3r:
ready to merge

PR #3409, opened 2026-02-05
[JEWEL-1029] [JEWEL-1230] Reimplement `Dropdown` component

wellingtoncosta:
## Summary

This PR introduces a significant refactoring of `PopupMenu` to use `PopupContainer` instead of directly using `Popup`, consolidating visual styling (shadows, borders, backgrounds) and scrolling logic in one place. Previously, `MenuContent` had its own implementation of these features; now `PopupContainer` handles all popup styling consistently across the application, with `MenuContent` focusing solely on rendering menu items. Both menu content and ad content are now in the same scrollable area within `PopupContainer`.

To support this refactoring, `PopupContainer` was enhanced with scrolling support via `maxHeight`, `useIntrinsicWidth` for width control, and key event handling. The `PopupMenu` API now accepts both `menuStyle` and `popupContainerStyle` parameters to better separate menu item styling from container styling.

Additionally, this PR introduces `MenuComboBox`, a modern replacement for the deprecated `Dropdown` component that combines `ComboBox` UI with `MenuContent` for richer popup functionality including icons, keybindings, separators, and submenus. The component provides better keyboard navigation, focus management, and accessibility support.

Showcase samples were updated to demonstrate `MenuComboBox` usage.

## Screen recording

https://github.com/user-attachments/assets/46678c13-f20a-4b9a-9f07-94e795ea74df

## Release notes

### ⚠️ Important Changes                                                                                                                                                                                          
                                                                                                                                                                                                                   
* **JEWEL-1230** `Dropdown` component is now deprecated in favor of `MenuComboBox`        
                                                                                                                                                                                                                   
### New features                                                                                                                                                                                                 
                                                                                                                                                                                                                   
* **JEWEL-1029** Added scrolling support to `PopupContainer` with optional `maxHeight` parameter and vertical scrollbar
* **JEWEL-1029** Added `useIntrinsicWidth` parameter to `PopupContainer` for controlling width measurement (set to `false` for `SubcomposeLayout`-based components like `LazyColumn`)                                                
* **JEWEL-1029** Added key event handling to `PopupContainer` via `onPreviewKeyEvent` and `onKeyEvent` parameters
* **JEWEL-1029** Added `PopupMenu` overloads accepting both `menuStyle` and `popupContainerStyle` for better separation of menu item styling from container styling
* **JEWEL-1230** Introduced experimental `MenuComboBox` as modern `Dropdown` replacement with support for menu items with icons, keybindings, separators, and submenus

### Deprecated API

* **JEWEL-1230** Deprecated `Dropdown` in favor of `MenuComboBox` with `DeprecationLevel.WARNING`
* **JEWEL-1029** Deprecated older `PopupContainer`, `PopupMenu`, and `ComboBox` overloads in favor of new variants with enhanced parameters

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to new/deprecated public UI APIs and changes to popup sizing/scrolling/focus behavior that could affect menus and combo boxes across the library.
> 
> **Overview**
> Introduces experimental `MenuComboBox` (menu-backed dropdown) built on `ComboBox` + `MenuContent`, and deprecates the older `Dropdown` API with an automated `ReplaceWith`.
> 
> Refactors `PopupMenu` to render via `PopupContainer` (instead of raw `Popup`), splitting `menuStyle` vs `popupContainerStyle` and adding `maxHeight`-based scrolling support; `MenuContent` is simplified to item rendering only.
> 
> Extends `PopupContainer` with `useIntrinsicWidth`, `maxHeight`, and key-event hooks, and extends experimental `ComboBox` to accept `PopupProperties`, popup key handling, and intrinsic-width control; updates `ListComboBox` popup focusability handling and fixes stale `onPopupVisibleChange` captures.
> 
> Updates the showcase to demonstrate `MenuComboBox` (icons, separators, submenus, disabled state) and refreshes API dumps accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d2c6822f6371fa49d3e7e2aefa77c14a31f471e1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
DanielSouzaBertoldi: I think it's best to make this distinction as early as possible in the KDoc, since the idea is to strongly suggest users to use this variant from now on, right?


```suggestion
 * This version of ComboBox allows for complete customization of the label area through a composable function and also includes additional parameters for controlling popup behavior, keyboard events, and width measurement.
```

And also add the reason why this variant is the one users should stick with 90% of the time

DanielSouzaBertoldi:
```suggestion
 * @param useIntrinsicPopupWidth Whether to use intrinsic width measurement for the popup. Set it to `true` when the popup
 *   content must size itself to its widest item (e.g., `MenuContent`). Set to false (default) when the content contains
 *   `SubcomposeLayout`-based components (`LazyColumn`, `LazyRow`, etc.) which don't support intrinsic measurements
```

DanielSouzaBertoldi, line 301: In the KDoc it's stated that this parameter has a default value of `false`, but that's not the case in the actual code.

DanielSouzaBertoldi, line 306: Once your PR is merged, I'll have to make sure to add a variant that does not let the user set the `maxPopupHeight` but instead should be high enough for an arbitrary number of rows to appear at once in my PR (#3384)

DanielSouzaBertoldi: Why is the default for `focusable` false? Thinking as a user, I'd probably like to have the popup automatically gain focus once I click something that opens it.

Also, this property is set as `true` by default in the `PopupMenu` variant that you added

Or is there some context here I'm missing? 🤯

DanielSouzaBertoldi:
```suggestion
    onPopupKeyEvent: ((KeyEvent) -> Boolean)? = null,
```

Import `androidx.compose.ui.input.key.KeyEvent` 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: The popup will only become scrollable if the popup height is smaller than what the content needs to be totally visible, maybe we can make this a little bit clearer


```suggestion
 * @param maxHeight The maximum height of the menu, beyond which the content becomes scrollable.
```

DanielSouzaBertoldi: Same point as before

DanielSouzaBertoldi: I think we can get rid of these comments

DanielSouzaBertoldi, line 1197: nit: for the sake of consistency with your prior changes, rename this parameter to `menuStyle`

DanielSouzaBertoldi: This code is duplicated both in here and on line 411. We could create a function that maps a `PopupContainerStyle` with values from `MenuStyle`:

nit: delete the comment above, the code is self-explanatory 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
DanielSouzaBertoldi, line 48: nit: for consistency, if you opt to make the change I suggested earlier about making it clear when the popup becomes scrollable, make sure to add the info here too 🙏🏻

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
```diff
@@ -0,0 +1,124 @@
...
+    outline: Outline = Outline.None,
+    maxPopupHeight: Dp = Dp.Unspecified,
+    maxPopupWidth: Dp = Dp.Unspecified,
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    style: ComboBoxStyle = JewelTheme.comboBoxStyle,
+    menuStyle: MenuStyle = JewelTheme.menuStyle,
+    onPopupVisibleChange: (visible: Boolean) -> Unit = {},
+    content: MenuScope.() -> Unit,
+) {
+    val popupManager = remember { PopupManager(onPopupVisibleChange) }
```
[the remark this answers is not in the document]
DanielSouzaBertoldi: Or use `rememberUpdatedState` if the manager is "heavy" or holds its own internal state that you don't want to lose 😄 this ensures the manager will always call the latest version of the function without needing to be re-instantiated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
DanielSouzaBertoldi:
```suggestion
* When a `maxHeight` is specified and the content exceeds it, the container becomes scrollable
* and displays a vertical scrollbar.
```

DanielSouzaBertoldi: nit

```suggestion
 * @param useIntrinsicWidth Whether to use `IntrinsicSize.Max` for width measurement. Set to `true` when you want the popup
 *   to size to its widest content. Set to false when content contains `SubcomposeLayout`-based components (`LazyColumn`,
 *   `LazyRow`, etc.) which don't support intrinsic measurements
```

DanielSouzaBertoldi: Import `androidx.compose.ui.input.key.KeyEvent` 😄 


```suggestion
    onPreviewKeyEvent: ((KeyEvent) -> Boolean)? = null,
    onKeyEvent: ((KeyEvent) -> Boolean)? = null,
```

DanielSouzaBertoldi, line 220: nit: add KDoc here too since the previous `PopupContainer` now has one 😛 OOORRRRR since both of them are now hidden, it doesn't really matter. In this case, the best course of action is to just remove the KDoc from the previous `PopupContainer`

DanielSouzaBertoldi: Import `androidx.compose.ui.input.key.KeyEvent` 😄 


```suggestion
    onPreviewKeyEvent: ((KeyEvent) -> Boolean)? = null,
    onKeyEvent: ((KeyEvent) -> Boolean)? = null,
```

DanielSouzaBertoldi: What do you mean by "branches" here?

DanielSouzaBertoldi: Sorry, I don't get it. If users don't set a `maxHeight` then the content will scroll independently from the `adContent`?

Or by "original behavior" you mean the fact that maybe there's no vertical scroll at all?

DanielSouzaBertoldi (changes requested):
Awesome stuff! 🚀 

Just some minor changes + a little nit-picking 😈

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
wellingtoncosta: The `if` / `else` branches

wellingtoncosta: I reworked this to be more intuitive rephrased these comments. Also found a subtle bug in some combo box samples regarding measurement. Things are now more clear and working.

wellingtoncosta:
@DanielSouzaBertoldi I resolved all your comments (I hope so), so please take a look on this PR again once you have a chance. Let me know if you have more questions or concerns.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
DanielSouzaBertoldi: Not sure if this was changed. Every other code here sets `focusable` as true, but this is still false. Is this expected?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: I mean, this is just my opinion. If you don't agree please let me know :)

DanielSouzaBertoldi: Still duplicated

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
DanielSouzaBertoldi: Now its missing the part about the content being scrollable if it's bigger than the maximum size. I'll let you decide if this bit of info is important or not

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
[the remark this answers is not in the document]
DanielSouzaBertoldi: Wait, this wasn't fixed yet.

DanielSouzaBertoldi:
@wellingtoncosta I noticed some changes are still pending. Let me know once you reply to my comments/add the changes, please

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
DanielSouzaBertoldi: This code is still not keying on `onPopupVisibleChange` or updated to use `rememberUpdatedState` instead.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
DanielSouzaBertoldi: It's always a good idea to add tests to check if the bugs you found have been fixed for good. It also protect us in the future if the bug happens to come again after some changes 😎

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -356,41 +526,52 @@ public fun MenuContent(
...
-                        selectedSubMenu = selectedSubMenu,
-                        setSelectedSubMenu = { selectedSubMenu = it },
-                    )
-                }
-            }
-
-            adContent?.let { PopupAd(modifier = Modifier.fillMaxWidth()) { it() } }
+            MenuStyle(isDark = style.isDark, colors = style.colors, metrics = adjustedMetrics, icons = style.icons)
+        } else {
+            style
```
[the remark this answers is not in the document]
wellingtoncosta: I actually removed this useless code. I was trying to fix a sizing issue with that code, but now it's useless.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
wellingtoncosta: Now it's fixed haha

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
wellingtoncosta: I think it's fine since we have a mention about that in the KDoc.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
wellingtoncosta: Now it's fixed 😄

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
wellingtoncosta: Fixed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
```diff
@@ -58,17 +58,28 @@ import org.jetbrains.jewel.ui.theme.dropdownStyle
...
  * @see EditableListComboBox
  */
+@Deprecated(
+    message =
+        "Use MenuComboBox instead for a more robust implementation with better keyboard navigation and accessibility",
+    replaceWith =
+        ReplaceWith(
+            "MenuComboBox(labelContent = content, modifier = modifier, enabled = enabled, outline = outline, items = menuContent)",
+            "org.jetbrains.jewel.ui.component.MenuComboBox",
+        ),
```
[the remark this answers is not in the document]
wellingtoncosta: Fixed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
wellingtoncosta: Fixed

wellingtoncosta: Fixed

wellingtoncosta: Removed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
```diff
@@ -112,13 +229,99 @@ public fun PopupContainer(
...
+                    .background(colors.background, popupShape)
+                    .clip(popupShape)
+                    .thenIf(useIntrinsicWidth) { width(IntrinsicSize.Max) }
+                    .thenIf(maxHeight != Dp.Unspecified) { heightIn(max = maxHeight) }
+
+            val movableContent = remember { movableContentOf { content() } }
+            val movableAdContent =
+                adContent?.let {
+                    remember { movableContentOf { PopupAd(modifier = Modifier.fillMaxWidth()) { adContent() } } }
+                }
```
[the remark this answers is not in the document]
wellingtoncosta: Fixed

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
wellingtoncosta: It has no default value now and the KDoc doesn't mention any default value.

DanielSouzaBertoldi (approved):
Awesome work! 🥳

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -356,41 +480,18 @@ public fun MenuContent(
...
-                    shadowSize = style.metrics.shadowSize,
-                    shadowColor = colors.shadow,
-                    borderWidth = style.metrics.borderWidth,
-                    borderColor = colors.border,
-                )
-                .background(colors.background, menuShape)
-                .width(IntrinsicSize.Max)
-                .onHover { localMenuController.onHoveredChange(it) }
+    Column(
+        modifier = modifier.padding(style.metrics.contentPadding).onHover { localMenuController.onHoveredChange(it) }
```
rock3r, line 509: Wouldn't this leave `ContextMenu` unstyled now that `MenuContent` no longer owns the popup container chrome/scrolling? `ContextMenu.kt` still renders `MenuContent` directly inside a raw `Popup`, so I'd expect context menus to lose background/border/shadow/scroll behavior unless `ContextMenu` is migrated too.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
```diff
@@ -58,17 +58,35 @@ import org.jetbrains.jewel.ui.theme.dropdownStyle
...
  * @see ListComboBox
  * @see EditableComboBox
  * @see EditableListComboBox
  */
+@Deprecated(
+    message =
+        "Use MenuComboBox instead for a more robust implementation with better keyboard navigation and accessibility",
+    replaceWith =
+        ReplaceWith(
+            """MenuComboBox(
```
rock3r, line 78: The release notes currently sell this as a smoother auto-migration than it really is. `Dropdown.content` is `BoxScope.() -> Unit`, while `MenuComboBox.labelContent` takes `() -> Unit`, so forwarded/stored lambdas will still need manual adaptation (for example `labelContent = { Box(content = content) }`).

That minor pain is acceptable for an experimental API, but I'd still soften the automatic-replacement claim in the release notes.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
```diff
@@ -112,13 +231,103 @@ public fun PopupContainer(
...
+            val currentAdContent by rememberUpdatedState(adContent)
+
+            val movableContent = remember { movableContentOf { currentContent() } }
+            val movableAdContent = remember {
+                movableContentOf {
+                    currentAdContent?.let { ad -> PopupAd(modifier = Modifier.fillMaxWidth()) { ad() } }
+                }
+            }
+
+            if (useIntrinsicWidth) {
```
rock3r, line 312: Wouldn't `maxHeight` now only become actually scrollable when `useIntrinsicWidth` is `true`? The KDoc says specifying `maxHeight` makes the popup scrollable, but in the default `useIntrinsicWidth = false` branch we only apply `heightIn(max = maxHeight)` and skip the `verticalScroll` / `VerticalScrollbar` wrapper. That looks like a regression for direct `PopupContainer(maxHeight = ...)` callers with non-lazy content.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -79,21 +70,81 @@ import org.jetbrains.jewel.foundation.state.SelectableComponentState
...
+    menuStyle: MenuStyle = JewelTheme.menuStyle,
+    popupContainerStyle: PopupContainerStyle = JewelTheme.popupContainerStyle,
+    popupProperties: PopupProperties = PopupProperties(focusable = true),
+    adContent: (@Composable () -> Unit)? = null,
+    content: MenuScope.() -> Unit,
+) {
+    val density = LocalDensity.current
+
+    val popupPositionProvider =
+        remember(menuStyle.metrics.offset, menuStyle.metrics.menuMargin, horizontalAlignment, density) {
```
rock3r: `popupContainerStyle` is now a first-class parameter, but this overload still derives popup positioning from `menuStyle.metrics.{offset,menuMargin}`. I feel a bit confused as to what I should expect here, really. But it's weird we have both and we read from the old one — I suppose I would expect that we take from the new style where possible, and deprecate the fields in the older one and stop using them.

Also, `Submenu` is still using the default/theme container style rather than propagating a caller-supplied `popupContainerStyle`. Looks like a bug, as style overrides won't propagate from root menus to their submenus.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -356,41 +480,18 @@ public fun MenuContent(
...
-                modifier = Modifier.fillMaxHeight().align(Alignment.CenterEnd),
+        var selectedSubMenu by remember { mutableStateOf<SubmenuItem?>(null) }
+        items.forEach { item ->
+            MenuItem(
+                item = item,
+                style = style,
+                showIcons = anyItemHasIcon,
+                showKeybindings = anyItemHasKeybinding,
+                selectedSubMenu = selectedSubMenu,
+                setSelectedSubMenu = { selectedSubMenu = it },
```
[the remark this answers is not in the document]
rock3r: Oh just saw Cursor also commented the same lol

rock3r (changes requested):

platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/ComboBoxes.kt
```diff
@@ -318,6 +326,135 @@ private fun CustomComboBoxes() {
...
+                        style = JewelTheme.defaultTextStyle,
+                        modifier = Modifier.padding(comboBoxStyle.metrics.contentPadding),
+                    )
+                },
+                modifier = Modifier.widthIn(max = 200.dp).fillMaxWidth(),
+                content = {
+                    stringItems.forEachIndexed { index, item ->
+                        selectableItem(selected = false, onClick = { selectedItem = index }) { Text(item) }
+                    }
+                },
```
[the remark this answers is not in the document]
DanielSouzaBertoldi: Fixed + added test

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: ### Popup Container Style vs Menu Style

I'd say we better use `PopupContainerStyle` as the source of truth for popups. How far a popup should appear from whatever is creating it is a popup concern, not a menu one (now that we are distinguishing both). For submenus, we should keep reading from `menuStyle.offset` because is specifically about how it can position itself relative to its parent menu item, so, it's a menu-level concern.

Given this, I noticed that `readPopupContainerStyle` in `IntUiBridgePopupContainer` has a `menuMargin` different than `IntUiBridgeMenu`:

Image

So this was also fixed

The thing though is that.... deprecating just a property (`menuMargin` in this case) is a pain...

### Style fixed ✅ 

| Before | After |
|---|---|
| Image | Image |

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: Fixed

| Before | After |
|---|---|
| Image | Image |

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
DanielSouzaBertoldi: Fixed!

| Before | After |
|---|---|
| image | image |

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/MenuComboBox.kt
```diff
@@ -0,0 +1,119 @@
...
+    val currentOnPopupVisibleChange by rememberUpdatedState(onPopupVisibleChange)
+    val popupManager = remember { PopupManager({ visible -> currentOnPopupVisibleChange(visible) }) }
+    val menuController = remember {
+        DefaultMenuController(
+            onDismissRequest = {
+                popupManager.setPopupVisible(false)
+                true
+            }
+        )
+    }
```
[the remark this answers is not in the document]
DanielSouzaBertoldi: If popupManager was keyed on another variable then the stale callback thing would have a chance to happen, I'll give you that.

However, keying menuController to popupManager would basically be a no-op. `popupManager` is remembered without keys so it's stable for the entire lifetime of the composable.

rock3r:
bugbot run

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/styling/MenuStyling.kt
```diff
@@ -265,22 +265,42 @@ public class MenuItemColors(
...
     public val contentPadding: PaddingValues,
     public val offset: DpOffset,
     public val shadowSize: Dp,
     public val borderWidth: Dp,
     public val itemMetrics: MenuItemMetrics,
     public val submenuMetrics: SubmenuMetrics,
 ) {
+    // To avoid source/binary breaking change for clients that read this property
+    @Deprecated(message = "Use PopupContainerStyle.metrics.menuMargin instead.", level = DeprecationLevel.WARNING)
+    public val menuMargin: PaddingValues = PaddingValues()
```
rock3r: This preserves the old `menuMargin` symbol, which is great for compat, but not the old value, which may be a compat issue. The hidden constructor still accepts `menuMargin`, and existing compiled callers may also read `menuStyle.metrics.menuMargin`, but now always see `PaddingValues()` instead of the value they provided. This can lead to unexpected behaviours.

Can we keep storing the old value for compatibility, even if new code should use `PopupContainerStyle.metrics.menuMargin` and this value is effectively never used? In any case, I reckon the deprecation message/KDoc should make that explicit.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
```diff
@@ -112,13 +231,105 @@ public fun PopupContainer(
...
+
+            val movableContent = remember { movableContentOf { currentContent() } }
+            val movableAdContent = remember {
+                movableContentOf {
+                    currentAdContent?.let { ad -> PopupAd(modifier = Modifier.fillMaxWidth()) { ad() } }
+                }
+            }
+
+            val needsScrolling = maxHeight != Dp.Unspecified
+            val scrollState = if (needsScrolling || useIntrinsicWidth) rememberScrollState() else null
```
rock3r: Wouldn't this make intrinsic-width popups scrollable even when no `maxHeight` was requested? `PopupMenu`, `ContextMenu`, `MenuComboBox`, and submenus all use intrinsic width, so they all get `verticalScroll`/scrollbar wiring even with `maxHeight = Dp.Unspecified`.

I think intrinsic width and scrolling should stay separate: use the intrinsic-width layout branch if needed, but only create/apply the scroll state when `needsScrolling` is true.

> [!NOTE]
> This is an AI finding, I did not manage to confirm it, but seems legit?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ComboBox.kt
```diff
@@ -401,9 +484,11 @@ internal fun ComboBoxImpl(
                             .then(popupModifier)
                             .onClick { popupManager.setPopupVisible(false) },
                     horizontalAlignment = horizontalPopupAlignment,
-                    popupProperties = PopupProperties(focusable = false),
+                    useIntrinsicWidth = useIntrinsicPopupWidth,
```
rock3r, line 485: Shouldn't we also pass the resolved `maxHeight` to `PopupContainer(maxHeight = ...)` here? Right now `ComboBoxImpl` caps the outer modifier with `heightIn(max = maxHeight)`, but the new scrolling contract lives in `PopupContainer.maxHeight`.

This is currently masked a bit by intrinsic width also creating a scroll state, but if that is fixed, `MenuComboBox(maxPopupHeight = ...)` with non-lazy `MenuContent` would be height-capped without the container knowing it should make the content scrollable.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Button.kt
rock3r, line 976: Not sure if this is indeed how it works, but it seems to me this may be missing the max height; this call caps the popup height via `splitButtonPopupModifier.heightIn(max = maxPopupHeight)`, but the menu branch doesn't pass `maxHeight` to `PopupMenu`, so `PopupContainer` doesn't own scrolling for overflowing menu content.

Can we pass `maxHeight = maxPopupHeight` here?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
rock3r: Follow-up note: @wellingtoncosta would you be able to amend the PR description? Nobody else can do it 😅

rock3r (changes requested):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
DanielSouzaBertoldi: Yes, I do think this is a valid point. `useIntrinsicWidth` should only expand the container to fit its content 🤔 if we were applying an HorizontallyScrollbar then that'd be a different story, but that's not the case here. Scroll should only be applied when the height of the popup is not enough to display everything it has to

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Button.kt
DanielSouzaBertoldi: Yes sir 🫡 this is basically the same fix as the one before this comment. Thanks!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Dropdown.kt
wellingtoncosta: Removed the "which provides better keyboard navigation and accessibility support with automatic replacement provided" part.

rock3r (approved):
Thanks, all my concerns have been addressed.

DanielSouzaBertoldi:
[Sebastiano Poggi (@rock3r)](https://github.com/rock3r) [Wellington Costa (@wellingtoncosta)](https://github.com/wellingtoncosta) [Nebojsa Vuksic (@nebojsa-vuksic)](https://github.com/nebojsa-vuksic) it's been too long since this was last reviewed. I resolved conflicts, but better to be safe than sorry. Can you guys take a look at this again pretty please? 🥺

wellingtoncosta:
@DanielSouzaBertoldi the code looks good to me (I know I'm biased to say this), but we have some checks failing — especially those related to the `api-dump` files.

DanielSouzaBertoldi:
@wellingtoncosta thanks! i mean, i'm sure you don't remember most of the code since it's been soooo long since you've opened it :P made sure to ask for your review because of that lol

yeeeahhh spectre tests are failing but Seb is already figuring it out
bazel one failed because of rate limit
the annotate breaking api changes with ijp dumps one is supposed to fail since we are changing the api surface level but it's not introducing any real breaking changes

rock3r:
Fix for Spectre is https://github.com/JetBrains/intellij-community/pull/3645
Not really a Spectre issue, just the nth case of IJPL changes breaking shit silently and us having tests that tell us IJPL broke something

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/PopupContainer.kt
rock3r: Could we clarify in the KDoc that `maxHeight` only enables scrolling when `useIntrinsicWidth = true`? Otherwise, the content needs to handle scrolling itself. Not blocking on this.

rock3r (approved):
Rechecked the changes since my previous approval. Looks good, just a non-blocking KDoc clarification on the existing thread.

DanielSouzaBertoldi:
Fixed the KDoc, will fix the conflict and if everything is a-ok then i'll ask the bot to merge it 😎

DanielSouzaBertoldi:
Ready to merge

PR #3415, opened 2026-02-10
[JEWEL-1076, IJPL-176416] New Icons API

jakub-senohrabek:
This PR adds new icons api that allows cross-api icons (compose, swing), and icons serialization. The implementation is separated from the API (and the rendering part from data part) to allow creation of Icons without having to depend on UI code. (useful when creating icons on backend)

Make sure to start with intellij.platform.icons.api module, which contains the core api. (also check the rendering submodule). Afterwards, check the impl module (start with intellij submodule) and jewel implementation. (changed jewel modules)

:warning: Do not merge this in usual way, this is already in ultimate branch :warning:

GitOrigin-RevId: 32b987b2a7bba316208aefd998a523f6a74ace30


develar:
We do not have configured codex bot here to review, right ;)?

platform/icons-api/legacy-icon-support/resources/intellij.platform.icons.api.legacyIconSupport.xml
```diff
@@ -0,0 +1,9 @@
+<!-- Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license. -->
+<idea-plugin visibility="internal">
+  <dependencies>
+    <!-- region Generated dependencies - run `Generate Product Layouts` to regenerate -->
+    <module name="intellij.libraries.kotlinx.serialization.core"/>
+    <module name="intellij.platform.icons.api"/>
```
jreznot: Do we expect that every plugin requiring it will need to add explicit dependency? I believe it must be exposed transitively from some platform modules

platform/icons-api/legacy-icon-support/src/org/jetbrains/icons/legacyIconSupport/SwingIcon.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.jetbrains.icons.IconManager
+import org.jetbrains.icons.design.IconDesigner
+import org.jetbrains.icons.icon
+import org.jetbrains.icons.modifiers.IconModifier
+import org.jetbrains.icons.modifiers.fillMaxSize
+
+/**
+ * Layer for embedding swing icons.
+ */
+@ExperimentalIconsApi
```
jreznot: Does it require constant opt-ins in every usage? May be plain `@Experimental` instead?

develar:
Why does this change have status DRAFT? Is it ready for review?

platform/icons-impl/src/org/jetbrains/icons/impl/InPlaceDeferredIconResolver.kt
```diff
@@ -0,0 +1,36 @@
...
+  override val deferredIcon: WeakReference<DefaultDeferredIcon>,
+  val evaluator: suspend () -> Icon
+): DeferredIconResolver {
+  var resolvedIcon: Icon? = null
+  private val deferredValue = CompletableDeferred<Icon>()
+  private val isPending = AtomicBoolean(false)
+
+  override suspend fun resolve(): Icon {
+    val resolved = resolvedIcon
+    if (resolved != null) return resolved
```
rock3r: I think this can break:

• `InPlaceDeferredIconResolver.resolve()` returns immediately when `resolvedIcon != null`
• But `DefaultDeferredIconRenderer` only flips from placeholder/null to the resolved renderer in `whenDone()`
• New renderers start with `isDone = false`, `renderer = icon.placeholder?.createRenderer(...)`

So if a renderer is created after the deferred icon was already resolved, it never gets `whenDone()`, and `resolve()` now short-circuits without re-notifying listeners. That means the new renderer can stay stuck on the placeholder/ null forever.

There’s a second broken path here too:
• `DeferredIconResolverService.getOrCreateDeferredIcon()` returns `resolver?.deferredIcon?.get() ?: DefaultDeferredIcon(identifier, placeholder)`
• If the `resolver` is still in the map but its weak ref was cleared, callers get a fresh `DefaultDeferredIcon` that is not connected to the existing resolver state

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Icon.kt
```diff
@@ -76,11 +88,51 @@ public fun Icon(
...
+ * @param loadingStrategy specifies if the function should block the thread while loading the icon or show placeholder
+ * or render blank instead.
  */
-@Suppress("ComposableParamOrder") // To fix in JEWEL-929
+@Composable
+public fun Icon(
+    icon: Icon,
+    contentDescription: String?,
+    modifier: Modifier = Modifier,
+    loadingStrategy: LoadingStrategy = LoadingStrategy.BlockThread
```
rock3r: This seems unused, is it intended?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/icon/RendererBasedIconPainter.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.icons.rendering.Dimensions
+import org.jetbrains.icons.rendering.IconRenderer
+import org.jetbrains.icons.rendering.ScalingContext
+import org.jetbrains.jewel.foundation.InternalJewelApi
+
+@InternalJewelApi
+@ApiStatus.Internal
+public class RendererBasedIconPainter(private val iconRenderer: IconRenderer, private val scaling: ScalingContext) : Painter() {
+    override val intrinsicSize: Size = iconRenderer.calculateExpectedDimensions(scaling).toComposeSize()
```
rock3r: This is only computed once, is it intended? I would imagine this could break with animated or deferred icons if the dimensions change (e.g., the placeholder has a different size from the final icon)

platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/SwingIcon.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.awt.Graphics
+import java.awt.Graphics2D
+import java.awt.image.BufferedImage
+
+@ApiStatus.Experimental
+class SwingIcon(
+  val icon: Icon
+): javax.swing.Icon {
+  private val renderer by lazy {
+    IconRendererManager.getInstance().createRenderer(icon, RenderingContext.Empty) // TODO listen for updates to redraw Icon
```
rock3r, line 20: Just noting — this TODO will need to be addressed, or it'll break rendering animated/deferred icons in Swing

platform/icons-impl/intellij/src/org/jetbrains/icons/impl/intellij/rendering/SwingIcon.kt
```diff
@@ -0,0 +1,77 @@
...
+    IconRendererManager.getInstance().createRenderer(icon, RenderingContext.Empty) // TODO listen for updates to redraw Icon
+  }
+  private val dimensions by lazy {
+    renderer.calculateExpectedDimensions(SwingScalingContext(1f))
+  }
+
+  override fun paintIcon(c: Component?, g: Graphics, x: Int, y: Int) {
+    val scaling = getScaling(g)
+    val boundsSize = if (c != null) {
+      Dimensions(c.width, c.height)
```
rock3r: That’s not how `javax.swing.Icon` usually behaves. A Swing icon should paint into its own icon bounds, not expand to the host component’s full size. In things like buttons/labels with text, this risks stretching the icon to the component size instead of respecting `getIconWidth()`/`getIconHeight()`.

image

Local repro from the playground: this is the same `row(spacing = 25.percent) { ... }` icon rendered through `toSwingIcon()` under two different host sizes.

What’s striking is that the icon’s *internal layout* changes based on the available host width: instead of behaving like a small icon composed of two children plus spacing, it gets horizontally redistributed/squeezed as the host gets wider. That strongly suggests the renderer is resolving layout against the surrounding component bounds rather than a stable icon canvas / `getIconWidth()` × `getIconHeight()`.

So this does not just look theoretically “non-Swing-like” — it produces visibly wrong results for bounds-relative/layout-based icons in practice.

platform/icons-api/src/org/jetbrains/icons/IconManager.kt
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Deferred icon allows apis to return an Icon that takes some time to compute;
+ * optional placeholder can be included to allow rendering it before the actual icon is ready.
+ *
+ * To cache such icons and synchronize them over the network, some identifier should be given.
+ * Implementations might try to prefix the identifier with the source pluginId/moduleId to avoid clashes.
+ *
+ * If the identifier is not passed, an automatic one is created, however, this will prevent the result
+ * from being cached, as a new one is generated per each deferredIcon() call.
```
rock3r: Due to how the deserialization chain is currently set up, it looks like in the IDE implementation, deserialized deferred icons are not just “not yet remote-capable” — they are completely broken. I know there are still a few TODOs floating around in this area, but worth noting this as a critical missing area so we don't forget about it.

platform/icons-api/src/org/jetbrains/icons/DeferredIcon.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import org.jetbrains.annotations.ApiStatus
+
+/**
+ * Deferred Icon takes time to resolve; therefore, it is postponed to be resolved later.
+ * Placeholder icon can be set to provide a visual representation while the actual icon is being resolved.
+ * Unlike the older API, to force evaluation or get the resolved icon, IconManager should be used.
+ *
+ * @see IconManager.deferredIcon
+ * @see IconManager.forceEvaluationnnnn
```
rock3r:
```suggestion
 * @see IconManager.forceEvaluation
```

platform/icons-api/src/com/intellij/platform/icons/design/Shape.kt
```diff
@@ -0,0 +1,27 @@
+// Copyright 2000-2026 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.icons.design
+
+import kotlinx.serialization.Serializable
+import org.jetbrains.annotations.ApiStatus
+
+@ApiStatus.Experimental
+@Serializable
+sealed interface Shape {
```
rock3r, line 6: What about rounded rect? Seems pretty common

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Icon.kt
```diff
@@ -76,11 +88,51 @@ public fun Icon(
...
  * @param modifier optional [Modifier] for this Icon.
- * @param iconClass The class to use for resolving the icon resource. Defaults to `key.iconClass`.
- * @param tint tint to be applied to the icon. If [Color.Unspecified] is provided, then no tint is applied.
- * @param hint [PainterHint] to be passed to the painter.
+ * @param loadingStrategy specifies if the function should block the thread while loading the icon or show placeholder
+ * or render blank instead.
  */
-@Suppress("ComposableParamOrder") // To fix in JEWEL-929
+@Composable
+public fun Icon(
```
rock3r, line 91: I assume the breaking changes will be undone and you'll add new overloads here?

platform/icons-impl/src/org/jetbrains/icons/impl/rendering/layers/IconIconLayerRenderer.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.jetbrains.icons.rendering.Bounds
+import org.jetbrains.icons.rendering.Dimensions
+import org.jetbrains.icons.rendering.PaintingApi
+import org.jetbrains.icons.rendering.RenderingContext
+import org.jetbrains.icons.rendering.ScalingContext
+import org.jetbrains.icons.rendering.createRenderer
+import org.jetbrains.icons.impl.layers.IconIconLayer
+import org.jetbrains.icons.impl.rendering.modifiers.applyTo
+
+class IconIconLayerRenderer(
```
rock3r: `NestedIconLayerRenderer`? And `NestedIconLayer` as a consequence

platform/icons-api/rendering/src/org/jetbrains/icons/rendering/IconRenderer.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.icons.Icon
+
+@ApiStatus.Experimental
+interface IconRenderer {
+  val icon: Icon
+  @ApiStatus.Internal
+  fun render(api: PaintingApi)
+  @ApiStatus.Internal
+  fun calculateExpectedDimensions(scaling: ScalingContext): Dimensions
```
rock3r: I think the intrinsic sizing contract is still under-specified for bounds-relative icons. For icons built from `fillMaxSize()`, percent-based layout, or relative badge sizing, it's not clear what `calculateExpectedDimensions()` is supposed to return as the canonical size. Compose/Jewel can render such icons under explicit host bounds, while Swing often relies on `iconWidth`/`iconHeight`, so the same icon can legitimately end up being rendered at different visual sizes in different frontends. That seems bad to me.

Can we define this contract more explicitly? E.g., should these icons:
- have a fixed default canvas/intrinsic size,
- be treated as having no meaningful intrinsic size unless the caller supplies bounds,
- or be normalized/restricted for frontends that require intrinsic measurement?

Without a clearer contract here, consistency across IntelliJ Swing/Jewel bridge/Jewel standalone (and remote frontends!) feels underspecified.

platform/icons-api/rendering/src/org/jetbrains/icons/rendering/PaintingApi.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.icons.design.Color
+import org.jetbrains.icons.filters.ColorFilter
+
+/**
+ * Abstraction of painting API, this is used to define icons or graphics that are customizable
+ * but also reusable between different environments, where graphic api differ.
+ */
+@ApiStatus.Internal
+interface PaintingApi {
```
rock3r: I’m a bit worried that `PaintingApi` is currently too minimal/incomplete to cover all reasonable use cases.

Right now it only exposes `drawImage`, `drawCircle`, `drawRect`, `withCustomContext`, and a very small `DrawMode` (`Fill`, `Clear`, `Stroke`). That seems barely enough for the currently implemented built-ins, but it already appears limiting for plausible icon use cases:
- no `drawLine` / path primitive
- no transform support (rotation/scale/translate), which makes even something like a diagonal strikeout hard to express
- no clipping/masking API
- no proper compositing model beyond the special-case `Clear`
- `Stroke` itself looks underspecified (e.g., there is no stroke width), and backend support is already inconsistent as is

On that last point:
- `ComposePaintingApi` still has `TODO Support strokes` and maps stroke drawing to fill
- `SwingPaintingApi` still has `TODO apply alpha & color filters` for image drawing
- `getUsedBounds()` also looks half-baked still

So my concern is less “this should necessarily become a huge immediate-mode graphics API” and more: is this API complete enough to be the foundational cross-runtime painter contract for Swing/Jewel bridge/Jewel standalone/remote-dev scenarios?

If the intent is to keep `PaintingApi` deliberately narrow, I think that should probably be paired with a clearer plan for adding higher-level declarative primitives above it (e.g. lines/slashes/rotation-capable layers). Otherwise we may quickly hit functionality gaps — all while the low-level substrate is also not fully specified.

This can be a follow-up, but it needs to be properly decided and tracked before we say "we'll come back to this later".

platform/icons-api/src/org/jetbrains/icons/modifiers/CutoutMarginModifier.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.jetbrains.icons.design.IconUnit
+
+@Serializable
+@ApiStatus.Experimental
+/**
+ * Add cutout margin to the specific layer, which will clear the surrounding area
+ * Currently supported only by shape and image layers, image layer will not consider
+ * internal svg shape and will cut out rectangular area.
+ */
+class CutoutMarginModifier(
```
rock3r: I’m not sold on `cutoutMargin` as a public abstraction. It seems to encode a very specific rendering trick (“clear a slightly larger version of this layer before painting it”) rather than a more general/icon-semantic concept.

A few things worry me here:
- its semantics are not really uniform across layer types (the KDoc already notes shape/image-specific behavior, and image cutout is only rectangular — if this were an interface, it would have a few impls with methods throwing `UnsupportedOperationException` — never a good sign)
- the effect is highly paint-order-dependent, since it only clears previously drawn content and can itself be covered by later layers. This can be fine but it's pretty unintuitive from the API shape
- it feels closer to an imperative compositing instruction than to a stable declarative style/modifier concept, which breaks the "purity" of the design
- its existence suggests a higher-level need (“make this overlay readable against the base icon”) that might deserve a more principled API than exposing this mechanism directly
- it's not really configurable (e.g., I can't set how wide that cutout margin actually is)

This works well for badges, but I’m not sure it scales as a clean public primitive for a portable, serializable icon model.

platform/icons-api/src/org/jetbrains/icons/IconManager.kt
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * @see org.jetbrains.icons.deferredIcon
+   */
+  fun deferredIcon(placeholder: Icon?, identifier: String? = null, classLoader: ClassLoader? = null, evaluator: suspend () -> Icon): Icon
+
+  suspend fun forceEvaluation(icon: DeferredIcon): Icon
+
+  /**
+   * Converts specific Icon to swing Icon.
+   * ! This is an expensive operation and can include image loading, reuse the instance if possible. !
```
rock3r: The KDoc here made me think this is a _dangerous_ API and we need to consider an async/caching story; this sounds a lot like a footgun otherwise...

- `toSwingIcon()` appears to create a fresh `SwingIcon` wrapper per call, and I don’t see a top-level cache from `org.jetbrains.icons.Icon` to `javax.swing.Icon`. The returned `SwingIcon` does have per-instance lazy caching (`renderer`, `dimensions`), so reusing the same instance helps, but that is a caller-side reuse pattern rather than a stronger conversion/cache contract.
- Likewise, I don’t see a real async loading/update path here — the Swing path looks fundamentally synchronous/lazy, and even has a `TODO` about listening for updates to redraw. So the current behavior seems closer to “expensive compatibility adapter with per-instance lazy work” than to “cached/asynchronously loaded Swing conversion”.
- Re: caching, if we're relying on the `CachedImageIcon` caching here, shouldn't it be at least documented?

Would it make sense to clarify the intended contract here?

platform/icons-api/src/org/jetbrains/icons/IconManager.kt
```diff
@@ -0,0 +1,88 @@
...
+   */
+  fun toSwingIcon(icon: Icon): javax.swing.Icon
```
rock3r: I’m a bit concerned about the portability story of the Swing escape hatch (`swingIcon(...)` / `javax.swing.Icon.toNewIcon()`), especially given the broader goals here around serializable / wire-friendly icon descriptions.

As implemented, this does not seem to semantically convert a legacy Swing icon into a native new-icon model. It wraps the old `javax.swing.Icon` as a custom layer and later renders it via IntelliJ-specific legacy-icon/image machinery. That is a perfectly reasonable migration bridge, but it also seems much less portable than a “real” new icon:
- Arbitrary `javax.swing.Icon` instances are opaque rendering objects, not stable declarative icon descriptions
- Support appears heavily dependent on IntelliJ-specific runtime/rendering infrastructure
- The serialization story for legacy-wrapped icons looks much weaker than for native new-icon layers — would it work for remote dev? And how? I'm not sure how well such icons can survive transport across process/machine boundaries

So I think it would help to clarify whether this escape hatch is intentionally only a local/bridge-mode migration tool, rather than something expected to participate fully in the portable/wire-friendly icon model. Right now the API shape makes it look more first-class than it may actually be in cross-runtime scenarios.

This also ties in with the comment on the KDoc above — if this is a full-fledge interop layer, it should not rely on Swing impl details for caching, for example.

platform/icons-impl/intellij/src/org/jetbrains/icons/impl/intellij/rendering/SwingPaintingApi.kt
```diff
@@ -0,0 +1,230 @@
...
+      srcX,
+      srcY,
+      srcX + (srcWidth ?: imageWidth),
+      srcY + (srcHeight ?: imageHeight),
+      null,
+    )
+  }
+}
+
+fun PaintingApi.swing(): SwingPaintingApi? = this as? SwingPaintingApi
```
rock3r: I’m not sure why `fun PaintingApi.swing(): SwingPaintingApi?` needs to exist at all at this stage.

The whole value of `PaintingApi` seems to be that it is the backend-neutral rendering contract shared across different runtimes/frontends. This helper feels like an official escape hatch back into one concrete backend implementation, which weakens that abstraction boundary.

If there is already a concrete use case that requires reaching through `PaintingApi` into `SwingPaintingApi`, it would be helpful to document that motivation here, and consider exposing it directly (and in Compose too!). Otherwise, this reads as “we might want to pierce the abstraction later”, which seems like exactly the kind of convenience API that encourages backend-specific rendering logic to creep into code that is supposed to stay portable.

Given how easy this helper would be to add later if/when a real need appears, I’d lean toward not exposing it until that need is proven.

platform/icons-api/src/org/jetbrains/icons/design/Color.kt
```diff
@@ -0,0 +1,81 @@
...
+  fun toHex(): String
+
+  companion object {
+    val Transparent: Color = RGBA(0f, 0f, 0f, 0f)
+  }
+}
+
+@ApiStatus.Experimental
+@Serializable
+class RGBA(
```
rock3r: Was it intentional to make `RGBA(...)` the only first-class way to construct colors in the new icons API?

Right now I only see the core model exposing `RGBA(red, green, blue, alpha)` plus `toHex()`, but not corresponding convenience constructors/parsers for other common forms like RGB, hex strings, HSB/HSV/HSL, etc. That makes the API feel a bit bare for something that is likely to be authored directly.

Relatedly, I think the color space should be documented explicitly here. My assumption is that these channel values are meant to be sRGB, but that feels important enough to say outright — especially in an API that is trying to be portable/serializable across runtimes. “RGBA” without an explicit color-space contract is too underspecified.

platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/IntelliJIconManager.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.jetbrains.icons.impl.intellij.design.IntelliJIconDesigner
+import org.jetbrains.icons.impl.intellij.rendering.IntelliJIconRendererManager
+import org.jetbrains.icons.impl.intellij.rendering.SwingIcon
+import org.jetbrains.icons.impl.intellij.rendering.images.IntelliJImageResourceProvider
+import org.jetbrains.icons.rendering.IconRendererManager
+import org.jetbrains.icons.rendering.ImageResourceProvider
+import java.lang.ref.WeakReference
+import kotlin.getValue
+
+class IntelliJIconManager : DefaultIconManager() {
```
rock3r, line 35: I’m trying to understand whether the current deferred-icon resolver architecture is intentionally minimal, or just not finished yet.

From the code, the main strategy selection point seems to be `DefaultIconManager.createDeferredIconResolver(...)`, which suggests the design *could* support different resolver kinds. But in practice I only see one real implementation path (`InPlaceDeferredIconResolver`), and in the IntelliJ impl the “non-local” branch currently just fails:

```kotlin
if (evaluator == null) {
  throw NotImplementedError("Remote Icon evaluation is not supported")
}
```

Given the broader emphasis here on identifiers/caching/synchronisation / wire-friendliness, that makes the current IntelliJ behavior feel more like a TODO than a fully specified architecture. I seem to remember that being a known issue, but worth underlining again.

Is the lack of actual resolver pluggability here intentional for now — i.e. the design is deliberately “single local resolver only” at this stage — or is this accidental/incomplete, with the expectation that IntelliJ will eventually plug in additional resolver strategies for deserialized / remote-backed deferred icons?

Right now the code shape suggests extensibility, but the actual IJ implementation seems to stop at the first local case and leave the rest unsupported.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Icon.kt
rock3r, line 91: I think there is a broader caching/sharing/instance-stability question here that is still unclear to me, and this composable seems like the most visible place where it shows up.

A few related concerns:

- If callers build icons directly in composition via `icon { ... }`, it seems they currently need to `remember { ... }` them themselves to avoid recreating the model, renderer, and painter churn. That sounds like the correct usage, but if so, it would be good to document it explicitly in KDoc, because it is not obvious from the API shape.
- More generally, I find the difference in behavior between different icon kinds worrying. Deferred icons have identifier-based sharing, but ordinary icons seem to rely mostly on caller-side instance reuse. Since icons can be composed from other icons, this feels like a place where subtle inconsistencies and perf bugs will creep in, especially as this is not well-documented and may change at any time.
- On the Compose side, this makes me think we probably do need some centralized caching/sharing story — at least for common model instances and perhaps some resource/renderer-level reuse. Right now it looks fairly local/per-instance. This is a known limitation of the existing Jewel `IconKey` API — we never addressed it because we've been waiting for this work to happen for a while. It's a critical gap that needs addressing.
- On the Swing side, the equivalent concerns seem even more footgun-prone (`toSwingIcon()` returning fresh wrappers, caller-side reuse expectations, etc.), which makes the asymmetry between frontends quite concerning.

Stepping back: should the intended behavior be more uniform for all icon kinds? It feels like we probably want an `AllIcons`-style equivalent for the new model layer as the baseline way of ensuring stable/shared model instances, even if more sophisticated caching of composed models/rendered results comes later.

That under-the-hood caching part feels like a plausible v2 enhancement rather than something that necessarily requires API changes now. However, I think the expected sharing/caching model should be made much clearer — otherwise different icon kinds may continue to behave opaquely differently across Compose and Swing.

platform/icons-impl/intellij/src/org/jetbrains/icons/impl/intellij/rendering/images/IntelliJImageResource.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.jetbrains.icons.rendering.ImageScale
+import org.jetbrains.icons.rendering.RescalableImageResource
+
+internal class IntelliJImageResource(
+  private val holder: SwingImageResourceHolder,
+  private val imageModifiers: ImageModifiers? = null
+): RescalableImageResource, CachedGPUImageResourceHolder() {
+  private val expectedDimensions by lazy { holder.getExpectedDimensions() }
+
+  override fun scale(scale: ImageScale): BitmapImageResource {
```
rock3r: I think the new image-icon path may currently be regressing an important part of the existing IJPL icon behavior around size-specialized SVG variants.

As far as I can tell, if I author `image("foo.svg", ...)`, and there is also a `foo@20x20.svg` on disk, painting the icon at 20×20 here does **not** seem to pick the specialized asset. Instead, this path appears to rerender the base `foo.svg` at the requested scale.

That looks like a regression relative to the current IJPL icon stack, which does have explicit support/utilities for loading custom size variants. It also seems out of line with current Jewel resource loading behavior, where `@20x20` variants are a known concept (see the `Size` `PainterHint` IIRC).

Was this omission intentional? If not, this feels like a behavior-contract breach that could cause subtle visual regressions exactly in the cases where teams bothered to provide size-specialized SVGs in the first place.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/icon/PathImageResourceLocation.kt
```diff
@@ -0,0 +1,36 @@
...
+    public fun loadData(imageModifiers: ImageModifiers?): ByteArray {
+        val knownMods = imageModifiers as? DefaultImageModifiers
+        val finalPath = applyPathModifiers(path, knownMods)
+        val resourceStream = if (classLoader != null) {
+            classLoader.getResourceAsStream(finalPath)
+        } else ClassLoader.getSystemResourceAsStream(path) 
+        return resourceStream?.readBytes() ?: error("Resource not found: $finalPath")
+    }
+    
+    private fun applyPathModifiers(path: String, modifiers: DefaultImageModifiers?): String {
```
rock3r, line 25: This looks like a regression relative to Jewel’s current resource-loading behavior.

At the moment this path seems to support only exact-path loading plus `_dark` specialization, but not size-specialized variants like `foo@20x20.svg`. Current Jewel resource loading/painter hints already treat `@WxH` variants as a first-class concept, so the new icon path appears to be losing that capability rather than preserving it.

That creates an awkward cross-mode inconsistency:
- old/current Jewel resource loading knows about sized variants
- current IJPL icon utilities know about sized variants
- but the new icon image path seems not to

Was that intentional? If not, this looks like a real regression in image selection behavior, not just an implementation detail.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/icon/ComposeImageResourceProvider.kt
```diff
@@ -0,0 +1,137 @@
...
+@ApiStatus.Internal
+public class ComposeImageResourceProvider : ImageResourceProvider {
+    override fun loadImage(location: ImageResourceLocation, imageModifiers: ImageModifiers?): ImageResource {
+        // TODO Support image modifiers
+        if (location is PathImageResourceLocation) {
+            val extension = location.path.substringAfterLast(".").lowercase()
+            val data = location.loadData(imageModifiers)
+            val stream = ByteArrayInputStream(data)
+            return when (extension) {
+                "svg" -> ComposePainterImageResource(patchSvg(imageModifiers, stream).decodeToSvgPainter(Density(1f)), imageModifiers)
```
rock3r: One thing worth clarifying/documenting here is that the SVG rendering pipeline now diverges quite a bit across runtimes:

- in the IntelliJ/Swing path, SVGs are rendered to raster images and cached through the existing IJPL SVG cache machinery
- here in the standalone Compose path, `decodeToSvgPainter(...)` keeps an `SVGDOM`/painter representation and uses a per-painter draw cache keyed by draw size

That divergence may be perfectly acceptable, but it means the new icon system is not just sharing one rendering strategy across runtimes — it is sharing the model while relying on fairly different SVG realization pipelines underneath. It would be good to know whether that difference is intentional and expected. I don't think that would be the right direction to go in, but there may be aspects I have not considered.

platform/icons-impl/src/org/jetbrains/icons/impl/rendering/layers/IconIconLayerRenderer.kt
```diff
@@ -0,0 +1,37 @@
...
+  private val layer: IconIconLayer,
+  private val renderingContext: RenderingContext
+) : IconLayerRenderer {
+  private val renderer = layer.icon.createRenderer(renderingContext.adjustTo(layer))
```
rock3r: This looks like the place where nested icon renderers inherit the parent rendering context/update flow, which suggests deferred/animated children are expected to report updates through the same root flow.

That seems plausible by construction, but I couldn’t find tests that actually exercise nested update behavior — e.g., animated icons inside composed/layout icons, deferred icons nested inside other icons, or combinations of both. Given how subtle these lifecycle/update interactions can get, and how error prone this all is, do we have dedicated coverage for that somewhere?

I only found a very basic rendering smoke test, but not tests validating that nested update flows combine correctly or continue to behave correctly as composition gets more complex.

platform/icons-api/src/org/jetbrains/icons/design/IconDesigner.kt
```diff
@@ -0,0 +1,49 @@
...
+  align: IconAlign = IconAlign.TopRight,
+  cutout: IconUnit = 1.5.dp relativeTo 20.dp,
+  modifier: IconModifier = IconModifier,
+) {
+  shape(color, shape, modifier.size(size).align(align).cutoutMargin(cutout))
+}
+
+@ApiStatus.Experimental
+interface IconAnimationDesigner {
+  fun frame(duration: Long, builder: IconDesigner.() -> Unit)
```
rock3r: I think the animation API needs to state a clearer bounds/size contract.

Right now `frame(duration) { ... }` appears to allow each frame to contain an arbitrary icon/layer set, which in turn suggests that different frames may have different measured bounds. That seems dangerous for host layout/intrinsic-size stability, especially across Swing/Jewel/standalone frontends, and particularly because of the issues I've already found with cross-frontend consistency.

Do we actually intend to support frame-by-frame size changes? If yes, I think that needs to be explicitly specified and handled throughout the rendering/layout pipeline.

However, I reckon the safer contract is that all animation frames must have the same effective size, and that this size should be determined upfront (e.g., from an explicit animation size parameter, or from a precomputed max over all frames). It then needs to be treated as stable for the lifetime of the animated icon, clipping whatever falls outside those bounds.

Whatever is the intended contract, it should be enforced and documented at animation creation time rather than left implicit like we do in the current renderer behavior

platform/icons-api/src/org/jetbrains/icons/IconManager.kt
```diff
@@ -0,0 +1,88 @@
...
+   * Use convenience top-level method icon {} instead if possible.
+   */
+  fun icon(designer: IconDesigner.() -> Unit): Icon
+
+  /**
+   * @see org.jetbrains.icons.deferredIcon
+   */
+  fun deferredIcon(placeholder: Icon?, identifier: String? = null, classLoader: ClassLoader? = null, evaluator: suspend () -> Icon): Icon
+
+  suspend fun forceEvaluation(icon: DeferredIcon): Icon
```
rock3r: This is at a minimum named in a misleading way, since this is a no-op when a deferred icon was already evaluated. But depending on the intent, this may just not be working as expected — did you intend this as a way to force a re-evaluation of an icon?

If yes, the implementation breaks when we get to `InPlaceDeferredIconResolver.resolve()` as there is a short-circuit there: `if (resolved != null) return resolved`

If no, this should be called something more like `awaitEvaluation`, `resolveIfNeeded`, or similar, since it doesn't actually seem to force anything, and it's not really like you can do anything but wait for resolution to happen as it would anyway.

The only real use case for this seems to be when power saving mode is on and the deferred icons are not resolved automatically. Tbh that seems like a surprising use case to me, since the issue should be handled at the usage site, not systematically.

I also don’t see the power-save behavior documented anywhere in the deferred-icon API/docs/KDoc.

That seems rather important because it materially affects the behavioral contract: callers might reasonably assume deferred icons will eventually resolve automatically, whereas in power save mode they apparently may not. If that is an intentional part of the design, I think it should be stated explicitly rather than only discoverable from the IntelliJ implementation and assumed to be a bug.

In any case, having to have such an escape hatch for that scenario shows pretty clearly to me that it's not a great idea to have it in the first place. What do you do in that case? Call this function on ALL your icons, defeating the whole purpose of the feature, and in the process leaking an unrelated concern into UI code?

I understand this last part is a change outside the scope of this PR, but could be a good opportunity to revisit it, imo.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Icon.kt
```diff
@@ -76,11 +88,51 @@ public fun Icon(
...
+    val renderer = remember(icon) {
+        icon.createRenderer(context)
+    }
+    val painter = remember(icon, renderer) {
+        RendererBasedIconPainter(
+            renderer,
+            scaling = scalingContext
+        )
+    }
+    key(updateIndex) {
```
rock3r: This currently turns every deferred/animated icon update into a Compose state write + recomposition:

- `updateFlow` callback writes `updateIndex`
- `key(updateIndex)` rebuilds this subtree
- then we redraw

That seems heavier than necessary if the intended contract is that these updates are paint-only.

I’m working with the assumption that animated icons are **not** allowed to change size frame-by-frame (which I think is the safer contract), and I suspect we may need the same assumption for deferred icons too: once created, they should have a stable expected size, even before resolution. Without that, redraw-only invalidation would not be enough because icon updates could also require layout invalidation. Recompositions are expensive and can cause cascading relayouts which we should try to avoid where possible.

Assuming stable size, I think this should avoid recomposition on every animation tick / deferred update and invalidate **draw** instead.

Two concrete options:

### 1) Minimal refactor: keep a stable remembered painter and let the painter subscribe to the update flow

This follows the same general pattern as Compose’s own stateful painters (`VectorPainter`, `DrawablePainter`): the painter owns a small piece of mutable state that is read from `onDraw()`, so updating it invalidates the draw phase without forcing recomposition of the composable subtree.

At a high level, the composable would then look like this:

```kotlin
@Composable
public fun Icon(
    icon: Icon,
    contentDescription: String?,
    loadingStrategy: LoadingStrategy = LoadingStrategy.BlockThread, // TODO use it!
) {
    val scalingContext = RendererBasedIconPainter.inferScalingContext()
    val isDark = JewelTheme.isDark

    val updateFlow = remember(scope) {
        IconRendererManager.createUpdateFlow(scope) {
            // No Compose state write here
        }
    }

    val context = remember(updateFlow, isDark) {
        IconRendererManager.createRenderingContext(
            updateFlow = updateFlow,
            defaultImageModifiers = DefaultImageModifiers(
                isDark = isDark,
                // Others?
            ),
        )
    }

    val renderer = remember(icon, context) { icon.createRenderer(context) }
    val painter = remember(scalingContext, renderer, updateFlow, scope) {
        RendererBasedIconPainter(
            iconRenderer = renderer,
            scaling = scalingContext,
            updateFlow = updateFlow,
            scope = scope,
        )
    }

    Icon(
        painter = painter,
        contentDescription = contentDescription,
        modifier = modifier,
        // etc
    )
}
```

Then in `RendererBasedIconPainter.kt`:

```kotlin
public class RendererBasedIconPainter(
    private val iconRenderer: IconRenderer,
    private val scaling: ScalingContext,
    private val updateFlow: MutableIconUpdateFlow,
    private val scope: CoroutineScope,
) : Painter(), RememberObserver {

    // This is now safe since we enforce animation and deferred sizes to be fixed over time
    override val intrinsicSize: Size =
        iconRenderer.calculateExpectedDimensions(scaling).toComposeSize()

    private var drawTick by mutableIntStateOf(0)
    private var updateJob: Job? = null

    override fun onRemembered() {
        if (updateJob == null) {
            updateJob = scope.launch {
                updateFlow.collect {
                    // Will eventually overflow — it's fine, we don't care
                    // about the actual value here, only its changes
                    drawTick++
                }
            }
        }
    }

    override fun onForgotten() {
        updateJob?.cancel()
        updateJob = null
    }

    override fun onAbandoned() {
        onForgotten()
    }

    override fun DrawScope.onDraw() {
        // Draw-phase state read; updates invalidate draw, not composition
        // This is a trick to force a redraw when there is a new draw tick
        // conceptually similar to key(...) but without recomposition
        drawTick

        val api = ComposePaintingApi(this, scaling = scaling)
        val layerRect = Rect(Offset.Zero, Size(size.width, size.height))
        drawIntoCanvas { canvas ->
            canvas.withSaveLayer(layerRect, obtainPaint()) {
                iconRenderer.render(api)
            }
        }
    }
}
```

That would keep recomposition for actual structural changes (`icon`, theme, density, etc.), but not for animation ticks.

### 2) Longer-term: skip `Painter` for this path and render via a custom draw node / `DrawModifierNode`

That gives the most direct model: subscribe to `updateFlow` in the node and call `invalidateDraw()` on each icon update.

```kotlin
private class IconRenderNode(
    var renderer: IconRenderer,
    var scaling: ScalingContext,
    var updateFlow: MutableIconUpdateFlow,
) : Modifier.Node(), DrawModifierNode {

    private var job: Job? = null

    override fun onAttach() {
        job = coroutineScope.launch {
            updateFlow.collect {
                invalidateDraw()
            }
        }
    }

    override fun onDetach() {
        job?.cancel()
        job = null
    }

    override fun ContentDrawScope.draw() {
        val api = ComposePaintingApi(this, scaling = scaling)
        renderer.render(api)
    }
}
```

That is probably the cleanest end state if this path is expected to handle lots of animated icons efficiently.

But either way, I think the key point is the same: **if icon updates are supposed to be paint-only, this composable should not need to recompose on every update**.

platform/icons-api/rendering/src/com/intellij/platform/icons/rendering/IconRenderer.kt
rock3r, line 1: I think this PR needs much stronger test coverage before merge.

The new icon system is no longer just "paint an image"; it is now a compositional tree with layout, sizing, modifiers, deferred resolution, animation, and toolkit-specific update/repaint behavior. That means a lot of correctness now lives in the alignment between **measurement** and **rendering**, and that is exactly where I would expect subtle regressions.

I only found a very basic smoke test that renders a new icon to a `BufferedImage`, but I could not find dedicated coverage for the cases that seem most failure-prone:

- bounds propagation through nested icon/layout trees
- percentage-based sizing relative to parent bounds
- measurement vs rendering consistency for `row` / `column` / `box`
- spacing / alignment / margin interactions in nested layouts
- placeholder vs resolved deferred-icon size changes
- animation frame sizing / bounds stability
- nested deferred + animated update propagation
- Compose-side invalidation/repaint behavior for deferred/animated icons
- Swing-hosted legacy/custom layers participating correctly in sizing and repaint

Even if some of these behaviors are intentionally unsupported, I think that contract still needs to be made explicit and then locked down with tests. Right now the absence of coverage here makes me pretty nervous, because these are exactly the kinds of issues that tend to look fine in simple demos and then break in real UI usage.

My strong suggestion would be:
- add **unit tests** for layout/measurement/modifier semantics wherever possible
- add **Compose UI tests** for repaint/invalidation behavior on the Jewel side
- add at least a few **Swing integration tests** for `toSwingIcon()` / legacy-layer interop / deferred repaint behavior

platform/icons-impl/src/com/intellij/platform/icons/impl/rendering/layers/ShapeIconLayerRenderer.kt
rock3r, line 1: `calculateExpectedDimensions()` always returns `0 × 0`, but `render()` fully respects bounds/modifiers and will paint a visible shape as soon as the host supplies explicit bounds.

Is the intent that shape layers are purely bounds-relative and therefore have no meaningful intrinsic size? If so, that seems too restrictive for a public primitive — e.g. it should be possible to describe a 4×4 rectangle/circle regardless of host bounds. If not, this looks outright broken for intrinsic measurement.

image

Local repro from the playground. This icon is just a 12dp rectangle shape, but `toSwingIcon()` reports `0 × 0`. That seems very hard to reconcile with the current public API unless shape layers are intentionally considered to have no intrinsic size at all — which would be a surprising and very limiting contract for a general shape primitive.

platform/icons-impl/src/com/intellij/platform/icons/impl/rendering/layers/ImageIconLayerRenderer.kt
rock3r, line 1: Broad question: what is the intended contract for layer modifiers in intrinsic measurement?

Rendering applies `width` / `height` / `margin` / `align` / etc. through `layer.modifier.applyTo(...)`, but this `calculateExpectedDimensions()` path just returns the raw image size. I see the same pattern in several renderers, which suggests icons can be measured with one set of semantics and painted with another.

Is that intentional, or is this an oversight?

platform/icons-impl/src/com/intellij/platform/icons/impl/rendering/layers/LayoutIconLayerRenderer.kt
rock3r, line 1: `calculateExpectedDimensions()` for rows/columns ignores `spacing`, so the intrinsic size under-reports the actual footprint immediately.

More importantly, the measurement semantics here do not seem to match rendering semantics: measurement sums child widths/heights, but `render()` distributes the available extent into equal slots. So a row/column is measured like a content-sized layout, but painted like an equal-slot layout.

Was that intentional?

image

Screenshot from the local playground. This is the same `row(spacing = 25.percent) { ... }` icon rendered through `toSwingIcon()` under two host sizes. Instead of behaving like a small icon composed of two children plus spacing, it gets horizontally redistributed/squeezed based on the available host width. That strongly suggests the row layout is being resolved against host/component bounds rather than a stable icon canvas, and also that row measurement/render semantics are not aligned.

rock3r, line 1: In the `Box` branch, `finalBounds` is computed but then ignored; children are rendered against `api` rather than a nested API bound to the box's final bounds.

That makes it look like box-layer modifiers (`size`, `align`, `margin`, etc.) do not actually constrain/reposition the nested content. Is that really intended?

image

Local playground repro. The two icons differ only in where the same `size(50.percent).align(IconAlign.BottomRight)` modifier is attached:

- left: modifier on the `box` layer
- right: modifier on the child icon directly

These should be much closer if the box’s computed bounds are actually constraining/repositioning its nested content. The visible mismatch seems consistent with the `Box` render path computing `finalBounds` and then ignoring them.

platform/icons-impl/src/com/intellij/platform/icons/impl/InPlaceDeferredIconResolver.kt
rock3r, line 1: One more edge case here: what happens if `evaluator()` throws?

As written:
- `isPending` stays `true`
- `deferredValue` is never completed exceptionally
- `resolvedIcon` stays `null`

So after one failed evaluation, later callers seem to fall into `deferredValue.await()` and potentially wait forever. That would effectively wedge the resolver after the first failure.

Should this complete the deferred exceptionally (and/or reset the pending state) so failure is surfaced deterministically rather than poisoning future resolves?

platform/icons-impl/src/com/intellij/platform/icons/impl/DeferredIconResolverService.kt
rock3r, line 1: This looks like the place where the weak-ref/shareability problem should probably be solved explicitly.

If the resolver is still alive but `resolver.deferredIcon.get()` is already `null`, this currently returns a fresh `DefaultDeferredIcon(identifier, placeholder)`. That seems to break the core shareability contract of deferred icons, because the durable state (pending/resolved/error) then lives in one object while callers get handed a different icon instance that is no longer properly attached to it.

I think the cleaner fix is for the resolver to own the canonical deferred state independently of any one `DefaultDeferredIcon` instance, and treat the icon object as a lightweight handle/view over that state. Then if the old weakly-referenced icon instance was GC'd, a new one could be created and rebound to the existing resolver state instead of starting a disconnected lifecycle.

Was something like that the intended shape here?

platform/icons-impl/src/com/intellij/platform/icons/impl/rendering/DefaultDeferredIconRenderer.kt
rock3r, line 1: This seems like the deferred-icon equivalent of the animation size-contract problem.

Right now `calculateExpectedDimensions()` just delegates to whichever renderer is active at the moment, so a deferred icon can report one intrinsic size while the placeholder is shown and a different one after resolution. That makes deferred completion a potential layout event rather than a paint-only update.

I think the safer contract is that deferred icons must have a stable expected size known upfront for their whole lifetime (e.g. explicit size, or some precomputed expected bounds), and that this contract should be documented at the deferred-icon API level as well. Otherwise every frontend needs a real size-change / relayout story here, not just repaint/update notifications.

platform/icons-impl/src/com/intellij/platform/icons/impl/rendering/layers/LayoutIconLayerRenderer.kt
rock3r, line 1: What is the intended contract for invalid/degenerate layout cases here?

A few edge cases seem problematic:
- `nestedRenderers.count()` can be zero here, which would divide by zero
- if spacing is larger than the available extent, `remainingSize / nestedRenderers.count()` can go negative

That means this code can produce invalid child bounds, and Compose in particular is not tolerant of negative draw sizes/bounds. Do we want to clamp to zero, reject such layouts up front, or define some other fallback behavior?

platform/icons-impl/src/org/jetbrains/icons/impl/rendering/modifiers/MarginIconModifier.kt
rock3r, line 1: This can produce negative `width` / `height` when the margins exceed the available parent bounds.

Do we have a defined contract for that? If negative bounds are allowed to leak further down the pipeline, Compose can crash on them during draw. It seems like this path should either clamp to zero or validate/reject impossible layouts earlier.

platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/SwingIconLayerRenderer.kt
rock3r, line 1: I think this may be missing an update/repaint notification.

When the wrapped legacy `DeferredIcon` finishes, this path updates its cached `image`, but I don't see it notifying the new icon renderer/frontend via `renderingContext.updateFlow.triggerUpdate()` (or equivalent). If so, the visual update would depend on some unrelated repaint happening later.

That seems especially risky for the migration/interop path, because `javax.swing.Icon.toNewIcon()` can wrap exactly these old deferred icons.

platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/SwingIcon.kt
rock3r, line 1: I don't think `x` / `y` should be affecting the backing-layer dimensions here.

In Swing, `paintIcon(c, g, x, y)` uses `x`/`y` as the paint offset, not as a reduction in the icon's available size. Painting the same icon at `(0, 0)` vs `(10, 10)` should translate it, not shrink/crop the backing buffer by 10 pixels in each dimension.

As written, this seems likely to produce clipped/incorrect output for icons painted at non-zero coordinates.

platform/icons-impl/intellij/src/org/jetbrains/icons/impl/intellij/rendering/custom/CustomIconLayerRendererProvider.kt
rock3r, line 1: Is this a bug? The extension point says custom providers are supposed to implement `createRenderer(...)`, but this path checks `extension.handles(layer)` and then unconditionally returns `SwingIconLayerRenderer(...)` instead of delegating to `extension.createRenderer(...)`.

If I am reading this correctly, custom renderer providers can never actually provide their own renderer implementation here.

platform/icons-impl/src/com/intellij/platform/icons/impl/layers/LayoutIconLayer.kt
rock3r, line 1: `spacing` seems to be missing from both `equals()` and `hashCode()` here.

That means two layout layers that differ only by spacing compare equal, even though spacing materially affects rendering. Given how much this new icon model seems to rely on value-like model objects, caching/reuse, and caller-side remembering, this looks like a real correctness bug rather than just an implementation detail.

rock3r (commented):
As we talked about offline, I’m adding this here mostly to document that conversation in the review.

A broader architectural concern I still have is how this is supposed to work cleanly in **Jewel standalone**, which is a real target for this API and not optional.

Right now it still feels like some important parts of icon resource resolution are effectively trapped inside the IntelliJ Platform icon-loading stack (`IconLoader` / `util-ui/icons`), especially around path transformation / mapping / variant selection. That makes me nervous, because I don’t think we want either of these outcomes:

- Jewel standalone depending directly on the full IJPL icon-loading implementation
- or duplicating the same path-mapping logic in a separate standalone stack

Given that standalone support is a hard requirement, would it make sense to explicitly extract the *resource-path resolution* part into a smaller backend-neutral module that both IJPL and Jewel standalone can depend on?

By “path resolution” I mean the logic that takes a logical icon path and environment/context and produces the effective resource candidates / transformations (e.g. theme/UI variants, size-specialized variants, any other canonical path mapping rules). The actual loading/caching/rendering backends can still remain platform-specific, but the mapping rules themselves seem like they want to be shared.

Otherwise I worry we end up with an awkward split where:
- the icon *model* is portable
- the *rendering* backends are separate
- but the *resource resolution semantics* are still IntelliJ-specific and therefore either unavailable or reimplemented differently in standalone

That feels like exactly the sort of thing that will cause subtle divergence over time. Since standalone support is non-negotiable here, extracting that path-resolution layer seems like the cleaner long-term shape.

And to be explicit: I do think this is part of the implementation work required for this feature, and it will need to be addressed in this PR before merge.

rock3r:
I think this PR still needs much stronger test coverage before merge.

The new icon system is no longer just "paint an image"; it is now a compositional tree with layout, sizing, modifiers, deferred resolution, animation, and toolkit-specific update/repaint behavior. That means a lot of correctness now lives in the alignment between **measurement** and **rendering**, and that is exactly where I would expect subtle regressions.

I only found a very basic smoke test that renders a new icon to a `BufferedImage`, but I could not find dedicated coverage for the cases that seem most failure-prone:

- bounds propagation through nested icon/layout trees
- percentage-based sizing relative to parent bounds
- measurement vs rendering consistency for `row` / `column` / `box`
- spacing / alignment / margin interactions in nested layouts
- placeholder vs resolved deferred-icon size changes
- animation frame sizing / bounds stability
- nested deferred + animated update propagation
- Compose-side invalidation/repaint behavior for deferred/animated icons
- Swing-hosted legacy/custom layers participating correctly in sizing and repaint

Even if some of these behaviors are intentionally unsupported, I think that contract still needs to be made explicit and then locked down with tests. Right now the absence of coverage here makes me pretty nervous, because these are exactly the kinds of issues that tend to look fine in simple demos and then break in real UI usage.

My strong suggestion would be:
- add **unit tests** for layout/measurement/modifier semantics wherever possible
- add **Compose UI tests** for repaint/invalidation behavior on the Jewel side
- add at least a few **Swing integration tests** for `toSwingIcon()` / legacy-layer interop / deferred repaint behavior

rock3r:
PS I have a few tweaks/samples to propose that I referred to in the review https://github.com/JetBrains/intellij-community/compare/public/jakub.senohrabek/new-icons-api...rock3r:sebp/new-icons-tweaks

rock3r:
Superseeded

PR #3419, opened 2026-02-11
[JEWEL-1205] Fix Menu Items Looking Too Tall in Presentation Mode

DanielSouzaBertoldi:
# Context

Our menu items are looking a little too big when opening a ✨ Jewel ✨ menu in presentation mode. Curiously, sub menu items don't have the same issue. Check the evidences section below to see how menu items were being rendered.

Now, the fix is really, really, REALLY simple. HOWEVER, it took me a **really long time** to finally crack this case open and I'm kinda mad because of it. If you wish to read just one more novel of mine (I promise this won't be the last), check the collapsible section below (and don't forget to bring a glass of whatever is your favorite drink to drink while code reviewing this --I promise I won't judge you for this, even if you decide on drinking sparkling water you goofball)

<details><summary>The curve ball tale 🕵🏻 </summary>

Alright first of all, one of the things that really stood out to me was that `MenuItemBase` had a `.defaultMinSize(minHeight = itemMetrics.minHeight)` while `MenuSubmenuItem` did not.

> _Interesting._

I thought. I commented out this modifier from `MenuItemBase`, ran the IDE project, tested with presentation mode both on and off, and the issue was fixed!

> _Nice._

I celebrated. This looked like the world famous "quick win" that developers are so eager to have around. But then, it hit me: if I were to remove this modifier, what would happen to `itemMetrics.minHeight`? Would we need to just delete it for good? No, that doesn't seem to be the right thing to do.

Then, the next step of this quest started: trying to find out where the code of Intellij's Menu was and how they were applying the height for each menu item. This by itself took a long time, since there are so many intertwined classes and whatnot. Anyways, this is the boring part so I'll skip ahead.

I finally struck gold when I came across two classes: `BegMenuItemUI.java` and `IdeaMenuUI.java`. The former is the class that is used to render menu items in an `ActionPopupMenu`, exactly what I needed. The latter is basically a helper class.

Inside the `getPreferredSize()` function from `BegMenuItemUI`, there was this call at the end:

`IdeaMenuUI.patchPreferredSize(comp, rect.getSize());`

Here's the code `IdeaMenuUI.patchPreferredSize()`:

```java
public static @NotNull Dimension patchPreferredSize(Component c, Dimension preferredSize) {
  if (ExperimentalUI.isNewUI() && !IdeaPopupMenuUI.isMenuBarItem(c)) {
    JBInsets outerInsets = IdeaPopupMenuUI.isPartOfPopupMenu(c)
                           ? JBUI.CurrentTheme.PopupMenu.Selection.outerInsets()
                           : JBUI.CurrentTheme.Menu.Selection.outerInsets();
    return new Dimension(preferredSize.width, JBUI.CurrentTheme.List.rowHeight() + outerInsets.height());
  }

  return preferredSize;
}
```

> _No way._

I muttered. This code is basically proving to us that, if the user is using the new ui, then the height for the menu item is just:

`JBUI.CurrentTheme.List.rowHeight() + outerInsets.height()`

Which we were already doing. Of couse, we don't calculate both and set the result as the total height of a menu item, but we do apply the `outerInsets` + the min height for the menu item given the value in `rowHeight()`.

That means that our code was correct after all. So why was this behavior happening? That didn't make any sense. We could "fix it" by just removing the `minHeight` modifier, but why?

After adding some logs (basically just adding a print statement inside `onSizeChanged {}`), I could guarantee that the min height we were setting was way smaller than the items actual size:

```
[MenuItemBase] minHeight: 24.0.dp, actual height: 48  // presentation mode disabled
[MenuItemBase] minHeight: 42.0.dp, actual height: 147 // presentation mode enabled
```

Nothing was making sense anymore. The min height is too small. It's not like we are forcing the item to have the same height as `minHeight`.

I went over IntelliJ's code over and over, tried to add the same calculations on Jewel, the bug kept happening. I finally gave up since I ran out of ideas and asked the OG Gemini about it and....... I couldn't believe the answer.

It was a double scaling error all along. At first I was skeptical but once I tried using the unscaled value of `rowHeight()`, the bug disappeared. No need to delete the `minHeight` modifier. I was flabbergasted to say the least.

It then proceeded to explain to me why the double scaling:

`BUI.CurrentTheme.List.rowHeight()` already returns an scaled value. So for instance, it can be scaled by `1.0f` for "normal" mode and `1.75f` for presentation mode. When we call `.dp` from Compose, it also scales the value by a given factor (taking into account your screen display/pixel density).

So, the base height for rows is `24`. In presentation mode it gets scaled up to `42` (24x1.75). Then, Compose does its thing and scales up by `3.5f` in my case (2.0f - retina display * 1.75 - from IJ's). `42 * 3.5f = 147`. Bam. That's exactly the actual height of a menu item when presentation mode was enabled.

The reason `minHeight` caused this is because we were basically forcing Compose to apply this scaling factor on `42` instead of the unscaled value of `24`.

I never would've thought that this could be a scaling problem. Well, you live and you learn right? At least next time I'll also be suspicious of scaling issues.

Thanks for reading, leave your review on a piece of paper and mail it to me 😎 

</details>

# Changes
- Use the **unscaled value** of `JBUI.CurrentTheme.List.rowHeight()` in `org.jetbrains.jewel.bridge.theme.IntUiBridgeMenu`
- Also took this opportunity to refactor a bit of menu item/sub menu item code. They had quite a bit of duplicated code. Yep, this means I added yet another layer of Composable
   - Previously sub menu items didn't have this scaling problem because we weren't setting the minHeight. This was fixed with this refactor
- Created a handy `unscaledDp` function in `BridgeUtils`

# Evidences

|  | Screen Recording |
|--------|--------|
| Before | <video src="https://github.com/user-attachments/assets/ffcdfdb3-7322-4627-8187-2cb4a515302b" /> |
| After | <video src="https://github.com/user-attachments/assets/e833f8ed-48e0-465f-b4c6-ecec9b52400c" /> |

### Bug fixes
 * Menu items now keep their size when rendered in Presentation Mode

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
@@ -42,6 +42,7 @@ import com.intellij.ui.scale.JBUIScale.scale
 import com.intellij.util.ui.JBDimension
 import com.intellij.util.ui.JBFont
 import com.intellij.util.ui.JBInsets
+import com.intellij.util.ui.JBUI
 import com.intellij.util.ui.JBValue
 import java.awt.Dimension
 import java.awt.Insets
@@ -520,3 +521,58 @@ public fun retrieveEditorColorScheme(): EditorColorsScheme {
     val manager = EditorColorsManager.getInstance() as EditorColorsManagerImpl
     return manager.schemeManager.activeScheme ?: DefaultColorSchemesManager.getInstance().firstScheme
 }
+
+/**
+ * Retrieves a pre-scaled integer value from the current LaF as a non-negative [Dp], or returns [Dp.Unspecified] if the
+ * key is not found.
+ *
+ * Any negative values read from the LaF will be coerced to `0.dp`.
+ *
+ * Use this instead of [retrieveIntAsNonNegativeDpOrUnspecified] for keys whose values are explicitly rescaled by
+ * `patchHiDPI` in `LafManagerImpl`. See [unscaledDp] for the full list of affected keys and an explanation of why
+ * unscaling is needed.
+ *
+ * @param key The key to look up the integer with.
+ * @return The unscaled integer value from the LaF as a [Dp], or [Dp.Unspecified] if the key is not found.
+ * @see unscaledDp
+ */
+public fun retrieveUnscaledIntAsNonNegativeDpOrUnspecified(key: String): Dp =
+    retrieveUnscaledIntAsDpOrUnspecified(key).takeIf { it.isSpecified }?.safeValue() ?: Dp.Unspecified
+
+/**
+ * Retrieves a pre-scaled integer value from the current LaF as a [Dp], or returns [Dp.Unspecified] if the key is not
+ * found.
+ *
+ * Use this instead of [retrieveIntAsDpOrUnspecified] for keys whose values are explicitly rescaled by `patchHiDPI` in
+ * `LafManagerImpl`. See [unscaledDp] for the full list of affected keys and an explanation of why unscaling is needed.
+ *
+ * @param key The key to look up the integer with.
+ * @return The unscaled integer value from the LaF as a [Dp], or [Dp.Unspecified] if the key is not found.
+ * @see unscaledDp
+ */
+public fun retrieveUnscaledIntAsDpOrUnspecified(key: String): Dp {
+    val rawValue = UIManager.get(key)
+    if (rawValue is Int) return rawValue.unscaledDp
+    return Dp.Unspecified
+}
+
+/**
+ * Converts a pre-scaled JBUI value into a Compose [Dp].
+ *
+ * Certain integer values in UIDefaults are explicitly rescaled by `patchHiDPI` in `LafManagerImpl` after the theme JSON
+ * is applied. This means they already reflect the current JBUI scale factor. For example, `List.rowHeight` is stored as
+ * `24` in the theme JSON, but `patchHiDPI` overwrites it with `scale(24)`. So in Presentation Mode (scale=1.75),
+ * [com.intellij.util.ui.JBUI.CurrentTheme.List.rowHeight] returns `42`, not `24`.
+ *
+ * Passing such a pre-scaled value directly to `.dp` causes double-scaling:
+ * 1. `patchHiDPI` scales the stored value (e.g. 24 → 42 in Presentation Mode)
+ * 2. Compose scales it again via `LocalDensity` (e.g. `42.dp` → way too big)
+ *
+ * The keys currently explicitly rescaled by `patchHiDPI` are: `List.rowHeight`, `Table.rowHeight`, `Tree.rowHeight`,
+ * VCS log row height, `Tree.leftChildIndent`, `Tree.rightChildIndent`, and `SettingsTree.rowHeight`.
+ *
+ * For all other keys — those NOT rescaled by `patchHiDPI` — the value in UIDefaults is the raw logical integer from the
+ * theme JSON, and `.dp` is correct. Do NOT use this extension for those (e.g. via [retrieveIntAsDp] or similar).
+ */
+public val Int.unscaledDp: Dp
+    get() = JBUI.unscale(this).dp
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeLazyTree.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeLazyTree.kt
@@ -8,7 +8,7 @@ import androidx.compose.ui.unit.takeOrElse
 import com.intellij.util.ui.JBUI
 import org.jetbrains.jewel.bridge.dp
 import org.jetbrains.jewel.bridge.retrieveColorOrUnspecified
-import org.jetbrains.jewel.bridge.retrieveIntAsNonNegativeDpOrUnspecified
+import org.jetbrains.jewel.bridge.retrieveUnscaledIntAsNonNegativeDpOrUnspecified
 import org.jetbrains.jewel.ui.component.styling.LazyTreeIcons
 import org.jetbrains.jewel.ui.component.styling.LazyTreeMetrics
 import org.jetbrains.jewel.ui.component.styling.LazyTreeStyle
@@ -34,8 +34,8 @@ internal fun readLazyTreeStyle(): LazyTreeStyle {
             backgroundSelectedActive = selectedElementBackground,
         )
 
-    val leftIndent = retrieveIntAsNonNegativeDpOrUnspecified("Tree.leftChildIndent").takeOrElse { 7.dp }
-    val rightIndent = retrieveIntAsNonNegativeDpOrUnspecified("Tree.rightChildIndent").takeOrElse { 11.dp }
+    val leftIndent = retrieveUnscaledIntAsNonNegativeDpOrUnspecified("Tree.leftChildIndent").takeOrElse { 7.dp }
+    val rightIndent = retrieveUnscaledIntAsNonNegativeDpOrUnspecified("Tree.rightChildIndent").takeOrElse { 11.dp }
 
     return LazyTreeStyle(
         colors = itemColors,
@@ -49,7 +49,8 @@ internal fun readLazyTreeStyle(): LazyTreeStyle {
                         selectionBackgroundCornerSize = CornerSize(JBUI.CurrentTheme.Tree.ARC.dp / 2),
                         iconTextGap = 2.dp,
                     ),
-                elementMinHeight = retrieveIntAsNonNegativeDpOrUnspecified("Tree.rowHeight").takeOrElse { 24.dp },
+                elementMinHeight =
+                    retrieveUnscaledIntAsNonNegativeDpOrUnspecified("Tree.rowHeight").takeOrElse { 24.dp },
                 chevronContentGap = 2.dp, // See com.intellij.ui.tree.ui.ClassicPainter.GAP
             ),
         icons =
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
@@ -31,6 +31,7 @@ import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
@@ -821,35 +822,17 @@ internal fun MenuItemBase(
     style: MenuStyle = JewelTheme.menuStyle,
     @Suppress("DEPRECATION") content: @Composable (itemState: MenuItemState) -> Unit,
 ) {
-    var itemState by
-        remember(interactionSource) {
-            @Suppress("DEPRECATION") mutableStateOf(MenuItemState.of(selected = selected, enabled = enabled))
-        }
-
-    remember(enabled, selected) { itemState = itemState.copy(selected = selected, enabled = enabled) }
+    val itemState by rememberMenuItemState(selected, enabled, interactionSource)
 
     val focusRequester = remember { FocusRequester() }
-
-    LaunchedEffect(interactionSource) {
-        interactionSource.interactions.collect { interaction ->
-            when (interaction) {
-                is PressInteraction.Press -> itemState = itemState.copy(pressed = true)
-                is PressInteraction.Cancel,
-                is PressInteraction.Release -> itemState = itemState.copy(pressed = false)
-                is HoverInteraction.Enter -> {
-                    itemState = itemState.copy(hovered = true)
-                    focusRequester.requestFocus()
-                }
-
-                is HoverInteraction.Exit -> itemState = itemState.copy(hovered = false)
-                is FocusInteraction.Focus -> itemState = itemState.copy(focused = true)
-                is FocusInteraction.Unfocus -> itemState = itemState.copy(focused = false)
-            }
-        }
-    }
-
     val menuController = LocalMenuController.current
     val localInputModeManager = LocalInputModeManager.current
+
+    LaunchedEffect(itemState.isHovered) {
+        if (itemState.isHovered) {
+            focusRequester.requestFocus()
+        }
+    }
 
     Box(
         modifier =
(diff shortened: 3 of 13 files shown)


plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
```diff
@@ -508,6 +522,79 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
...
+          popupMenu.setTargetComponent(comp)
+          popupMenu.component.show(comp, 0, comp.height)
+        }
+      }
+        .align(AlignY.CENTER)
+
+      compose {
+        var showJewelMenu by remember { mutableStateOf(false) }
+
+        Box(modifier = Modifier.height(150.dp)) {
```
DanielSouzaBertoldi, line 742: Had to set a fixed height so that our Popup can be rendered in its entirety instead of smushed

daaria-s:
oh I remember this bug, nice research :grin:

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
```diff
@@ -520,3 +521,13 @@ public fun retrieveEditorColorScheme(): EditorColorsScheme {
     val manager = EditorColorsManager.getInstance() as EditorColorsManagerImpl
     return manager.schemeManager.activeScheme ?: DefaultColorSchemesManager.getInstance().firstScheme
 }
+
+/**
+ * Converts a raw Swing dimension (which might already be scaled by JBUI) into a Compose Dp.
+ *
+ * Use this whenever you fetch an Int value from a scaled property in JBUI. If you don't, you'll get "Double Scaling":
```
daaria-s: just curious, is it true that there might be more places where we need to use it but we don't know about them?

daaria-s (approved):

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
DanielSouzaBertoldi: Not that I was able to find yet 😮 

We do fetch the `rowHeight()` value in our `readFromLaF()` function over in `BridgeGlobalMetrics` but we are already unscaling it there

DanielSouzaBertoldi:
Hmmm come to think of it, maybe it's interesting to create a Detekt rule for this? I can't think of any scenarios where using a scaled `dp` value would make sense 🤔

rock3r:
New detekt rules get my blanket "hell yea"

rock3r:
@DanielSouzaBertoldi can we haz that detekt rule? :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -872,52 +852,24 @@ internal fun MenuItemBase(
...
-                horizontalArrangement = Arrangement.spacedBy(4.dp),
-                verticalAlignment = Alignment.CenterVertically,
-            ) {
-                if (canShowIcon) {
-                    val iconModifier = Modifier.size(style.metrics.itemMetrics.iconSize)
-                    if (iconKey != null) {
-                        Icon(
-                            key = iconKey,
-                            contentDescription = null,
-                            modifier = iconModifier.thenIf(!enabled) { disabledAppearance() },
```
rock3r, line 903: I think we've lost the disabledAppearance here, unless I'm missing something

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
rock3r: Hmm we do unscaling for `Insets` in this file *see `Insets.toPaddingValues()`), so if the `Int`s we get are indeed scaled, we need to check. I don't know if it's a potential issue, but it seems like `retrieveIntAsDp*()`s do need a closer look to validate whether they're fetching the right values or not.

Otherwise you're saying "it's fine to use `retrieveIntAsDp*`s everywhere except in this one case where it would do a double scale", which seems a bit odd in principle. Effectively, `JBUI.CurrentTheme.List.rowHeight()` does the same thing we do, calling `JBUI.getInt()` which calls `UIManager.get()` — `retrieveIntAsDp*()` also does the same, then calling `.dp` on the value (which is where the double scaling may happen).

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -1201,3 +1169,32 @@ public value class MenuItemState(public val state: ULong) : SelectableComponentS
         }
     }
 }
+
+@Suppress("DEPRECATION") // Not really deprecated, MenuItemState will be made internal
+@Composable
+internal fun rememberMenuItemState(
```
rock3r, line 1185: Wouldn't this subtly change `MenuSubmenuItem` behavior? Before the refactor it reset `selected = false` on `enabled` changes, whereas the shared helper now preserves `selected`. So if a submenu is currently open / hover-selected and then becomes disabled, I think it would stay selected (and keep the submenu open) until something else dismisses it. Is that intentional?

rock3r (changes requested):

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/BridgeUtils.kt
DanielSouzaBertoldi: Alright alright, I had to go down a rabbit hole to find out what the heck is going on.

Yes, it's true that the only value being scaled in the `rowHeight()` function is the fallback value:

```java
static int rowHeight() {
  int defaultHeight = JBUIScale.scale(24); // <---- HERE!!!!
  int result = getInt("List.rowHeight", defaultHeight);
  return result <= 0 ? defaultHeight : result;
}
```

`UIManager` doesn't care about scaling at all. It just tries to fetch a value from a `theme.json` file given a key.

So who the hell is scaling the `"List.rowHeight"` value?

At first I thought maybe the theme was declaring a scaled value by default, but that's not the case for `expUI_light*`/`expUI_dark*`:

```json
"List": {
  "rowHeight": 24,
  ```

After jumping from class to class and trying to make sense of this, I finally found the culprit. In `LafManagerImpl` there's this function:

```kotlin
private fun patchRowHeight(defaults: UIDefaults, key: String, prevScale: Float) {
  val value = defaults.get(key)
  val rowHeight = if (value is Int) value else 0
  if (rowHeight <= 0) {
    LOG.warn("$key = $value in ${UIManager.getLookAndFeel().name}; it may lead to performance degradation")
  }
  val custom = intSystemPropertyValue("ide.override.$key", -1)
  defaults.put(key, if (custom >= 0) scale(custom) else if (rowHeight <= 0) 0 else scale((rowHeight / prevScale).toInt()))
}
```

Which, wouldn't you know it, scales the `rowHeight` property. This function is called by `patchHiDPI(defaults: UIDefaults)`, a function that only "fixes" the predefined row height of a very small set of properties:

```kotlin
val prevRowHeightScale = prevScale
patchRowHeight(defaults, "List.rowHeight", prevRowHeightScale)
patchRowHeight(defaults, "Table.rowHeight", prevRowHeightScale)
patchRowHeight(defaults, JBUI.CurrentTheme.Tree.rowHeightKey(), prevRowHeightScale)
patchRowHeight(defaults, JBUI.CurrentTheme.VersionControl.Log.rowHeightKey(), prevRowHeightScale)
```

So, to answer your point @rock3r , `retrieveIntAsDp*()` functions are correct and they do return the correct values. The only "gotcha" here is these 4 properties where the code overrides their value by scaling them.

A bit further down the function body, the code also scales these other properties:

```kotlin
val intKeys = hashSetOf("Tree.leftChildIndent", "Tree.rightChildIndent", "SettingsTree.rowHeight")
val dimensionKeys = hashSetOf("Slider.horizontalSize",
                              "Slider.verticalSize",
                              "Slider.minimumHorizontalSize",
                              "Slider.minimumVerticalSize")
```

Comparing with our code, we only currently just fetch the values from `Tree.rowHeight`/`Tree.leftChildIndent`/`Tree.rightChildIndent` out of all the properties in the function.

So, those are also fixed! Check the screenshots below 😎 

| Before | After |
|--------|--------|
| image | image |

Final takeaway: there's no way around this. For each component we have to check if they do have some sneaky code that scales them.

Updated the code to include a `retrieveUnscaledIntAsDpOrUnspecified`/`retrieveUnscaledIntAsNonNegativeDpOrUnspecified` function and a thorough explanation of this whole situation!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: In practice, external state changes that disable an item also tend to invalidate the items list (lambda recapture), so the submenu closes anyway. Buuut the fix is really simple so I'd rather we cover all edge cases. Fixed ✅

DanielSouzaBertoldi:
@rock3r on the topic of the detekt rule, given everything that I learned about this problem in particular... I'm afraid it doesn't make sense :(

There are only a handful of properties where they value get overridden. We'd have to keep checking the code to see if they added new properties to `patchHiDPI` or not and update the rule accordingly.

Also, there's two ways AFAIK that we can use to fetch the value of a property: either by using a corresponding function in `JBUI`, such as `JBUI.CurrentTheme.List.rowHeight()` or by using the property key, `UIManager.get("List.rowHeight")`

The first one is alright to check for, we'd need just to check for specific chain calls in the AST, but the second case requires string literal matching inside a call chain which can be very brittle...

Instead, I propose what I already pushed in the latest changes: have the KDoc explicitly list all cases where we do need unscaled values.

What do you think?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -1004,32 +933,22 @@ internal fun MenuSubmenuItem(
...
-                Box(Modifier.weight(1f)) { content(itemState) }
-
+        MenuItemLayout(
+            itemState = itemState,
+            style = style,
+            iconKey = iconKey,
+            canShowIcon = showIcon,
+            trailingContent = {
+                @Suppress("DEPRECATION") // Not really deprecated, MenuItemColors will be made internal
                 Icon(
```
rock3r, line 947: Looks like the submenu chevron lost the stateful painter hint in the refactor. The leading icon path still handles disabled/tinted state, but without `Stateful(itemState)` a custom `MenuIcons.submenuChevron` won’t resolve `Disabled`/`Hovered`/`Focused`/`Pressed` SVG variants anymore.

Can we keep passing the state hint here?

rock3r (changes requested):

rock3r (approved):
Thanks, looks good!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -1048,6 +969,61 @@ internal fun MenuSubmenuItem(
...
+            verticalAlignment = Alignment.CenterVertically,
+        ) {
+            if (canShowIcon) {
+                val iconModifier = Modifier.size(itemMetrics.iconSize)
+                if (iconKey != null) {
+                    Icon(
+                        key = iconKey,
+                        contentDescription = null,
+                        modifier = iconModifier.thenIf(!enabled) { disabledAppearance() },
+                        tint = itemColors.iconTintFor(itemState).value,
```
nebojsa-vuksic: Really nice cleanup pulling the shared layout into `MenuItemLayout`! One behavior change I wanted to flag here: this now applies `tint` to the leading icon, whereas before neither `MenuItemBase` nor `MenuSubmenuItem` tinted the leading icon (only the submenu chevron used `iconTint`).

In the IDE bridge theme this is a no op, since `iconTint` there is `Color.Unspecified` and `Icon` treats that as "no tint". In the standalone IntUi default theme though, `iconTint` is `gray(7)` (see `IntUiMenuStyling.kt`), so leading icons in standalone menu items would now be flat tinted gray, which would recolor any colored action icons.

Would it make sense to keep the leading icon untinted and apply `iconTint` only to the chevron (via `trailingContent`), to preserve the previous appearance? If the tinting is actually intended, please ignore me.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeLazyTree.kt
```diff
@@ -34,8 +34,8 @@ internal fun readLazyTreeStyle(): LazyTreeStyle {
             backgroundSelectedActive = selectedElementBackground,
         )
 
-    val leftIndent = retrieveIntAsNonNegativeDpOrUnspecified("Tree.leftChildIndent").takeOrElse { 7.dp }
-    val rightIndent = retrieveIntAsNonNegativeDpOrUnspecified("Tree.rightChildIndent").takeOrElse { 11.dp }
+    val leftIndent = retrieveUnscaledIntAsNonNegativeDpOrUnspecified("Tree.leftChildIndent").takeOrElse { 7.dp }
```
nebojsa-vuksic, line 37: Just a heads up that this extends the fix to the lazy tree as well. I think it is correct: `patchHiDPI` in `LafManagerImpl` rescales `Tree.leftChildIndent`, `Tree.rightChildIndent`, and `Tree.rowHeight`, so these were double scaled the same way the menu height was, and the unscaled value produces the same result at scale 1.0. Since it is slightly outside the menu scope of the PR and there is no tree specific evidence in the description, it might be worth a quick visual check of trees in Presentation Mode to be safe.

plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
```diff
@@ -39,6 +46,7 @@ import com.intellij.openapi.ui.popup.LightweightWindowEvent
 import com.intellij.openapi.util.IconLoader
 import com.intellij.ui.EditorNotificationPanel
 import com.intellij.ui.InlineBanner
+import com.intellij.openapi.util.NlsActions
```
nebojsa-vuksic: Tiny nit: `NlsActions` looks like it ends up unused (the actions are built from plain `String` text), and it is also slightly out of order among the `com.intellij.*` imports. Might be worth removing to avoid an inspection warning.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -1201,3 +1177,32 @@ public value class MenuItemState(public val state: ULong) : SelectableComponentS
         }
     }
 }
+
+@Suppress("DEPRECATION") // Not really deprecated, MenuItemState will be made internal
+@Composable
+internal fun rememberMenuItemState(
```
nebojsa-vuksic, line 1185: Nice extraction. Just a testing note: the submenu `selected` and `enabled` handling was restructured a little here (the old code reset `selected` to false on any `enabled` change and synced `selected` from the param in a separate `remember`, whereas this routes through `rememberMenuItemState` plus the explicit `remember(enabled) { if (!enabled) ... }`). It looks equivalent for the common case, but the existing menu tests (`MenuContentShortcutRegistrationTest`) only cover shortcut registration, so this state path is not exercised. A quick manual check of submenu open and close, especially when toggling enabled while a submenu is open, would be reassuring.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: Oh, great catch. I'm already applying the tint for submenu items in `MenuSubmenuItem`, so I can just remove this call here. Thanks :)

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/IntUiBridgeLazyTree.kt
DanielSouzaBertoldi: Weird, this doesn't seem to affect Trees at all 🤔 

| Default View | Presentation Mode |
|---|---|
| image | image |

plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
DanielSouzaBertoldi: It's weird that the IDE didn't flag this unused import. Oh well, thanks for noticing!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: I just tested this by making the "More options" become unavailable after 5 seconds:

https://github.com/user-attachments/assets/0f81f966-174d-45f9-a408-a38a65853fe7

So everything's fine :)

I also took this opportunity to create a `MenuSubmenuItemTest`. It's outside the scope for this PR but heck, since I changed the inner workings why not

nebojsa-vuksic (approved):

rock3r:
@DanielSouzaBertoldi can you rebase/fix up the CI so we can ship?

nebojsa-vuksic (approved):

DanielSouzaBertoldi:
Ready to merge

DanielSouzaBertoldi:
Ready to merge

PR #3421, opened 2026-02-12
[JEWEL-1239] Render failed to load images as hyperlinks

AlexVanGogen:
> [!WARNING]
>
> This PR is built on top of [image scaling PR](https://github.com/JetBrains/intellij-community/pull/3387). It's very hard to decouple them as they both heavily change the same source code. Please disregard the earlier commit message, only the last one is relevant to this change. I'll update the branch once the above PR gets merged. 

Images that fail to load are now rendered as clickable hyperlinks instead of showing raw Markdown syntax or blank placeholders.

## Changes

  Failed Image Handling:
  - Images that fail to load are replaced with clickable links showing the alt text (or URL if no alt text)
  - A broken image indicator (🖼) is prepended to the link text

  Loading States:
  - Added `ImageRenderResult` sealed class with `Loading`, `Success`, and `Failed` states
  - Loading indicator is shown while images are being fetched
  - Previously failed images skip the loading indicator on retry to prevent flickering

  API Changes:
  - New `ImageRendererExtension.renderImage` method returning `ImageRenderResult` for explicit state handling
  - `Deprecated renderImageContent` method (preserved for backward compatibility)

#### Switch between successful and failed loading (valid/invalid image source)

https://github.com/user-attachments/assets/59a0c524-beb9-46aa-8ac7-0055b230497d

#### One broken URL doesn't affect other images in a block

https://github.com/user-attachments/assets/ecfcb0ae-64f5-4c31-bb08-5a2d9b3558ac

#### Changing alt text affects failed image rendering and doesn't cause flickering (loading indicator <-> broken link) (don't ask why badges from Jewel README didn't render :))

https://github.com/user-attachments/assets/6a7c5759-32df-4a6f-abcb-e325644e893f

### ⚠️ Important Changes
 - Images that fail to load are now rendered as clickable links instead of showing raw Markdown syntax

### New features
 - **JEWEL-1239** Render failed to load images as hyperlinks ([#3421](https://github.com/JetBrains/intellij-community/pull/XXXX))

### Deprecated API
- Deprecated `ImageRendererExtension.renderImageContent`; use `renderImage` instead, which provides explicit loading/success/failed states

### DIFF
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
@@ -18,10 +18,12 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.ProvidableCompositionLocal
 import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.movableContentOf
 import androidx.compose.runtime.mutableStateMapOf
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.mutableStateSetOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.staticCompositionLocalOf
@@ -46,9 +48,11 @@ import androidx.compose.ui.layout.MeasureScope
 import androidx.compose.ui.layout.onFirstVisible
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.LinkAnnotation
 import androidx.compose.ui.text.Placeholder
 import androidx.compose.ui.text.TextLayoutResult
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.rememberTextMeasurer
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
@@ -80,6 +84,7 @@ import org.jetbrains.jewel.markdown.MarkdownBlock.ListItem
 import org.jetbrains.jewel.markdown.MarkdownBlock.Paragraph
 import org.jetbrains.jewel.markdown.MarkdownBlock.ThematicBreak
 import org.jetbrains.jewel.markdown.WithInlineMarkdown
+import org.jetbrains.jewel.markdown.extensions.ImageRenderResult
 import org.jetbrains.jewel.markdown.extensions.MarkdownRendererExtension
 import org.jetbrains.jewel.markdown.rendering.MarkdownStyling.List.Ordered.NumberFormatStyles
 import org.jetbrains.jewel.markdown.rendering.MarkdownStyling.List.Unordered.BulletCharStyles
@@ -243,16 +248,14 @@ public open class DefaultMarkdownBlockRenderer(
         softWrap: Boolean = true,
         maxLines: Int = Int.MAX_VALUE,
     ) {
-        val originalImages = renderedImages(block)
-
-        val renderedContent = rememberRenderedContent(block, inlinesStyling, enabled, onUrlClick)
+        val (loadedImages, renderedContent) = rememberRenderedContent(block, inlinesStyling, enabled, onUrlClick)
         val textColor = inlinesStyling.textStyle.color.takeOrElse { LocalContentColor.current }
         val mergedStyle = inlinesStyling.textStyle.merge(TextStyle(color = textColor))
         val density = LocalDensity.current
 
         TextWithScalableInlineContent(
             text = renderedContent,
-            inlineContent = originalImages,
+            inlineContent = loadedImages,
             density = density,
             modifier = modifier,
             overflow = overflow,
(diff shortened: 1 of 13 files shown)


platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -71,34 +92,62 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
+            return ImageRenderResult.Failed
+        }
+
+        val result = imageResult
+        if (result == null) {
+            // Anti-flickering when typing: no loading indication for an image previously failed to load
+            return if (previouslyFailed) {
+                ImageRenderResult.Failed
+            } else {
+                ImageRenderResult.Loading(createLoadingIndicator())
```
DanielSouzaBertoldi, line 111: Question: shouldn't we use some sort of debounce here? I'm thinking in situations where users want to type a new valid link.

I mean, change `blabla.com/my-image-1` to `blabla.com/my-image-1000`, where both links are valid.

With this code, if users type the first `0` after `1`, then the request will load, if it's fast enough it'll fail and the hyperlink will show up instead, right? Maybe just chill for a bit before trying again.

I think that creating a debounce and using it as the key for the state variables would fix this

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -751,27 +770,61 @@ public open class DefaultMarkdownBlockRenderer(
...
+                    failedSources.remove(imageSource)
+                }
 
-        for (image in getImages(blockInlineContent)) {
-            val renderedImage = imagesRenderer?.renderImageContent(image)
-            if (renderedImage == null) {
-                map.remove(image.source)
-            } else {
-                map[image.source] = renderedImage
+                is ImageRenderResult.Loading -> {
```
rock3r, line 809: Shouldn't `Loading` clear a previous failed state here?

Right now we only remove the source from `failedSources` on `Success`, so after an image has failed once we keep rebuilding it as the fallback link throughout later `Loading` states. That means a renderer using the new `ImageRenderResult` contract can't really express `Failed -> Loading -> Success`: the transient loading UI stays stuck in the failed visual state until the success arrives.

This will still eventually recover on success, but the intermediate state seems wrong.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -884,6 +1063,60 @@ private fun getImages(input: WithInlineMarkdown): List<InlineMarkdown.Image> = b
...
+            val imageSource = image?.source.orEmpty()
+            val altText = image?.alt?.ifEmpty { null } ?: imageSource
+
+            val index =
+                if (enabled) {
+                    val link =
+                        LinkAnnotation.Clickable(
+                            tag = imageSource,
+                            linkInteractionListener = { onUrlClick?.invoke(imageSource) },
+                            styles = styling.textLinkStyles,
```
rock3r, line 1103: Shouldn't we merge the link styles with the surrounding text style here, like the normal inline-link path does?

`DefaultInlineMarkdownRenderer.renderLink()` uses `currentTextStyle.smartMerge(styling.textLinkStyles, enabled = true)`, but this rebuild path applies `styling.textLinkStyles` / `styling.linkDisabled` directly. So if a failed image sits in a heading or any other styled inline context, the fallback link can lose the expected typography and no longer match a normal link in the same place.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -884,6 +1063,60 @@ private fun getImages(input: WithInlineMarkdown): List<InlineMarkdown.Image> = b
...
+
+private fun rebuildWithFailedImagesAsLinks(
+    old: AnnotatedString,
+    failedImages: Set<String>,
+    block: WithInlineMarkdown,
+    styling: InlinesStyling,
+    enabled: Boolean,
+    onUrlClick: ((String) -> Unit)?,
+): AnnotatedString {
+    val imagesBySource = getImages(block).associateBy { it.source }
```
rock3r, line 1079: I don't think we can key this by `source` anymore.

If the same image URL appears multiple times in a block with different alt texts, `associateBy` keeps only one `InlineMarkdown.Image`, so every failed occurrence ends up using that one alt/fallback text. Coalescing the loading for identical sources seems fine, but the fallback text is occurrence-specific and shouldn't get collapsed with it.

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
rock3r: @AlexVanGogen I think this is still not fully addressed. There is also the "hidden loading after a failure" issue I commented on, compounding this

rock3r (changes requested):

rock3r:
Nice direction overall — the fallback-link behavior and the explicit image render states make sense, and the added coverage is helpful.

The main things I found are:
- `Failed -> Loading -> Success` is not really represented correctly right now: after an image has failed once, the source stays in `failedSources` until `Success`, so later `Loading` states still render as the fallback link instead of showing the transient loading UI.
- Failed-image links don't inherit the surrounding inline text style the same way normal links do, so they can lose the expected typography in headings or other styled inline contexts.
- Repeated images with the same `src` conflate their fallback text: keying by source and then using `associateBy` means multiple failed occurrences can end up sharing the same alt / fallback text even though that text is occurrence-specific.

Nit: the release notes in the PR description still link to `pull/XXXX`, and I think this user-visible change reads more like a bug fix than a new feature.

DanielSouzaBertoldi:
This one is next on the line, let's goooo

nebojsa-vuksic (approved):

DanielSouzaBertoldi:
The PR #3579 continues Sasha's legacy, addressing all comments so far :)

PR #3431, opened 2026-02-19
[JEWEL-1257] Add support for new islands palette

daaria-s:
This PR adds support for the Islands color palette format in `ThemeColorPalette`.

Islands palettes use indexed steps starting from 10 and incrementing by 10 (10, 20, 30, …) instead of consecutive numbering (1, 2, 3, …).

- The `ThemeColorPalette` constructor now accepts a new `isIslands` parameter to distinguish between Islands and non-Islands palette types.
- All color accessor methods (`grayOrNull()`, `blueOrNull()`, etc.) now support both indexing schemes depending on the palette type.
- The Bridge theme reader detects Islands themes via `IslandsState.isEnabled()` and reads colors using the appropriate key format.

### ⚠️ Important Changes
 * `ThemeColorPalette` now includes a new `isIslands` parameter to distinguish between Islands and non-Islands theme palettes.
 * Updated all color accessor methods in `ThemeColorPalette` to support both indexing schemes.
 * Updated `BridgeThemeColorPalette` to parse Islands palette keys when applicable.
 * Added tests for `BridgeThemeColorPalette` covering both Islands and non-Islands palette formats.
 * `IntUIDarkTheme` and `IntUILightTheme` now include new `isIslands` argument, their generator was updated for that.

### Deprecated API
* `ThemeColorPalette` without `isIslands` parameter is deprecated, now `isIslands` should be specified for the palette

### New features
 * Added support for the Islands color palette format in `ThemeColorPalette`.

---

> [!NOTE]
> **Medium Risk**
> Updates the public `ThemeColorPalette` API and theme parsing to handle a second palette key/indexing scheme, which could affect any consumers relying on palette lookups or indexing. Risk is mitigated by keeping a hidden deprecated constructor and adding targeted bridge tests for both formats.
> 
> **Overview**
> **Adds support for the new Islands palette format** across Jewel theme infrastructure.
> 
> `ThemeColorPalette` now carries an `isIslands` flag and updates `*OrNull` accessors plus `lookup()` to handle both classic `gray1`/`ColorPalette.Gray1` and Islands-style `gray-10`/`ColorPalette.gray-10` indexing; a hidden deprecated constructor preserves source/binary compatibility.
> 
> The IDE LaF bridge now detects Islands via `IslandsState.isEnabled()` and reads palette keys/indices accordingly, with a new comprehensive `BridgeThemeColorPaletteTest` covering both formats. Generated Int UI themes and the theme descriptor generator were updated to pass `isIslands = false`, the Jewel Gradle setup is bumped to IDEA `261-EAP-SNAPSHOT` and uses `useInstaller = false`, and Bazel wintools test targets were renamed to `community-wintools-*`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 64f693dfe1ac3773cd889b036993175148bb2a68. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
```diff
@@ -39,7 +42,33 @@ public class ThemeColorPalette(
     public val purple: List<Color>,
     public val teal: List<Color>,
     public val rawMap: Map<String, Color>,
+    private val isIslands: Boolean,
 ) {
+    @Suppress("DEPRECATION")
+    @Deprecated("Use the constructor with isIslands parameter", level = DeprecationLevel.HIDDEN)
```
DanielSouzaBertoldi, line 48: Please add the "Deprecated API" section in your PR's release notes about this 😄

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
```diff
@@ -344,6 +391,7 @@ public class ThemeColorPalette(
 
     override fun toString(): String {
         return "ThemeColorPalette(" +
+            "isIslands=$isIslands, " +
```
DanielSouzaBertoldi: This is a very very silly nit, so feel very free to just ignore it, but I'd respect the order of parameters in the constructor when converting it to a string. In other words, the `isIslands` should be after `rawMap`, mirroring its place in the constructor.

Also, not sure if you manually made these changes but you can do `cmd+N` and a popup will appear asking if you want the IDE to generate `equals()`, `hashCode()` and `toString()` 😄

platform/jewel/gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ composeDesktop = "1.10.0"
 detekt = "1.23.8"
 dokka = "2.0.0"
 filepicker = "3.1.0"
-idea = "2025.3.2"
+idea = "261.21849.20"
```
DanielSouzaBertoldi: Why do we need to update this dependency? I thought the islands theme was released in 2025.2.3 😯

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,336 @@
...
+    fun `should save windows popup border color to ThemeColorPalette`() {
+        val windowsPopupBorder = java.awt.Color(100, 100, 100)
+
+        withUiDefaults(mapOf("ColorPalette.windowsPopupBorder" to windowsPopupBorder)) {
+            val palette = ThemeColorPalette.readFromLaF()
+            assertEquals(windowsPopupBorder.toComposeColor(), palette.windowsPopupBorder)
+        }
+    }
+
+    private fun createNonIslandThemeColors(colorName: String, count: Int): Map<String, java.awt.Color> {
```
DanielSouzaBertoldi: Please import `java.awt.Color` 😄 

```suggestion
    private fun createNonIslandThemeColors(colorName: String, count: Int): Map<String, Color> {
```

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,336 @@
...
+        val expectedColors = mutableMapOf<String, java.awt.Color>()
```
DanielSouzaBertoldi: Same thing here

```suggestion
        val expectedColors = mutableMapOf<String, Color>()
```

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,336 @@
...
+        for (i in 1..count) {
+            val color =
+                java.awt.Color(
```
DanielSouzaBertoldi: Also here

```suggestion
                Color(
```

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,336 @@
...
+                    (i * 10) % 256,
+                )
+            val key = "ColorPalette.$colorName$i"
+            expectedColors[key] = color
+        }
+
+        return expectedColors
+    }
+
+    private fun createIslandThemeColors(colorName: String, count: Int): Map<String, java.awt.Color> {
```
DanielSouzaBertoldi:
```suggestion
    private fun createIslandThemeColors(colorName: String, count: Int): Map<String, Color> {
```

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,336 @@
...
+        val expectedColors = mutableMapOf<String, java.awt.Color>()
+        for (i in 10..count * 10 step 10) {
+            val color =
+                java.awt.Color(
```
DanielSouzaBertoldi:
```suggestion
                Color(
```

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,336 @@
...
+                )
+            val key = "ColorPalette.$colorName$i"
+            expectedColors[key] = color
+        }
+
+        return expectedColors
+    }
+
+    private fun createIslandThemeColors(colorName: String, count: Int): Map<String, java.awt.Color> {
+        val expectedColors = mutableMapOf<String, java.awt.Color>()
```
DanielSouzaBertoldi:
```suggestion
        val expectedColors = mutableMapOf<String, Color>()
```

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,336 @@
...
+                    val actual = palette.tealOrNull(i)
+                    assertEquals("Color $colorName$i is not correct in the palette", expected, actual)
+                }
+            }
+        }
+    }
+
+    @Test
+    fun `should save windows popup border color to ThemeColorPalette`() {
+        val windowsPopupBorder = java.awt.Color(100, 100, 100)
```
DanielSouzaBertoldi: Please import `java.awt.Color` 😄 

```suggestion
        val windowsPopupBorder = Color(100, 100, 100)
```

DanielSouzaBertoldi (changes requested):

platform/jewel/gradle/libs.versions.toml
daaria-s: because I need to use new IslandsState.isEnabled/setEnabled API, it was added recently and is in 261 only

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPalette.kt
```diff
@@ -45,28 +48,36 @@ public fun ThemeColorPalette.Companion.readFromLaF(): ThemeColorPalette {
...
         rawMap = rawMap,
+        isIslands = isIslands,
     )
 }
 
-private fun readPaletteColors(colorName: String): Map<String, Color> {
-    val defaults = uiDefaults
-    val allKeys = defaults.keys
-    val colorNameKeyPrefix = "ColorPalette.$colorName"
+@Suppress("UnreachableCode")
```
daaria-s, line 55: There was a false positive error in detekt

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/clipboard/JewelBridgeClipboard.kt
```diff
@@ -39,6 +39,7 @@ internal class JewelBridgeClipboard : Clipboard {
     // itself. Hence, we build a facade Clipboard that delegates to the IntelliJ CopyPasteManager.
     override val nativeClipboard: NativeClipboard by lazy { JewelAwtClipboardBridge(copyPasteManager) }
 
+    @Suppress("UnreachableCode")
```
daaria-s, line 42: also false positive in :detektMain

platform/jewel/foundation/api-dump.txt
```diff
@@ -700,7 +700,8 @@ f:org.jetbrains.jewel.foundation.theme.JewelThemeKt
 f:org.jetbrains.jewel.foundation.theme.ThemeColorPalette
 - sf:$stable:I
 - sf:Companion:org.jetbrains.jewel.foundation.theme.ThemeColorPalette$Companion
-- <init>(java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.Map):V
+- b:<init>(java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.Map):V
+- <init>(java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.List,java.util.Map,Z):V
```
daaria-s, line 704: jewel checks said it is a breaking API change, is it really it?

DanielSouzaBertoldi: Not really :)

The `b` prefix means this constructor is now marked as `ACC_SYNTHETIC` in the Java bytecode, so this constructor is now only available "under the hood". Developers using either Kotlin or Java will not be able to use it, but the compiler can still find it declared in the bytecode.

Now, the new constructor has all the parameters from the synthetic constructor + a `Z`, which is your new `isIslands: Boolean`.

Though, I'm not sure if you really have to add `@Suppress("DEPRECATION")` to your `@Deprecated` constructor in `ThemeColorPalette.kt`

platform/jewel/ide-laf-bridge/build.gradle.kts
```diff
@@ -35,7 +35,7 @@ repositories {
 dependencies {
     api(projects.ui) { exclude(group = "org.jetbrains.kotlinx") }
     intellijPlatform {
-        intellijIdea(libs.versions.idea)
+        intellijIdea(libs.versions.idea) { useInstaller = false }
```
DanielSouzaBertoldi, line 38: What does this `useInstaller` do and why do we need to set it to false now? 😯

DanielSouzaBertoldi (approved):
Great work!

platform/jewel/ide-laf-bridge/build.gradle.kts
nebojsa-vuksic: It's required in order to use a snapshot version of platform. I don't know implementation details behind it though. It's probably worth asking @hsz

hsz: By default, IDE releases available in JetBrains CDN are resolved (user-facing releases, installers). If you want to access EAP releases, you have to opt out of installer artifacts. See https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-dependencies-extension.html#target-versions-multi-os-archives

DanielSouzaBertoldi: Thanks for the answer guys!

rock3r:
@daaria-s can you rebase/solve conflicts?

platform/jewel/gradle/libs.versions.toml
```diff
@@ -6,8 +6,8 @@ composeDesktop = "1.10.0"
 detekt = "1.23.8"
 dokka = "2.0.0"
 filepicker = "3.1.0"
-idea = "2025.3.2"
-intelliJPlatformBuild = "253.30387.90"
+idea = "261-EAP-SNAPSHOT"
```
rock3r, line 9: Since we're at 2026.1.2 stable (`261.24374.151`), we should pin to a stable/named version instead of `EAP-SNAPSHOT`.

```suggestion
idea = "261.24374.151"
```

Also, this platform version bump should be explicitly documented in the PR description and Release Notes, as it will drop Jewel bridge support for IDEs older than 2026.1.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPalette.kt
```diff
@@ -45,28 +48,36 @@ public fun ThemeColorPalette.Companion.readFromLaF(): ThemeColorPalette {
...
+private fun readPaletteColors(colorName: String, isIslands: Boolean): Map<String, Color> {
+    val defaults: UIDefaults? = uiDefaults
```
rock3r, line 57: Can we make this non-nullable now? It looks like the explicit `UIDefaults?` cast and the Elvis fallback `?: return TreeMap()` below are workarounds for a new `@NotNull` annotation from the platform bump. Since `UIManager.getDefaults()` shouldn't be null, we can probably drop the safe calls and the redundant `?: return TreeMap()`.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
```diff
@@ -288,6 +332,7 @@ public class ThemeColorPalette(
      * @return The [Color] associated with the key, or `null` if the key is not found.
      */
     public fun lookup(colorKey: String): Color? {
```
rock3r, line 342: Nit: the KDoc above this method explicitly mentions `The key can be in the format "colorNameN" (e.g., "gray1", "blue12")`. Could we update it to also mention the new Islands format `"colorName-N"` (e.g., `"gray-10"`)?

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
```diff
@@ -62,13 +91,14 @@ public class ThemeColorPalette(
     /**
      * Retrieves a gray color from the palette by its index, or `null` if the index is out of bounds.
      *
-     * Palette indices start at 1; how many entries exist for a color depends on the Look and Feel. Some LaFs may only
-     * have a partial palette, or none at all.
+     * Palette indices start at 1 (or 10 for Islands themes); how many entries exist for a color depends on the Look and
```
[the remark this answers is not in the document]
daaria-s: okay that makes sense

daaria-s:
> @daaria-s can you rebase/solve conflicts?

done

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/theme/BridgeThemeColorPaletteTest.kt
```diff
@@ -0,0 +1,344 @@
...
+        }
+    }
+
+    private fun inIslandsTheme(block: () -> Unit) {
+        // set islands theme
+        IslandsState.setEnabled(true, false)
+        try {
+            block()
+        } finally {
+            // unset islands theme
```
rock3r, line 340: Can we restore the previous global state here instead of always resetting to false/removing the keys? This is a plain JUnit test class, not an IJ fixture test, and Gradle can reuse the same test worker JVM across classes.

So if the current `UIDefaults` already had one of these palette keys, `withUiDefaults()` overwrites and then deletes it; similarly, if Islands was enabled before the test, `inIslandsTheme()` disables it afterwards. Snapshotting the previous `UIDefaults` values and the previous `IslandsState.isEnabled()` / `isCustomEnabled()` values would keep the test isolated.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
```diff
@@ -39,7 +42,33 @@ public class ThemeColorPalette(
     public val purple: List<Color>,
     public val teal: List<Color>,
     public val rawMap: Map<String, Color>,
+    private val isIslands: Boolean,
```
rock3r, line 45: Non-blocking follow-up: with this flag private, theme-agnostic callers can’t tell whether the typed accessors expect classic indices (`blueOrNull(4)`) or Islands indices (`blueOrNull(40)`). That mirrors the raw Swing key problem, but Jewel now has enough typed structure that we could probably offer something nicer.

I filed [JEWEL-1331](https://youtrack.jetbrains.com/issue/JEWEL-1331) to track additive API options, e.g., exposing the palette format or adding format-independent ordinal accessors. No action needed here.

rock3r (changes requested):
Turns out I had a review in drafts since March 18th. Yay me. Anyway, cleaned up and updated, here it is :D

rock3r:
Superseded by #3523, which carries this change forward from a rebased branch with the outstanding review feedback addressed.

PR #3442, opened 2026-02-27
[JEWEL-1267] Support sized images

AlexVanGogen:
## Summary

Add support for parsing and rendering sized images in Jewel Markdown when image dimensions are provided via HTML `width` and `height` attributes or native Markdown image attribute blocks (`{width=100 height=50%}`), including both absolute pixel values and percentage-based dimensions with proportional scaling when only one dimension is specified.

Fixes JEWEL-1267

## Changes

 * New `DimensionSize` sealed interface with `Pixels` and `Percent` value classes to represent image dimension specifications
 * Extended `InlineMarkdown.Image` with optional `width` and `height` properties
 * Added parsing of Markdown image attribute blocks: `alt{width=100 height=50%}` — supports unquoted, single-quoted, and double-quoted attribute values, with `px` and `%` suffixes
 * Added `parseHtmlSizeValue()` for parsing HTML size attributes (e.g., `"100"`, `"100px"`, `"50%"`) into `DimensionSize` values, with negative percentage values treated as invalid
 * Refactored `Coil3ImageRendererExtensionImpl` to compute placeholder sizes based on specified dimensions — when both are given they're used directly, when only one is specified the other scales proportionally, and when none are specified the original image size is used
 * Updated `api-dump-experimental.txt` with new `DimensionSize` API entries
 * Added comprehensive unit tests:
   * 10 tests for Markdown image attribute block parsing
   * 12 tests for Compose image rendering with dimensions
   * Extended HTML converter tests for `` tag parsing

#### Percentage and pixel dimensions (50%, 540px/304px)
one

#### Mixed dimension combinations (pixel+percentage, percentage+pixel)
one-27 at 12 19 37

#### Original images

Screenshot 2026-03-10 at 20 02 46

#### Using sized width/height

Screenshot 2026-03-10 at 20 03 16

### New features
 * Added support for sized images in Markdown rendering — HTML `` tags with `width` and `height` attributes and native Markdown image attribute blocks (`{width=100 height=50%}`) are now parsed and rendered at the specified dimensions, with proportional scaling when only one dimension is provided (https://github.com/JetBrains/intellij-community/pull/3442)

### DIFF
--- a/platform/jewel/markdown/core/api-dump-experimental.txt
+++ b/platform/jewel/markdown/core/api-dump-experimental.txt
@@ -1,3 +1,28 @@
+*:org.jetbrains.jewel.markdown.DimensionSize
+*f:org.jetbrains.jewel.markdown.DimensionSize$Percent
+- org.jetbrains.jewel.markdown.DimensionSize
+- bsf:box-impl(I):org.jetbrains.jewel.markdown.DimensionSize$Percent
+- s:constructor-impl(I):I
+- equals(java.lang.Object):Z
+- s:equals-impl(I,java.lang.Object):Z
+- sf:equals-impl0(I,I):Z
+- f:getValue():I
+- hashCode():I
+- s:hashCode-impl(I):I
+- s:toString-impl(I):java.lang.String
+- bf:unbox-impl():I
+*f:org.jetbrains.jewel.markdown.DimensionSize$Pixels
+- org.jetbrains.jewel.markdown.DimensionSize
+- bsf:box-impl(I):org.jetbrains.jewel.markdown.DimensionSize$Pixels
+- s:constructor-impl(I):I
+- equals(java.lang.Object):Z
+- s:equals-impl(I,java.lang.Object):Z
+- sf:equals-impl0(I,I):Z
+- f:getValue():I
+- hashCode():I
+- s:hashCode-impl(I):I
+- s:toString-impl(I):java.lang.String
+- bf:unbox-impl():I
 *:org.jetbrains.jewel.markdown.InlineMarkdown
 *f:org.jetbrains.jewel.markdown.InlineMarkdown$Code
 - org.jetbrains.jewel.markdown.InlineMarkdown
@@ -41,12 +66,16 @@
 - org.jetbrains.jewel.markdown.WithInlineMarkdown
 - sf:$stable:I
 - <init>(java.lang.String,java.lang.String,java.lang.String,java.util.List):V
+- <init>(java.lang.String,java.lang.String,java.lang.String,org.jetbrains.jewel.markdown.DimensionSize,org.jetbrains.jewel.markdown.DimensionSize,java.util.List):V
+- b:<init>(java.lang.String,java.lang.String,java.lang.String,org.jetbrains.jewel.markdown.DimensionSize,org.jetbrains.jewel.markdown.DimensionSize,java.util.List,I,kotlin.jvm.internal.DefaultConstructorMarker):V
 - <init>(java.lang.String,java.lang.String,java.lang.String,org.jetbrains.jewel.markdown.InlineMarkdown[]):V
 - equals(java.lang.Object):Z
 - f:getAlt():java.lang.String
+- f:getHeight():org.jetbrains.jewel.markdown.DimensionSize
 - getInlineContent():java.util.List
 - f:getSource():java.lang.String
 - f:getTitle():java.lang.String
+- f:getWidth():org.jetbrains.jewel.markdown.DimensionSize
 - hashCode():I
 *f:org.jetbrains.jewel.markdown.InlineMarkdown$Link
 - org.jetbrains.jewel.markdown.InlineMarkdown
--- a/platform/jewel/markdown/core/metalava/core-api-0.36.0.txt
+++ b/platform/jewel/markdown/core/metalava/core-api-0.36.0.txt
@@ -1,6 +1,21 @@
 // Signature format: 4.0
 package org.jetbrains.jewel.markdown {
 
+  @SuppressCompatibility @org.jetbrains.annotations.ApiStatus.Experimental @org.jetbrains.jewel.foundation.ExperimentalJewelApi public sealed interface DimensionSize {
+  }
+
+  @kotlin.jvm.JvmInline public static final value class DimensionSize.Percent implements org.jetbrains.jewel.markdown.DimensionSize {
+    ctor @KotlinOnly public DimensionSize.Percent(int value);
+    method public int getValue();
+    property public int value;
+  }
+
+  @kotlin.jvm.JvmInline public static final value class DimensionSize.Pixels implements org.jetbrains.jewel.markdown.DimensionSize {
+    ctor @KotlinOnly public DimensionSize.Pixels(int value);
+    method public int getValue();
+    property public int value;
+  }
+
   @SuppressCompatibility @org.jetbrains.annotations.ApiStatus.Experimental @org.jetbrains.jewel.foundation.ExperimentalJewelApi public sealed interface InlineMarkdown {
   }
 
@@ -37,16 +52,20 @@ package org.jetbrains.jewel.markdown {
   }
 
   @SuppressCompatibility @org.jetbrains.annotations.ApiStatus.Experimental @org.jetbrains.jewel.foundation.ExperimentalJewelApi @org.jetbrains.jewel.foundation.GenerateDataFunctions public static final class InlineMarkdown.Image implements org.jetbrains.jewel.markdown.InlineMarkdown org.jetbrains.jewel.markdown.WithInlineMarkdown {
-    ctor public InlineMarkdown.Image(String source, String alt, String? title, java.util.List<? extends org.jetbrains.jewel.markdown.InlineMarkdown> inlineContent);
-    ctor public InlineMarkdown.Image(String source, String alt, String? title, org.jetbrains.jewel.markdown.InlineMarkdown... inlineContent);
+    ctor public InlineMarkdown.Image(String source, String alt, String? title, java.util.List<? extends org.jetbrains.jewel.markdown.InlineMarkdown> inlineContent, optional org.jetbrains.jewel.markdown.DimensionSize? width, optional org.jetbrains.jewel.markdown.DimensionSize? height);
+    ctor public InlineMarkdown.Image(String source, String alt, String? title, org.jetbrains.jewel.markdown.InlineMarkdown[] inlineContent, optional org.jetbrains.jewel.markdown.DimensionSize? width, optional org.jetbrains.jewel.markdown.DimensionSize? height);
     method public String getAlt();
+    method public org.jetbrains.jewel.markdown.DimensionSize? getHeight();
     method public java.util.List<org.jetbrains.jewel.markdown.InlineMarkdown> getInlineContent();
     method public String getSource();
     method public String? getTitle();
+    method public org.jetbrains.jewel.markdown.DimensionSize? getWidth();
     property public String alt;
+    property public org.jetbrains.jewel.markdown.DimensionSize? height;
     property public java.util.List<org.jetbrains.jewel.markdown.InlineMarkdown> inlineContent;
     property public String source;
     property public String? title;
+    property public org.jetbrains.jewel.markdown.DimensionSize? width;
   }
 
   @SuppressCompatibility @org.jetbrains.annotations.ApiStatus.Experimental @org.jetbrains.jewel.foundation.ExperimentalJewelApi @org.jetbrains.jewel.foundation.GenerateDataFunctions public static final class InlineMarkdown.Link implements org.jetbrains.jewel.markdown.InlineMarkdown org.jetbrains.jewel.markdown.WithInlineMarkdown {
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
@@ -1,0 +1,21 @@
+package org.jetbrains.jewel.markdown
+
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+
+/** Represents a size value for an image dimension (width or height). Can be either a pixel value or a percentage. */
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public sealed interface DimensionSize {
+    /** A loaded image should be exactly [value] pixels in the specified dimension. */
+    @JvmInline
+    public value class Pixels(public val value: Int) : DimensionSize {
+        override fun toString(): String = "${value}px"
+    }
+
+    /** A loaded image should be [value]% of the specified dimension of the loaded image. */
+    @JvmInline
+    public value class Percent(public val value: Int) : DimensionSize {
+        override fun toString(): String = "$value%"
+    }
+}
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
@@ -112,13 +112,33 @@ public sealed interface InlineMarkdown {
         public val alt: String,
         public val title: String?,
         override val inlineContent: List<InlineMarkdown>,
+        public val width: DimensionSize? = null,
+        public val height: DimensionSize? = null,
     ) : InlineMarkdown, WithInlineMarkdown {
         public constructor(
             source: String,
             alt: String,
             title: String?,
             vararg inlineContent: InlineMarkdown,
-        ) : this(source, alt, title, inlineContent.toList())
+            width: DimensionSize? = null,
+            height: DimensionSize? = null,
+        ) : this(source, alt, title, inlineContent.toList(), width, height)
+
+        @Deprecated("Use a constructor with width and height parameters instead.", level = DeprecationLevel.HIDDEN)
+        public constructor(
+            source: String,
+            alt: String,
+            title: String?,
+            vararg inlineContent: InlineMarkdown,
+        ) : this(source, alt, title, inlineContent.toList(), null, null)
+
+        @Deprecated("Use a constructor with width and height parameters instead.", level = DeprecationLevel.HIDDEN)
+        public constructor(
+            source: String,
+            alt: String,
+            title: String?,
+            inlineContent: List<InlineMarkdown>,
+        ) : this(source, alt, title, inlineContent, null, null)
 
         override fun equals(other: Any?): Boolean {
             if (this === other) return true
@@ -130,6 +150,8 @@ public sealed interface InlineMarkdown {
             if (alt != other.alt) return false
             if (title != other.title) return false
             if (inlineContent != other.inlineContent) return false
+            if (width != other.width) return false
+            if (height != other.height) return false
 
             return true
         }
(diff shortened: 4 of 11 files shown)


AlexVanGogen:
Created using a skill from [here](https://github.com/JetBrains/intellij-community/pull/3429) :P

(except for images, those had to be added manually)

AlexVanGogen:
TIL some preview providers also support given width and height: https://about.gitlab.com/releases/2022/12/22/gitlab-15-7-released/#change-the-dimensions-of-images-in-markdown

Even though GitHub doesn't do it by default, we could do it almost for free (all that's required to add is a simple regex matcher on the text literal appended to the native image). So I'd put this PR on hold until this functionality is supported.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
```diff
@@ -120,6 +124,13 @@ public sealed interface InlineMarkdown {
             vararg inlineContent: InlineMarkdown,
         ) : this(source, alt, title, inlineContent.toList())
 
+        public constructor(
```
DanielSouzaBertoldi, line 136: Any particular reason why this public api wasn't marked as `@Deprecated` and `DeprecationLevel.Hidden`?

Given that `width` and `height` have a default value of `null`, I think we could be a bit more agressive here.

Also, if you decide on deprecating this constructor for real, I'd recommend moving both `width`/`height` to under `inlineContent` to avoid source breaking change 😄

platform/jewel/markdown/extensions/images/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImplTest.kt
```diff
@@ -25,11 +25,13 @@ import coil3.test.FakeImageLoaderEngine
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.test.StandardTestDispatcher
+import org.jetbrains.jewel.markdown.DimensionSize
 import org.jetbrains.jewel.markdown.InlineMarkdown
 import org.junit.Rule
 import org.junit.Test
 
 @OptIn(ExperimentalCoilApi::class)
+@Suppress("LargeClass")
```
DanielSouzaBertoldi, line 34: We really should suppress this detekt rule for test classes... Not asking you to do this in this PR, just leaving out this "mental" note 😄

DanielSouzaBertoldi (approved):
No major blocks, this is pretty cool stuff, congrats!

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
AlexVanGogen: Fixed, thanks!

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
```diff
@@ -51,45 +52,114 @@ public fun Node.readInlineMarkdown(markdownProcessor: MarkdownProcessor): List<I
...
+
+            is CMStrongEmphasis ->
+                InlineMarkdown.StrongEmphasis(openingDelimiter, readInlineMarkdown(markdownProcessor))
+
+            is CMCode -> InlineMarkdown.Code(literal)
+            is CMHtmlInline -> InlineMarkdown.HtmlInline(literal)
+            is CMImage -> {
+                val inlineContent = readInlineMarkdown(markdownProcessor)
+                val attrs = (next as? CMText)?.literal
+                if (attrs?.startsWith('{') == true) {
```
rock3r, line 87: Looks like this can lose text after the image attribute block. `attrs` is the whole following `CMText` node, so something like `foo{width=100} trailing text` no longer passes `isValidImageAttributes()` because the text node does not end with `}`, but we still advance to `next.next` and drop the trailing text.

Can we parse/consume only the leading balanced `{...}` prefix and keep the remaining suffix as text?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/html/Converters.kt
```diff
@@ -39,11 +40,32 @@ private object ImageConverter : HtmlElementConverter {
...
+    if (trimmed.isEmpty()) return null
+
+    if (trimmed.endsWith("%")) {
+        val digits = trimmed.dropLast(1).trim()
+        val value = digits.toIntOrNull() ?: return null
+        if (value < 0) return null
+        return DimensionSize.Percent(value)
+    }
+
+    val digits = trimmed.takeWhile { it.isDigit() }
```
rock3r, line 62: This accepts any value that starts with digits as pixels, so `100em`, `100px;`, or even `50%;` become pixel sizes. That seems broader than the documented/summarized support for unitless pixels, `px`, and `%`, and can silently reinterpret unsupported units as pixels.

Can we make this strict instead — accept only a non-negative integer, integer + `px`, or integer + `%`, and return `null` for the rest?

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -80,25 +85,98 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
+    }
+
+    /**
+     * Computes the placeholder size for the image.
+     *
+     * If both dimensions are specified, those are used. If only one dimension is specified, the other is scaled
+     * proportionally based on the loaded image's aspect ratio. If no dimensions are specified, the loaded image's
+     * original dimensions are used. While loading, a minimal placeholder is used unless dimensions are specified.
+     *
+     * For percentage values, they are treated as a percentage of the original image size.
```
rock3r, line 102: Question: should percentages be relative to the intrinsic image size? This is internally consistent with the implementation/tests, but it is surprising for HTML-like input: users usually read `` as 50% of the available/container width, not 50% of the bitmap's original width.

If container-relative sizing is not feasible here, could we at least document/release-note that `Percent` means a percentage of the loaded image's intrinsic dimension? Otherwise the same `50%` markup renders very differently depending on source image resolution.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
```diff
@@ -112,13 +112,33 @@ public sealed interface InlineMarkdown {
         public val alt: String,
         public val title: String?,
         override val inlineContent: List<InlineMarkdown>,
+        public val width: DimensionSize? = null,
```
rock3r, line 115: Could we document the new public dimension API? The semantics are not obvious from the type alone: whether percentages are intrinsic- or container-relative, what happens when only one axis is specified, and what "pixels" map to in the inline text placeholder all matter for callers constructing `InlineMarkdown.Image` directly.

A short KDoc on `width`/`height`, plus slightly more precise `DimensionSize` docs, would make the contract much clearer.

platform/jewel/markdown/core/metalava/core-api-0.36.0.txt
```diff
@@ -1,6 +1,21 @@
 // Signature format: 4.0
 package org.jetbrains.jewel.markdown {
 
+  @SuppressCompatibility @org.jetbrains.annotations.ApiStatus.Experimental @org.jetbrains.jewel.foundation.ExperimentalJewelApi public sealed interface DimensionSize {
```
rock3r, line 4: Since this is such an old PR, these Metalava dumps were generated against a now-stale API version. The current `jewel.release.version` will be `0.39.0` shortly, but the new `DimensionSize`/`Image` signatures are added to the `0.36.0` archive while `core-api-0.37.0.txt` still has the old signatures.

Can we regenerate/update the dump for the active/forthcoming Jewel API version instead, and drop the accidental `0.36.0` archive changes?

platform/jewel/markdown/core/api-dump-experimental.txt
```diff
@@ -41,12 +66,16 @@
 - org.jetbrains.jewel.markdown.WithInlineMarkdown
 - sf:$stable:I
 - <init>(java.lang.String,java.lang.String,java.lang.String,java.util.List):V
+- <init>(java.lang.String,java.lang.String,java.lang.String,org.jetbrains.jewel.markdown.DimensionSize,org.jetbrains.jewel.markdown.DimensionSize,java.util.List):V
```
rock3r, line 69: This IJ API dump still seems to reflect the pre-fix constructor order: it lists the new `InlineMarkdown.Image` constructor/default helper as `(String, String, String, DimensionSize, DimensionSize, List...)`, while the source now puts `inlineContent` before `width`/`height` to avoid the source-compatibility trap Daniel pointed out.

Can we regenerate the IJ dumps from the current source as well?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+
+/** Represents a size value for an image dimension (width or height). Can be either a pixel value or a percentage. */
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public sealed interface DimensionSize {
+    /** A loaded image should be exactly [value] pixels in the specified dimension. */
+    @JvmInline
+    public value class Pixels(public val value: Int) : DimensionSize {
```
rock3r, line 12: Looks like the public API allows constructing invalid sizes directly, even though the parsers reject them. For example, external code can still create `DimensionSize.Pixels(-1)` or `DimensionSize.Percent(-50)`, and the renderer will pass those through to placeholder sizing.

Can we enforce the accepted range here, or validate/drop invalid values before rendering? Either way, it would be good for the direct API contract to match the parsed Markdown/HTML path.

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -80,25 +85,98 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
+    @Composable
+    private fun computePlaceholder(
+        imageResult: SuccessResult?,
+        specifiedWidth: DimensionSize?,
+        specifiedHeight: DimensionSize?,
+    ): Placeholder {
+        val density = LocalDensity.current
+
+        // At least one dimension is unspecified or requires the image to compute; return the "empty" placeholder
+        if (imageResult == null) {
```
rock3r, line 113: Nit/question: while loading, could we already reserve the pixel dimensions that are explicitly requested? Right now we only do that when both axes are pixel values. For `{width=200}` we still use the `0.sp x 1.sp` placeholder until the image loads, even though the requested width is already known.

We cannot avoid all jumps before we know the intrinsic aspect ratio, but reserving the known axis and using a minimal value for the unknown one would improve the easy case.

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -80,25 +85,98 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
-            imageResult?.let {
-                val imageSize = it.image
-                with(LocalDensity.current) {
-                    // `toSp` ensures that the placeholder size matches the original image size in pixels.
-                    // This approach doesn't allow images from appearing larger with different screen scaling,
-                    // but simply maintains behavior consistent with standalone AsyncImage rendering.
+            computePlaceholder(imageResult = imageResult, specifiedWidth = image.width, specifiedHeight = image.height)
+
+        return InlineTextContent(placeholder) {
+            Image(painter = painter, contentDescription = image.title, modifier = Modifier.fillMaxSize())
```
rock3r, line 91: Looks like the placeholder is sized to the requested width/height, but the actual `Image` still uses the default `ContentScale.Fit`. So when both dimensions are specified with a different aspect ratio, the image content is fitted inside the box rather than stretched/squished to those dimensions.

The tests/comments describe those cases as stretched, so should this pass `contentScale = ContentScale.FillBounds`? If preserving aspect ratio is intentional, then I think the tests/docs should avoid implying that explicit two-axis dimensions stretch the image pixels.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
```diff
@@ -0,0 +1,21 @@
...
+        override fun toString(): String = "${value}px"
+    }
+
+    /** A loaded image should be [value]% of the specified dimension of the loaded image. */
+    @JvmInline
+    public value class Percent(public val value: Int) : DimensionSize {
```
rock3r, line 18: Could we also mark the nested public value classes as experimental? `DimensionSize` itself has `@ApiStatus.Experimental`/`@ExperimentalJewelApi`, but `Pixels` and `Percent` do not, unlike the other public nested Markdown model types.

This also leaks into the API dumps: tools/callers looking at the directly constructed value classes do not see the experimental marker.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
```diff
@@ -51,45 +52,114 @@ public fun Node.readInlineMarkdown(markdownProcessor: MarkdownProcessor): List<I
...
-        is CMHtmlInline -> InlineMarkdown.HtmlInline(literal)
-        is CMImage -> {
-            val inlineContent = readInlineMarkdown(markdownProcessor)
-            InlineMarkdown.Image(
-                source = destination,
-                alt = inlineContent.renderAsSimpleText().trim(),
-                title = title,
-                inlineContent = inlineContent,
-            )
+    convertToInlineMarkdown(markdownProcessor).first
```
rock3r, line 59: I'm a bit worried about making `toInlineMarkdownOrNull()` sibling-aware. Its KDoc says it converts this node and returns only one `InlineMarkdown?`, but `convertToInlineMarkdown()` can now peek at `this.next` for image attributes. Public callers that walk `node.next` themselves have no way to know that the following text node was logically consumed, so they can get both `Image(width=...)` and a separate `Text("{width=...}")`.

Could we keep this public single-node API node-local, and reserve sibling-aware image-attribute parsing for `readInlineMarkdown()` where we can also advance to the returned next node? It should be up to the inline parsers to figure out if they need more nodes.

rock3r (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/html/Converters.kt
DanielSouzaBertoldi: In HTML just having something like `width = 50` is treated as being `50px`. So we can be a bit more lenient in that regard and always fallback to pixel as the unit in this case. What do you think?

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
DanielSouzaBertoldi: We need to declare the size of the `Placeholder` from `InlineTextContent` upfront, before the text is laid out, so we unfortunately cannot mirror the HTML behavior faithfully and make it take the percentage of the available container size :(

I'll document it here properly + in the release notes

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
DanielSouzaBertoldi: Now, this got me thinking... Cap percentage to just 100% or let users go wild and make images look as big as they want to? Then they could scale beyond the intrinsic size like 200% or whatever

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
DanielSouzaBertoldi: I think we can simplify this to just this:

```kotlin
if (imageResult == null) {
    // For pixel values, we can show them immediately; for percentage we need the image
    val pixelWidth = (specifiedWidth as? DimensionSize.Pixels)?.value
    val pixelHeight = (specifiedHeight as? DimensionSize.Pixels)?.value
    
    return with(density) {
        Placeholder(
            width = pixelWidth?.toSp() ?: 1.sp,
            height = pixelHeight?.toSp() ?: 1.sp,
            placeholderVerticalAlign = PlaceholderVerticalAlign.Bottom,
        )
    }
}
```

Then it doesn't matter if users declared just one or the other, both, or neither. Besides, `1.sp` of width as a fallback value is a bit more "cleaner" to me than just... nothing. Jumps will jump, nothing we can do about that anywho

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
DanielSouzaBertoldi: I say let's mirror HTML in this case. If you declare both width and height, then the image fills the bounds instead of preserving aspect ratio

image

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
DanielSouzaBertoldi: Hmmmm.... In this case we can just move the InlineMarkdownImage logic to `readInlineMarkdown()`. Got it

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
rock3r: I think letting users go above 100% is fine — `200%` is a reasonable way to say "twice the intrinsic size" with the current semantics. The bit I'd avoid is negative or otherwise invalid values, since those don't have a useful rendering meaning and the parsers already reject them.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/html/Converters.kt
rock3r: Yeah, unitless numbers being pixels makes sense to me too — I don't want to reject `width=50`. I meant the suffix handling after that: values like `100em`, `100px;`, or `50%;` currently become pixels because we only take the leading digits. I'd rather keep unitless numbers + explicit `px`/`%`, but reject unsupported or malformed suffixes instead of silently falling back to pixels.

PR #3449, opened 2026-03-05
[JEWEL-1285] Update JDialogRenderer to Support Custom Window Shape

DanielSouzaBertoldi:
# Context
During the development of the Got It component, we ran into a limitation of our current `JDialogRenderer`. Currently, it can only render windows with rounded corners. While this works perfectly for menus and other types of popups, it fails to provide the best UX for the Got It component since it has a custom shape.

So, this PR paves the way so that our Got It component can be rendered in all its glory!

# Changes
- Created new variations of `Popup` that take `windowShape: ((IntSize) -> java.awt.Shape)? = null,`. Deprecated variations without it.
- Added `windowShape: ((IntSize) -> java.awt.Shape)? = null,` lambda parameter to `PopupRenderer` and passed it down to both `JDialogRenderer` and `JBPopupRenderer`
- Updated `JDialogRenderer` code to apply the custom window shape defined by the factory lambda if:
   1. `compose.interop.blending` is disabled
   2. `windowShape` is not null

# Demo
| Scenario | Screenshot |
|--------|--------|
| Before | image |
| After | image |
| Custom Popup Enabled, Window Shape Provided but Compose Blending Disabled | <video src="https://github.com/user-attachments/assets/0173c710-ebff-403b-a733-838079a9e423" /> |

| Scenario | Screenshot |
|--------|--------|
| Arrow pointing up  | image |
| Arrow pointing down | image |
| Arrow pointing left | image |

Curiously, this minor artifact is not present when the arrow is to the right. No idea why, though.

### New features
 * New variations of `Popup` is available which take a `windowShape` factory. With this, you can create a custom shape that'll be applied to the window which holds the popup. Bear in mind that this factory is only useful in the following scenario:
    * You have the system property `jewel.customPopupRender` flag enabled
    * Your project is running as a Desktop app. IJP plugins are unaffected by this change
    * Your project does not have the system property `compose.interop.blending` enabled
* Keep in mind that forcing to clip the window to a shape has some limitations: Swing's `WindowTranslucency` can't render concave corners properly since it forces pixels to be either completely off or on. This means that depending on your custom shape, you can have jagged edges. For this reason, it's highly encouraged to also enable `compose.interop.blending` in your system properties.

### Deprecated API
 * `Popup` components without the `windowShape` parameter are now deprecated.


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -110,6 +248,7 @@ public interface PopupRenderer {
         onPreviewKeyEvent: ((KeyEvent) -> Boolean)?,
         onKeyEvent: ((KeyEvent) -> Boolean)?,
         cornerSize: CornerSize,
+        windowShape: ((IntSize) -> java.awt.Shape)? = null,
```
daaria-s: Wouldn't it break any existing external implementations of PopupRenderer?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
```diff
@@ -46,22 +46,158 @@ import org.jetbrains.jewel.foundation.JewelFlags
...
     popupPositionProvider: PopupPositionProvider,
+    cornerSize: CornerSize,
     onDismissRequest: (() -> Unit)? = null,
     properties: PopupProperties = PopupProperties(),
     onPreviewKeyEvent: ((KeyEvent) -> Boolean)? = null,
     onKeyEvent: ((KeyEvent) -> Boolean)? = null,
     content: @Composable () -> Unit,
 ) {
-    Popup(popupPositionProvider, ZeroCornerSize, onDismissRequest, properties, onPreviewKeyEvent, onKeyEvent, content)
+    if (JewelFlags.useCustomPopupRenderer) {
```
daaria-s: Do we need this duplication? Looks like we can delegate it like:
`Popup(popupPositionProvider, cornerSize, onDismissRequest, properties, onPreviewKeyEvent, onKeyEvent, windowShape = null, content)`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
DanielSouzaBertoldi: Ooof, yes this breaks external implementations. Great catch!

I'll update the code to have a deprecated function with the old signature. Thanks!

daaria-s (approved):
Thank you for the updates! LGTM

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
AlexVanGogen: It's not entirely true: `blendingEnabled` can be `true` even without `compose.interop.blending`. Having a transparent `Panel.background` is enough.

(And I'm not sure that dialog panels cannot have a custom background, different from `Panel.background`, but it's out of scope for this PR anyway)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
AlexVanGogen: WDYT about mentioning `compose.interop.blending` and transparent panel here? They also affect whether `windowShape` gets invoked

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
DanielSouzaBertoldi: Removed `(via compose.interop.blending)` part, thanks!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Popup.kt
DanielSouzaBertoldi: Yeah, I agree. `windowShape` is not just limited by our custom flag, but it fails silently if the panel background is transparent. Updated the KDoc. Thanks!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -240,6 +267,33 @@ private fun JPopupImpl(
...
+                                        // java.awt.GraphicsDevice.WindowTranslucency.PERPIXEL_TRANSLUCENT mode
+                                        // (per-pixel alpha). Calling Window.setShape() would switch it to
+                                        // PERPIXEL_TRANSPARENT (hard pixel clip), breaking antialiasing at the edges.
+                                        // Compose's own drawing + transparent background is sufficient.
+                                        return@JPopupMeasurePolicy
+                                    }
+                                    // Without blending, fall back to Window.setShape() to at least
+                                    // clip the rectangular window boundary to the balloon outline.
+                                    // Note: this uses PERPIXEL_TRANSPARENT mode, which has known
+                                    // antialiasing limitations at concave corners (e.g. arrow junction).
```
rock3r, line 285: Worth a one-off log?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -240,6 +267,33 @@ private fun JPopupImpl(
...
+                                    val logicalSize =
+                                        IntSize(
+                                            floor(size.width / dialog.density()).toInt(),
+                                            floor(size.height / dialog.density()).toInt(),
+                                        )
+                                    try {
+                                        dialog.shape = currentWindowShape(logicalSize)
```
rock3r, line 292: Looks like the explicit AWT shape can stick around if `windowShape` changes back to `null` while this remembered `JDialog` is still alive.

In the non-null path we assign `dialog.shape = currentWindowShape(logicalSize)`, but the null path only goes through the rounded-corners fallback and never clears the shape. So a popup that recomposes from a custom shape to no custom shape could keep the previous native clip, especially when `cornerSize == ZeroCornerSize` and `applyRoundedCorners()` returns immediately.

Can we reset `dialog.shape` before the non-custom-shape fallback?

rock3r (changes requested):

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
DanielSouzaBertoldi: Hell yeah 😎

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
DanielSouzaBertoldi: Great catch! Will do 🙏🏻

rock3r (approved):

nebojsa-vuksic (approved):

DanielSouzaBertoldi:
Ready to merge

PR #3451, opened 2026-03-06
[JEWEL-1150] Create Got It Tooltip Component

DanielSouzaBertoldi:
> [!IMPORTANT]
> This PR is a direct continuation of  #[3449](https://github.com/JetBrains/intellij-community/pull/3449)! While that PR is not merged, just ignore these three files changed: `JBPopupRenderer.kt`, `Popup.kt` and `JDialogRenderer.kt`

# Context

This PR adds the greatest component EVER to Jewel's arsenal. Don't let the "tooltip" part in its name fool you. This does **NOT** act like a tooltip **at all**, since a tooltip by definition should only be active while your cursor is hovering a component, and this weird guy loves to behave like your plain 'ol popup. It won't disappear unless you click its main button (actually there are other ways to close it which you'll find out during code review, yes this your encouragement to go through the code carefully and find out all its features. Or just check the screen recordings below if you feel like cheating and don't mind a guilty conscience), and hangs around basically begging for the user's attention.

The cool thing though is that you are free to choose where it's going to show up below, on top or at the start/end of another component, and it has a very cool custom shape that resembles a balloon. Everybody loves balloons!

## Current limitations
1. This PR does not add support for keyboard shortcuts.
2. This PR does not add support for Lottie animations (since they don't provide support for CfD AFAIK)
3. The main button in IJP's version can be colored with a vertical linear gradient. Ours, for now, does not.
4. We do not support contrast colors.

# Changes

* Added `GotItTooltip` composable with balloon popup, auto-positioning, and configurable `GotItBody` DSL
* Added `BalloonShape` for drawing the balloon outline with directional arrow tip
* Added `BalloonPopupPositionProvider` for positioning the balloon relative to its anchor
* Extracted `calculateBalloonPosition` and `createBalloonOutline` as `@InternalJewelApi` top-level functions in `BalloonInternals`
* Added bridge theming via `readGotItTooltipStyle` / `readGotItButtonStyle` in `IntUiBridgeGotItTooltip`
* Added standalone (Int UI) theme defaults

# Demo

## Themes
| | Light | Dark |
|---|--------|--------|
| IJP | image | image |
| Jewel | image | image |

### Main Options
| | IJP | Jewel |
|---|--------|--------|
| Header | image | image |
| Image w/ Borders | image |  image |
| Image w/o Borders | image | image |
| Icon | image | image |
| Step (smaller than 10) | image | image |
| Step (greater than 9) | image |image |
| Max Width (150) | image | image |
| Max Width w/ Image | image | image |

### Buttons
| | IJP | Jewel |
|---|--------|--------|
| Without Buttons | image | image |
| Only Default Primary Button | image | image |
| W/ Secondary Button | image | image |
| Primary Button Hover | <video src="https://github.com/user-attachments/assets/f4da01db-3b00-4196-ac12-602e362af5cb" /> | <video src="https://github.com/user-attachments/assets/bbb2dc88-0ebe-4d54-81e2-6abb34c315bd" /> |
| Primary Button Click | <video src="https://github.com/user-attachments/assets/1452a052-8a85-4235-8cac-f1cb3aaef87a" /> | <video src="https://github.com/user-attachments/assets/4ab6f59a-1b94-4eec-b831-bc86a8c26118" /> |
| Secondary Button Hover | <video src="https://github.com/user-attachments/assets/4d869c12-6c4c-49f2-96ea-c36cbad5adb5" /> | <video src="https://github.com/user-attachments/assets/4cf4b4b5-d1e7-43b5-b852-2bad951c797b" /> |
| Secondary Button Click | <video src="https://github.com/user-attachments/assets/c7ba0240-60dc-475a-a4a7-9c85b09df7ef" /> | <video src="https://github.com/user-attachments/assets/5e9f93c3-e79f-4346-8477-bbca145e71ed" /> |
| Link | image | image |
| Browser Link | image | image |

_**Note:** Yes, on click shouldn't show a blue outline. However, that would take a bit of refactoring on the Button composable side which is not part of this PR!_ 

### Body text
| | IJP | Jewel |
|---|--------|--------|
| Plain Text | image | image |
| Code | image | image |
| Icon | image | image |
| Bold text | image | image |
| Link | image | image |
| External Link | image | image |
| Hovering Inline Links | <video src="https://github.com/user-attachments/assets/a57191ad-793a-40a2-9c15-98fc579ca146" /> | <video src="https://github.com/user-attachments/assets/974df5c3-8600-4861-853e-9f06eec712d2" /> |

### Behavioral
| | IJP | Jewel |
|---|--------|--------|
| Link (Browser) | <video src="https://github.com/user-attachments/assets/9fc2e7e9-a610-41eb-bf2c-5e9fc3d8fe7e" /> | <video src="https://github.com/user-attachments/assets/ac077d0d-bfc6-4bf4-ba14-ea77e3c5f8a0" /> |
| Link (Internal Action) | <video src="https://github.com/user-attachments/assets/17cc887c-912c-4601-b34a-027da23c1a27" /> | <video src="https://github.com/user-attachments/assets/cd6099b0-f660-4737-81a2-d285dd583af2" /> |
| Timeout (3 seconds) | <video src="https://github.com/user-attachments/assets/0bf2f9e4-cd02-4776-a63c-057717bbb6fb" /> | <video src="https://github.com/user-attachments/assets/99a22830-3c66-4426-8098-285ae7b29c3a" /> |
| Max Width w/ Image | image | <video src="https://github.com/user-attachments/assets/d28ce7c9-5d0e-40ed-a466-d9a4ffdaefcd" /> |
| Text < 5 lines w/o image/max width | image | image |
| Text < 5 lines w image | image | image |
| Text < 5 lines w max width | image | <video src="https://github.com/user-attachments/assets/d62d89a6-f9cb-4988-a83a-5b0039a80613" /> |
| Text >=5 lines w/o image/max width | image | <video src="https://github.com/user-attachments/assets/7ddea4fe-6387-49d3-a184-e6a4e2090963" /> |
| Text >=5 lines w image | image | image |
| Text >=5 lines w max width | image | image |
| `onEscapePress` defined | DISCUSSION PENDING. CHECK COMMENTS | <video src="https://github.com/user-attachments/assets/77f47800-2bd1-4456-88c0-027d247a0dd6" /> |
| `onEscapePress` undefined | <video src="https://github.com/user-attachments/assets/eba78edb-c7f4-4083-a83f-1d6f9cf51238" /> | <video src="https://github.com/user-attachments/assets/42a1ba5b-5b6c-4cf5-86f8-0a330aa11e4d" /> |

### Got It in its full glory (aka all possible customizations at once)

| | Light | Dark |
|---|--------|--------|
| IJP | image | image |
| Jewel | image | image |

### Popup Renderer

| | Compose | Jewel w/o Compose Interop Blending | Jewel w/ Compose Interop Blending |
|---|--------|--------|--------|
| Light | <video src="https://github.com/user-attachments/assets/e599dbda-410c-4d28-b662-93c6502e6e06" /> | <video src="https://github.com/user-attachments/assets/488bb4b4-b706-4337-8909-e03631879504" /> | <video src="https://github.com/user-attachments/assets/2279afa9-d4f6-48fa-9fdb-6a89a7fe0791" /> |
| Dark | <video src="https://github.com/user-attachments/assets/685c3f7c-e827-4458-be2c-65303cc05874" /> | <video src="https://github.com/user-attachments/assets/1ea53a42-054b-4051-8ecf-620c121d105c" /> | <video src="https://github.com/user-attachments/assets/ade00eea-fb4a-4d75-a284-6b48affd6ac4" /> |

### New features
 * Added new `GotItTooltip` component with a balloon popup, rich body DSL (text, link, shortcut segments), optional header, primary/secondary buttons, step indicator, timeout auto-dismiss, and keyboard escape support. Check the showcase for all possible customizations.


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
```diff
@@ -0,0 +1,971 @@
...
+    style: GotItTooltipStyle = LocalGotItTooltipStyle.current,
+    content: @Composable () -> Unit,
+) {
+    val layoutDirection = LocalLayoutDirection.current
+    val cornerRadiusPx = style.metrics.cornerRadius.value.roundToInt()
+    val currentOnDismiss by rememberUpdatedState(onDismiss)
+    val currentOnShow by rememberUpdatedState(onShow)
+    val currentOnEscapePress by rememberUpdatedState(onEscapePress)
+    val shouldListenForEscapeKeyPress by
+        rememberUpdatedState(visible && currentOnEscapePress != null || buttons.hasNoButtons())
```
DanielSouzaBertoldi: So, we have a decision to make here. I humbly ask for your input :)

In `GotItComponentBuilder` from IJP, there's this function:

```
/**
 * Sets the action to be executed when the escape key is pressed.
 * Note that the popup will be closed after the action execution.
 * If escape action is not set, the popup won't be closed on the escape key.
 */
fun onEscapePressed(action: () -> Unit): GotItComponentBuilder {
  this.escapeAction = action
  return this
}
```

The last line in the comment states pretty convincingly that, if users don't set a escape action, then the popup will not be closed on the escape key.

This however, is false. It doesn't matter if you set up a escape action or not, the popup will always be dismissed when users press escape.

The funny thing is, it doesn't say anything about how the component can be closed if there's no buttons, so that's why I added a "escape hatch" for such cases. If there are no buttons, then users can (and should, unless developers externally set `visible` to `false` on some other condition) close the popup when pressing the escape key.

So, I decided on follow what's actually written in the code instead of its actual behavior.

What do you think? Stick to the code intention or the actual behavior?

rock3r:
> Yes, on click shouldn't show a blue outline. However, that would take a bit of refactoring on the Button composable side which is not part of this PR!

I think you could override the focus outline colour or size with a composition local on the button(s) container so that should be easy (if slightly painful) but I'm also ok with tracking it for a follow-up

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
rock3r: I think we should match the actual behaviour, which is that esc always escapes it (and perhaps file an IJPL bug for the KDoc mismatch in Swing?)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
```diff
@@ -0,0 +1,971 @@
...
+/**
+ * A Got It tooltip component that provides contextual information about new or changed features.
+ *
+ * Got It tooltips are specialized popups (called "Balloons" in the Swing implementation) that:
+ * - Appear near UI elements to guide users
+ * - Can display icons or step numbers
+ * - Support rich content (headers, links, images, buttons)
+ * - Auto-hide after a timeout (optional, mutually exclusive with buttons)
+ * - Display an arrow pointing to the target element
+ * - Can limit how many times they appear (showCount)
```
rock3r: Could we drop or rephrase this bullet? I don't see a `showCount`/equivalent parameter in the Jewel API added here — visibility is fully caller-controlled via `visible`. As written, the generated docs suggest built-in show-count support similar to Swing, but users won't find an API for it.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
```diff
@@ -0,0 +1,971 @@
...
+    style: GotItTooltipStyle,
+) {
+    var imageWidth by remember { mutableStateOf<Dp?>(null) }
+    val density = LocalDensity.current
+    val editorTextStyle = JewelTheme.editorTextStyle
+    val externalLinkIconKey = LocalLinkStyle.current.icons.externalLink
+
+    // Auto-extension (280 → 328 dp) is only allowed when there is no image and no custom maxWidth
+    val allowWidthExtension = image == null && maxWidth == null
+    var currentWidth by remember(maxWidth) { mutableStateOf(maxWidth ?: DEFAULT_TOOLTIP_WIDTH) }
```
rock3r: Looks like this width state can get stuck at the extended width if the tooltip body changes while the composable stays alive. A long body can set `currentWidth` to `DEFAULT_TOOLTIP_EXTENDED_WIDTH`, but the state is only remembered by `maxWidth`, so a later shorter body with the same `maxWidth == null` won't reset back to `DEFAULT_TOOLTIP_WIDTH`.

Can we include the content-driving inputs here, or otherwise reset the width when `body`/`image`/the auto-extension inputs change?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltipBody.kt
```diff
@@ -0,0 +1,252 @@
...
+    @JvmInline value class Code(val text: String) : GotItBodySegment
+
+    data class InlineLink(val text: String, val action: () -> Unit) : GotItBodySegment
+
+    data class BrowserLink(val text: String, val uri: String) : GotItBodySegment
+
+    data class Icon(val contentDescription: String?, val content: @Composable () -> Unit) : GotItBodySegment
+}
+
+@VisibleForTesting
```
rock3r, line 95: Since this is public for the separate `ui-tests` module, could we mark it as internal Jewel API too? There are existing visible-for-testing public helpers using `@ApiStatus.Internal` + `@InternalJewelApi`; without those, this test/rendering helper shows up as normal published API and becomes something external consumers may depend on. Same for `buildInlineContent` below.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
```diff
@@ -0,0 +1,971 @@
...
+        var result = label.hashCode()
+        result = 31 * result + action.hashCode()
+        return result
+    }
+
+    override fun toString(): String = "GotItButton(label='$label', action=$action)"
+
+    public companion object {
+        /** Default "Got it" button with no extra action */
+        public val Default: GotItButton = GotItButton(label = "Got it")
```
rock3r: Can we avoid hardcoding this user-visible label in common UI API? Swing uses `IdeBundle.message("got.it.button.name")`, and Jewel already has `LocalMessageResourceResolverProvider` for IDE bundle strings in UI components.

With the current singleton, bridge/IJP usage gets an English-only default in localized IDE builds, and the public API bakes that string in. Maybe the default label could be resolved at render time via resource/theme plumbing, with a standalone fallback if needed?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltipBody.kt
```diff
@@ -0,0 +1,252 @@
...
+                    withStyle(SpanStyle(color = styleColors.link)) { append(segment.text) }
+                }
+            is GotItBodySegment.BrowserLink -> {
+                withLink(LinkAnnotation.Url(url = segment.uri, styles = linkStyles)) {
+                    withStyle(SpanStyle(color = styleColors.link)) { append(segment.text) }
+                }
+                append(" ")
+                appendInlineContent(browserLinkArrowId(index), "↗")
+            }
+            is GotItBodySegment.Icon -> appendInlineContent(iconInlineId(index), segment.contentDescription ?: "\uFFFD")
```
rock3r: I think this makes decorative inline icons visible to text/accessibility consumers as `�`. If `contentDescription` is `null`, I would expect the icon to be decorative rather than announced/copied/searched as the Unicode replacement character.

Can we use empty alternate text if Compose allows it, or otherwise avoid exposing U+FFFD for decorative icons? The current test seems to lock in the undesirable behaviour.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
```diff
@@ -0,0 +1,971 @@
...
+
+    Column {
+        when (link) {
+            is GotItLink.Regular -> {
+                Link(text = link.label, onClick = link.action, style = overriddenLinkStyle)
+            }
+
+            is GotItLink.Browser -> {
+                val uriHandler = LocalUriHandler.current
+                ExternalLink(
```
rock3r, line 802: Could we keep the same safe URI handling as `ExternalLink(text, uri, ...)` here? That overload routes through `openUri(...)`, which catches/logs `IllegalArgumentException` and `IOException`. This manual `uriHandler.openUri(link.uri)` bypasses that, so a bad/unsupported URI can throw out of the click handler, and then `link.action()` won't run either.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
```diff
@@ -0,0 +1,971 @@
...
+                        image = image,
+                        maxWidth = maxWidth,
+                        onDismiss = onDismiss,
+                        style = style,
+                    )
+
+                    LaunchedEffect(Unit) { currentOnShow() }
+
+                    if (timeout.isFinite()) {
+                        LaunchedEffect(Unit) {
```
rock3r: Should this be keyed by `timeout`? If the tooltip stays visible and the caller changes from one finite timeout to another, `LaunchedEffect(Unit)` keeps the original delay running, so dismissal still happens at the old time. `LaunchedEffect(timeout)` should restart the timer while the existing finite guard still cancels it when switching to `Duration.INFINITE`.

platform/jewel/ui/metalava/ui-api-0.37.0.txt
```diff
@@ -15,7 +15,7 @@ package org.jetbrains.jewel.ui {
...
....RadioButtonStyle radioButtonStyle, org.jetbrains.jewel.ui.component.styling.ScrollbarStyle scrollbarStyle, org.jetbrains.jewel.ui.component.styling.SegmentedControlButtonStyle segmentedControlButtonStyle, org.jetbrains.jewel.ui.component.styling.SegmentedControlStyle segmentedControlStyle, org.jetbrains.jewel.ui.component.styling.SelectableLazyColumnStyle selectableLazyColumnStyle, org.jetbrains.jewel.ui.component.styling.SimpleListItemStyle simpleListItemStyle, org.jetbrains.jewel.ui.component.styling.SliderStyle sliderStyle, org.jetbrains.jewel.ui.component.styling.TextAreaStyle textAreaStyle, org.jetbrains.jewel.ui.component.styling.TextFieldStyle textFieldStyle, org.jetbrains.jewel.ui.component.styling.TooltipStyle tooltipStyle, org.jetbrains.jewel.ui.component.styling.DropdownStyle undecoratedDropdownStyle, org.jetbrains.jewel.ui.component.styling.SpeedSearchStyle speedSearchStyle, org.jetbrains.jewel.ui.component.styling.SearchMatchStyle searchMatchStyle, org.jetbrains.jewel.ui.component.styling.PopupAdStyle popupAdStyle, org.jetbrains.jewel.ui.component.styling.ButtonStyle defaultSlimButtonStyle, org.jetbrains.jewel.ui.component.styling.ButtonStyle outlinedSlimButtonStyle);
```
rock3r: These will need re-generating once 0.39 dumps land on master

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/DefaultComponentStyling.kt
```diff
@@ -124,7 +130,95 @@ public class DefaultComponentStyling(
     public val popupAdStyle: PopupAdStyle,
     public val defaultSlimButtonStyle: ButtonStyle,
     public val outlinedSlimButtonStyle: ButtonStyle,
+    public val gotItTooltipStyle: GotItTooltipStyle,
+    public val gotItButtonStyle: ButtonStyle,
 ) : ComponentStyling {
+    @Deprecated(
+        message = "Use the constructor with `gotItButtonStyle` and `gotItButtonStyle`.",
```
rock3r: Nit: copy-paste mistake — this should say `gotItTooltipStyle` and `gotItButtonStyle`, not `gotItButtonStyle` twice.

rock3r (changes requested):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
DanielSouzaBertoldi: Ooof, that was a leftover for when I started working on this. Later down the line I realized that users can control how many times this tooltip can show by themselves, trying to provide such behavior in a declarative framework would be useless. Thanks for pointing this out!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltipBody.kt
DanielSouzaBertoldi: Unfortunately Compose does not accept `alternateText` that's empty :/ from their KDoc:

```
 * @param alternateText The text to be replaced by the inline content. It's displayed when the
 *   inlineContent parameter of [BasicText] doesn't contain [id]. Accessibility features will also
 *   use this text to describe the inline content.
 * @throws IllegalArgumentException if [alternateText] has zero length.
```

But I guess we could use [zero-width space unicode character](https://unicode-explorer.com/c/200B), that's `\u200B`. It satisfies both the Compose constraint and the a11y concern

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
rock3r: @DanielSouzaBertoldi was this addressed? Code still looks the same as far as i can tell

rock3r (changes requested):
Need to address `rememberUpdatedState(visible && currentOnEscapePress != null || buttons.hasNoButtons())` change, then will be good to go

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/gotit/GotItTooltip.kt
DanielSouzaBertoldi: Not yet, I'll file the IJPL bug now, update our code and record the behavior :)

(Edited because after talking to Gemini "evidences" is not a common way of phrasing this sort of stuff in english, which I think it's a shame but I digress)

DanielSouzaBertoldi: Bug filed: https://youtrack.jetbrains.com/issue/IJPL-248104/Outdated-Got-It-Component-onEscapePressed-KDoc

rock3r:
Can you fix the CI? :)

rock3r (approved):

DanielSouzaBertoldi:
Ready to merge

DanielSouzaBertoldi:
@rock3r @nebojsa-vuksic Asking for another quick round of reviews :)

Given that we had some i18n errors when trying to merge the PR, I had to rearrange some code. To make it easier to understand what changed here's a list of the main changes:

1 - [GotItTooltip](https://github.com/JetBrains/intellij-community/pull/3451/changes#diff-65d5ce0a21f63b888ffee0e4e9a95a3a1990b0b8c42ff2ac966ddd0b8897ef5c)
  * This was the tip of the iceberg. I deleted the companion object entirely (it only held a `Default` val whose label was a hardcoded string, "Got It")
  * Replaced `GotItButtons.Default` (a val) `with @Composable fun default()`, which reads `LocalMessageResourceResolverProvider` and resolves `got.it.button.name`. A static can't read a `CompositionLocal`, which is why the literal existed in the first place
  * Changed the `buttons` parameter to `buttons: GotItButtons = GotItButtons.default()` in both `GotItTooltip` overloads
  * KDoc for `@param primary` now points at `GotItButtons.default` instead of quoting a literal

2 - [MessageResourceResolver](https://github.com/JetBrains/intellij-community/pull/3451/changes#diff-14546531642f3035ca6ad853b60a9c86e08d224dda03054e4442558ec001c827)
  * Added the `@Nls` annotation on `resolveIdeBundleMessage`, so resolved messages carry the NLS contract into `GotItButton(@Nls label)`.

3 - [GotIt.kt](https://github.com/JetBrains/intellij-community/pull/3451/changes#diff-06b53c3ad125e88afd22b25aee1574e170e4871d3025b4fef4c6c52b4e257ca6)
  * The `usePrimaryDefault` branch now uses `GotItButtons.default().primary`, hoisted above `remember(defaultPrimaryButton)`, instead of resolving the bundle key by hand
  * Added a `@file:Suppress("HardCodedStringLiteral")` because the demo is intentionally not localized

rock3r (approved):
:shipit:

PR #3454, opened 2026-03-09
[JEWEL-1277] Add FrontMatter metadata block support for Markdown

AlexVanGogen:
## Summary

Add support for YAML front matter metadata blocks in Jewel's Markdown rendering. Front matter content (delimited by `---` markers at the beginning of a file) is parsed and rendered as a table where keys form the first column and values form the second column. It differs from other renderers that prefer using rows (likely because then the keys will render in bold), but it helps keeping all the information in a Markdown file readable without coming down to horizontal scrolling.

Multi-value (list) entries are rendered as Markdown lists (opposite to one-row tables in other renderers), again, to keep the vertical structure of the block.

Fixes JEWEL-1277

## Changes

* New `front-matter` extension module (`intellij.platform.jewel.markdown.extensions.frontMatter`) with a `FrontMatterProcessorExtension` that integrates with Jewel's Markdown processor pipeline
* Custom block parser (`FrontMatterBlockParser`) supporting plain key-value pairs, YAML lists, quoted strings, and block scalars (literal `|` and folded `>` with all chomping indicators)
* Refactored `gfm-tables` Table model classes (`TableBlock`, `TableCell`, `TableHeader`, `TableRow`) from `internal` to `public` experimental API to allow reuse by front matter rendering
* Changed `TableBlock.header` to nullable (`TableHeader?`) to support headerless tables (used for nested value tables)
* Changed `TableCell.content` from `List<InlineMarkdown>` to `MarkdownBlock` to support block-level content in cells (e.g., nested tables)
* Updated `GitHubTableBlockRenderer` to render block content in cells via `RenderBlock` instead of `RenderParagraph`
* Added 22 unit tests covering various front matter scenarios including block scalars, lists, quoted values, and edge cases

### Tested scenarios

| Use case | Screenshot |
|---|---|
| Simple key-value block | Screenshot 2026-04-01 at 19 21 11 |
| Single key block | Screenshot 2026-04-01 at 19 21 47 |
| Empty block | Screenshot 2026-04-01 at 19 23 28 |
| List | Screenshot 2026-04-01 at 19 24 08 |
| Single item | Screenshot 2026-04-01 at 19 24 39 |
| Mixed scalar and list values | Screenshot 2026-04-01 at 19 25 11 |
| Block parsing stops on non-front-matter syntax | Screenshot 2026-04-01 at 19 25 49 |
| Block not at document start | Screenshot 2026-04-01 at 19 26 44 |
| List items with colon | Screenshot 2026-04-01 at 19 27 12 |
| Scalar with newlines | Screenshot 2026-04-01 at 19 32 30 |
| Scalar with leading lines | Screenshot 2026-04-01 at 19 28 27 |
| Strip chomping | Screenshot 2026-04-01 at 19 28 59 |
| Keep chomping | Screenshot 2026-04-01 at 19 31 21 |
| Folding | Screenshot 2026-04-01 at 19 33 50 |
| Folding + strip chomping | Screenshot 2026-04-01 at 19 34 23 |
| Folding + keep chomping | Screenshot 2026-04-01 at 19 35 03 |
| Folding + blank lines | Screenshot 2026-04-01 at 19 35 38 |
| Empty scalar | Screenshot 2026-04-01 at 19 36 15 |
| Empty scalar + key | Screenshot 2026-04-01 at 19 37 14 |
| Non-empty scalar + key | Screenshot 2026-04-01 at 19 37 42 |
| Unquote quotes | Screenshot 2026-04-01 at 19 38 05 |

### Typing in

https://github.com/user-attachments/assets/a74c2697-4a34-4891-a488-8a75d302fa05

### New features
* Added YAML front matter metadata block support for Markdown documents — front matter is rendered as a key-value table (#3454)


platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
```diff
@@ -0,0 +1,113 @@
...
+                                    rowIndex = rowIndex,
+                                    columnIndex = 1,
+                                    content = createValueContent(entry),
+                                    alignment = null,
+                                ),
+                            ),
+                    )
+                }
+
+            return TableBlock(header = null, rows = rows)
```
rock3r, line 70: I don't think front matter should emit `gfm-tables`' `TableBlock`. This makes one Markdown extension depend on another extension's model/renderer, which feels like a slippery slope towards dependency hell for consumers.

Semantically, front matter is not a GFM table; it just happens to be rendered table-like. Could this be a front-matter-specific `MarkdownBlock.CustomBlock` with its own renderer using `BasicTableLayout` instead?

The current shape also means adding a new front matter extension requires publicizing/generalizing the GFM table model API, which feels like the wrong dependency direction. If there's useful shared table rendering code, I'd rather extract that into a lower-level/shared layer than make extensions depend on each other.

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
```diff
@@ -0,0 +1,278 @@
...
+        val block = currentBlock as? CurrentBlock.KeyValue ?: return
+        val key = block.currentKey ?: return
+        val node = FrontMatterNode(key, block.currentValues.toList(), block.currentIsList)
+        fmBlock.appendChild(node)
+        block.currentKey = null
+        block.currentValues.clear()
+        block.currentIsList = false
+    }
+
+    override fun closeBlock() {
```
rock3r, line 218: Shouldn't we only emit a front matter block after we've seen the closing delimiter?

The PR description says front matter is delimited by `---` markers, but `closeBlock()` currently flushes the collected entries even if the parser stopped because it hit non-front-matter Markdown instead of a closing `---`. So a document like this would still render `title` as front matter:

```markdown
---
title: This was meant to be normal markdown
# Heading
```

Could we require the closing delimiter before rendering the collected metadata, and otherwise leave this as regular Markdown?

rock3r (changes requested):
@DanielSouzaBertoldi this will require a substantial rework to get in, can you do it plz? <3

DanielSouzaBertoldi:
Check the PR #3556 for the updated PR for this feature :)

PR #3466, opened 2026-03-20
[JEWEL-1290] DefaultBanner Should not Hardcode Padding

DanielSouzaBertoldi:
# Context

`DefaultBanner` has a default padding value of `10.dp`. This PR changes it to use whatever is set in `style.metrics.padding` instead. It also stops filling the entire available width and instead just wraps its content.

# Changes

- Changed `.padding(10.dp)` to `.padding(style.metrics.padding)` in the Row that wraps the banner content.
- Add `.width(IntrinsicSize.Max)` to the `Column` that wraps the banner content.

### ⚠️ Important Changes
 * `*DefaultBanner` no longer fills the available width by default; it now wraps its content instead. If you want banners to span the full width, be sure to add `modifier = Modifier.fillMaxWidth()`.

### Bug fixes
 * Paddings for `*DefaultBanner` components are now customizable

# Screenshots

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/DefaultBanner.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/DefaultBanner.kt
@@ -6,6 +6,7 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.RowScope
 import androidx.compose.foundation.layout.Spacer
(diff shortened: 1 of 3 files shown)


daaria-s (approved):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/DefaultBanner.kt
```diff
@@ -850,10 +851,10 @@ private fun DefaultBannerImpl(
     modifier: Modifier = Modifier,
     content: @Composable (() -> Unit),
 ) {
-    Column(modifier = modifier) {
+    Column(modifier = modifier.width(IntrinsicSize.Max)) {
```
rock3r, line 870: Could we document that users cannot use subcomposition inside the content slot because of this? `width(IntrinsicSize.Max)` asks the banner subtree for intrinsic measurements, so `SubcomposeLayout`-based composables (`BoxWithConstraints`, lazy layouts, etc.) can fail at runtime.

I reckon a custom `Layout` would be a better long-term solution, as it would allow us to support it properly, but I’m okay with that coming in a follow-up. For this PR, can we mention the restriction in the `Default*Banner` KDoc and release notes so users know which content is safe for now?

rock3r (changes requested):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/DefaultBanner.kt
DanielSouzaBertoldi: Fixed + [improvement filed](https://youtrack.jetbrains.com/issue/JEWEL-1312/Replace-IntrinsicSize.Max-in-DefaultBanner-with-a-Custom-Layout-to-Support-Subcomposition-based-Content)!

rock3r:
Looks good, we can ship after 3390 ships.

rock3r (approved):

nebojsa-vuksic (approved):

rock3r:
@DanielSouzaBertoldi same here, fix up then we can ship!

DanielSouzaBertoldi:
Why does CI hate me 😭

DanielSouzaBertoldi:
Ready to merge

PR #3471, opened 2026-03-25
[JEWEL-1075] Fix crash in ScrollableContainer when used with IntrinsicSize

DanielSouzaBertoldi:
(this description was generated by `jewel-pr-preparer`)

# Context

`VerticallyScrollableContainer` (and its horizontal counterpart) crashed with `IllegalArgumentException` when being used inside a Composable that expects `IntrincisSize`. This is the crash stacktrace:

```
Caused by: java.lang.IllegalArgumentException: Can't represent a width of 0 and height of 2147483647 in Constraints
	at androidx.compose.ui.unit.ConstraintsKt.throwInvalidConstraintException(Constraints.kt:401)
	at androidx.compose.ui.unit.ConstraintsKt.createConstraints(Constraints.kt:425)
	at androidx.compose.ui.unit.Constraints$Companion.fixedHeight-OenEA2s(Constraints.kt:244)
	at org.jetbrains.jewel.ui.component.ScrollableContainerKt$ScrollableContainerImpl$2$1.measure-3p2s80s(ScrollableContainer.kt:779)
	at androidx.compose.ui.layout.MeasurePolicy.maxIntrinsicHeight(MeasurePolicy.kt:164)
	at androidx.compose.ui.node.IntrinsicsPolicy.maxIntrinsicHeight(IntrinsicsPolicy.kt:53)
	at androidx.compose.ui.node.LayoutNode.maxIntrinsicHeight(LayoutNode.kt:764)
	at androidx.compose.ui.node.InnerNodeCoordinator.maxIntrinsicHeight(InnerNodeCoordinator.kt:140)
```

Compose's default `MeasurePolicy` intrinsic fallback re-runs the measure block with `Constraints.Infinity` on the queried axis. The old implementation used a SAM lambda for `Layout { }`, so the measure block would run during intrinsic queries. Inside that block, the scrollbar was being measured with `Constraints.fixedHeight(Int.MAX_VALUE)`, an illegal Constraints value that throws immediately.

# Changes                                                                                                                                                                                                      

- Changed the measurement order in `ScrollableContainer`: instead of measuring scrollbars with `Constraints.fixedHeight`/`Width(Int.MAX_VALUE)` to query their size, the implementation now queries scrollbar intrinsic sizes first by converting the Layout SAM lambda to an explicit `MeasurePolicy` object with proper intrinsic overrides, then measures content, then measures scrollbars with concrete dimensions.
- This alone was not enough; the scrollbar `Layout` lambdas in `Scrollbar.kt` were themselves using SAM lambdas with no intrinsic overrides as well, so Compose's default fallback would re-run their measure blocks with 
`Constraints.Infinity`, causing the same crash one level deeper. Both `verticalMeasurePolicy` and `horizontalMeasurePolicy` were converted to explicit `MeasurePolicy` objects that report `thumbThickness` as the intrinsic size on their fixed axis and 0 on the scroll axis.
- Added `check()` assertions (matching `LazyColumn`'s behavior) that throw `IllegalStateException` with a descriptive message when the scroll axis has an unbounded constraint during actual layout, giving developers clear guidance instead of a cryptic crash deep in Compose internals.                                                                                                                                
- Added comprehensive tests to `ScrollableContainerTest` covering: no crash in `IntrinsicSize.Max`/`IntrinsicSize.Min` parents, correct reported intrinsic dimensions (including scrollbar space for the macOS `AlwaysVisible` style), scrolling still works after the fix, and unbounded-axis nesting produces the expected IllegalStateException.

# Release notes

## Bug fixes

- `VerticallyScrollableContainer` and `HorizontallyScrollableContainer` no longer crash with `IllegalArgumentException` when placed inside a layout that queries intrinsic measurements.

### DIFF
--- a/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainerTest.kt
+++ b/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainerTest.kt
@@ -2,14 +2,20 @@
 package org.jetbrains.jewel.ui.component
 
 import androidx.compose.foundation.ScrollState
+import androidx.compose.foundation.horizontalScroll
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.requiredWidth
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
@@ -17,6 +23,8 @@ import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
@@ -27,10 +35,11 @@ import androidx.compose.ui.layout.Layout
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.test.assertHeightIsEqualTo
 import androidx.compose.ui.test.assertWidthIsEqualTo
-import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.junit4.v2.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.unit.Constraints
 import androidx.compose.ui.unit.dp
+import kotlin.test.assertFailsWith
 import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 import kotlin.time.Duration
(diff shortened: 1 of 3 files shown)


platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainerTest.kt
```diff
@@ -1479,6 +1488,215 @@ class ScrollableContainerTest {
...
+        rule.setContent {
+            val scrollState = rememberScrollState()
+            IntUiTheme {
+                Row(Modifier.height(IntrinsicSize.Max).width(300.dp)) {
+                    HorizontallyScrollableContainer(scrollState = scrollState) {
+                        Row { repeat(100) { Text("Item $it") } }
+                    }
+                }
+            }
+            assertTrue(scrollState.canScrollForward)
```
nebojsa-vuksic: This assertion runs inside the `setContent { }` composable lambda, i.e. during composition before measurement. `ScrollState`'s default `maxValue` is `Int.MAX_VALUE`, so `canScrollForward` (`value < maxValue`) is `true` regardless of whether scrolling actually works — the test passes vacuously and doesn't exercise the fix.

The vertical counterpart (`vertical container inside IntrinsicSize Max column ... still allows scrolling`) does it correctly: it hoists `val scrollState = ScrollState(0)` out of `setContent` and asserts after it returns. Suggest mirroring that here.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainer.kt
```diff
@@ -824,102 +829,218 @@ private fun ScrollableContainerImpl(
...
+                        }
+                    val height = contentPlaceable.height + hScrollbarHeight
+
+                    // Now that content is measured and the container dimensions are known (bounded
+                    // even if incomingConstraints were unbounded), measure the scrollbars with
+                    // concrete sizes.
+                    val verticalScrollbarPlaceable =
+                        if (accountForVerticalScrollbar) {
+                            val verticalScrollbarConstraints =
+                                Constraints.fixedHeight(height - sizeOffsetWhenBothVisible)
```
nebojsa-vuksic: Nit: the scrollbar is now measured against the container `height` (`contentPlaceable.height + hScrollbarHeight`) instead of the incoming viewport `maxHeight` as before. In the normal scrollable case these are equal, so no change — but when both scrollbars are visible and content is very small, `height - sizeOffsetWhenBothVisible` can go negative, and `Constraints.fixedHeight(negative)` throws. The old code used `maxHeight` here, which was safe from this. A `.coerceAtLeast(0)` guard here (and on the `fixedWidth` line below) would close that corner.

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainerTest.kt
```diff
@@ -1479,6 +1488,215 @@ class ScrollableContainerTest {
         rule.waitForIdle()
     }
 
+    @Test
+    fun `vertical container inside column with IntrinsicSize Max does not crash`() {
```
rock3r, line 1492: Could we add coverage for the `IntrinsicSize.Min` case too? The PR description says the tests cover both `IntrinsicSize.Max` and `IntrinsicSize.Min` parents, but the new tests only test the `Max` path.

Since the implementation now overrides both min and max intrinsic measurements, it would be good to have at least a small no-crash/size check for `Column(Modifier.width(IntrinsicSize.Min))` and the horizontal equivalent as well.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainer.kt
```diff
@@ -824,102 +829,218 @@ private fun ScrollableContainerImpl(
...
+                //   - HorizontallyScrollableContainer → height delegates to content + scrollbar
+                // Scroll-axis (must be bounded by the parent — no natural size):
+                //   - Always returns 0
+
+                override fun IntrinsicMeasureScope.maxIntrinsicWidth(
+                    measurables: List<IntrinsicMeasurable>,
+                    height: Int,
+                ): Int {
+                    val contentWidth = measurables[0].maxIntrinsicWidth(height)
+                    if (verticalScrollbar == null) return contentWidth
```
rock3r, line 993: Is this intentionally different from the non-intrinsic path? In "normal" layout we only reserve macOS `AlwaysVisible` scrollbar space when `verticalScrollbarVisible`/`scrollState.canScroll` is true, but the intrinsic path always accounts for the scrollbar width whenever the vertical scrollbar exists, even if it would not show (can't scroll).

That means an `IntrinsicSize` parent can reserve scrollbar space for non-scrollable content, even though the "normal" measurement wouldn't. If that's the intended trade-off because intrinsic queries can't reliably know if it's scrollable, could we document that here and lock it in the test?

rock3r (changes requested):

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainerTest.kt
DanielSouzaBertoldi: Ooof, that was an oversight on my part. THanks for pointing this out!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ScrollableContainer.kt
DanielSouzaBertoldi: Exactly what you said, it's an intentional trade-off here because intrinsic queries are meant to give a stable answer before measurement, so that we can tell exactly how much space to allocate.

If the override checked `verticalScrollbarVisible `, a `Column(Modifier.width(IntrinsicSize.Max))` containing a `VerticallyScrollableContainer` would change width as content crossed the scrollability threshold which is bad 😭 If we always reserve the track in the intrinsic path, then we don't risk any UI jumps

I'll add a comment making this design decision clearer

rock3r (approved):

nebojsa-vuksic (approved):

DanielSouzaBertoldi:
Ready to merge

PR #3475, opened 2026-04-01
[JEWEL-1298] Fix JDialogRenderer Density Issues

DanielSouzaBertoldi:
(this description was generated by `jewel-pr-preparer`)

`JDialogRenderer` had two density-related bugs affecting popup positioning and content rendering when the app overrides `LocalDensity`.

  ## Changes

  * Replace `LocalDensity.current` with `dialog.density()` (the actual screen scale from `graphicsConfiguration`) when converting Compose layout coordinates to AWT logical pixels in `withDensity` — the screen scale factor is always the correct ratio here, independent of any app-level density override.
  * Fix `ProvideValuesFromOtherContext` to provide the ComposePanel's defaults as the outer context and the caller's composition locals as the inner context, so the caller's values (including any `LocalDensity` override) correctly win over ComposePanel defaults instead of being silently discarded.

  ## Release notes

  ### Bug fixes

  * Fixed `JDialogRenderer` incorrectly using `LocalDensity` for AWT coordinate conversion, causing popups to appear at the wrong position or size when the app overrides the Compose density.
  * Fixed `ProvideValuesFromOtherContext` swallowing caller-provided composition locals (such as `LocalDensity`) in favor of `ComposePanel` defaults.

## Demo

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/5c6b2b07-9058-4e4c-b06c-695ac46e4f80" /> | <video src="https://github.com/user-attachments/assets/6884e6ba-64ea-43c5-9df4-ac2385f4c6d1" /> |


DanielSouzaBertoldi:
Yo, btw, why do we have both `kotlinter` and `ktfmt`?

We should really just stick to one of them, otherwise we get situations like in this PR, where `ktfmt` formats the code in a certain way that `kotlinter` does not like at all:

image

Even though this is pretty rare, it can happen.

I took a look at the `.editorconfig` file and most ktlinter rules are disabled anyways:

image

rock3r (approved):

rock3r:
@DanielSouzaBertoldi we can disable it if we want to, but it does check more than what ktfmt enforces IIRC

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -419,7 +421,7 @@ private fun Rectangle.withDensity(density: Float): Rectangle =
 @Composable
 private fun ProvideValuesFromOtherContext(context: CompositionLocalContext, content: @Composable () -> Unit) {
     val existingContext = currentCompositionLocalContext
-    CompositionLocalProvider(context) { CompositionLocalProvider(existingContext, content) }
+    CompositionLocalProvider(existingContext) { CompositionLocalProvider(context, content) }
```
daaria-s: I think I found an issue connected to this change, pls correct me if I'm wrong
If I enable Custom Popup Renderer and try to open **Sub menus**, it just disappears and i cannot open further submenus. Video is attached.
When I revert your change, it works fine.
After some discussion with AI, he told me that "a submenu is itself another popup, opened from inside the first popup, and for that nested popup to appear correctly, the submenu code must see the current popup host component, but because of the change it sees the original parent window component." Does it make any sense? :smile: 

https://github.com/user-attachments/assets/ebb38fa2-6ae7-43eb-a7c6-5eec1ba3fb2b

DanielSouzaBertoldi: Yeah, that's definitely a bug! Great catch 😄 

The problem is that we were providing the local compositions of the parent window to submenus.

The fix was pretty simple, use the local compositions from the parent but override the `LocalComponent` to be `this`, pointing to the current component itself, not the parent. With this change I realized that `ProvideValuesFromOtherContext` is kinda redundant now so I also removed it from the code (one less abstraction layer, hell yeah). Here's the result:

https://github.com/user-attachments/assets/fb810331-99d6-478c-a2e7-e3babef1a22c

Thank you so much for pointing this out! :)

DanielSouzaBertoldi:
@rock3r I disabled both `ktlint_standard_argument-list-wrapping` and `ktlint_standard_max-line-length`. Let ktfmt handle wrapping and detekt can check max line length :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -412,16 +410,6 @@ private fun Rectangle.withDensity(density: Float): Rectangle =
...
 
-/**
- * When inheriting from another context, we need to ensure that values already provided in the current context are not
- * overridden.
- */
-@Composable
-private fun ProvideValuesFromOtherContext(context: CompositionLocalContext, content: @Composable () -> Unit) {
-    val existingContext = currentCompositionLocalContext
-    CompositionLocalProvider(context) { CompositionLocalProvider(existingContext, content) }
-}
```
DanielSouzaBertoldi, line 423: This code was a bit redundant. The `context` parameter had its value assigned by `val compositionLocalContext by rememberUpdatedState(currentCompositionLocalContext)`

Then `existingContext` also fetched the information from `currentCompositionLocalContext`.

At the end of the day they were both providing the same values 😵‍💫

daaria-s (approved):

rock3r (approved):
Ready to merge

PR #3476, opened 2026-04-02
[JEWEL-1276] Fix popup focus-stealing in JDialogRenderer

rock3r:
Focusable popups rendered via `JDialogRenderer` steal key-window status on macOS, causing the parent window's title bar and traffic-light buttons to grey out as if the window were inactive. This happens because `focusableWindowState` is `true` when `setVisible(true)` is called, making the `JDialog` the macOS key window and deactivating the parent.

Notice how with this PR the semaphore icons remain active:

image

## Changes

* Apply the same `focusableWindowState` toggle technique used in IntelliJ's `AbstractPopup`/`DialogPopupWrapper`: set `focusableWindowState = false` and `isAutoRequestFocus = false` synchronously before `isVisible = true`, then restore `focusableWindowState = true` via `SwingUtilities.invokeLater`. The popup shows without becoming the key window, but can still accept focus on explicit interaction.
* Change the `AWTEventListener` keyboard gate from `dialog.isActive` to `dialog.isVisible`, so `onKeyEvent`/`onPreviewKeyEvent` callbacks fire without requiring the dialog to hold key-window status.

### Bug fixes

* Fixed focusable popups stealing key-window status on macOS, which caused the parent window title bar to appear inactive when a popup was shown via `JDialogRenderer`.

### DIFF
--- a/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
+++ b/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
@@ -318,7 +318,7 @@ private fun JPopupImpl(
                     }
                 }
                 is AWTKeyEvent -> {
-                    if (!dialog.isActive) return@AWTEventListener
+                    if (!dialog.isVisible) return@AWTEventListener
 
                     val composeEvent = event.toComposeKeyEvent()
 
@@ -344,9 +344,15 @@ private fun JPopupImpl(
     DisposableEffect(dialog) {
         dialog.contentPane = composePanel
 
-        dialog.isAutoRequestFocus = currentProperties.focusable
+        // JEWEL-1276 this weird code duplicates the JBPopup/AbstractPopup logic to avoid focus stealing on macOS.
+        // The actual hack is implemented in LocalPopupComponentFactory and uses the fact that macOS only sets the
+        // "key window" when a new window/popup is created. We don't want our popups to do it by default, so we
+        // start them as non-focusable, and then make them focusable a short while later.
+        dialog.focusableWindowState = false
+        dialog.isAutoRequestFocus = false
         dialog.isVisible = true
         dialog.size = composePanel.preferredSize
+        SwingUtilities.invokeLater { dialog.focusableWindowState = true }
 
         onDispose {
             dialog.isVisible = false


platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -344,9 +344,12 @@ private fun JPopupImpl(
     DisposableEffect(dialog) {
         dialog.contentPane = composePanel
 
-        dialog.isAutoRequestFocus = currentProperties.focusable
+        // JEWEL-1276 this duplicates the JBPopup/AbstractPopup logic to avoid focus stealing on macOS
+        dialog.focusableWindowState = false
+        dialog.isAutoRequestFocus = false
         dialog.isVisible = true
         dialog.size = composePanel.preferredSize
+        SwingUtilities.invokeLater { dialog.focusableWindowState = true }
```
DanielSouzaBertoldi, line 355: I mean, I get the solution but it still feels kinda weird to me lol

at the end of the day `focusableWindowState` is being set to true anyways, so I don't get how doing things this way actually stops macos from changing the key window to the popup 😵‍💫

DanielSouzaBertoldi (commented):
Besides not understanding the MacOS quirk at play here, it works and it's the way JB has been doing for a while now so that's fine by me 😎

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
rock3r: This patch stands on the shoulders of giants. It's a crazy hack that works and has worked for 20 years so I think it's fine lol

rock3r: A more reasoned response is that apparently the point is _when_ it's set. I reckon macOS uses only the INITIAL state of the flag to decide which one is the key window, so this effectively just says "yeah look this is not it" then macOS looks away and bam! We change it to the value we need, and macOS is none the wiser

DejanMilicic (approved):

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -344,9 +344,12 @@ private fun JPopupImpl(
     DisposableEffect(dialog) {
         dialog.contentPane = composePanel
 
-        dialog.isAutoRequestFocus = currentProperties.focusable
+        // JEWEL-1276 this duplicates the JBPopup/AbstractPopup logic to avoid focus stealing on macOS
```
AlexVanGogen: I tried to find what's the original and found [this](https://github.com/JetBrains/intellij-community/blob/98c2003fb1fa6428dc70e3e6efe7d611617eb5bd/platform/platform-impl/src/com/intellij/ui/popup/LocalPopupComponentFactory.kt#L87). @rock3r WDYT about mentioning it in the comment? Considering it does look hacky-ish

AlexVanGogen (approved):

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
rock3r: The comment mentions the entry points, from where you eventually end up in there. Is that not enough?

AlexVanGogen: Idk how easily discoverable it is if start from `JBPopup` :) I've found it by looking up usages

AlexVanGogen: I also thought at first that "duplicates" relates only to the two setters below and not to `invokeLater` thing. Might be a personal issue, though :)

rock3r: Improved the comment from your feedback :)

nebojsa-vuksic (approved):

rock3r:
Ready to merge

PR #3481, opened 2026-04-09
Make getMethodHandlesImplLookup Handle Exceptions

alonalbert:
On Android API < 26,:
```
debugProcess.findClass(evaluationContext, "java.lang.invoke.MethodHandles\$Lookup", null)
```

Will throw and this will make Evaluate Expression fail.

https://youtrack.jetbrains.com/issue/IDEA-388179/Evaluate-Expression-Throws-on-Android-API-25

---

> [!NOTE]
> **Low Risk**
> Change is localized to debugger evaluation utility code and only adds exception handling/fallback behavior; low risk aside from potentially masking unexpected class-loading errors.
> 
> **Overview**
> Prevents Evaluate Expression failures on older Android runtimes by making `getMethodHandlesImplLookup` tolerate `findClass` throwing (e.g., API < 26) and return `null` instead.
> 
> Callers already treat a `null` lookup as a helper-invocation failure, so evaluation now cleanly falls back instead of crashing when `MethodHandles$Lookup` is unavailable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c2b1596e4171febbd682313c0683e5a0db1242c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/java/debugger/impl/src/com/intellij/debugger/engine/MethodInvokeUtils.kt
+++ b/java/debugger/impl/src/com/intellij/debugger/engine/MethodInvokeUtils.kt
@@ -76,12 +76,12 @@ object MethodInvokeUtils {
   }
 
   fun getMethodHandlesImplLookup(evaluationContext: EvaluationContextImpl): ObjectReference? {
-    val theClass = evaluationContext.debugProcess.findClass(evaluationContext,
-                                                            "java.lang.invoke.MethodHandles\$Lookup",
-                                                            null)
-    val theField = DebuggerUtils.findField(theClass,
-                                           "IMPL_LOOKUP")
-    return theClass?.getValue(theField) as? ObjectReference
+    val theClass = runCatching {
+      // On Android API < 26, this will throw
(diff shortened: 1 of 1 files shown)


mvicsokolova:
Thank you for the fix, the old behaviour was actually incorrect 👍🏻. I've provided more error logging to `com.intellij.debugger.engine.MethodInvokeUtils#getMethodHandlesImplLookup` for both cases when the class or the field value cannot be found, and I would push the whole change to master myself, if that's ok

gorrus:
Also, it makes sense to completely disable the invocation helper for Android with registry `debugger.evaluate.method.helper`=off.

alonalbert:
Why disable for Android completely? It seems to work with API >=26

mvicsokolova:
The fix was merged to master: [IDEA-388179](https://youtrack.jetbrains.com/issue/IDEA-388179)

gorrus:
> Why disable for Android completely? It seems to work with API >=26

we have never tested it on android

PR #3491, opened 2026-04-23
[JEWEL-1307] Circumvent crash when selecting Tooltip (and Text) inside a SelectionContainer

DanielSouzaBertoldi:
# Context
There's a crash in Jetpack Compose that happens where gesture being translated from text coordinates to the selection container coordinates don't share the same root node, which is the case for our tooltip and popups in general.

Filed issues about this bug:

https://youtrack.jetbrains.com/projects/CMP/issues/CMP-9161
https://issuetracker.google.com/issues/300781578

# Fix
- Wrap the tooltip in a `DisabledSelection` as a workaround.
- Also removed the internal `FlagAwareTooltipArea` in favor of using `JewelTooltipArea` directly.

# Screenshots/Screen Recordings

## Tooltip
| Scenario | Screen Recording |
|--------|--------|
| Crash | <video src="https://github.com/user-attachments/assets/4a57510f-2076-4b09-9725-f5dcdd183b0f" /> |
| Compose Popup | <video src="https://github.com/user-attachments/assets/bd622015-75a0-4c79-a74f-cf9eb0fff362" /> |
| Jewel Popup | <video src="https://github.com/user-attachments/assets/67232132-6cea-4745-9d96-60c2dc1ac4b1" /> |

The main use case where the crash can most commonly happen is this one, using tooltips. However, it is technically viable to have a `SelectionContainer` wrapping other components that do open a popup, like ComboBoxes and Menus. Why would you ever do such thing I've no idea, but here's proof that they are not crashing:

---

## Menu
| Scenario | Screen Recording |
|--------|--------|
| Crash | <video src="https://github.com/user-attachments/assets/77b329e9-cb34-4217-8a11-5364d012f335" /> |
| Compose Popup | <video src="https://github.com/user-attachments/assets/3f5e83dc-4d6e-4635-9975-ac0c44e82fe8" /> |
| Jewel Popup | <video src="https://github.com/user-attachments/assets/624f45a1-75d0-4466-83c4-fe3e36c89c8f" /> |

## ComboBox
| Scenario | Screen Recording |
|--------|--------|
| Crash | <video src="https://github.com/user-attachments/assets/2fbda99a-c104-45e9-8abb-410d2f29d264" /> |
| Compose Popup | <video src="https://github.com/user-attachments/assets/168ff99b-c730-43e1-be3e-a242ab02aaae" /> |
| Jewel Popup | <video src="https://github.com/user-attachments/assets/9a973ebd-2f5a-4133-b826-74c2f18f8695" /> |

## SpeedSearch
Funnily enough I wasn't able to reproduce the crash in SpeedSearch. Tried with both our native popup and Compose's but no luck.

There seems to be a bug with the selection though, it stops working after the speed search popup is rendered for the first time.

Anywho, have fun watching be try to force the crash using our native popup for 37 seconds:

https://github.com/user-attachments/assets/1dea946e-5e74-4b97-8d5b-6e27cd159184

### DIFF
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TooltipArea.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TooltipArea.kt
@@ -2,7 +2,6 @@
 package org.jetbrains.jewel.ui.component
 
 import androidx.compose.foundation.ExperimentalFoundationApi
-import androidx.compose.foundation.TooltipArea as ComposeTooltipArea
 import androidx.compose.foundation.TooltipPlacement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.shape.CornerSize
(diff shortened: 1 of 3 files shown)


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TooltipArea.kt
```diff
@@ -29,55 +28,10 @@ import androidx.compose.ui.unit.dp
...
- * @param tooltip Composable content of the tooltip.
- * @param modifier The modifier to be applied to the layout.
- * @param delayMillis Delay in milliseconds.
- * @param tooltipPlacement Defines the position of the tooltip.
- * @param content Composable content that the current tooltip is set to.
- * @see org.jetbrains.jewel.ui.component.Popup
- * @see androidx.compose.ui.window.Popup
- */
-@Composable
-internal fun FlagAwareTooltipArea(
```
DanielSouzaBertoldi, line 50: Honestly we don't even need this composable.

`Popup` already has the same `JewelFlags` check, so we can safely remove this guy.

rock3r:
> There seems to be a bug with the selection though, it stops working after the speed search popup is rendered for the first time

LOL have fun figuring this out XD But the text selection not working is I think because the items that the Text is inside are clickable, and that may be intercepting the events to use for click? Who knows.

rock3r (approved):
Nice one, thanks!

DanielSouzaBertoldi:
@rock3r that might be a possible cause, I'll file an bug for this on YT so we don't forget to take a look someday

daaria-s (approved):

rock3r:
Ready to merge

nebojsa-vuksic (approved):

PR #3493, opened 2026-04-27
Fix Error vs Failure Reporting

alonalbert:
Reverse conditions

IDEA-388810

### DIFF
--- a/platform/testFramework/bootstrap/src/com/intellij/tests/bazel/TestData.java
+++ b/platform/testFramework/bootstrap/src/com/intellij/tests/bazel/TestData.java
@@ -88,7 +88,7 @@ class TestData {
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


platform/testFramework/bootstrap/src/com/intellij/tests/bazel/TestData.java
```diff
@@ -102,7 +102,7 @@ public boolean isFailure() {
       return true;
     }
```
k15tfu, line 103: Should we also move this to `TestData.isSkipped()`, the same way it was done in this commit https://github.com/bazel-contrib/rules_jvm/commit/df44bd92954be8fdc5b8b7cdf323ff98cb62baea?

shalupov: I'll do that

k15tfu (approved):

PR #3495, opened 2026-04-29
[JEWEL-1309] Bump detekt from v1 to v2.0.0-alpha.3

DanielSouzaBertoldi:
# Context

In order to bump CMP to 1.11.0-beta03, we have to first bump Detekt to 2.0.0-alpha.3. This PR does exactly that, fixes all migration issues and updates gradle/IML/Bazel files accordingly.

# The Surprise

With this update, we cannot run tests from `detekt-plugin` module through JPS anymore. The reason for that is because Detekt 2.0 updated its list of transitive dependencies which result in a clash of missing APIs from either IJP or `kotlin-compiler` directly 😬 Here's a more detailed explanation:

Detekt v1 used to depend on kotlin-compiler-embeddable ([check POM here](https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/detekt-api/1.23.8/detekt-api-1.23.8.pom))

v2 ditched the embeddable version for the "slim" kotlin-compiler ([check POM here](https://repo1.maven.org/maven2/dev/detekt/detekt-api/2.0.0-alpha.3/detekt-api-2.0.0-alpha.3.pom))

This seemingly innocent change screws up our test suite for custom detekt rules, and we start getting this exception:

Error occurred during initialization of VM
java.lang.Error: com.intellij.util.lang.UrlClassLoader.<init>(java.lang.ClassLoader)
	at java.lang.ClassLoader.initSystemClassLoader(java.base@25.0.2/ClassLoader.java:1901)
	at java.lang.System.initPhase3(java.base@25.0.2/System.java:1988)
Caused by: java.lang.NoSuchMethodException: com.intellij.util.lang.UrlClassLoader.<init>(java.lang.ClassLoader)
	at java.lang.Class.getConstructor0(java.base@25.0.2/Class.java:3187)
	at java.lang.Class.getDeclaredConstructor(java.base@25.0.2/Class.java:2491)
	at java.lang.ClassLoader.initSystemClassLoader(java.base@25.0.2/ClassLoader.java:1888)
	at java.lang.System.initPhase3(java.base@25.0.2/System.java:1988)

From what I've gathered, that's because `kotlin-compiler` bundles its own `UrlClassLoader` without a `ClassLoader` constructor. When this class ends up in the classpath, anything trying to instantiate `UrlClassLoader(parent)`  blows up. Apparently JPS needs this specific `UrlClassLoader` to start the test JVM.

[...]

PR #3496, opened 2026-04-30
[JEWEL-1225] Bump CMP version to 1.11.0

DanielSouzaBertoldi:
> [!IMPORTANT]
> This is a direct continuation of #3495 !  

# Context

CMP 1.11.0 just dropped and we now have access to fixes and new features! Hell yeah.

# Changes
- Removed references to `Key.Home` since it got removed from the API surface
- Changed `Cursor` imports from `org.jetbrains.skiko.Cursor` to `java.awt.Cursor` which is the correct type expected for `PointerInput`
- `ContextMenuItem` now has its own `enabled` property. We don't need to declare our own variable for that for `ContextMenuItemOption`
- Default focus clear behavior changed. Now focus is cleared automatically when clicking a non-focusable element. Our `EditableComboBox` had to be changed to reflect this.

### ⚠️ Important Changes
 * Jewel now targets CMP 1.11.0, which comes with its [own set of breaking changes](https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.11.0). This means that from Jewel 0.38+, it'll be transitively binary-incompatible with artifacts compiled against CMP 1.10.x and lower.


platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Borders.kt
```diff
@@ -39,7 +40,6 @@ import org.jetbrains.jewel.ui.icons.AllIconsKeys
 import org.jetbrains.jewel.ui.outline
 import org.jetbrains.jewel.ui.theme.colorPalette
 import org.jetbrains.jewel.ui.typography
-import org.jetbrains.skiko.Cursor
```
DanielSouzaBertoldi, line 42: `PointerIcon` now expects a `java.awt.Cursor`

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
```diff
@@ -155,10 +155,10 @@ public class ContextSubmenu(label: String, public val submenu: () -> List<Contex
 public class ContextMenuItemOption(
     public val icon: IconKey? = null,
     public val actionType: ContextMenuItemOptionAction? = null,
-    public val enabled: Boolean = true,
```
DanielSouzaBertoldi, line 158: `androidx.compose.foundation.ContextMenuItem` now has its own `enabled` property:

```kt
// Avoid breaking backwards compatibility with subclasses that already have an `enabled` field
private val _enabled = enabled
val enabled: Boolean
    @JvmName("contextMenuItemBaseClassEnabled") get() = _enabled
```

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
```diff
@@ -148,13 +148,7 @@ public fun EditableComboBox(
     ) {
         CompositionLocalProvider(LocalContentColor provides style.colors.contentFor(comboBoxState).value) {
             Row(
-                modifier =
-                    Modifier.onFocusChanged {
-                        comboBoxState = comboBoxState.copy(focused = it.isFocused)
-                        if (!it.isFocused) {
-                            popupManager.setPopupVisible(false)
-                        }
-                    },
```
DanielSouzaBertoldi, line 157: ~~This one was interesting. With CMP 1.11, the behavior for popups mirrors Android's Compose.~~

~~By default, now focus is cleared when clicking a non-focusable element. There's even a feature flag that we can set to `false` fo fallback to the 1.10 behavior: `ComposeUiFlags.isClearFocusOnMouseDownEnabled`.~~

~~For us, this means that we don't really need to call `setPopupVisibile` anymore since CMP does that for us now.~~

~~If you want, you can [read the discussion on YT](https://youtrack.jetbrains.com/issue/CMP-10127) about this new behavior!~~

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextContextMenu.kt
```diff
@@ -38,7 +38,7 @@ public object TextContextMenu : TextContextMenu {
                                 actionType = CutMenuItemOptionAction,
                                 enabled = textManager.selectedText.isNotEmpty(),
                                 label = localization.cut,
-                                action = it,
+                                action = it.execute,
```
DanielSouzaBertoldi, line 41: `TextManager` properties now return an:

```kt
class Action(val enabled: Boolean, val execute: () -> Unit)
```

Instead of just a lambda, which now resides in `execute`

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/Keybindings.kt
```diff
@@ -89,10 +89,10 @@ public open class DefaultSelectableColumnKeybindings : SelectableColumnKeybindin
         get() = isCtrlPressed
 
     override val KeyEvent.isSelectFirstItem: Boolean
-        get() = (key == Key.MoveHome || key == Key.Home) && !isContiguousSelectionKeyPressed
```
DanielSouzaBertoldi, line 92: > 'val Home: Key' is deprecated. `Key.Home` was mapped to the keyboard "Home" key in error. It is meant to be the "system" home key on Android, and should never be delivered to applications. For the keyboard "Home" key use `Key.MoveHome`. For the Android system "Home" key (unlikely to be needed), use `Key.SystemHome`.

platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/DecoratedWindow.kt
```diff
@@ -95,6 +95,7 @@ public fun DecoratedWindow(
         var decoratedWindowState by remember { mutableStateOf(DecoratedWindowState.of(window)) }
 
         DisposableEffect(window) {
+            window.isClearFocusOnMouseDownEnabled = false
```
DanielSouzaBertoldi, line 98: With CMP 1.11, the behavior for popups mirrors Android's Compose.

By default, now focus is cleared when clicking a non-focusable element.

This, however, breaks a [number of our composables + our custom modifiers](https://youtrack.jetbrains.com/issue/CMP-10127/Opening-a-Popup-with-focusable-false-incorrectly-clears-Compose-focus-from-the-parent-container#focus=Comments-27-13714978.0-0) that track a window/panel activation. Given the nature of this feature, the safest and simplest option for us is to simply disable the feature altogether.

You can click the link above to read the entire discussion about this :)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/EditableComboBox.kt
DanielSouzaBertoldi: Disregard everything in this comment, we had to undo this.

You can find new information reading this comment [just a bit below in this page](https://github.com/JetBrains/intellij-community/pull/3496#discussion_r3182501056).

platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/DecoratedWindow.kt
rock3r: Looks like we should be doing this in the other standalone surfaces, too. This covers `DecoratedWindow`, and `JewelComposePanelWrapper` covers the bridge helper, but the plain standalone `Window` / `singleWindowApplication` wrappers and the raw `ComposePanel` in `JDialogRenderer` still inherit CMP’s new default.

If clear-focus-on-mouse-down breaks Jewel activation/popup tracking, should we disable it there as well? Also, does this need to be done more centrally in the bridge too? Would that just be done at the `JewelComposePanel` level, so it applies everywhere?

platform/jewel/gradle/libs.versions.toml
```diff
@@ -2,8 +2,8 @@
 assertJ = "3.26.3"
 coil = "3.2.0"
 commonmark = "0.24.0"
-composeDesktop = "1.10.0"
-detekt = "1.23.8"
+composeDesktop = "1.11.0-beta03"
```
rock3r: This was obviously right when the PR was opened, but now that CMP 1.11.0 final is out, can we bump this from `1.11.0-beta03` to `1.11.0` before merging? That keeps the published Jewel dependency and the migration/release-note messaging aligned with the stable CMP release.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
rock3r: Looks like this removes `ContextMenuItemOption.getEnabled()` from the subclass ABI. This is exactly what the comment in the snippet you posted here says the JvmName is there to avoid, yet we are not doing the same.

Source still sees `enabled` via `ContextMenuItem`, but the upstream getter is intentionally named `contextMenuItemBaseClassEnabled`, so it won’t satisfy binaries compiled against Jewel’s `ContextMenuItemOption.enabled`.

Can we keep `public val enabled` here and pass it to the superclass? That should preserve the old getter while still using CMP’s new constructor.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextContextMenu.kt
rock3r: Should we pass through `it.enabled` here too? CMP moved the action availability into `TextContextMenu.Action`; this unwraps `execute`, but Jewel still recomputes enabled state for Cut/Copy and leaves Paste/Select All default-enabled.

That can ignore upstream policy for cases like password fields, read-only fields, empty clipboard, or already-selected-all text. Could we use `enabled = it.enabled` for all four actions?

rock3r (changes requested):
Could we add PR release notes before this lands? This isn’t just an internal dependency bump: CMP 1.11 has breaking changes, and Jewel 0.38+ will be transitively binary-incompatible with artifacts compiled against CMP 1.10.x and lower. It’d be useful to link https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.11.0 and call out any other user-visible transitive behavior changes relevant to Jewel users.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
DanielSouzaBertoldi: Unfortunately our case is a bit more complex than just keeping `public val enabled` :/

See, our `enabled` variable can't be shadowed without `override`, but it also can't be overridden because CMP's `enabled` is `final`.

The only way that I can think of to avoid binary breaking changes is to add a DeprecationLevel.Hidden property with a different name but `@get:JvmName("getEnabled")`.

What do you think?

platform/jewel/decorated-window/src/main/kotlin/org/jetbrains/jewel/window/DecoratedWindow.kt
DanielSouzaBertoldi: Ooops, missed those. My bad! Yes, we really should disable this new behavior in every single place that provides jewel. On brigde, we can just disable it in the `init` block in `JewelComposePanelWrapper`. Thanks for pointing this out!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/ContextMenu.kt
rock3r: Seems reasonable yeah

platform/jewel/ui/metalava/ui-api-0.37.0.txt
```diff
@@ -345,10 +345,9 @@ package org.jetbrains.jewel.ui.component {
   public final class ContextMenuItemOption extends androidx.compose.foundation.ContextMenuItem {
     ctor public ContextMenuItemOption(optional org.jetbrains.jewel.ui.icon.IconKey? icon, optional org.jetbrains.jewel.ui.component.ContextMenuItemOptionAction? actionType, optional boolean enabled, String label, kotlin.jvm.functions.Function0<kotlin.Unit> action);
     method public org.jetbrains.jewel.ui.component.ContextMenuItemOptionAction? getActionType();
-    method public boolean getEnabled();
```
DanielSouzaBertoldi, line 348: Our current Metalava setup doesn't track JVM bytecode, that's why it thinks this function has been removed but good, but that's not the case.

Also, metalava started complaining about a bunch of unrelated changes, so I ignored those for this PR. Bear in mind that the Annotate Metalava changes job will fail though

platform/jewel/ui/metalava/ui-api-0.37.0.txt
```diff
@@ -345,10 +345,9 @@ package org.jetbrains.jewel.ui.component {
...
     method public org.jetbrains.jewel.ui.icon.IconKey? getIcon();
     property public org.jetbrains.jewel.ui.component.ContextMenuItemOptionAction? actionType;
-    property public boolean enabled;
+    property @Deprecated public boolean enabledCompat;
```
DanielSouzaBertoldi, line 350: Metalava only tracks the Kotlin property name, not the renamed JVM getter.

platform/jewel/ui/metalava/ui-api-0.37.0.txt
rock3r: 🐻

rock3r (approved):

nebojsa-vuksic (approved):

PR #3498, opened 2026-05-06
Fix: Address potential deadlock in DirectoryLock (IJPL-179219)

vishalga-google:
This pull request enhances the reliability of the IDE startup process, especially during updates or rapid restarts, by making DirectoryLock.java more tolerant of lingering processes.

**Problem:**

Currently, if a previous IDE process is still in the process of shutting down when a new instance is launched, the directory lock check can fail immediately, leading to a "Start Failed" dialog. This can occur even if the old process would have exited moments later, causing unnecessary interruption.

**Solution:**

This change introduces a more graceful handling mechanism in `DirectoryLock`:

**1.  Active Waiting for Process Exit:** If a process ID from a previous run is found in the lock file, the system now checks if this process is still active. If it is, the new instance will wait up to 1000 milliseconds for the existing process to terminate, using `ProcessHandle.onExit().get()`. This allows a short window for the old process to clean up and release locks.

**2. Enhanced Retry Mechanism:** The lock acquisition logic now attempts up to 3 times:

**- Attempt 1:** Checks the lock. If locked by an active process, waits for it to exit as described above. If the process exits within the timeout, it proceeds to the next attempt to give time for file cleanup. If the process _fails_ to exit within the timeout, startup is aborted to prevent deadlocks.

**- Attempts 2 & 3:** These attempts include short 200ms pauses. This helps in cases where the process has exited, but the operating system hasn't fully released file locks or synchronized filesystem changes, particularly on Windows.

**3. Stale Lock File Cleanup:** As a fallback, if the process indicated by the lock file is confirmed to be no longer running, but the lock files still exist, they are forcibly removed to allow startup to proceed.

This refined approach aims to:
- Add no overhead in the normal case (no conflicting process).
- Successfully start the IDE when a previous instance is slightly slow to shut down.
- Recover from crashes where stale lock files were left behind.
- Still fail relatively quickly (~1 second) if a process is genuinely unresponsive.

**Changes:**

- `platform/platform-impl/bootstrap/src/com/intellij/platform/ide/bootstrap/DirectoryLock.java:`
1. Increased the main loop iterations from 1 to 3.
2. Integrated `ProcessHandle` to detect and wait for the existing process to exit.
3. Added clear logging regarding the waiting process.
4. Included specific handling for `TimeoutException` and `InterruptedException` during the wait.

- `platform/platform-tests/testSrc/com/intellij/platform/ide/bootstrap/DirectoryLockTest.java:`
1. Added `activatingWithDelayedExitingProcess()` test: Simulates a scenario where a process holds the lock and takes some time to exit. Verifies that the lock is eventually acquired.
2. Added `activatingWithHungProcessTimesOut`() test: Simulates a scenario where the process holding the lock never exits. Verifies that the startup attempt times out and throws `CannotActivateException`.

Testing:

The new unit tests in `DirectoryLockTest.java` validate the new logic for handling delayed exits and unresponsive processes.

---

> [!NOTE]
> **Medium Risk**
> Changes IDE startup locking/activation behavior by adding retries and a timed wait on a competing process, which could affect startup timing and edge-case lock acquisition failures.
> 
> **Overview**
> Makes `DirectoryLock.lockOrActivate()` more tolerant of rapid restarts by retrying lock acquisition (up to 3 attempts) and, when the lock file points to an IDE process with the same command, waiting up to 1s for that process to exit before failing.
> 
> Adds new tests that spawn a dummy Java process to verify startup succeeds when the competing process exits shortly after, and still throws `CannotActivateException` when the competitor remains hung past the timeout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35621bdac81b03ce063a5f22fa933e8c067c1857. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/platform/platform-impl/bootstrap/src/com/intellij/platform/ide/bootstrap/DirectoryLock.java
+++ b/platform/platform-impl/bootstrap/src/com/intellij/platform/ide/bootstrap/DirectoryLock.java
@@ -175,7 +175,7 @@ public final class DirectoryLock {
       var command = ProcessHandle.current().info().command().orElse("???");
       LOG.debug("current command: " + command);
 
-      for (int attempt = 0; attempt < 1; attempt++) {
+      for (int attempt = 0; attempt < 3; attempt++) {
         try {
           return tryListen();
         }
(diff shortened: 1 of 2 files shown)


trespasserw:
I have removed the retry loop not just because it was incorrectly implemented, rather because repeated deleting port and lock files makes the state more complicated (it is possible to delete an active UDS file, but `bind()` attempts will fail). Besides, delays increase the "dark" (no-UI) startup time, provoking users to start more instances.

But waiting for a competing process to quit does make sense indeed. I have add this logic and the tests along with some smaller tweaks.

After some time, I'll try to compare reports from 261 and 262, to see if there's any effect.

PR #3503, opened 2026-05-08
[JEWEL-1313] Add syntax highlighting to FencedCodeBlock in standalone

DanielSouzaBertoldi:
### Context
So you know when you are coding an awesome chat plugin or a plugin that kinda acts like a tutorial for beginner programmers (heck, or even for "advanced" programmers for that matter) and then you add Jewel's markdown to your plugin and then you add a code block to your markdown file/string and you see how beautiful and elegant your plugin looks highlighting all these cool keywords and patterns from that awesome programming language you love and care for deeply and you are so grateful that without realizing, a voice in the back of you head says something weird like "thanks Jewel team, this is awesome"?

And then you feel so inspired that now you actually want to break out of IntelliJ's boundaries and go all in on also creating a desktop app that users can use whenever they want, but then you go to test your markdown file and... Where's all the awesome syntax highlighting? Why does everything look so.... bland? Why... why does it have to be like this? Why can't the standalone version of Jewel be as awesome as the bridge version? Are the devs that lazy? Why so many questions?!

You go through all the idk how many stages of grief until you accept that, yes, maybe standalone is not as cool as bridge, or maybe the devs are indeed lazy, and your inspiration dies right there and then. Maybe your idea was meant to be limited to the IDE after all... you lose all hope and some days you don't even feel like continuing development because... what's the point anymore?

_WELL_, let me tell you that your concern has **FINALLY** been addressed!

With this PR, the code blocks in your markdown can look as they should: beautiful, awesome, and dare I say it – sexy. ooh lala!

Hell yeah now THAT's what I'm talking about. The feud between standalone and bridge is about to be over ~until the next big feature~ 😎 

If that's not enough to convince you, click the collapsible section below to read the boring AI version of everything I said so far:

<details><summary>Boring AI summary zzzzzz</summary>
<p>

`SimpleCodeHighlighter` is a lightweight, dependency-free `CodeHighlighter` for standalone **(non-IDE)** Jewel apps. Until now, the only built-in option was `NoOpCodeHighlighter`, meaning fenced code blocks in rendered Markdown were displayed as plain text. This PR adds a regex-based highlighter that works without any IntelliJ Platform dependencies and wires it up as the default in `ProvideMarkdownStyling`.

</p>
</details> 

### Changes
- Add `SimpleCodeHighlighter` — a `CodeHighlighter` implementation driven by an ordered list of regex `TokenRules`. Rules are matched by earliest-start position; ties go to rule order. The highlighter emits a single `AnnotatedString` synchronously.
- Add `LanguageGrammar` — a named set of `TokenRules` matched case-insensitively against the fenced code block language tag.
- Add `TokenRule` and `TokenType` — the rule/token primitives. `TokenRule.Companion` has factory functions for the common token types (`keyword`, `string`, `comment`, `type`, `constant`, `number`, `builtin`, `functionCall`, `functionDeclaration`, `typeDeclaration`).
- Add `SyntaxHighlightColors` — a token color palette with `light()` and `dark()` presets matching IntelliJ's `expUI light` and `expUI dark` editor schemes.
- Add `BuiltInLanguageGrammars` with built-in grammars for Kotlin and Java, ported from the IntelliJ TextMate bundles.
- Change `ProvideMarkdownStyling` default `codeHighlighter` from `NoOpCodeHighlighter` to `SimpleCodeHighlighter` (theme-aware via `isDark`). The explicit-codeHighlighter overload remains unchanged.
- Add developer docs in `docs/standalone-code-highlighting.md` covering the rule engine, grammar authoring, factory functions, and regex engine limitations.
- Add unit tests for `SimpleCodeHighlighter`, `TokenRule`, and the built-in Kotlin/Java grammars.
- Add `coroutines-core` and `coroutines-test` test dependencies to `int-ui-standalone-tests`.

**Disclaimer**: this is just a "kickoff" kind of PR. I added just Kotlin and Java because these languages are GOATED (I added Java more out of respect otherwise we wouldn't have Kotlin today) but the idea is that, once this PR is merged, we can gradually add more and more languages

### ⚠️ Important Changes
 * `ProvideMarkdownStyling` now defaults to `SimpleCodeHighlighter` instead of `NoOpCodeHighlighter`. Fenced code blocks in rendered Markdown are highlighted automatically in standalone apps. Pass an explicit `codeHighlighter` to opt out or substitute a different implementation.
                      
### New features
  * New SimpleCodeHighlighter in int-ui-standalone — a dependency-free, regex-based CodeHighlighter for standalone apps. Supports custom grammars via additionalGrammars (searched before built-ins, so they can override built-in behavior). Built-in grammars included for Kotlin and Java.

### Screen recordings

| Title | Video |
|--------|--------|
| Java/Kotlin syntax highlighting | <video src="https://github.com/user-attachments/assets/537f7677-e8c2-4970-a8c2-ce9163f38d81" /> |
| Epic feature announcement video | <video src="https://github.com/user-attachments/assets/addb48f0-842a-4714-b978-4c81f6ba1d65" /> |

#### Light mode
| Language | GitHub (light default) | Jewel Bridge (islands light) | Jewel Standalone |
| Kotlin | image | image | image |
| Java | image | image | image |

#### Dark mode
| Language | GitHub (dark default) | Jewel Bridge (Islands Dark) | Jewel Standalone |
| Kotlin | image | image | image |
| Java | image | image | image |


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -660,7 +660,9 @@ public open class DefaultMarkdownBlockRenderer(
     ) {
         val content = block.content
         val highlighter = LocalCodeHighlighter.current
-        val highlightedCode by highlighter.highlight(content, mimeType).collectAsState(AnnotatedString(content))
+        val highlightedCode by remember(content, mimeType, highlighter) {
```
DanielSouzaBertoldi: Pretty simple optimization: this way, we avoid having to recompose code blocks that didn't have its contents changed in any way :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/JavaLanguageGrammar.kt
```diff
@@ -0,0 +1,44 @@
...
+                // Comments must come first
+                TokenRule.comment("/\\*[\\s\\S]*?\\*/"),
+                TokenRule.comment("//[^\n]*"),
+                // Strings (text blocks first — Java 13+)
+                TokenRule.string("\"\"\"[\\s\\S]*?\"\"\""),
+                TokenRule.string("\"(?:[^\"\\\\]|\\\\.)*\""),
+                TokenRule.string("'(?:[^'\\\\]|\\\\.)*'"),
+                // entity.name.function.java — identifier immediately before (
+                // Must precede keyword rules so method names are colored before keywords match
+                TokenRule.functionCall("\\b([A-Za-z_][A-Za-z0-9_]*)\\s*(?=\\()"),
```
nebojsa-vuksic: Thanks for this, the rule engine reads really nicely! I think there might be a small ordering issue here, though. Since the `functionCall` rule sits above the keyword rules and the tokenizer breaks earliest_start ties by rule order, control flow keywords that are followed by `(` end up matching `functionCall` first and get colored as function calls rather than keywords.

I ran a faithful port of the tokenizer with these exact patterns to double check, and saw:

* `if (x) {}` colors `if` as FUNCTION_CALL
* `while (x) {}` colors `while` as FUNCTION_CALL
* `switch (x) {}` colors `switch` as FUNCTION_CALL
* `catch (E e) {}` colors `catch` as FUNCTION_CALL

The Kotlin grammar already avoids this by placing `functionCall` after the keyword rules (it even has a dedicated `control keywords before parenthesis are not colored as function calls` test), and that ordering also matches the table in `standalone-code-highlighting.md`. Would it make sense to move this rule below the keyword, constant, and type rules to mirror Kotlin? I confirmed that `void main()` and `println()` still resolve to FUNCTION_CALL after the move, so real declarations and calls should be unaffected. A small regression test such as `highlight("if (x) {}")` asserting that `if` stays a keyword would be a nice safety net.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/KotlinLanguageGrammar.kt
```diff
@@ -0,0 +1,51 @@
...
+                TokenRule.comment("/\\*[\\s\\S]*?\\*/"),
+                TokenRule.comment("//[^\n]*"),
+                // Strings
+                TokenRule.string("\"\"\"[\\s\\S]*?\"\"\""),
+                TokenRule.string("\"(?:[^\"\\\\]|\\\\.)*\""),
+                TokenRule.string("'(?:[^'\\\\]|\\\\.)*'"),
+                // fun <name> — entity.name.function.kotlin
+                TokenRule.functionDeclaration("\\b(fun)\\s+([A-Za-z_][A-Za-z0-9_]*)"),
+                // class/interface/object/enum <name> — entity.name.type.class.kotlin
+                TokenRule.typeDeclaration("\\b(class|interface|object|enum)\\s+([A-Za-z_][A-Za-z0-9_]*)"),
```
nebojsa-vuksic: One edge case I think is worth handling: `enum class X` seems to be mis-highlighted by this rule. Because `enum` is one of the group 1 alternatives and `class` matches the group 2 identifier shape, `typeDeclaration` ends up consuming `enum class` before the keyword rule gets a chance to run.

I verified this with a faithful port of the tokenizer using this exact pattern:

* `enum class Foo` colors `enum` as KEYWORD, `class` as BUILTIN (the type name color), and leaves `Foo` unstyled
* for comparison, `class MyClass` is correct: `class` is KEYWORD and `MyClass` is BUILTIN

Since `enum class` is the only way to declare an enum in Kotlin, it is a fairly common construct, and it looks like it is currently untested (the suite covers `class MyClass` only). Would it be worth excluding the declaration starters from the group 2 identifier, or handling the `enum class` pair explicitly, and adding an `enum class Foo` test? Happy to help if useful.

platform/jewel/markdown/int-ui-standalone-styling/src/main/kotlin/org/jetbrains/jewel/intui/markdown/standalone/IntUiProvideMarkdownStyling.kt
```diff
@@ -119,7 +126,7 @@ public fun ProvideMarkdownStyling(
     imageSourceResolver: ImageSourceResolver,
     markdownStyling: MarkdownStyling,
     markdownBlockRenderer: MarkdownBlockRenderer,
-    codeHighlighter: CodeHighlighter,
+    codeHighlighter: CodeHighlighter = SimpleCodeHighlighter(SyntaxHighlightColors.light()),
```
nebojsa-vuksic: Small thing I noticed: this overload's new default always uses `SyntaxHighlightColors.light()`, regardless of the active theme, so a dark mode app that relies on the default here would get light syntax colors. The other two overloads switch on `isDark` via `remember(isDark) { ... dark() else light() }`. Since this default is evaluated in a `@Composable`, it could read `JewelTheme.isDark` for consistency. It is fairly minor since callers usually pass the highlighter explicitly (as the sample does), but it might be nice to resolve the theme awareness inconsistency.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
```diff
@@ -0,0 +1,95 @@
...
+                if (bestMatch == null || match.range.first < bestMatch.range.first) {
+                    bestMatch = match
+                    bestRule = rule
+                }
+            }
+
+            if (bestMatch == null || bestRule == null) break
+
+            for ((groupIndex, tokenType) in bestRule.captures) {
+                // if  the regex doesn't have as many groups as bestMach, skip silently
```
nebojsa-vuksic: Tiny nit: there is a small typo and a double space in this comment. `if  the regex doesn't have as many groups as bestMach` could become `if the regex doesn't have as many groups as bestMatch`.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
```diff
@@ -0,0 +1,95 @@
...
+public class SimpleCodeHighlighter(
+    public val colors: SyntaxHighlightColors,
+    public val additionalGrammars: List<LanguageGrammar>,
+) : CodeHighlighter {
+    public constructor(colors: SyntaxHighlightColors) : this(colors, emptyList())
+
+    @Deprecated(message = "Use highlight(code, language) instead.", level = DeprecationLevel.HIDDEN)
+    override fun highlight(code: String, mimeType: MimeType?): Flow<AnnotatedString> =
+        flowOf(AnnotatedString(code))
+
```
rock3r, line 52: Looks like this deprecated overload is weaker than the bridge highlighter's compatibility path. In bridge, `highlight(code, mimeType)` still resolves known MIME types back to an IDE language and highlights them; here any MIME-looking fence routed through `RenderCodeWithMimeType` stays plain text.

Can we keep the same compatibility behavior for standalone too, at least for the built-in Kotlin/Java mappings, and also avoid using the deprecated overload in the renderer when we still have the raw fence info string? That would make ` ```text/x-kotlin ` behave consistently with ` ```kotlin ` and keeps the new raw-language path as the preferred one.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
```diff
@@ -0,0 +1,95 @@
...
+    override fun highlight(code: String, language: String): Flow<AnnotatedString> {
+        if (language.isBlank()) return flowOf(AnnotatedString(code))
+        val grammar = findGrammar(language) ?: return flowOf(AnnotatedString(code))
+        return flowOf(tokenize(code, grammar))
+    }
+
+    private fun findGrammar(language: String): LanguageGrammar? {
+        val normalized = language.trim().lowercase()
+        return (additionalGrammars + BuiltInLanguageGrammars.all).firstOrNull { normalized in it.names }
```
rock3r: Could we resolve the fence info string the same way the bridge native path does? The Markdown parser preserves the full info string, so something like ` ```kotlin startline=3 ` reaches the highlighter as `"kotlin startline=3"` and won't match `names = listOf("kotlin", "kt", "kts")`.

Bridge's `CodeFenceLanguageGuesser` first tries the full value, then progressively strips trailing space-separated customizations until a language matches. That seems like the right behavior here too, especially since custom grammars could have multi-word names. As a side note, the bridge TextMate fallback probably wants the same behavior eventually too; it currently uses only the first token.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
```diff
@@ -0,0 +1,95 @@
...
+    public constructor(colors: SyntaxHighlightColors) : this(colors, emptyList())
+
+    @Deprecated(message = "Use highlight(code, language) instead.", level = DeprecationLevel.HIDDEN)
+    override fun highlight(code: String, mimeType: MimeType?): Flow<AnnotatedString> =
+        flowOf(AnnotatedString(code))
+
+    override fun highlight(code: String, language: String): Flow<AnnotatedString> {
+        if (language.isBlank()) return flowOf(AnnotatedString(code))
+        val grammar = findGrammar(language) ?: return flowOf(AnnotatedString(code))
+        return flowOf(tokenize(code, grammar))
```
rock3r: This tokenizes eagerly before the `Flow` is returned, so Markdown rendering does the regex scan while creating the flow from composition:

```kotlin
remember(content, language, highlighter) {
    highlighter.highlight(content, language)
}
```

Can we make the flow lazy and do the tokenization off the main thread, aligned with the bridge highlighter? The bridge does the actual annotation from inside the flow and uses `withContext(highlightDispatcher)`, which avoids making large/many fenced blocks do all the work synchronously on the composition path.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/LanguageGrammar.kt
```diff
@@ -0,0 +1,27 @@
...
+ * that appears first in [rules] wins, so higher-priority rules (e.g., comments and strings) should come first.
+ *
+ * @param names All lowercase names and aliases for this language (e.g., `["kotlin", "kt", "kts"]`). These are matched
+ *   against the info string of a Markdown fenced code block.
+ * @param rules Priority-ordered list of [TokenRule]s. Each rule's [TokenRule.captures] maps capture group indices to
+ *   [TokenType]s. A single rule can color multiple parts of a match independently.
+ */
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+@Immutable
```
rock3r, line 27: Question: should these collection-backed types defensively copy their inputs if they are marked `@Immutable`? `LanguageGrammar` exposes `List`s, and `TokenRule` exposes a `Map`; those are read-only interfaces, but callers can still pass mutable collections and mutate them after construction.

I'd lean toward snapshotting the collections in the constructors and documenting that behavior, so the public API actually satisfies the `@Immutable` contract. Otherwise we may need to drop the annotation from the collection-backed types.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/TokenRule.kt
```diff
@@ -0,0 +1,102 @@
...
+ * - Key `0` refers to the entire match.
+ * - Keys `1`, `2`, etc. refer to specific capturing groups.
+ *
+ * A single rule can produce multiple spans — for example, a rule that matches `fun myFunc` can color `fun` as
+ * [TokenType.KEYWORD] (group 1) and `myFunc` as [TokenType.FUNCTION_CALL] (group 2) in one pass.
+ *
+ * Patterns use Java's `java.util.regex` engine, which supports lookahead (`(?=...)`) and fixed-width lookbehind
+ * (`(?<=...)`, `(?<!...)`). This covers the vast majority of tmLanguage-style patterns.
+ *
+ * **Known limitations compared to PCRE/Oniguruma** (used by TextMate grammars):
```
rock3r, line 26: I think Java regex does support [possessive quantifiers](https://docs.oracle.com/javase/tutorial/essential/regex/quant.html) (`*+`, `++`, `?+`) — are you sure these need to be listed as unsupported?

If not, could we remove this from the limitation list here and in `standalone-code-highlighting.md`? Otherwise future grammar authors may rewrite valid imported patterns unnecessarily.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
rock3r: This might be an AI false positive, so take it with a pinch of salt, but... I (gippy) wonder if remembering the bridge highlighter flow can make bridge Markdown keep using a stale editor color scheme after theme changes.

`IntelliJCodeHighlighter.createHighlightingFlow()` currently captures `EditorColorsManager.getInstance().globalScheme` when the flow is created, and this `remember(content, language, highlighter)` keeps the same flow while those keys stay unchanged. If that is real, it probably deserves a separate issue/fix rather than blocking this PR.

platform/jewel/markdown/int-ui-standalone-styling/src/main/kotlin/org/jetbrains/jewel/intui/markdown/standalone/IntUiProvideMarkdownStyling.kt
```diff
@@ -53,7 +54,10 @@ public fun ProvideMarkdownStyling(
                 MarkdownBlockRenderer.light(markdownStyling)
             }
         },
-    codeHighlighter: CodeHighlighter = remember { NoOpCodeHighlighter },
+    codeHighlighter: CodeHighlighter =
```
rock3r, line 59: The KDoc in this file still describes the old behaviour. After this PR, standalone Markdown does have syntax highlighting by default through `SimpleCodeHighlighter`.

Could we update it to mention the new default and say that callers can pass `NoOpCodeHighlighter` or their own `CodeHighlighter` to opt out/replace it? The same text appears on the next overload too.

rock3r (changes requested):

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/JavaLanguageGrammar.kt
DanielSouzaBertoldi: Sure, I don't recall exactly why I moved the `functionCall` rule above `keyword`, but I was able to confirm this bug too.

image

Moving the line to after the `keyword` is enough to apply the correct color for the examples you mentioned. Thanks!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/KotlinLanguageGrammar.kt
DanielSouzaBertoldi: Yep, I agree :) removing `enum` from the `typeDeclaration` regex is enough to fix this:

| Before | After |
|---|---|
| image | image |

I'll also add a test. Thanks!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
DanielSouzaBertoldi: The thing though is that we are actively deprecating the MimeType APIs :( that's why I opted to set the level as Hidden, so that users are not encouraged at all to use this overload

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
DanielSouzaBertoldi: Great idea! Will implement :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
rock3r: Ah missed that bit. Then yeah this is fine

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/LanguageGrammar.kt
DanielSouzaBertoldi: Oh damn, you are right! Completely missed this. I'll snapshot the collections since this is basically a config object anywho

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/TokenRule.kt
DanielSouzaBertoldi: Oof, Claude created this KDoc so confidently that I thought it was true. Didn't double check it, my bad. Will remove mentions of this limitation from the code, thanks :)

rock3r: This was a GPT finding, which I was also not sure about but went and validated it :D

rock3r: By the way, if this means we can improve/simplify our lexers, we should...

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
DanielSouzaBertoldi: Hmmmm after digging deeper gippy is correct! We are basically snapshotting the color scheme once `createHighlightingFlow()` is called the first time. The fix is pretty simple though, just move the colorScheme line to `highlightAndEmit`. 

Filed a bug: https://youtrack.jetbrains.com/issue/JEWEL-1343/Fix-stale-colorScheme-in-IntelliJCodeHighlighter

🕺🏻

rock3r: Thanks Gippy <3

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
rock3r: I can only see first-token-matching implemented in `findGrammar`, is it supposed to be done now?

rock3r (changes requested):

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
```diff
@@ -0,0 +1,101 @@
...
+                if (bestMatch == null || match.range.first < bestMatch.range.first) {
+                    bestMatch = match
+                    bestRule = rule
+                }
+            }
+
+            if (bestMatch == null || bestRule == null) break
+
+            for ((groupIndex, tokenType) in bestRule.captures) {
+                // if the regex doesn't have as many groups as bestMach, skip silently
```
rock3r:
```suggestion
                // if the regex doesn't have as many groups as bestMatch, skip silently
```

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
DanielSouzaBertoldi: Oh, right. I totally forgot to reply here with my reasoning 😛 

Basically, if there's any customization then they'll appear after a space, right? Like, using your own example:

`kotlin startline=3`

So, to separate the language from everything else, we just need to get the first string before an whitespace :)

rock3r: Yep but my point was to align to the `CodeFenceLanguageGuesser` behaviour...

DanielSouzaBertoldi: Not gonna lie, for bridge I just copied and pasted the code from IntelliJ's markdown, and after reading the specification for both [CommonMark](https://spec.commonmark.org/0.31.2/#info-string) and [GFM](https://github.github.com/gfm/#fenced-code-blocks), nobody really cares how to handle the info string.

Having to keep trimming down the attributes just to find a language seems a bit wasteful in my opinion, plus for all supported languages in IntelliJ's textmate none of them have a language name with a whitespace, and I really doubt that users will create a custom language grammar that expects one.

Buuuut anywho, since there's no spec at all, then we can keep logic parity between the two

rock3r: Fair pushback, I am ok with postponing that too

DanielSouzaBertoldi: Nah it's fine my guy, just making sure to let you know my opinion, but given that there's no spec, keeping the logic parity is fine by me ❤️ just pushed the changes

rock3r: Much love <3

DanielSouzaBertoldi:
@rock3r @nebojsa-vuksic after the last changes, I made sure to also better improve the `LanguageGrammar` API by having it declare a `name` and `aliases`. This way for languages that do not have any other aliases like TOML it's just a matter of creating a `LanguageGrammar(name = "toml")` instead of `LanguageGrammar(names = listOf("toml")`

Kinda of a whatever change but something about the `names` nomenclature was bugging me out. I also updated the `standalone-code-highlighting` to mirror this change :)

rock3r (approved):

rock3r:
Thanks man, you rock <3

DanielSouzaBertoldi:
Ready to merge

PR #3507, opened 2026-05-15
[IJPL-43620] [platform] Custom Ordering for Status Bar Items

endrju19:
Adds drag-and-drop reordering of widgets in the status bar.

User-chosen order is stored as `Map<widgetId, Int>` — an absolute index per widget. On each drop, dense indices for currently-visible widgets are snapshotted. During sort, after `LoadingOrder.sortByLoadingOrder`, a stable sort by `widgetOrder[id] ?: Int.MAX_VALUE` places customized widgets in their user-chosen order; uncustomized widgets trail in their default LoadingOrder positions. Entries for currently-invisible widgets are preserved so a re-enabled widget returns to its saved position.

This matches the pattern used by tool window ordering (`DesktopLayout.setAnchor` / `normalizeOrder`).

Continuation of #3447 — same DnD wiring (`WidgetTransferable` / `WidgetTransferHandler`) and the `WIDGET_IT` fix in `IdeStatusBarImplKt#wrap` needed for DnD detection. The earlier model (replaying a list of `source → target` moves) cascade-shifted unrelated widgets on inverse drags; the absolute-index model eliminates that structurally.

11 unit tests in `WidgetSorterTest` cover the regression and edge cases.

Supersedes #3447.


platform/platform-impl/src/com/intellij/openapi/wm/impl/status/IdeStatusBarImpl.kt
aydarin: After moving 1 widget it saves (in my case) an order state for all 24 widgets (though I see only 7 in my status bar). That prevents the default order being changed. As a result, every newly installed/visible widget is automatically located to the very right if any customisations happened before.
The previous approach allowed that.

Do we need to save the order number for all of the visible widgets? I would suggest saving the order number for the moved widgets. So whenever a user drops the widget, we save the new index for it + update the index of previously moved widgets. 

Example:
1. Default order: A-B-C-D-E
2. User moves C to the index 1, then they see in status bar: A-C-B-D-E
3. **New** - The persisted state map: (C -> 1)
4. User moves D to the index 0, then they see in status bar: D-A-C-B-E
5. **New** - We persist the new index for D and update the persisted state for all previously customized widgets. In this case we save the new index for C. Then the persisted state is: (D -> 0, C -> 2)

For reordering widgets you take all the widgets with customized orders and apply the change from the smallest index to the biggest one.

What do you think?

aydarin: Some widget are not draggable. Would it be possible to fix it?
https://github.com/user-attachments/assets/0a744452-68f9-4380-8dbc-c0a04f284b0b

aydarin (changes requested):

PR #3509, opened 2026-05-18
[vcs-log] consider other branch-like filters when checking if the log is filtered by a single branch

juliabeliaeva:
This is a fix for [IJPL-245209](https://youtrack.jetbrains.com/issue/IJPL-245209) issue that ensures that current branch highlighting in Git Log is visible when single branch filter is accompanied by a revision filter.


iasemenov (approved):

iasemenov:
Merged, thank you!

PR #3512, opened 2026-05-19
IJPL-245319 quote $new_name in fish shell integration

msmorgan:
Fish's test-require-arg feature flag causes an error in `[ $new_name ]` when `$new_name` expands to nothing. Fixed by using `-n` and quoting the variable.

Also switched to `test -n` and `contains` as idiomatic equivalents.


KonstantinHudyakov:
Thank you for the pull request! It is merged: https://github.com/JetBrains/intellij-community/commit/111c13bfda72eefc2cf9239de6292d48a1b89662

PR #3515, opened 2026-05-21
Write Shard Status File Early

alonalbert:
Shard status must be written even if no tests are actually executed so write it early.

[Bug BAZEL-3206](https://youtrack.jetbrains.com/issue/BAZEL-3206/JUnit5BazelRunner-Doesnt-Write-Shard-Status-File-if-No-Tests-Are-Executed)


k15tfu (approved):

shalupov:
in master as https://github.com/JetBrains/intellij-community/commit/9312d30d279d90f58ebde158c973077a8f986a03

PR #3521, opened 2026-05-26
[gradle] IDEA-389806 honor IDE parallel model fetch setting and preserve legacy fallback on Gradle 9.4+

joshfriend:
Commit 4549aa118b unconditionally added `-Didea.parallelModelFetch.enabled=true` for Gradle 9.4+ and kept passing `-Dorg.gradle.tooling.parallel.ignore-legacy-default=true` whenever the IDE setting was off. Two problems:

1. The IDE 'Parallel model fetch' setting was silently ignored on Gradle 9.4+, so batched fetch was forced on with no opt-out.
2. `ignore-legacy-default=true` strips Gradle's documented fallback from `org.gradle.tooling.parallel` to `org.gradle.parallel`. Projects that already set `org.gradle.parallel=true` then run the batched fetch sequentially, regressing IDE sync by ~60% on large projects.

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleProjectResolver.java
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleProjectResolver.java
@@ -599,11 +599,21 @@ public final class GradleProjectResolver implements ExternalSystemProjectResolve
       executionSettings.withArgument("-Didea.resolveSourceSetDependencies=true");
     }
     boolean parallelModelFetch = executionSettings.isParallelModelFetch();
-    if (!parallelModelFetch) {
-      executionSettings.withArgument("-Dorg.gradle.tooling.parallel.ignore-legacy-default=true");
-    }
-    if (parallelModelFetch || GradleVersionUtil.isGradleAtLeast(resolverContext.getGradleVersion(), "9.4")) {
+    // Gradle 9.4 introduced org.gradle.tooling.parallel, decoupling Tooling API
+    // parallelism from task parallelism. See:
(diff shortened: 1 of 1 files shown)


plugins/gradle/src/org/jetbrains/plugins/gradle/service/project/GradleProjectResolver.java
```diff
@@ -599,12 +599,24 @@ private static void configureExecutionArgumentsAndVmOptions(
...
+    // Gradle 9.4 introduced org.gradle.tooling.parallel, decoupling Tooling API
+    // parallelism from task parallelism. See:
+    // https://docs.gradle.org/9.4.0/release-notes.html#new-property-for-tooling-api-parallelism-control
+    boolean gradleSupportsParallelModelFetch =
+      GradleVersionUtil.isGradleAtLeast(resolverContext.getGradleVersion(), "9.4");
+    if (parallelModelFetch) {
+      // IDE explicitly opted into batched model fetch.
       executionSettings.withArgument("-Didea.parallelModelFetch.enabled=true");
     }
+    else if (!gradleSupportsParallelModelFetch) {
```
nskvortsov: In this branch, IDEA model builders will never work in parallel, regardless of any flags (including `org.gradle.tooling.parallel`).  See https://github.com/JetBrains/intellij-community/blob/45d0d66edf8a44e4675a11ea142e01bc1d258011/plugins/gradle/tooling-extension-impl/src/com/intellij/gradle/toolingExtension/impl/util/GradleModelProviderUtil.java#L27

nskvortsov (changes requested):
The root cause of the problem is that previously IDEA and Android model builders were running via different code paths, allowing to control the parallelism separately. Now it is not the case anymore.

PR #3523, opened 2026-06-01
[JEWEL-1257] Add support for new islands palette

rock3r:
This PR adds support for the Islands color palette format in `ThemeColorPalette`.

Islands palettes use indexed steps starting from 10 and incrementing by 10 (10, 20, 30, …) instead of consecutive numbering (1, 2, 3, …).

- The `ThemeColorPalette` constructor now accepts a new `isIslands` parameter to distinguish between Islands and non-Islands palette types.
- All color accessor methods (`grayOrNull()`, `blueOrNull()`, etc.) now support both indexing schemes depending on the palette type.
- The Bridge theme reader detects Islands themes via `IslandsState.isEnabled()` and reads colors using the appropriate key format.

### ⚠️ Important Changes
 * `ThemeColorPalette` now includes a new `isIslands` parameter to distinguish between Islands and non-Islands theme palettes.
 * `ThemeColorPalette` color accessor methods now support both classic and Islands indexing schemes.
 * `BridgeThemeColorPalette` now parses Islands palette keys when applicable.
 * When using color accessors, use the native index scheme for the palette: classic themes use `1, 2, 3, …`, while Islands themes use `10, 20, 30, …`.
   * For example, the Islands equivalent of `blueOrNull(1)` is `blueOrNull(10)`, and `redOrNull(3)` becomes `redOrNull(30)`.
   * Note that at the time of writing the official Islands themes also contain the non-Island palette style entries, but that is not guaranteed to be the case for third party themes or in the future.

### Deprecated API
* The `ThemeColorPalette` constructor without the `isIslands` parameter was removed from the visible API; specify `isIslands` when constructing palettes.
  * The API is retained for binary compatibility for the time being, but please migrate as soon as possible.

### New features
 * Added support for the Islands color palette format in `ThemeColorPalette`.

### DIFF
--- a/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
+++ b/platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
@@ -4,8 +4,11 @@ import androidx.compose.runtime.Immutable
 import androidx.compose.ui.graphics.Color
 import org.jetbrains.jewel.foundation.GenerateDataFunctions
 
-private val colorKeyRegex: Regex
+private val nonIslandsColorKeyRegex: Regex
     get() = "([a-z]+)(\\d+)".toRegex(RegexOption.IGNORE_CASE)
+
+private val islandsColorKeyRegex: Regex
+    get() = "([a-z]+)-(\\d+)".toRegex(RegexOption.IGNORE_CASE)
 
 /**
  * A palette of colors provided by the theme.
@@ -14,7 +17,7 @@ private val colorKeyRegex: Regex
  *
  * Note that not all Look and Feel themes are guaranteed to have a full palette, and some may not have one at all. The
  * number of colors in each list depends on the implementation in the LaF. It is therefore important to use the
- * `*OrNull` accessors to avoid [IndexOutOfBoundsException]s.
+ * `*OrNull` accessors to avoid [NullPointerException]s.
  *
  * @property gray A list of gray colors, from lightest to darkest.
  * @property blue A list of blue colors.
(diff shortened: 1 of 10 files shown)


rock3r:
Ready for review: all CI checks are now passing except the API-change annotation check, which is expected for this API update.

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
```diff
@@ -237,18 +278,20 @@ public class ThemeColorPalette(
         "This can throw exceptions if the LaF does not have a full palette, use purpleOrNull() instead",
         ReplaceWith("purpleOrNull(index)"),
     )
-    public fun purple(index: Int): Color = purple[index - 1]
+    @Suppress("UnsafeCallOnNullableType")
+    public fun purple(index: Int): Color = purpleOrNull(index)!!
```
DanielSouzaBertoldi, line 282: I know this (and the others) were deprecated a long time ago by now, but this change effectively makes the `* @throws IndexOutOfBoundsException if the index is out of bounds.` part in the KDoc obsolete.

Now, if for whatever reason the double bangs fail, it'll actually throw an `NullPointerException`.

I'm down to updating the KDoc as well even though they are all deprecated

DanielSouzaBertoldi (approved):
LGTM! 🕺🏻 

Just one minor point I wanted to raise, but no blockers

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/theme/ThemeColorPalette.kt
rock3r: Good point, let me do that KDoc update

rock3r: Updated the deprecated accessor KDocs to document the possible NullPointerException instead of IndexOutOfBoundsException.

DanielSouzaBertoldi (approved):
Cool stuff, thanks!

nebojsa-vuksic (approved):

rock3r:
ready to merge

PR #3524, opened 2026-06-02
Optimize Islands border painting allocation churn

rsauciuc:
... to minimize allocations per repaint cycle:
  * reuse a single 'sharedPath' (Path2D.Float) instance.
  * reuse a single 'CachedBoundsShape' proxy wrapper with pre-allocated mutable 'cachedBounds2D' and 'cachedBounds' fields to avoid Rectangle/Rectangle2D allocations on background graphics fills.

This is particularly important for components nesting Compose panels: when Compose animations are running at full speed (60/120fps), there's a flurry of allocations (and extra work) from frequent repaints.

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/application/impl/islands/IslandsUICustomization.kt
+++ b/platform/platform-impl/src/com/intellij/openapi/application/impl/islands/IslandsUICustomization.kt
@@ -97,12 +97,16 @@ import java.awt.Insets
 import java.awt.Point
 import java.awt.Rectangle
 import java.awt.RenderingHints
+import java.awt.Shape
 import java.awt.Toolkit
 import java.awt.event.AWTEventListener
 import java.awt.event.HierarchyEvent
 import java.awt.event.WindowAdapter
 import java.awt.event.WindowEvent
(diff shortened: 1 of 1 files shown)


rsauciuc:
@rock3r does this look reasonable?

rock3r (approved):
LGTM but will leave to @alexander-lobas as he's the Swing expert :)

bulenkov:
@rsauciuc Thank you so much for the patch. Could you point to a test or a specific place in Android Studio we can use to check the before/after state?

rsauciuc:
Hi Konstantin! 

This is showing up for Gemini in Android Studio, because it's a Compose UI. There's a bit too much GPU and CPU activity when the agent is waiting and we're just animating some shimmer/spinner/progress indicator, or just scrolling through the timeline. I used async-profiler to get clues. The Compose bits are rendering at almost 120fps, we have a few recomposition bugs on our side, but I also noticed the border painting is too eager to repaint and allocate objects. I can share the profiles if it's helpful.

rock3r:
You can probably repo with Junie or AIA toolwindows too

PR #3525, opened 2026-06-02
[JEWEL-1138] Scale images to fit available width

DanielSouzaBertoldi:
> [!IMPORTANT]
>  This PR replaces [the one by](https://github.com/JetBrains/intellij-community/pull/3387) @AlexVanGogen, addressing open concerns :)

## Description

Images in paragraphs and headings are now scaled down to fit the available width when they would otherwise overflow.

This is implemented via TextWithScalableInlineContent, which wraps a two-layer layout:

- An outer Layout that provides custom intrinsic measurements (min/max width and height). This is necessary because SubcomposeLayout does not support intrinsic measurement, but parent containers like tables rely on it.
- An inner SubcomposeLayout that defers Text composition until constraints are known, so placeholder dimensions can be scaled to the actual available width before the text is laid out.

Images are rendered as `InlineTextContent` within the existing `Text` composable. There's no content splitting, no `FlowRow`, no separate image composables. The `ImageRendererExtension` interface (e.g. the `Coil3` implementation) must return placeholders using `TextUnitType.Sp` units so they can be correctly converted to pixels and scaled. Non-Sp placeholders are skipped with a warning log.

The only con we have now is that this approach requires the `ImageRendererExtension` implementation to use `Sp` placeholder units. This is documented in the interface `KDoc` and enforced with a runtime warning, though.

**Image-only paragraphs/headings:** these now also go through `TextWithScalableInlineContent` and scale to fit the container width, matching the behavior for mixed text+image content. This replaces the previous `FlowRow`-based approach that was mentioned in earlier release notes.

## Screen Recording

https://github.com/user-attachments/assets/07ba1ffe-0d13-4ab5-b6e3-2ef36ab4f291

### Bug fixes
 * Markdown images now scales images to fit available width and adjusts vertical space


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,21 +852,288 @@ public open class DefaultMarkdownBlockRenderer(
...
+                        style = mergedStyle,
+                        softWrap = false,
+                        maxLines = 1,
+                        placeholders = placeholders,
+                    )
+
+                measured.size.width
+            }
+
+        val minIntrinsicWidth =
```
DanielSouzaBertoldi, line 926: There's still some discussion to be had about the calculations here. Following up [on this one](https://github.com/JetBrains/intellij-community/pull/3387/changes#r3196953666):

> For image-only paragraphs/headings, textWithoutInlineContent becomes effectively empty, so minIntrinsicWidth returns 0. A parent using IntrinsicSize.Min or otherwise querying minIntrinsicWidth could then measure the image at zero width.

I had this to say:

Scaling an image down to fix 0 is semantically defensible. Technically the content can render at 0 width, even if it's effectively invisible. As a user, what's a useful expected lower bound in this case? Should we return 1? The image width? The downside is that any non-zero value would prevent scaling below that point, which arguably defeats the purpose of this feature :/

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,21 +852,288 @@ public open class DefaultMarkdownBlockRenderer(
...
+            remember(textWithoutInlineContent, mergedStyle, density, fontFamilyResolver, layoutDirection) {
+                val resolvedStyle = resolveDefaults(mergedStyle, layoutDirection)
+                ceil(
+                        ParagraphIntrinsics(
```
DanielSouzaBertoldi, line 930: The old code splitted the text by whitespace, measured each result and then returned the max value measured. For CJK text which contains no spaces, any text would be measured as one indivisible token at full width. So we couldn't line break character by character.

Now we just delegate the width calculation to Compose, less of a hassle for us.

We need to wrap the result in a `ceil().toInt()` because the Compose API returns a Float and we need to round up the minimum size, guaranteeing we are always reserving at least as much space required

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,21 +852,288 @@ public open class DefaultMarkdownBlockRenderer(
...
+            .toSortedSet(compareBy(InlineContentLayoutState::id))
+
+    /** Scales the inline content placeholders by the given scale factor. */
+    private fun scaleInlineContent(
+        content: Map<String, InlineTextContent>,
+        availableWidth: Int,
+        density: Density,
+    ): Map<String, InlineTextContent> {
+        return content.mapValues { (_, inlineContent) ->
+            if (inlineContent.placeholder.width.type != TextUnitType.Sp) {
```
DanielSouzaBertoldi, line 1122: This code [addresses this comment](https://github.com/JetBrains/intellij-community/pull/3387#discussion_r3196694312)

> This scaling path seems to assume all image placeholders use sp units. ImageRendererExtension exposes raw Compose InlineTextContent, and Placeholder.width / height are TextUnits that can also be em. Density.roundToPx() throws for non-sp text units (Only Sp can convert to Px), while passing the placeholder directly to Text lets Compose resolve em placeholders relative to the text style.
>
> The built-in Coil renderer uses sp, so it is fine, but a custom image renderer returning e.g. Placeholder(width = 1.em, height = 1.em, ...) would now crash once this scaling path runs. Can we either handle em placeholders here, or make the ImageRendererExtension contract explicitly require sp placeholders and guard this path with a clearer failure mode?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -820,21 +852,288 @@ public open class DefaultMarkdownBlockRenderer(
...
+                                textAlign = textAlign,
+                            )
+                        }
+
+                    val placeable = measurables.firstOrNull()?.measure(constraints)
+
+                    layout(
+                        width = placeable?.width ?: constraints.minWidth,
+                        height = placeable?.height ?: constraints.minHeight,
+                        alignmentLines =
```
DanielSouzaBertoldi, line 1055: This code [addresses this comment](https://github.com/JetBrains/intellij-community/pull/3387#discussion_r3196959183)

> This wrapper should probably forward the text alignment lines. Before this change the paragraph rendered a Text directly, so parents using baseline alignment could read FirstBaseline / LastBaseline. In the scalable path we now have Layout -> SubcomposeLayout -> Text, but both layout(...) calls place the child without passing any alignmentLines, so the markdown block no longer exposes the text baselines when this path is used.
> 
> Could we forward the measured child’s FirstBaseline / LastBaseline through both wrapper layers when specified, or add coverage if baseline alignment is meant to remain supported?

platform/jewel/markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRendererTest.kt
```diff
@@ -136,4 +138,11 @@ public class DefaultMarkdownBlockRendererTest {
             assertEquals("second:https://example.com", clickedUrl)
         }
     }
+
+    @Test
+    public fun `appendInlineContent uses expected annotation tag`() {
```
DanielSouzaBertoldi, line 143: This [test addresses this comment](https://github.com/JetBrains/intellij-community/pull/3387#discussion_r3196947589)

> Since this mirrors Compose's internal INLINE_CONTENT_TAG, could we leave a comment here and maybe add a small guard test? I get why we need the exact tag to filter inline-content annotations, but the constant is internal to Compose (InlineTextContent.kt), so a future Compose upgrade could change it without a compiler error.
>
> If that happens, regular Text rendering may still work, but getPlaceholders() / withoutInlineContent() would stop seeing inline-content ranges and our custom intrinsic calculations would silently diverge. A comment plus a tiny test around appendInlineContent("id") would make this dependency explicit.

---

I opted to using `appendInlineContent` directly instead of making `getPlaceHolders()`/`withoutInlineContent()` internals just for this. This test is enough to make us aware of any changes on Compose side

DanielSouzaBertoldi:
I'm unable to run ApiCheckTest for some reason, since I can't build the project through JPS. I'll update the api-dump files once I finally can

rock3r:
Can you fix conflicts and improve the release notes? If anything they should mention this is about Markdown images, otherwise it's just way too vague :D

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
DanielSouzaBertoldi: Replying to [this comment by](https://github.com/JetBrains/intellij-community/pull/3387#discussion_r3347212395) @rock3r :

> I would imagine it should report the min intrinsic size based on the inline image's size (which would be ~zero if it's not specified and the image is not loaded yet, that's not something we can improve on)

---

My main concern though is that images are scalable, so the minimum/maximum intrinsic size is basically whatever the image is currently being capable of rendering given the bounds it has 🤔 idk, this is a bit confusing to me honestly, could you please elaborate a bit on this?

DanielSouzaBertoldi: Like, I get it's a bit odd from an usability standpoint, but given that scalable images have no minimum width per se, to me it's ok to return ~0 for image-only paragraphs 😵‍💫

rock3r: The image's intrinsic size is:
* for bitmaps, its resolution
* for SVGs, its declared size, or if that is missing, the view box size as declared in the SVG

So, I do not really follow the problem with the images being scalable; we are talking about their intrinsic sizes, which are a very specific value, at the net of any scaling applied to them

DanielSouzaBertoldi: I get what you're trying to say but I think the distinction here is that `maxIntrinsicWidth` already captures the image's natural pixel dimensions. `minIntrinsicWidth` is a separate concept. It's the minimum width at which content renders without clipping. Since the whole point of this PR is that images scale down, they never really clip, they just get smaller. So `minIntrinsicWidth` is really about the text (the widest unbreakable word). If we returned the image's pixel width there instead, we'd basically be telling parent layouts "don't go below 300px" whiiiiich is basically what this feature is supposed to avoid, no?

rock3r: Yes, you can technically scale any image up or down as much as you want, but their _intrinsic_ size is fixed. It's one value for both min and max, and that value is exactly the image size. So, semantically, setting zero is wrong; however, since we are adapting into a different system, I understand what you mean; I would not let it be 0x0, as that seems wrong, but perhaps 1x1? Mostly because I want to avoid an image disappearing entirely in the case where is no room to show it and so it's sized to 0x0. I mean, 1x1 is also bad, but perhaps less bad?

DanielSouzaBertoldi: Honestly, idk man :( I mean, I can coerce the value to be 1, but 1px of size is invisible to the naked eye and I feel like it's not going to solve the issue we are discussing here to be fair.

Like, I get that, for instance, an image that is 300x200 has a minimum intrinsic size of 300x200. That's the resolution for the image, and I agree!

But in the code, when we create the `MeasurePolicy` we are dealing with the "minimum layout" needed before the content starts to get clipped. Since we are scaling them based on the space the image can occupy, the actual minimum is 0 (or 1 for convention).

Also, `minIntrinsicWidth` is only queried when a parent is sizing itself (like `IntrinsicSize.Min`). The actual measure pass just uses whatever constraints the parent has. So returning 1 doesn't really help... If the space is 0, the image still measures at 0. Besides, we would be returning a false floor without the benefit (no visibility at all when a image is this smal)

IMO intrinsic queries should communicate the contract the component implements, not the contract of the primitive it wraps.

For a compromise, I added some comments explaining the reasoning behind it.

Buuuut I'm happy to keep discussing this 😄

rock3r: Nah, I said I understand — it's fine. If we see it breaks something, we'll change it.

rock3r (approved):

nebojsa-vuksic (approved):

nebojsa-vuksic:
Ready to merge

PR #3526, opened 2026-06-02
Do Not Log an Error if ObjectCollectedException

alonalbert:
See https://youtrack.jetbrains.com/issue/KTIJ-39019

### DIFF
--- a/plugins/kotlin/jvm-debugger/coroutines/src/org/jetbrains/kotlin/idea/debugger/coroutine/CoroutineStackFrameInterceptor.kt
+++ b/plugins/kotlin/jvm-debugger/coroutines/src/org/jetbrains/kotlin/idea/debugger/coroutine/CoroutineStackFrameInterceptor.kt
@@ -22,6 +22,7 @@ import com.intellij.xdebugger.impl.XDebugSessionImpl
 import com.sun.jdi.ArrayReference
 import com.sun.jdi.Location
 import com.sun.jdi.LongValue
+import com.sun.jdi.ObjectCollectedException
 import com.sun.jdi.ObjectReference
 import com.sun.jdi.Value
 import org.jetbrains.kotlin.idea.debugger.base.util.evaluate.DefaultExecutionContext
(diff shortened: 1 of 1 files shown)


plugins/kotlin/jvm-debugger/coroutines/src/org/jetbrains/kotlin/idea/debugger/coroutine/CoroutineStackFrameInterceptor.kt
```diff
@@ -271,9 +271,13 @@ internal fun callMethodFromHelper(
     } catch (e: MethodNotFoundException) {
         fileLogger().warn(e)
     } catch (e: Exception) {
-        val helperExceptionStackTrace = MethodInvokeUtils.getHelperExceptionStackTrace(context.evaluationContext, e)
-        DebuggerUtilsImpl.logError("Exception from helper: ${e.message}", e,
-                                   *listOfNotNull(helperExceptionStackTrace).toTypedArray()) // log helper exception if available
+        if (Registry.`is`("debugger.collectible.value.retries.error", true)) {
```
zuevmaxim: actually this flag is not supposed to be used for this purpose. Due to many problems with helper methods on ART, maybe it will be better to disable them specifically for android for now?
We would like to have this suppression android specific, as it does not appear on other platforms

alonalbert: I was also tempted to disable method helpers, but I don't understand the specifics.

I'm open to this if you think it's safe.

FYI, the ART team is investigating the  ObjectCollectedException issue on our end.

alonalbert: Hmm, I got an email with a comment from Egor but I don't see it here in the conversation:

```
I will replace it to

        if (e is ObjectCollectedException || e.cause is ObjectCollectedException) {
            return null
        }
I suppose that should be enough and valid, as we already tried the retry approach

```

alonalbert:
I updated the commit.

plugins/kotlin/jvm-debugger/coroutines/src/org/jetbrains/kotlin/idea/debugger/coroutine/CoroutineStackFrameInterceptor.kt
zuevmaxim: Sorry, it was me, but then we started an internal discussion regarding such a solution. I think it's OK, but I will add a warning. It will not break the tests

zuevmaxim:
514164e5803f3c8287691787222f40e769b9fc6f

PR #3527, opened 2026-06-03
[JEWEL-1267] Support sized images

DanielSouzaBertoldi:
> [!NOTE]
> This PR supersedes  #3442 addressing remaining comments :)

## Summary

Add support for parsing and rendering sized images in Jewel Markdown when image dimensions are provided via HTML `width` and `height` attributes or native Markdown image attribute blocks (`{width=100 height=50%}`), including both absolute pixel values and percentage-based dimensions with proportional scaling when only one dimension is specified.

Fixes JEWEL-1267

## Changes

 * New `DimensionSize` sealed interface with `Pixels` and `Percent` value classes to represent image dimension specifications
 * Extended `InlineMarkdown.Image` with optional `width` and `height` properties
 * Added parsing of Markdown image attribute blocks: `alt{width=100 height=50%}` — supports unquoted, single-quoted, and double-quoted attribute values, with `px` and `%` suffixes
 * Added `parseHtmlSizeValue()` for parsing HTML size attributes (e.g., `"100"`, `"100px"`, `"50%"`) into `DimensionSize` values, with negative percentage values treated as invalid
 * Refactored `Coil3ImageRendererExtensionImpl` to compute placeholder sizes based on specified dimensions — when both are given they're used directly, when only one is specified the other scales proportionally, and when none are specified the original image size is used
 * Updated `api-dump-experimental.txt` with new `DimensionSize` API entries
 * Added a ton of unit tests.

### HTML

image

#### Original images

image

### New features
 * Added support for sized images in Markdown rendering — HTML `` tags with `width` and `height` attributes and native Markdown image attribute blocks (`{width=100 height=5px}`) are now parsed and rendered at the specified dimensions, with proportional scaling when only one dimension is provided. For now we only support pixel/unitless dimensions. (https://github.com/JetBrains/intellij-community/pull/3527)

### DIFF
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
@@ -1,0 +1,43 @@
+package org.jetbrains.jewel.markdown
+
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+
+/**
+ * Represents a size value for an image dimension (width or height).
+ *
+ * Currently only pixel values ([Pixels]) are supported.
+ */
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public sealed interface DimensionSize {
+    /** A loaded image should be exactly [value] pixels in the specified dimension. */
+    @ApiStatus.Experimental
+    @ExperimentalJewelApi
+    @JvmInline
+    public value class Pixels(public val value: Int) : DimensionSize {
+        init {
+            require(value >= 0) { "Value cannot be negative." }
+        }
+
+        override fun toString(): String = "${value}px"
+    }
+
+    // TODO[JEWEL-1333]: Explore percentage sizing further since the first iteration resized the image in relation
+    //  to the percentage instead of making the image fill the required percentage of the available Canvas space.
+}
+
+internal fun String.parseDimensionSize(): DimensionSize? {
+    val trimmed = trim()
+    if (trimmed.isEmpty()) return null
+
+    val number = trimmed.takeWhile { it.isDigit() }
+    val convertedNumber = number.toIntOrNull() ?: return null
+    val normalizedUnit = trimmed.substringAfter(number).trim().trimEnd(';').lowercase()
+
+    return when (normalizedUnit) {
+        "" -> DimensionSize.Pixels(convertedNumber)
+        "px" -> DimensionSize.Pixels(convertedNumber)
+        else -> null
+    }
+}
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
@@ -104,6 +104,22 @@ public sealed interface InlineMarkdown {
         override fun toString(): String = "HtmlInline(content='$content')"
     }
 
+    /**
+     * An inline image node, corresponding to `![alt](source "title")` in Markdown.
+     *
+     * Standard renderer implementations apply the following sizing rules based on [width] and [height]:
+     * - If you specify both sizes, the image is rendered at exactly those dimensions, stretching if the aspect ratio
+     *   differs from the original.
+     * - If you specify only one of them, the other dimension is scaled proportionally to preserve the aspect ratio.
+     * - If you don't specify either, the image is rendered at its intrinsic (loaded) size.
+     *
+     * @param source The URL or path of the image.
+     * @param alt The plain-text alternative description of the image.
+     * @param title The optional tooltip title of the image.
+     * @param inlineContent The parsed inline nodes that make up the alt text.
+     * @param width The optional display width. See [DimensionSize] for supported value types.
+     * @param height The optional display height. See [DimensionSize] for supported value types.
+     */
     @ApiStatus.Experimental
     @ExperimentalJewelApi
     @GenerateDataFunctions
@@ -112,13 +128,33 @@ public sealed interface InlineMarkdown {
         public val alt: String,
         public val title: String?,
         override val inlineContent: List<InlineMarkdown>,
+        public val width: DimensionSize? = null,
+        public val height: DimensionSize? = null,
     ) : InlineMarkdown, WithInlineMarkdown {
         public constructor(
             source: String,
             alt: String,
             title: String?,
             vararg inlineContent: InlineMarkdown,
-        ) : this(source, alt, title, inlineContent.toList())
+            width: DimensionSize? = null,
+            height: DimensionSize? = null,
+        ) : this(source, alt, title, inlineContent.toList(), width, height)
+
+        @Deprecated("Use a constructor with width and height parameters instead.", level = DeprecationLevel.HIDDEN)
+        public constructor(
+            source: String,
+            alt: String,
+            title: String?,
+            vararg inlineContent: InlineMarkdown,
+        ) : this(source, alt, title, inlineContent.toList(), null, null)
+
+        @Deprecated("Use a constructor with width and height parameters instead.", level = DeprecationLevel.HIDDEN)
+        public constructor(
+            source: String,
+            alt: String,
+            title: String?,
+            inlineContent: List<InlineMarkdown>,
+        ) : this(source, alt, title, inlineContent, null, null)
 
         override fun equals(other: Any?): Boolean {
             if (this === other) return true
@@ -130,6 +166,8 @@ public sealed interface InlineMarkdown {
             if (alt != other.alt) return false
             if (title != other.title) return false
             if (inlineContent != other.inlineContent) return false
+            if (width != other.width) return false
+            if (height != other.height) return false
 
             return true
         }
@@ -139,6 +177,8 @@ public sealed interface InlineMarkdown {
             result = 31 * result + alt.hashCode()
             result = 31 * result + (title?.hashCode() ?: 0)
             result = 31 * result + inlineContent.hashCode()
+            result = 31 * result + (width?.hashCode() ?: 0)
+            result = 31 * result + (height?.hashCode() ?: 0)
             return result
         }
 
@@ -147,7 +187,9 @@ public sealed interface InlineMarkdown {
                 "source='$source', " +
                 "alt='$alt', " +
                 "title=$title, " +
-                "inlineContent=$inlineContent" +
+                "inlineContent=$inlineContent, " +
+                "width=$width, " +
+                "height=$height" +
                 ")"
         }
     }
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
@@ -1,5 +1,8 @@
 package org.jetbrains.jewel.markdown.processing
 
+import kotlin.text.first
+import kotlin.text.indexOf
+import kotlin.text.last
 import org.commonmark.node.Code as CMCode
 import org.commonmark.node.Delimited
 import org.commonmark.node.Emphasis as CMEmphasis
@@ -15,9 +18,11 @@ import org.commonmark.parser.beta.ParsedInline
 import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.foundation.util.JewelLogger
+import org.jetbrains.jewel.markdown.DimensionSize
 import org.jetbrains.jewel.markdown.InlineMarkdown
 import org.jetbrains.jewel.markdown.WithInlineMarkdown
 import org.jetbrains.jewel.markdown.WithTextContent
+import org.jetbrains.jewel.markdown.parseDimensionSize
 
 /**
  * Reads all supported child inline nodes into a list of [InlineMarkdown] nodes, using the provided [markdownProcessor]
@@ -25,7 +30,7 @@ import org.jetbrains.jewel.markdown.WithTextContent
  *
  * @param markdownProcessor Used to parse the inline contents as needed.
  * @return A list of the contents as parsed [InlineMarkdown].
- * @see toInlineMarkdownOrNull
+ * @see convertToInlineMarkdown
  */
 @ApiStatus.Experimental
 @ExperimentalJewelApi
@@ -33,10 +38,20 @@ public fun Node.readInlineMarkdown(markdownProcessor: MarkdownProcessor): List<I
     val inlines = buildList {
         var current = this@readInlineMarkdown.firstChild
         while (current != null) {
-            val inline = current.toInlineMarkdownOrNull(markdownProcessor)
+            val (inline, next) = current.convertToInlineMarkdown(markdownProcessor)
+            if (inline is InlineMarkdown.Image && next is CMText) {
+                val result = parseImageWithAttributes(inline, next)
+                if (result != null) {
+                    val (imageWithAttrs, remainder) = result
+                    add(imageWithAttrs)
+                    if (remainder != null) add(InlineMarkdown.Text(remainder))
+                    current = next.next
+                    continue
+                }
+            }
             if (inline != null) add(inline)
 
-            current = current.next
+            current = next
         }
     }
     return markdownProcessor.convertHtmlInlines(inlines)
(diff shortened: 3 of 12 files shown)


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
```diff
@@ -25,18 +29,27 @@ import org.jetbrains.jewel.markdown.WithTextContent
...
+ * @see convertToInlineMarkdown
  */
 @ApiStatus.Experimental
 @ExperimentalJewelApi
 public fun Node.readInlineMarkdown(markdownProcessor: MarkdownProcessor): List<InlineMarkdown> {
     val inlines = buildList {
         var current = this@readInlineMarkdown.firstChild
         while (current != null) {
-            val inline = current.toInlineMarkdownOrNull(markdownProcessor)
+            val (inline, next) = current.convertToInlineMarkdown(markdownProcessor)
```
DanielSouzaBertoldi, line 41: Addresses [this](https://github.com/JetBrains/intellij-community/pull/3442#discussion_r3335254409):

> I'm a bit worried about making toInlineMarkdownOrNull() sibling-aware. Its KDoc says it converts this node and returns only one InlineMarkdown?, but convertToInlineMarkdown() can now peek at this.next for image attributes. Public callers that walk node.next themselves have no way to know that the following text node was logically consumed, so they can get both Image(width=...) and a separate Text("{width=...}").
>
> Could we keep this public single-node API node-local, and reserve sibling-aware image-attribute parsing for readInlineMarkdown() where we can also advance to the returned next node? It should be up to the inline parsers to figure out if they need more nodes.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
```diff
@@ -0,0 +1,55 @@
...
+
+internal fun String.parseDimensionSize(): DimensionSize? {
+    val trimmed = trim()
+    if (trimmed.isEmpty()) return null
+
+    val number = trimmed.takeWhile { it.isDigit() }
+    val convertedNumber = number.toIntOrNull() ?: return null
+    val normalizedUnit = trimmed.substringAfter(number).trimEnd(';')
+
+    return when (normalizedUnit) {
```
DanielSouzaBertoldi, line 38: Addresses [this comment](https://github.com/JetBrains/intellij-community/pull/3442#discussion_r3335254357)

> Can we make this strict instead — accept only a non-negative integer, integer + px, or integer + %, and return null for the rest?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+internal fun String.parseDimensionSize(): DimensionSize? {
+    val trimmed = trim()
+    if (trimmed.isEmpty()) return null
+
+    val number = trimmed.takeWhile { it.isDigit() }
+    val convertedNumber = number.toIntOrNull() ?: return null
+    val normalizedUnit = trimmed.substringAfter(number).trimEnd(';')
```
rock3r: Can we keep this strict, but trim whitespace and normalize case instead of trimming semicolons? I think the parser shape should be roughly: extract the leading number, then `drop(number.length).trim().lowercase()` and accept only `""`, `"px"`, or `"%"`.

That keeps useful leniency for `100`, `100px`, `100PX`, and maybe `100 px`, but still rejects malformed suffixes like `100px;`, `100pxfoo`, or `50%;` instead of treating them as valid sizes.

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -80,25 +86,102 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
+    ): Placeholder {
+        val density = LocalDensity.current
+
+        // At least one dimension is unspecified or requires the image to compute
+        if (imageResult == null) {
+            // For pixel values, we can show them immediately; for percentage we need the image
+            val pixelWidth = (specifiedWidth as? DimensionSize.Pixels)?.value
+            val pixelHeight = (specifiedHeight as? DimensionSize.Pixels)?.value
+
+            return with(density) {
```
DanielSouzaBertoldi, line 128: Addresses [this comment](https://github.com/JetBrains/intellij-community/pull/3442#discussion_r3335254389).

> Nit/question: while loading, could we already reserve the pixel dimensions that are explicitly requested? Right now we only do that when both axes are pixel values. For {width=200} we still use the 0.sp x 1.sp placeholder until the image loads, even though the requested width is already known

rock3r:
One higher-level note on `%` sizes: can we make the release notes explicit that percentages here are relative to the intrinsic loaded image size, not the available/container width like authors would normally expect from browser/HTML behavior?

I'm also wondering whether we should drop `%` support from this PR and come back to it later. The semantics are inherently a bit wrong for Markdown written with browser behavior in mind, and `width=50%` producing "half of the bitmap's original width" may be weird enough that supporting only pixel/unitless sizes for now is the safer first step.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
```diff
@@ -0,0 +1,55 @@
...
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public sealed interface DimensionSize {
+    /** A loaded image should be exactly [value] pixels in the specified dimension. */
+    @ApiStatus.Experimental
+    @ExperimentalJewelApi
+    @JvmInline
+    public value class Pixels(public val value: Int) : DimensionSize {
+        init {
+            require(value >= 0) { "Value cannot be negative." }
```
DanielSouzaBertoldi, line 20: Addresses [this comment](https://github.com/JetBrains/intellij-community/pull/3442#discussion_r3335254385).

> Looks like the public API allows constructing invalid sizes directly, even though the parsers reject them. For example, external code can still create DimensionSize.Pixels(-1) or DimensionSize.Percent(-50), and the renderer will pass those through to placeholder sizing.
>
> Can we enforce the accepted range here, or validate/drop invalid values before rendering? Either way, it would be good for the direct API contract to match the parsed Markdown/HTML path.

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -80,25 +86,102 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
-                    )
-                }
+            computePlaceholder(imageResult = imageResult, specifiedWidth = image.width, specifiedHeight = image.height)
+
+        return InlineTextContent(placeholder) {
+            Image(
+                painter = painter,
+                contentDescription = image.title,
+                modifier = Modifier.fillMaxSize(),
+                contentScale =
```
DanielSouzaBertoldi, line 96: Addresses [this comment](https://github.com/JetBrains/intellij-community/pull/3442#discussion_r3335254396).

> Looks like the placeholder is sized to the requested width/height, but the actual Image still uses the default ContentScale.Fit. So when both dimensions are specified with a different aspect ratio, the image content is fitted inside the box rather than stretched/squished to those dimensions.
>
> The tests/comments describe those cases as stretched, so should this pass contentScale = ContentScale.FillBounds? If preserving aspect ratio is intentional, then I think the tests/docs should avoid implying that explicit two-axis dimensions stretch the image pixels.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
DanielSouzaBertoldi: But wait, should semicolons be treated as invalid? HTML does accept values with `;`

image

That's why I opted to just trim semicolons, to mirror HTML's behavior

DanielSouzaBertoldi: Now, for cases like `100PX` / `100 px` etc I agree. for `100pxfoo` we are already ignoring the result tho 😎

DanielSouzaBertoldi:
@rock3r I mean, given how `InlineTextContent` works we can't properly make the image scale to a specific percentage of the available space in the canvas. At least I couldn't think of any solutions to this 🤔 Given that my approach to this was to mirror HTML as best as we could, we can drop percentage scaling for now, although scaling by percentage is way more useful than just pixels IMO

DanielSouzaBertoldi:
Also: this PR can't be merged until I'm able to build the project and run ApiCheckTest :(

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
rock3r: Ah I misread the code, you're right that we're already ignoring pxfoo! As for the semicolon, yeah it's fine, at this point I prefer aligning better with HTML rendering.

rock3r:
Yeah, I know they need to fix the IJC build first...

Re: proper % support, I _think_ there is no good way to support it, unless we do hacks like capture the width of the text component and do it relative to that, but then again that probably won't work in cases like intrinsic widths etc. I'd rather not ship something today and think it through better than ship something weird, surprising, and overall not very useful (% of image intrinsic size seems a pointless capability?)

DanielSouzaBertoldi:
Alrighty, I'll drop support for percentage now and file an issue to experiment further with this 😎

rock3r:
Looks like the PR text/release notes and API dumps still need one more pass after dropping `%` support.

The description still mentions `Percent`, `{width=100 height=50%}`, `"50%"`, percentage-based dimensions, and proportional scaling from percentages. The release note also still links to #3442 instead of this PR. Can we update those to describe the current scope: unitless/pixel image dimensions only?

The generated API dumps also look stale relative to the latest source: `core-api-0.37.0.txt` and `api-dump-experimental.txt` still include `DimensionSize.Percent`, and `api-dump-experimental.txt` still has the old `InlineMarkdown.Image` constructor order with `DimensionSize, DimensionSize, List`. Can we regenerate the dumps from the latest source after the `%` removal?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/DimensionSize.kt
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        override fun toString(): String = "${value}px"
+    }
+
+    // TODO[JEWEL-1333]: Explore percentage sizing further before shipping.
```
rock3r: Can we avoid shipping the commented-out `Percent` API block? The TODO/follow-up issue reference is useful, but keeping the whole disabled public API shape in source makes this look half-shipped and easy to accidentally resurrect stale details from later.

I'd rather keep just a short TODO explaining why percentage support is deferred, and let the follow-up issue/design bring back the API when the semantics are settled.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
```diff
@@ -106,3 +159,32 @@ internal fun List<InlineMarkdown>.renderAsSimpleText(): String = buildString {
...
+    if (width == null && height == null) return null // Cases like { random text } should be ignored
+
+    val remainder = textLiteral.substring(endAttrIdx + 1)
+    // If there's no text after the attributes we can safely skip the CMText node. Otherwise, remove the
+    // attributes from this CMText literal and return the remainder.
+    val nextNode =
+        if (remainder.isBlank()) {
+            next.next
+        } else {
+            next.literal = remainder
```
rock3r: Note: this makes `readInlineMarkdown()` mutate the underlying CommonMark node, so it is not idempotent anymore. If a caller reuses the same parsed tree and calls `readInlineMarkdown()` again, the attribute block has already been stripped from this `CMText` and the second conversion can produce different `InlineMarkdown` output.

Can we avoid mutating the source node here to begin with? If that gets too complicated, could we at least document on `readInlineMarkdown()` that it may mutate the input tree while consuming image attributes? At least it's bad but not surprising

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
```diff
@@ -104,6 +104,16 @@ public sealed interface InlineMarkdown {
...
     }
 
+    /**
+     * An inline image node, corresponding to `![alt](source "title")` in Markdown.
+     *
+     * @param source The URL or path of the image.
+     * @param alt The plain-text alternative description of the image.
+     * @param title The optional tooltip title of the image.
+     * @param inlineContent The parsed inline nodes that make up the alt text.
+     * @param width The optional display width. See [DimensionSize] for additional details on the underlying behavior.
```
rock3r: Can we make the public docs for the new sizing API a bit more explicit? The current KDoc points to `DimensionSize`, which is fine, but callers constructing `InlineMarkdown.Image` directly still have to infer the important behavior from the renderer/tests (which they won't do): pixel-only support for now, proportional scaling when only one axis is specified, and exact/fill-bounds sizing when both axes are specified.

This probably does not all have to live on `width`/`height`; documenting the standard renderer semantics on `DimensionSize`, `InlineMarkdown.Image`, or `ImageRendererExtension.renderImageContent()` would work too.

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -80,25 +86,94 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
+    private fun computePlaceholder(
+        imageResult: SuccessResult?,
+        specifiedWidth: DimensionSize?,
+        specifiedHeight: DimensionSize?,
+    ): Placeholder {
+        val density = LocalDensity.current
+
+        // At least one dimension is unspecified or requires the image to compute
+        if (imageResult == null) {
+            // For pixel values, we can show them immediately; for percentage we need the image
```
rock3r: Nit: this still mentions percentage sizes, but `Percent` support was removed for this iteration. Could we rephrase this to just say that known pixel dimensions can be reserved while loading, and unspecified dimensions fall back to the minimal placeholder?

rock3r (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/ProcessingUtil.kt
DanielSouzaBertoldi: Oh right, good catch! Yeah I can make the readInlineMarkdown idempotent once again, no biggie. Thanks for pointing this out :)

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/InlineMarkdown.kt
DanielSouzaBertoldi: Tell me what you think of the update KDocs please :)

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
DanielSouzaBertoldi: Yep, + there's no need to cast the variable to DimensionSize.Pixel since it's the only type we support now 😛 updated the code

rock3r (approved):
LGTM, thanks mate <3 One nit: you added some extra whitespace in `DimensionSize` L6, L8; `InlineMarkdown` L109; and `Coil3ImageRendererExtensionImpl` L123. Can you clean it up?

nebojsa-vuksic (approved):

DanielSouzaBertoldi:
Ready to merge

DanielSouzaBertoldi:
Ready to merge

PR #3529, opened 2026-06-03
Disable AI agents CLI install prompts in Android Studio

jrlogsdon:
Change-Id: Ibf1f1e39cfa66b5575b0050ba852fc5983c63c1d

### DIFF
--- a/plugins/terminal/frontend/src/com/intellij/terminal/frontend/action/TerminalAgentsSelectorPresentationUtil.kt
+++ b/plugins/terminal/frontend/src/com/intellij/terminal/frontend/action/TerminalAgentsSelectorPresentationUtil.kt
@@ -4,6 +4,7 @@ package com.intellij.terminal.frontend.action
 import com.intellij.icons.AllIcons
 import com.intellij.openapi.actionSystem.Presentation
 import com.intellij.openapi.actionSystem.ex.ActionUtil
+import com.intellij.openapi.application.ex.ApplicationInfoEx
 import com.intellij.openapi.util.text.HtmlChunk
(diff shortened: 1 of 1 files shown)


plugins/terminal/frontend/src/com/intellij/terminal/frontend/action/TerminalAgentsSelectorPresentationUtil.kt
```diff
@@ -34,6 +35,11 @@ object TerminalAgentsSelectorPresentationUtil {
     mode: TerminalAgentMode,
     showNewBadge: Boolean,
   ) {
+    if (PlatformUtils.isAndroidStudio() &&
```
jreznot: I would suggest checking IDE vendor instead, so we do not affect any fork

jreznot: `ApplicationInfoEx.getInstanceEx().isVendorJetBrains`

jreznot:
Merging on our side!

KonstantinHudyakov:
Hello! Thank you for highlighting this issue!
In case of applying your change, AI Agents button would still be shown and would open empty popup if there are no available agents.
So, I've merged the fix that adds this condition in the different place:
https://github.com/JetBrains/intellij-community/commit/915d9f49af6cf6aae1ae4c1c5567a7014d5a6231

jrlogsdon:
Thanks for the clarification and quick fix! We will cherry-pick that change instead!

PR #3531, opened 2026-06-04
Fix NullPointer in StatisticsEventLogProviders at startup

jrlogsdon:
This was introduced in IntelliJ commit: 7a6e6be7bd0 and causes an issue for Android Studio at startup due to property initialization order for non-JetBrains IDEs (calculateEventLogProvider behaves differently for Android Studio compared to IntelliJ).

Change-Id: Ibddd632793a86995c3f193c46f1cc1e3b23e098d

### DIFF
--- a/platform/statistics/src/com/intellij/internal/statistic/eventLog/StatisticsEventLogProvidersHolder.kt
+++ b/platform/statistics/src/com/intellij/internal/statistic/eventLog/StatisticsEventLogProvidersHolder.kt
@@ -18,15 +18,15 @@ import java.util.concurrent.atomic.AtomicReference
 
 @Service(Service.Level.APP)
 internal class StatisticsEventLogProvidersHolder(coroutineScope: CoroutineScope) {
+  // Cache for empty providers by [recorderId] to avoid creating new instances on every call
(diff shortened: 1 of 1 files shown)


jreznot:
> calculateEventLogProvider behaves differently for Android Studio compared to IntelliJ

Does it mean you have patched this place in sources? Could we please incorporate this behavior branch to the platform if so? At the moment it is too fragile to maintain this contract from our side, given that we have no any test covering it

jrlogsdon:
This is the patch that we have created on our side that unblocked us for the 2026.1 merge. We would like to get this into intellij-community to prevent divergence in future merges.

Luckily, the failure was pretty immediate on our side so tests caught it (plugin loading failed).

gharrma:
Re: "calculateEventLogProvider behaves differently for Android Studio compared to IntelliJ"

To clarify, the behavior difference is not due to a patch, but rather it's due to the fact that this function queries some plugin vendor metadata, and Android Studio has different metadata resulting in a different code path being taken.

jreznot:
I understand that, but we have no any chance here to prevent it from happening in the future, since there is no any test on our side. Is there any way for us to have one?

jreznot:
Merged

PR #3533, opened 2026-06-04
IDEA-384446 Allow executing different Gradle test tasks on a module

bric3:
### Summary

This pull request ensures users can execute different Gradle test tasks on the same module from the context menu without having to clear the existing run configuration. Precisely, always shows the task chooser (if multiple test tasks are available). 

Also, this change adjusts the naming conventions for the run configurations with the test task name to improve usability.

Screenshot 2026-06-04 at 16 06 39
Screenshot 2026-06-04 at 16 07 20

Should fix https://youtrack.jetbrains.com/issue/IDEA-384446/Gradle-Cannot-Execute-different-test-task-on-a-module-after-running-one

### Important Changes

- Prevent reusing existing context configurations before showing the task chooser when multiple Gradle test tasks are available.
- Maintain naming consistency: the chooser action retains the name of the default test task, and the selected task determines the new configuration name.
- Avoid appending a unique suffix if an equivalent configuration for the selected test task already exists.

### DIFF
--- a/plugins/gradle/java/src/execution/test/runner/AbstractGradleTestRunConfigurationProducer.kt
+++ b/plugins/gradle/java/src/execution/test/runner/AbstractGradleTestRunConfigurationProducer.kt
@@ -2,6 +2,7 @@
 package org.jetbrains.plugins.gradle.execution.test.runner
 
 import com.intellij.execution.JavaRunConfigurationExtensionManager
+import com.intellij.execution.RunManager
 import com.intellij.execution.actions.ConfigurationContext
 import com.intellij.execution.actions.ConfigurationFromContext
 import com.intellij.openapi.util.Ref
@@ -16,6 +17,8 @@ import org.jetbrains.plugins.gradle.util.cmd.node.GradleCommandLineTasks
 import org.jetbrains.plugins.gradle.util.createTestWildcardFilter
 import java.util.StringJoiner
 import java.util.function.Consumer
+
+private const val DEFAULT_TEST_TASK_NAME = "test"
 
 abstract class AbstractGradleTestRunConfigurationProducer<E : PsiElement, Ex : PsiElement> : GradleTestRunConfigurationProducer() {
 
@@ -33,9 +36,23 @@ abstract class AbstractGradleTestRunConfigurationProducer<E : PsiElement, Ex : P
 
   protected abstract fun getAllTestsTaskToRun(context: ConfigurationContext, element: E, chosenElements: List<Ex>): List<TestTasksToRun>
 
+  override fun findOrCreateConfigurationFromContext(context: ConfigurationContext): ConfigurationFromContext? {
+    val configurationFromContext = super.findOrCreateConfigurationFromContext(context) ?: return null
+    val element = getElement(context) ?: return configurationFromContext
+    if (allTestsTaskToRun(context, element).map { it.tasksToRun.testName }.toSet().size > 1) {
+      (configurationFromContext.configuration as GradleRunConfiguration).name =
+        suggestConfigurationName(context, element, emptyList())
+    }
+    return configurationFromContext
+  }
+
   private fun getAllTasksAndArguments(context: ConfigurationContext, element: E, chosenElements: List<Ex>): List<GradleCommandLineTasks> {
     return getAllTestsTaskToRun(context, element, chosenElements)
       .map { it.toTasksAndArguments() }
+  }
+
+  private fun allTestsTaskToRun(context: ConfigurationContext, element: E): List<TestTasksToRun> {
+    return getAllTestsTaskToRun(context, element, emptyList())
   }
 
   override fun doSetupConfigurationFromContext(
@@ -69,7 +86,11 @@ abstract class AbstractGradleTestRunConfigurationProducer<E : PsiElement, Ex : P
     val module = context.module ?: return false
     val externalProjectPath = resolveProjectPath(module) ?: return false
     val element = getElement(context) ?: return false
-    val allTasksAndArguments = getAllTasksAndArguments(context, element, emptyList())
+    val allTestsTaskToRun = allTestsTaskToRun(context, element)
+    if (allTestsTaskToRun.map { it.tasksToRun.testName }.toSet().size > 1) {
+      return false
+    }
+    val allTasksAndArguments = allTestsTaskToRun.map { it.toTasksAndArguments() }
     val tasksAndArguments = configuration.commandLine.tasks.tokens
     return externalProjectPath == configuration.settings.externalProjectPath &&
            tasksAndArguments.isNotEmpty() && allTasksAndArguments.isNotEmpty() &&
@@ -96,14 +117,44 @@ abstract class AbstractGradleTestRunConfigurationProducer<E : PsiElement, Ex : P
           .mapValues { it.value.map(TestTasksToRun::testFilter).toSet() }
           .map { createTasksAndArguments(it.key, it.value) }
 
-        runConfiguration.name = suggestConfigurationName(context, element, elements)
-        setUniqueNameIfNeeded(project, runConfiguration)
         runConfiguration.settings.taskNames = chosenTasksAndArguments.flatMap { it.tokens }
         runConfiguration.settings.scriptParameters = if (chosenTasksAndArguments.size > 1) "--continue" else ""
+        runConfiguration.name = suggestConfigurationName(context, element, elements)
+          .withGradleTestTaskName(chosenTestsToRun)
+        if (!hasSameExistingConfiguration(runConfiguration)) {
+          setUniqueNameIfNeeded(project, runConfiguration)
+        }
 
         super.onFirstRun(configuration, context, startRunnable)
       }
     }
+  }
+
+  private fun String.withGradleTestTaskName(chosenTestsToRun: List<List<TestTasksToRun>>): String {
+    val testTaskName = chosenTestsToRun.flatten().map { it.tasksToRun.testName }.distinct().singleOrNull()
+    if (testTaskName == null || testTaskName == DEFAULT_TEST_TASK_NAME) return this
+    if (endsWith(".$DEFAULT_TEST_TASK_NAME'")) {
+      return removeSuffix(".$DEFAULT_TEST_TASK_NAME'") + ".$testTaskName'"
+    }
+    if (endsWith("'")) {
+      return dropLast(1) + ".$testTaskName'"
+    }
+    return "$this.$testTaskName"
+  }
+
+  private fun hasSameExistingConfiguration(configuration: GradleRunConfiguration): Boolean {
+    val taskNames = configuration.commandLine.tasks.tokens
+    val scriptParameters = configuration.settings.scriptParameters
+    return RunManager.getInstance(configuration.project)
+      .getConfigurationSettingsList(configurationFactory.type)
+      .any { settings ->
+        val existingConfiguration = settings.configuration as? GradleRunConfiguration ?: return@any false
+        existingConfiguration !== configuration &&
+        existingConfiguration.name == configuration.name &&
+        existingConfiguration.settings.externalProjectPath == configuration.settings.externalProjectPath &&
+        existingConfiguration.commandLine.tasks.tokens == taskNames &&
+        existingConfiguration.settings.scriptParameters == scriptParameters
+      }
   }
 
   /**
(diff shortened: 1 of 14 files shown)


bric3:
My minimal reproducer project:

gradle/wrapper/gradle-wrapper.properties
```properties
distributionBase=GRADLE_USER_HOME
distributionPath=wrapper/dists
distributionUrl=https\://services.gradle.org/distributions/gradle-9.5.1-bin.zip
zipStoreBase=GRADLE_USER_HOME
zipStorePath=wrapper/dists
```

gradle/gradle-daemon-jvm.properties
```properties
toolchainVersion=21
```

settings.gradle.kts
```Gradle Kotlin DSL
plugins {
    id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
}

rootProject.name = "malformed-parameters-oracle-8"
```

build.gradle.kts
```Gradle Kotlin DSL
plugins {
    id("java")
}

group = "io.github.bric3.reproducer"
version = "1.0-SNAPSHOT"

repositories {
    mavenCentral()
}

dependencies {
    implementation("net.bytebuddy:byte-buddy:1.18.1")
    testImplementation(platform("org.junit:junit-bom:5.10.0"))
    testImplementation("org.junit.jupiter:junit-jupiter")
    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}

java {
    toolchain {
        // The bug starts reproducing when compiled from JDK 21 (JDK 20 passes)
        languageVersion.set(JavaLanguageVersion.of(21))
    }
}

val oracle8Launcher = javaToolchains.launcherFor {
    languageVersion = JavaLanguageVersion.of(8)
    vendor = JvmVendorSpec.ORACLE
}
val corretto8Launcher = javaToolchains.launcherFor {
    languageVersion = JavaLanguageVersion.of(8)
    vendor = JvmVendorSpec.AMAZON
}
val zulu8Launcher = javaToolchains.launcherFor {
    languageVersion = JavaLanguageVersion.of(8)
    vendor = JvmVendorSpec.AZUL
}

tasks {
    withType<JavaCompile>().configureEach {
        options.release.set(8)
        // workaround for JDK-8170063 JDK-8058322, maybe JDK-8341145 in Oracle 8
        options.compilerArgs.add("-parameters")
    }
}

testing {
    suites {
        val oracle8Tests by registering(JvmTestSuite::class) {
            targets {
                all {
                    testTask.configure {
                        testClassesDirs = sourceSets.test.get().output.classesDirs
                        classpath = sourceSets.test.get().runtimeClasspath
                        javaLauncher = oracle8Launcher
                        useJUnitPlatform()
                    }
                }
            }
        }
        val corretto8Tests by registering(JvmTestSuite::class) {
            targets {
                all {
                    testTask.configure {
                        testClassesDirs = sourceSets.test.get().output.classesDirs
                        classpath = sourceSets.test.get().runtimeClasspath
                        javaLauncher = corretto8Launcher
                        useJUnitPlatform()
                    }
                }
            }
        }
        val zulu8Tests by registering(JvmTestSuite::class) {
            targets {
                all {
                    testTask.configure {
                        testClassesDirs = sourceSets.test.get().output.classesDirs
                        classpath = sourceSets.test.get().runtimeClasspath
                        javaLauncher = zulu8Launcher
                        useJUnitPlatform()
                    }
                }
            }
        }
    }
}

tasks.check {
    dependsOn(testing.suites)
}
```

src/test/java/SomeTest.java
```java
import org.junit.jupiter.api.Test;

public class ShouldNotRaiseMalformedParametersExTest {
    @Test
    void should_assert() { }
}
```

nskvortsov:
Thank you for a proposed solution.

I believe we can use something like this. 
I see a blocking issue in this PR, though.  Matching an existing configuration from the context should still be supported because it may have additional configuration made by a user via UI.
With the current solution, if a created configuration is adjusted e.g., a `-Dkey=val` parameter added to the command line, the followup invocatoin will spam new run configurations.
image

Please investigate and see, if this can be fixed.

plugins/gradle/java/src/execution/test/runner/AbstractGradleTestRunConfigurationProducer.kt
```diff
@@ -96,16 +117,46 @@ abstract class AbstractGradleTestRunConfigurationProducer<E : PsiElement, Ex : P
...
+        if (!hasSameExistingConfiguration(runConfiguration)) {
+          setUniqueNameIfNeeded(project, runConfiguration)
+        }
 
         super.onFirstRun(configuration, context, startRunnable)
       }
     }
   }
 
+  private fun String.withGradleTestTaskName(chosenTestsToRun: List<List<TestTasksToRun>>): String {
```
nskvortsov: I find the proposed naming for a run configuration to be confusing. When applied to a class it gives a false impression there is some method: `T1.oracle8Tests`.

IDEA should be more explicit about what happens. Something like `$this via :$testTaskName and 3 other tasks`. After all, it is a user visible text, not an internal FQN

plugins/gradle/java/testSources/execution/test/producer/GradleTestRunConfigurationProducerTest.kt
```diff
@@ -231,6 +231,30 @@ class GradleTestRunConfigurationProducerTest : GradleTestRunConfigurationProduce
     assertGutterRunActionsSize(projectData["project"]["org.example.TestCaseWithMain"]["test2"].element, 2)
   }
 
+  @Test
+  fun `test configuration with one of multiple test tasks keeps chosen existing configuration name`() {
```
nskvortsov: Thank you for providing a test. Please check other tests in the same category. For me, the `GradleTestRunConfigurationProducerTest#test configuration tests for directory` fails and needs adjustment.

nskvortsov (changes requested):

bric3:
@nskvortsov Good catch on properties, I completely missed that scenario. It seems there's a challenge though: the run configuration action looks up an existing configuration before invoking `onFirstRun`

https://github.com/JetBrains/intellij-community/blob/94a03c8bc8e4111911f70f7a7b8a1b80fbe74e62/platform/execution-impl/src/com/intellij/execution/actions/BaseRunConfigurationAction.java#L134-L139

but the Gradle task chooser runs inside `onFirstRun`

   https://github.com/JetBrains/intellij-community/blob/94a03c8bc8e4111911f70f7a7b8a1b80fbe74e62/plugins/gradle/java/src/execution/test/runner/AbstractGradleTestRunConfigurationProducer.kt#L89

which is called within action's `perform` 

https://github.com/JetBrains/intellij-community/blob/94a03c8bc8e4111911f70f7a7b8a1b80fbe74e62/platform/execution-impl/src/com/intellij/execution/actions/BaseRunConfigurationAction.java#L170

It also captures the settings before the chooser callback 

https://github.com/JetBrains/intellij-community/blob/94a03c8bc8e4111911f70f7a7b8a1b80fbe74e62/platform/execution-impl/src/com/intellij/execution/actions/BaseRunConfigurationAction.java#L168-L169

As such this requires some changes in the base run action. And possibly in the subclass RunContextAction as well.

----

Also sorry, for the failing tests, I struggled at that time to have the bazel command working and probably skewed my way of doing the tests.

plugins/gradle/java/src/execution/test/runner/AbstractGradleTestRunConfigurationProducer.kt
bric3: I think when selecting a test task that not named `test`, then it's a good idea to display the test task first, as it is an intentional choice, also I think it should appear first, so it may have less chance to be trimmed.

Image

vs if the task appears first

Image

And, I overlooked the individual test selection, either classes or methods, (I simply forgot this case), today it's named like this (and adds a ` (N)` suffix when running another test task)

Image.

I think having the test task should also appear.

That said to have a better name, even if it's only user visible I think it has some impacts on the class hierarchy of the producers (e.g. the kotlin ones `AbstractKotlin*ClassGradleConfigurationProducer`).

plugins/gradle/java/testSources/execution/test/producer/GradleTestRunConfigurationProducerTest.kt
bric3: Sorry, the bazel project had some issues when I tried to work on this. I somewhat managed to dirty patch the build file to try running tests. This must have slipped through.

bric3 (commented):
Thanks a lot for pointing out all of these. However, fixing these this makes the PR touch much more files, this broader change makes me uneasy 😓.

I somewhat found a way to handle configuration matching based on the choosen task. But again this required me to change other files in the platform module (it's related to how configuration are selected and applied in `onFirstRun`).

nskvortsov:
Some details are important and cannot be omitted. And covering them may require larger efforts. Thank you for your time and attention to this matter.

bric3:
@nskvortsov Btw I already worked, toward the first fixes (point 1), I'm moving forward to address point 2.

1. > With the current solution, if a created configuration is adjusted e.g., a -Dkey=val parameter added to the command line, the followup invocatoin will spam new run configurations.

2. > I find the proposed naming for a run configuration to be confusing. When applied to a class it gives a false impression there is some method: `T1.oracle8Tests`.

bric3:
@nskvortsov Done

Note, as I started working on top of this e633b6f75ff7f1a45c3f087daae3d05c3e7999bd `master` commit, I had to make a dirty patch to the `BUILD.bazel` run relevant tests, but I may have missed some. I hope the Bazel setup on later `master` commits is fixed.

<details><summary><code>plugins/gradle/java/BUILD.bazel</code> diff</summary>

```diff
Index: plugins/gradle/java/BUILD.bazel
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/plugins/gradle/java/BUILD.bazel b/plugins/gradle/java/BUILD.bazel
--- a/plugins/gradle/java/BUILD.bazel	(revision Staged)
+++ b/plugins/gradle/java/BUILD.bazel	(date 1785459836298)
@@ -380,3 +380,17 @@
     runtime_deps = [":java-tests_test_lib"],
 )
 ### auto-generated section `test intellij.gradle.java.tests` end
+
+# Temporary local target for focused verification in this checkout.
+jps_test(
+    name = "java-tests_test_dirty",
+    data = ["@kotlin_test_deps//:kotlin-dist-for-ide.jar"],
+    jvm_flags = [
+        "-Didea.load.plugins.id=com.intellij.java,com.intellij.gradle,org.jetbrains.plugins.gradle,JUnit,org.intellij.groovy,com.intellij.groovy.scripting,intellij.bookmarks.plugin",
+    ],
+    runtime_deps = [
+        "//:main",
+        ":java-tests_test_lib",
+        "//plugins/kotlin/base/pluginTestFramework:kotlin-base-plugin-testFramework",
+    ],
+)
```

</details>

The naming rule I ended p with is: when multiple Gradle test tasks are available, put the explicitly selected task first in the configuration name. The remainder describes the test scope : 

| Producer/scope | Pattern | Example |
|---|---|---|
| Directory/module | `<tasks> in [<Gradle path>]` | `myFancyTest in [:foo:bar]` |
| Package | `<tasks> in <package>` | `myFancyTest in com.example.tests` |
| Test class | `<tasks> for <class>` | `myFancyTest for FooTest` |
| Test method | `<tasks> for <class.method>` | `myFancyTest for FooTest.shouldWork` |
| Pattern/multi-selection | `<tasks> for <tests>` | `myFancyTest for FooTest and 2 more` |
| Kotlin MPP class/method | Same `for` pattern | `jvmTest for FooTest.shouldWork` |

_I tried to avoid the ` in ` / ` for ` variations (at least to simplify the code a bit) but it looked odd to me when reading the config name._

When there are multiple selected elements, the first element name is shown then followed by `and N more`.

The legacy name is retained when **only one test task** is available, which should be a task named `test`. In this case it would look like, `FooTest`, `FooTest.shouldWork`, or `Tests in 'foo'`.
_Maybe that choice can be revisited._

Naming needs to happen after task selection. If task selection finds and reuses an existing configuration, its custom name and edited options are preserved as you noticed in a previous comment rather than replaced. 

The common formatting lives in `GradleTestRunConfigurationProducer`, with the producer-specific scopes supplied by the directory, package, class, method, pattern, and Kotlin MPP producers.

nskvortsov:
I made a couple of changes on top of this, covering some edge cases in Kotlin Multiplatform and Android run configurations. Running the internal CI and code review.

PR #3535, opened 2026-06-04
IDEA-390181 Proxy credentials encrypted with hardcoded AES key in source code

ishaq2321:
## Summary

Replace the hardcoded AES key in HttpConfigurable.java with a randomly generated per-installation key.

## Problem

HttpConfigurable encrypts proxy credentials stored on disk at proxy.settings.pwd using a hardcoded AES key discoverable from source code. Anyone with access to the repository or a decompiled binary can decrypt stored proxy usernames and passwords across all installations.

## Fix

1. Generate a random 128-bit AES key per installation, persisted at proxy.settings.pwd.key
2. Existing credentials encrypted with the old key are automatically migrated on first access (try new key, fall back to legacy key, re-encrypt with new key)
3. New installations get a unique random key with no legacy migration needed

## YouTrack

https://youtrack.jetbrains.com/issue/IDEA-390181

### DIFF
--- a/platform/platform-api/src/com/intellij/util/net/HttpConfigurable.java
+++ b/platform/platform-api/src/com/intellij/util/net/HttpConfigurable.java
@@ -50,6 +50,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.security.Key;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


platform/platform-api/src/com/intellij/util/net/HttpConfigurable.java
```diff
@@ -74,6 +75,31 @@
 public class HttpConfigurable implements PersistentStateComponent<HttpConfigurable>, Disposable {
   private static final Logger LOG = Logger.getInstance(HttpConfigurable.class);
   private static final Path PROXY_CREDENTIALS_FILE = PathManager.getOptionsDir().resolve("proxy.settings.pwd");
+  private static final Path PROXY_ENCRYPTION_KEY_FILE = PathManager.getOptionsDir().resolve("proxy.settings.pwd.key");
```
trespasserw, line 78: Thank you for your contribution! Unfortunately, we are not interested in improving this code, because it is expected to be removed in the near future.

ishaq2321:
The PR is unnecessary for the maintainers will be removing this code in near future.

PR #3534, opened 2026-06-04
IDEA-382706 Auto-detect JUnit report files upon opening

bric3:
Enhances the JUnit plugin by implementing the following features:
- Automatically recognize JUnit-style XML report files when opened.
- Display editor notifications that offer import options for such files.
- Add a new context menu action to view reports as test results.

Fixes https://youtrack.jetbrains.com/issue/IDEA-382706/Auto-detect-a-JUnit-report-file-when-opening-it

Screenshot 2026-06-04 at 18 26 24
image

PR #3536, opened 2026-06-05
IDEA-354352 preserve JUnit XML report details

bric3:
Importing external JUnit XML reports previously relied on a single SAX text buffer, so `failure` text could be overwritten by later output or `metadata` elements. Maven and Gradle reports also commonly use a single `testsuite` root, which the Ant/JUnit XML extension did not recognize before the detection work.

Keep the existing SM runner import path, but make the Ant/JUnit XML handler capture result elements, output, properties, and Surefire retry/flaky elements independently. Extra metadata is surfaced as ordinary test output so existing report presentation can show it without new UI fields.

Fixes 
* https://youtrack.jetbrains.com/issue/IDEA-354352/IDEA-cant-import-test-xml-produced-by-Gradle
* https://youtrack.jetbrains.com/issue/IDEA-380741/Import-tests-form-file-is-broken-on-this-testsuite-xml-report

Screenshot 2026-06-05 at 17 36 58

Screenshot 2026-06-05 at 17 37 20

By the way one of the only "spec" I found is https://github.com/testmoapp/junitxml, so I may have missed something.

----

I wonder if `AntTestContentHandler` should be renamed to something more JUnit report related than related to "Ant".

Related to 
* #3534

### DIFF
--- a/java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
+++ b/java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
@@ -25,6 +25,7 @@ import org.xml.sax.helpers.DefaultHandler;
 import java.io.Reader;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.List;
 
 public class AntTestContentHandler extends DefaultHandler {
@@ -79,7 +80,7 @@ public class AntTestContentHandler extends DefaultHandler {
   private final List<TestProblem> myProblems = new ArrayList<>();
   private TestIgnoredEvent myIgnoredEvent;
   private int myPropertiesLevel;
-  private final ArrayDeque<TextElement> myTextElements = new ArrayDeque<>();
+  private final Deque<TextElement> myTextElements = new ArrayDeque<>();
 
   public AntTestContentHandler(GeneralTestEventsProcessor processor) {
     myProcessor = processor;
--- a/java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
+++ b/java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
@@ -23,6 +23,9 @@ import org.xml.sax.SAXException;
 import org.xml.sax.helpers.DefaultHandler;
 
 import java.io.Reader;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.List;
 
 public class AntTestContentHandler extends DefaultHandler {
   public static final class AntTestOutputExtension implements ImportTestOutputExtension {
@@ -38,11 +41,11 @@ public class AntTestContentHandler extends DefaultHandler {
 
         @Override
         public void startElement(String name, String nsPrefix, String nsURI, String systemID, int lineNr) {
-          rooName[0] = name;
+          rooName[0] = getElementName(name);
         }
       });
-      return TESTSUITES.equals(rooName[0]) ? new AntTestContentHandler(processor)
-                                           : null;
+      return TESTSUITES.equals(rooName[0]) || TESTSUITE.equals(rooName[0]) ? new AntTestContentHandler(processor)
+                                                                            : null;
     }
   }
 
@@ -55,18 +58,28 @@ public class AntTestContentHandler extends DefaultHandler {
   private static final String DURATION = "time";
   private static final String ERROR = "error";
   private static final String FAILURE = "failure";
+  private static final String RERUN_ERROR = "rerunError";
+  private static final String RERUN_FAILURE = "rerunFailure";
+  private static final String FLAKY_ERROR = "flakyError";
+  private static final String FLAKY_FAILURE = "flakyFailure";
   private static final String SKIPPED = "skipped";
   private static final String IGNORED = "ignored";
   private static final String OUT = "system-out";
   private static final String ERR = "system-err";
+  private static final String PROPERTIES = "properties";
+  private static final String PROPERTY = "property";
+  private static final String MESSAGE = "message";
+  private static final String TYPE = "type";
+  private static final String VALUE = "value";
 
   private final GeneralTestEventsProcessor myProcessor;
   private final Stack<String> mySuites = new Stack<>();
   private String myCurrentTest;
   private String myDuration;
-  private String myStatus;
-  private final StringBuilder currentValue = new StringBuilder();
-  private boolean myErrorOutput = false;
+  private final List<TestProblem> myProblems = new ArrayList<>();
+  private TestIgnoredEvent myIgnoredEvent;
+  private int myPropertiesLevel;
+  private final ArrayDeque<TextElement> myTextElements = new ArrayDeque<>();
 
   public AntTestContentHandler(GeneralTestEventsProcessor processor) {
     myProcessor = processor;
@@ -74,7 +87,8 @@ public class AntTestContentHandler extends DefaultHandler {
 
   @Override
   public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
-    if (TESTSUITE.equals(qName)) {
+    String name = getElementName(qName);
+    if (TESTSUITE.equals(name)) {
       String nameValue = attributes.getValue(NAME);
       final String suiteName = nameValue == null ? "" : StringUtil.unescapeXmlEntities(nameValue);
       String packageValue = attributes.getValue(PACKAGE);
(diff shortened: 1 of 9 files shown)


madlexa:
Hi! Really nice PR.

> I wonder if AntTestContentHandler should be renamed to something more JUnit report related than related to "Ant".

The "Ant" name is historical. This is the classic Ant <junitreport> / JUnit XML format, same convention as the TeamCity xml-report plugin's [AntJUnitReportParser](https://github.com/JetBrains/teamcity-xml-tests-reporting/blob/master/agent/src/jetbrains/buildServer/xmlReportPlugin/parsers/antJUnit/AntJUnitReportParser.java)

java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
```diff
@@ -55,26 +58,37 @@ public void startElement(String name, String nsPrefix, String nsURI, String syst
...
   private final Stack<String> mySuites = new Stack<>();
   private String myCurrentTest;
   private String myDuration;
-  private String myStatus;
-  private final StringBuilder currentValue = new StringBuilder();
-  private boolean myErrorOutput = false;
+  private final List<TestProblem> myProblems = new ArrayList<>();
+  private TestIgnoredEvent myIgnoredEvent;
+  private int myPropertiesLevel;
+  private final ArrayDeque<TextElement> myTextElements = new ArrayDeque<>();
```
madlexa: Could you please declare the field as `Deque<TextElement>` rather than the `ArrayDeque` impl?

java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
```diff
@@ -84,77 +98,281 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+                                             element.label,
+                                             text);
+        myProblems.add(problem);
       }
-      else {
-        myProcessor.onUncapturedOutput(currentText, myErrorOutput ? ProcessOutputTypes.STDERR : ProcessOutputTypes.STDOUT);
+    }
+  }
+
+  private void emitProperty(@Nullable String name, @Nullable String value) {
```
madlexa, line 223: It looks like `emitOutput`, `emitProperty`, `createTestFailedEvent`, etc. mostly repeat what the platform already does in the "service messages -> events" path. You can see this in (MyServiceMessageVisitor)[https://github.com/JetBrains/intellij-community/blob/86269f133a51cdafeff50e48bb1314cf896b70f2/platform/smRunner/src/com/intellij/execution/testframework/sm/runner/OutputToGeneralTestEventsConverter.java#L373]:

- `createTestFailedEvent` -> `visitTestFailed`
- `emitOutput` -> `visitTestStdOut` / `visitTestStdErr`
- suite-level output -> `onUncapturedOutput`
- started / finished / ignored → `visitTestStarted` / `visitTestFinished` / `visitTestIgnored`

What do you think about producing [##teamcity](https://www.jetbrains.com/help/teamcity/service-messages.html) via [ServiceMessageBuilder](https://github.com/JetBrains/intellij-community/blob/master/platform/smRunner/src/com/intellij/execution/testframework/sm/ServiceMessageBuilder.java) and feeding [OutputToGeneralTestEventsConverter](https://github.com/JetBrains/intellij-community/blob/master/platform/smRunner/src/com/intellij/execution/testframework/sm/runner/OutputToGeneralTestEventsConverter.java)`.process(text, key)`? Then these methods mostly go away, and the `expected/actual` **diff** comes for free.


Example XML->service-messages mapping (as a reference, NOT as a dependency): https://github.com/JetBrains/teamcity-xml-tests-reporting

bric3:
Thanks @madlexa, is there anything I can do to help move this forward ?

madlexa (changes requested):

java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
bric3: Hmm, I'm not sure I'm ready to follow that path, for one reason mainly, the proposed change is mostly about opening and seeing details of the xml report within IJ, i.e. when opening them from local report file ; and I have no experience with Teamcity.

Please correct me if I'm off track as I don't fathom as well the whole platform.

That said, IIUC the _service-message visitor_ allows to translate TeamCity-style text messages into test-runner events. And that would mean something like that I believe (more or less): 

1. junit xml report gets parsed
2. `AntTestContentHandler`
   * produces `##teamcity[...]` text events
3. `OutputToGeneralTestEventsConverter` interprets that via `MyServiceMessageVisitor`
4. `GeneralTestEventsProcessor` handles the display

But then it creates an indirection that I'm not sure is worth it, and the event would still needs parsing. Also, what if the JUnit report is richer than the team city events (I have no idea today), e.g. Gradle recently added the support for attachment, but other data exists today (the properties for example). They are not really surfaced today in the IJ interface, but it might be something to surface one day.

* https://docs.gradle.org/9.4.0/release-notes.html#additional-test-data-capture
* https://github.com/testmoapp/junitxml#properties-for-suites-and-cases

bric3: Additional notes, while looking at team city handling it seems the message needs to be "preparsed" in some way to set proper attributes. e.g. expected/actual.

bric3: Ping @madlexa ;)

madlexa: Hi @bric3 , thanks for your patience!

OK, if that change is hard to do, I withdraw my suggestion - let's keep your approach.

Nice work here: result elements parsed separately from output and properties, support for a single `<testsuite>` root from `Maven/Gradle`, surefire `retry/flaky` elements, and — most importantly — tests with real report fixtures. 👍

One last thing, if you don't mind. The failed event is currently created with `null, nul`l for the comparison texts, so the diff viewer never opens for imported reports. Could you **fill** `expected/actual`- take them from the XML when the writer provides them, and otherwise parse them from the failure message? You don't need to share any code, but there is a similar logic you can look at for reference: [ComparisonFailureData.registerSMAttributes(...)](https://github.com/JetBrains/intellij-community/blob/master/plugins/junit_rt/src/com/intellij/junit4/ExpectedPatterns.java) in java/java-runtime/src/com/intellij/rt/execution/junit/ComparisonFailureData.java, and the patterns in AbstractExpectedPatterns / plugins/junit_rt/src/com/intellij/junit4/ExpectedPatterns.java.

It would also be great to assert in your `FileComparisonFailure` fixture test that the test proxy really gets the comparison texts, not only the message and the stack trace.

madlexa (changes requested):

java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
bric3: 🙇 Thank you to take the time to review my contributions!

That's a fair improvement. I've looked a bit into what's done with junit report accross the board, and I didn't found any explicit elements or attributes wrote in the report for actual / expected, so I would err on avoiding supporting it for no reasons there. However, parsing the message makes sense, I've worked on it today.

However, doing this search was profitable since it allowed me to improve the tested surface. I'll push something next probably.

java/execution/impl/src/com/intellij/execution/AntTestContentHandler.java
```diff
@@ -84,77 +99,281 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+    else if (TESTCASE.equals(name)) {
+      TestFailedEvent failedEvent = createTestFailedEvent();
+      if (failedEvent != null) {
+        myProcessor.onTestFailure(failedEvent);
+      }
+      else {
+        emitNonPrimaryProblems();
+        if (myIgnoredEvent != null) {
+          myProcessor.onTestIgnored(myIgnoredEvent);
         }
```
bric3, line 182: **question:** @madlexa While I was adding more tests, I was wondering about surefire's diagnostic elements (`rerunFailure`, `rerunError`, `flakyFailure`, `flakyError`). Currently, I chose the code to emit the content of those element 
* either on `stderr` when it's flaky (because the testcase ultimately passed, but flaked before)
* or as additional stacktrace element when the rerun failed.

On second thought I'm not sure it's the right approach, I wonder if the parser should emit "synthetic" test events, to surface the retries/flakes ? 

----

FYI the `TestFailedEvent`, and other events for test finished and ignored are not-flexible enough to add this kind of information (retries, flakes, etc.). Also, as far as I know this seems like mostly surefire and gradle at this time. But having a way to surface that in relation to the test makes sense in my opinion (that's why I went with the stderr, and stack trace first).

I'm not sure Gradle write report this way

bric3:
Unsure if I should comment here, as the PR still has writing for reply, but I replied in review comment threads.

@madlexa In particular I'd like some guidance on this https://github.com/JetBrains/intellij-community/pull/3536#discussion_r3790561705

Thanks in advance

PR #3537, opened 2026-06-09
IJPL-246928 Fix opening of files with non-CP-1252 names on Windows

sprigogin:


trespasserw:
I applied a bit more generic variant of the fix. Thanks!

PR #3538, opened 2026-06-10
IDEA-359309 Error reported for Groovy enum constants using map expression

DidierLoiseau:
Do not report error if Groovy >= 2.2 is used (+ some cleanup)

**Note**: While I updated the tests, I never actually managed to run them, so I did it blindly. Within IntelliJ, it says “_The test uses classes from Java plugin but Java plugin wasn't loaded_”, and in the terminal `./tests.cmd --module intellij.groovy --test "*Groovy*"` says “_No tests were found in 'intellij.groovy' module classpath w/ simple patterns 'null', patterns '*Groovy*', or groups 'null'_”.


BasLeijdekkers:
Thanks for your contribution, I have merged it with some minor changes. It should appear on GitHub after some time.

PR #3540, opened 2026-06-12
IDEA-378401 Copy Gradle project path via Copy Reference on build scripts

bric3:
Register a `qualifiedNameProvider` (requires the `refactoring` module) that resolves Gradle identity paths from the `ExternalProject` tree when synced, and falls back to module data. This covers composite builds, buildSrc, custom build file names, and settings scripts without a separate _Copy Path_ menu item.

https://youtrack.jetbrains.com/issue/IDEA-378401/Copy-build.gradle-project-path-reference

With this change, the _Copy Reference_ is now proposed, while it wasn't before.

Screenshot 2026-06-12 at 16 51 16
Screenshot 2026-06-12 at 14 55 42
Screenshot 2026-06-12 at 14 56 14
Screenshot 2026-06-12 at 14 56 56


plugins/gradle/src/org/jetbrains/plugins/gradle/action/GradleBuildScriptQualifiedNameProvider.kt
```diff
@@ -0,0 +1,129 @@
...
+import java.nio.file.Path
+import java.util.ArrayDeque
+import kotlin.io.path.pathString
+
+/**
+ * Feeds **Copy Reference** via [com.intellij.ide.actions.FqnUtil]: Gradle identity path for the project that owns
+ * this script, following Gradle's tooling model ([ExternalProject.getBuildFile], [ExternalProject.getProjectDir],
+ * [ExternalProject.getIdentityPath]).
+ */
+class GradleBuildScriptQualifiedNameProvider : QualifiedNameProvider {
```
nskvortsov, line 28: Could you please elaborate a bit more on why was a (more generic) `QualifiedNameProvider` chosen over (more precise) `CopyPathProvider`? 

The latter fits the task perfectly and allows a more clear description, e.g. "Copy Gradle Path" in the menu.

nskvortsov (changes requested):

plugins/gradle/src/org/jetbrains/plugins/gradle/action/GradleBuildScriptQualifiedNameProvider.kt
bric3: I chose this it seems to me that's what is used `CopyReferenceAction` (i.e. I followed the _Copy Reference_ action), and I landed on QualifiedNameProvider. For me a project path is a "reference". This avoided to create a new action.

But I'm ok as well if a dedicated “Copy Gradle Path” is preferable.

nskvortsov: After giving it another thought, let’s keep the `QualifiedNameProvider`. It binds automatically to various keys and has a small footprint.

nskvortsov (approved):
I will be running the CI checks shortly

bric3:
Any news on the CI?

nskvortsov:
Sorry for the delay.
The CI run surfaced a new issue, related to IDEA's [Split Mode](https://plugins.jetbrains.com/docs/intellij/split-mode-and-remote-development.html) requirements. Basically, the `QualifiedNameProvider` implementations now _must_ be placed in a backend module, which is not yet introduced in our codebase. 
I do not expect any changes from your side. We are working on introducing the required module.

bric3:
@nskvortsov thanks for the feedback. I'll wait for it. Is there something public to specifically track this ongoing change about the qualified name provider?

nskvortsov:
Not really. Will create it soon

nskvortsov:
I ended up applying this work myself. Running through internal checks now.

bric3:
@nskvortsov Cool, thanks a bunch

PR #3541, opened 2026-06-12
Handle ClassNotFoundException When Checking isInstance ScopeCoroutine

alonalbert:
If loading ScopeCoroutine throws ClassNotFoundException, it means callerFrame can't be an instance.

See https://youtrack.jetbrains.com/issue/IDEA-390477


zuevmaxim:
042704e6fe51626acc9185bbed4f473d9ec50d6b

PR #3542, opened 2026-06-15
Add support for relative paths in the compose-based markdown preview editor

tnorbye:


plugins/markdown/compose/src/main/kotlin/com/intellij/markdown/compose/preview/MarkdownComposePanel.kt
```diff
@@ -127,7 +130,20 @@ internal class MarkdownComposePanel(
...
+      val nioPath = baseDir?.toNioPathOrNull()
+
+      val defaultResolver = if (nioPath != null) {
+        ImageSourceResolver.create(nioPath, logResolveFailure = true)
+      } else {
+        ImageSourceResolver.create()
+      }
+
+      MarkdownComposeImageSourceResolver(baseDir, defaultResolver)
+    }
```
DanielSouzaBertoldi, line 141: Our `ImageSourceResolver` `create(rootDir: Path, logResolveFailure: Boolean)` overload creates a `DefaultImageSourceResolver` with all capabilities that we provide(`PlainUri`, `AbsolutePath`, `RelativePathInResources` and `RelativePath`).

So I feel like this code could just be reduced to this:

```kt
val imageSourceResolver = remember(virtualFile) {
      val nioPath = virtualFile?.parent?.toNioPathOrNull()
      if (nioPath != null) ImageSourceResolver.create(nioPath, logResolveFailure = true)
      else ImageSourceResolver.create()
  }
```

One thing that might be worth it to apply in our `RelativePath` is the code that handles malformed `file:` strings, but I'd argue that malformed URIs should be treated as malformed instead of worked around for the sake of providing a "less resistance" path, but I'd be happy to discuss this 😄 

Or is there something I'm missing here that makes our `DefaultImageSourceResolver` not enough for your use case?

plugins/markdown/compose/src/main/kotlin/com/intellij/markdown/compose/preview/MarkdownComposeImageSourceResolver.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.markdown.rendering.ImageSourceResolver
+
+internal class MarkdownComposeImageSourceResolver(private val baseDir: VirtualFile?, private val defaultResolver: ImageSourceResolver) :
+  ImageSourceResolver {
+
+  override fun resolve(rawDestination: String): String? {
+    // If the image source looks like a file URL and we have a base directory,
+    // attempt to resolve it locally first.
+    var clean = rawDestination
```
nebojsa-vuksic: nit: Could we parse standard file URLs via `java.net.URI` / `Path` first instead of stripping the prefix manually?

That would also avoid a couple of issues here: 
1. absolute file:// resolution is currently gated on `baseDir != null`, even though absolute paths do not need a base dir
2. the tests currently cover helpers/fallbacks but not the core parsing branches (`file:///...`, `file://...`, relative `file://image.png`, encoded paths, etc.). A few targeted tests around those cases would make this much safer.

plugins/markdown/compose/intellij.markdown.compose.preview.iml
```diff
@@ -28,6 +28,7 @@
     <content url="file://$MODULE_DIR$">
       <sourceFolder url="file://$MODULE_DIR$/src/main/kotlin" isTestSource="false" />
       <sourceFolder url="file://$MODULE_DIR$/src/main/resources" type="java-resource" />
+      <sourceFolder url="file://$MODULE_DIR$/test/src" isTestSource="true" />
```
nebojsa-vuksic: If I remember correctly this will fail IJ project structure tests, because now its required to have a separate JPS module for tests.

plugins/markdown/compose/src/main/kotlin/com/intellij/markdown/compose/preview/MarkdownComposePanel.kt
tnorbye: Yeah, I had done it this way because I was trying to support file://relative links, but apparently that is not common, so definitely not worth the extra processing here, so it's much simpler now. Which means I could drop the tests too.

plugins/markdown/compose/src/main/kotlin/com/intellij/markdown/compose/preview/MarkdownComposeImageSourceResolver.kt
tnorbye: Yeah I did that, before looking more carefully at the DefaultImageSourceResolver and realizing I could use it if I just drop the file://relative handling, which doesn't seem worth it, so this is gone now.

plugins/markdown/compose/intellij.markdown.compose.preview.iml
tnorbye: Yep -- I've deleted this now since we're just reusing the DefaultImageSourceResolver.

nebojsa-vuksic (approved):

DanielSouzaBertoldi (commented):
leaving my humble approval here, thanks for discussing this with us :)

DanielSouzaBertoldi (approved):

PR #3544, opened 2026-06-18
[JEWEL-1339] Decouple .main.kts CI scripts from runner Kotlin by pinning Kotlin 2.3.20

nebojsa-vuksic:
The Jewel Checks workflow ran its `.main.kts` scripts through the `/usr/bin/env kotlin` shebang, so they used whatever Kotlin the GitHub runner image happened to ship. A runner image refresh moved that to Kotlin 2.4.0, which has a `.main.kts` regression resolving top-level helpers/extensions imported across a `@file:Import` boundary (FIR crash `Expected FirResolvedTypeRef with ConeKotlinType but was FirUserTypeRefImpl`), breaking the API-dump and Metalava jobs. The runner image drift was the trigger; the underlying bug is in Kotlin 2.4.0's `.main.kts` import handling. This change makes the Jewel CI scripts independent of the runner-provided Kotlin and deterministic across runner image refreshes.

## Changes

* Add a local composite action `.github/actions/setup-kotlin` that installs a pinned Kotlin compiler (`version` input), caches it per OS and version via `actions/cache`, self-heals if the cached install is incomplete, and exposes the `bin` directory as an output.
* Use it in every `jewel-checks.yml` job that runs a `.main.kts` script — annotate API dumps, check API dumps, and Metalava — invoking the pinned `kotlin` binary directly instead of relying on the shebang/PATH.
* Pin the version via a workflow-level `KOTLIN_VERSION: 2.3.20` env, matching Jewel's build Kotlin.

### DIFF
--- a/.github/actions/setup-kotlin/action.yml
+++ b/.github/actions/setup-kotlin/action.yml
@@ -1,0 +1,54 @@
+name: Set up Kotlin
+description: >
+  Installs a pinned Kotlin compiler and exposes its bin directory. The download is cached per OS and version, so it is
+  only fetched on a cache miss (first run on a runner, or after a version bump) rather than on every run.
+
+inputs:
+  version:
+    description: The Kotlin version to install (e.g., 2.3.20).
+    required: true
+
+outputs:
+  bin-dir:
+    description: Absolute path to the installed Kotlin bin directory.
+    value: ${{ steps.expose.outputs.bin-dir }}
+
+runs:
+  using: composite
+  steps:
+    - name: Restore cached Kotlin ${{ inputs.version }}
+      id: cache
+      uses: actions/cache@v4
+      with:
+        key: kotlin-${{ runner.os }}-${{ inputs.version }}
+        path: ${{ runner.temp }}/kotlin-${{ inputs.version }}
+
+    - name: Install Kotlin ${{ inputs.version }}
+      shell: bash
+      run: |
+        KOTLIN_HOME="${RUNNER_TEMP}/kotlin-${{ inputs.version }}"
+        KOTLINC="${KOTLIN_HOME}/bin/kotlinc"
+
+        # Trust the restored cache only if the compiler is present and reports the expected version.
+        kotlin_cached_install_is_valid() {
+          [ -x "${KOTLINC}" ] || return 1
+          "${KOTLINC}" -version 2>&1 | grep -qF "kotlinc-jvm ${{ inputs.version }} "
+        }
+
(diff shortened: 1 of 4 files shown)


.github/actions/setup-kotlin/action.yml
```diff
@@ -0,0 +1,46 @@
+name: Set up Kotlin
+description: >
+  Installs a pinned Kotlin compiler and exposes its bin directory. The download is cached per OS and version, so it is
+  only fetched on a cache miss (first run on a runner, or after a version bump) rather than on every run.
+
+inputs:
+  version:
+    description: The Kotlin version to install (e.g. 2.3.20).
```
rock3r: The nit-tiest of nits... 
```suggestion
    description: The Kotlin version to install (e.g., 2.3.20).
```

.github/actions/setup-kotlin/action.yml
```diff
@@ -0,0 +1,46 @@
...
+      with:
+        key: kotlin-${{ runner.os }}-${{ inputs.version }}
+        path: ${{ runner.temp }}/kotlin-${{ inputs.version }}
+
+    - name: Install Kotlin ${{ inputs.version }}
+      shell: bash
+      run: |
+        KOTLIN_HOME="${RUNNER_TEMP}/kotlin-${{ inputs.version }}"
+        # Self-healing: trust the restored cache only if the launcher is actually present, otherwise (re)download.
+        if [ -x "${KOTLIN_HOME}/bin/kotlin" ]; then
```
rock3r: Probably unnecessary, but what about validating the actual version as a further guard?

rock3r (approved):

DanielSouzaBertoldi (approved):
LGTM! Quick question though, do we need the Release Notes section? I thought that the github actions code was only used for internal development, i.e, we don't expose these things to users :)

rock3r:
Yeah we can drop it. Seems like the skill was a bit over-eager :D

rock3r:
One thing I just thought of is that we should document this pinning and have a kotlin version bump checklist so we do not forget to do this in the future

.github/actions/setup-kotlin/action.yml
nebojsa-vuksic: Makes sense. Let me update the check 👍

nebojsa-vuksic:
> One thing I just thought of is that we should document this pinning and have a kotlin version bump checklist so we do not forget to do this in the future

@rock3r I've added consistency test too. It's a better guard. Now `jewel/gradle/libs.versions.toml` is source of truth.

nebojsa-vuksic:
Ready to merge

PR #3547, opened 2026-06-24
[JEWEL-1345] Improve Markdown block renderer API

rock3r:
This PR allows custom Markdown block renderers to better support HTML parsing, text alignment, and image rendering without removing any existing binary API. Some renderer paths that were previously private implementation details are now exposed as overridable hooks, while the deprecated lowercase renderer shims and legacy `HtmlBlock` API remain in place for compatibility.

## Changes

* Added overridable uppercase renderer hooks for attributed HTML blocks and image-only paragraph rendering.
* Made the default renderer's rendered-image lookup extensible for custom block renderers.
* Kept raw CommonMark HTML blocks intentionally non-rendered by default while preserving the existing `MarkdownBlock.HtmlBlock` API.
* Preserved one-off logging for unsupported custom blocks, scoped per renderer instance.
* Updated renderer tests, Metalava signatures, and IntelliJ Platform ABI dumps for the additive API changes.

## Testing

* `cd platform/jewel && ./gradlew :markdown:core:compileKotlin :markdown:core:compileTestKotlin :markdown:extensions:gfm-tables:compileTestKotlin --no-daemon`
* `cd platform/jewel && ./gradlew :markdown:core:checkMetalavaExperimentalApi --no-daemon`

### New features

* Custom Markdown block renderers can override attributed HTML-block rendering directly via `RenderHtmlBlockWithAttributes`.
* Custom Markdown block renderers can override image-only paragraph rendering via `RenderImagesOnlyParagraph`.

### DIFF
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
@@ -60,7 +60,8 @@ import org.jetbrains.jewel.markdown.scrolling.ScrollingSynchronizer
  *   provided by the [MarkdownParserFactory], but you can provide your own if you need to customize the parser — e.g.,
  *   to ignore certain tags. If [markdownMode] is `MarkdownMode.WithEditor`, make sure you set
  *   `includeSourceSpans(IncludeSourceSpans.BLOCKS)` on the parser.
- * @param parseEmbeddedHtml If `true`, a subset of native HTML elements will be parsed as Markdown blocks.
+ * @param parseEmbeddedHtml If `true`, a subset of native HTML elements will be parsed as Markdown blocks. See
+ *   [MarkdownHtmlConverter].
  */
 @ApiStatus.Experimental
 @ExperimentalJewelApi
@@ -71,7 +72,7 @@ public class MarkdownProcessor(
         MarkdownParserFactory.create(optimizeEdits = markdownMode is MarkdownMode.EditorPreview, extensions),
     private val parseEmbeddedHtml: Boolean = false,
 ) {
-    @Suppress("UnusedPrivateProperty") // languageRecognizer is only here for binary compat reasons
+    @Suppress("UnusedPrivateProperty", "DEPRECATION") // languageRecognizer is only here for binary compat reasons
     @Deprecated(
         "`languageRecognizer` is not necessary anymore. Use the constructor without it.",
         replaceWith = ReplaceWith("MarkdownProcessor(extensions, markdownMode, commonMarkParser)"),
@@ -85,6 +86,7 @@ public class MarkdownProcessor(
         parseEmbeddedHtml: Boolean = false,
     ) : this(extensions, markdownMode, commonMarkParser, parseEmbeddedHtml)
 
+    @Suppress("DEPRECATION")
     @Deprecated("Use a version with a `parseEmbeddedHtml` parameter", level = DeprecationLevel.HIDDEN)
     public constructor(
         extensions: List<MarkdownProcessorExtension> = emptyList(),
@@ -94,6 +96,7 @@ public class MarkdownProcessor(
         languageRecognizer: (String) -> MimeType? = { MimeType.Known.fromMarkdownLanguageName(it) },
     ) : this(extensions, markdownMode, commonMarkParser, languageRecognizer, false)
 
+    @Suppress("DEPRECATION")
     @Deprecated("Use a version with a `parseEmbeddedHtml` parameter", level = DeprecationLevel.HIDDEN)
     public constructor(
         extensions: List<MarkdownProcessorExtension> = emptyList(),
@@ -351,7 +354,7 @@ public class MarkdownProcessor(
             intermediateHtmlBlocks.mapNotNull { htmlElement ->
                 htmlConverter.convert(this@MarkdownProcessor, htmlElement) { newMdBlock, lines ->
                     when (newMdBlock) {
-                        is MarkdownBlock.HtmlBlock -> return@convert newMdBlock
+                        is MarkdownBlock.HtmlBlockWithAttributes -> return@convert newMdBlock
                         is MarkdownBlock.ListItem ->
                             return@convert newMdBlock // we convert the list item's inner block instead
                         else -> {
@@ -370,7 +373,7 @@ public class MarkdownProcessor(
         MarkdownBlock.BlockQuote(processChildren(this))
 
     private fun Heading.toMarkdownHeadingOrNull(): MarkdownBlock.Heading? {
-        if (level < 1 || level > 6) return null
+        if (level !in 1..6) return null
         return MarkdownBlock.Heading(inlineContent = readInlineMarkdown(this@MarkdownProcessor), level = level)
     }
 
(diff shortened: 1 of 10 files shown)


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/MarkdownBlockRenderer.kt
```diff
@@ -614,36 +283,37 @@ public interface MarkdownBlockRenderer {
...
-     * Renders a [HtmlBlock] into a Compose UI. Since Compose can't render HTML out of the box, this might result in a
-     * no-op (e.g., in [DefaultMarkdownBlockRenderer.render]).
+     * Renders a [MarkdownBlock.HtmlBlockWithAttributes] into a Compose UI. When
+     * [org.jetbrains.jewel.markdown.processing.MarkdownProcessor.parseEmbeddedHtml] is `false`, raw HTML blocks are
+     * exposed as [MarkdownBlock.HtmlBlockWithAttributes] wrapping a paragraph with raw HTML inline content. When it is
+     * `true`, supported HTML is converted into Markdown blocks and wrapped with attributes when applicable.
      *
      * @param block The HTML block to render.
-     * @param styling The [`HtmlBlock`][MarkdownStyling.HtmlBlock] styling to use to render.
+     * @param onUrlClick
```
DanielSouzaBertoldi: Missing param's description

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -119,58 +112,45 @@ public open class DefaultMarkdownBlockRenderer(
...
+                    blockRenderer.RenderCustomBlock(
                         block = block,
                         blockRenderer = this,
                         inlineRenderer = inlineRenderer,
                         enabled = enabled,
                         modifier = modifier,
                         onUrlClick = onUrlClick,
                     )
+                } else {
+                    logUnsupportedBlock(block)
```
DanielSouzaBertoldi, line 173: Given that this is called inside the `RenderBlock` function, then the log is on a per composition basis, not per-screen. Maybe we could make it per-screen?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
```diff
@@ -121,24 +122,6 @@ public sealed interface MarkdownBlock {
         override fun toString(): String = "Heading(inlineContent=$inlineContent, level=$level)"
     }
 
-    @ApiStatus.Experimental
-    @ExperimentalJewelApi
-    @GenerateDataFunctions
-    public class HtmlBlock(public val content: String) : MarkdownBlock {
```
DanielSouzaBertoldi, line 139: Now this class is basically orphaned. We still reference it in `MarkdownStyling` and `MarkdownFactory` for instance, though. I'd figure we could get rid of it for good or file a follow-up issue in YT since there are some classes outside Jewel that also reference it `com.intellij.markdown.compose.preview.JcefLikeMarkdownStyling` being one example

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
```diff
@@ -380,6 +383,14 @@ public class MarkdownProcessor(
     private fun IndentedCodeBlock.toMarkdownCodeBlockOrNull(): CodeBlock.IndentedCodeBlock =
         CodeBlock.IndentedCodeBlock(literal.trimEnd('\n'))
 
+    private fun HtmlBlock.toMarkdownHtmlBlockWithAttributesOrNull(): MarkdownBlock.HtmlBlockWithAttributes? {
+        if (literal.isBlank()) return null
+        return MarkdownBlock.HtmlBlockWithAttributes(
+            mdBlock = MarkdownBlock.Paragraph(listOf(InlineMarkdown.HtmlInline(literal.trimEnd('\n')))),
+            attributes = emptyMap(),
+        )
+    }
```
DanielSouzaBertoldi, line 389: Hmmmmm this is a behavioral breaking change, right? Like, the old code just mapped this to a `HtmlBlock` and when it reached the renderer, it just didn't render anything (`// HTML blocks are intentionally not rendered`).

Now we do render any HTML block, even if the flag is disabled.

And by "render" in the case where the flag is disabled, we just show the raw html code since it's wrapped in a `MarkdownBlock.Paragraph`.

If this is intentional, a renderer test + mentioning this behavioral breaking change would be awesome :)

DanielSouzaBertoldi (commented):
Great stuff, love to see us getting rid of old code 😈 but there are some things I feel we should discuss a bit

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/MarkdownBlock.kt
rock3r: Yeah it's breaking changes fest so may as well. I could also split the removal from the rest, and do a separate PR for those. Probably a better idea...

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
rock3r: Not sure how you'd do "per-screen" without knowing what constitutes a screen in the user code to begin with. I think this is fine to be once per renderer. Very hard to do otherwise in a meaningful way.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
rock3r: Oh yeah I did some changes right before pushing and this slipped through! Good catch. I'll revert the behaviour change.

DanielSouzaBertoldi (approved):
Good stuff 😎

rock3r:
Took out the deprecated API removal to a separate PR, #3560, so this can be merged in 262 for 0.40 (the deprecated API removal will only land on master for 263+ so we do not do breaking changes outside of major IJPL bumps)

DanielSouzaBertoldi (approved):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/processing/MarkdownProcessor.kt
rock3r: Reverted for real™ now :D

rock3r:
Updated the Metalava API dumps to target the 0.41.0 Jewel release instead of 0.40.0. The PR branch was rebased onto the latest master and the API changes were moved from `core-api-0.40.0.txt` to `core-api-0.41.0.txt`.

DanielSouzaBertoldi (approved):

rock3r:
Ready to merge

PR #3548, opened 2026-06-26
[terminal] IJPL-165734 terminal was restarted after drag in editor splits

bric3:
In my last comment in this issue, I noticed that when dragging a reset terminal tab back to the tool window, the terminal "session" was revived.

https://youtrack.jetbrains.com/issue/IJPL-165734/Terminal-restarts-when-dragged-after-Move-to-Editor

This made me explore the the way terminal and editors were working. In my manual testing I also stumbled on an issue with moving an terminal _editor tab_ to another window, I took the opportunity to fix it. I had some trouble with editor splits. So unsure if my fix is ok.

Finally one thing I didn;t fix, but can take a look at it, is that when dragging a terminal tab to the editor tab, it doesn't trigger the "split" ability of editors. It's only when a terminal editor tab is already in the editors that it can be dragged to split.
(However if editors are already split, then dragging to one of the existing split works, but you can't create new ones)

Note: While testing this I noticed that _Split Right / Down_ needed special treatment, so I took inspiration from `FileEditorManagerImpl.moveNoSplitFileIfNeeded` and `MoveEditorToOppositeTabGroupAction`.

https://github.com/user-attachments/assets/a78e4cf8-374c-4b92-a5c8-a4962c830e17

### DIFF
--- a/platform/platform-api/src/com/intellij/ui/tabs/impl/JBTabsImpl.kt
+++ b/platform/platform-api/src/com/intellij/ui/tabs/impl/JBTabsImpl.kt
@@ -2638,6 +2638,7 @@ open class JBTabsImpl internal constructor(
       return
     }
 
+    processRemove(tab = info, forcedNow = true, setSelectedToNull = selectedInfo == info)
     if (visibleInfos.isEmpty()) {
       removeDeferredNow()
     }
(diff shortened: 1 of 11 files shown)


KonstantinHudyakov:
Rejected in favour of more localised fix, please see the details in the issue:
https://youtrack.jetbrains.com/issue/IJPL-165734/Terminal-restarts-when-dragged-after-Move-to-Editor#focus=Comments-27-13963209.0-0

bric3:
Nice, thank you very much @KonstantinHudyakov for taking a look at my PR :)

bric3:
Just a cross reference for me :)

* 469506a4f0a2d5ea488ab545ea4e3b5f3e7ef4d4
* 8c4b53e91ef68e71f8dcba262b68d1fc812f1f2b
* fe43a62fef647767bd172f9e8e5c23cd92adf9b5

PR #3552, opened 2026-06-29
[JEWEL-1327] Fix Markdown Ordered List Crash

DanielSouzaBertoldi:
# Context

We have a crash that occurs whenever users try to start an ordered list with 0. [CommonMark does list this scenario as valid](https://spec.commonmark.org/0.31.2/#list-items) and funnily enough our processor does test for these scenarios exactly like in the specifications but in our renderer we require such values to be `>0`.

This PR fixes this scenario, allowing all format number styles to accept `0` as a value, but for roman/alphabetical formats we fallback to `0`, like github does:

```
0. a
1. b
    0. c
    1. d
        0. e
        1. f
```

Is rendered as:

0. a
1. b
    0. c
    1. d
        0. e
        1. f

# Changes

- Removed the check for `number > 0` in all number format style.
- Added a new test class just for rendering ordered lists `OrderedListRenderingTest`.
- Updated `MarkdownTestTheme` to use roman format as secondary level and alphabetical as the third to mirror our default prod behaviour (and to make it easier to test stuff, ngl)

# Screen Recordings
| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/d0c1025f-b1e0-49dd-bf85-5d2f1d059ee3" /> | <video src="https://github.com/user-attachments/assets/af1a57ce-0bea-4ae3-94e6-4e6dfbabd6c6" /> |

**Note:** Yes, this is [the default behavior for CommonMark](https://github.com/commonmark/commonmark-java/blob/main/commonmark/src/main/java/org/commonmark/internal/ListBlockParser.java#L94). It only interrupts a paragraph if it starts with `1`, so that's why it renders both zeros in the secondary and third levels in the same line as the previous list item.

### Bug fixes
 * Markdown ordered lists won't crash if you start your list with 0. For number formats such as roman and alphabetical, an item with index 0 will render as 0 and then render as the expected number format (i.e, `i` for roman, `a` for alphabetical)

### DIFF
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/MarkdownStyling.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/MarkdownStyling.kt
@@ -471,7 +471,7 @@ public class MarkdownStyling(
 
                     public object Decimal : NumberFormatStyle {
                         override fun formatNumber(number: Int): String {
-                            require(number > 0) { "Input must be a positive integer" }
+                            require(number >= 0) { "Input must not be a negative integer" }
 
                             return number.toString()
                         }
@@ -479,7 +479,8 @@ public class MarkdownStyling(
 
                     public object Roman : NumberFormatStyle {
                         override fun formatNumber(number: Int): String {
-                            require(number > 0) { "Input must be a positive integer" }
+                            // Roman numerals can't represent 0; just render it as the literal "0".
+                            if (number < 1) return number.toString()
 
                             val values = intArrayOf(1000, 900, 500, 400, 100, 90, 50, 40, 10, 9, 5, 4, 1)
                             val symbols = arrayOf("m", "cm", "d", "cd", "c", "xc", "l", "xl", "x", "ix", "v", "iv", "i")
@@ -498,7 +499,8 @@ public class MarkdownStyling(
 
                     public object Alphabetical : NumberFormatStyle {
                         override fun formatNumber(number: Int): String {
-                            require(number > 0) { "Input must be a positive integer" }
+                            // Letters can't represent 0; just render it as the literal "0".
+                            if (number < 1) return number.toString()
 
                             var num = number
                             return buildString {
--- a/platform/jewel/markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/rendering/OrderedListRenderingTest.kt
+++ b/platform/jewel/markdown/core/src/test/kotlin/org/jetbrains/jewel/markdown/rendering/OrderedListRenderingTest.kt
@@ -1,0 +1,113 @@
+// Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the
+// Apache 2.0 license.
+package org.jetbrains.jewel.markdown.rendering
+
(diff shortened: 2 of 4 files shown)


nebojsa-vuksic (approved):

rock3r:
@DanielSouzaBertoldi can you add release notes please

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/MarkdownStyling.kt
```diff
@@ -471,15 +471,16 @@ public class MarkdownStyling(
 
                     public object Decimal : NumberFormatStyle {
                         override fun formatNumber(number: Int): String {
-                            require(number > 0) { "Input must be a positive integer" }
+                            require(number >= 0) { "Input must not be negative value" }
```
rock3r:
```suggestion
                            require(number >= 0) { "Input must not be a negative integer" }
```

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/MarkdownStyling.kt
```diff
@@ -498,7 +499,8 @@ public class MarkdownStyling(
 
                     public object Alphabetical : NumberFormatStyle {
                         override fun formatNumber(number: Int): String {
-                            require(number > 0) { "Input must be a positive integer" }
+                            // Roman numerals can't represent 0; Just render it as the literal "0".
```
rock3r:
```suggestion
                            // Letters can't represent 0; just render it as the literal "0".
```

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/MarkdownStyling.kt
```diff
@@ -471,15 +471,16 @@ public class MarkdownStyling(
...
+                            require(number >= 0) { "Input must not be negative value" }
 
                             return number.toString()
                         }
                     }
 
                     public object Roman : NumberFormatStyle {
                         override fun formatNumber(number: Int): String {
-                            require(number > 0) { "Input must be a positive integer" }
+                            // Roman numerals can't represent 0; Just render it as the literal "0".
```
rock3r:
```suggestion
                            // Roman numerals can't represent 0; just render it as the literal "0".
```

rock3r (commented):
LGTM, apart from some nits and missing the release notes. Can you confirm it renders the same as [this gist](https://gist.github.com/rock3r/1b2c98ef3af4a77776fcb45809ef6f4c)?

DanielSouzaBertoldi:
| Jewel | GitHub |
|---|---|
| Screenshot 2026-06-30 at 14 34 11 | gist |

@rock3r apart from the spacing between items, the number format and indentation are spot on :)

I'll file an issue to look at the spacing between each item in the list though

rock3r (approved):
pending the small nits :)

DanielSouzaBertoldi:
@rock3r alright, just pushed the changes + added release notes (completely forgot about it, damn)

DanielSouzaBertoldi:
Ready to merge

PR #3554, opened 2026-06-30
[KTIJ-24906]: Add all imports for ReplaceWith

kshulzh:
https://youtrack.jetbrains.com/issue/KTIJ-24906/ReplaceWith-doesnt-add-several-imports

https://github.com/user-attachments/assets/e2e949f4-2777-4e33-8163-6deb7e08060c

Hi all, I have a question regarding second part of provided video should I try to avoid import unused declarations?

### DIFF
--- a/plugins/kotlin/code-insight/fixes-k2/src/org/jetbrains/kotlin/idea/k2/codeinsight/fixes/replaceWith/DeprecatedSymbolUsageFixBase.kt
+++ b/plugins/kotlin/code-insight/fixes-k2/src/org/jetbrains/kotlin/idea/k2/codeinsight/fixes/replaceWith/DeprecatedSymbolUsageFixBase.kt
@@ -241,7 +241,8 @@ abstract class DeprecatedSymbolUsageFixBase(
 
                     val typeElement = typeReference?.typeElement as? KtUserType ?: return null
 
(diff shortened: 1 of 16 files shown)


akozlova (changes requested):
I don't think it's possible to avoid unused imports as part of the fix; they should be cleaned on optimize imports step after refactorings are done. But all imports are actually used when we create a class reference, can't we include only those additional imports which are not used in the class reference?
Because this addImport call performs direct psi modification, meaning that in the worst case it would add import which is already covered by on demand import (*). Let's try to avoid the additional pressure when no additional imports were added (the common case, I guess)
Thanks!

kshulzh:
hello @akozlova, I updated some files. Used addImportFor to add sammple imports and addImport for star imports

akozlova:
It's better but still you process imports which are already part of the replacement signature for the second time. Can't we filter out those which were already part of the fully qualified name of the class? I am not actually sure that we need to support the case from the issue at all (I hope we would have a spec what can and what can not be written in the annotation) but I want that the normal cases do not have performance problems because of that

kshulzh:
Hello, I added filter for additional imports and some test cases for conflicting imports

akozlova:
The MR was merged. Thank you for your contribution!

PR #3556, opened 2026-07-02
[JEWEL-1277] Add FrontMatter metadata block support for Markdown

DanielSouzaBertoldi:
> [!IMPORTANT]
> This PR supersedes #3454 ! It does all the changes Seb suggested :) 

## Summary

Add support for YAML front matter metadata blocks in Jewel's Markdown rendering. Front matter content (delimited by `---` markers at the beginning of a file) is parsed and rendered as a table where keys form the first column and values form the second column. It differs from other renderers that prefer using rows (likely because then the keys will render in bold), but it helps keeping all the information in a Markdown file readable without coming down to horizontal scrolling.

Multi-value (list) entries are rendered as Markdown lists (opposite to one-row tables in other renderers), again, to keep the vertical structure of the block.

## Changes

* New `front-matter` extension module (`intellij.platform.jewel.markdown.extensions.frontMatter`) with a `FrontMatterProcessorExtension` that integrates with Jewel's Markdown processor pipeline.
* `FrontMatter` custom block is self-contained, its own `FrontMatterBlockRenderer` uses `BasicTableLayout` in `:foundation`.
* Added `FrontMatterStyling` + int-ui-standalone/ide-laf-bridge factories.
* Custom block parser (`FrontMatterBlockParser`) supporting plain key-value pairs, YAML lists, quoted strings, and block scalars (literal `|` and folded `>` with all chomping indicators).
* Parser now requires a closing `---`. An unterminated block ignores the entire markdown in it.
* Added tons of unit tests covering various front matter scenarios including block scalars, lists, quoted values, and edge cases.
* `TableCell`, `TableHeader`, `TableRow` and `TableBlock` are now experimental (as they should've been from the start)

### Tested scenarios

| Use case | Screenshot |
|---|---|
| Simple key-value block |  Screenshot 2026-07-02 at 15 48 17 |
| Single key block | Screenshot 2026-07-02 at 15 58 56 |
| Empty block | Screenshot 2026-07-02 at 15 59 26 |
| List | Screenshot 2026-07-02 at 15 59 46 |
| Single item | Screenshot 2026-07-02 at 16 00 01 |
| Mixed scalar and list values | Screenshot 2026-07-02 at 16 00 17 |
| Block parsing doesn't render if no closing `---` is found | Screenshot 2026-07-02 at 16 00 43 |
| Block not at document start | Screenshot 2026-07-02 at 16 01 36 |
| List items with colon | Screenshot 2026-07-02 at 16 01 49 |
| Scalar with newlines | Screenshot 2026-07-02 at 16 02 30 |
| Scalar with leading lines | Screenshot 2026-07-02 at 16 02 56 |
| Strip chomping | Screenshot 2026-07-02 at 16 03 14 |
| Keep chomping | Screenshot 2026-07-02 at 16 03 25 |
| Folding | Screenshot 2026-07-02 at 16 03 51 |
| Folding + strip chomping | Screenshot 2026-07-02 at 16 04 11 |
| Folding + keep chomping | Screenshot 2026-07-02 at 16 04 51 |
| Folding + blank lines | Screenshot 2026-07-02 at 16 05 12 |
| Empty scalar | Screenshot 2026-07-02 at 16 05 25 |
| Empty scalar + key | Screenshot 2026-07-02 at 16 05 40 |
| Non-empty scalar + key | Screenshot 2026-07-02 at 16 05 56 |
| Unquote quotes | Screenshot 2026-07-02 at 16 06 14 |
| block with full-line comment | frontmatter-01 |
| block with "comment" in the value | frontmatter-02 |
| Flow-type sequence | frontmatter-03 |
| Flow-type sequence empty | frontmatter-04 |
| Flow-type sequence items with quotes | frontmatter-05 |
| Flow-type sequence only one item | frontmatter-06 |
| Flow-type sequence wrapped in double quotes | frontmatter-07 |

> [!NOTE]
> The only major difference in my implementation compared to Sasha's is that now that we are not using Gfm Table to render the table, we do not have alternating background colors. Here's what I mean:

| Before | After |
|---|---|
| image | Screenshot 2026-07-02 at 16 00 17 |

Is this something we wish to replicate? Maybe. Please let me know your thoughts, GitHub does seem to alternate the background color, [as can be seen here](https://github.com/JetBrains/intellij-community/blob/master/.claude/skills/jewel-release-helper/SKILL.md).

### Typing in

https://github.com/user-attachments/assets/5133a7ec-ea06-464d-bc12-239e968e01b3

### ⚠️ Important Changes
 * `TableCell`, `TableHeader`, `TableRow` and `TableBlock` are now experimental.

### New features
* Added YAML front matter metadata block support for Markdown documents — front matter is rendered as a key-value table. Nested tables are not supported as of yet.


platform/jewel/markdown/extensions/gfm-tables/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/github/tables/TableBlock.kt
```diff
@@ -1,8 +1,12 @@
 // Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
 package org.jetbrains.jewel.markdown.extensions.github.tables
 
+import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
 import org.jetbrains.jewel.markdown.MarkdownBlock
 
+@ApiStatus.Experimental
+@ExperimentalJewelApi
 internal data class TableBlock(val header: TableHeader, val rows: List<TableRow>) : MarkdownBlock.CustomBlock {
```
DanielSouzaBertoldi, line 10: These APIs should've been experimental from the start. Does this deserve its own release notes? Maybe important changes?

platform/jewel/markdown/extensions/front-matter/intellij.platform.jewel.markdown.extensions.frontMatter.iml
```diff
@@ -0,0 +1,46 @@
...
+      <excludeFolder url="file://$MODULE_DIR$/bin" />
+      <excludeFolder url="file://$MODULE_DIR$/build" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="library" name="kotlin-stdlib" level="project" />
+    <orderEntry type="module" module-name="intellij.libraries.kotlinx.coroutines.core" />
+    <orderEntry type="library" name="jetbrains-annotations" level="project" />
+    <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="module" module-name="intellij.platform.jewel.markdown.core" />
+    <orderEntry type="module" module-name="intellij.platform.jewel.markdown.extensions.gfmTables" />
```
rock3r: Looks like this dependency is unused now that front matter no longer reuses the GFM table model/renderer. The Gradle module and plugin descriptor don't depend on `gfmTables` anymore, but the JPS module still does, and that also generated the `BUILD.bazel` dependency.

Can we drop this from the `.iml` and regenerate the Bazel file? Otherwise the module graph still keeps part of the coupling we were trying to avoid.

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
```diff
@@ -0,0 +1,90 @@
...
+    private object FrontMatterBlockProcessorExtension : MarkdownBlockProcessorExtension {
+        override fun canProcess(block: CustomBlock): Boolean = block is FrontMatterBlock
+
+        override fun processMarkdownBlock(
+            block: CustomBlock,
+            processor: MarkdownProcessor,
+        ): MarkdownBlock.CustomBlock? {
+            val frontMatterBlock = block as FrontMatterBlock
+            // Only treat the collected content as front matter if it was closed by a trailing "---" delimiter.
+            if (!frontMatterBlock.sawClosingDelimiter) return null
```
rock3r, line 47: I think this only half-addresses the unclosed-delimiter case. We no longer emit a `FrontMatter` block, but by the time we return `null` here the custom parser has already consumed the opening `---` and any key-like lines, so they disappear from the rendered Markdown instead of being treated as normal Markdown.

For example, with:

```markdown
---
title: This was meant to be normal markdown
# Heading
```

GitHub renders the first line as a thematic break, then `title: ...` as a paragraph, then the heading. With the current implementation the consumed front-matter-looking lines are dropped. Can we either only claim the block once we know there is a closing delimiter, or preserve/re-emit the consumed lines through the normal Markdown path?

Reference rendering:

Image

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
```diff
@@ -0,0 +1,279 @@
...
+        val listItemMatch = LIST_ITEM_REGEX.matchEntire(trimmed)
+        // If a list item, add it to the current block
+        if (listItemMatch != null) {
+            if (block.currentKey == null) return false
+            block.currentIsList = true
+            block.currentValues.add(parseStringValue(listItemMatch.groupValues[1]))
+            return true
+        }
+
+        val keyValue = trimmed.split(':', limit = 2)
```
rock3r, line 107: Looks like comment-only YAML lines break an otherwise valid front matter block. A line like `# generated by tooling` has no `:`, so `parseKeyLine()` returns `false`, the parser stops before seeing the closing delimiter, and then the already-consumed content can be dropped by the unclosed-block path.

Can we ignore YAML comment lines inside front matter? They are common enough in metadata blocks that this seems worth supporting, or at least testing explicitly. Can be split off to a follow-up if it's too much pain to do here.

Reference render:
Image

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
```diff
@@ -0,0 +1,279 @@
...
+        if (keyValue.size < 2) {
+            // not a "key: value" pair
+            return false
+        }
+
+        val (key, value) = keyValue.map { it.trim() }
```
rock3r, line 113: I reckon nested mappings need a clearer behaviour here. Since we trim the key, indented mappings are promoted to top-level entries:

```yaml
metadata:
  author: example-org
  version: "1.0"
```

This would render as an empty `metadata` row plus separate top-level `author` and `version` rows, which is misleading for common front matter shapes like skill metadata. Ideally, we should support nested maps properly; if that is too much for this PR, could we at least avoid flattening them silently and split proper support to a follow-up?

Reference render:
Image

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
```diff
@@ -0,0 +1,279 @@
...
+                    currentKey = key,
+                    indicator = blockScalarMatch.groupValues[1][0],
+                    chomping = blockScalarMatch.groupValues[2].firstOrNull(),
+                    minimumIndent = line.indentWidth() + 1,
+                )
+            return true
+        }
+
+        // Simple "key: value"
+        currentBlock = CurrentBlock.KeyValue(currentKey = key, currentValues = mutableListOf(parseStringValue(value)))
```
rock3r, line 147: Could we also account for flow-style YAML sequences or at least document that they are not supported yet?

`tags: [a, b]` is a fairly common front matter shape, but this path treats it as a scalar and renders the literal `[a, b]` instead of a list. I'm fine with splitting support to a follow-up, but then the supported list syntax should probably be explicit in tests/docs, so users don't expect all simple YAML lists to work.

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+}
+
+private object FrontMatterParserExtension : ParserExtension, TextContentRendererExtension {
+    override fun extend(parserBuilder: Builder) {
+        parserBuilder.customBlockParserFactory(FrontMatterBlockParser.Factory())
+    }
+
+    override fun extend(rendererBuilder: TextContentRenderer.Builder) {
```
rock3r, line 89: Shouldn't the text renderer output the front matter as-is here?

Right now the parser extension creates `FrontMatterBlock` / `FrontMatterNode`, but the `TextContentRendererExtension` is a no-op, so CommonMark text rendering has no renderer for those custom nodes and drops the whole block. That seems surprising for text extraction/search-ish paths; I'd expect the original front matter, including delimiters, to survive in text output. Don't recall if this was the case in Sasha's PR too, but in case, I had missed it :)

platform/jewel/markdown/extensions/gfm-tables/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/github/tables/TableBlock.kt
rock3r: Yes please

rock3r:
Re "Is this something we wish to replicate? Maybe" I'd say yes, but not something we need to do right now — could be a follow-up

rock3r (changes requested):

rock3r (approved):

rock3r (changes requested):
Wrong tab LOL sorry

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
DanielSouzaBertoldi: Oh, right. Yeah, it IS possible but honestly it's more of a headache than it should.

We have two main limitations:

1. The CommonMark API streams the content, so it's not possible for us to tell it to "go back" some lines or to completely ignore what our custom factory already consumed. Once we consume, CommonMark will not re-feed it to other parsers, we'd have to buffer what we are consuming to try and circumvent this
2. Our `processMarkdownBlock` function returns just a single `CustomBlock`, so there's no way for us to turn our FrontMatterBlock into a divider + paragraph + heading (using your example). Each one of those are a `MarkdownBlock` :/

So, to "fix" both issues, instead of doing `if (!frontMatterBlock.sawClosingDelimiter) return null` in our parser, we'd have to return a new custom block (which implies a new renderer) that will hold the re-parsed markdown blocks text in the raw buffer.

I.e, do something like:

```
if (!frontMatterBlock.sawClosingDelimiter) {
  val fallback = processor.processMarkdownDocument(frontMatterBlock.rawText)
  return RawMarkdownPassthrough(fallback)
}
```

This can be a bit sneaky though because we must make sure to not re-parse the `rawText` with the FrontMatter extension.

So, the final recipe for us to do this is basically: add a raw buffer, a new custom block + renderer, a re-parse call and a recursion guard, just for us to be able to improve a malformed-input. Given our constraints, I don't feel it's justifiable right now. A follow-up, maybe?

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: Great catch! Will implement + add unit test and a screenshot, thanks :)

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: After analyzing the code, this would require a substantial rewrite of the parser code. The model and renderer itself would be kind of a trivial change, but making our parser indent aware is the real risk here. I'm up to filing a follow-up.

But what do you mean exactly by "avoid flattening them silently"? Like, adding a log that we don't support nested key/value pairs?

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: This one is easier than the others so I'll implement here now 😎 keep em coming

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
rock3r: I mean them being rendered as if they were top-level keys, which is wrong. For example:

```md
---
metadata:
  author: Daniel
  version: "1.0"
---
```

Right now gets rendered as if it were:

```md
---
metadata: |
author: Daniel
version: "1.0"
---
```

But that is completely wrong, as the equivalent flattened representation would be:


```md
---
metadata.author: Daniel
metadata.version: "1.0"
---
```

Clearer?

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
rock3r: What about a hacky solution where a front matter block has an optional "attached coda" block that represents whatever was ingested together by the parser but should be rendered separately as a paragraph or something? Our front matter renderer extension would then render the coda and pretend it is its own content even if it isn't...

Yes, it's a hack, but I am not sure what other way we have to do this short of changing a lot of stuff like you said. But I suspect this is a late Friday idea, with many drawbacks and edge cases :D

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
DanielSouzaBertoldi: Not verbatim, no.

By the time rendering happens, the parser has already discarded the raw source and kept only parsed key/values, so the original text (fences, comments, formatting, quoting) can't be reproduced. Rather than silently drop the block (the old no-op) or fake it with `---` fences around a normalized body (which would look verbatim but isn't), I can make the text renderer emit an honest normalized reconstruction (that is, key: value per entry) so the content survives for extraction/search without misrepresenting fidelity.

True "as-is" output would require the parser to retain raw lines; which is something more fitting for a follow-up

rock3r: Reconstructed is fine tbh

DanielSouzaBertoldi: [Done!](https://github.com/JetBrains/intellij-community/pull/3556/changes#diff-1c1b2112fdde4a28a74511f782990053e444517d96a5984546ca25c6de35e0dfR103)

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: Alright, so I had to change our test + screenshot that rendered `#` as a markdown heading, since now we are just ignoring the line.

Also, to make it easier for us, my code [just ignores full-lines comments](https://github.com/JetBrains/intellij-community/pull/3556/changes#diff-a3a526e869f14f0a8ddcf7eeda325e066ac42198a8283179d8e40192df3b7735R94).

So, in the examples below:

```
title: foo # note
title: foo#note
title: "foo # bar"
url: [http://mycoolsite.com#section](http://mycoolsite.com/#section)
color: #fff
```

They'll all render as scalar types. Mainly because then we'd have to ask serious philosophical questions like: "what is a comment?" and having to deal with every single answer would make us go insane in the short term

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: Yep, thanks! 😄 I'll see what I can do about this

rock3r: @DanielSouzaBertoldi this still needs an answer (even in the form of a new issue for the follow-up)

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: [Done!](https://github.com/JetBrains/intellij-community/pull/3556/changes#diff-a3a526e869f14f0a8ddcf7eeda325e066ac42198a8283179d8e40192df3b7735R138)

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
rock3r: @DanielSouzaBertoldi this still needs an answer (even in the form of a new issue for the follow-up)

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: Yeah I just read your reply now, I'll check the code to see if we can flatten them like `metadata.nestedProperty`. If it's a cheap impl then I'll make the changes, otherwise I'll file a follow-up along with the rest :)

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
```diff
@@ -0,0 +1,306 @@
...
+    override fun closeBlock() {
+        flushAll()
+    }
+
+    private fun flushAll() {
+        flushBlockScalar()
+        flushCurrentKey()
+    }
+
+    private fun splitTopLevelCommas(inner: String): List<String> = buildList {
```
DanielSouzaBertoldi, line 241: This entire function was created by Claude. I first tried the "naive" way of just doing `inner.split(',')` but then that kinda "broke" the result for strings like `["a", "b, c", "d"]` since the middle comma was also being splitted. The list would then render as 4 items instead of 3.

Claude came in clutch and created this weird code here that does track quotes and fixes the issue entirely

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterBlockParser.kt
DanielSouzaBertoldi: We'll need a follow-up :/ The fundamental issue is making the code indent-aware nonetheless, same issue with trying to render a malformatted frontmatter block as plain markdown. I'll file a follow-up

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatterProcessorExtension.kt
DanielSouzaBertoldi: Hmmmm yeah, this could work but we'd still face the same issues I listed above. We'll need to buffer the raw markdown (including the fence ``` which our factory consumes), re-parse the raw markdown using a recursion guard and add a renderer path.

We could though make FrontMatter render its own raw markdown instead of creating a new custom block just for this, like:

```kt
sealed interface FrontMatter : CustomBlock {
    data class Table(entries: List<Entry>) : FrontMatter
    data class Raw(block: List<MarkdownBlock>) : FrontMatter
}
```

Given the effort necessary, I'd say we push this to another PR :)

DanielSouzaBertoldi (commented):
I'll file an [issue with three improvements needed](https://youtrack.jetbrains.com/issue/JEWEL-1353/Improve-FrontMatter-Capabilities) here:

- Nested tables (make the parser indent-aware)
- Render malformed FrontMatter blocks as raw markdown (will require a recursion guard, new render path and a buffer that'll store each line FrontMatter consumes)
- Alternate background colors for each row

Plus, I also updated release notes to make it clear that we do not support nested tables yet

rock3r (approved):

rock3r:
Nit: the _Parser now requires a closing --- [...]_ remark in the PR description is technically not true anymore, if you want to clarify it

DanielSouzaBertoldi:
@rock3r true!

Updated to say `* Parser now requires a closing `---`. An unterminated block ignores the entire markdown in it.`

Since, well, for a FrontMatter block to show it does need to have a closing `---` anywho 😛

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatter.kt
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * The parsed content of a YAML front matter metadata block.
+ *
+ * Front matter is rendered as a headerless two-column table where the first column contains the
+ * [entries' keys][Entry.key] and the second column their [content][Entry.value]. Scalar values are rendered as
+ * paragraphs, while list values are rendered as unordered lists, preserving the vertical structure of the block.
+ */
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public data class FrontMatter(val entries: List<Entry>) : MarkdownBlock.CustomBlock {
```
nebojsa-vuksic: Are we sure we want to mark this as a `data` class? Every new property will break binary compatibility.

rock3r: Oh that's a good catch, I had completely missed it!

rock3r (changes requested):
No data classes in the API please :)

platform/jewel/markdown/extensions/front-matter/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/frontmatter/FrontMatter.kt
DanielSouzaBertoldi: Done!

rock3r (approved):

DanielSouzaBertoldi:
Ready to merge

PR #3557, opened 2026-07-03
[KTIJ-25214]: Remove matching } when backspacing { in empty lambda

kshulzh:
https://youtrack.jetbrains.com/issue/KTIJ-25214/Automatically-delete-right-pairing-curly-brace-for-an-empty-lambda-argument

https://github.com/user-attachments/assets/d5314d56-40f6-41cc-b5c8-12d0ba4482de

### DIFF
--- a/plugins/kotlin/base/code-insight/minimal/src/org/jetbrains/kotlin/idea/editor/KotlinLambdaBackspaceHandler.kt
+++ b/plugins/kotlin/base/code-insight/minimal/src/org/jetbrains/kotlin/idea/editor/KotlinLambdaBackspaceHandler.kt
@@ -1,0 +1,64 @@
+package org.jetbrains.kotlin.idea.editor
+
+import com.intellij.codeInsight.editorActions.BackspaceHandlerDelegate
+import com.intellij.codeInsight.highlighting.BraceMatchingUtil
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiFile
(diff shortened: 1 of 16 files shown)


FredyH:
@kshulzh thanks for your contribution!
Could you please convert the `KotlinLambdaBackspaceHandler.java‎` class to Kotlin? We want to avoid adding new Java files to the code (at least in the Kotlin plugin).

Also, could you please add three more test cases to ensure that the handler will not delete the braces in the following circumstances:
1. Pressing backspace with one space between the `{` and the caret
2. Pressing backspace with a comment in the block of the lambda
3. Pressing backspace with some code in the lambda

kshulzh:
> Kirill Shulzhenko (@kshulzh) thanks for your contribution! Could you please convert the `KotlinLambdaBackspaceHandler.java‎` class to Kotlin? We want to avoid adding new Java files to the code (at least in the Kotlin plugin).
> 
> Also, could you please add three more test cases to ensure that the handler will not delete the braces in the following circumstances:
> 
> 1. Pressing backspace with one space between the `{` and the caret
> 2. Pressing backspace with a comment in the block of the lambda
> 3. Pressing backspace with some code in the lambda

Hi @FredyH, test cases added, handler from Java to Kotlin translated

FredyH:
Merged as: `543025e9a9b156fcfef9ddf0ff5cbf3cd58dd83d`

PR #3559, opened 2026-07-06
IJPL-248963 fix stale parent propagation in VcsLogJoiner

SergejSalnikov:
Root commits (no parents) in the first block were incorrectly kept in the unresolved-hashes set because the removal was guarded by `!commit.getParents().isEmpty()`. This caused `getFirstUnTrackedIndex` to scan the saved log looking for these root commit IDs, throwing `VcsLogRefreshNotEnoughDataException` when they were not found.

Additionally, `getAllNewCommits` was deduplicated against only the green-index range of the saved log rather than the full unsafe block size. When the red index exceeded the green index, commits in the gap were not checked, potentially producing duplicate entries in the merged result.

Fix both issues by:
- unconditionally removing all first-block commit IDs from the unresolved set (via a collected `firstBlockIds` set),
- computing new commits against `savedLog[0..unsafeBlockSize)` instead of `savedLog[0..greenIndex)`, ensuring correct deduplication regardless of which index is larger.

https://youtrack.jetbrains.com/issue/IJPL-248963


whitec10ud:
Thanks a lot for the contribution! We really appreciate you taking the time to investigate this and submit a fix.

The issue ([[IJPL-248963](https://youtrack.jetbrains.com/issue/IJPL-248963/VcsLogJoiner-Stale-Parent-Propagation-Bug)](https://youtrack.jetbrains.com/issue/IJPL-248963/VcsLogJoiner-Stale-Parent-Propagation-Bug)) has been fixed on our side and the fix will be available in the next IDE release.

For now, we decided to go with a smaller, targeted fix for this specific issue rather than incorporating the broader refactorings from this PR. Nevertheless, thank you again for the contribution and for bringing this to our attention!

PR #3561, opened 2026-07-06
[JEWEL-1275] Support Custom ScrollbarAdapter Injection in Scrollbars

DanielSouzaBertoldi:
# Context

While the default adapter we have in our Scrollbars are sufficient for uniform lists, they can cause significant visual instability in containers with heterogeneous item dimensions (e.g., content blocks with wildly varying heights or widths). In these cases, the default estimation logic leads to a "jumping" scrollbar thumb that changes size and position abruptly as new items are measured. To resolve this, developers need to provide specialized adapters (e.g., using caching or stable-index strategies) to ensure a smooth, professional user experience.

# Changes

- Added new overloads for `VerticallyScrollableContainer`, `HorizontallyScrollableContainer`, `VerticalScrollbar` and `HorizontalScrollbar`.
- Added a new showcase that addresses [JEWEL-1311](https://youtrack.jetbrains.com/issue/JEWEL-1311/VerticalScrollbar-goes-out-of-screen-bounds-when-LazyVerticalGrid-contains-items-with-GridItemSpan), comparing how the default scrollbar adapter behaves compared to a custom one for grids that contains items with different spans.

# Screen recording

https://github.com/user-attachments/assets/7f9aefd1-ff60-49e8-9daf-9132ccc82ac2

### New features
 * Now it's possible to provide your own scrollbar adapter for out scrollable components: ``VerticallyScrollableContainer`, `HorizontallyScrollableContainer`, `VerticalScrollbar` and `HorizontalScrollbar`.

### Deprecated API
 * Overloads for the components `VerticallyScrollableContainer`, `HorizontallyScrollableContainer`, `VerticalScrollbar` and `HorizontalScrollbar` that do not take a `adapter: ScrollAdapter` are now deprecated.


platform/jewel/ui/metalava/ui-api-0.39.0.txt
```diff
@@ -3456,6 +3462,7 @@ package org.jetbrains.jewel.ui.icons {
     field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey Frame;
     field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey Freeze;
     field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey FreezeAll;
+    field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey InspectionSeverity;
```
rock3r, line 3465: Is this because somebody else updated the icons and forgot to update the dump? Fine if it is, just mildly annoyed this still happens.

rock3r (approved):

platform/jewel/ui/metalava/ui-api-0.39.0.txt
DanielSouzaBertoldi: Yep... :/

nebojsa-vuksic (approved):

rock3r:
Ready to merge

nebojsa-vuksic:
On `BaseScrollbar`, `userAdapter ?: when(scrollState){…}`: the conditional `rememberScrollbarAdapter` call is fine (Compose handles it, and `remember(adapter, …) { SliderAdapter(...) }` already resets state correctly when the adapter identity changes), so no concern there.

One thing worth considering: now that a provided `adapter` bypasses the `when(scrollState)` type gate, a mismatched adapter/scrollState pair (which the KDoc warns about) has no runtime signal at all - the thumb (driven by the adapter) and wheel-scroll-on-track (driven by `scrollState` directly) silently desync. Would a debug-time `require`/check that the adapter is backed by `scrollState` be worth adding, to fail loudly instead of rendering a subtly broken scrollbar?

rock3r:
Hm that's a good point Nebojsa but I am not sure how we'd enforce it. I don't recall, does the adapter expose the state it's attached to?

DanielSouzaBertoldi:
@nebojsa-vuksic  @rock3r we unfortunately can't enforce this :/ `ScrollbarAdapter` is just an opaque interface:

```
package androidx.compose.foundation.v2


// kdocs removed for the sake of simplicity
interface ScrollbarAdapter {
    val scrollOffset: Double
    val contentSize: Double
    val viewportSize: Double
    suspend fun scrollTo(scrollOffset: Double)
}
```

The `rememberScrollbarAdapter` that CMP provides returns a `ScrollableScrollbarAdapter` that is internal, so we can't downcast them. And, giving the properties in the interface, we can't do `require(adapter.state == scrollState)`, there's no `state` at all. We could use reflection against internal CMP class names, but that's fragile.

I considered dropping the `scrollState` when a `userAdapter` is provided, but that wouldn't work since we need the scroll state to set other stuff like checking the visibility with `scrollState.isScrollInProgress` and `Modifier.scrollable(state = scrollState)`.

I'd argue that having this technical decision declared in the KDoc is enough given the predicament we're in. Upstream CMP scrollbars take only an adapter, so they never face this :/

PR #3562, opened 2026-07-06
[JEWEL-1334] Gradle Scripts Migration to Bazel (Part 1)

DanielSouzaBertoldi:
# Context

This is the first PR in a long list of them (idk exactly how many of them yet) which aims to migrate all possible tooling we have for gradle to Bazel.

With this PR, we have the barebones files we'll need moving forward. It also adds the `standalone_sample` target so that we don't need to run `./gradlew :samples:standalone:run` anytime we want to test something on the standalone side of things

### DIFF
--- a/.bazelproject
+++ b/.bazelproject
@@ -20,3 +20,4 @@ import_run_configurations:
   tools/bazelRunConfigurations/Run___build_android_studio.xml
   tools/bazelRunConfigurations/Run___build_i_build_target.xml
   tools/bazelRunConfigurations/Run___build_idea_community.xml
+  tools/bazelRunConfigurations/Run___build_jewel_standalone_sample.xml
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -140,6 +140,8 @@ java_binary(
 ## see community/build/tests-options.bzl
 exports_files(["intellij.idea.community.main.iml"])
 
+exports_files(glob(["build.txt"]), visibility = ["//visibility:public"])
+
 ### auto-generated section `build intellij.idea.community.main` start
 
 jvm_library(
--- a/platform/jewel/jewel_deps.MODULE.bazel
+++ b/platform/jewel/jewel_deps.MODULE.bazel
@@ -1,0 +1,22 @@
+http_file = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
+
+http_file(
+    name = "jewel_deps_kotlinpoet",
+    downloaded_file_path = "kotlinpoet.jar",
+    sha256 = "585b5673a86301e5e06d307a3327d5a73e213cb7600d1c856604ede016f1d656",
+    url = "https://https://cache-redirector.jetbrains.com/repo1.maven.org/maven2/com/squareup/kotlinpoet-jvm/2.1.0/kotlinpoet-jvm-2.1.0.jar",
+)
+
+http_file(
+    name = "jewel_deps_ktfmt",
+    downloaded_file_path = "ktfmt.jar",
+    sha256 = "d13324023754112797d828e2f92117196190f7617d16ad1d1765e5be768cc5c2",
+    url = "https://cache-redirector.jetbrains.com/repo1.maven.org/maven2/com/facebook/ktfmt/0.55/ktfmt-0.55-with-dependencies.jar",
+)
+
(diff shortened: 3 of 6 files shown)


platform/jewel/jewel_deps.MODULE.bazel
```diff
@@ -0,0 +1,22 @@
...
+    sha256 = "d13324023754112797d828e2f92117196190f7617d16ad1d1765e5be768cc5c2",
+    url = "https://repo1.maven.org/maven2/com/facebook/ktfmt/0.55/ktfmt-0.55-with-dependencies.jar",
+)
+
+http_file(
+    name = "jewel_deps_detekt",
+    downloaded_file_path = "detekt.jar",
+    sha256 = "3afe89a11120303c73c9bdda3d8fe558dd9070a6937d27819ddc04b275381245",
+    url = "https://repo1.maven.org/maven2/io/gitlab/arturbosch/detekt/detekt-cli/1.23.8/detekt-cli-1.23.8-all.jar"
+)
```
DanielSouzaBertoldi, line 22: This class holds the dependencies we'll need for these (and future) tools. Most of them are fat jars so that we don't have to explicitly list every single transitive dependency for each one of them

.bazelproject
```diff
@@ -19,4 +19,5 @@ run_config_run_with_bazel: false
 import_run_configurations:
   tools/bazelRunConfigurations/Run___build_android_studio.xml
   tools/bazelRunConfigurations/Run___build_i_build_target.xml
-  tools/bazelRunConfigurations/Run___build_idea_community.xml
\ No newline at end of file
+  tools/bazelRunConfigurations/Run___build_idea_community.xml
+  tools/bazelRunConfigurations/Run___build_jewel_standalone_sample.xml
```
DanielSouzaBertoldi, line 23: maybe this is a bit of a stretch but with this the jewel standalone sample will be automatically added to the configurations list 😛 no need to run it first so that it gets added

BUILD.bazel
```diff
@@ -1194,3 +1194,5 @@ exports_files(
     visibility = ["//visibility:public"],
 )
 ### auto-generated section `iml intellij.idea.community.main.tests` end
+
+exports_files(glob(["build.txt"]), visibility = ["//visibility:public"])
```
DanielSouzaBertoldi: Our Generate Themes script has to read the build version. Since this is a kick off PR, figured I made this change in visibility right now. The script itself will come later.

platform/jewel/jewel_deps.MODULE.bazel
```diff
@@ -0,0 +1,22 @@
+http_file = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
+
+http_file(
+    name = "jewel_deps_kotlinpoet",
+    downloaded_file_path = "kotlinpoet.jar",
+    sha256 = "585b5673a86301e5e06d307a3327d5a73e213cb7600d1c856604ede016f1d656",
+    url = "https://repo1.maven.org/maven2/com/squareup/kotlinpoet-jvm/2.1.0/kotlinpoet-jvm-2.1.0.jar",
```
shalupov: please use https://cache-redirector.jetbrains.com/repo1.maven.org/maven2/... prefix, so we won't get random failures when maven central rate limits us again

.bazelproject
shalupov: No objections, but probably you need to make this change along with adding tools/bazelRunConfigurations/Run___build_jewel_standalone_sample.xml ?

shalupov (changes requested):

.bazelproject
DanielSouzaBertoldi: You're right! When I unstashed my work for some reason the file itself was missing lol just pushed it, thanks :)

shalupov (approved):

rock3r (approved):

nebojsa-vuksic (approved):

rock3r:
@DanielSouzaBertoldi should we ship?

DanielSouzaBertoldi:
@rock3r let's do it 😎 part 2 coming right up!

DanielSouzaBertoldi:
Ready to merge

mfilippov (approved):
LGTM

PR #3563, opened 2026-07-07
[JEWEL-1319] Draw circular progress indicator on Canvas

rock3r:
`CircularProgressIndicator` currently builds spinner frames by generating SVG strings, converting them to byte arrays, and decoding them into painters. This replaces that pipeline with direct Canvas drawing while preserving the existing public API, styling, frame timing, size variants, and intended visual output.

## Changes

* Reimplemented `CircularProgressIndicator` and `CircularProgressIndicatorBig` on top of a shared Canvas-based implementation.
* Preserved `CircularProgressStyle` color fallback and `frameTime` behavior.
* Kept the existing `loadingDispatcher` parameter for API compatibility, though it is no longer needed now that there is no background SVG decoding work.
* Matched the original SVG segment winding by keeping the animated rotation positive and placing individual segments with negative per-segment rotation.
* Derived segment geometry from the actual draw scope size so the spinner scales correctly with the final Canvas constraints.

## Performance validation

* Local temporary recomposition harness, not committed: advanced the Compose test clock by 80 frames at 16 ms per frame and compared an old-style composition-read animation path with the new Canvas draw-phase path. Result: old-style path recomposed 11 times; new Canvas path recomposed 1 time.
* Local throwaway JVM micro-harness, not committed: compared old SVG string + byte-array preparation against the new geometry calculation path. This intentionally excluded `decodeToSvgPainter`, so it undercounts the removed work. Old SVG string + byte preparation measured roughly 7.7–11.3 µs/op; new geometry calculation measured roughly 0.04–0.12 µs/op.
* No screenshots are included because this is intended to preserve the existing spinner visuals; the change is focused on removing unnecessary composition and SVG decode work.

## Testing

* `cd platform/jewel && ./gradlew check --continue --no-daemon`
* `cd platform/jewel && ./gradlew detekt detektMain detektTest --continue --no-daemon`
* `cd platform/jewel && ./gradlew :ui:checkMetalavaApi --no-daemon --max-workers=1`
* `cd platform/jewel && ./gradlew :ui:validatePublicApi --no-daemon --max-workers=1`
* `cd platform/jewel && ./gradlew :ui:test --tests org.jetbrains.jewel.ui.component.CircularProgressIndicatorRecompositionHarnessTest --no-daemon` using a temporary local-only test harness, removed after collecting the counts.
* `git diff --check`

Could not complete `./bazel.cmd build //platform/jewel/ui:ui` locally because Bazel dependency fetching was blocked with `Operation not permitted` while downloading `rules_kotlin` into the local `MonorepoBazel` cache.

### Bug fixes

* Improved `CircularProgressIndicator` performance by drawing the spinner directly on Canvas instead of generating and decoding SVG frames.

### DIFF
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/CircularProgressIndicator.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/CircularProgressIndicator.kt
@@ -1,32 +1,29 @@
 package org.jetbrains.jewel.ui.component
 
-import androidx.compose.animation.core.InfiniteRepeatableSpec
 import androidx.compose.animation.core.LinearEasing
 import androidx.compose.animation.core.RepeatMode
-import androidx.compose.animation.core.VectorConverter
-import androidx.compose.animation.core.animateValue
+import androidx.compose.animation.core.animateFloat
+import androidx.compose.animation.core.infiniteRepeatable
 import androidx.compose.animation.core.rememberInfiniteTransition
 import androidx.compose.animation.core.tween
-import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.Canvas
 import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.produceState
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


nebojsa-vuksic (approved):

DanielSouzaBertoldi (approved):

rock3r:
Ready to merge

PR #3564, opened 2026-07-07
[JEWEL-1355] Select first match when the selected item is filtered out

DanielSouzaBertoldi:
## Context

Typing a character triggers two reactions from the same state change, and they run on independent reactive graphs with nothing coordinating their order. One re-filters the list, recomposes, and drops the previously selected item from the selection when it no longer matches. The other, a `LaunchedEffect` in `SpeedSearchableLazyColumnScrollEffect`, reacts to the new `matchingIndexes` and decides what to select. Both hop through `flowOn(dispatcher)`, so the selection effect can run either while the old selection is still present or after the filter has already cleared it.

Walking through typing "Angular" while "Spring Boot" is selected (Spring Boot is not an Angular match, so it gets filtered out): when the effect runs while the old selection is still there, it picks the visible match closest to that selection, lands on "Angular", and the choice sticks. When the filter runs first, the effect reaches its selection logic with an empty selection. The "pick the closest visible match" step needs an existing selection to measure distance from, so with nothing to anchor to it produces nothing, and control slips into the next branch, which was written for matches that are scrolled off-screen. That branch computes the first match after the last visible item, which skips index 0 and selects "AngularJS". It then stays wrong, because on the following round "AngularJS" is itself a valid match and gets kept.

So the outcome depended purely on which reaction won the race. The correct answer was never guaranteed, only likely (wrap your head around this!).

```mermaid
flowchart TD
    K["User types 'Angular'"] --> S["attach(): one snapshot sets<br/>currentMatcher + matchingIndexes"]

    S --> A["Graph A — derivedStateOf re-filters the list<br/>LazyColumn recomposes<br/>'Spring Boot' no longer matches → selectedKeys becomes empty"]
    S --> B["Graph B — scroll effect reacts to matchingIndexes"]

    A -. "race: no ordering between the two" .-> B

    B --> Q1{"Is a currently selected<br/>item still a match?"}
    Q1 -- yes --> KEEP["Keep it selected"]
    Q1 -- "no (selection empty)" --> Q2{"bestVisibleMatch:<br/>visible match closest to<br/>an existing selection"}

    Q2 -- "selection exists →<br/>closest match found" --> GOOD["Select 'Angular' (index 0)"]
    Q2 -- "selection empty →<br/>produces null" --> FALL["Fall-through branch<br/>(meant for OFF-screen matches):<br/>first match after visibleItemsRange.last"]

    FALL --> BUG["Selects 'AngularJS' (index 1)<br/>and keeps it on the next round"]

    Q2 -. "FIX: ?: indexOfVisibleMatches.firstOrNull()" .-> FIX["Select first visible match<br/>= 'Angular' (index 0)"]

    style BUG fill:#4a1f1f,stroke:#cc4444,color:#fff
    style FALL fill:#4a3a1f,stroke:#cc9944,color:#fff
    style FIX fill:#1f4a2a,stroke:#44cc77,color:#fff
    style GOOD fill:#1f4a2a,stroke:#44cc77,color:#fff
```

## Changes

* In `SpeedSearchableLazyColumnScrollEffect`, when there is no current selection to anchor to, fall back to the first visible match (`?.first ?: indexOfVisibleMatches.firstOrNull()`) instead of leaking into the off-screen-scroll branch. First-match selection no longer depends on winning the race: whichever reaction runs first, the empty-selection case now resolves to the first result on its own. When a selection does exist, behavior is unchanged.
* Added a regression test that reproduces the losing side of the race deterministically. Instead of hoping CI timing lines up, it clears the selection first (the exact state the re-filter produces), types a query with two visible matches, and asserts the first one is selected. With the fix reverted the test selects "AngularJS" and fails; with the fix it passes. The previously flaky `navigation should cycle through filtered items only` test is stable for the same reason.

This also matches the IntelliJ Platform behavior: both `SpeedSearchBase` (no selection starts the scan at index 0 and returns the first match) and `ListWithFilter` / `ListPopupImpl` (the previously selected row leaving the filtered set falls back to row 0) resolve to the first result in this case.

* Documented the full selection algorithm on `SpeedSearchableLazyColumnScrollEffect`. With the fix in place, the no-matching-selection behavior forms one rule: a forward scan from the top of the viewport that wraps to the top of the list (topmost visible match → first match after the viewport → first match from the top). This preserves the user's rough position and mirrors Swing's `SpeedSearchBase.findElement` rather than `ListWithFilter`'s always-select-row-0.
* Added three tests pinning that rule with a viewport scrolled mid-list: matches inside the viewport (topmost visible wins), matches only below (first match below the viewport is preferred over the global first), and matches only above (wraps to the first match from the top, not the nearest one above).

## Screen recordings

No way for me to screen record this since it's a race condition :( but there is a regression unit test that verifies the fix is working as intended :)

### Bug fixes

* Fixed `SpeedSearchArea` over a filtered list selecting the second match instead of the first when the previously selected item was filtered out of the results.

### DIFF
--- a/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchAreaFilteringTest.kt
+++ b/platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchAreaFilteringTest.kt
@@ -37,6 +37,7 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.resetMain
 import kotlinx.coroutines.test.setMain
+import org.jetbrains.jewel.foundation.lazy.SelectableLazyListState
 import org.jetbrains.jewel.foundation.lazy.rememberSelectableLazyListState
 import org.jetbrains.jewel.foundation.search.EmptySpeedSearchMatcher
 import org.jetbrains.jewel.foundation.search.filter
@@ -73,6 +74,8 @@ class SpeedSearchAreaFilteringTest {
     @get:Rule val rule = createComposeRule()
 
     private val testDispatcher = UnconfinedTestDispatcher()
+
+    private var capturedListState: SelectableLazyListState? = null
 
     private val ComposeContentTestRule.onLazyColumn
         get() = onNodeWithTag("LazyColumn")
@@ -209,6 +212,20 @@ class SpeedSearchAreaFilteringTest {
         onLazyColumn.performKeyPress(Key.DirectionUp, rule = this)
         onLazyColumnItem("Angular").assertIsDisplayed().assertIsSelected()
     }
+
+    @Test
+    fun `when the previously selected item is filtered out, the first visible match is selected`() =
+        runFilteringComposeTest {
+            capturedListState!!.selectedKeys = emptySet()
+            waitForIdle()
+
+            // Type "Angular" - two matches, both visible, first one at index 0 of the filtered list.
+            onLazyColumn.performKeyPress("Angular", rule = this)
+
+            // The first match must be selected, never the second one.
+            onLazyColumnItem("Angular").assertIsDisplayed().assertIsSelected()
+            onLazyColumnItem("AngularJS").assertIsDisplayed()
+        }
 
     @Test
     fun `partial match filtering should work correctly`() = runFilteringComposeTest {
@@ -442,6 +459,7 @@ class SpeedSearchAreaFilteringTest {
             val speedSearchState = rememberSpeedSearchState()
             capturedSpeedSearchState = speedSearchState
             val listState = rememberSelectableLazyListState()
+            capturedListState = listState
 
             // Filter the list based on the current matcher - same pattern as showcase
             val filteredItems by remember { derivedStateOf { listEntries.filter(speedSearchState.currentMatcher) } }
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
@@ -281,6 +281,22 @@ internal class SpeedSearchableLazyColumnScopeImpl(
     }
 }
 
+/**
+ * Keeps the list selection in sync with the speed-search matches.
+ *
+ * Whenever [SpeedSearchState.matchingIndexes] changes while the search is active, the selection is decided as follows:
+ * 1. If a currently selected item still matches, keep it selected (scrolling to it if it is offscreen).
+ * 2. Otherwise, if a selection exists, select the visible match closest to it.
+ * 3. Otherwise, select the topmost visible match.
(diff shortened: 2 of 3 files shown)


platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchAreaFilteringTest.kt
```diff
@@ -210,6 +213,20 @@ class SpeedSearchAreaFilteringTest {
         onLazyColumnItem("Angular").assertIsDisplayed().assertIsSelected()
     }
 
+    @Test
+    fun `when the previously selected item is filtered out, the first visible match is selected`() =
+        runFilteringComposeTest {
+            capturedListState!!.selectedKeys = emptySet()
+            waitForIdle()
```
DanielSouzaBertoldi, line 220: This may seem dumb but it reproduce the states that appeared on CI: the item that was selected before typing gets filtered out, so the selection set is empty at the moment the new matches arrive. This is the interleaving that used to leak into the "scroll to an off-screen match" branch and select a later match instead of the first one.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
```diff
@@ -358,7 +359,7 @@ internal fun SpeedSearchableLazyColumnScrollEffect(
                                 ?.let { visibleMatchIndex to it }
                         }
                         .minByOrNull { (it.first - it.second).absoluteValue }
-                        ?.first
+                        ?.first ?: indexOfVisibleMatches.firstOrNull()
```
rock3r, line 378: How are we deciding which match to select here exactly? Looks like this first picks a visible match, then the first one after the viewport, then the first one before it — so it’s neither consistently the first match nor the closest one.

I could be okay with preserving the user’s rough position, that might even feel nicer, but then can we make that intentional and add coverage for matches above, inside, and below the viewport?

rock3r (changes requested):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/search/SpeedSearchableLazyColumn.kt
DanielSouzaBertoldi: Yeah the code may seem a bit over the place but it's actually doing what its JBUI counterpart does (`com.intellij.ui.SpeedSearchBase.findElement()` if you want to check it out), which is basically to just scan forward from the top of the viewport and wrap around to the top of the list, so "topmost visible match" is the first hit of that scan, "first after the viewport" is the scan continuing past the visible range, and "first before it" is the wrap.

I added a KDoc making this behavior more explicit + added 3 tests covering the scenarios you mentioned in your second option :)

rock3r: Cool cool, that's fine then. Thanks! I was just a bit thrown off by the behaviour as it seemed a bit underdefined

rock3r (approved):

nebojsa-vuksic (approved):

rock3r:
Ready to merge

PR #3565, opened 2026-07-07
IJPL-249299 fix: Keep stripped empty lines when copying a vertical section in editor (and reworked terminal)

bric3:
Fix block selection copy when the selected rectangle crosses lines shorter than the selected columns.

See https://youtrack.jetbrains.com/issue/IJPL-249299/Column-selection-copy-blank-lines-are-stripped-from-pasted-text

The editor can omit carets for rows that are fully in virtual space. Copy then used to concatenate only caret selections, so those logical rows disappeared from the clipboard. This preserves those rows as empty lines while keeping raw rectangular text available for IDE paste paths and copy preprocessors.

The fix stores the logical block-selection row bounds, reuses them in `TextBlockTransferable`, and applies the same conversion through editor copy, direct copy-paste helper usage, and terminal copy-on-selection.
I believe the newly created VirtualSelectionText to be interesting, but it's integration in the API feels clumsy as well.

https://github.com/user-attachments/assets/64111ac1-5826-482a-85d8-40638d9f139f

----

I honestly asked some help from agents for this, but I'm quite unsure about the outcome, for example I choose to add a new internal API in `TextBlockTransferable` because that's a class that is available in both modules, and there's already a static method there. But this doesn't look great is it surfaces an internal API.


bric3:
I'd also reach out to @KonstantinHudyakov as it might interest him for the terminal "component".

PR #3566, opened 2026-07-09
[collab] Add a hover Jump to Source icon to the review changes tree

jiec-msft:
## What

Adds a hover "Jump to Source" icon to file rows of the code review changes tree (used by the GitHub pull request view). Hovering a file row now shows an `EditSource` pencil icon next to the viewed-state checkbox; a single click opens the file in the editor.

## Why

In the PR changes tree, opening the actual file currently requires the context menu (*Jump to Source*) or F4 — a double click is already taken by the diff preview. The Commit / Local Changes tree solves the same problem with hover icons (`ChangesViewNodeAction` + `HoverChangesTree`), but that mechanism does not apply here because the review tree uses its own renderer (`CodeReviewProgressRenderer`). This adds the equivalent affordance to that renderer.

## How

- `CodeReviewProgressRenderer`: a fixed-width icon slot (same pattern as the viewed-state checkbox slot, with an invisible placeholder) is reserved for every leaf row, so the row layout never shifts on hover. The icon only renders for the hovered row and only when the file resolves to a local `VirtualFile` (deleted files show nothing).
- `CodeReviewProgressTreeModel`: the existing checkbox hit-test click handler is generalized into `installRendererComponentClickHandler` and reused for the new icon; a click navigates via `OpenFileDescriptor`.

## Validation

- `bazel build //platform/collaboration-tools:collaboration-tools` passes.
- Manually reviewed against the checkbox hover behavior for layout-shift and hit-testing consistency.

## Note

I could not find an existing YouTrack ticket for this. I understand features need prior agreement per CONTRIBUTING.md — happy to file a YouTrack issue and link it here, or adjust/split the change as you prefer.

## Test

Consistent with context menu:

(1) The icon shows when the `Jump to Source` is in the context menu (right click menu)

image

(2) No icon shows for the file if the file don't have `Jump to Source` in the context menu.

image

### DIFF
--- a/platform/collaboration-tools/src/com/intellij/collaboration/ui/codereview/CodeReviewProgressRenderer.kt
+++ b/platform/collaboration-tools/src/com/intellij/collaboration/ui/codereview/CodeReviewProgressRenderer.kt
@@ -4,7 +4,11 @@ package com.intellij.collaboration.ui.codereview
 import com.intellij.collaboration.ui.HorizontalListPanel
 import com.intellij.collaboration.ui.layout.SizeRestrictedSingleComponentLayout
 import com.intellij.collaboration.ui.util.DimensionRestrictions
+import com.intellij.collaboration.util.RefComparisonChange
+import com.intellij.collaboration.util.filePath
+import com.intellij.icons.AllIcons
 import com.intellij.openapi.vcs.changes.ui.ChangesBrowserNode
+import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.ui.CellRendererPanel
 import com.intellij.ui.ClientProperty
 import com.intellij.ui.ColoredTreeCellRenderer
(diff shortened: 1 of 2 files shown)


jiec-msft:
Based on recent commit author on GitHub plugin, invite @bartekjansz @nataliaponomareva-cloud to take a look, thanks!

iasemenov:
Thanks a lot for the contribution! It would be great if you could create a YouTrack issue for this.
Meanwhile, we'll discuss the implementation with the designers to align the UI a little bit better

iasemenov:
Thank you for the contribution! This was implemented as https://youtrack.jetbrains.com/issue/IJPL-251890

jiec-msft:
@iasemenov Thank you for following up and make it better! Next time, I would create a YouTrack issue.

PR #3573, opened 2026-07-11
[KTIJ-38341]: Mark functional interface argument contexts as blocking

kshulzh:
https://youtrack.jetbrains.com/issue/KTIJ-38341/False-positive-Possibly-blocking-call-in-non-blocking-context-could-lead-to-thread-starvation
Here I added FunctionalInterface lambdas inspection support 
image
I have a question about the full blocking/nonblocking inspection. It looks like inline functions are analyzed incorrectly. In the following example, the Thread.sleep call inside the inline run block is not highlighted, while the subsequent Thread.sleep call is
image


FredyH:
@kshulzh Thanks for the contribution. Unfortunately, your change came at the same time as I was reworking the non-blocking/blocking detection in Kotlin in [KTIJ-32716](https://youtrack.jetbrains.com/issue/KTIJ-32716).
The fix for that issue also fixes [KTIJ-38341](https://youtrack.jetbrains.com/issue/KTIJ-38341) by no longer going through the argument mapping altogether, which means this change is no longer compatible.

PR #3574, opened 2026-07-13
[github] IJPL-250160 Reduce redundant REST /pulls requests after push

joshfree:
YouTrack: [IJPL-250160](https://youtrack.jetbrains.com/issue/IJPL-250160)

After a successful push, `GHPushNotificationCustomizer` calls `GET /repos/{owner}/{repo}/pulls` to decide whether to offer "Create pull request" or "Open existing pull request". It only needs to know whether **0, 1, or >1** PR exists, but `PullRequests.find(...)`  sets no `per_page` and fetches the default 30. The PR-creation lookup (`GHPRCreationServiceImpl.findOpenPullRequest`) uses only the first result but also fetches 30. Re-pushing the same branch head repeats the identical request, unnecessarily burning GitHub API calls for a result that hasn't changed.

This fix means smaller responses and fewer requests per push — less network work and modestly lighter use of the user's GitHub API rate limit.

 ### Changes

 - Add an optional `pagination: GithubRequestPagination?` parameter to `GithubApiRequests.Repos.PullRequests.find`.
 - Pass `per_page=1` from the creation lookup (first match only) and `per_page=2` from the push notification (`1` would break the ">1" case).
 - Cache the push-notification lookup (time- and size-bounded, keyed by `(server, repository, remote branch, pushed revision)`) so re-pushing an unchanged head reuses the previous result.

 Behavior is unchanged: both `find` callers keep the same first-match / count semantics.

 ### Tests

 - `GithubApiRequestsPullRequestsFindTest` — `find` targets `/pulls`, omits `per_page` by default, emits `per_page`/`page` when paginated.
 - `GHPushNotificationLookupCacheTest` — hit within TTL, expiry, key sensitivity, overwrite/refresh, size eviction.

### DIFF
--- a/plugins/github/github-core/src/org/jetbrains/plugins/github/api/GithubApiRequests.kt
+++ b/plugins/github/github-core/src/org/jetbrains/plugins/github/api/GithubApiRequests.kt
@@ -615,11 +615,13 @@ object GithubApiRequests {
         state: GithubIssueState? = null,
         baseRef: String? = null,
         headRef: String? = null,
+        pagination: GithubRequestPagination? = null,
       ): GithubApiRequest<GithubResponsePage<GHPullRequestRestIdOnly>> =
         Get.jsonPage<GHPullRequestRestIdOnly>(getUrl(repository, urlSuffix, urlQuery {
           param("state", state?.toString())
           param("base", baseRef)
           param("head", headRef)
+          param(pagination)
         }))
           .withOperation(GithubApiRequestOperation.RestGetPullRequests)
           .withOperationName("find pull requests")
--- a/plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationCustomizer.kt
+++ b/plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationCustomizer.kt
@@ -149,7 +149,7 @@ internal class GHPushNotificationCustomizer(private val project: Project) : GitP
       remoteBranch = remoteBranchName,
       sourceRevision = sourceRevision,
     )
-    lookupCache.get(cacheKey, System.currentTimeMillis())?.let { return it }
+    lookupCache.get(cacheKey)?.let { return it }
 
     val accountManager = serviceAsync<GHAccountManager>()
     val token = accountManager.findCredentials(account) ?: return emptyList()
@@ -162,7 +162,7 @@ internal class GHPushNotificationCustomizer(private val project: Project) : GitP
                           // only the 0/1/>1 distinction matters, so two results are enough
                           pagination = GithubRequestPagination(pageSize = 2))
       ).items
-      lookupCache.put(cacheKey, existingPrs, System.currentTimeMillis())
+      lookupCache.put(cacheKey, existingPrs)
       existingPrs
     }
     catch (ce: CancellationException) {
--- a/plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationCustomizer.kt
+++ b/plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationCustomizer.kt
@@ -22,6 +22,7 @@ import org.jetbrains.plugins.github.api.GithubApiRequests.Repos.PullRequests
 import org.jetbrains.plugins.github.api.data.GithubIssueState
 import org.jetbrains.plugins.github.api.data.pullrequest.GHPullRequestRestIdOnly
 import org.jetbrains.plugins.github.api.data.pullrequest.toPRIdentifier
+import org.jetbrains.plugins.github.api.data.request.GithubRequestPagination
 import org.jetbrains.plugins.github.api.executeSuspend
 import org.jetbrains.plugins.github.authentication.accounts.GHAccountManager
 import org.jetbrains.plugins.github.authentication.accounts.GithubAccount
@@ -37,6 +38,8 @@ import java.util.concurrent.CancellationException
 private val LOG = logger<GHPushNotificationCustomizer>()
 
 internal class GHPushNotificationCustomizer(private val project: Project) : GitPushNotificationCustomizer {
+  private val lookupCache = GHPushNotificationLookupCache()
+
   override suspend fun getActions(
     repository: GitRepository,
     pushResult: GitPushRepoResult,
@@ -68,7 +71,7 @@ internal class GHPushNotificationCustomizer(private val project: Project) : GitP
       return emptyList()
     }
 
-    val existingPrs = findExistingPullRequests(projectMapping, account, remoteBranch)
+    val existingPrs = findExistingPullRequests(projectMapping, account, remoteBranch, pushResult.sourceBranch)
     return when (existingPrs.size) {
       0 -> {
         listOf(GHPRCreatePullRequestNotificationAction(project, projectMapping, account))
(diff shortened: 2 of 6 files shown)


plugins/github/github-core/src/org/jetbrains/plugins/github/api/GithubApiRequests.kt
```diff
@@ -615,11 +615,13 @@ object GithubApiRequests {
         state: GithubIssueState? = null,
         baseRef: String? = null,
         headRef: String? = null,
+        pagination: GithubRequestPagination? = null,
       ): GithubApiRequest<GithubResponsePage<GHPullRequestRestIdOnly>> =
         Get.jsonPage<GHPullRequestRestIdOnly>(getUrl(repository, urlSuffix, urlQuery {
           param("state", state?.toString())
           param("base", baseRef)
           param("head", headRef)
+          param(pagination)
```
anuchandy, line 624: The other usage of `params` function seems to pass a string param (likely describing the param name for a better user message/logging) - do we need to align this?

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationLookupCache.kt
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * The cache key includes the pushed [Key.sourceRevision]: when a push carries a concrete
+ * commit hash (detached/revision push) a new revision is a natural cache miss, so genuinely
+ * new state is always re-fetched. For ordinary branch pushes the revision is the branch ref
+ * (stable between pushes), so entries additionally expire after [ttlMs] to bound staleness.
+ *
+ * All access is synchronized; the customizer is a project-level extension whose
+ * `getActions` is a suspend function that may be invoked concurrently.
+ */
+internal class GHPushNotificationLookupCache @VisibleForTesting constructor(
```
anuchandy: Since this module already depends on the IntelliJ platform, Guava's CacheBuilder (com.google.common.cache, bundled with the platform - no new dependency) likely available and gives us cache for free: expireAfterWrite for TTL, maximumSize for the memory bound, and it's thread-safe by default. 

Worth considering as an alternative to the hand-rolled cache. (It also offers single-flight loading via get(key, loader), which would de-dup concurrent misses - something the current split get/put doesn't.

Otherwise the class `GHPushNotificationLookupCache` looks good to me - probably we should add a comment for get and put indicating - " The returned list is the same instance held in the cache (not a copy) and MUST NOT be mutated; mutating it would corrupt the entry returned to later callers.."

srnagar: +1 to reusing existing cache instead of creating a new one here.

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationLookupCache.kt
```diff
@@ -0,0 +1,80 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code
```
srnagar: It seems to match the other files but for new file, should we update the copyright time range?

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationLookupCache.kt
```diff
@@ -0,0 +1,80 @@
...
+  @Synchronized
+  fun put(key: Key, value: List<GHPullRequestRestIdOnly>, nowMs: Long) {
+    entries.entries.removeAll { isExpired(it.value, nowMs) }
+    // Re-insert so the freshest key moves to the end of the insertion-ordered map.
+    entries.remove(key)
+    entries[key] = Entry(value, nowMs)
+    val iterator = entries.keys.iterator()
+    while (entries.size > maxEntries && iterator.hasNext()) {
+      iterator.next()
+      iterator.remove()
```
srnagar: We should consider evicting the least recently accessed entry in the cache instead of the oldest entry to get added to the cache as that might evict an entry that has a high hit rate.

plugins/github/github-core/src/org/jetbrains/plugins/github/api/GithubApiRequests.kt
joshfree: This is the pattern established in the file (lines 121, 240, 505, 585, 756) since pagination contains both `page` and `per_page`; I just kept it consistent with the rest of the file.

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationLookupCache.kt
joshfree: thanks! switched to Guava CacheBuilder

joshfree:
Thanks for the review @anuchandy and @srnagar - I appreciate it

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationLookupCache.kt
```diff
@@ -0,0 +1,70 @@
...
+  constructor() : this(DEFAULT_TTL_MS, DEFAULT_MAX_ENTRIES, Ticker.systemTicker())
+
+  internal data class Key(
+    val server: String,
+    val repositoryPath: String,
+    val remoteBranch: String,
+    val sourceRevision: String,
+  )
+
+  private val cache: Cache<Key, List<GHPullRequestRestIdOnly>> =
```
iasemenov: I don't think we need to abstract away the cache, since it's just a direct wrapper at this point and the test is effectively testing the library instead of our code

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationCustomizer.kt
```diff
@@ -125,21 +128,42 @@ internal class GHPushNotificationCustomizer(private val project: Project) : GitP
...
-    val repository = repositoryMapping.repository
-    val remoteBranchName = branch.nameForRemoteOperations
-
     return try {
-      executor.executeSuspend(PullRequests.find(repository, GithubIssueState.open, baseRef = null, headRef = repository.repositoryPath.owner + ":" + remoteBranchName)).items
+      val existingPrs = executor.executeSuspend(
+        PullRequests.find(repository, GithubIssueState.open, baseRef = null,
+                          headRef = repository.repositoryPath.owner + ":" + remoteBranchName,
+                          // only the 0/1/>1 distinction matters, so two results are enough
+                          pagination = GithubRequestPagination(pageSize = 2))
```
iasemenov, line 153: I'm not sure the server can ever return more than one result with these filters

iasemenov (changes requested):

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationLookupCache.kt
iasemenov: Also `Caffeine` cache implementation is generally recommended over the Guava one
This is non-documented arcane knowledge, unfortunately :(

joshfree: Good call — removed the `GHPushNotificationLookupCache` wrapper and its test, and inlined the cache directly in `GHPushNotificationCustomizer` (commit d0608b0). It now uses Caffeine (`com.github.benmanes.caffeine.cache`, bundled with the platform) instead of Guava, with the same 1-minute expire-after-write / 100-entry bounds. Caffeine's TinyLFU eviction also addresses the earlier note about evicting by recency/frequency rather than insertion order. Thanks for the Caffeine pointer.

plugins/github/github-core/src/org/jetbrains/plugins/github/notification/GHPushNotificationCustomizer.kt
joshfree: These filters constrain `head` (`owner:branch`) and `state=open` but leave `base` unconstrained (`baseRef = null`). GitHub allows multiple simultaneously-open PRs from the same head branch when they target different base branches, so this call can legitimately return more than one row — that's why the customizer treats `>1` as ambiguous (offers no action) and fetches `pageSize = 2` to tell 0 / 1 / many apart. I've added a comment documenting this in d0608b0. If you'd prefer the customizer only consider PRs against a single base, I'm happy to constrain `base` and simplify this to a boolean existence check instead — just let me know which you'd rather.

joshfree:
@iasemenov this is ready for another look — pushed d0608b0 addressing your feedback:

- Dropped the `GHPushNotificationLookupCache` wrapper and its test; inlined a Caffeine cache directly in `GHPushNotificationCustomizer` (same 1-min / 100-entry bounds).
- Replied inline on the `>1` results question — it's reachable because `base` is unconstrained, but happy to constrain it and simplify to an existence check if you'd prefer.

I don't have permission to formally re-request review on this repo, so flagging here instead. Thanks!

iasemenov:
I've looked a little deeper into it, and I'm inclined to only merge the pagination part, not the caching.
The reasons are:
* push operation (even for heavy pushers) is rare enough, that this check is unlikely to cause a huge strain on the API
* PR creation is not strictly tied to a revision push, so a cached check might not be accurate, which is not a great UX

I would like to propose the following alternative:
* we will look into dropping this request completely in favour of parsing the Git response for the "create PR prompt"
* we will also look into making this request conditional once the API migration is complete

Would this be acceptable?

iasemenov (changes requested):

joshfree:
Thanks for digging in — that reasoning makes sense, and I'm happy to go pagination-only.

I've pushed a commit ( + "64aede5" + ) that **drops the caching** (removed the Caffeine cache, its key type, and the revision plumbing) and **keeps just the pagination**: `per_page=1` for the PR-creation lookup (first match only) and `per_page=2` for the push-notification check (enough to tell 0 / 1 / >1 apart). The `GithubApiRequestsPullRequestsFindTest` pagination test stays; no caching code or tests remain.

On your two proposals — both sound great, and they're better than caching:

1. **Dropping the request entirely** by parsing the Git push response for the "create PR" prompt — agreed this is the ideal end state; happy to help however's useful.
2. **Making the request conditional** (ETag / `If-None-Match` -> `304`) once the API migration completes — strongly in favor; that's the direction we're driving for these polling paths generally.

(I don't have permission to formally re-request review here, so flagging in-thread. Thanks again!)

joshfree:
Rebased onto current `master` and resolved the overlap in `184167c`.

Upstream now already exposes the `pagination` parameter on `PullRequests.find`, so the refreshed diff is narrower: it passes `per_page=1` for the first-match creation lookup and `per_page=2` where the notification needs to distinguish 0 / 1 / multiple PRs, plus the request URL regression tests. No cache code remains.

All prior review conversations are resolved. Ready for another look when you have capacity.

iasemenov:
The changes were merged, thank you!

PR #3575, opened 2026-07-14
https://youtrack.jetbrains.com/issue/IJPL-173018/Scroll-to-results-du…

momomo:
Scroll to results during typing  in project view option - stop jumping around in Project View

https://youtrack.jetbrains.com/issue/IJPL-173018/Scroll-to-results-during-typing-in-project-view-option-stop-jumping-around-in-Project-View

This does not attempt to solve it in a clean way .. but does it through the registry which is good enough for me

### DIFF
--- a/platform/core-api/src/com/intellij/openapi/application/ApplicationListener.java
+++ b/platform/core-api/src/com/intellij/openapi/application/ApplicationListener.java
@@ -19,7 +19,7 @@ public interface ApplicationListener extends EventListener {
   }
 
   /**
-   * This method is called before restarting an application (e.g., after installing a plugin). 
+   * This method is called before restarting an application (e.g., after installing a plugin).
    * Return {@code true} if the application can be restarted, or {@code false} to cancel restarting.
    */
   default boolean canRestartApplication() {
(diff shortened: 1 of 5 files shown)


stachenov:
Please remove unnecessary formatting changes from the commit, as it makes it harder to review (and creates unnecessary conflict risks).

momomo:
Sorry, i can't spend more time on this .. i am sure you can reindent it easily ...

momomo:
Looks easy enough here; 

image

stachenov:
Implemented manually.

momomo:
@Sergei you should consider following hte same pattern of: 

image

and adding a button here: 

image

Also you should consider naming the registry key similary, Scroll to results while typing ...

stachenov:
That's a visible UI change, I can't just do it on my own. It would require a much more complicated internal process, with the first step being actually getting the feature to some release plan.

A registry key is the most I can do as a single dev. No approval required because it's not officially supported.

If this issue gets more attention, maybe we can get a proper UI solution someday.

momomo:
Ok, yeah it works for me ... but for you it could be a good opportunity to show drive .. just contact the responsible person for UI, used to be Dmitry Jemerov <yole@jetbrains.com> but I think he quit, and just propose the addition. Its consistent with what exists now and you get to walk around with your underwear at home waiting for answers :) 

They are going to say, thats awesome Sergei! Good thinking! You are a UI expert. Come take my job! Such ambitious guy! 

It will also most likely succeed as it follows existing pattern.

PR #3577, opened 2026-07-17
Check for library nullability before importing

karimai-gg:


HackerMadCat:
Hi!

Could you please provide any reproducer for this issue? Is it possible to reproduce it with a pure Gradle Java project?

We have `org.jetbrains.plugins.gradle.importing.GradleDependenciesImportingTest#testProjectWithUnresolvedDependency` test, which tests a scenario with an unresolved dependency. For unresolved dependency, we create an empty `Library`. Therefore, it shouldn't be possible to have `LibraryOrderEntry` without `Library` in the Gradle sync.

As I can see, it is an external system issue that allows specifying `LibraryDependencyData` without corresponding `LibraryData`. Could you please create an external system test that specifies data nodes and executes `com.intellij.platform.externalSystem.testFramework.ExternalSystemTestUtilKt#importData`?

karimai-gg:
hi,
I shared the repro steps with details on the discrepencies noticed via Slack.

karimai-gg:
@nskvortsov Sergey told me to reach out to you with the updates to this PR. I have added some tests after I managed to repro the issue on a local project and verified the root cause of the issue.

PTAL and let me know you need further info.
Thanks.

nskvortsov:
The solutoin LGTM, running the CI pipleline

nskvortsov:
The PR fails test [testLocalFileDepsImportedAsModuleLibraries_nonExistentPath](https://github.com/JetBrains/intellij-community/blob/bc52ec06806ddb7aaa791813f8d840a6520ff245/plugins/gradle/testSources/org/jetbrains/plugins/gradle/importing/GradleDependenciesImportingTest.java#L569), please take a look,

karimai-gg:
Hi,
This appears to be failing from upstream already. I have a clean branch and running this test fails with "Library entity source must be from data services".
This behavior has been verified across multiple machines too. 
I am not sure if this is related this other bug [that](https://youtrack.jetbrains.com/issue/IDEA-392182) noticed. I haven't verified.

karimai-gg:
Hi,
If you still are having failures only when my patch is applied, then could you please consider sharing more details about the failures or any helpful information ?
Thank you.

nskvortsov:
This test passes successfully in our repository. With the PR it fails because of a difference in the entity source:

image

plugins/gradle/src/org/jetbrains/plugins/gradle/service/syncAction/impl/extensions/GradleDependencySyncExtension.kt
```diff
@@ -67,6 +68,20 @@ class GradleDependencySyncExtension : GradleSyncExtension {
...
+      syncStorage.resolve(moduleEntity.symbolicId)?.dependencies
+        ?.filterIsInstance<LibraryDependency>()
+        ?.filterNot { syncStorage.contains(it.library) }
+        ?.forEach {
+          val existingProjectEntity = projectStorage.resolve(it.library)
+          existingProjectEntity?.let {
+            val added = syncStorage.addEntity(it.createEntityTreeCopy())
+            syncStorage.modifyLibraryEntity(added) {
+              this.entitySource = moduleEntity.entitySource
+            }
```
nskvortsov: This entity source copy is what fails the `testLocalFileDepsImportedAsModuleLibraries_nonExistentPath` assertion. Why would you need to introduce it?

nskvortsov (changes requested):
I think I see the reason for the test failure. Please adjust

karimai-gg:
Hi Nikita
thanks a lot for the feedback and for taking the time to go through it.

Looking at the file again I realised that this patch I am trying to apply here was already merged as of https://github.com/JetBrains/intellij-community/pull/3355/.
This has landed with IDEA 262 now and can see that we also have it now, which is the part of the fix that we need to close the issue.

We can consider that this PR and the bug is now obsolete.

Thank you.

nskvortsov:
Thank you for the follow-up!

PR #3578, opened 2026-07-17
KTIJ-39564: `uastParent` resolves to `UBlockExpression` instead of `UYieldExpression` for nested when expression

francescoo22:
This PR fixes the issue described in KTIJ-39564


francescoo22:
cc. @jsjeon @BartvHelvert

PR #3579, opened 2026-07-17
[JEWEL-1239] Render failed to load images as hyperlinks

DanielSouzaBertoldi:
> [!NOTE]
> This PR supersedes #3421 :) 

# Context

Images that fail to load are now rendered as clickable hyperlinks instead of showing raw Markdown syntax or blank placeholders.

## Changes

Failed Image Handling:
  - Images that fail to load are replaced with clickable links showing the alt text (or URL if no alt text)
  - A broken image indicator (🖼) is prepended to the link text
  - Repeated images with the same URL but different alt text render independently

Loading States:
  - Added `ImageRenderResult` sealed interface with `Loading`, `Success`, and `Failed` states
  - A loading indicator is shown while an image is being fetched; when the image specifies
      width/height, the loading placeholder reserves that size so the layout doesn't shift on load
  - Source changes are debounced (~300ms) so editing a URL doesn't fire loads for transient,
      invalid intermediate values and flash the broken-link fallback
  - Switching between valid images keeps the previous one on screen (stale-while-revalidate) rather than flashing the loading indicator (it causes flickering which I do not like at alllll)

API Changes:
  - New `ImageRendererExtension.renderImage` method returning `ImageRenderResult` for explicit state handling
  - `Deprecated renderImageContent` method (preserved for backward compatibility)

## Evidence

| Scenario | Failed image | Debounce while typing | Failed -> Loading -> Success | Swapping Valid Images | Same URL, different alt text | Failed image inside of heading |
|---|---|---|---|---|---|---|
| Screen Recording | <video src="https://github.com/user-attachments/assets/4d936944-1d77-493f-aa80-e8070fb043e6" /> | <video src="https://github.com/user-attachments/assets/4b2142c4-e882-4d26-ac70-69e67577d71d" /> | <video src="https://github.com/user-attachments/assets/3e14c038-1513-4465-822e-319a0aee2b60" /> | <video src="https://github.com/user-attachments/assets/8fe5ceb7-1b27-4066-8520-d418012290ba" /> | <video src="https://github.com/user-attachments/assets/825ca7dd-a069-445f-a73f-4c56d2e2b277" /> | <video src="https://github.com/user-attachments/assets/5c5ec092-5caf-4bea-b81f-fba09c356d12" /> |

| |  Screenshot |
|--------|--------|
| Preserving style applied in a broken link | image |
| Preserving outer link destination URL | image |
| Using destination as alt if no alt is provided | image |

### ⚠️ Important Changes
 - Images that fail to load are now rendered as clickable links instead of showing raw Markdown syntax

### New features
 - In Markdown, images that fail to load are now rendered as hyperlinks

### Deprecated API
- Deprecated `ImageRendererExtension.renderImageContent`; use `renderImage` instead, which provides explicit loading/success/failed states

### DIFF
--- a/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
+++ b/platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
@@ -18,9 +18,11 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.ProvidableCompositionLocal
 import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.movableContentOf
 import androidx.compose.runtime.mutableStateMapOf
+import androidx.compose.runtime.mutableStateSetOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.runtime.withCompositionLocal
@@ -51,6 +53,7 @@ import androidx.compose.ui.platform.LocalFontFamilyResolver
 import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.LinkAnnotation
 import androidx.compose.ui.text.ParagraphIntrinsics
 import androidx.compose.ui.text.Placeholder
 import androidx.compose.ui.text.TextLayoutResult
@@ -92,6 +95,7 @@ import org.jetbrains.jewel.markdown.MarkdownBlock.ListItem
 import org.jetbrains.jewel.markdown.MarkdownBlock.Paragraph
 import org.jetbrains.jewel.markdown.MarkdownBlock.ThematicBreak
 import org.jetbrains.jewel.markdown.WithInlineMarkdown
+import org.jetbrains.jewel.markdown.extensions.ImageRenderResult
 import org.jetbrains.jewel.markdown.extensions.MarkdownRendererExtension
 import org.jetbrains.jewel.markdown.rendering.MarkdownStyling.List.Ordered.NumberFormatStyles
 import org.jetbrains.jewel.markdown.rendering.MarkdownStyling.List.Unordered.BulletCharStyles
@@ -226,16 +230,14 @@ public open class DefaultMarkdownBlockRenderer(
         softWrap: Boolean = true,
         maxLines: Int = Int.MAX_VALUE,
     ) {
-        val originalImages = renderedImages(block)
-
-        val renderedContent = rememberRenderedContent(block, inlinesStyling, enabled, onUrlClick)
+        val (loadedImages, renderedContent) = rememberRenderedContent(block, inlinesStyling, enabled, onUrlClick)
         val textColor = inlinesStyling.textStyle.color.takeOrElse { LocalContentColor.current }
         val mergedStyle = inlinesStyling.textStyle.merge(TextStyle(color = textColor))
         val density = LocalDensity.current
 
         TextWithScalableInlineContent(
             text = renderedContent,
-            inlineContent = originalImages,
+            inlineContent = loadedImages,
             density = density,
             modifier = modifier,
             overflow = overflow,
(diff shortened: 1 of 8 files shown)


platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -1036,6 +1073,64 @@ private fun getImages(input: WithInlineMarkdown): List<InlineMarkdown.Image> = b
...
+            val imageSource = image?.source.orEmpty()
+            val altText = image?.alt?.ifEmpty { null } ?: imageSource
+
+            val index =
+                if (enabled) {
+                    val link =
+                        LinkAnnotation.Clickable(
+                            tag = imageSource,
+                            linkInteractionListener = { onUrlClick?.invoke(imageSource) },
+                            styles = styling.textStyle.smartMerge(styling.textLinkStyles, enabled = true),
```
rock3r: Nit: this rebuild applies only the link and block text styles to the replacement. A failed image nested in emphasis or strong text (for example, `*alt*`) loses that enclosing inline style. Could we preserve the effective style of the replaced range? Fine to leave this edge case out of scope if intentional, or push to a follow-up if it becomes annoying

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
```diff
@@ -48,61 +61,96 @@ internal class Coil3ImageRendererExtensionImpl(private val imageLoader: ImageLoa
...
-                        .data(resolvedImageSource)
-                        // make sure image doesn't get downscaled to the placeholder size
-                        .size(Size.ORIGINAL)
-                        .build(),
-                imageLoader = imageLoader,
-                onLoading = { hasError = false },
-                onSuccess = { successState ->
+
+        // Debounce source changes as to avoid trying to fetch images needlessly while the user is not done typing.
+        var debouncedSource by remember { mutableStateOf(resolvedImageSource) }
```
rock3r, line 76: Could we key this image-renderer composition by `image.inlineContentId()`? The unkeyed remembered state follows call position, so inserting, removing, or reordering image occurrences can temporarily assign an already-loaded image or failure state to a different occurrence.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -1036,6 +1073,64 @@ private fun getImages(input: WithInlineMarkdown): List<InlineMarkdown.Image> = b
...
+            }
+            val image = imagesById[annotation.item]
+            val imageSource = image?.source.orEmpty()
+            val altText = image?.alt?.ifEmpty { null } ?: imageSource
+
+            val index =
+                if (enabled) {
+                    val link =
+                        LinkAnnotation.Clickable(
+                            tag = imageSource,
```
rock3r: Should a failed image nested in an outer link preserve that outer link’s destination? For `[alt](https://destination.example)`, this replacement creates a link to `broken-image-url`; normal HTML retains the enclosing anchor’s `https://destination.example` target when the image fails.

MDN references courtesy of Terra:

- [MDN `<a>`: missing icon within an outer link](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#external_links_and_linking_to_non-html_resources)
- [MDN ``: image link](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#image_link)
- [MDN ``: `alt`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#alt)

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
```diff
@@ -632,27 +634,62 @@ public open class DefaultMarkdownBlockRenderer(
...
-        for (image in getImages(blockInlineContent)) {
-            val renderedImage = imagesRenderer?.renderImageContent(image)
-            if (renderedImage == null) {
-                map.remove(image.source)
-            } else {
-                map[image.source] = renderedImage
+                is ImageRenderResult.Loading -> {
+                    failedSources.remove(imageId)
+                    // Show loading indicator if provided
+                    if (result.content != null) {
```
rock3r, line 677: Shouldn’t `Loading(null)` remove any existing `map[imageId]` entry? As it is, a custom renderer that refreshes an already-loaded image and returns `Loading(null)` keeps displaying the (now stale) success content, rather than the placeholder text.

rock3r (changes requested):

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
DanielSouzaBertoldi: Yeah I do understand the issue but the thing is that if we key the debouncedSource state to the image inline content id that would effectively destroy the debounce completely :(

That's because we check: `if (debouncedSource != resolvedImageSource) {` which will always be true and will get rid of the debounce delay.

If I key the inlineContentId to `displayed`, it'll re-init to null and the spinner comes back instead of holding the old image.

And the id has the source in it, so it changes on every keystroke regardless. The only thing that stays put when you edit a URL in place is the call position, which is why we lean on it, and also why it bleeds on reorder. The markdown gets reparsed from scratch every keystroke so there's no occurrence identity that survives an edit.

The only real fix I can think of is a parse-stable id per occurrence in the model, which is a lot bigger tho, so I'd rather play it safe for now

rock3r: Ok, let's go ahead with the eventual consistency then; please do track the outstanding improvement as an issue so we do not forget forever, and add a comment mentioning it

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
DanielSouzaBertoldi: [Fixed](https://github.com/JetBrains/intellij-community/pull/3579/changes#diff-54df9284972bb6066a7474093c18e590db7bbf834493e3bf26aa99246aacd27cR1129) + [added unit test](https://github.com/JetBrains/intellij-community/pull/3579/changes#diff-4e8035542c0f9a3d905612abf4002462f49f94245575548628333ae7e0a237a9R1189)

platform/jewel/markdown/extensions/images/src/main/kotlin/org/jetbrains/jewel/markdown/extensions/images/Coil3ImageRendererExtensionImpl.kt
DanielSouzaBertoldi: Added a comment explaining this limitation and linking to [JEWEL-1364](https://youtrack.jetbrains.com/projects/JEWEL/issues/JEWEL-1364/Markdown-image-render-state-bleeds-across-occurrences-on-insert-remove-reorder)

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
DanielSouzaBertoldi: [Fixed](https://github.com/JetBrains/intellij-community/pull/3579/changes#diff-54df9284972bb6066a7474093c18e590db7bbf834493e3bf26aa99246aacd27cR1120) + [added unit tests](https://github.com/JetBrains/intellij-community/pull/3579/changes#diff-4e8035542c0f9a3d905612abf4002462f49f94245575548628333ae7e0a237a9R1263)

Note: i made it so that if users don't provide an alt text, we show the `destination` as the alt text instead of the `broken-image-url`, since clicking on the link will open `destination`, not `broken-image-url`

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/rendering/DefaultMarkdownBlockRenderer.kt
DanielSouzaBertoldi: Fixed + [added unit test](https://github.com/JetBrains/intellij-community/pull/3579/changes#diff-4e8035542c0f9a3d905612abf4002462f49f94245575548628333ae7e0a237a9R1321)

(Also updated the PR description with screenshots of the three fixed scenarios)

rock3r (approved):
🚀

nebojsa-vuksic (approved):

rock3r:
@DanielSouzaBertoldi if you fix metalava we can ship (we are on 0.41 now)

DanielSouzaBertoldi:
Yeah I'll update all the metalava dumps from all open PRs I have access to to point to 0.41 :) Thanks for the heads up!

DanielSouzaBertoldi:
> Could not GET 'https://d2s4y8xcwt8bet.cloudfront.net/io/ktor/ktor-websocket-serialization-jvm/3.1.0/ktor-websocket-serialization-jvm-3.1.0.pom'. Received status code 503 from server: Service Unavailable

Come on CI stop failing 😭

DanielSouzaBertoldi:
Ready to merge

PR #3581, opened 2026-07-18
IDEA-347862 Lombok: Fix false-negative for redundant slf4j definition inspection

mibac138:
The inspection only worked on classes without a package (or if the getLogger call used the fully qualified class name) because `PsiClass.getQualifiedName` returns the whole name.

```java
initializer.getText() // LoggerFactory.getLogger(org.example.Test.class)
  .contains(format(LOGGER_INITIALIZATION, containingClass.getQualifiedName())) // LoggerFactory.getLogger(Test.class)
```

The fix swaps the simple text-based comparison to a `CallMatcher` with some further validation. Also added some tests to verify that it's working correctly.

cc @mikosik (assigned to the issue in YouTrack)


amaembo:
Thank you. The patch looks good. I will merge it.

amaembo:
Merged internally: cf3558fedac9e32b4947f0c4a95e4a9b09f9ac8b
Unfortunately, I messed with commit settings, so your name doesn't appear in the commit metadata. Sorry for that.

PR #3582, opened 2026-07-21
[JEWEL-1367] Replace managing-youtrack snippets with a youtrack-cli CLI

rock3r:
The `managing-youtrack` skill was ~625 lines of copy-paste `curl` snippets. Snippets cannot be tested, cannot fail loudly, and drift silently from the API — and this set had drifted into three defects, each confirmed against the live instance.

## Context

* The documented commands endpoint, `POST /api/issues/<ID>/commands`, does not exist. It returns `404 No subresource for path commands`, so every state change, assignee change and issue link the skill described via commands would have failed. Commands are global: `POST /api/commands`, with the target issues in the body.
* The worked examples were built on `JEWEL-1`, which has been deleted and returns 404 — as was the matching `<PROJECT>-1` trick the skill recommended for discovering a project's internal id.
* Every write depended on the reader remembering a shell-escaping convention ("write JSON to a temp file, pass it with `-d @file`, never interpolate") rather than on anything actually enforcing it.

## Changes

* Adds `scripts/yt.py`, a stdlib-only CLI — no venv, no lockfile, no install step. It builds request bodies in Python, which removes shell escaping as a category rather than mitigating it, and covers the previous skill's full surface plus attachments (list, upload, download, delete), which it did not document at all.
* Pins the host in code and refuses any cross-host redirect, so the skill's URL-pinning rule is enforced rather than merely documented.
* Maps errors onto exit codes (3 auth, 4 not found, 5 validation, 6 transient) so callers can branch without parsing prose, and retries 429/5xx with jittered backoff.
* Resolves the token from `--token-op-path`, `$YOUTRACK_TOKEN`, or `$YOUTRACK_TOKEN_OP_PATH`, the first and last reading the secret through the 1Password CLI. Only the `op://` path ever reaches argv; the token is held in memory, scrubbed from error output, and never logged, printed, or written to disk.
* Handles the 1Password CLI's failure modes explicitly, since both exit 1 and only the message separates them: an unanswered prompt hits op's own ~60s deadline with `authorization timeout`, while a declined one returns `authorization prompt dismissed` — and the latter tells the agent to stop and ask rather than retry. Anything else points at the agent sandbox blocking the 1Password CLI from reaching the desktop app, with instructions to retry unsandboxed before involving the user.
* Restructures the skill along the `skill-creator` guidance: `SKILL.md` drops from 625 to 176 lines and keeps only what an agent would otherwise get wrong, the full command surface moves to `references/cli-reference.md`, the `curl` escape hatch to `references/raw-api.md`, and `evals/evals.json` adds 10 behavioural evals covering happy paths, trigger near-misses and pressure cases. Every mutating eval is plan-only.
* Adds `scripts/test_yt.py`: 52 unit tests that need no network, plus an opt-in read-only `--live` smoke check. Beyond those, the CLI was verified against the live instance — all three token sources, exit codes for missing and invalid tokens, read paths matching the raw `curl` output they replace, `command apply --dry-run` leaving JEWEL-1367 and JEWEL-525 untouched, and an attachment download byte-exact at 288108 bytes.


rock3r:
Review note: the CLI and its tests have been reviewed by three models already, it is probably not worth spending time reading them again (I tried but my python is very bad). Focus on the skill/docs :)

.claude/skills/managing-youtrack/references/cli-reference.md
```diff
@@ -0,0 +1,226 @@
...
+python3 $YT user search jane --top 10 --format table
+
+python3 $YT project get <PROJECT>                    # -> {"shortName":"...","id":"<INTERNAL-ID>",...}
+python3 $YT project fields <PROJECT> --format table   # required flags and allowed types
+
+python3 $YT saved-queries --top 50
+```
+
+`project get` queries `/api/admin/projects`. Never derive a project id by fetching
+`<PROJECT>-1` — that issue frequently does not exist.
```
DanielSouzaBertoldi: "frequently"? I thought it didn't exist at all 🤯

.claude/skills/managing-youtrack/SKILL.md
```diff
@@ -1,624 +1,246 @@
...
+Paths below are relative to this skill's own directory. Point `YT` at the CLI, adjusting the
+prefix if you are running from elsewhere:
 
 ```bash
-curl -s -L --max-redirs 3 \
-  -H @- \
-  -H "Accept: application/json" \
-  "https://youtrack.jetbrains.com/api/issues/<ISSUE-ID>/customFields?fields=id,name,value(name)" \
-  <<< "Authorization: Bearer ${YOUTRACK_TOKEN}"
+YT=scripts/yt.py
```
DanielSouzaBertoldi: Isn't this a bit brittle? Line 25 states "relative to this skill's own directory", but won't agents invoking this skill almost always have cwd = repo root? the script file itself is in `.claude/skills/managing-youtrack`, we usually run claude in `intellij-community`, so maybe we should make the script path more explicit by telling the agent to resolve skill's absolute path first rather than a relative path that's only true when inside this .claude folder 🤔

DanielSouzaBertoldi (approved):
Awesome work!

With this, we could even start to think about ways to use the CLI to automatically set YT issues as done/fixed along with the jewel version once a PR gets merged.

The problem is figuring out all the little details in between tho 😛 how to precisely tell what version is the fix for, how to check if a closed PR was actually merged or not (checking if the commit exists in the repo is enough i guess) and when to check (a cron job? everytime a pr gets closed?)

But I digress. No blockers, just one minor suggestion to make this as bullet-proof as possible

.claude/skills/youtrack-cli/references/cli-reference.md
```diff
@@ -0,0 +1,229 @@
...
+
+All examples assume `YT` points at the CLI and a token source is set. Use the **absolute** path to
+this skill's `scripts/yt.py` so it works from any working directory — replace `<SKILL-DIR>` with
+this skill's own directory (or a harness-provided variable such as `$CLAUDE_SKILL_DIR` under Claude
+Code):
+
+```bash
+YT="<SKILL-DIR>/scripts/yt.py"
+export YOUTRACK_TOKEN_OP_PATH="op://VAULT/ITEM/FIELD"
+```
```
DanielSouzaBertoldi, line 14: Nice! Thanks for updating this ♥️

DanielSouzaBertoldi (approved):

.claude/skills/youtrack-cli/references/cli-reference.md
rock3r: Thanks to you for spotting it wasn't great. I think my melting brain just missed it when I checked this file :(

.claude/skills/managing-youtrack/references/cli-reference.md
rock3r: That may have been overfitting :) But we did have JEWEL-1 back when we were on GitHub, not sure why it's gone in YouTrack :P

rock3r: Oh maybe it was a PR :)

rock3r:
@nebojsa-vuksic lmk if you want to hold this for review or we can ship

.claude/skills/youtrack-cli/references/cli-reference.md
DanielSouzaBertoldi: It's all good my friend :) Glad we were able to make the skill even more resilient now 💪🏻

rock3r:
Ready to merge

PR #3583, opened 2026-07-21
IDEA-350451 IDEA-332430 Lombok: Add inspection to detect unused logger annotation

mibac138:
Checks all logger annotations supported by Lombok. Includes tests.


plugins/lombok/src/main/resources/intellij.lombok.xml
```diff
@@ -191,6 +191,16 @@
       implementationClass="de.plushnikov.intellij.plugin.inspection.RedundantSlf4jDefinitionInspection"
       key="inspection.redundant.slf.4.j.definition.display.name" bundle="messages.LombokBundle"/>
 
+    <localInspection
+      language="JAVA"
+      groupPath="Java,Lombok"
+      groupKey="inspection.lombok.group.name.redundant.definitions"
+      groupBundle="messages.LombokBundle"
+      shortName="UnusedLogger"
```
amaembo: Please use more comprehensive 'shortName' like "UnusedLoggerLombok", because short names have global namespace (in inspection profile configuration and in suppressions)

plugins/lombok/src/main/resources/messages/LombokBundle.properties
```diff
@@ -10,6 +10,7 @@ dialog.title.attention=Attention!
 dialog.message.this.element.cannot.be.renamed=This element cannot be renamed.
 inspection.message.lombok.annotation.deprecated.not.supported=Lombok''s annotation ''{0}'' is deprecated and not supported by Lombok plugin anymore. Use ''{1}'' instead.
 inspection.message.slf4j.logger.defined.explicitly=Slf4j Logger is defined explicitly. Use Lombok @Slf4j annotation instead.
+inspection.message.logger.not.used=Logger generated by @{0} is not used
```
amaembo: Let's use single quotes around the annotation, like `Logger generated by ''@{0}'' is not used`

plugins/lombok/src/main/resources/inspectionDescriptions/UnusedLogger.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+Reports Lombok logger annotations whose generated logger field is not used.
```
amaembo: Please add a code example where the warning is issued. See, e.g. java/java-impl/resources/inspectionDescriptions/ReassignedVariable.html for formatting example. 

Also
```html
<!-- tooltip end -->
<p><small>New in 2026.3</small></p>
```
Would be nice

plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/inspection/UnusedLoggerInspection.java
```diff
@@ -0,0 +1,62 @@
...
+      if (!annotation.hasQualifiedName(name)) return;
+
+      var psiClass = PsiTreeUtil.getParentOfType(annotation, PsiClass.class, false);
+      if (psiClass == null) return;
+
+      var logger = psiClass.findFieldByName(AbstractLogProcessor.getLoggerName(psiClass), false);
+      if (logger == null || logger.getNavigationElement() != annotation || !logger.hasModifierProperty(PsiModifier.PRIVATE)) return;
+      if (ReferencesSearch.search(logger, new LocalSearchScope(psiClass)).findFirst() != null) return;
+
+      var loggerName = name.substring(name.lastIndexOf('.') + 1);
```
amaembo: com.intellij.openapi.util.text.StringUtil#getShortName(java.lang.String)?

plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/inspection/UnusedLoggerInspection.java
```diff
@@ -0,0 +1,62 @@
...
+      checkFor(annotation, LombokClassNames.COMMONS_LOG);
+      checkFor(annotation, LombokClassNames.CUSTOM_LOG);
+      checkFor(annotation, LombokClassNames.JBOSS_LOG);
+      checkFor(annotation, LombokClassNames.FLOGGER);
+      checkFor(annotation, LombokClassNames.SLF_4_J);
+      checkFor(annotation, LombokClassNames.XSLF_4_J);
+    }
+
+    private void checkFor(@NotNull PsiAnnotation annotation, String name) {
+      if (!annotation.hasQualifiedName(name)) return;
```
amaembo: I would create a hashset of all the annotation names and use `String name = annotation.getQualifiedName(); if (!set.contains(name)) return;` here, to avoid calling `checkFor` many times. WDYT?

plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/inspection/UnusedLoggerInspection.java
```diff
@@ -0,0 +1,62 @@
...
+
+      var psiClass = PsiTreeUtil.getParentOfType(annotation, PsiClass.class, false);
```
amaembo: I think it would be more robust to use `annotation.getAnnotationOwner()` (which should be a `PsiModifierList`), and then `getParent()` and check if it's the class. Otherwise, we may end up doing something strange (e.g., if annotation is mistakenly applied to a method; compilation errors don't prevent inspections from running).

plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/inspection/UnusedLoggerInspection.java
```diff
@@ -0,0 +1,62 @@
...
+      if (psiClass == null) return;
+
+      var logger = psiClass.findFieldByName(AbstractLogProcessor.getLoggerName(psiClass), false);
+      if (logger == null || logger.getNavigationElement() != annotation || !logger.hasModifierProperty(PsiModifier.PRIVATE)) return;
+      if (ReferencesSearch.search(logger, new LocalSearchScope(psiClass)).findFirst() != null) return;
+
+      var loggerName = name.substring(name.lastIndexOf('.') + 1);
+      holder.problem(annotation, LombokBundle.message("inspection.message.logger.not.used", loggerName))
+        .highlight(ProblemHighlightType.LIKE_UNUSED_SYMBOL)
+        .fix(new RemoveAnnotationQuickFix(annotation, psiClass))
```
amaembo: I think com.intellij.codeInsight.daemon.impl.quickfix.DeleteElementFix is better here (but please check). RemoveAnnotationQuickFix has a lot of logic which is not necessary in this inspection.

amaembo (changes requested):
Thank you for your pull request. Please address the comments.

plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/inspection/UnusedLoggerInspection.java
mibac138: Good point, I've adjusted the code to this approach

plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/inspection/UnusedLoggerInspection.java
mibac138: DeleteElementFix works fine, but contrary to RemoveAnnotationQuickFix, it doesn't remove unused imports. I've left the fix as it is and added a test to verify that the unused import is removed. Let me know if you'd like to use the simpler quickfix anyways

plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/inspection/UnusedLoggerInspection.java
mibac138: Yeah, that's a cleaner approach. I've switched to that :)

amaembo:
Pushed, thank you! 04d7057e61d0a8478d7901e4b9be8818aaa948f9

PR #3585, opened 2026-07-22
[JEWEL-1369] Fix stroked icons rendering incorrectly across frontends

rock3r:
> [!IMPORTANT]
> **This is about the new cross-frontend icons API** — `com.intellij.platform.icons`, in `platform/icons-api` / `platform/icons-impl` — the one designed to render the same icon through multiple frontends (Swing, Compose) and the one Jewel's Compose standalone frontend draws through. **The legacy icon path (`IconLoader`, `SVGLoader` and its colour patchers) is not touched by this PR** and none of the defects below apply to it.

Stroking an icon — the palette substitution that turns an icon into a monochrome outline in a given color — rendered differently depending on which frontend drew it. The Swing frontend and Jewel's Compose standalone frontend each carried their own copy of the same substitution, and each copy was wrong in a different way, so the same icon could come out blanked in a Swing toolbar, gray in a Compose one, and correct in neither. That is precisely the failure the new API exists to prevent: one icon, one appearance, whichever frontend renders it.

Fixes [JEWEL-1369](https://youtrack.jetbrains.com/issue/JEWEL-1369) / IJPL-176416.

Six distinct defects were behind that:

* **The Swing frontend blanked every fill.** It replaced `fill` unconditionally with `transparent` and added a `stroke` attribute. An icon drawn from filled shapes — which is most of them — became invisible rather than stroked.
* **The stroke patcher was silently dropped.** The patchers were combined outwards from `svgPatcher`, which is `null` whenever an icon carries no explicit patcher (the common case), so the null-safe call discarded the stroke patcher entirely — while `isStroke` still forced the stroke variant to load, leaving the icon in its authored gray.
* **Only `fill` was patched, never `stroke`, in both frontends.** New UI icons come in two authoring styles, `<path fill="#6C707E">` and `<path fill="none" stroke="#6C707E">`; patching one attribute leaves every icon of the other style untouched.
* **The palette was incomplete.** It was missing two colours the legacy Swing stroke patcher (`com.intellij.ui.icons.stroke`) has — `#A8ADBD` and `#CED0D6` — plus the keyword spellings `black` and `white`, which an SVG may use in place of `#000000` / `#FFFFFF`. Icons painted in any of those were left un-stroked.
* **Stroking resolved dark artwork in a dark theme.** A `_dark` variant is authored in its own grays (`inlaySettings_dark.svg` is `#9DA0A8`), which are not stroke palette colours, so a stroked icon in a dark theme kept them. The legacy path avoids this by loading the light variant; both frontends now do the same.
* **The Swing frontend compared color values case-sensitively.** SVGs are authored uppercase (`#6C707E`) while the palette is built from `Color.toHex()`, which is lowercase. Every palette color containing a hex letter silently failed to match and was left un-stroked; colors that happen to be all digits (`#818594`) worked by accident. SVG hex colors are case-insensitive per spec, and the Compose frontend already compared that way.

## Changes

* Adds `strokeSvgPatcher(stroke)` in `platform/icons-impl` as the single definition of the stroke substitution. Both frontends now call it, which is what stops them drifting apart again. It is the only new public declaration — the design palettes it is built from stay private.
* The substitution is *palette-only*, never blanket: background tints become `transparent`, foreground tints become the stroke color, and anything painted outside those two sets is deliberately left alone — so filled artwork survives instead of being erased.
* The substitution applies to both the `fill` and the `stroke` attribute, covering both New UI authoring styles. The foreground palette also gains the two colours and the two keyword spellings it was missing relative to the legacy patcher, so the two agree on what counts as a palette colour.
* Condition evaluation moved onto `SvgPatchOperation.matches`, so both frontends resolve an operation identically rather than each deciding for itself. It folds case only for a plain colour literal on a colour-valued attribute — a `url(#Id)` paint reference and an `id` are case-sensitive and must not be conflated.
* The combine starts from the icon's own `svgPatcher` and falls back to the stroke patcher. That fixes the dropped-patcher bug while keeping the icon's own patcher running *first*, so an icon that deliberately recolours a palette colour keeps it.
* Jewel’s `ComposeImageResourceProvider` (the Compose frontend) drops its duplicate palette and its `fill -> transparent` blanket replacement, and delegates to the shared patcher.
* Adds no public platform API: `strokeSvgPatcher` and `SvgPatchOperation.matches` are both `@ApiStatus.Internal`, and `api-dump.txt` is unchanged.
* `StrokeRenderTest` pins the behavior at the pixel level through the *real* Swing path — `imageIcon(...).toSwingIcon().paintIcon(...)` — rasterizing real New UI icons and counting red vs. authored-gray pixels: `softWrap` (fill-authored), `save` and `inlaySettings` (stroke-authored) all recolor fully, with none of the authored gray left behind. `StrokePatcherTest` pins the patch shape itself, spelling out the colours it expects rather than reading them back from the patcher's own palette, so dropping an entry fails the test instead of quietly shrinking what it checks. `SvgPatchOperationMatchesTest` pins the comparison rule directly. The render tests also cover a dark theme and patcher ordering; each was mutation-checked to fail when the behaviour it describes is reverted.

## Screenshots

Real New UI icons rendered through the Swing frontend — `imageIcon(...).toSwingIcon().paintIcon(...)` — asked to stroke in `#3574F0`, at master vs. this branch. Before the fix, stroking is a no-op: the "before" row is pixel-identical to the unstroked reference.

Icon stroking before and after

### Bug fixes

* Fixed icon stroking rendering incorrectly: icons drawn from filled shapes were blanked out entirely, icons authored with `stroke` rather than `fill` were left in their authored gray, and stroking was silently skipped for icons that carry no explicit SVG patcher. Stroking now recolors both the `fill` and the `stroke` attribute, matches palette colors case-insensitively (including keyword spellings such as `black`), and renders identically in the Swing and Compose frontends.

### DIFF
--- a/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/icon/PathImageResourceLocation.kt
+++ b/platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/icon/PathImageResourceLocation.kt
@@ -38,7 +38,10 @@ public class PathImageResourceLocation(public val path: String, public val class
             append(path.substringBeforeLast('/', ""))
             append('/')
             append(path.substringBeforeLast('.').substringAfterLast('/'))
-            if (modifiers.isDark) {
+            // A stroked icon loads its light artwork even in a dark theme, matching the Swing frontend: the stroke
+            // palette describes the light variants, so a `_dark` variant would come back with colors it does not know
+            // and would be left in its authored color instead of being recolored.
+            if (modifiers.isDark && modifiers.stroke == null) {
                 append("_dark")
             }
             append('.')
--- a/platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/Convertors.kt
+++ b/platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/Convertors.kt
@@ -55,8 +55,7 @@ private class ProxySvgAttributePatcher(
         }
         SvgPatchOperation.Operation.Replace -> {
           if (operation.conditional) {
-            val matches = attributes[operation.attributeName] == operation.expectedValue
-            if (matches == !operation.negatedCondition) {
+            if (operation.matches(attributes[operation.attributeName]) != operation.negatedCondition) {
               attributes.replace(operation.attributeName, operation.value!!)
             }
           } else {
@@ -65,8 +64,7 @@ private class ProxySvgAttributePatcher(
         }
         SvgPatchOperation.Operation.Remove -> {
           if (operation.conditional) {
-            val matches = attributes[operation.attributeName] == operation.expectedValue
-            if (matches == !operation.negatedCondition) {
+            if (operation.matches(attributes[operation.attributeName]) != operation.negatedCondition) {
               attributes.remove(operation.attributeName)
             }
           } else {
--- a/platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/images/SwingImageResourceHolder.kt
+++ b/platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/images/SwingImageResourceHolder.kt
@@ -7,8 +7,8 @@ import com.intellij.ui.scale.ScaleContext
 import com.intellij.platform.icons.impl.intellij.rendering.toAwtFilter
 import com.intellij.platform.icons.impl.intellij.rendering.toIJPatcher
 import com.intellij.platform.icons.impl.patchers.DefaultSvgPatcher
+import com.intellij.platform.icons.impl.patchers.strokeSvgPatcher
 import com.intellij.platform.icons.impl.rendering.DefaultImageModifiers
-import com.intellij.platform.icons.patchers.svgPatcher
 import com.intellij.platform.icons.rendering.Dimensions
 import com.intellij.platform.icons.rendering.ImageModifiers
 import java.awt.Image
@@ -26,18 +26,22 @@ internal fun ImageModifiers?.toLoadParameters(): LoadIconParameters {
     filters.add(colorFilter.toAwtFilter())
   }
   val knownModifiers = this as? DefaultImageModifiers
-  val strokePatcher = knownModifiers?.stroke?.let { stroke ->
-    svgPatcher {
-      replace("fill", "transparent")
-      add("stroke", stroke.toHex())
-    }
-  }
+  val strokePatcher = knownModifiers?.stroke?.let { strokeSvgPatcher(it) }
   val ijModifiers = this as? IntelliJImageModifiers
-  val colorPatcher = ((this?.svgPatcher?.combineWith(strokePatcher)) as? DefaultSvgPatcher)?.toIJPatcher(ijModifiers?.legacyPatcherProvider)
+  // The icon's own patcher runs first and the stroke substitution after it, so an icon that explicitly recolors a
+  // palette color keeps that color: the stroke operation no longer matches what the explicit one already replaced.
+  // The elvis is what keeps a stroke-only icon patched at all — `svgPatcher` is null whenever an icon carries no
+  // explicit patcher, and combining outwards from null would discard the stroke patcher entirely.
+  val combinedPatcher = this?.svgPatcher?.combineWith(strokePatcher) ?: strokePatcher
+  val colorPatcher = (combinedPatcher as? DefaultSvgPatcher)?.toIJPatcher(ijModifiers?.legacyPatcherProvider)
+  val isStroke = knownModifiers?.stroke != null
   return LoadIconParameters(
     filters = filters,
-    isDark = knownModifiers?.isDark ?: false,
+    // A stroked icon loads its light artwork even in a dark theme. The palette describes the light variants, so
+    // resolving `_dark` first would hand the patcher colors it does not know and the icon would keep its authored
+    // ones — and the artwork is about to be recolored wholesale anyway, so which variant it started from is moot.
+    isDark = !isStroke && knownModifiers?.isDark == true,
     colorPatcher = colorPatcher,
-    isStroke = knownModifiers?.stroke != null
+    isStroke = isStroke
   )
 }
--- a/platform/icons-impl/intellij/tests/test/com/intellij/platform/icons/impl/intellij/StrokePatcherTest.kt
+++ b/platform/icons-impl/intellij/tests/test/com/intellij/platform/icons/impl/intellij/StrokePatcherTest.kt
@@ -1,0 +1,117 @@
+// Copyright 2000-2026 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.platform.icons.impl.intellij
+
+import com.intellij.platform.icons.impl.design.DefaultSRGB
+import com.intellij.platform.icons.impl.patchers.DefaultSvgPatcher
+import com.intellij.platform.icons.impl.patchers.SvgPatchOperation
+import com.intellij.platform.icons.impl.patchers.strokeSvgPatcher
+import com.intellij.testFramework.junit5.TestApplication
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertNotNull
+
+/**
+ * Pins the shape of the stroke patch itself: which colors it covers, which attributes it touches, and how selectively.
+ *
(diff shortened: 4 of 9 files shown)


platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/icon/PathImageResourceLocation.kt
```diff
@@ -38,7 +38,10 @@ public class PathImageResourceLocation(public val path: String, public val class
             append(path.substringBeforeLast('/', ""))
             append('/')
             append(path.substringBeforeLast('.').substringAfterLast('/'))
-            if (modifiers.isDark) {
+            // A stroked icon loads its light artwork even in a dark theme, matching the Swing frontend: the stroke
+            // palette describes the light variants, so a `_dark` variant would come back with colors it does not know
+            // and would be left in its authored color instead of being recolored.
+            if (modifiers.isDark && modifiers.stroke == null) {
```
DanielSouzaBertoldi, line 44: Alright, since I've no familiarity whatsoever with how icons are rendered in both swing and compose, I asked Claude to guide me through the changes while bombarding it with questions (starting by the fact that "stroke" here can mean two different things: the svg attribute or the "recoloring" that the patcher does on top of the svg colors).

Then, after dissecting the concepts and the code, we noticed something that might be outside the scope of this PR but I wanted to raise the question so that we can be aware of this.

Looking at the `com.intellij.ui.icons.RasterizedImageDataLoader` class (the loader that `DataLoaderImageResourceHolder`, one of the `SwingImageResourceHolder` implementations, delegates to), there's specific code that checks whether an icon `HAS_STROKE`:

```kt
if (parameters.isStroke && (imageFlags and ImageDescriptor.HAS_STROKE) == ImageDescriptor.HAS_STROKE) {
  effectivePath = "${name}_stroke.$ext"
  nonSvgScale = 1f
}
```

Reading the code, `parameters.isStroke` means the caller asked to recolor the stroke for this icon and `HAS_STROKE` is a build-time flag that `ImageCollector` sets when the icon ships a dedicated `_stroke.svg` variant on disk. So, the legacy image loader swaps the path of the icon if we're recoloring and a pre-drawn stroke variant exists.

Using the `icons/expui/run/run.svg` example, there *is* a variation of this icon with `_stroke`:

| run.svg | run_stroke.svg |
|--------|--------|
| image | image |

But in the Compose renderer class (`org.jetbrains.jewel.ui.icon.ComposeImageResourceProvider`) we do not have the same check. The `run.svg` has these two attributes: `fill="#F2FCF3" stroke="#208A3C"`. `F2FCF3` (used as the `fill` color) *is* in the `strokeBackgroundPalette` list, so it's recolored as transparent, awesome! But the green stroke, #208A3C is not part of the expected palette and we don't check if there's a `_stroke` variant of an icon (we only check for `_dark`), so we ignore the green color and final result is a transparent background color with a green "outline".

Here's the difference in rendering the run icon in both standalone and ide:

| Compose | Swing |
|--------|--------|
| image | image |

Here's the git patch of this test:

<details><summary>Really cool git patch</summary>
<p>

```
Index: platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Icons.kt
IDEA additional info:
<+>UTF-8
diff --git a/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Icons.kt b/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Icons.kt
--- a/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Icons.kt	(revision 350b96dcb881c15c718a9d3a510d67c91dae1205)
+++ b/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/Icons.kt	(date 1784897699809)
@@ -161,6 +161,75 @@
             }
         }
 
+        Column {
+            // JEWEL-1369 review: stroke parity across the two Compose stroke pipelines, on a HAS_STROKE icon.
+            //
+            // Actions.Execute resolves to expui/run/run.svg in New UI. Its ink is #208A3C — a semantic green that is
+            // deliberately NOT in the stroke palette — and it ships a dedicated white run_stroke.svg variant.
+            //
+            //  * System A (the `Stroke` painter hint) is a `_stroke` suffix hint: it loads the white run_stroke.svg
+            //    variant, so it strokes cleanly.
+            //  * System B (the new `IconModifier.stroke` API) only knows the `_dark` suffix (PathImageResourceLocation),
+            //    so it never reaches run_stroke.svg. It palette-synthesizes from the green base instead — and #208A3C
+            //    is not a palette color, so the green survives un-recolored.
+            //
+            // Both cells ask for the SAME white stroke on the SAME icon and should be identical white. They aren't:
+            // the hint is white, the modifier stays green. (The Swing frontend matches the hint, via run_stroke.svg.)
+            Text("Stroke parity — Actions.Execute (HAS_STROKE, #208A3C base), asked to stroke white on blue:")
+
+            val strokeParityBackground =
+                if (JewelTheme.isDark) {
+                    JewelTheme.colorPalette.blueOrNull(4) ?: Color(0xFF375FAD)
+                } else {
+                    JewelTheme.colorPalette.blueOrNull(4) ?: Color(0xFF3574F0)
+                }
+
+            Row(
+                modifier = Modifier.fillMaxWidth().padding(horizontal = 16.dp),
+                horizontalArrangement = Arrangement.spacedBy(24.dp),
+            ) {
+                Column(
+                    horizontalAlignment = Alignment.CenterHorizontally,
+                    verticalArrangement = Arrangement.spacedBy(4.dp),
+                ) {
+                    Box(Modifier.size(24.dp), contentAlignment = Alignment.Center) {
+                        Icon(contentDescription = "Execute base (unstroked)") { iconKey(AllIconsKeys.Actions.Execute) }
+                    }
+                    Text("base (green)")
+                }
+                Column(
+                    horizontalAlignment = Alignment.CenterHorizontally,
+                    verticalArrangement = Arrangement.spacedBy(4.dp),
+                ) {
+                    Box(
+                        Modifier.size(24.dp).background(strokeParityBackground, shape = RoundedCornerShape(4.dp)),
+                        contentAlignment = Alignment.Center,
+                    ) {
+                        Icon(
+                            key = AllIconsKeys.Actions.Execute,
+                            contentDescription = "Execute via Stroke hint (System A)",
+                            hint = Stroke(Color.White),
+                        )
+                    }
+                    Text("hint — white ✓")
+                }
+                Column(
+                    horizontalAlignment = Alignment.CenterHorizontally,
+                    verticalArrangement = Arrangement.spacedBy(4.dp),
+                ) {
+                    Box(
+                        Modifier.size(24.dp).background(strokeParityBackground, shape = RoundedCornerShape(4.dp)),
+                        contentAlignment = Alignment.Center,
+                    ) {
+                        Icon(contentDescription = "Execute via IconModifier.stroke (System B)") {
+                            iconKey(AllIconsKeys.Actions.Execute, modifier = IconModifier.stroke(Color.White))
+                        }
+                    }
+                    Text("modifier — stays green ✗")
+                }
+            }
+        }
+
         Column {
             var checked by remember { mutableStateOf(true) }
 

```



</p>
</details>

So, all of this just to bring up two main points:

1. ~~Given that the idea of this fix is to render icons identically in both Swing and Compose frontends, maybe we could plug this fix in this PR too?!~~
2. The Compose icon render code now has two pipelines: using hints or icon modifiers. Which one of them are we going to actually urge users to use? Or do they complement each other?

DanielSouzaBertoldi: Disregard, I just saw your screenshots in #3588 that shows exactly this scenario being fixed as well as some others

DanielSouzaBertoldi (approved):
Checked all the major issues fixed here, nothing to add :) The only two points I had have already been addressed in #3588 so it's all fine

It was cool to learn more about how icons are rendered in IJP/Jewel tho.

but... with this code if an icon is drawn with just the background colors in the `strokeBackgroundPalette` list the icon will render nothing, since it'll be 100% transparent. But I guess that icon designers are aware of this so it's not really a risk after all.

Good stuff 🚀

rock3r:
> but... with this code if an icon is drawn with just the background colors in the strokeBackgroundPalette list the icon will render nothing, since it'll be 100% transparent. But I guess that icon designers are aware of this so it's not really a risk after all.

Yeah this is how the IJPL works so 🤷

jakub-senohrabek (approved):

rock3r:
Ready to merge

nebojsa-vuksic (approved):

PR #3588, opened 2026-07-23
[JEWEL-1370] Align stroking on the new icons API with the legacy path

rock3r:
> [!IMPORTANT]
> ~~**This PR is stacked on #3585 and requires it to be merged first.** It branches off `ijpl-176416-icons-stroke` and builds on the shared stroke patcher that PR introduces.~~ Merged!

> [!IMPORTANT]
> **This is about the new cross-frontend icons API** — `com.intellij.platform.icons`, in `platform/icons-api` / `platform/icons-impl` — the one designed to render the same icon through multiple frontends (Swing, Compose) and the one Jewel's Compose standalone frontend draws through. **The legacy icon path (`IconLoader`, `SVGLoader` and its colour patchers) is not touched by this PR.** It is the reference implementation the new API is being aligned *to*, and none of the defects below apply to it.

Follow-up to #3585. Comparing the two implementations side by side surfaced three further behaviours where the new API still does not match the legacy one — long-standing gaps, not regressions from #3585.

Fixes [JEWEL-1370](https://youtrack.jetbrains.com/issue/JEWEL-1370).

* **Hand-authored `_stroke.svg` variants took the wrong patch.** Such a variant is a separate drawing of the same glyph, already reduced to an outline by hand and authored in white. The legacy path only recolours it; the new API applied the full palette substitution, background-drop included. Reducing an already-reduced drawing can only take artwork away.
* **The Compose frontend never looked for that variant.** A stroked `expui/run/run.svg` rendered the base file — a background-tint fill inside an off-palette green outline. The fill dropped out, the green survived, and the requested stroke colour appeared nowhere.
* **Opacity and alpha were ignored.** `fill` and `stroke` carry no alpha channel — alpha belongs on `fill-opacity` / `stroke-opacity` — so a translucent stroke colour was written out as `#RRGGBBAA`, and an opacity the document was authored with went on shading a colour that had just been replaced.
* **Short hex forms never matched.** Values were compared literally, so `fill="#fff"` never matched the palette's `#ffffff`.

## Changes

* Adds `authoredStrokeSvgPatcher(stroke)` beside `strokeSvgPatcher`: a white-to-target recolour with no palette reduction, which is what the legacy path applies to an authored variant. Each frontend picks between the two from the file it resolved. Stroking such a variant in opaque white patches nothing at all, matching the legacy short-circuit, so a variant carrying its own opacity is left exactly as authored.
* Jewel's `PathImageResourceLocation` prefers the `_stroke` variant when stroking, falling back to the base file for icons that ship none, the way the Swing frontend already did.
* A colour written into a colour attribute is split across it and its paired `-opacity`, which is set or removed from the new colour's alpha — what the legacy `newSvgPatcher` does via its `alphaProvider`. It covers `stop-color` and `flood-color` alongside `fill` and `stroke`, since those pair with an opacity attribute too.
* Colour conditions compare canonically, expanding the three- and four-digit hex shorthands. The comparison is on the colour alone — an alpha the literal carries is not part of its identity, since the substitution replaces the opacity of what it matched along with the colour. A literal of a length that names no colour, such as five digits, keeps comparing as text rather than being expanded into a colour it does not name (legacy `toCanonicalColor` throws on those; see IJPL-250996).
* Both fixes above land in shared code rather than in either frontend: colour comparison extends `SvgPatchOperation.matches`, which #3585 already made the one place a condition is resolved, and colour writing gets the matching `writeSvgAttribute` that both frontends now write through. A frontend that split the pair differently would draw the same icon differently, which is the same argument that put `matches` there.
* `authoredStrokeSvgPatcher`, `AUTHORED_STROKE_VARIANT_SUFFIX` and `writeSvgAttribute` are the new declarations, all `@ApiStatus.Internal` like `strokeSvgPatcher`, so `api-dump.txt` is unchanged. No Jewel public API change.
* Tests: `SvgAttributeWriterTest` and `PathImageResourceLocationTest` are new, pinning how a colour is written and which file a stroked icon resolves — including that it takes light artwork in a dark theme, composing with the rule #3585 added. `SvgPatchOperationMatchesTest` gains the shorthand, alpha and invalid-length cases; `StrokePatcherTest` covers the authored-variant patch, keeping its property that expected colours are spelled out rather than read back from the patcher's own palette. `StrokeRenderTest` adds two cases through the real Swing path: no pixel of a stroked `cursorText.svg` still comes out at the `0.7` opacity it was authored with (18 did before), and stroking `run.svg` leaves none of its off-palette green behind.

## Screenshots

Each row names the defect and says what the `before` column is showing. Rendered through the real Swing path at the parent commit and at this one, except the `run.svg` row, which rasterizes the two documents the Compose frontend actually produces for it.

Before/after for the three aligned behaviours, plus an unchanged control

Two further changes cannot be shown this way, because no shipped icon reaches them and one of them is invisible to this rasterizer. Both are pinned by tests:

| authored | stroke colour | before | after |
|---|---|---|---|
| `fill="#6C707E"` | `#FF333380` | `fill="#ff333380"` — an alpha channel in an attribute that has none | `fill="#ff3333"` `fill-opacity="0.5019608"` |
| `fill="#6C707E80"` | `#FF3333` | unchanged — the palette never matched it, so the shape kept its authored grey | `fill="#ff3333"` |

JSVG happens to tolerate the eight-digit form in the first row, so it renders the same either way today; Skia's behaviour on the Compose side is untested, which is the point of not relying on it.

### ⚠️ Important Changes

* Writing a colour into a colour attribute from an `SvgPatcher` now also normalises the paired opacity attribute — `fill-opacity` for `fill`, and likewise for `stroke`, `stop-color` and `flood-color`: the written colour's alpha moves onto it, and it is removed when that colour is opaque. Custom patchers that relied on the document's authored opacity surviving will see it replaced.

### Bug fixes

* Stroked icons rendered through the new cross-frontend icons API now behave like the legacy `IconLoader` path: icons shipping a hand-authored `_stroke.svg` variant are resolved and recoloured in the Compose frontend too, a shape's authored `fill-opacity` / `stroke-opacity` no longer shades the stroke colour, a translucent stroke colour is no longer written into `fill` / `stroke` as `#RRGGBBAA`, and three-digit shorthands such as `#fff` now match the stroke palette.

### DIFF
--- a/platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/Convertors.kt
+++ b/platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/Convertors.kt
@@ -6,11 +6,16 @@ import com.intellij.util.SVGLoader
 import com.intellij.platform.icons.design.BlendMode
 import com.intellij.platform.icons.filters.ColorFilter
 import com.intellij.platform.icons.impl.filters.TintColorFilter
+import com.intellij.platform.icons.impl.patchers.AUTHORED_STROKE_VARIANT_SUFFIX
 import com.intellij.platform.icons.impl.patchers.DefaultSvgPatcher
 import com.intellij.platform.icons.impl.patchers.SvgPatchOperation
+import com.intellij.platform.icons.impl.patchers.authoredStrokeSvgPatcher
+import com.intellij.platform.icons.impl.patchers.strokeSvgPatcher
+import com.intellij.platform.icons.impl.patchers.writeSvgAttribute
 import com.intellij.platform.icons.swing.toAwtColor
 import java.awt.Color
 import java.awt.image.RGBImageFilter
+import com.intellij.platform.icons.design.Color as DesignColor
 
 internal fun ColorFilter.toAwtFilter(): RGBImageFilter {
   return when (this) {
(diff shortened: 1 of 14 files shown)


rock3r:
(note: @jakub-senohrabek verbally approved this PR)

DanielSouzaBertoldi (approved):
All the concerns I found in #3585 were addressed in this one and I couldn't find anything to suggest/improve. Awesome work! 😎

rock3r:
That's two reviews, including @jakub-senohrabek's verbal go-ahead. Ready to merge!

wellingtoncosta (approved):

PR #3589, opened 2026-07-24
KTIJ-26390 don't flag internal usage across friend test modules

stokado:
KotlinInternalInJavaInspection compared PsiElement.module directly, so a Java test module could get a false positive when referencing an internal Kotlin declaration in its friend production module. Use getKaModule/directFriendDependencies instead.

[YouTrack Ticket](https://youtrack.jetbrains.com/issue/KTIJ-26390/False-positive-inspection-Usage-of-Kotlin-internal-declaration-from-different-module-is-reported-in-java-test-and-java)


stokado:
Not sure if I wired up test data correctly, but I tested in using both Gradle and IntelliJ build systems

akozlova:
I'd rename `isSameModule` to `isSameModuleOrFriends` or similar. 
Tests won't run this way automatically, one needs to run Generate Kotlin Tests configuration. Then the test class should be modified and you'll be able to check it. 
Thank you!
P.S. Sorry for the late response

stokado:
Hello Anna,
Thank you for the review.

I renamed method as you suggested.
Unfortunately I couldn't find `Generate Kotlin Tests` configuration. 

I tried to click green arrow in `plugins/kotlin/util/test-generator-all/test/org/jetbrains/kotlin/idea/testGenerator/all/GenerateAllTests.kt`, which created bazel run configuration, but it fails with following logs

image


```
com.intellij.platform.testFramework.core.FileComparisonFailedError: 'K2MultiFileInspectionTestGenerated.java' is not up to date
Use 'Generate Kotlin Tests' run configuration to regenerate tests
```

Could you help me create this task with right configuration?

akozlova:
Seems that `isUpToDateCheck` is `true` for you, could you please just set it to false locally and regenerate that class `K2MultiFileInspectionTestGenerated`?

stokado:
Thank you, it worked.
Looks like everything's in place.

akozlova:
Everything should work fine, thanks!

akozlova:
merged manually

PR #3590, opened 2026-07-24
Use --repo_env instead of --action_env for cpp toolchain flags

gharrma:
These env vars should be set with --repo_env since they are consumed by a repository rule. Using --action_env no longer works because Bazel 9+ sets --incompatible_repo_env_ignores_action_env by default.

cc @shalupov 

See also the example in https://github.com/bazelbuild/rules_cc/blob/main/README.md#default-autoconfigured-toolchain.


k15tfu (approved):

k15tfu:
Merged, thank you! https://github.com/JetBrains/intellij-community/commit/590966932ec7c1771025fc67873593c2c65a08e9

PR #3591, opened 2026-07-27
[JEWEL-1322] Fix Markdown colors not following IDE theme switches

DanielSouzaBertoldi:
# Context

Look, the main issue that this PR is supposed to fix is to use the correct colors when switching themes on the IDE. That's done, but while wiring up the GFM table and alert extensions in the devkit Compose demo to test the fix, I found that `Markdown()` never reads `LocalMarkdownBlockRenderer` aaaaand that is a problem, imo.

Its `blockRenderer` parameter defaulted to a plain `DefaultMarkdownBlockRenderer` with no renderer extensions, so any renderer configured through `ProvideMarkdownStyling` was ignored unless you also passed it to every single `Markdown()` call. Nothing fails loudly, either. The processor still parses tables and alerts into custom blocks and the default renderer just doesn't know them and skips them. Meanwhile `LazyMarkdown` reads the composition local, so the exact same markdown kept its tables there and lost them in `Markdown()`.

And then that one line fix got a [review comment](https://github.com/JetBrains/intellij-community/pull/3591#discussion_r3675413075) that made it bigger. Turns out `markdownStyling` only ever fed `blockVerticalSpacing`, everything else already came from the renderer. So the moment the renderer stops being rebuilt from that param, passing a styling explicitly would move the spacing and do nothing to the blocks 😭 That's worse than either option, so the renderer decides styling now, full stop, and `markdownStyling` is deprecated.

Soooo, this PR has three fixes:
  
1. A one line fix for the actual problem described in the issue on YT
2. Changing the default parameter of Markdown to reference `JewelTheme.markdownBlockRenderer` instead of creating a plain `DefaultMarkdownBlockRenderer`
3. Making the renderer the only source of styling, which is where most of the diff went

## Changes

* `isLightTheme` is now a getter that re-reads `JBColor.isBright()`, so bridge table and alert colors follow IDE theme switches.
* Both `Markdown()` overloads now default `blockRenderer` to `JewelTheme.markdownBlockRenderer`, like `LazyMarkdown` already did. Passing an explicit parameter still wins.
* New overloads of both `Markdown` variants and `LazyMarkdown` that don't take `markdownStyling`. They read the block styling and the block spacing off `blockRenderer.rootStyling`.
* The old ones are deprecated at `WARNING` level, with a `ReplaceWith` that folds the styling into the renderer via `createCopy(rootStyling = ...)`. The quickfix keeps your styling instead of dropping it on the floor. Calls that don't pass the param resolve to the new overload, so only the people actually using it get warned.
* The deprecated overloads go through a private `withStyling` helper. It reuses the renderer if `rootStyling` already matches, and it's `remember`ed so recomposing doesn't hand a brand new renderer identity downstream. Extensions keep whatever styling they were built with, so custom blocks don't follow along. That bit is documented on the params.
* New `MarkdownStylingSourceTest`, nine tests. Which styling wins, the reuse-or-copy branch in `withStyling`, and the `remember` surviving a recomposition. Plus the regression test for the original bug: a table rendered through `Markdown()` with the renderer supplied only via the composition local.
* `createMarkdownTestStyling` takes a `blockVerticalSpacing` param now, the tests need stylings that disagree about it.

## Screenshots

| Before | After |
|--------|--------|
| image | image |

As you can see, luckily for the alerts we don't really have any problems when switching themes.

### ⚠️ Important Changes

* `Markdown` now takes its default block renderer from `LocalMarkdownBlockRenderer` instead of creating a plain renderer with no extensions. If you call `Markdown()` outside `ProvideMarkdownStyling` and relied on the old implicit default, wrap the call in `ProvideMarkdownStyling` or pass a `blockRenderer` explicitly.
* The block renderer is now the single source of Markdown styling, including the spacing between blocks. If you pass a `markdownStyling` that disagrees with your renderer's `rootStyling`, you no longer get part of the output styled differently from the rest.

### Bug fixes

* `Markdown` no longer ignores the block renderer set up via `ProvideMarkdownStyling`, so GFM tables, alerts, and other extension blocks render without passing `blockRenderer` to every call
* Bridge Markdown table stripes and alert colors now follow IDE theme switches instead of staying on the theme the IDE started with

### Deprecated API

* The `markdownStyling` parameter on `Markdown` and `LazyMarkdown` is deprecated. Pass a `MarkdownBlockRenderer` that already carries the styling instead (normally through `ProvideMarkdownStyling`)


platform/jewel/markdown/extensions/gfm-tables/src/test/kotlin/org/jetbrains/jewel/markdown/extensions/github/tables/GitHubTableBlockRendererRememberKeysTest.kt
```diff
@@ -12,9 +13,13 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.test.ExperimentalTestApi
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
-import androidx.compose.ui.test.runComposeUiTest
+import androidx.compose.ui.test.v2.runComposeUiTest
```
rock3r, line 16: Huh we should likely do this across the repo, good point

rock3r: I filed [JEWEL-1383](https://youtrack.jetbrains.com/issue/JEWEL-1383/Migrate-Jewel-UI-tests-to-v2-API) for ya :D

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
```diff
@@ -64,7 +63,7 @@ public fun Markdown(
     onUrlClick: (String) -> Unit = {},
     markdownStyling: MarkdownStyling = JewelTheme.markdownStyling,
     processor: MarkdownProcessor = JewelTheme.markdownProcessor,
-    blockRenderer: MarkdownBlockRenderer = DefaultMarkdownBlockRenderer(markdownStyling),
+    blockRenderer: MarkdownBlockRenderer = JewelTheme.markdownBlockRenderer,
```
rock3r, line 71: Wouldn’t this leave an explicit `markdownStyling` parameter value only partially applied? Before this change, the implicit and wrong `DefaultMarkdownBlockRenderer(markdownStyling)` used it for the rendered blocks too; now, the outer column uses it, while the local renderer keeps its own `rootStyling`. This could lead to different styling at different places if the local block renderer has a custom style.

Could we keep the local renderer for its extensions but apply the explicitly passed styling to it as well? A `MarkdownBlockRenderer.withStyling(markdownStyling)` helper could delegate to `createCopy(rootStyling = markdownStyling)` only when it differs, avoiding a new instance for the usual case:

```kotlin
private fun MarkdownBlockRenderer.withStyling(styling: MarkdownStyling): MarkdownBlockRenderer =
    if (rootStyling == styling) this else createCopy(rootStyling = styling)
```

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
```diff
@@ -110,7 +109,7 @@ public fun Markdown(
     selectable: Boolean = false,
     onUrlClick: (String) -> Unit = {},
     markdownStyling: MarkdownStyling = JewelTheme.markdownStyling,
-    blockRenderer: MarkdownBlockRenderer = DefaultMarkdownBlockRenderer(markdownStyling),
+    blockRenderer: MarkdownBlockRenderer = JewelTheme.markdownBlockRenderer,
```
rock3r, line 167: Ditto

rock3r (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
DanielSouzaBertoldi: I'm afraid the rabbit hole is deeper than we thought 😬 

Looking at the code, I don't think that `withStyling` alone closes the gap and tbf, it can't.

`createCopy` in `DefaultMarkdownBlockRenderer` passes `rendererExtensions` directly, and by then the extensions have already captured their own `rootStyling`, meaning a copy carrying the explicit styling would restyle the built-in blocks but leave table cells and alert bodies on whatever their extension was constructed with.

Then I checked `markdownStyling` and it basically just feeds `blockVerticalSpacing` and nothing else (`Markdown.kt`, line 119 and `LazyMarkdown` lines 171 and 177). So the weird thing here is that the actual block style is coming from the renderer itself, the `markdownStyling` parameter does not style anything at all. Our `LazyMarkdown` has been broken like this since forever 😭 

My humble suggestion (spoiler alert: it introduces major breaking changes :) ) is to read spacing off `blockRenderer.rootStyling` and drop `markdownStyling` from these composables, so the renderer is the only thing deciding styling.

DanielSouzaBertoldi: Or, we do like you said (add a `withStyling`) but with a `remember` wrapped around it given that default argument expressions re-run on every recomposition

In any case, I'd rather discuss this further before proceeding with the changes 🕺🏻

rock3r: What about keeping the current API with the `withStyling` fix, but deprecate it, and add an overload that only takes in the block spacing from the local styling?

DanielSouzaBertoldi: Now we are talking 😎 that also works

DanielSouzaBertoldi: Given that this PR is really small as is, i dont think we need a follow up. We can afford a few more LOC here 😛

DanielSouzaBertoldi: @rock3r added the new overloads + updated the PR description to reflect the changes. I'll resolve this conversation for now but please feel free to add new suggestions/comments/feedback/life pro tips :)

rock3r (approved):

rock3r:
@DanielSouzaBertoldi note that this requires a metalava dump update!

DanielSouzaBertoldi:
Metalava dumps updated, thanks :)

DanielSouzaBertoldi:
Ready to merge

PR #3592, opened 2026-07-29
[JEWEL-1050] Create palette color accessors with default fallback

DanielSouzaBertoldi:
# Context

Currently there's no ergonomic way to get a sensible default color when the current LaF's palette is incomplete or missing, which is the common case for most 3rd-party themes. Consumers are basically forced to write something like `JewelTheme.palette.grayOrNull(10) ?: MyPaletteCopy.gray(10, JewelTheme.isDark)`, effectively keeping a private copy of Jewel's default palette values that goes stale the moment the real palette changes.

We can try and do better than this, which this PR will (hopefully) manage to do so :)

Now, instead of doing:

`val background = JewelTheme.palette.grayOrNull(14) ?: MyPaletteCopy.gray(14, JewelTheme.isDark)`

Users will just have to do:

`val background = JewelTheme.palette.grayOrDefault(14)`
  
#  Changes

- New `@Composable` extension functions on `ThemeColorPalette`: `grayOrDefault`, `blueOrDefault`, `greenOrDefault`, `redOrDefault`, `yellowOrDefault`, `orangeOrDefault`, `purpleOrDefault`, `tealOrDefault`. Each falls back through the palette's own entry, then Jewel's built-in default palette (light/dark and Islands-aware), then `Color.Unspecified` if the index is out of bounds even there.
- `DefaultColorPalette` (in `:ui`, with Light, Dark, IslandsLight, IslandsDark) is generated from the same `expUI_light`/`expUI_dark`/`ManyIslandsLight`/`ManyIslandsDark` theme JSONs already used elsewhere, via a new `intelliJColorPaletteGenerator` Gradle plugin and `DefaultColorPaletteGeneratorTask`. It's marked `@InternalJewelApi` and excluded from the public API dumps since it's implementation data is behind the accessors above, not meant to be used directly.
- `ThemeColorPalette.isIslands` is now public, so the new accessors (and any other consumer) can tell whether a palette uses the Islands color key format.
- `int-ui-standalone`'s generated `IntUiLightTheme`/`IntUiDarkTheme` reference `DefaultColorPalette.Light`/`.Dark` directly now, instead of each embedding its own copy of the same palette literals.
- A good chunk of the colors in the Islands theme JSONs are aliases pointing at another key rather than literal hex values, sometimes several levels deep. The generator follows these chains down to an actual  color instead of choking on them.
- Added tests covering palette-provided values, each fallback table (light/dark/Islands), the out-of-bounds case, and the gap-compacting fix above (`ThemeColorPaletteTest`, `ThemeColorPaletteOrDefaultTest`).
- Fixed a pre-existing bug in `BridgeThemeColorPalette.readFromLaF` where a palette with a gap in the middle of its sequence (e.g. a LaF declaring `Gray1` and `Gray3` but not `Gray2`) silently returned the next color over instead of falling back correctly, so our list-building was compacting the hole away instead of preserving it.
- Split `:ui`'s generated/ source root into generated/icons and generated/theme (one per generator), fixing an IDE "package directive doesn't match file location" warning caused by the two generators' outputs sharing one overlapping root.

### New features

  * Added `grayOrDefault`, `blueOrDefault`, `greenOrDefault`, `redOrDefault`, `yellowOrDefault`, `orangeOrDefault`, `purpleOrDefault`, and `tealOrDefault` accessors to `ThemeColorPalette`. They fall back to Jewel's built-in default palette when the current Look and Feel doesn't provide a color for the requested index, so you don't need to keep your own copy of the default values around just for a sensible fallback (handy for 3rd-party or incomplete themes).
  * `ThemeColorPalette.isIslands` is now public, so you can tell whether a palette uses the Islands color key format.

### Bug fixes
 * Fixed a bug where a Look and Feel palette with a gap in the middle of its color sequence (e.g. declaring Gray1 and Gray3 but not Gray2) would return the wrong color for the missing index instead of falling back to a sensible default.


platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.Nested
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.TaskAction
```
DanielSouzaBertoldi, line 23: Alright. This file basically holds the `readColors` logic from `IntUiThemeDescriptorReader` while also updating the algorithm to take into account the ManyIslands variations as well. Keep this in mind when reviewing it :)

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
```diff
@@ -0,0 +1,153 @@
...
+abstract class PaletteSource {
+    @get:InputFile
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    abstract val themeFile: RegularFileProperty
+
+    @get:Input
+    abstract val propertyName: Property<String>
+
+    @get:Input
+    abstract val islands: Property<Boolean>
```
DanielSouzaBertoldi, line 34: Can't call this guy `isIslands` because Kotlin compiles it to a JVM getter literally called `isIslands()` and that conflicts with Gradle.

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
```diff
@@ -0,0 +1,153 @@
...
+                    source.islands.get(),
+                )
+                .build()
+        )
+    }
+
+    /** Groups colors by family (e.g. "Gray1"/"gray-160" both group under "Gray"), sorted ascending by index. */
+    private fun buildColorGroupBlocks(colors: Map<String, String>): List<CodeBlock> =
+        colors.entries
+            .groupBy { it.key.replace("""-?\d+$""".toRegex(), "").replaceFirstChar { c -> c.uppercase() } }
```
DanielSouzaBertoldi, line 126: Islands theme defines color names following the format: `<color>-<number>`, while the old theme had the format `<Color><Number>`.

Old theme: `Gray1`
Islands: `gray-10`

This code will turn both into `Gray`.

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
```diff
@@ -0,0 +1,153 @@
...
+            .filterKeys { colorGroups.contains(it) }
+            .mapValues { (_, entries) -> entries.sortedBy { it.key.takeLastWhile(Char::isDigit).toInt() } }
```
DanielSouzaBertoldi, line 128: Had to add this sort logic because the old `readColors` code just assumed all colors were declared in the JSON file in ascending order (tbf they were). But then islands came in and was all like "what if my json file declared colors in descending order?" and now we have to sort it out.

For the old theme this is a no-op

For islands themes (or any other theme file) we can guarantee that the result will be in ascending order :)

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
```diff
@@ -0,0 +1,153 @@
...
+
+        val aliased = colors[value] ?: error("Unresolved color alias '$value'")
+        return resolveHex(aliased, colors, depth + 1)
+    }
+}
+
+@Serializable
+data class IntelliJColors(
+    val colors: Map<String, String> = emptyMap(),
+)
```
DanielSouzaBertoldi, line 157: Meh we only care about colors declared in the JSON file so no need to add anything else here to be serialized.

platform/jewel/ui/generated/theme/org/jetbrains/jewel/ui/theme/DefaultColorPalette.kt
```diff
@@ -0,0 +1,1389 @@
...
+// *** DO NOT EDIT MANUALLY! Use intelliJColorPaletteGenerator task to update. ***
+//
+package org.jetbrains.jewel.ui.theme
+
+import androidx.compose.ui.graphics.Color
+import kotlin.Suppress
+import org.jetbrains.jewel.foundation.theme.ThemeColorPalette
+
+@Suppress("LargeClass")
+public object DefaultColorPalette {
```
DanielSouzaBertoldi, line 15: Yeahh.... one giant file.

Also, I decided to have this file in `:ui` so that we don't have to duplicate the same file in both `int-ui-standalone` and `ide-laf-bridge`.

I'm deliberately trading the theme-agnostic vibe of `:ui` for a single source of truth. The alternative risked two independently-generated copies of the same literal table silently drifting apart, enforced only by build-wiring discipline rather than the type system.

Let me know what you think :)

(this file had to live either in `:ui` or `:foundation`)

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
nebojsa-vuksic, line 34: Can we name it `isIslandsTheme`?

nebojsa-vuksic (approved):

platform/jewel/ui/generated/theme/org/jetbrains/jewel/ui/theme/DefaultColorPalette.kt
```diff
@@ -0,0 +1,1425 @@
...
```
rock3r, line 15: Please keep the generated palette table out of the stable API!

This generated object and its four properties are added to `ui`'s stable API dump, even though they are just implementation data behind the new `*OrDefault` accessors. Please make this object internal and remove it from the public API baseline. We never intended this to ship as API (and it does not have any KDoc anyway, which is a requirement for public APIs)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/theme/ThemeColorPalette.kt
```diff
@@ -0,0 +1,164 @@
...
+private fun ThemeColorPalette.defaultTable(): ThemeColorPalette =
+    when {
+        isIslands && JewelTheme.isDark -> DefaultColorPalette.IslandsDark
+        isIslands && !JewelTheme.isDark -> DefaultColorPalette.IslandsLight
+        !isIslands && JewelTheme.isDark -> DefaultColorPalette.Dark
+        else -> DefaultColorPalette.Light
+    }
+
+@Composable
+private fun ThemeColorPalette.orDefault(index: Int, accessor: ThemeColorPalette.(Int) -> Color?): Color =
```
rock3r, line 21: `readFromLaF()` turns each palette map into `values.toList()`, which compacts holes. For example, a 3p LaF containing only `Gray1` and `Gray3` makes `grayOrNull(2)` return `Gray3`, and this accessor returns the wrong color instead of falling back. Retain the index gaps (and have the nullable accessor treat them as absent) before relying on this fallback contract. (I think this is pre-existing, but we should fix it)

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
rock3r: What about JvmName?

platform/jewel/buildSrc/src/main/kotlin/intellij-color-palette-generator.gradle.kts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+val extension = container<ColorPaletteGenerator> { ColorPaletteGenerator(it, project) }
+
+extensions.add("intelliJColorPaletteGenerator", extension)
+
+val targetDir: Provider<Directory> = project.provider { project.layout.projectDirectory.dir("generated/theme") }
+
+val generateDefaultColorPalette =
+    tasks.register<DefaultColorPaletteGeneratorTask>("generateDefaultColorPalette") {
```
rock3r, line 26: This only wires the generated table into Gradle. The Bazel target still compiles the checked-in `generated/**/*.kt` but has no way to generate the palette, so a Bazel-only update to a palette can ship the stale values as last read by Gradle.

Please add Bazel generation/input wiring (or at least a checked-in-output verification) so both build systems keep `DefaultColorPalette` synchronized.

rock3r (changes requested):
The PR introduces the public fallback behavior without tests. Please cover a palette-provided value, a missing value falling back to the correct light/dark and Islands table, and `Color.Unspecified` when neither palette contains the requested index. Also add tests for the gaps compacting behaviour to avoid regressions

platform/jewel/buildSrc/src/main/kotlin/intellij-color-palette-generator.gradle.kts
DanielSouzaBertoldi: Soooo here's what I was thinking:

I'm about to create the part 3 gradle -> bazel migration PR, which will also migrate the IntelliJThemeGenerator script. If I were to add a Bazel version in this PR, I'd have to also create the necessary BUILD.bazel files and folder structure, which for this PR meehhh I'm not really a fan of.

Having said that, I'll shift gears and focus on bazel part 3 first so we can merge it ASAP and then add the bazel version of this same script without any friction.

Wdyt?

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
DanielSouzaBertoldi: Unfortunately, no :(

We will hit the same wall: Kotlin compiler generates a `is*` version of the variable anyways

DanielSouzaBertoldi: Nope :(

If I try to add this annotation, I get the error:

> JvmName annotation is not applicable to this declaration.

Looks like this annotation can't be applied to abstract variables

platform/jewel/buildSrc/src/main/kotlin/intellij-color-palette-generator.gradle.kts
rock3r: Ok with postponing to that PR, as long as you don't forget :P

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/theme/ThemeColorPalette.kt
DanielSouzaBertoldi: Had no clue this could happen, damn. Thanks for pointing it out, I'll fix this :)

platform/jewel/buildSrc/src/main/kotlin/intellij-color-palette-generator.gradle.kts
DanielSouzaBertoldi: I gotchu fam 😎

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
rock3r: Lol poor Parth (you must be used to this I suppose, Parth... also, hi Parth!)

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/theme/ThemeColorPalette.kt
rock3r: We missed it before too. This is GPT putting me on the scent of something being afoot

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/theme/IntelliJColorPaletteGeneratorTask.kt
DanielSouzaBertoldi: WHAT?!!! HAHAHHAA sorry man, didn't even noticed GitHub tagged you when I added my reply 😭

DanielSouzaBertoldi:
@rock3r alright, added a bunch of unit tests, fixed the laf bug, made the default palette object internal :) the bazel version of the script will be done in the next bazel migration PR

platform/jewel/ui/generated/theme/org/jetbrains/jewel/ui/theme/DefaultColorPalette.kt
```diff
@@ -0,0 +1,1429 @@
+// Generated by the Jewel Int UI Color Palette Generator for IJP 263
+// *** DO NOT EDIT MANUALLY! Use intelliJColorPaletteGenerator task to update. ***
+//
+package org.jetbrains.jewel.ui.theme
```
DanielSouzaBertoldi, line 4: So, here's a "fun" thing I noticed:

since the icons API is also generated in the ui module, when I added this generated code in `generated/theme/org/jetbrains...` the `theme` directory as treated as a package (thanks to `<sourceFolder url="file://$MODULE_DIR$/generated" isTestSource="false" generated="true" />` in the `iml` file!)

So this meant that the actual package would have to be `theme.org.jetbrains...` which is a whatever kind of thing but at the same time it doesn't follow the naming convention for packages in the repo.

To circumvent this, I had to move this generated code out of `generated/` and into a sibling directory called `generated-theme`.

Let me know if you have a better idea, if you'd rather have the generated code under `theme.org.jetbrains.yaddayadda` or some other alternative :)

rock3r: Wouldn't setting the generated sources root to `theme` instead of `generated` also work? Or, alternatively, drop the `theme` folder level (but I'd rather do the former)

rock3r (approved):
LGTM, but will leave it to you @DanielSouzaBertoldi if you wanna tweak the generated folder thing before shipping

platform/jewel/ui/generated/theme/org/jetbrains/jewel/ui/theme/DefaultColorPalette.kt
DanielSouzaBertoldi: Wait, your first idea is to have a folder structure like this:

```
ui/
├─ generated/
│  ├─ org.jetbrains.jewel.ui.icons/
├─ theme/
│  ├─ org.jetbrains.jewel.ui.docs/
├─ src/
│  ├─ blabla/
├─ other files
```

Or do you mean like this?

```
ui/
├─ generated/
│  ├─ theme/
│  │  ├─ org.jetbrains.jewel.ui.docs/
│  ├─ org.jetbrains.jewel.ui.icons/
├─ src/
│  ├─ blabla/
├─ other files
```

rock3r: I mean:

```
ui/
├─ generated/
│  ├─ theme/
│  │  ├─ org.jetbrains.jewel.ui.docs/
│  ├─ icons/
│  │  ├─ org.jetbrains.jewel.ui.icons/
│  ├─ ...
├─ src/
│  ├─ blabla/
├─ other files
```

rock3r: We can register both theme and icons as generated sources dirs, so we keep em separate and clean, and they all get merged anyway

DanielSouzaBertoldi: image

Done!

DanielSouzaBertoldi:
@rock3r re-requesting your review just to play it safe :P

rock3r (approved):
Looks fine, thank you!

DanielSouzaBertoldi:
Ready to merge

PR #3593, opened 2026-07-29
Invoke ModuleToGradleProjectConnector before DataServices

karimai-gg:
So it can properly prepare for data deletion in case when the gradle.phased.sync.disabled isn't set.


nskvortsov:
Declined after discussion at https://youtrack.jetbrains.com/issue/IDEA-392182/ModuleToGradleProjectConnector-runs-after-the-data-services-have-run#focus=Comments-27-14235524.0-0

PR #3594, opened 2026-07-29
Invoke ProjectDataImportExtension#finalizeImportData after DataServic…

karimai-gg:
…e#postProcess


nskvortsov:
Thank you for the fix. Checking it with our CI.

PR #3595, opened 2026-07-30
[JEWEL-1384] Fix coroutines NoSuchMethodError in standalone artifacts

DanielSouzaBertoldi:
We got a [crash report](https://youtrack.jetbrains.com/issue/JEWEL-1374) that happens when users targets Jewel `0.38.*` or `0.39.*` versions and explicitly targets coroutines 1.11.0+. The crash only happens on packaged standalone apps that call `runBlocking`:

```
- Exception in thread "main" `java.lang.NoSuchMethodError: 'java.lang.Object kotlinx.coroutines.BuildersKt.runBlockingK$default(kotlin.coroutines.CoroutineContext, kotlin.jvm.functions.Function2, int,
  java.lang.Object)'`
```

### TL;DR:
Stock coroutines 1.11.0 [reestructured the runBlocking function](https://github.com/Kotlin/kotlinx.coroutines//blob/1.11.0/kotlinx-coroutines-core/concurrent/src/Builders.concurrent.kt#L156) so Kotlin call sites compile to `BuildersKt.runBlockingK` rather than `runBlocking`, while Java callers get a separate overload that keeps the old name.

After we updated our standalone library POM file to also depend on the new icons api, we also inadvertently added the transitive dependencies from the new icons API. One of the transitive dependency is the IJP's fork of coroutines (`org.jetbrains.intellij.deps.kotlinx:kotlinx-coroutines-core-jvm`).

So, effectively we have two `kotlinx-coroutines-core-jvm` libraries. Gradle doesn't get rid of one over the other because their group id is different so it assumes they are completely different libraries.

Except they are not, so when we package the distribution of our code, the coroutines fork wins the classpath race (looks like jpackage writes the classpath alphabetically, so `1.10.2-intellij-1` lands before `1.11.0`), the bytecode tries to call `runBlockingK` function which does not exist in the fork and bang. Crash.

---
  
`icons-api`, `icons-api-rendering` and `icons-impl` all declare a dependency on the IntelliJ Platform fork of coroutines (`org.jetbrains.intellij.deps.kotlinx:kotlinx-coroutines-core-jvm`). The fork uses a different `groupId` than the stock artifact, so Gradle and Maven do not see the two as the same module. No conflict resolution happens and consumers get both jars. Both contain the `kotlinx.coroutines.*` packages, so whichever jar comes first on the classpath shadows the other. jpackage writes the classpath alphabetically, and `1.10.2-intellij-1` sorts before stock `1.11.0`, so the fork wins.

The fork lags upstream. Stock 1.11.0 declares `runBlocking` with `@JvmName("runBlockingK")`, so Kotlin call sites compile to `BuildersKt.runBlockingK`. The fork has no such method and every `runBlocking` call fails to link.

This does not reproduce under `./gradlew run`, because Gradle orders the runtime classpath by dependency graph rather than alphabetically, and stock happens to come first there. It only appears in the packaged artifact, so it passes CI and reaches users.

The icons POMs have declared the fork since MRI-4371 made those modules publishable to Maven Central. Nothing consumed them from outside the IDE until [JEWEL-1374] declared them in the Jewel POMs, which is why this surfaced in 0.38.1 and 0.39.1.

None of the three modules use anything the fork adds: `IntellijCoroutines`, `ParallelismCompensation`, `softLimitedParallelism`, `runBlockingWithParallelismCompensation`, `SoftLimitedDispatcher`, `FlowValueWrapperInternal` and `probeJobCreated` have zero hits in their bytecode. `icons-api` references no coroutines at all, `icons-api-rendering` only uses `CoroutineScope`, and `icons-impl` only uses stock API. Stock coroutines is a safe substitute.

## Changes

* Add `JewelMavenArtifacts.patchPlatformDependencies`, which repoints the coroutines dependency of the published Icons API modules from the fork to the stock artifact. The `artifactId` is the same in both groups, and dropping the `-intellij-N` suffix gives the stock version the fork is based on (`1.10.2-intellij-1` becomes `1.10.2`), so normal conflict resolution can pick a single version.
* Wire `patchPlatformDependencies` into `IdeaCommunityProperties.mavenArtifacts.patchDependencies` for `isPublishedPlatformDependency` modules, which previously fell through unpatched.
* Add `checkNoCoroutinesForkIsPublished` to `JewelMavenArtifacts.validate`, failing the build if any generated POM declares the fork `groupId`. It reads the generated POMs rather than the patched dependency lists, so it also catches the fork arriving transitively.
* Fix the "Testing publishing locally" section of the releasing guide. The `BuildContextImpl`/`ULTIMATE_HOME` workaround it describes no longer exists, since `JewelMavenArtifactsBuildTarget` already passes the community root. The publish script path was also wrong, the build number is optional, and the tolerated exit code 1 was undocumented.

The dependency is repointed rather than dropped because `icons-impl` declares no Compose dependency and uses `BuildersKt`, `Job`, `channels`, `flow` and friends at runtime. Dropping it would leave anyone depending on `icons-impl` directly with no coroutines provider at all.

## Verification

Published locally with `../scripts/publishJewelStandaloneToMavenLocal.cmd`. All three icons POMs now declare `org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.10.2`, and no generated POM mentions the fork `groupId`.

A standalone reproducer that calls `runBlocking` and compiles against stock coroutines 1.11.0 now packages with a single coroutines jar and launches. Its call site still compiles to `runBlockingK$default`, so the fix is not hiding the problem by falling back to an older API.

<details><summary>Want to test this yourself? Check the steps here</summary>
<p>

1. Check out this branch
2. Run the script in `platform/jewel/scripts/publishJewelStandaloneToMavenLocal.cmd`
3. Clone the reproducer here: https://github.com/DanielSouzaBertoldi/JewelIconsApiRunBlockingCrashReproducer
4. Run `./gradlew packageDistributionForCurrentOS`
5. Open the generated app in `build/compose/binaries/main/app/JewelStandaloneReproducer`
    * The crash should not happen if the root `build.gradle` file targets Jewel `0.40.0-263.SNAPSHOT` (if it can't find the version, check what version was locally generated in step 2)
    * Change it to `0.39.1-262.9437.29`, rerun the app, and it should crash

</p>
</details> 

## Screen Recordings

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/353e2741-72c5-406a-9ee9-4f8079012c9a"/> | <video src="https://github.com/user-attachments/assets/1b69c15e-f008-48aa-9916-9dabc48f3e28" /> |

### Bug fixes

 * Fixed a `NoSuchMethodError` crash on startup in packaged Jewel Standalone apps, caused by the IntelliJ Platform fork of `kotlinx-coroutines-core` being published as a transitive dependency of the Icons API modules alongside the stock artifact. Projects that worked around this with a `replacedBy` module rule can now remove it.

### DIFF
--- a/build/src/org/jetbrains/intellij/build/JewelMavenArtifacts.kt
+++ b/build/src/org/jetbrains/intellij/build/JewelMavenArtifacts.kt
@@ -21,8 +21,34 @@ import org.jetbrains.jps.model.module.JpsDependencyElement
 import org.jetbrains.jps.model.module.JpsModule
 import org.jetbrains.jps.model.module.JpsModuleDependency
 import kotlin.io.path.exists
+import kotlin.io.path.name
+import kotlin.io.path.readText
 
 private const val GROUP_ID: String = "org.jetbrains.jewel"
+
+/**
+ * The IJP fork of kotlinx.coroutines, and the stock artifact it is forked from.
+ *
+ * The fork shares the `kotlinx.coroutines.*` packages with the stock artifact but has a different groupId, so
+ * neither Gradle nor Maven can conflict-resolve the two. Consumers get both jars, and whichever one sorts first
+ * on the classpath wins. The fork lags upstream, so when it wins, anything stock added since the fork point is
+ * missing: stock 1.11.0 compiles `runBlocking` to `BuildersKt.runBlockingK` via `@JvmName`, the fork has no such
+ * method, and every `runBlocking` call in a consumer throws `NoSuchMethodError`.
+ *
+ * The fork must never reach standalone consumers. See [JewelMavenArtifacts.patchPlatformDependencies].
+ */
+private const val IJP_COROUTINES_FORK_GROUP: String = "org.jetbrains.intellij.deps.kotlinx"
+private const val STOCK_COROUTINES_GROUP: String = "org.jetbrains.kotlinx"
+
+/** The fork appends this to the stock version it is based on, e.g. `1.10.2-intellij-1` for stock `1.10.2`. */
+private const val IJP_COROUTINES_FORK_VERSION_SUFFIX: String = "-intellij"
+
+/**
+ * Artifact ID prefix used by the upstream coroutines artifacts and by the IJP fork. Matching on it rather than on
+ * [IJP_COROUTINES_FORK_GROUP] also covers a fork that moves to another groupId. This is a best-effort guard: a fork
+ * published under an unrelated artifact ID would still get through.
+ */
+private const val COROUTINES_ARTIFACT_PREFIX: String = "kotlinx-coroutines"
 
 /**
  * Each entry represents a prefix for Platform dependencies which are being published to Maven Central.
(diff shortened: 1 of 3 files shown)


build/src/org/jetbrains/intellij/build/JewelMavenArtifacts.kt
```diff
@@ -180,6 +196,32 @@ internal object JewelMavenArtifacts {
...
+    dependencies: List<MavenArtifactDependency>,
+  ): List<MavenArtifactDependency> {
+    check(isPublishedPlatformDependency(module))
+    return dependencies.map { dependency ->
+      val coordinates = dependency.coordinates
+      if (coordinates.groupId != IJP_COROUTINES_FORK_GROUP) return@map dependency
+      dependency.copy(
+        coordinates = coordinates.copy(
+          groupId = STOCK_COROUTINES_GROUP,
+          version = coordinates.version.substringBefore("-intellij"),
```
rock3r: Nit: do we want to insert a 'fail loudly' guard if the version here doesn't have the expected format? E.g. if at some point they switch to `-ij` or whatever else, we should not leave that in

rock3r (approved):
Nice fix, thanks dude!

nebojsa-vuksic:
@DanielSouzaBertoldi Everything looks good to me. Great job! Can we also add the tests for `patchPlatformDependencies`?

build/src/org/jetbrains/intellij/build/JewelMavenArtifacts.kt
DanielSouzaBertoldi: I mean, as much as we can try there's no way to guarantee that we will guard against every single possible variation. We can make this more resilient but there will be some risk if the fork drops "coroutines" in the name or something even more crazy (and even more unlikely)

So I added two guards that try their best to avoid this situation:

1. If the version doesn't contain the `-intellij` suffix we strip, so a switch to `-ij` blows up instead of silently publishing a stock version that doesn't exist.
2. If any kotlinx-coroutines-* artifact whose group isn't `org.jetbrains.kotlinx`, so a fork moving to a different groupId gets caught too.

Maybe overkill, but better safe than sorry right?

DanielSouzaBertoldi:
@nebojsa-vuksic added tests :)


Btw, any real reason why we can't use `projectHome = COMMUNITY_ROOT.communityRoot` for the tetsts in `MavenArtifactsBuilderTest`?

build/src/org/jetbrains/intellij/build/JewelMavenArtifacts.kt
rock3r: Indeed. Unlikely to be needed, but still. Better safe than sorry

DanielSouzaBertoldi:
Ooops clicked the wrong button :(

rock3r (approved):

DanielSouzaBertoldi:
Ready to merge

PR #3598, opened 2026-08-03
[JEWEL-1387] Fix native window lookup on JBR 25

rock3r:
JBR 25 uses `LWCWindowPeer`, which inherits `getPlatformWindow()` rather than declaring it. Jewel’s lookup only checked the runtime class, causing native macOS window chrome updates to silently stop working.

## Changes

* Search the AWT peer hierarchy for `getPlatformWindow()` and the native `ptr` field.
* Mark the reflection helpers as test-only internal APIs.
* Add a regression test covering inherited reflective members.

### Bug fixes

* Fixed native macOS window-chrome color and full-screen-button updates when running standalone Jewel applications on JBR 25.


DanielSouzaBertoldi (approved):
Thanks for finding and fixing this bug! JNA will always bite us when we are least expecting :(

rock3r:
Ready to merge

PR #3600, opened 2026-08-05
[JEWEL-1342] Move Jewel and YouTrack skills to .agents/skills

rock3r:
Three agent skills — `jewel-pr-preparer`, `jewel-release-helper`, and the YouTrack CLI — were authored in `.claude/skills`, which holds generated stubs rather than canonical sources. That made them visible to a single harness only. This moves all three to `.agents/skills`, the harness-neutral canonical tree.

The YouTrack skill is also renamed `managing-youtrack` → `youtrack-community`, because its Community-only scope was previously left to be inferred. IJAI-345 had replaced a working, self-contained Community skill with a stub pointing at `.agents/skills/youtrack/scripts/youtrack.ts` — a path that does not exist in `intellij-community` and cannot be made to exist from here, since the `youtrack` skill is monorepo-only. In the monorepo that pointer resolves and the breakage is invisible; in a standalone Community checkout it resolves to nothing, leaving a skill whose entire content was an instruction to use something the reader demonstrably does not have. JEWEL-1367 refilled the hole; this change makes the boundary explicit in the name and in the skill's first paragraph, so the next consolidation cannot quietly hollow it out again.

## Changes

* Move `jewel-pr-preparer` and `jewel-release-helper` from `.claude/skills` to `.agents/skills`, with their scripts intact, and regenerate their `.claude/skills` stubs.
* Move the YouTrack CLI skill to `.agents/skills/youtrack-community`, carrying `yt.py`, its test suite, references, and evals unchanged, and regenerate its `.claude/skills` stub.
* State the Community-only scope in the skill body and description, and direct monorepo users to the `youtrack` skill.
* Shorten the three skills' descriptions to single lines. The guide renderer hard-fails above 160 UTF-8 bytes per skill and 6 KiB per edition (`.ai/render-guides.mjs:310-311`); these used folded scalars of 414, 347 and 416 bytes. The full set now measures 2885 bytes, longest 145.
* Regenerate `.agents/skills/INDEX.md`, which now lists all 34 skills.

### Why each skill appears twice in the diff

The diff adds each skill under both `.agents/skills` and `.claude/skills`, which looks like duplication but is the layout this repository already uses. `.agents/skills` holds the canonical source; `.claude/skills` holds a generated stub carrying the `<!-- Generated by community/.ai/render-guides.mjs -->` marker, plus a copy of the skill's bundled assets so that harness-specific skill-directory variables resolve to a directory containing the scripts. Every existing skill is checked in that way — `kotlin-ui-dsl`, added in IJPL-252121, is the most recent example — and `AgentGuidelinesConsistencyTest` fails the build when a canonical source has no committed stub. The stubs here are generated by the renderer, not hand-written; the counts line up at 34 sources, 34 stubs and 34 index rows, with no orphans in either direction.

### Various notes

Worth flagging that the 160-byte per-skill limit is an arbitrary local constant, not a platform requirement — the agentskills.io spec allows up to 1024 characters. Those two are not directly comparable: the renderer counts UTF-8 bytes, the spec counts characters, and what actually consumes context is tokens. A non-ASCII character can cost several bytes while still being one character, so a byte budget silently penalises em dashes, accented names and non-Latin scripts more than plain ASCII, and none of the three units maps cleanly onto the cost the budget is presumably trying to control. Compressing descriptions this far also risks degrading skill recall, since the description is the only signal an agent has when deciding whether a skill applies. These are shortened to satisfy the renderer, not because the limit is justified; it would be worth backing that constant with actual retrieval evals, and measuring in tokens if context is the real constraint.

Guide outputs are unchanged by this PR. `render-guides.mjs` cannot currently run in a standalone Community checkout — it derives the repository root two levels above itself, which only holds in the ultimate layout where it sits at `community/.ai` — but that fix is out of scope here and has been handed to @develar as a separate patch.

## Verification

The YouTrack CLI's 124 unit tests pass from the new location, `evals.json` parses, and no references to the old skill name remain.


nebojsa-vuksic (approved):

rock3r:
Ready to merge

DanielSouzaBertoldi (approved):

wellingtoncosta (approved):

rock3r:
FYI — had to manually run the render script again and re-duplicate the skills from .agents into .claude, as that is the process. Changed lines grew from +24 -20 to a lot more because of that, but it's only a mechanical copy. No other changes.

PR #3601, opened 2026-08-05
KTIJ-39829 UAST: Inline reified properties from library fail to resolve to a fake light method

francescoo22:


francescoo22:
@KirpichenkovPavel @jsjeon

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
```diff
@@ -161,23 +166,38 @@ internal fun toPsiClass(
...
+                            callableSymbol.setter ?: callableSymbol
+                        } else {
+                            callableSymbol.getter ?: callableSymbol
+                        }
+                    } else {
+                        callableSymbol
+                    }
+
+                    val methodName = when (actualSymbol) {
+                        is KaPropertyAccessorSymbol -> actualSymbol.javaMethodName ?: ""
```
KirpichenkovPavel: The empty string fallbacks might be dangerous for properties with legit `javaMethodName == null`.
```kotlin
inline val <reified T : Any> T.`reified val`: String
    get() = "test"
```

WDYT about returning `null` PsiMethod in this case instead?

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
```diff
@@ -188,6 +208,20 @@ private fun fakePsiMethodForReifiedInline(
     return null
 }
 
+context(_: KaSession)
+private fun KaDeclarationSymbol.isReified(): Boolean = when (this) {
+    is KaPropertyAccessorSymbol -> containingDeclaration?.isReified() == true
+    else -> typeParameters.any { it.isReified }
+}
+
+context(_: KaSession)
```
KirpichenkovPavel: Nit: context seems unnecessary here.

KirpichenkovPavel (changes requested):

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
francescoo22: fixed

plugins/kotlin/uast/uast-kotlin-fir/src/org/jetbrains/uast/kotlin/internal/firKotlinInternalUastUtils.kt
francescoo22: I'm not a big fan of returning null because we would only partially fix the issue by doing that. 

```
inline val <T : Any> T.`non reified val`: String
    get() = "test"
```

Non reified inline properties from external libraries resolve correctly, so probably also reified ones should. I pushed a workaround for it that seems to work. Let me know what you think.

francescoo22: I also changed the `else` branch a bit by returning a `null` psi method instead of an empty string when name can't be found. In theory, we should never end up in the `else` branch at all, so I guess returning null psi element if we cannot even find a name in that case can be safe.

PR #3602, opened 2026-08-06
[JEWEL-1360] Gradle Scripts to Bazel Migration - Part 3

DanielSouzaBertoldi:
## Context

Part 3 of the Jewel Bazel migration: ports `annotate-api-dump-changes.main.kts`, `check-api-dumps.main.kts`, `compare-branches.main.kts`, and `validate-maven-artifacts.main.kts` to Bazel-buildable Kotlin, alongside their `.main.kts` originals under `platform/jewel/scripts/`. Each ported command takes an injected `CommandRunner` instead of shelling out directly — a real `DefaultCommandRunner` in production, a `FakeCommandRunner` in tests — and `CompareBranches`/`ValidateMavenArtifacts` take a couple of small injectable filesystem lambdas (`pathExists`, `isDirectory`, `isFile`) for the same reason.

Every script got built and tested for real via `bazel build`/`bazel test`, and checked line-by-line against its live `.main.kts` original. That turned up a few real bugs along the way, fixed here instead of left to rot further:

* `JewelBazelScriptsUtils.kt`'s `runCommand` never actually enforced `timeoutAmount` — `process.waitFor()` had no timeout argument, so a hung subprocess would block forever instead of aborting. It now uses the timed `waitFor` overload and kills the process on timeout.
* `CheckApiDumps.kt` had the exact "false green" bug a recent Jewel PR just fixed in `check-api-dumps.main.kts`. A stale `tests.cmd` invocation (which `tests.cmd` now rejects) combined with a missing `set -e` meant the check reported success without the test ever running. Restored `--module`/`--test`, `set -e -o pipefail`, and single-quoted shell-literal path escaping, and pulled the command/script-building logic out into testable functions while at it.
* `branchExists`, in both `JewelBazelScriptsUtils.kt` and the original `utils.main.kts`, was missing `exitOnError = false` — unlike `isDirectoryGitRepo` sitting right next to it. Since `git rev-parse --verify` exits non-zero for exactly the case `branchExists` exists to catch, `compare-branches`' own "Branch not found" message was dead code in both scripts. Fixed in both, not just the port.

A few smaller things got cleaned up too: a "Checking aginst base commit" typo, `requireGhTool` not going through the shared `exitWithError` helper like every other error path in the file, and a dropped trailing period in `CheckApiDumps`' `--verbose` help text.

## Changes

* Ports `AnnotateApiDumpChanges.kt`, `CheckApiDumps.kt`, `CompareBranches.kt`, and `ValidateMavenArtifacts.kt`, plus a shared `CliktUtils.kt` (`exitWithError`) used by all four Clikt-based commands.
* Adds a full test suite: `FakeCommandRunner` plus one test file per script, exercising the extracted pure/testable functions (`processDiff`, `buildApiCheckCommand`/`buildApiCheckScript`, `fetchCommits`, `crossValidateDependencies`, `cleanup`).
* Fixes the three bugs above (subprocess timeout, `CheckApiDumps` false green, `branchExists`), plus the minor typo/consistency fixes.

### DIFF
--- a/platform/jewel/build-scripts/bazel/BUILD.bazel
+++ b/platform/jewel/build-scripts/bazel/BUILD.bazel
@@ -1,5 +1,15 @@
-load("@rules_java//java:defs.bzl", "java_binary")
-load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_import", "kt_jvm_library")
+load("@rules_java//java:defs.bzl", "java_binary", "java_library")
+load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_import", "kt_jvm_library", "kt_jvm_test")
+
+java_library(
+    name = "clikt-dependencies",
+    exports = [
+        "@lib//:fleet-clikt",
+        "@lib//:fleet-clikt-core",
+        "@lib//:kotlinx-coroutines-core",
+    ],
+    visibility = ["//visibility:public"],
+)
 
 kt_jvm_import(
     name = "kotlinpoet",
@@ -20,6 +30,15 @@ kt_jvm_library(
 )
 
 kt_jvm_library(
+    name = "clikt-utils",
+    srcs = ["src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CliktUtils.kt"],
+    deps = [
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+    ],
+)
+
+kt_jvm_library(
     name = "ktfmt-runner-lib",
     srcs = ["src/main/kotlin/org/jetbrains/jewel/scripts/bazel/KtfmtRunner.kt"],
     deps = [
@@ -34,6 +53,45 @@ kt_jvm_library(
     deps = [
         ":jewel-bazel-scripts-utils",
         "@lib//:kotlinx-coroutines-core",
+    ],
+)
+
+kt_jvm_library(
+    name = "annotate-api-dumps",
+    srcs = ["src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChanges.kt"],
+    deps = [
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+        ":clikt-utils",
+    ],
+)
+
+kt_jvm_library(
+    name = "check-api-dumps",
+    srcs = ["src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CheckApiDumps.kt"],
+    deps = [
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+        ":clikt-utils",
+    ],
+)
+
+kt_jvm_library(
+    name = "compare-branches",
+    srcs = ["src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CompareBranches.kt"],
+    deps = [
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+    ],
+)
+
+kt_jvm_library(
+    name = "validate-maven-artifacts",
+    srcs = ["src/main/kotlin/org/jetbrains/jewel/scripts/bazel/ValidateMavenArtifacts.kt"],
+    deps = [
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+        ":clikt-utils",
     ],
 )
 
@@ -55,4 +113,123 @@ java_binary(
     name = "jewelVersionUpdater",
     main_class = "org.jetbrains.jewel.scripts.bazel.JewelVersionUpdaterKt",
     runtime_deps = [":jewel-version-updater"],
+)
+
+java_binary(
+    name = "annotateApiDumpChanges",
+    runtime_deps = [":annotate-api-dumps"],
+    main_class = "org.jetbrains.jewel.scripts.bazel.AnnotateApiDumpChangesKt",
+)
+
+java_binary(
+    name = "checkApiDumps",
+    runtime_deps = [":check-api-dumps"],
+    main_class = "org.jetbrains.jewel.scripts.bazel.CheckApiDumpsKt",
+)
+
+# Run this on CLI like: bazel run //platform/jewel/build-scripts/bazel:compareBranches -- branch1 branch2
+java_binary(
+    name = "compareBranches",
+    runtime_deps = [":compare-branches"],
+    main_class = "org.jetbrains.jewel.scripts.bazel.CompareBranchesKt",
+)
+
+# Run this on CLI like: bazel run //platform/jewel/build-scripts/bazel:validateMavenArtifacts -- branch1 branch2
+java_binary(
+    name = "validateMavenArtifacts",
+    runtime_deps = [":validate-maven-artifacts"],
+    main_class = "org.jetbrains.jewel.scripts.bazel.ValidateMavenArtifactsKt",
+)
+
+# Tests
+kt_jvm_test(
+    name = "bazel-test-utils",
+    srcs = ["src/test/kotlin/org/jetbrains/jewel/scripts/bazel/BazelTestUtils.kt"],
+)
+
+kt_jvm_test(
+    name = "jewelBazelScriptsUtilsTest",
+    srcs = [
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/JewelBazelScriptsUtilsTest.kt",
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt",
+    ],
+    associates = [":jewel-bazel-scripts-utils"],
+    deps = [
+        "@rules_kotlin//kotlin/compiler:kotlin-test",
+        "@lib//:junit4",
+        "@lib//:kotlinx-coroutines-test",
+    ],
+    test_class = "org.jetbrains.jewel.scripts.bazel.JewelBazelScriptsUtilsTest",
+)
+
+kt_jvm_test(
+    name = "annotateApiDumpChangesTest",
+    srcs = [
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChangesTest.kt",
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt",
+    ],
+    associates = [":annotate-api-dumps"],
+    deps = [
+        "@rules_kotlin//kotlin/compiler:kotlin-test",
+        "@lib//:junit4",
+        "@lib//:kotlinx-coroutines-test",
+        ":jewel-bazel-scripts-utils",
+        ":bazel-test-utils",
+    ],
+    test_class = "org.jetbrains.jewel.scripts.bazel.AnnotateApiDumpChangesTest",
+)
+
+kt_jvm_test(
+    name = "checkApiDumpsTest",
+    srcs = [
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/CheckApiDumpsTest.kt",
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt",
+    ],
+    associates = [":check-api-dumps"],
+    deps = [
+        "@rules_kotlin//kotlin/compiler:kotlin-test",
+        "@lib//:junit4",
+        "@lib//:kotlinx-coroutines-test",
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+    ],
+    test_class = "org.jetbrains.jewel.scripts.bazel.CheckApiDumpsTest",
+)
+
+kt_jvm_test(
+    name = "compareBranchesTest",
+    srcs = [
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/CompareBranchesTest.kt",
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt",
+    ],
+    # Tests don't need a real workspace root (git calls go through FakeCommandRunner, path-existence checks go
+    # through the injected pathExists), just some valid directory to pass the isDirectory check.
+    env = {"BUILD_WORKSPACE_DIRECTORY": "."},
+    associates = [":compare-branches"],
+    deps = [
+        "@rules_kotlin//kotlin/compiler:kotlin-test",
+        "@lib//:junit4",
+        "@lib//:kotlinx-coroutines-test",
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+    ],
+    test_class = "org.jetbrains.jewel.scripts.bazel.CompareBranchesTest",
+)
+
+kt_jvm_test(
+    name = "validateMavenArtifactsTest",
+    srcs = [
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/ValidateMavenArtifactsTest.kt",
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt",
+    ],
+    associates = [":validate-maven-artifacts"],
+    deps = [
+        "@rules_kotlin//kotlin/compiler:kotlin-test",
+        "@lib//:junit4",
+        "@lib//:kotlinx-coroutines-test",
+        ":jewel-bazel-scripts-utils",
+        ":clikt-dependencies",
+        ":bazel-test-utils",
+    ],
+    test_class = "org.jetbrains.jewel.scripts.bazel.ValidateMavenArtifactsTest",
 )
--- a/platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChanges.kt
+++ b/platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChanges.kt
@@ -1,0 +1,272 @@
+@file:Suppress("RAW_RUN_BLOCKING", "SSBasedInspection")
+package org.jetbrains.jewel.scripts.bazel
+
+import com.github.ajalt.clikt.command.SuspendingCliktCommand
+import com.github.ajalt.clikt.command.main
+import com.github.ajalt.clikt.core.Context
+import com.github.ajalt.clikt.parameters.options.flag
+import com.github.ajalt.clikt.parameters.options.option
+import java.io.File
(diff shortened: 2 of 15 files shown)


platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChanges.kt
```diff
@@ -0,0 +1,272 @@
+@file:Suppress("RAW_RUN_BLOCKING", "SSBasedInspection")
```
DanielSouzaBertoldi, line 1: all the files in the `/scripts/bazel` folder were directly "translated" from their `/scripts/*.main.kts` counterparts.

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CheckApiDumps.kt
```diff
@@ -0,0 +1,105 @@
...
+import java.io.File
+import kotlin.time.Duration.Companion.minutes
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+
+// tests.cmd needs both the JPS module that holds the test class, and the test pattern itself.
+private const val TEST_MODULE = "intellij.platform.testFramework.monorepo.tests"
+private const val TEST_CLASS = "com.intellij.platform.testFramework.monorepo.api.ApiCheckTest"
+
```
DanielSouzaBertoldi, line 18: This script has all the fixes introduced in #3586

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CheckApiDumps.kt
```diff
@@ -0,0 +1,105 @@
...
+internal fun buildApiCheckCommand(): String = "./tests.cmd --module $TEST_MODULE --test $TEST_CLASS"
+
+/**
+ * Builds the shell script that runs [command] and tees its output to [outputFile], from [repoRoot].
+ *
+ * The pipeline's exit status must be [command]'s and not tee's, hence `pipefail`; `errexit` makes a failing `cd`
+ * fatal, too.
+ */
+internal fun buildApiCheckScript(repoRoot: File, outputFile: File, command: String): String =
```
DanielSouzaBertoldi, line 37: This function and `buildApiCheckCommand` above were extracted out of `run` in the original script so that they are unit-testable

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CompareBranches.kt
```diff
@@ -0,0 +1,378 @@
...
+                help =
+                    "The start date for the commit range (yyyy-mm-dd). " +
+                        "If omitted, it will be inferred from the latest release in RELEASE NOTES.md.",
+            )
+            .defaultLazy {
+                val since = getLatestReleaseDate(getJewelRoot() ?: File(".").canonicalFile)
+                if (since.isNullOrBlank()) {
+                    throw UsageError(
+                        "Error: --since is required if RELEASE NOTES.md does not exist or contain a release date."
+                    )
```
DanielSouzaBertoldi, line 62: The original script called both `println` and `exitProcess`.

I opted to use Clikt's `UsageError` instead since it renders a usage synopsis for free

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/JewelBazelScriptsUtils.kt
```diff
@@ -137,7 +137,7 @@ suspend fun isDirectoryGitRepo(directory: File, runner: CommandRunner = DefaultC
  * @return true if the branch exists, false otherwise.
  */
 suspend fun branchExists(branch: String, directory: File, runner: CommandRunner = DefaultCommandRunner): Boolean =
-    runner("git rev-parse --verify $branch", directory).isSuccess
+    runner("git rev-parse --verify $branch", directory, exitOnError = false).isSuccess
```
DanielSouzaBertoldi, line 140: Asked Claude to review my PR and it caught this subtle bug:

`git rev-parse --verify <branch>` exits non-zero when a branch does not exist. Without the `exitOnError = false`, the command runner throws before `branchExists` can ever return `false`, making the caller's own "Branch not found" error message unreachable 🤯

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/JewelBazelScriptsUtils.kt
```diff
@@ -215,7 +215,14 @@ private suspend fun runCommand(
             if (!streamOutput) {
                 CompletableFuture.supplyAsync { process.inputStream.bufferedReader().readText() }
             } else null
-        val exitCode = process.waitFor()
+
+        val finishedInTime = process.waitFor(timeoutAmount.inWholeSeconds, TimeUnit.SECONDS)
```
DanielSouzaBertoldi, line 219: Claude also noticed I forgot to add a timeout for `process.waitFor` like the original script does. Now we use the timeout both in here and on line 226 below

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CompareBranches.kt
```diff
@@ -0,0 +1,378 @@
...
+import kotlinx.coroutines.runBlocking
+
+// Run it like: bazel run //platform/jewel/build-scripts/bazel:compareBranches -- branch1 branch2
+fun main(args: Array<String>) {
+    runBlocking { CompareBranchesCommand().main(args) }
+}
+
+internal class CompareBranchesCommand(
+    private val commandRunner: CommandRunner = DefaultCommandRunner,
+    private val pathExists: (File) -> Boolean = File::exists,
```
DanielSouzaBertoldi, line 26: This is a test seam only. Production always uses the real `File` method. We should only swap this only for test fakes :)

Every other command that has this sort of parameter in the constructor is for this same exact same reason

platform/jewel/build-scripts/bazel/BUILD.bazel
```diff
@@ -56,3 +114,117 @@ java_binary(
...
+)
+
+# Tests
+kt_jvm_test(
+    name = "jewelBazelScriptsUtilsTest",
+    srcs = [
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/JewelBazelScriptsUtilsTest.kt",
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt",
+    ],
+    associates = [":jewel-bazel-scripts-utils"],
```
DanielSouzaBertoldi, line 156: We use `associate` here so that the test module can have access to `internal` functions :)

platform/jewel/build-scripts/bazel/BUILD.bazel
```diff
@@ -56,3 +114,115 @@ java_binary(
...
+
+kt_jvm_test(
+    name = "compareBranchesTest",
+    srcs = [
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/CompareBranchesTest.kt",
+        "src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt",
+    ],
+    # Tests don't need a real workspace root (git calls go through FakeCommandRunner, path-existence checks go
+    # through the injected pathExists), just some valid directory to pass the isDirectory check.
+    env = {"BUILD_WORKSPACE_DIRECTORY": "."},
```
DanielSouzaBertoldi, line 207: `CompareBranchesCommand.run()` unconditionally reads `BUILD_WORKSPACE_DIRECTORY`, which Bazel only ever sets for `bazel run`, never `bazel test`. Without this `env` override, every test in that target would crash immediately on startup, not because of anything the test itself does, but because the target under test assumes a `bazel run` context.

platform/jewel/scripts/utils.main.kts
```diff
@@ -388,7 +388,7 @@ suspend fun isDirectoryGitRepo(directory: File): Boolean =
  * @return true if the branch exists, false otherwise.
  */
 suspend fun branchExists(branch: String, directory: File): Boolean =
-    runCommand("git rev-parse --verify $branch", directory).isSuccess
+    runCommand("git rev-parse --verify $branch", directory, exitOnError = false).isSuccess
```
DanielSouzaBertoldi, line 392: [Same fix here](https://github.com/JetBrains/intellij-community/pull/3602#discussion_r3735636875) in the gradle version. Until we are ready to fully migrate to bazel, might as well fix what's wrong on the gradle side too

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChangesTest.kt
```diff
@@ -0,0 +1,514 @@
+package org.jetbrains.jewel.scripts.bazel
```
DanielSouzaBertoldi, line 1: hell yeah now we got tests for our scripts folks 😎 ofc sometimes we can't fully test a script but at least we can check input -> output and test specific functions

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChangesTest.kt
```diff
@@ -0,0 +1,514 @@
...
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+import kotlinx.coroutines.test.runTest
+import org.junit.After
+import org.junit.Test
+
+private val testFile = File("api-dump-test.txt")
+
+class AnnotateApiDumpChangesTest {
+    private val tmpDir = File(System.getenv("TEST_TMPDIR") ?: System.getProperty("java.io.tmpdir"))
```
rock3r: This is a bit risky: when `TEST_TMPDIR` isn't set — e.g., someone runs this test from the IDE instead of Bazel — `tmpDir` becomes the *system* temp dir, and `tearDown()` then nukes everything in it, other processes' files included.

Can we use a dedicated subdirectory and only delete that?

```suggestion
    private val tmpDir =
        File(System.getenv("TEST_TMPDIR") ?: System.getProperty("java.io.tmpdir"))
            .resolve("annotate-api-dump-changes-test").apply { mkdirs() }
```

Same thing in `ValidateMavenArtifactsTest`.

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/FakeCommandRunner.kt
```diff
@@ -0,0 +1,20 @@
...
+
+class FakeCommandRunner(private val handler: (command: String) -> CmdResult = { CmdResult.Success("") }) :
+    CommandRunner {
+    val calls = mutableListOf<String>()
+
+    override suspend fun invoke(
+        command: String,
+        workingDir: File?,
+        timeoutAmount: Duration,
+        exitOnError: Boolean,
```
rock3r, line 14: The fake ignores `exitOnError`, while `DefaultCommandRunner` throws on failure when it's true. That means tests can't distinguish a call that *expects* failure from one that doesn't — for example, the `branchExists` regression test in `JewelBazelScriptsUtilsTest` would still pass even if someone removed `exitOnError = false` again, which is the very bug this PR fixes.

Can we make the fake model throw when the handler returns `Failure` and `exitOnError` is true (closest to the real behaviour), or just record the flag alongside each call so tests can assert it? Either works, up to you.

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChanges.kt
```diff
@@ -0,0 +1,272 @@
...
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.runBlocking
+
+private val chunkHeaderRegex = "^@@ \\-([0-9]+)(?:,[0-9]+)? \\+([0-9]+)".toRegex()
+
+fun main(args: Array<String>) {
+    runBlocking { AnnotateApiDumpChangesCommand().main(args) }
+}
+
+class AnnotateApiDumpChangesCommand(private val commandRunner: CommandRunner = DefaultCommandRunner) :
```
rock3r: Nit: missing `internal` here — the other three commands are all `internal`, and the tests reach this one via `associates` anyway.

platform/jewel/build-scripts/bazel/BUILD.bazel
```diff
@@ -1,5 +1,15 @@
-load("@rules_java//java:defs.bzl", "java_binary")
-load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_import", "kt_jvm_library")
+load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library", "kt_jvm_import", "kt_jvm_test")
```
rock3r: Nit/question: the load order changed here. This file is hand-written (no `.iml`, generator never touches it) and was sorted "correctly" before. Was the reorder intentional? If it's just editor noise, can we revert to sorted order?

Side question, should these scripts get an iml module (that does not ship) too, so the IDE does code completion and indexing? Or is it unimportant since the Bazel project would include them?

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChangesTest.kt
```diff
@@ -0,0 +1,514 @@
...
+                .trimIndent()
+
+        val result = processDiff(apiDiff, testFile, true)
+
+        assertEquals(4, result.size)
+    }
+
+    @Test
+    fun `processDiff breaking change found should have correct data`() {
+        val experimentalFlag = listOf(true, false).random()
```
rock3r: Nit: nondeterministic test input — each run only exercises one of the two branches, picked by coin flip. If this assertion ever fails it'll look flaky and only reproduce ~50% of the time. Can we assert both values deterministically (two tests, or a small loop)? Given there are dedicated `formatLog` tests for both stable and experimental below, even a fixed `false` here would do.

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/JewelBazelScriptsUtilsTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+
+class JewelBazelScriptsUtilsTest {
+    @Test
+    fun `DefaultCommandRunner aborts a command that exceeds timeoutAmount instead of blocking until it exits`() =
+        runTest {
+            var elapsed = 0L
+
+            assertFailsWith<IllegalStateException> {
```
rock3r, line 22: This won't work on Windows — there's no `sleep` there, so it would throw an  `IOException` instead of an `IllegalStateException`.

Now, we've never really cared about running these scripts on Windows, so it's fine; I'd just make the Unix-only assumption explicit with a short comment, or an explicit error if ran on Windows.

rock3r (changes requested):

platform/jewel/build-scripts/bazel/BUILD.bazel
DanielSouzaBertoldi: I've been juggling a bunch of git stashes so some of them were in sorted order while others weren't :(

Also, honestly I'm not a fan of creating a IML file for this new module. It is supposed to be "pure" Bazel, given that the end goal is to get "rid" of JPS in the future.

Besides, this repo has been broken on JPS for a really long time now, so I doubt folks are that concerned with the JPS side of things in IJC 😬

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChangesTest.kt
DanielSouzaBertoldi: Wow, had no idea. Thanks for raising this concern!

platform/jewel/build-scripts/bazel/BUILD.bazel
shalupov: I think it's fine to do it without imls. JPS model is still used in community, but compilation by *JPS* is broken. BUILD.bazel scripts style should be controlled (and it is for other things) with buildifier

shalupov (approved):

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChangesTest.kt
DanielSouzaBertoldi: Created a [BazelTestUtils file](https://github.com/JetBrains/intellij-community/pull/3602/changes#diff-9018e1a3feb418eb003d389f03b3e45f7bf6605b27a16afcfa22c979c0a18594) that both tests now use

platform/jewel/build-scripts/bazel/src/test/kotlin/org/jetbrains/jewel/scripts/bazel/JewelBazelScriptsUtilsTest.kt
DanielSouzaBertoldi: Added `Assume.assumeFalse()` to skip this test on Windows machines

mfilippov (approved):
LGTM

rock3r (approved):

DanielSouzaBertoldi:
Ready to merge

wellingtoncosta (approved):

PR #3606, opened 2026-08-10
IDEA-254862 Gradle: accept transformed auxiliary artifacts

NichtStudioCode:
This PR allows transformed source and javadoc artifacts to be attached during project import.

Currently, `AuxiliaryArtifactResolverImpl.classify()` checks for Gradle's internal `ModuleComponentArtifactIdentifier`. Transform outputs use `TransformedComponentFileArtifactIdentifier`, so valid artifacts are discarded.

This PR uses `ModuleComponentIdentifier` instead, which works with transformed artifacts. Additionally, it's also not internal Gradle API.

This fixes the first problem that I mentioned in my [youtrack comment](https://youtrack.jetbrains.com/issue/IDEA-254862/Gradle-import-and-auto-download-of-sources-does-not-respect-artifact-transformers#focus=Comments-27-14232476.0-0) on [IDEA-254862](https://youtrack.jetbrains.com/issue/IDEA-254862) and passes the `transformed-source-artifact-id.zip` reproducer attached there.

I've also added a unit test for this. It currently uses `java.lang.reflect.Proxy` instead of mockito because mockito is not on the test classpath for this module, but I can change it to use mockito if that is required.


honourosis:
Looks good to me. Thank you for your contribution.

honourosis (approved):

PR #3607, opened 2026-08-10
    Use safer regex for end of line patterns

karimai-gg:
In order to avoid ending up with unnecessary branching logic that is very unoptimal for longer stacktraces.
this impacts Windows specifically that has \r\n as EOL, so for each line, we would have 2 patterns to evaluate against that would match the previous regex.
This blows the complexity to an exponential level.
    
Fixes IDEA-392543

### DIFF
--- a/plugins/gradle/src/org/jetbrains/plugins/gradle/issue/GradleIssueFailureImpl.kt
+++ b/plugins/gradle/src/org/jetbrains/plugins/gradle/issue/GradleIssueFailureImpl.kt
@@ -83,8 +83,7 @@ private object GradleIssueFailureLocationResolver {
 
   private val ERROR_LOCATION_IN_FILE_PATTERN = Pattern.compile("(?:Build|Settings) file '([^']*)' line: (\\d+)")
   private val ERROR_IN_FILE_PATTERN = Pattern.compile("(?:Build|Settings) file '([^']*)'")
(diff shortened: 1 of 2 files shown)


plugins/gradle/src/org/jetbrains/plugins/gradle/issue/GradleIssueFailureImpl.kt
nskvortsov: On the short lines this will be a tiny win. But I will keep it.

nskvortsov (approved):

nskvortsov:
Thank you, running the CI now.

plugins/gradle/src/org/jetbrains/plugins/gradle/issue/GradleIssueFailureImpl.kt
karimai-gg: I found that actually this .* does eat the \r character at EOL ( and I have a suspicion even the ones above that have [^'] could go on a multiline matching). I think it's best to replace .* by [^\r\n] so we make sure if nothing fits within the same line then we stop the matching, and similarily do [^'\r\n] for other patterns.

Our benchmarks are still failing with my changes, which means we are still consuming a lot of res doing these ops on multiline stacktraces. 

I ma tyring to optimise these with the suggestions above and will let you know.

PR #3610, opened 2026-08-12
[JEWEL-1396] Keep Escape in Jewel ComboBox popups

rock3r:
Escape closes a `SpeedSearchableComboBox` popup before users can clear the active speed-search query. In standalone custom `JDialogRenderer` popups, global AWT listeners can compete for the same native Escape event: non-focusable parent/child popups can swallow it, while focusable popups can miss events delivered to their owner window.

## Changes

* `SpeedSearchableComboBox` gives speed search first claim on Escape: the first press hides and clears the query, and the second press closes the popup.
* `JDialogRenderer` dispatches native key events by owning popup dialog and forwards owner-window events for focusable popups.
* Native key events handled by popup content or used for dismissal are consumed before they can reach host dialogs.
* `SpeedSearchArea` omits the renderer dismissal callback when `dismissOnLoseFocus = false`.
* `PopupRenderer` has explicit `windowShape` overload implementations in the default, JBPopup, and test renderers. `JDialogRenderer` applies native window shapes; other renderers explicitly do not.
* Component and renderer contract coverage includes ComboBox Escape propagation, speed-search dismissal callbacks, and `SpeedSearchableComboBox` behavior in the showcase.
* An opt-in Spectre 0.5.0 headful lane in `int-ui-standalone-tests` uses a real Jewel standalone window, synthetic AWT injection, and `moveTo()` pointer routing to cover hovered ComboBox dismissal, focusable menu dismissal through the owner window, and the two-step speed-search Escape sequence.

## Screen recording

 Before | After
 --- | ---
 <video src="https://github.com/user-attachments/assets/28d7fae1-933b-4b67-bb83-a043de832de6" /> | <video src="https://github.com/user-attachments/assets/c27ac9a0-6ee5-451c-bed7-ce89c3fd1372" />

## Spectre test scope

Spectre coverage is intentionally limited to the dedicated Gradle `:int-ui:int-ui-standalone-tests:spectreTest` task in this PR. JEWEL-1390 covers Bazel integration and broader standalone E2E use. Bridge/IDE integration is TBD; Spectre already builds on the IDE starter, so it should fit that path, but it is outside this change's scope. JEWEL-1397 tracks the broader standalone GUI E2E infrastructure.

## Validation

* `platform/jewel ./gradlew check --continue --no-daemon`
* `platform/jewel ./gradlew detekt detektMain detektTest --continue --no-daemon`
* `platform/jewel ./gradlew checkMetalavaApi --continue --no-daemon`
* `./bazel.cmd test //platform/testFramework/monorepo:monorepo-tests_test --test_filter=com.intellij.platform.testFramework.monorepo.api.ApiCheckTest`
* Focused Bazel regressions for `SpeedSearchAreaDismissRequestTest`, `ComboBoxTest`, and `SpeedSearchableComboBoxTest`
* Affected Bazel builds for `ui`, `ide-laf-bridge`, `int-ui-standalone`, and related test modules
* `platform/jewel ./gradlew :int-ui:int-ui-standalone-tests:spectreTest --no-daemon`
* Manual standalone dialog and showcase smoke testing

### Bug fixes

* The first Escape press in a speed-searchable ComboBox clears the speed-search query; the second press closes the popup.
* Standalone custom popups route Escape to the appropriate popup renderer, preserving focusable menu behavior and preventing Escape from leaking to host dialogs.

### DIFF
--- a/platform/jewel/gradle/libs.versions.toml
+++ b/platform/jewel/gradle/libs.versions.toml
@@ -22,6 +22,7 @@ ktfmtGradlePlugin = "0.23.0"
 metalava = "1.0.0-alpha13"
 jsoup = "1.21.2"
 mockk = "1.14.5"
+spectre = "0.5.0"
 
 [libraries]
 assertj-core = { module = "org.assertj:assertj-core", version.ref = "assertJ"}
@@ -52,6 +53,8 @@ ktor-client-java = { module = "io.ktor:ktor-client-java", version = "3.0.3" }
 metalava = { module = "com.android.tools.metalava:metalava", version.ref = "metalava" }
 jsoup = { module = "org.jsoup:jsoup", version.ref = "jsoup" }
 mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
+spectre-core = { module = "dev.sebastiano.spectre:spectre-core", version.ref = "spectre" }
+spectre-testing = { module = "dev.sebastiano.spectre:spectre-testing", version.ref = "spectre" }
 
 # Plugin libraries for build-logic's convention plugins to use to resolve the types/tasks coming from these plugins
 detekt-gradlePlugin = { module = "dev.detekt:detekt-gradle-plugin", version.ref = "detekt" }
--- a/platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
+++ b/platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
@@ -1,0 +1,212 @@
+@file:OptIn(ExperimentalJewelApi::class)
+
+package org.jetbrains.jewel.intui.standalone.popup
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.awt.ComposeWindow
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.window.application
+import dev.sebastiano.spectre.core.ComposeAutomator
+import dev.sebastiano.spectre.core.RobotDriver
+import dev.sebastiano.spectre.testing.runSpectreTest
+import java.awt.event.KeyEvent
+import java.util.concurrent.atomic.AtomicReference
+import kotlin.concurrent.thread
+import kotlin.test.assertTrue
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.delay
+import org.jetbrains.jewel.foundation.ExperimentalJewelApi
+import org.jetbrains.jewel.foundation.JewelFlags
+import org.jetbrains.jewel.intui.standalone.theme.IntUiTheme
+import org.jetbrains.jewel.intui.standalone.window.Window as JewelWindow
+import org.jetbrains.jewel.ui.component.ComboBox
+import org.jetbrains.jewel.ui.component.DefaultButton
+import org.jetbrains.jewel.ui.component.PopupManager
+import org.jetbrains.jewel.ui.component.PopupMenu
+import org.jetbrains.jewel.ui.component.SpeedSearchArea
+import org.jetbrains.jewel.ui.component.Text
+import org.jetbrains.jewel.ui.component.rememberSpeedSearchState
+import org.jetbrains.jewel.ui.component.search.SpeedSearchableComboBox
+import org.junit.jupiter.api.Test
+
+// Note: these tests run intentionally Gradle-only until JEWEL-1390
+class CustomPopupRendererSpectreTest {
+    @Test
+    fun `escape closes a hovered combo box`(): Unit = runSpectreTestWithCustomPopupRenderer {
+        val app = SpectreTestApplication()
+        app.start()
+        try {
+            val automator = ComposeAutomator.inProcess(RobotDriver.synthetic(app.awaitWindow()))
+
+            automator.click(automator.waitForNode(tag = REGULAR_COMBO_TAG))
+            automator.waitForNode(tag = REGULAR_POPUP_VISIBLE_TAG, text = "true")
+
+            // The pointer remains inside the ComboBox after click, reproducing the hover path that
+            // previously let the popup's JDialogRenderer consume Escape as a no-op dismissal.
+            automator.pressKey(KeyEvent.VK_ESCAPE)
+            automator.waitForNode(tag = REGULAR_POPUP_VISIBLE_TAG, text = "false")
+        } finally {
+            app.stop()
+        }
+    }
+
+    @Test
+    fun `escape closes a focusable menu when the owner window receives the key`(): Unit =
+        runSpectreTestWithCustomPopupRenderer {
+            val app = SpectreTestApplication()
+            app.start()
+            try {
+                val automator = ComposeAutomator.inProcess(RobotDriver.synthetic(app.awaitWindow()))
+
+                val menuButton = automator.waitForNode(tag = MENU_BUTTON_TAG)
+                automator.click(menuButton)
+                automator.waitForNode(tag = MENU_VISIBLE_TAG, text = "true")
+
+                // Spectre routes synthetic key events to the window under its injected pointer. Move that pointer back
+                // to the owner window to cover focusable popups that have not taken native focus yet.
+                automator.moveTo(menuButton)
+                automator.waitForIdle()
+
+                automator.pressKey(KeyEvent.VK_ESCAPE)
+                automator.waitForNode(tag = MENU_VISIBLE_TAG, text = "false")
+            } finally {
+                app.stop()
+            }
+        }
+
+    @Test
+    fun `escape prioritizes speed search when dismiss on lose focus is disabled`(): Unit =
+        runSpectreTestWithCustomPopupRenderer {
+            val app = SpectreTestApplication()
+            app.start()
+            try {
+                val automator = ComposeAutomator.inProcess(RobotDriver.synthetic(app.awaitWindow()))
+
+                automator.click(automator.waitForNode(tag = SPEED_SEARCH_COMBO_TAG))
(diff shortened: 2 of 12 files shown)


nebojsa-vuksic:
The default-renderer part looks good - deriving the two-step behaviour from `hideSearch()` returning `false` when there is nothing left to clear is neat.

My concerns are all in the standalone `JDialogRenderer` path (`jewel.customPopupRender=true`), and I think they are one issue rather than four: `onDismissRequest`'s nullability is answering two different questions. The renderer reads `dismissRequest != null` as "does this popup care about dismissal?", but the call sites use the body to answer "should it dismiss right now?" - `if (dismissOnLoseFocus)` in `SpeedSearchArea`, `if (!chevronHovered)` in `ComboBox`. So a popup that decided not to dismiss still counts as `dismissed`, and we `event.consume()` anyway.

Since this is a global `AWTEventListener` on the `Toolkit`, that consume stops more than the host dialog - it also stops the focused component's own Compose handlers, including the `onPreviewKeyEvent` this PR adds to `ListComboBoxImpl`, so the headline behaviour never runs in this renderer. Would making the callback report its outcome be worth it, e.g. `onDismissRequest: (() -> Boolean)?` and consuming only when it returns `true`? That covers the `SpeedSearchArea` and `ComboBox` cases directly. Details inline.

Two smaller things:

1. the description says we preserve `dismissOnLoseFocus = false` behaviour, but with the callback always non-null Escape gets consumed while nothing is hidden, so it becomes a dead key there
2. the new tests never set `JewelFlags.useCustomPopupRenderer`, which defaults to `false`, so the whole `JDialogRenderer` half is untested. I see JEWEL-1397 covers the native E2E lane - could we still add a unit-level test that flips the flag, given that is where these regressions are?

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
```diff
@@ -421,7 +425,10 @@ private fun SpeedSearchInput(
             }
         }
 
-    Popup(popupPositionProvider = rememberComponentRectPositionProvider(anchor, alignment)) {
+    Popup(
+        popupPositionProvider = rememberComponentRectPositionProvider(anchor, alignment),
+        onDismissRequest = { if (dismissOnLoseFocus) speedSearchState.hideSearch() },
```
nebojsa-vuksic: Are we sure this preserves the `dismissOnLoseFocus = false` behaviour the description mentions? The body does, but the callback is non-null either way and the renderer keys off existence: `dismissed = ... && dismissRequest != null`, then `event.consume()`.

So with `dismissOnLoseFocus = false` in a standalone app, Escape invokes a no-op lambda, still counts as dismissed, and gets consumed - the overlay stays, the component's `onPreviewKeyEvent(::processKeyEvent)` never runs, and the host dialog does not see it either. Escape ends up doing nothing, where before it cleared the query.

Line 195 in this file already checks the flag outside the callback. Could we do the same here?

```suggestion
        onDismissRequest = if (dismissOnLoseFocus) ({ speedSearchState.hideSearch() }) else null,
```

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -418,16 +418,21 @@ private fun JPopupImpl(
...
-                    if (!dialog.isVisible) return@AWTEventListener
+                    if (event.isConsumed || !dialog.isVisible) return@AWTEventListener
 
                     val composeEvent = event.toComposeKeyEvent()
 
                     val consumed =
                         currentOnPreviewKeyEvent?.invoke(composeEvent) == true ||
                             currentOnKeyEvent?.invoke(composeEvent) == true
+                    val dismissRequest = currentOnDismissRequest
+                    val dismissed = !consumed && composeEvent.isDismissRequest() && dismissRequest != null
```
nebojsa-vuksic, line 433: I don't think the first-Escape-clears-search behaviour actually reaches this renderer. `ComboBoxImpl` goes through `PopupContainer`, which forwards only `onDismissRequest` and no `onPreviewKeyEvent`/`onKeyEvent`, so `consumed` is always `false` here and the first Escape sets `dismissed` and closes the whole popup. The `consume()` below then keeps the new `onPreviewKeyEvent` on `ListComboBoxImpl` from ever seeing it.

It also only resolves this way because the combo popup registers its listener before the speed-search overlay, and `AWTEventMulticaster` notifies in registration order - if that order flipped, the intended behaviour would work. Could we base the decision on something other than "a dismiss callback exists"?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -418,16 +418,21 @@ private fun JPopupImpl(
                     }
                 }
                 is AWTKeyEvent -> {
-                    if (!dialog.isVisible) return@AWTEventListener
+                    if (event.isConsumed || !dialog.isVisible) return@AWTEventListener
```
nebojsa-vuksic: Question rather than a claim here, since I reasoned it from the code and did not run it.

With the new `consume()` below, this guard makes the first-registered popup win over later ones. `PopupMenuImpl` (`Menu.kt:301`) and `Submenu` (`Menu.kt:1097`) pass the same `handlePopupMenuOnKeyEvent`, both dialogs are visible while a submenu is open, and that handler returns `true` for Down while calling `focusManager.moveFocus(Next)` in its own scene. If the parent's listener is notified first, it would handle Down in the parent scene and consume, and the submenu's listener would return early here - so arrow navigation inside a submenu would stop moving the visible selection.

Does that ordering hold in practice, or is something keeping the parent's listener from firing while the submenu has focus?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/popup/JDialogRenderer.kt
```diff
@@ -418,16 +418,21 @@ private fun JPopupImpl(
...
                             currentOnKeyEvent?.invoke(composeEvent) == true
+                    val dismissRequest = currentOnDismissRequest
+                    val dismissed = !consumed && composeEvent.isDismissRequest() && dismissRequest != null
 
-                    if (!consumed && composeEvent.isDismissRequest()) {
-                        currentOnDismissRequest?.invoke()
+                    if (dismissed) {
+                        dismissRequest.invoke()
+                    }
+                    if (consumed || dismissed) {
```
nebojsa-vuksic, line 438: `dismissed` only means a dismiss callback existed and the key was a dismiss request, not that we actually dismissed - but we consume either way. `ComboBox` passes `onDismissRequest = { if (!chevronHovered) popupManager.setPopupVisible(false) }`, so pressing Escape with the pointer over the chevron leaves the popup open and still swallows the Escape, and the host dialog can no longer be closed with it.

Would `onDismissRequest: (() -> Boolean)?`, consuming only when it returns `true`, work here? That would cover the `SpeedSearchArea` case too.

rock3r:
>the new tests never set `JewelFlags.useCustomPopupRenderer`, which defaults to false, so the whole `JDialogRenderer` half is untested. I see JEWEL-1397 covers the native E2E lane - could we still add a unit-level test that flips the flag, given that is where these regressions are?

No, that is nonsense, the separate issue exists exactly because we can't currently run headful test, and the flag does absolutely nothing in headless tests (`jps_test`). The Bazel build only allows running headful tests if using JetBrains Starter, which is not an option for Jewel standalone. Since that is a larger change, this PR has to ship the `JDialogRenderer` change without tests, which will be added by JEWEL-1397 as per its description. A test that flips a flag that is not used by anyone is, as Claude would say, "testing theatre".

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
rock3r: I'll check, this may have been a typo.

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
```diff
@@ -0,0 +1,210 @@
...
+            automator.pressKey(KeyEvent.VK_ESCAPE)
+            automator.waitForNode(tag = MENU_VISIBLE_TAG, text = "false")
+        } finally {
+            app.stop()
+        }
+    }
+
+    @Test
+    fun `escape prioritizes speed search when dismiss on lose focus is disabled`(): Unit = runSpectreTest {
+        assertTrue(JewelFlags.useCustomPopupRenderer, "spectreTest must enable JDialogRenderer")
```
wellingtoncosta: I'd say to create a helper function like `runSpectreTestWithCustomPopupRenderer` just to embbed this check since you did it in the other tests abve.

wellingtoncosta (approved):

platform/jewel/gradle/libs.versions.toml
```diff
@@ -22,6 +22,7 @@ ktfmtGradlePlugin = "0.23.0"
 metalava = "1.0.0-alpha13"
 jsoup = "1.21.2"
 mockk = "1.14.5"
+spectre = "0.5.0"
```
DanielSouzaBertoldi, line 25: Oh nice 😎 

Question: shouldn't we wire `spectreTest` into `check`? Otherwise the tests you created won't run "by default" no? If that's intentional, then maybe updating Claude's skill to also run `spectreTest` along with `detekt` and `detektMain` would be enough as well.

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/SpeedSearchArea.kt
DanielSouzaBertoldi: Shouldn't we update this KDoc to also note that now it controls whether to lose focus when a user presses the escape key? Clicking outside the search bar would also dismiss it. Reinforce all these hidden behaviors would benefit users :D

DanielSouzaBertoldi (approved):
Just a couple of questions, no blockers. Nice stuff!

platform/jewel/gradle/libs.versions.toml
rock3r: Intended for now, will be properly wired in Bazel checks (and not in Gradle) in JEWEL-1390. I'm ok with this not running on the CI in the meantime, it was mostly to prove the point right now that things do work :D JEWEL-1390 will be the next step setting up the Spectre dependency and infra in Bazel, using these tests as the first example. Will work on that asap after this ships.

DanielSouzaBertoldi (approved):
Let's 🚢 this

rock3r:
Ready to merge 😎 since I've addressed all comments and there was nothing major to change

PR #3612, opened 2026-08-14
[IJPL-252958] Fix Compose Resources string fold/preview.

sonic-onkaringale:
When a Compose Multiplatform string resource (`Res.string.xxx`) is referenced in Kotlin code, the IDE folds the reference and shows a preview of its resolved text. 

This preview is generated by `ComposeResourcesFoldingBuilder.getResourcePlaceholderText()`, which iterates `psiElements` (one per locale variant of the resource) and returns the text of the *first* non-empty match with no preference for the base/default (unqualified `values/`) resource. 

In practice, this list is not ordered with the base `values/strings.xml` first, so the folded preview ends up showing whatever locale-qualified variant happens to sort first (e.g. `values-ar` before `values` in the resource item list), even when the base `values/strings.xml` holds the actual default text.

**Steps to reproduce**
1. Create a Compose Multiplatform project with composeResources.
2. Add `res/values/strings.xml` with a string key, e.g. `<string name="foo">Hello</string>`.
3. Add `res/values-ar/strings.xml` with the same key translated, e.g. `<string name="foo">مرحبا</string>`.
4. In Kotlin code, reference `stringResource(Res.string.foo)`.
5. Observe the folded inline preview shows the Arabic text instead of "Hello".

**Expected behavior**
The folded preview should prefer the text from the unqualified `values/` (default) resource folder when one exists, falling back to whatever is available otherwise.

***I verified this locally by compiling and patching the code, the patch resolves the issue and the fold now correctly shows the default locale's text.***

**Environment**
- IDE: IntelliJ IDEA 2026.2.1
- Compose Multiplatform plugin version: 263.20260731.0
- OS: macOS 26
- Project uses composeResources with 16 locale folders (values, values-ar, values-de, ...)
Screenshot 2026-08-14 at 11 02 51 AM
Screenshot 2026-08-14 at 10 59 51 AM

### DIFF
--- a/plugins/compose/intellij.compose.ide.plugin.resources/src/com/intellij/compose/ide/plugin/resources/folding/ComposeResourcesFoldingBuilder.kt
+++ b/plugins/compose/intellij.compose.ide.plugin.resources/src/com/intellij/compose/ide/plugin/resources/folding/ComposeResourcesFoldingBuilder.kt
@@ -62,19 +62,37 @@ private fun KtSimpleNameExpression.getFoldingDescriptor(): FoldingDescriptor? {
 }
 
 private fun getResourcePlaceholderText(psiElements: List<PsiElement>): String? {
-  for (psiElement in psiElements) {
-    val xmlTag = psiElement.parent.parent as? XmlTag ?: continue
-    val textValue = when (xmlTag.name) {
-      ResourceType.STRING.typeName -> xmlTag.value.text.trim()
-      ResourceType.PLURAL_STRING.typeName -> {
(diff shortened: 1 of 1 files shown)


sonic-onkaringale:
Recorrected my commit.
Eliminated Eager Sorting ( O(N log(N)) to Short-Circuit O(N) )

sonic-onkaringale:
Reference for YouTrack : [IJPL-252958](https://youtrack.jetbrains.com/issue/IJPL-252958)

PR #3613, opened 2026-08-15
KTIJ-26409 : Highlight overridden/implemented functions by type

ApoloApps:
Fixes https://youtrack.jetbrains.com/issue/KTIJ-26409/Highlight-overridden-implemented-functions-by-type

https://github.com/user-attachments/assets/bef8a073-4b28-4dac-a860-15c94acb54fb


vladimirdolzhenko:

I've spotted some behavior inconsistency:
Let's have the following code:
```
interface I1 {
    fun foo1()
}
interface I2 {
    fun foo2()
}
class ClassX: I1, I2 {
    override fun <caret>foo1() {
    }

    override fun foo2() {
//        foo1()
    }
}
```

In such case `I1` is highlighted as intended.
image


Now, let's uncomment `foo1()` call in `foo2()` and move the caret back to `foo1`:

image


it seems it is flaky and cut-paste class code make it work again:

image


Another problem, when we move cursor inside `foo2` method on `foo1()` call:

image


it highlights usages of `foo1` - that is correct, but in this case super-classes/interfaces should not be highlighted.

Could you please investigate and fix these issues ?

vladimirdolzhenko (changes requested):

I've spotted some behavior inconsistency:
Let's have the following code:
```
interface I1 {
    fun foo1()
}
interface I2 {
    fun foo2()
}
class ClassX: I1, I2 {
    override fun <caret>foo1() {
    }

    override fun foo2() {
//        foo1()
    }
}
```

In such case `I1` is highlighted as intended.
image


Now, let's uncomment `foo1()` call in `foo2()` and move the caret back to `foo1`:

image


it seems it is flaky and cut-paste class code make it work again:

image


Another problem, when we move cursor inside `foo2` method on `foo1()` call:

image


it highlights usages of `foo1` - that is correct, but in this case super-classes/interfaces should not be highlighted.

Could you please investigate and fix these issues ?

ApoloApps:
>Now, let's uncomment foo1() call in foo2() and move the caret back to foo1:

I can't seem to reproduce your 1st issue with this PR branch. I don't think it's related to my code but rather some cache issue I can't seem to reproduce.

The second issue is reproduced and I'll be commiting soon

https://github.com/user-attachments/assets/9378da45-915c-4235-bab6-c45ec2cfeeef


The second issue is reproduced and I'll be commiting soon

PR #3614, opened 2026-08-16
PY-91641 Expose generated type information through the Python IDE

starnotes-xj:
Add a generic "Generate/Update Python Type Information" action backed by the experimental `Pythonid.typeInformationGenerator` extension point.

The Python IDE owns only the reusable orchestration:

- discover the project/module interpreter;
- select the first applicable companion-provider on a background thread;
- ask for confirmation and run generation with cancellable progress;
- refresh the SDK, packages, skeletons, and daemon analysis after success.

The provider contract now lives in the public Python OpenAPI module. Providers own their dependency discovery and generation details, and may be supplied by companion plugins. Applicability checks are documented as lightweight and side-effect free. Provider-specific failures are logged without exposing arbitrary process output as UI text.

This update intentionally removes the SageMath provider and the Sage `.sage` preparse action from JetBrains core. Sage-specific behavior remains in the companion plugin https://github.com/starnotes-xj/sage-ide-support, while generated Sage type data and preparsing remain in the external `sage-pycharm-stubgen` tool. This keeps Python core independent of one runtime ecosystem and leaves the new SPI reusable for other generated type-information providers.

The generic action is hidden when no provider is registered, so a plain Python installation does not show an inert menu entry. The companion plugin can register its provider against the OpenAPI contract when that integration is implemented.

Issue: https://youtrack.jetbrains.com/issue/PY-91641

Validation: XML parsing, message-key validation, Bazel source/dependency queries, and `git diff --check` pass. Full Bazel compilation is currently blocked before Kotlin compilation by the local external LLVM dependency cache: `libcxxabi`/`libunwind` source globs are empty.


starnotes-xj:
Follow-up idea for the EP design (no code change needed in this PR —
offering a contract for what a `typeInformationGenerator` returns).

The engine already writes a stable, machine-readable report next to the
generated stubs (`<output>/generation-report.json`).  It would make the EP
more complete for reviewers/users if this file were part of the EP
contract, so PyCharm can surface results after "Generate/Update":

```jsonc
{
  "sage_version": "10.9",
  "sage_package": "<env>/lib/python3.13/site-packages/sage",
  "output_root": "<env>/sage_typings",
  "discovered": 576,          // .pyx modules found
  "generated": 576,           // stubs written
  "failed": 0,
  "factory_inferred": [       // "name -> dotted return type", runtime-verified
    "GF -> sage.rings.finite_rings.finite_field_base.FiniteField",
    "..."
  ],
  "factory_unresolved": [],   // factories whose static type is not importable
  "dynamic_unresolved": [],   // sage.all exports that fell back to Any
  "docstrings": { "modules_processed": 175, "docstrings_attached": 1555,
                  "return_types_added": 440, "curated_applied": 681, "...": "..." },
  "failures": []
}
```

Suggested minimal addition to `PyTypeInformationGenerator` (or the EP's
KDoc): a documented optional report path — e.g.
`TypeInformationGenerationResult` with the fields above (the three lists
are the actionable ones: `failed` / `factory_unresolved` /
`dynamic_unresolved` map 1:1 to what the `--strict` audit refuses).

Reference numbers from a real run against Sage 10.9 (WSL): 576/576
generated, 60 factories inferred, 0 unresolved, 440 return annotations
upgraded.  Happy to adjust the schema proposal if you'd prefer a subset.

PR #3615, opened 2026-08-17
IDEA: detect AI coding CLIs for plugin suggestions

wangmingliang-ms:
## Summary

Add the executable basenames for three AI coding CLIs to `EnvironmentDependencyCollector`:

- `copilot` — GitHub Copilot CLI
- `codex` — OpenAI Codex CLI
- `claude` — Claude Code

This lets IntelliJ detect the commands from `PATH` and resolve Marketplace plugins that declare matching `dependencySupport` features such as `executable:copilot`.

## Motivation

GitHub Copilot for IntelliJ plans to declare these executable mappings in its main plugin descriptor so users of AI coding CLIs can discover the IDE integration through the existing **Plugins | Suggested** experience.

## Testing

- `git diff --check`
- reviewed the existing cross-platform `EnvironmentScannerTest`, which covers executable suffix handling and executable-file checks used by this collector
- no scanner behavior is changed; this patch only extends the collector's data list

### DIFF
--- a/platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/pluginsAdvertisement/EnvironmentDependencyCollector.kt
+++ b/platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/pluginsAdvertisement/EnvironmentDependencyCollector.kt
@@ -14,6 +14,11 @@ internal class EnvironmentDependencyCollector : DependencyCollector {
     "podman",
     "terraform",
 
+    // AI coding CLIs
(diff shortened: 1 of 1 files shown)


jreznot:
Thanks for the suggestion, and sorry for the disappointing answer — we currently don't plan to support detecting those CLIs for plugin suggestions. That decision isn't really a technical one; it depends on JetBrains' broader product direction, so it's outside what we can commit to here.

PR #3617, opened 2026-08-17
[JEWEL-1391] Add ten built-in grammars to the standalone highlighter

DanielSouzaBertoldi:
## Context

Now that we have syntax highlighting for markdown code blocks in Standalone, it's time to add a bunch of grammars so users can be happy and then they can spread the word about our awesome project and then they can talk to their moms about it and their moms are gonna be like "that's pretty cool but idk what any of these words mean" (unless their mom was a programmer back in the day then they can sort of understand it a lil bit better i guess)

Anywho, **AI was used HEAVILY** to port regex from the TextMate plugin to our solution, especially because TextMate has a more "complete" architecture for checking syntax (they basically use a tree and can create rules like "a keyword can only happen inside a functionCall" or stuff like that) while ours is a flat regex lookup and there are some downsides to that. But we can get away with 95% of all regex rules they have so that's a pretty huge win in my book 😎 

Letting AI do its thing does not mean I did not review the decisions it took during the whole process. There were quite a couple of things wrong that are now fixed (I hope I found all of them). Having said that, I encourage you to ask your own AI to review the regex port (I'll even give you a prompt you can use instead of having to ask it to do that using your own words). Check the **Review Notes** section below.

## Changes

- Adds ten built-in grammars: `C`, `CSS`, `HTML`, `JavaScript`, `JSX`, `JSON`, `Python`, `Shell`, `SQL` and `YAML`. All patterns are ported verbatim from the `TextMate` bundles in `plugins/textmate/lib/bundles/`; each file header lists where the flat rule engine forced a deviation and why.
- Adds `TokenType.PROPERTY_KEY` and `TokenType.OPERATOR`, with `TokenRule.propertyKey()` and `TokenRule.operator()`. Both map to IntelliJ keys that had no equivalent before, `DEFAULT_INSTANCE_FIELD` and `DEFAULT_OPERATION_SIGN`.
- Fixes quadratic tokenizing. `SimpleCodeHighlighter` now caches each rule's next match, since rules with nothing left to match rescanned to end-of-input at every cursor position. Shell went from 13.3s to 62ms on 36KB, with byte-identical spans.
- Re-sources both built-in palettes from IntelliJ's Light and Dark schemes. They previously used the legacy Default and Darcula, which no shipped theme selects, so light-mode function names came out unstyled.
- Corrects scope mapping against `TextMateDefaultColorsProvider`. `storage.type` is a keyword rather than a type, and `keyword.operator` has its own key. This changes Kotlin and Java, which shipped earlier.

### AI Time Babyyy

Here's a handy prompt to give to your AI so that we can try and cover all bases. Feel free to change the prompt however you want it (it assumes you are checked into this branch):

[regex-review-prompt.md](https://github.com/user-attachments/files/31179928/regex-review-prompt.md)

### Human Time Babyyy

Human eyes would help most on:

- [SimpleCodeHighlighter.kt](https://github.com/JetBrains/intellij-community/pull/3617/changes#diff-433bf8a5e2318b03b30d4d223e3abe7872f732c233449cce261c8d963b922991) (the cache, and its assumption that no rule uses \G)
- [TokenType.kt](https://github.com/JetBrains/intellij-community/pull/3617/changes#diff-433bf8a5e2318b03b30d4d223e3abe7872f732c233449cce261c8d963b922991), [TokenRule.kt](https://github.com/JetBrains/intellij-community/pull/3617/changes#diff-433bf8a5e2318b03b30d4d223e3abe7872f732c233449cce261c8d963b922991), [SyntaxHighlightColors.kt](https://github.com/JetBrains/intellij-community/pull/3617/changes#diff-62b5ca185e539c8a2288f3baeb64b29c7ff2b442dd8ce18a6605bf0fe9d82750) (public API surface)
- [JavaLanguageGrammar.kt](https://github.com/JetBrains/intellij-community/pull/3617/changes#diff-62b5ca185e539c8a2288f3baeb64b29c7ff2b442dd8ce18a6605bf0fe9d82750), [KotlinLanguageGrammar.kt](https://github.com/JetBrains/intellij-community/pull/3617/changes#diff-62b5ca185e539c8a2288f3baeb64b29c7ff2b442dd8ce18a6605bf0fe9d82750) (already-shipped languages whose colors change)
- [docs/standalone-code-highlighting.md](https://github.com/JetBrains/intellij-community/pull/3617/changes#diff-62b5ca185e539c8a2288f3baeb64b29c7ff2b442dd8ce18a6605bf0fe9d82750) (the porting guide contributors will follow)

#### Light mode
| Language | GitHub (light default) | Jewel Standalone | Jewel Bridge (light) |
| C | C github light | C standalone light | C bridge light |
| CSS | CSS github light | CSS standalone light | CSS bridge light |
| HTML | HTML github light | HTML standalone light | HTML bridge light |
| Java | Java github light | Java standalone light | Java bridge light |
| JavaScript | JavaScript github light | JavaScript standalone light | JavaScript bridge light |
| JSON | JSON github light | JSON standalone light | JSON bridge light |
| JSX | JSX github light | JSX standalone light | JSX bridge light |
| Kotlin | Kotlin github light | Kotlin standalone light | Kotlin bridge light |
| Python | Python github light | Python standalone light | Python bridge light |
| Shell | Shell github light | Shell standalone light | Shell bridge light |
| SQL | SQL github light | SQL standalone light | SQL bridge light |
| YAML | YAML github light | YAML standalone light | YAML bridge light |

#### Dark mode
| Language | GitHub (dark default) | Jewel Standalone | Jewel Bridge (dark) |
| C | C github dark | C standalone dark | C bridge dark |
| CSS | CSS github dark | CSS standalone dark | CSS bridge dark |
| HTML | HTML github dark | HTML standalone dark | HTML bridge dark |
| Java | Java github dark | Java standalone dark | Java bridge dark |
| JavaScript | JavaScript github dark | JavaScript standalone dark | JavaScript bridge dark |
| JSON | JSON github dark | JSON standalone dark | JSON bridge dark |
| JSX | JSX github dark | JSX standalone dark | JSX bridge dark |
| Kotlin | Kotlin github dark | Kotlin standalone dark | Kotlin bridge dark |
| Python | Python github dark | Python standalone dark | Python bridge dark |
| Shell | Shell github dark | Shell standalone dark | Shell bridge dark |
| SQL | SQL github dark | SQL standalone dark | SQL bridge dark |
| YAML | YAML github dark | YAML standalone dark | YAML bridge dark |

### ⚠️ Important Changes

- `SyntaxHighlightColors` takes two new constructor parameters, `propertyKey` and operator. Custom palettes need updating.
- The built-in palettes now mirror IntelliJ's Light and Dark editor schemes instead of the legacy Default and Darcula. Every color changed, and function names are now colored in light mode.
- Kotlin and Java highlighting changed to match the `TextMate` plugin's own scope mapping. Java primitives such as `int` are keywords rather than types, `instanceof` is an operator, Kotlin stdlib types such as `String` are builtins, and the name in class `Foo` is now a type.

### New features

- Ten new built-in grammars: C, CSS, HTML, JavaScript, JSX, JSON, Python, Shell, SQL and YAML.
- `TokenType.PROPERTY_KEY` and `TokenType.OPERATOR`, with `TokenRule.propertyKey()` and `TokenRule.operator()`.

### Bug fixes

- Highlighting no longer degrades quadratically with input size.

### DIFF
--- a/platform/jewel/int-ui/int-ui-standalone-tests/src/test/kotlin/org/jetbrains/jewel/intui/code/highlighting/languages/CSSGrammarTest.kt
+++ b/platform/jewel/int-ui/int-ui-standalone-tests/src/test/kotlin/org/jetbrains/jewel/intui/code/highlighting/languages/CSSGrammarTest.kt
@@ -1,0 +1,319 @@
+// Copyright 2000-2026 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package org.jetbrains.jewel.intui.code.highlighting.languages
+
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+import kotlin.test.assertTrue
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.test.runTest
+import org.jetbrains.jewel.intui.code.highlighting.colorAt
+import org.jetbrains.jewel.intui.code.highlighting.spansAt
+import org.jetbrains.jewel.intui.code.highlighting.testColors
+import org.jetbrains.jewel.intui.standalone.code.highlighting.SimpleCodeHighlighter
+import org.junit.jupiter.api.Test
+
+internal class CSSGrammarTest {
+    private val highlighter = SimpleCodeHighlighter(testColors)
+
+    private suspend fun highlight(code: String, language: String = "css") =
+        highlighter.highlight(code, language).first()
+
+    @Test fun `name is recognized`() = runTest { assertTrue(highlight("a { color: red; }").spanStyles.isNotEmpty()) }
+
+    @Test
+    fun `comment is colored as comment`() = runTest {
+        assertEquals(testColors.comment, highlight("/* a comment */").colorAt(0))
+    }
+
+    @Test
+    fun `property name is colored as property key`() = runTest {
+        // "color" starts at index 4 in "a { color: red; }"
+        assertEquals(testColors.propertyKey, highlight("a { color: red; }").colorAt(4))
+    }
+
+    @Test
+    fun `custom property is colored as property key`() = runTest {
+        // "--brand" starts at index 4 in "a { --brand: red; }"
+        assertEquals(testColors.propertyKey, highlight("a { --brand: red; }").colorAt(4))
+    }
+
+    @Test
+    fun `vendored property name is colored as property key`() = runTest {
+        assertEquals(testColors.propertyKey, highlight("a { -webkit-box-shadow: none; }").colorAt(4))
+    }
+
+    @Test
+    fun `element selector is colored as keyword`() = runTest {
+        assertEquals(testColors.keyword, highlight("a { color: red; }").colorAt(0))
+    }
+
+    @Test
+    fun `custom element selector is colored as keyword`() = runTest {
+        assertEquals(testColors.keyword, highlight("my-widget { color: red; }").colorAt(0))
+    }
+
+    @Test
+    fun `class selector is colored as type`() = runTest {
+        assertEquals(testColors.type, highlight(".foo { color: red; }").colorAt(0))
+    }
+
+    @Test
+    fun `class selector may contain digits and double hyphens`() = runTest {
+        assertEquals(testColors.type, highlight(".card--wide { color: red; }").colorAt(0))
+        assertEquals(testColors.type, highlight(".col-6 { color: red; }").colorAt(0))
+    }
+
+    @Test
+    fun `id selector is colored as type`() = runTest {
+        assertEquals(testColors.type, highlight("#foo { color: red; }").colorAt(0))
+    }
+
+    @Test
+    fun `an id selector that looks like a hex color is still a selector`() = runTest {
+        // `#foo` above never reached the hex rule, since `o` is not a hex digit. These do.
+        assertEquals(testColors.type, highlight("#abc { color: red; }").colorAt(0))
+        assertEquals(testColors.type, highlight("#abcdef { color: red; }").colorAt(0))
+        assertEquals(testColors.type, highlight("#abc, #def { color: red; }").colorAt(6))
+        assertEquals(testColors.type, highlight("a:hover #abc { color: red; }").colorAt(8))
+        assertEquals(testColors.type, highlight("a { color: red; } #abc { color: red; }").colorAt(18))
+    }
+
+    @Test
+    fun `a hyphenated id selector is colored whole, not just its hex-looking prefix`() = runTest {
+        val result = highlight("#abc-def { color: red; }")
+        assertEquals(testColors.type, result.colorAt(0))
+        assertEquals(testColors.type, result.colorAt(5))
+    }
+
+    @Test
+    fun `pseudo-class is colored as builtin`() = runTest {
+        val result = highlight("a:hover { color: red; }")
+        assertEquals(testColors.keyword, result.colorAt(0), "'a' is still an element selector")
+        assertEquals(testColors.builtin, result.colorAt(1), "':hover' should be a pseudo-class")
+        assertEquals(testColors.builtin, result.colorAt(2), "the name is part of the same match")
+    }
+
+    @Test
+    fun `pseudo-element is colored as builtin`() = runTest {
+        assertEquals(testColors.builtin, highlight("a::before { color: red; }").colorAt(1))
+    }
+
+    @Test
+    fun `pseudo-class is not mistaken for a property`() = runTest {
+        // The grammar's property-name list does not contain `hover`, and the pseudo-class rule owns the colon
+        assertNotEquals(testColors.propertyKey, highlight("a:hover { color: red; }").colorAt(0))
+    }
+
+    @Test
+    fun `functional pseudo-class colors its nth expression as a number`() = runTest {
+        val result = highlight("li:nth-child(2n+1) { color: red; }")
+        assertEquals(testColors.builtin, result.colorAt(2))
+        assertEquals(testColors.number, result.colorAt(13))
+    }
+
+    @Test
+    fun `functional pseudo-class colors its parity keyword as builtin`() = runTest {
+        assertEquals(testColors.builtin, highlight("li:nth-child(odd) { color: red; }").colorAt(13))
+    }
+
+    @Test
+    fun `negation pseudo-class is colored as builtin`() = runTest {
+        assertEquals(testColors.builtin, highlight("a:not(.x) { color: red; }").colorAt(1))
+    }
+
+    @Test
+    fun `a value that is also a tag name stays a value without a trailing semicolon`() = runTest {
+        // Regression: about twenty words are in both the tag-name and property-value lists. The bundle keeps
+        // them apart by nesting; flattened, `table` matched the tag rule because `\s` is in its lookahead set
+        // and `}` follows. Declarations ending in `;` were never affected.
+        for (value in listOf("table", "small", "ruby", "progress", "menu")) {
+            assertEquals(
+                testColors.builtin,
+                highlight("a { display: $value }").colorAt(13),
+                "'$value' without a trailing semicolon should still be a property value",
+            )
+            assertEquals(
+                testColors.builtin,
+                highlight("a { display: $value; }").colorAt(13),
+                "'$value' with a trailing semicolon should be a property value",
+            )
+        }
+    }
+
+    @Test
+    fun `the same words are still tag names in selector position`() = runTest {
+        // The other half of the guard: it must not cost us selector highlighting
+        for (tag in listOf("table", "small", "ruby", "progress", "menu", "code")) {
+            assertEquals(
+                testColors.keyword,
+                highlight("$tag { color: red; }").colorAt(0),
+                "'$tag' in selector position should be a tag name",
+            )
+        }
+    }
+
+    @Test
+    fun `hex color is colored as constant`() = runTest {
+        // "#fff" starts at index 11 in "a { color: #fff; }"
+        assertEquals(testColors.constant, highlight("a { color: #fff; }").colorAt(11))
+    }
+
+    @Test
+    fun `a hex color in value position is a color, not an id`() = runTest {
+        val result = highlight("a { color: #abc; }")
+        assertEquals(testColors.constant, result.colorAt(11))
+        assertNotEquals(testColors.type, result.colorAt(11), "'#abc' in a value is a color, not an id")
+        // Every shape the hex rule accepts, in the positions a value can take
+        assertEquals(testColors.constant, highlight("a { color: #abc }").colorAt(11))
+        assertEquals(testColors.constant, highlight("a { color:#abc; }").colorAt(10))
+        assertEquals(testColors.constant, highlight("a { color: #abcd; }").colorAt(11))
+        assertEquals(testColors.constant, highlight("a { color: #aabbccdd; }").colorAt(11))
+        assertEquals(testColors.constant, highlight("a { border: 1px solid #ff0000; }").colorAt(22))
+        assertEquals(
+            testColors.constant,
+            highlight("a { background: linear-gradient(#abc, #def); }").colorAt(32),
+            "hex inside a function call",
+        )
+    }
+
+    @Test
+    fun `document-rule functions are colored as function call`() = runTest {
+        // support.function.document-rule.css — each name starts at index 10 in `@document <name>("x") { }`
+        for (function in listOf("url-prefix", "domain", "regexp")) {
+            assertEquals(
+                testColors.functionCall,
+                highlight("@document $function(\"x\") { }").colorAt(10),
+                "'$function' should be colored as a function call",
+            )
+        }
+    }
+
+    @Test
+    fun `at-rule is colored as keyword`() = runTest {
+        assertEquals(testColors.keyword, highlight("@media screen").colorAt(0))
+        assertEquals(testColors.keyword, highlight("@font-face { }").colorAt(0))
+    }
+
+    @Test
+    fun `vendor-prefixed at-rule is colored as keyword`() = runTest {
+        assertEquals(testColors.keyword, highlight("@-webkit-keyframes spin { }").colorAt(0))
+    }
+
+    @Test
+    fun `important is colored as keyword`() = runTest {
+        assertEquals(testColors.keyword, highlight("!important").colorAt(0))
+        assertEquals(testColors.keyword, highlight("a { color: red !important; }").colorAt(15))
+    }
+
+    @Test
+    fun `numbers with units are colored as number`() = runTest {
+        // "10px" starts at index 11 in "a { width: 10px; }"
+        assertEquals(testColors.number, highlight("a { width: 10px; }").colorAt(11))
+        assertEquals(testColors.number, highlight("a { width: 50%; }").colorAt(11))
+        assertEquals(testColors.number, highlight("a { margin: -1.5em; }").colorAt(12))
+    }
+
+    @Test
+    fun `unit gets its own keyword span`() = runTest {
+        // The bundle scopes the unit keyword.other.unit.*.css, so "rem" is both inside the number span and its own
+        // keyword span. "rem" starts at index 12 in "a { width: 2rem; }".
+        assertTrue(highlight("a { width: 2rem; }").spansAt(12).any { it.color == testColors.keyword })
+    }
+
+    @Test
+    fun `decimal without a leading digit is a number, not a class`() = runTest {
+        assertEquals(testColors.number, highlight("a { opacity: .5; }").colorAt(13))
+        assertEquals(testColors.number, highlight("a { color: rgba(0, 0, 0, .5); }").colorAt(25))
+    }
+
+    @Test
+    fun `string is colored as string`() = runTest {
+        // The opening quote is at index 13 in "a { content: \"x\"; }"
+        assertEquals(testColors.string, highlight("a { content: \"x\"; }").colorAt(13))
+        assertEquals(testColors.string, highlight("a { content: '\\201C'; }").colorAt(13))
+    }
+
+    @Test
+    fun `color keyword is colored as builtin`() = runTest {
+        assertEquals(testColors.builtin, highlight("a { color: red; }").colorAt(11))
+        assertEquals(testColors.builtin, highlight("a { color: rebeccapurple; }").colorAt(11))
+        assertEquals(testColors.builtin, highlight("a { color: currentColor; }").colorAt(11))
+    }
+
+    @Test
+    fun `property value keyword is colored as builtin`() = runTest {
+        assertEquals(testColors.builtin, highlight("a { display: flex; }").colorAt(13))
+        assertEquals(testColors.builtin, highlight("a { list-style-type: lower-roman; }").colorAt(21))
+        assertEquals(testColors.builtin, highlight("a { font-family: helvetica; }").colorAt(17))
+        assertEquals(testColors.builtin, highlight("a { color: -moz-fixed; }").colorAt(11))
+    }
+
+    @Test
+    fun `function name is colored as function call`() = runTest {
+        assertEquals(testColors.functionCall, highlight("a { width: calc(1px); }").colorAt(11))
+        assertEquals(testColors.functionCall, highlight("a { color: rgba(0,0,0,.5); }").colorAt(11))
+        assertEquals(testColors.functionCall, highlight("a { width: min(1px, 2px); }").colorAt(11))
+        assertEquals(testColors.functionCall, highlight("a { clip-path: circle(1px); }").colorAt(15))
+        assertEquals(testColors.functionCall, highlight("a { transform: translateX(1px); }").colorAt(15))
+        assertEquals(testColors.functionCall, highlight("a { transition-timing-function: steps(2); }").colorAt(32))
+    }
+
+    @Test
+    fun `gradient url and var are colored as function calls`() = runTest {
+        assertEquals(testColors.functionCall, highlight("a { background: linear-gradient(red, blue); }").colorAt(16))
+        assertEquals(testColors.functionCall, highlight("a { background: url(x.png); }").colorAt(16))
+        assertEquals(testColors.functionCall, highlight("a { color: var(--x); }").colorAt(11))
+    }
+
+    @Test
+    fun `attribute selector names its attribute and operator`() = runTest {
+        assertEquals(testColors.propertyKey, highlight("a[href] { color: red; }").colorAt(2))
+        assertEquals(testColors.operator, highlight("a[href=\"x\"] { color: red; }").colorAt(6))
+    }
+
+    @Test
+    fun `media feature is colored as property key`() = runTest {
+        assertEquals(testColors.propertyKey, highlight("@media (min-width: 600px) { }").colorAt(8))
+    }
+
+    @Test
+    fun `media type and logical operator are colored`() = runTest {
+        val result = highlight("@media screen and (color) { }")
+        assertEquals(testColors.builtin, result.colorAt(7), "'screen' is a media type")
+        assertEquals(testColors.operator, result.colorAt(14), "'and' is a logical operator")
+    }
+
+    @Test
+    fun `media feature keyword is colored as builtin`() = runTest {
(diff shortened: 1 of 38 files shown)


platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
```diff
@@ -87,13 +87,33 @@ public class SimpleCodeHighlighter(
...
+        val nextMatch = arrayOfNulls<MatchResult>(grammar.rules.size)
+        val exhausted = BooleanArray(grammar.rules.size)
+
         while (i < code.length) {
             // Find the rule whose match starts earliest; ties broken by rule order
             var bestMatch: MatchResult? = null
             var bestRule: TokenRule? = null
 
-            for (rule in grammar.rules) {
-                val match = rule.find(code, i) ?: continue
```
DanielSouzaBertoldi, line 96: This was ok and all but the problem is that `find` would scan forward from `i` until it matches or reaches the end of the input. This is cool only if a rule still has matches ahead, so it would stop early if it found something. In this case, this is a very cheap processing time.

Now, in the worst case scenario, a rule that does not have any matches would scan the entire rest of the document at every subsequent cursor positon. That's not good, in fact it's no bueno.

At the end of the day, each such rule contributes an `O(n²)` term, so the whole tokenizer is `O(rules × n²)`. Ouch.

When porting SQL and CSS this issue became really apparent given that they have 50+ rules and most of them do not match anything in the code block.

The fix here is to just re-run `find()` once the cursor passes its last match, and if a search returns null, every later search will return null too so we can just give up searching for matches early.

With this, we can keep using `lookbehind` and `\A`, but **NOT** `\G`, since it anchors to the enclosing begin and we have no contexts. The cache adds a second reason to keep it out: cached matches and `\G` disagree about where the anchor is, so you'd get wrong spans, not slow ones, and the `standalone-code-highlighting.md` docs now say so explicitly.

platform/jewel/int-ui/int-ui-standalone/metalava/int-ui-standalone-api-0.41.0.txt
```diff
@@ -137,13 +137,15 @@ package org.jetbrains.jewel.intui.standalone.code.highlighting {
   }
 
   @SuppressCompatibility @androidx.compose.runtime.Immutable @org.jetbrains.annotations.ApiStatus.Experimental @org.jetbrains.jewel.foundation.ExperimentalJewelApi public final class SyntaxHighlightColors {
-    ctor @KotlinOnly public SyntaxHighlightColors(androidx.compose.ui.graphics.Color keyword, androidx.compose.ui.graphics.Color type, androidx.compose.ui.graphics.Color constant, androidx.compose.ui.graphics.Color functionCall, androidx.compose.ui.graphics.Color string, androidx.compose.ui.graphics.Color comment, androidx.compose.ui.graphics.Color number, androidx.compose.ui.graphics.Color builtin);
```
DanielSouzaBertoldi, line 140: This is not a breaking change since we did not ship this code yet :P

(the same is true for other metalava dumps and the BCV dump, aka api-dump-test)

platform/jewel/ui/metalava/ui-api-0.41.0.txt
```diff
@@ -3708,7 +3708,7 @@ package org.jetbrains.jewel.ui.icons {
     field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey Add;
     field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey AddJdk;
     field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey Add_20x20;
-    field @SuppressCompatibility public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey Alpha;
+    field @Deprecated @SuppressCompatibility @org.jetbrains.annotations.ApiStatus.ScheduledForRemoval public static final org.jetbrains.jewel.ui.icon.IntelliJIconKey Alpha;
```
DanielSouzaBertoldi, line 3711: Just ignore those... We are still having problems with internal changes to the icons api that do not update the metalava dumps :/

DanielSouzaBertoldi:
Jewel checks is failing because of timeout issues 😬 

> Java setup process failed due to: API rate limit exceeded for 52.234.43.177

I'll keep trying to fix it, but please do ignore the red cross in that job for now.

rock3r:
The rate limit issues sort themselves out :)

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
```diff
@@ -87,13 +87,33 @@ public class SimpleCodeHighlighter(
         val spans = mutableListOf<Span>()
         var i = 0
 
+        // Cache each rule's next match. Without this, rules with nothing left to match rescan to end-of-input
+        // at every cursor position and tokenizing goes quadratic.
+        //
+        // Safe because find() returns the leftmost match at or after the start index, so a cached match that
+        // still starts at or after `i` is what a fresh find would return. Re-find only once the cursor passes it.
+        // This would break for \G, which anchors to the search start, but ported grammars have to strip \G.
```
rock3r, line 95: Could we add a regression test that asserts the built-in grammars contain no `\G`? The cache is only correct without it: a future port containing `\G` produces wrong spans, not merely slower highlighting. This way we don't risk it slipping in while we're not looking :)

Keeping this check scoped to `BuiltInLanguageGrammars` leaves custom rules subject to the documented limitation, but hey, their problem.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SyntaxHighlightColors.kt
```diff
@@ -40,17 +42,21 @@ public class SyntaxHighlightColors(
     public val comment: Color,
     public val number: Color,
     public val builtin: Color,
+    public val propertyKey: Color,
```
rock3r, line 45: Could we keep custom palettes source- and binary-compatible here? Defaults of `Color.Unspecified` keep existing Kotlin calls compiling, but the old eight-argument JVM constructor still needs a hidden deprecated forwarding overload.

I'd rather keep the old concrete constructor hidden than make callers update for two token kinds that can safely inherit ambient text...

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/JSXLanguageGrammar.kt
```diff
@@ -0,0 +1,29 @@
...
+    listOf(
+        // support.class.component.js.jsx
+        // capitalized names are components, lowercase ones are DOM elements.
+        TokenRule("</?([A-Z][\\w\$.]*)(?=[\\s/>])", mapOf(1 to TokenType.TYPE)),
+        // entity.name.tag.js.jsx
+        TokenRule("</?([a-z][\\w.:-]*)(?=[\\s/>])", mapOf(1 to TokenType.KEYWORD)),
+        // entity.other.attribute-name.js.jsx
+        // requires `name="…"` with no spaces around `=`, which is JSX style. `x = "y"` in ordinary JS
+        // keeps its spaces and is left alone.
+        TokenRule.propertyKey("[A-Za-z_\$][\\w\$:.-]*(?==[\"'{])"),
```
rock3r: [K3] Wouldn't this also recognize compact ordinary JavaScript assignments? `const x="y"` in a JSX file styles `x` as a property key because this has no opening-tag context. TextMate only enables the attribute rule inside a tag; whitespace around `=` isn't a syntactic boundary. Can we prove tag context here, or leave attributes plain when flat matching can't do that safely?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/PythonLanguageGrammar.kt
```diff
@@ -0,0 +1,78 @@
...
+                        "hasattr|hash|id|isinstance|issubclass|iter|map|max|min|next|repr|reversed|round|" +
+                        "setattr|sorted|sum|zip|vars|dir|eval|exec|divmod|chr|ord|hex|oct|bin|callable)\\b"
+                ),
+                // entity.name.function.call — after keywords so `if (` stays a keyword
+                TokenRule.functionCall("\\b([A-Za-z_][A-Za-z0-9_]*)\\s*(?=\\()"),
+                // constant.numeric.python — the j suffix is an imaginary literal
+                TokenRule.number("\\b0[bB][01_]+\\b"),
+                TokenRule.number("\\b0[oO][0-7_]+\\b"),
+                TokenRule.number("\\b0[xX][0-9a-fA-F_]+\\b"),
+                TokenRule.number("\\b\\d[\\d_]*(?:\\.[\\d_]*)?(?:[eE][+-]?\\d+)?[jJ]?\\b"),
```
rock3r: [K3] Looks like this number pattern misses valid float spellings: `.5` and `.5e2` leave the leading dot plain, and valid Python `1e1_0` gets no number span at all. MagicPython has a leading-dot float branch and permits underscores in the exponent. Could we port both cases and add regressions for them?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/HTMLLanguageGrammar.kt
```diff
@@ -0,0 +1,165 @@
...
+                    mapOf(2 to TokenType.KEYWORD),
+                ),
+                // #tags-valid, meta.tag.custom.{start,end}.html
+                TokenRule("(<)($CUSTOM_TAG_NAME)(?=\\s|/?>)", mapOf(2 to TokenType.KEYWORD)),
+                TokenRule("(</)($CUSTOM_TAG_NAME)(?=\\s|/?>)", mapOf(2 to TokenType.KEYWORD)),
+                // #tags-invalid — the catch-all. Group 3 is invalid.illegal.unrecognized-tag.html, skipped
+                TokenRule("(</?)((\\w[^\\s>]*))(?<!/)", mapOf(2 to TokenType.KEYWORD)),
+                // #attribute — entity.other.attribute-name.html on the whole match.
+                // The trailing (?=\s*=) is ours, not the bundle's: see adaptation 5 in the header.
+                TokenRule.propertyKey(HTML5_ATTRIBUTES + ATTRIBUTE_VALUE_AHEAD),
```
rock3r: [K3] Could we add a leading `(?<![\w:-])` guard to these attribute alternatives too? The trailing guard prevents an overlong match, but `<div mytitle="x">` still starts matching at `title` and colors only that suffix as a property key. The same applies to the `style`, event-handler, and `data-` rules.

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/CLanguageGrammar.kt
```diff
@@ -0,0 +1,234 @@
...
+                TokenRule.keyword("(?<!\\w)switch(?!\\w)"),
+                TokenRule.keyword("(?<!\\w)case(?!\\w)"),
+                TokenRule.keyword("(?<!\\w)default(?!\\w)"),
+                // keyword.control.c
+                TokenRule.keyword("\\b(break|continue|do|else|for|goto|if|_Pragma|return|while)\\b"),
+                // #storage_types — the aggregate and asm words read as keywords; see the header
+                TokenRule.keyword("(?-mix:\\b(enum|struct|union)\\b)"),
+                TokenRule.keyword("\\b(?:__asm__|asm)\\b"),
+                // keyword.other.typedef.c, then storage.modifier.c
+                TokenRule.keyword("typedef"),
```
rock3r: [K3] Looks like this is missing word boundaries: `int mytypedefName;` colors the `typedef` substring as a keyword. Can we use the same boundary treatment as the neighboring keyword rules and add an identifier-substring regression?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/YAMLLanguageGrammar.kt
```diff
@@ -0,0 +1,100 @@
...
+                TokenRule("(?m)^(%)(YAML)([\\t ]++)(1\\.[0-3])", mapOf(2 to TokenType.KEYWORD, 4 to TokenType.NUMBER)),
+                // keyword.other.directive.tag.yaml, from #directives
+                TokenRule("(?m)^(%)(TAG)(?>([\\t ]++)((!)(?>[0-9A-Za-z-]*+(!))?+))?+", mapOf(2 to TokenType.KEYWORD)),
+                // entity.other.document.begin.yaml / .end.yaml. No TokenType corresponds to
+                // entity.other.document, so KEYWORD is our choice rather than the bundle's.
+                TokenRule.keyword("(?m)^---(?=[\\r\\n\\t ]|\\z)"),
+                TokenRule.keyword("(?m)^\\.{3}(?=[\\r\\n\\t ]|\\z)"),
+                // meta.map.key.yaml string.quoted.{double,single}.yaml entity.name.tag.yaml
+                TokenRule(
+                    KEY_INDENT + "(\"(?>[^\\\\\"]++|\\\\.)*+\"|'(?>[^']++|'')*+')" + KEY_END,
```
rock3r, line 76: [K3] Wouldn't `KEY_INDENT` make both key rules block-only? Valid flow maps such as `{foo: bar}` leave `foo` plain, while `{"foo": bar}` treats the key as a generic string rather than a property key. Could we cover plain and quoted flow-map keys too?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/CSSLanguageGrammar.kt
```diff
@@ -0,0 +1,410 @@
...
+                // string.quoted.double.css / string.quoted.single.css — begin `"`, end `"|(?<!\\)(?=$|\n)`, with
+                // #escapes as the body
+                TokenRule.string("(?m)\"(?:\\\\(?:[0-9a-fA-F]{1,6}|.)|[^\"\\\\\\r\\n])*+(?:\"|(?<!\\\\)(?=\$))"),
+                TokenRule.string("(?m)'(?:\\\\(?:[0-9a-fA-F]{1,6}|.)|[^'\\\\\\r\\n])*+(?:'|(?<!\\\\)(?=\$))"),
+                // constant.character.escape.codepoint.css / .newline.css / .css
+                TokenRule.constant("\\\\[0-9a-fA-F]{1,6}"),
+                TokenRule.constant("(?m)\\\\\$\\s*"),
+                TokenRule.constant("\\\\."),
+                // constant.other.color.rgb-value.hex.css — before the id-selector rule, since `#abc` is both
+                TokenRule.constant("(#)(?:[0-9a-fA-F]{3,4}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})\\b"),
```
rock3r: [K3] Wouldn't an ID selector such as `#abc { color: red; }` match this first? The hex-color and ID-selector rules both start at 0, and the flat engine uses rule order to break the tie, so `#abc` is colored as a constant. `#abc-def` also gets a partially colored `#abc` prefix. Can we establish declaration/value context here, or otherwise avoid treating ambiguous selector syntax as a color?

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SyntaxHighlightColors.kt
```diff
@@ -92,45 +102,55 @@ public class SyntaxHighlightColors(
...
+            "builtin=$builtin, " +
+            "propertyKey=$propertyKey, " +
+            "operator=$operator" +
             ")"
 
     /** Companion object for [SyntaxHighlightColors], holding the built-in palettes. */
     public companion object {
         /**
-         * Returns a [SyntaxHighlightColors] palette matching IntelliJ IDEA's "Default" (light) editor color scheme.
+         * Returns a [SyntaxHighlightColors] palette matching IntelliJ IDEA's "Light" editor color scheme, the one the
```
rock3r, line 113: `SimpleCodeHighlighter` still says that these palettes match IntelliJ's Default and Darcula schemes. Could we update that KDoc to say Light and Dark too?

rock3r (changes requested):

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SyntaxHighlightColors.kt
DanielSouzaBertoldi: I mean, if we manage to ship this in 0.41 then that's not an issue since this class was never shipped at all to consumers :P

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SimpleCodeHighlighter.kt
DanielSouzaBertoldi: Added a new test class `BuiltInLanguageGrammarsTest.kt`! Keep in mind I had to make the `BuiltInLanguageGrammars` object public and add the internal api annotations, though

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SyntaxHighlightColors.kt
DanielSouzaBertoldi: Done!

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/CLanguageGrammar.kt
DanielSouzaBertoldi: Damn, K3 just found a bug in TextMate's regex 🤯 

Image

With this, ours is an "improved" version now. At least for `typedef`s lol

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/CSSLanguageGrammar.kt
DanielSouzaBertoldi: Done

| Before | After |
|---|---|
| Image | Image |

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/HTMLLanguageGrammar.kt
DanielSouzaBertoldi: Done

| Before | After |
|---|---|
| Image | Image |

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/JSXLanguageGrammar.kt
DanielSouzaBertoldi: Done!

| Before | After |
|---|---|
| Image |   Image |

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/PythonLanguageGrammar.kt
DanielSouzaBertoldi: Done!

| Before | After |
|---|---|
| Image | Image |

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/YAMLLanguageGrammar.kt
DanielSouzaBertoldi: Done!

| Before | After |
|---|---|
| Image | Image |

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/languages/CLanguageGrammar.kt
rock3r: Yay open weights 😎

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/code/highlighting/SyntaxHighlightColors.kt
rock3r: Fair point, I got confused as to what is "new" and what isn't XD

rock3r (approved):
All comments addressed — thanks! The fixes check out mechanically (exact-grammar tokenization of every reported trigger plus regression cases), the new \G guard test covers the cache invariant, and the flow-map / selector / float ports look faithful to the bundles.

platform/jewel/int-ui/int-ui-standalone-tests/src/test/kotlin/org/jetbrains/jewel/intui/code/highlighting/languages/CSSGrammarTest.kt
```diff
@@ -0,0 +1,319 @@
...
+import org.jetbrains.jewel.intui.standalone.code.highlighting.SimpleCodeHighlighter
+import org.junit.jupiter.api.Test
+
+internal class CSSGrammarTest {
+    private val highlighter = SimpleCodeHighlighter(testColors)
+
+    private suspend fun highlight(code: String, language: String = "css") =
+        highlighter.highlight(code, language).first()
+
+    @Test fun `name is recognized`() = runTest { assertTrue(highlight("a { color: red; }").spanStyles.isNotEmpty()) }
```
wellingtoncosta, line 21: I've always liked this convention where, if a function has a single parameterless annotation, it can appear on the same line [[1](https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations)] 😎

wellingtoncosta (approved):

DanielSouzaBertoldi:
Ready to merge 🚀 Thanks for the help guys!

PR #3618, opened 2026-08-19
[JEWEL-1390] Add Spectre and the standalone headful test lane

rock3r:
Jewel has had no way to test real UI behaviour: the Compose UI unit tests run headless and never open a window, so anything depending on a native popup, focus, or key routing has been untestable. [Spectre](https://spectre.sebastiano.dev) drives a real Compose Desktop window with synthetic input, which closes that gap. This adds it as a test-only dependency and builds the standalone headful lane around it, with the JEWEL-1396 popup regression tests as its first users.

Spectre is declared in `jewel_deps.MODULE.bazel` rather than in `lib`, and exposed as the `testonly` target `//platform/jewel:spectre`, so Bazel refuses to let any production target depend on it and it cannot reach the IDE or a published artifact.

## Changes

* Declare Spectre and its runtime closure in `jewel_deps.MODULE.bazel`, exposed as `//platform/jewel:spectre` and `//platform/jewel:spectre-junit-console`.
* Add the `spectre_test` macro in `platform/jewel/spectre.bzl`. It cannot be a `jps_test`: that macro forces `-Djava.awt.headless=true` and puts the IntelliJ test runtime on the classpath, which would both break a headful test and mask the standalone-runtime leaks the lane exists to catch.
* Wire the JEWEL-1396 popup tests in as `//platform/jewel/int-ui/int-ui-standalone-tests:jewel-intUi-standalone-spectre-tests`.
* Add `StandaloneRuntimeSmokeTest`, asserting the lane's classpath carries no IntelliJ Platform classes apart from the icons API, which standalone Jewel depends on deliberately.
* Add CI jobs on Linux (under Xvfb) and Windows, both selecting targets by the `requires-display` tag so a new `spectre_test` target joins CI without a workflow change.
* Document the lane, its standalone-only scope, and the CI contract in `docs/bazel-build-tips.md`.

## Notes for reviewers

Three things here behave unlike a normal Bazel test target, and each was found by running on real hardware rather than by reasoning about it:

* `env_inherit = ["DISPLAY", "XAUTHORITY"]` — Bazel scrubs the environment of test actions, so a display set up around `bazel test` never reaches the test JVM. Without this every test dies in `XToolkit`'s static initialiser under Xvfb, while passing happily on macOS.
* Discovery by `--select-package` — a bare `--scan-classpath` scans only classpath *directories* and every Bazel classpath entry is a jar, while `--scan-classpath=<jar>` assumes a runfiles symlink tree and so finds nothing on Windows, where Bazel uses a runfiles manifest. `--fail-if-no-tests` is what turns either mistake into a red lane rather than a green one that ran nothing.
* The X11 package list — `libxrandr2` and `libxinerama1` are dlopened by `libawt_xawt` and `libskiko`, so no linker error would ever flag them as missing.

Verified green on macOS arm64, Ubuntu 26.04 x86_64 under Xvfb, and Windows 11 x86_64, all from this commit.

Headful coverage for the IJP bridge renderers is deliberately out of scope: Spectre automates a Compose Desktop window, whereas inside the IDE a Jewel popup is a `JBPopup` hosting a `ComposePanel` and the application under test is the IDE itself. That is JEWEL-1397, now rescoped to the in-IDE lane.

### DIFF
--- a/.github/workflows/jewel-checks.yml
+++ b/.github/workflows/jewel-checks.yml
@@ -288,6 +288,81 @@ jobs:
           fi
           echo "Bazel build files look up-to-date."
 
+  spectre_tests:
+    name: Jewel Spectre tests
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: .
+
+    steps:
+      - uses: actions/checkout@v7.0.1
+        name: Check out repository
+
+      - name: Install Xvfb
+        # libxrandr2 and libxinerama1 are dlopened, so nothing reports them missing.
+        # language=bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            xvfb \
+            libxrender1 \
+            libxtst6 \
+            libxi6 \
+            libxrandr2 \
+            libxinerama1 \
+            libgl1 \
+            libfontconfig1
+
+      - name: Run the Spectre headful UI tests
+        run: |
+          xvfb-run --auto-servernum --server-args="-screen 0 1920x1080x24" \
+            ./bazel.cmd test //platform/jewel/... \
+              --build_tests_only \
+              --test_tag_filters=requires-display \
+              --test_output=errors
+
+      - name: Upload Bazel test logs
+        uses: actions/upload-artifact@v7.0.1
+        if: failure()
+        with:
+          name: spectre-test-logs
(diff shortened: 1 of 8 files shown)


nebojsa-vuksic (approved):

rock3r:
Ready to merge :)

PR #3619, opened 2026-08-19
[JEWEL-1398] Update the outdated readme

rock3r:
The readme had gotten outdated: the stability warning predates our binary-compat guarantees, the stateful icon sample doesn't compile, the standalone setup is missing the Compose compiler plugin, and the two decorated-window modules are described the wrong way round.

## Changes

* Stability warning replaced with the real policy — stable = binary compat guaranteed, experimental = best effort and annotated, deprecated ≈ two platform majors, source compat mainly for named-parameter usage. The "3rd party Compose plugins aren't officially supported" caveat moves to its own warning.
* Icon sample fixed: `hints` has to be named (the next positional param is `modifier`), and `suffix()` is declared on `PainterProviderScope`.
* Added the missing `org.jetbrains.kotlin.plugin.compose`, and noted which version each of the three plugins takes.
* Depend on `jewel-decorated-window`; `int-ui-decorated-window` is a compat-only module with no code of its own.
* `FontFamily`, `EmbeddedFontFamily` and `asComposeFontFamily()` are Compose Multiplatform APIs, not ours.
* Latest release badge now reads from Maven Central instead of a hand-pinned `0.34.0`.
* Dropped `VERSIONS.md` — stuck at 0.34 while releases reached 0.40. Release notes and Maven Central already cover it.
* Renamed "Compose for Desktop" to "Compose Multiplatform" throughout. CfD is the historical name from when Compose only targeted desktop; the official docs use "Compose Multiplatform" and plain "desktop", and no longer use the old name at all.
* Linked [jewel-ui.dev](https://jewel-ui.dev).


wellingtoncosta (approved):

rock3r:
Ready to merge

nebojsa-vuksic (approved):

PR #3620, opened 2026-08-20
Fix non-deterministic order in product-info.json

gharrma:
Some v2 modules such as intellij.libraries.test.discovery are now included in the distro multiple times across different plugins. Thus it is no longer sufficient to sort by module name alone.

As a side note: the reason sorting is needed in the first place is because PluginManagerCore.getPluginSet().enabledPlugins has a non-deterministic order (due to directory traversal logic).

---

This is purely for improving build reproducibility. I've tested this change in the Android Studio build at Google.


trespasserw:
Merged, thanks!

PR #3621, opened 2026-08-20
[JEWEL-1399] Read Jewel reviewers from the published maintainers list

rock3r:
The `jewel-pr-preparer` skill hardcoded the reviewer list in two places, which had already drifted from the real team. It now reads https://jewel-ui.dev/data/maintainers.json instead. This also relaxes the review policy that was documented alongside those lists.

## Changes

* `suggest_reviewers.py` fetches the maintainers list at run time and groups candidates by `team` (`google`/`jetbrains`) rather than the old external/internal split. Round-robin by open review load, three suggestions, and team mixing are unchanged.
* The list is public input, so it is validated before use: https-only, no cross-host redirects, size-capped, `$schema_version` checked, strict patterns for `github_username` and `team`, bad entries skipped with a warning, handles shell-quoted in the printed `gh` command. A descriptive `User-Agent` is required because Cloudflare 403s the default `urllib` one.
* On fetch failure it falls back to a bundled verbatim snapshot and warns it may be stale; if that is unusable too it exits and asks for manual reviewers. A snapshot works on a fresh checkout (a cache would not) and does not block a PR on a network blip (a hard error would).
* **Review policy:** one Jewel team approval now suffices for a small or simple change, or one confined to `platform/jewel`; an internal approval is only required when a PR touches files outside it. That is deliberately narrower than the guide's Jewel-related paths, and the guide now says so. Requesting three reviewers is unchanged — only the approval bar moved. Updated `jewel-contribution-guide.md` and `pr-guide.md`.

`.agents/skills/jewel-pr-preparer/` is the canonical source; `.claude/…` is its `render-guides.mjs` output, hence the duplicated diff. The renderer needs an Ultimate checkout, so I reproduced its transform and verified it byte-identical against `master` before regenerating.

### DIFF
--- a/.claude/skills/jewel-pr-preparer/SKILL.md
+++ b/.claude/skills/jewel-pr-preparer/SKILL.md
@@ -350,16 +350,25 @@ gh pr edit <number> --repo JetBrains/intellij-community \
   --add-reviewer "<reviewer3>"
 ```
 
-Add 3 reviewers, mixing external and internal contributors. Known Jewel team reviewer handles:
+Always add 3 reviewers, mixing teams.
 
-- External reviewers (non-JetBrains): `DanielSouzaBertoldi`, `rock3r`
-- Internal reviewers (JetBrains): `DejanMilicic`, `nebojsa-vuksic`, `AlexVanGogen`, `daaria-s`
+The reviewer pool is **not** hardcoded here. It comes from the published Jewel maintainers list at
+<https://jewel-ui.dev/data/maintainers.json>, where every maintainer carries a `team` of either `jetbrains` or `google`. Maintainers from
+Touchlab work on Google's behalf and are recorded as `google`; the only distinction that matters for reviews is whether someone is a
+JetBrains employee. Never hardcode reviewer handles in this skill, in the helper script, or in a PR — if the roster is wrong, fix the
+published list instead.
 
 Do NOT ask for a review from the author of the PR, of course.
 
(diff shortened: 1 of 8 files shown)


.claude/skills/jewel-pr-preparer/SKILL.md
DanielSouzaBertoldi, line 1: Wait, why do we have the same three files in both `.agents/` and `.claude/`?

rock3r: May I introduce you to the .ai folder and its "skills renderer"? 🫠  tl;dr it copies files instead of symlinking them (I suppose for Windows compat reasons) from .agents to .claude

rock3r: You _need_ to run the renderer script (which by the way only works in the monorepo, as it hardcodes paths and was clearly never tested in intellij-community), which copies stuff around. In this case I did a decent approximation of what that would do, copying stuff around, but the reality is that, thanks to that whole thing, now we can't have automatic bot merges for skills changes, because they ALSO change files in the ultimate repo _and_ check they are synced with community. Anyway, I did the approximation to spare @nebojsa-vuksic some time — it should be correct in the community repo, then the script should also update other stuff in the monorepo, but that I can't help with.

DanielSouzaBertoldi (approved):

rock3r:
I'd say ready to merge — but will most likely need a manual merge from @nebojsa-vuksic

PR #438, opened 2016-08-22
Close All non-project

afdw:
This PR adds action in RMB menu of tabs that closes all non-project files.

PR #3622, opened 2026-08-20
[JEWEL-1396] Take popup dismissal policy from PopupProperties

rock3r:
The custom popup renderers worked out both *whether to dismiss a popup* and *whether to consume the Escape key* from the shape of `onDismissRequest`: non-null meant "dismiss and consume", null meant "do neither". A caller wanting to suppress dismissal had to withhold the callback, which suppressed every dismissal path at once. `ComboBox` did exactly that while the pointer was over it — a pointer concern — so Escape was consumed while nothing closed, and the key never reached the surrounding dialog.

Compose's own popup keeps these concerns apart, and Jewel now does the same:

| decision | channel |
| --- | --- |
| dismiss on click outside | `PopupProperties.dismissOnClickOutside` |
| dismiss on Escape | `PopupProperties.dismissOnBackPress` |
| consume Escape | `PopupProperties.focusable` |
| notify | `onDismissRequest`, which says only whether there is any way to act |

## Changes

* `ComboBox` and `EditableComboBox` suppress only the pointer path, via `dismissOnClickOutside`, and always pass a dismissal callback. Hovering no longer affects Escape. Clicking the chevron to close an open popup behaves exactly as before.
* Both renderers gate Escape dismissal on `dismissOnBackPress`, which neither honoured before, and consume Escape whenever the popup is focusable — dismissed or not — matching Compose's rule that any focusable popup consumes back events. A non-focusable popup no longer swallows a key it did not act on.
* `JBPopupRenderer` pushes property changes onto the live `AbstractPopup`. It captured them once at creation, so a popup whose properties change could not be honoured at all.
* `GotItTooltip` sets `dismissOnBackPress = false`: it handles Escape on its anchor, so the renderer must keep its hands off the key.
* `SpeedSearchArea` expresses `dismissOnLoseFocus` as `dismissOnClickOutside`, so Escape always hides the search input whatever the outside-dismissal policy is.
* New Spectre parity test asserting the custom renderer is indistinguishable from Compose's own popup across ten scenarios, including popups with no component logic behind them, where the renderer's own decision is the only thing being observed. A control scenario fails the test if the Escape probe itself breaks, and renderer-owned scenarios carry absolute expectations, so the differential comparison cannot pass vacuously.

## Verification

No screenshots: nothing here changes how anything looks, only how Escape and outside clicks are routed. Measured in a dev IDE built from sources, headful under Xvfb, plus the standalone Spectre suite. Two bugs confirmed by measuring the code both with and without the fix:

| scenario | before | after |
| --- | --- | --- |
| Escape with the pointer over a ComboBox chevron | popup stays open, key swallowed | popup closes |
| Escape on a `GotItTooltip`, bridge renderer | tooltip stuck open, key swallowed | tooltip closes |
| Chevron click on an open popup | closes, does not reopen | closes, does not reopen |
| Menus, submenus, speed search, editable combo box | — | unchanged |

`JBPopupRenderer` and Compose's own renderer now report identical Compose node counts across a full open/close/reopen cycle, and the parity test holds for all ten scenarios on both renderers.

### ⚠️ Important Changes

* The custom popup renderers now honour `PopupProperties.dismissOnBackPress`, which they previously ignored. A popup that sets it to `false` will no longer be dismissed by Escape, and will no longer consume the key unless it is focusable. Popups that relied on the previous unconditional Escape dismissal should set it explicitly.

### Bug fixes

* Fixed Escape not closing a `ComboBox` or `EditableComboBox` popup while the pointer was over the component. The key was also swallowed, so it could not reach the surrounding dialog.
* Fixed `GotItTooltip` not closing on Escape in the IDE, where the key was consumed by the popup renderer before the tooltip's own handler could run.

### DIFF
--- a/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
+++ b/platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
@@ -45,6 +45,7 @@ import com.intellij.openapi.ui.popup.JBPopupListener
 import com.intellij.openapi.ui.popup.LightweightWindowEvent
 import com.intellij.ui.ScreenUtil
 import com.intellij.ui.awt.RelativePoint
+import com.intellij.ui.popup.AbstractPopup
 import com.intellij.ui.scale.JBUIScale
 import java.awt.Component
 import java.awt.Dimension
@@ -118,6 +119,8 @@ private fun JBPopup(
     val currentContent = rememberUpdatedState(content)
     val currentPopupPositionProvider by rememberUpdatedState(popupPositionProvider)
     val currentOnDismissRequest by rememberUpdatedState(onDismissRequest)
+    val currentOnPreviewKeyEvent by rememberUpdatedState(onPreviewKeyEvent)
+    val currentOnKeyEvent by rememberUpdatedState(onKeyEvent)
     val currentProperties by rememberUpdatedState(properties)
 
     val owner = LocalComponent.current
@@ -182,17 +185,29 @@ private fun JBPopup(
             .setCancelOnClickOutside(currentProperties.dismissOnClickOutside)
             .setCancelOnWindowDeactivation(currentProperties.dismissOnClickOutside)
             .setLocateWithinScreenBounds(false)
+            // AbstractPopup cancels itself on Escape whenever the handler below declines the event, which would
+            // close the popup behind Compose's back and leave the two out of sync. Take sole ownership instead.
+            .setCancelKeyEnabled(false)
             .setKeyEventHandler { event ->
                 val composeEvent = event.toComposeKeyEvent()
                 val consumed =
-                    onPreviewKeyEvent?.invoke(composeEvent) == true || onKeyEvent?.invoke(composeEvent) == true
+                    currentOnPreviewKeyEvent?.invoke(composeEvent) == true ||
+                        currentOnKeyEvent?.invoke(composeEvent) == true
+                // Whether Escape dismisses is the popup's declared policy, not something inferred from the shape
+                // of its callbacks. A null callback means we have no way to dismiss, so we must not claim the key.
+                val dismissed =
+                    !consumed &&
+                        composeEvent.isDismissRequest() &&
+                        currentProperties.dismissOnBackPress &&
+                        currentOnDismissRequest != null
 
-                if (!consumed && composeEvent.isDismissRequest()) {
+                if (dismissed) {
                     isVisible = false
-                    true
-                } else {
-                    consumed
                 }
+                // A focusable popup swallows Escape whether or not it dismissed, matching Compose. A
+                // non-focusable one must let a key it did not act on reach whatever owns it.
+                val claimsDismissRequest = composeEvent.isDismissRequest() && currentProperties.focusable
+                consumed || dismissed || claimsDismissRequest
             }
             .addListener(
                 object : JBPopupListener {
@@ -206,6 +221,15 @@ private fun JBPopup(
                 }
             )
             .createPopup()
+    }
+
+    // The builder above only captures the properties as they were on first composition. AbstractPopup reads these
+    // flags live on every event, so pushing changes keeps a popup whose properties vary (a ComboBox suppressing
+    // click-outside dismissal while its chevron is hovered) in step without recreating the native window.
+    LaunchedEffect(popup, currentProperties) {
+        val abstractPopup = popup as? AbstractPopup ?: return@LaunchedEffect
+        abstractPopup.setCancelOnClickOutside(currentProperties.dismissOnClickOutside)
+        abstractPopup.setCancelOnWindowDeactivation(currentProperties.dismissOnClickOutside)
     }
 
     val rectValue = popupRectangle.value
--- a/platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
+++ b/platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
@@ -7,6 +7,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.text.input.TextFieldState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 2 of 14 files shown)


rock3r:
Back in draft as I investigate a potential further issue

rock3r:
all fine, good to review

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
```diff
@@ -182,17 +185,29 @@ private fun JBPopup(
...
-
-                if (!consumed && composeEvent.isDismissRequest()) {
+                    currentOnPreviewKeyEvent?.invoke(composeEvent) == true ||
+                        currentOnKeyEvent?.invoke(composeEvent) == true
+                // Whether Escape dismisses is the popup's declared policy, not something inferred from the shape
+                // of its callbacks. A null callback means we have no way to dismiss, so we must not claim the key.
+                val dismissed =
+                    !consumed &&
+                        composeEvent.isDismissRequest() &&
+                        currentProperties.dismissOnBackPress &&
```
DanielSouzaBertoldi, line 201: Shouldn't this be called `dismissOnEscapePress`? "Back" to me reads more like "backspace pressed" than "escape"

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
```diff
@@ -97,21 +107,101 @@ class CustomPopupRendererSpectreTest {
...
+            app.stop()
+        }
+    }
+}
+
+/**
+ * Waits until [find] matches nothing, across every window Spectre tracks.
+ *
+ * Spectre only ships a wait-for-presence helper, but a popup lives in its own native window, and these tests need to
+ * watch that window go away.
```
DanielSouzaBertoldi, line 189: I wonder if Spectre will ship something akin to this helper + others like:

```kt
private fun ComposeAutomator.hasTag(tag: String) = findByTestTag(tag).isNotEmpty()

private fun ComposeAutomator.hasText(text: String) = findByText(text).isNotEmpty()
```

someday.... 👀

wellingtoncosta (approved):

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/PopupRendererParitySpectreTest.kt
```diff
@@ -0,0 +1,349 @@
...
+    val bounds = a.waitForNode(tag = OPEN_TAG).boundsOnScreen
+    driver.click(bounds.x + bounds.width - 8, bounds.y + bounds.height / 2)
+    a.waitForNode(tag = POPUP_TAG)
+}
+
+private val CASES =
+    listOf(
+        // Control: no popup at all. If this does not report hostSawEscape=true, the probe itself is broken
+        // and every other hostSawEscape reading in this run is meaningless.
+        Case(
```
DanielSouzaBertoldi: Genuine question, because I really don't get it: how do we really know if the probe is broken?

Because like, if it's broken for the Compose Popup, it should also be for our JDialog, right?

In this case, both would set `hostSawEscape = false` and then the list of divergences (which is the only thing the `assertEquals` checks) will be empty regardless. I don't think that we can know if the probe is really false 🤔

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRenderer.kt
rock3r: It's Compose's name, not ours 🙂 `dismissOnBackPress` is `androidx.compose.ui.window.PopupProperties`' own property, documented as covering "the back or escape buttons on Android or the escape key on desktop". The point of this PR is that the renderers honour Compose's declared policy, so we read the property as upstream named it — a local alias would only obscure that. Agreed the name is Android-flavoured.

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
rock3r: From your lips to Spectre's ears 👀 `waitUntilGone` just above is already a private copy waiting on Spectre #438, and `hasTag`/`hasText` (in the parity test) would be natural companions. Happy to file an issue for them.

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/PopupRendererParitySpectreTest.kt
rock3r: Fair point — the control was a canary only in the intentions lol. The assertion is purely differential in nature, so a dead probe (e.g., `hostSawEscape` stuck to false) would have passed, exactly as you say. Bad miss on my part on the pre-PR review. I won't blame the agent for it, I should have caught it.

K3 fixed it in e675b1c; it says:

> The control case is now an explicit gate — it must report `hostSawEscape=true` on both renderers before the matrix runs, so a broken probe fails the test outright. The raw popup cases now carry absolute expectations too, since the renderer's own decision is the only thing being observed there. That includes a new `rawPopup(nonFocusable)` scenario proving an unconsumed Escape is forwarded to the host while a native popup window exists (`popupOpen=true hostSawEscape=true`) — the gap the no-popup control cannot cover.

Full Spectre suite green on both renderers.

EDIT: sorry this was posted without validation/rephrasing by mistake (bad, K3!) I have cleaned it up now

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
rock3r: Actually since I am preparing 0.6.0 why not doing it anyway :D

rock3r: Filed: rock3r/spectre#463 — `hasTag`/`hasText` presence helpers, sibling to the `waitUntilGone` issue.

rock3r: (that last one was K3, it keeps posting without asking me nor me telling it to)

DanielSouzaBertoldi (approved):
All my concerns have been addressed, thanks! ❤️

wellingtoncosta (approved):

rock3r:
Ready to merge

PR #3623, opened 2026-08-21
[JEWEL-1407] Use the official product name, Compose Multiplatform

rock3r:
JetBrains retired "Compose for Desktop": [kotlinlang.org/compose-multiplatform](https://kotlinlang.org/compose-multiplatform/) and the [compose-multiplatform](https://github.com/JetBrains/compose-multiplatform) readme both use "Compose Multiplatform" as the product name and plain "desktop" for the target, and neither uses the old name anywhere.

The readme used the retired name 13 times, including its title and two `CfD` abbreviations, while linking to jewel-ui.dev where every page uses the official one. Desktop is still named where it carries information, as "Compose Multiplatform for desktop" and "a Compose Multiplatform desktop app".

This was part of JEWEL-1398 but did not make it into the cherry-pick, so the rest of that work landed without it.


DanielSouzaBertoldi (approved):

rock3r:
Ready to merge

wellingtoncosta (approved):
Very tough to review, but approved

PR #3624, opened 2026-08-21
[JEWEL-1404] Clean up Jewel's codebase

DanielSouzaBertoldi:
**DO NOT FEAR THE LOC COUNT!!!!!!!!**

I ran the `generateAndroidStudioReleasesList` task and tweaked the code a bit now that each AS release has the name of an animal followed by a digit (i.e, Quail 1, Quail 2, Quail 3), the previous code ended up discarding them, blablabla I added more details down below.

90% of the LOC is just the splash screen list being updated :)

# Context

As we bump library versions, some APIs that were experimental turned stable, so they don't require an annotation anymore. Kotlin is always improving and shipping new features as well, which we can then change our code slightly to make use of those features.

But honestly, I just checked each file under `plaftorm/jewel` and verified if there were any warnings. If there were, I'd fix them if it made sense to do so, otherwise I would just ignore it. I'd say that 90% of all warnings should now be gone, except for cases where the IDE would suggest refactoring the code to something much more weirder to read. For instance:

<table>
<tr>
<td align="center"> How it is </td>
<td align="center"> How the IDE likes it </td>
</tr>
<tr>
<td>

```kt
/**
 * In SegmentedControl.kt, this code has this warning:
 * Introduce 'KeyEventType.KeyUp' as subject of 'when'
 */
when {
    KeyEventType.KeyUp == it.type && Key.DirectionRight == it.key -> {
        if (selectedButtonIndex < buttons.size - 1) {
            buttons[selectedButtonIndex + 1].onSelect()
        }
        true
    }

    KeyEventType.KeyUp == it.type && Key.DirectionLeft == it.key -> {
        if (selectedButtonIndex > 0) {
            buttons[selectedButtonIndex - 1].onSelect()
        }
        true
    }

    else -> false
}
```

</td>
<td>

```kt
/**
 * If we accept the automatic change, we end up with this:
 */
when (KeyEventType.KeyUp) {
    it.type if Key.DirectionRight == it.key -> {
        if (selectedButtonIndex < buttons.size - 1) {
            buttons[selectedButtonIndex + 1].onSelect()
        }
        true
    }
    it.type if Key.DirectionLeft == it.key -> {
        if (selectedButtonIndex > 0) {
            buttons[selectedButtonIndex - 1].onSelect()
        }
        true
    }
    else -> false
}
```

</td>
 </tr>
</table>

Which to me reads kinda weird. Though I did accept the same change but in another file so please let me know if you find that code awkward (it would also prove that you read this description so far 😈)

### Warning cleanup

* Dropped `@OptIn` and `@Suppress` annotations whose targets are now stable or no longer warn.
* `(x?.hashCode() ?: 0)` → `x.hashCode()`. The stdlib `Any?.hashCode()` already returns `0` for null.
* Subject-less `when` blocks now use a subject plus guard conditions.
* `Long` millisecond parameters moved to `kotlin.time.Duration`.
* `DisabledAppearance`: `nativeCanvas` → `skiaCanvas`, after the upstream rename.
* Fixed KDoc links pointing at renamed or unreachable symbols. `IconUtils` KDoc sat below its annotations, so it wasn't attached to anything; moved it above.
* Repointed 19 sample links from the deleted `samples/standalone/…/view/component/` to `samples/showcase/`.
* Deleted dead code: an uncalled private helper, two test assertions nobody ever used, an orphaned `package-info.java`.
* `SwingBridgeReader` rethrows `CancellationException` instead of retrying through it, and catches `Exception` rather than `Throwable`.
* `ImageSourceResolver` guards `Path.of()`, so a malformed destination falls through to the next capability instead of throwing out of `resolve()`.
* `InlineBanner`: added the missing `@Nls` on `title`.
* `RELEASE NOTES.md`: Standalone Version column for v0.35–v0.39.1. Master already covers v0.28–v0.34.

### Android Studio releases generator

* Made the output class, model package and indent configurable, so the generator can write somewhere other than the deleted sample's package.
* Applied it on the root build, pointed at DevKit's source tree.
* Fixed the splash lookup. Releases from Panda onward carry a drop counter (`Quail 3`) that tripped the digit guard, so everything new came out with no image.
* Stripped the counter from `displayText` too: entries now read `Android Studio Quail | 2026.1.3`. Legacy version-only names keep their numbers.
* Added the eight missing splash screens (Otter, OtterFeatureDrop, Panda, Quail).
* Regenerated `AndroidStudioReleases.kt`.

### Devkit Stuff

* Added a bunch of missing strings to `DevkitComposeBundle.properties`, removed all hardcoded strings
* Changed `@Nsl` to `@NslSafe` in the `AndroidStudio` data class because it's just used for a cool demo. If we were to add strings for each release in the properties file then we'd be screwed.
* Added `ExperimentalJewelApi` annotations in files that where missing it
* Found a crash when an `InlineBanner` tries to calculate its width with unbound constraints, filed [JEWEL-1415](https://youtrack.jetbrains.com/issue/JEWEL-1415/InlineBanner-crashes-when-measured-with-an-unbounded-width). For now I'm just setting an arbitrary width so users can still compare both swing/jewel UIs.
* Minor corrections suggested by the IDE such as using `Duration`, replacing references to deprecated APIs and using `JBUI.Borders.empty()` instead of `null`.

`ui/generated/…/DefaultColorPalette.kt` is regenerated output. The committed copy had drifted from the theme JSONs on master.

Also, some of the changes were made by the Bazel version of ktfmtFormat, so that's why some files only add blank lines and whatnot

### IMPORTANT NOTE!

This PR does NOT replace usages of the deprecated `SelectableLazyColumn` throughout our components (such as in `SpeedSearchableLazyColumn.kt` because doing so would require a really big refactor of the code, plus we gotta be careful with this one give that the new APIs had a very sneaky bug (though it was fixed ASAP back then).

Watch out.

### ⚠️ Important Changes

* `ImageSourceResolver.ResolveCapability.AbsolutePath` and `RelativePath` now return `null` when a destination can't be parsed as a path, rather than letting `InvalidPathException` escape `resolve()`. Resolution falls through to the next capability.

### Bug fixes

* The IDE theme bridge no longer swallows `ProcessCanceledException` while retrying to read theme data


platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChanges.kt
```diff
@@ -1,4 +1,5 @@
 @file:Suppress("RAW_RUN_BLOCKING", "SSBasedInspection")
+
```
DanielSouzaBertoldi, line 2: Any new blank lines added or indentation fixes were made by the Bazel's version of `ktfmtFormat` :)

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/demodata/AndroidStudioReleases.kt
```diff
@@ -68,6 +80,12 @@ open class AndroidStudioReleasesGeneratorTask : DefaultTask() {
 
     @get:Input val resourcesDirs = project.objects.setProperty<File>()
 
+    @get:Input val outputClassName = project.objects.property<String>()
+
+    @get:Input val modelPackage = project.objects.property<String>()
+
+    @get:Input val indent = project.objects.property<String>()
```
DanielSouzaBertoldi, line 87: The reason why I opted to make these properties configurable, is so the generator can write somewhere other than the deleted sample's package. The release list JSON was moved to the devkit package but the generator itself never got updated.

Now it doesn't matter, we can run the task directly passing the necesary paths and etc when running the task.

platform/jewel/buildSrc/src/main/kotlin/org/jetbrains/jewel/buildlogic/demodata/AndroidStudioReleasesReader.kt
```diff
@@ -117,6 +133,18 @@ internal object AndroidStudioReleasesReader {
         return "\"$splashPath\""
     }
 
+    // Releases starting from Panda carry a digit along their name (e.g., "Quail 3"), so we remove those
+    private fun String.withoutDropCounter(): String =
+        split(' ').filterNot { it.isEmpty() || it.all(Char::isDigit) }.joinToString(separator = " ")
+
```
DanielSouzaBertoldi, line 139: That's a personal choice of mine, I think it's best to list something like:

```
Android Studio Panda | 2025.3.4
Android Studio Panda | 2025.3.3
...
```

Instead of:


```
Android Studio Panda 4 | 2025.3.4
Android Studio Panda 3 | 2025.3.3
...
```

The IDE build number is enough to discern between them imo

platform/jewel/foundation/src/main/kotlin/org/jetbrains/jewel/foundation/lazy/tree/Tree.kt
```diff
@@ -154,7 +154,7 @@ public class Tree<T> internal constructor(public val roots: List<Element<T>>) {
              */
             public fun close() {
                 detachChildren()
-                children?.asSequence()?.filterIsInstance<Node<*>>()?.forEach { it.closeRecursively() }
+                closeRecursively()
```
DanielSouzaBertoldi, line 157: `closeRecursively` body is:

```kt
children?.forEach { if (it is Node) it.closeRecursively()
```

Which does exactly the same thing. It iterates `children`, picks out `Node`s and recurse into each.

platform/jewel/ide-laf-bridge/src/main/kotlin/org/jetbrains/jewel/bridge/SwingBridgeReader.kt
```diff
@@ -54,10 +55,14 @@ internal class SwingBridgeReader {
         while (counter < 20) {
             delay(20.milliseconds)
             counter++
-            runCatching { readThemeData() }
```
DanielSouzaBertoldi: The warning here was:

> Do not use runCatching in CLion code

The fix itself is to just get rid of `runCatching` and turn this into a `try`/`catch` block, but then while double checking my fixes with Claude it raised the concern that having just one `catch(_: Exception)` was just too broad and we should respect a `ProcessCancelledException` exception.

[Per the doc](https://www.jetbrains.com/help/inspectopedia/IncorrectCancellationExceptionHandling.html):

> `ProcessCanceledException` and its inheritors must not be caught, swallowed, logged*, or handled in any way. Instead, it must be rethrown so that the IntelliJ Platform infrastructure can handle it correctly.

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/view/component/package-info.java
DanielSouzaBertoldi, line 1: When we moved the samples to `samples/showcase/…/showcase/components/` this package was left behind. Nothing references it, so I just removed it.

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/Assertions.kt
```diff
@@ -4,14 +4,7 @@ import androidx.compose.ui.semantics.SemanticsProperties
...
 import androidx.compose.ui.test.assert
-import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.TextRange
 
-fun SemanticsNodeInteraction.assertEditableTextEquals(expected: String): SemanticsNodeInteraction =
-    assert(SemanticsMatcher.expectValue(SemanticsProperties.EditableText, AnnotatedString(expected)))
-
-fun SemanticsNodeInteraction.assertEditableTextEquals(expected: AnnotatedString): SemanticsNodeInteraction =
-    assert(SemanticsMatcher.expectValue(SemanticsProperties.EditableText, expected))
-
```
DanielSouzaBertoldi: We had this helpers for a long time and no one used it. There's no harm in keeping them but well, I'm taking no witnesses today 😛

platform/jewel/ui/generated/theme/org/jetbrains/jewel/ui/theme/DefaultColorPalette.kt
DanielSouzaBertoldi, line 1: Running `detekt` (that also runs `intellij-color-palette-generator.gradle.kts`) updated this file. Seems like `ManyIslandsLight/Dark.theme.json` changed and our palette went stale...

rock3r:
> if you find that code awkward

Kinda but not more than the older syntax tbh

rock3r (approved):
YAY SPRING CLEANING

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
```diff
@@ -92,7 +92,7 @@ private const val BANNER_CONTENT_SPACING = 8
 public fun InlineInformationBanner(
     @Nls text: String,
     modifier: Modifier = Modifier,
-    title: String? = null,
+    @Nls title: String? = null,
```
wellingtoncosta, line 95: I missed that one when I was working on JEWEL-1070

wellingtoncosta (approved):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/InlineBanner.kt
DanielSouzaBertoldi: No worries fam, we are always supporting each other here 😎

DanielSouzaBertoldi:
@rock3r @wellingtoncosta just letting you guys know that the rabbit splash screen for [canary builds image has been added](https://github.com/JetBrains/intellij-community/pull/3624/changes#diff-df9437a03e6775468ea5c7edf11f460a6195e04a610b1dae187af95f1d77e0de), and the [android release list was updated once more](https://github.com/JetBrains/intellij-community/pull/3624/changes#diff-df9437a03e6775468ea5c7edf11f460a6195e04a610b1dae187af95f1d77e0de) 🙏🏻

DanielSouzaBertoldi:
As soon as the pipeline finishes without any errors I'll merge this :)

DanielSouzaBertoldi:
Ready to merge

DanielSouzaBertoldi:
@rock3r @wellingtoncosta sorry guys, will need another round of code review from y'all. When I tried to merge this, merge bot got mad with a bunch of "Internationalization: Hardcoded strings" errors on the devkit side.

So I was like, screw it, let me fix stuff on there too. Just skip everything under `platform/jewel` (nothing changed) and focus on changes under `plugins/devkit` please 🙏🏻 

To make it easier, here's an overview of the major changes:

- Added a bunch of missing strings to `DevkitComposeBundle.properties`, removed all hardcoded strings
- Changed `@Nsl` to `@NslSafe` in the `AndroidStudio` data class because it's just used for a cool demo. If we were to add strings for each release in the properties file then we'd be screwed.
- Added `ExperimentalJewelApi` annotations in files that where missing it
- Found a crash when an `InlineBanner` tries to calculate its width with unbound constraints, filed [JEWEL-1415](https://youtrack.jetbrains.com/issue/JEWEL-1415/InlineBanner-crashes-when-measured-with-an-unbounded-width). For now I'm just setting an arbitrary width so users can still compare both swing/jewel UIs.
- Minor corrections suggested by the IDE such as using `Duration`, replacing references of deprecated APIs and using `JBUI.Borders.empty()` instead of `null`.

Thank you for your patience 😎

plugins/devkit/intellij.devkit.compose/src/demo/LoremData.kt
DanielSouzaBertoldi, line 1: Both strings here were moved to DevkitComposeBundle.properties

rock3r (approved):

wellingtoncosta (approved):

PR #3627, opened 2026-08-21
IJPL-245730: Make startup error text selectable

TR1028:
Wrap the success message in prepareMessage() so the text can be selected and copied by the user.


trespasserw:
Thanks, but the problem is fixed already in 046d723.

PR #3628, opened 2026-08-21
IJPL-2176 Support inode aliases in Linux file watcher

wilx:
## Issue

[IJPL-2176 – File watcher failed to start ("table error: collision" messages in the log)](https://youtrack.jetbrains.com/issue/IJPL-2176/File-watcher-failed-to-start-table-error-collision-messages-in-the-log)

## Summary

- allow Linux inotify watches to represent multiple path aliases for the same filesystem inode instead of terminating on a watch-descriptor collision
- revalidate aliases during self-events and periodic reconciliation so unlink, rename, atomic replacement, descendant removal, and mount-topology changes retain or rebuild the correct watches
- add native and IDE regression coverage for hard-link lifecycle transitions, registration races, disappearing descendants, and detached bind mounts

## Testing

- Clang CMake build: passed
- CTest: 1/1 passed
- privileged native suite with registration-race and bind-mount coverage: 11/11 passed
- Bazel `com.intellij.openapi.vfs.local.FileWatcherTest`: 45 discovered, 36 passed, 9 platform-specific assumption skips
- `git diff --check`: passed

### DIFF
--- a/native/fsNotifier/linux/CMakeLists.txt
+++ b/native/fsNotifier/linux/CMakeLists.txt
@@ -16,6 +16,19 @@ target_compile_definitions(fsnotifier PRIVATE VERSION="${VERSION}")
 target_compile_options(fsnotifier PRIVATE "-O2" "-Wall" "-Wextra" "-Wpedantic")
 target_link_options(fsnotifier PRIVATE "-static" "-flto=full")
 
+include(CTest)
+if(BUILD_TESTING)
+    target_compile_definitions(fsnotifier PRIVATE FSNOTIFIER_TESTING)
+    find_package(Python3 REQUIRED COMPONENTS Interpreter)
+    add_test(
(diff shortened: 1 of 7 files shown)


trespasserw:

Thanks for your contribution!

canepan:
I hope "soon" really means close in the future, because this issue has been there for at least 4 years: would it hurt to fix it and then progress with the rewrite?

Nodens-:
According to this: [youtrack.jetbrains.com link](https://youtrack.jetbrains.com/projects/IJPL/issues/IJPL-2176/File-watcher-failed-to-start-table-error-collision-messages-in-the-log)
the issue exists for at least 12 years... I have several C++ projects that are hitting this due to dependency hard links used in Fedora official packages. It would be much appreciated if this actually landed until the subsystem is rewritten.

PR #3629, opened 2026-08-21
[gradle-maven] don't suppress completion for non-coordinate literals in dependencies blocks

joshfriend:
`MavenDependenciesGradleCompletionContributor` called `stopHere()` for any string literal inside a dependencies block, including arguments of nested calls like `project(':path')` and `files('libs/dep.jar')` that are not maven coordinates. This suppressed completion from all other contributors for those literals. Skip such literals instead, and give the contributor registration an id so plugins can order relative to it.

My [spotlight plugin](https://github.com/joshfriend/spotlight) offers autocomplete contributions for `project(...)` and the gradle-maven plugin has broken that in 2026.2+


nskvortsov:
Apologize for the inconvenience caused, and thank you for your contribution!
Merged as 590762b986dcf5a4e82a93e14cfd2083adec3a34

PR #3633, opened 2026-08-29
IDEA-389429 Do not trigger auto-make for removals outside the project content

christianaurichzm:
## Problem

`BuildManager` decides whether VFS changes should trigger auto-make from `BulkFileListener.after`.

For a removal, this is too late. The `VirtualFile` has already been invalidated, so `ProjectFileIndex` can no longer determine its previous location. The existing code treats every invalid file as relevant:

```java
if (!eventFile.isValid()) {
  return true;
}
```

Removals that cannot affect compilation can therefore schedule a build. This includes files under excluded compiler output roots, ignored directories, and paths outside project content.

This is one source of the spurious auto-make scheduling observed in IDEA-389429. It does not account for every auto-make trigger during an external Maven build.

## Fix

Classify the removal side of VFS changes before they are applied.

A small `AsyncFileListener` handles `VFileDeleteEvent` and the source side of `VFileMoveEvent` from `prepareChange`. At this point the file still exists, so the file index can classify its original location. The listener schedules auto-make from `ChangeApplier.afterVfsChange()` when the removal is relevant.

The existing `BulkFileListener.after` path stays responsible for creations, content changes, copies, and the destination side of moves. Its event buffer and background executor are unchanged. The fallback that treated every invalid file as relevant is removed.

Before-state classification is limited to removals because their original location is lost after the VFS change. Other events remain on the existing post-VFS path.

## Regression coverage

`AutoMakeTriggerTest` covers both phases:

```text
removals
  source file                       -> triggers
  source root                       -> triggers
  compiler output file              -> does not trigger
  compiler output root              -> does not trigger
  excluded target root (mvn clean)  -> does not trigger
  ignored .git file                 -> does not trigger
  file outside project content      -> does not trigger

creations
  source file                       -> triggers
  compiler output file              -> does not trigger
```

Restoring the old rule that every removal triggers auto-make makes exactly the five negative removal cases fail: 9/9 becomes 4/9.

```bash
./tests.cmd --module intellij.java.compiler.tests.main \
  --test com.intellij.compiler.server.AutoMakeTriggerTest
```

## Verification

```text
AutoMakeTriggerTest           9 passed
Neighbouring compiler tests  23 passed
com.intellij.compiler.*      271 passed, 12 environment-related failures
```

The same 12 failures reproduce unchanged on `master`. Eleven require `intellij-deps-fastutil` in the local Maven repository. `CompilerBuildViewTest` expects a build step that this checkout does not produce.

A content-to-content move schedules auto-make from both its removal and addition sides. `BuildManagerPeriodicTask.schedule()` coalesces the requests, so two schedules result in one build.

## Scope

This change fixes incorrect auto-make scheduling for removals whose previous location cannot be classified in the post-VFS listener.

It does not detect Maven processes launched from the integrated terminal and does not claim to resolve IDEA-389429 in full.

Replaying the file operations from a real `mvn clean install -T1C` confirmed that removing `target` no longer triggers auto-make. Recreating an excluded `target` root can still be reported by `ProjectFileIndex` as project content until its excluded-root state is revalidated.

That is a separate project-model issue and is not changed by this patch.


christianaurichzm:
Rebased onto current master and force-pushed. Master changed the same auto-make trigger code in
3d863f8d5dd9 (IDEA-392693) and 0470a9d4cd76 (IDEA-393292), which caused the conflicts.

PR #3634, opened 2026-08-31
[JEWEL-1416] Match GroupHeader dividers with Swing

rock3r:
Align `GroupHeader` dividers with Swing `TitledSeparator`.

## Changes

* Read the group separator colors from the correct Look and Feel key, with the same fallback as Swing.
* Add enabled and disabled divider styling while retaining the previous binary API.
* Align the standalone light and dark divider defaults with Swing.
* Fix excessive spacing between the label and divider compared to Swing.
* Update the IntelliJ API dump and Jewel Metalava signatures.

Note: Jewel implements the New UI theme, not the Islands themes (for now), and Islands has a different colour for dividers.

## Screenshots

 Before | After
 --- | ---
 image | image

### Swing vs Jewel

| Light | Dark |
| --- | --- |
| image | image |

### New features

* Add enabled and disabled states to `GroupHeader` dividers.

### Bug fixes

* Align `GroupHeader` divider appearance with IntelliJ Swing group headers.

### DIFF
--- a/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiGroupHeaderStyling.kt
+++ b/platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiGroupHeaderStyling.kt
@@ -22,12 +22,14 @@ public fun GroupHeaderStyle.Companion.dark(
 ): GroupHeaderStyle = GroupHeaderStyle(colors, metrics)
 
 /** Creates an Int UI light [GroupHeaderColors] with the provided parameters. */
-public fun GroupHeaderColors.Companion.light(divider: Color = IntUiLightTheme.colors.gray(12)): GroupHeaderColors =
-    GroupHeaderColors(divider)
+public fun GroupHeaderColors.Companion.light(
+    divider: Color = IntUiLightTheme.colors.grayOrNull(12) ?: Color(0xFFCDCDCD)
+): GroupHeaderColors = GroupHeaderColors(divider)
 
 /** Creates an Int UI dark [GroupHeaderColors] with the provided parameters. */
-public fun GroupHeaderColors.Companion.dark(divider: Color = IntUiDarkTheme.colors.gray(3)): GroupHeaderColors =
-    GroupHeaderColors(divider)
+public fun GroupHeaderColors.Companion.dark(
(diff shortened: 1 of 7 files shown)


platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiGroupHeaderStyling.kt
```diff
@@ -22,12 +22,14 @@ public fun GroupHeaderStyle.Companion.dark(
 ): GroupHeaderStyle = GroupHeaderStyle(colors, metrics)
 
 /** Creates an Int UI light [GroupHeaderColors] with the provided parameters. */
-public fun GroupHeaderColors.Companion.light(divider: Color = IntUiLightTheme.colors.gray(12)): GroupHeaderColors =
-    GroupHeaderColors(divider)
+public fun GroupHeaderColors.Companion.light(
+    divider: Color = IntUiLightTheme.colors.grayOrNull(10) ?: Color.Unspecified
```
DanielSouzaBertoldi: Why is this `grayOrNull(10)`? Looking at the `expUI_light.theme.json` file the only `separatorColor` tokens that reference `Gray10` are `LicenseList` and `LicenseServer`.

The "main" `separatorColor` from `*` references `Gray12`, I can't seem to find any references to `GroupHeader` 🤔 

The same question applies to the dark variant below

rock3r:
Updated PR description too

platform/jewel/int-ui/int-ui-standalone/src/main/kotlin/org/jetbrains/jewel/intui/standalone/styling/IntUiGroupHeaderStyling.kt
rock3r: You were right, I got the wrong keys because I thought the exact match of `Group.separatorColor` in the parent chain (which exist in intellijlaf and darcula) would take precedence over the theme's `*` keys. This will never finish tripping me up. I also wasted a good amount of time due to the fact that Islands changes the separator colour compared to Int UI. Oh well. I at least fixed the double spacing issue we had while I was at it :)

DanielSouzaBertoldi (approved):
Let's goooooo 🚀

rock3r:
Ready to merge, then :)

PR #3636, opened 2026-09-01
[JEWEL-1392] Gradle Scripts to Bazel Migration - Part 4

DanielSouzaBertoldi:
## Context

Continues the Jewel Gradle → Bazel scripts migration started in Part 3 ([JEWEL-1360](https://github.com/JetBrains/intellij-community/pull/3602)). Ports six more scripts to Bazel: `intellij-theme-generator`, `intellij-color-palette-generator`, `android-studio-releases-generator`, `compare-performance`, `extract-release-notes`, and `run-performance-tests`. 

The last one didn't have a working Bazel version to port from because the original always shelled out to `./gradlew`, which the Bazel port can't touch at all, so we had to get a bit creative by using Bazel's own tools to achieve the same results:

* `ComparePerformance.kt` shelled out to the `jfr` CLI to read `.jfr` files. Under `bazel run`, the launched process uses a hermetic JBR toolchain whose `bin/` isn't on `PATH`, so this failed with `Cannot run program "jfr"` every time. [We asked the Bazel team about a PATH-based fix](https://github.com/JetBrains/intellij-community/pull/3636#discussion_r3905529466). They pushed back on PATH/`JAVA_HOME` assumptions and pointed out that `java.home` only guarantees the *current* JVM, not necessarily the one that produced the file being read.
* `--per-test` mode (one JFR file per test method) had no Bazel-native equivalent, since the original script discovered test methods via `--test-dry-run` and looped, and `JUnit5BazelRunner` doesn't expose discovery-without-execution. [The Bazel team's answer reframed the problem: method discovery isn't actually needed here](https://github.com/JetBrains/intellij-community/pull/3636#discussion_r3905550604), and we aren't tied to `JUnit5BazelRunner` for this.

Both problems come down to the same root cause: shelling out. This PR replaces both subprocess calls with the JDK's own `jdk.jfr` APIs used in-process.

## Changes

* `ComparePerformance.kt` now reads `.jfr` files via `jdk.jfr.consumer.RecordingFile` (new `JfrEventReader.kt`) instead of parsing `jfr print --json` output, dropping the `jfr`-on-`PATH` dependency entirely. Found a real latent bug along the way: `jdk.GCHeapSummary`'s `committedSize` lives nested under its `heapSpace` field, not at the top level, so the old JSON-based code always read 0 for it.
* `RunPerformanceTests.kt`'s `--per-test` mode is implemented for real now: a new `JfrTestExecutionListener` opens one `jdk.jfr.Recording` per test method through the JUnit Platform Launcher API directly and closes it when the test finishes. No discovery-without-execution step needed.
* Both `kt_jvm_library` targets touching `jdk.jfr.*` needed a `kotlinc_opts` bump to `jvm_target = "21"`. `rules_kotlin`'s toolchain-wide default of `1.8` otherwise hides `jdk.jfr.consumer`'s APIs from Kotlin source, the same way `javac --release` would.
* `run-performance-tests`/`runPerformanceTests` are now `testonly = True`, required transitively through `ui-tests_test_lib` (confirmed this propagates through `runtime_deps`, not just `deps`). Makes sense given what the script does now: it drives test code directly, so testonly is accurate, not a workaround.
* `--duration` only applies to aggregate mode now; `--per-test` recordings are bounded by each test's own execution time instead of a fixed cap.
* Also fixed along the way: `CommandRunner`'s command-string parsing was a naive `command.split(" ")` with no shell involved, so quoting in a command string was either meaningless or actively wrong. Replaced with `tokenizeCommand()`, a real POSIX-shell-style word splitter (single/double quotes, backslash escaping in double quotes). `ComparePerformance.kt` and `ValidateMavenArtifacts.kt` both wrapped a user-supplied path in a bare `File(path)`; `bazel run` doesn't guarantee the launched process's working directory matches where the user invoked it from, so a relative path broke silently. Both now resolve through a shared `resolvePath()` helper against `getBuildWorkspaceDirectory()` when the path isn't already absolute.

## Known limitation

Time/CPU/GC-sampled JFR metrics (jdk.ExecutionSample, jdk.CPULoad, GC events) usually come back empty per test method in --per-test mode. Not a bug: `jdk.CPULoad` has a fixed 1-second period even under "profile" settings, and `jdk.ExecutionSample` runs off a shared periodic clock, so recordings this short (verified on both a fast and a ~68ms/test UI test class) rarely land a sample. Allocation-based metrics (jdk.ObjectAllocationSample) are threshold-triggered instead, so they're unaffected. Documented on `JfrTestExecutionListener`.


platform/jewel/build-scripts/bazel/BUILD.bazel
```diff
@@ -142,7 +273,7 @@ java_binary(
 )
 
 # Tests
-kt_jvm_test(
+kt_jvm_library(
     name = "bazel-test-utils",
     srcs = ["src/test/kotlin/org/jetbrains/jewel/scripts/bazel/BazelTestUtils.kt"],
```
DanielSouzaBertoldi, line 278: `BazelTestUtils` is just a utility class, not a test class itself so we need to declare it as a library instead :)

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AndroidStudioReleasesGenerator.kt
```diff
@@ -0,0 +1,242 @@
...
+private const val DATA_URL = "https://jb.gg/android-studio-releases-list.json"
+private const val OUTPUT_CLASS_NAME = "com.intellij.devkit.compose.demo.releasessample.AndroidStudioReleases"
+private const val MODEL_PACKAGE = "com.intellij.devkit.compose.demo.releasessample"
+private const val INDENT = "  "
+private const val OUTPUT_FILE_PATH =
+    "plugins/devkit/intellij.devkit.compose/src/demo/releasessample/AndroidStudioReleases.kt"
+private const val RESOURCES_DIR_PATH = "plugins/devkit/intellij.devkit.compose/resources"
+
+// Run it like: bazel run //platform/jewel/build-scripts/bazel:generateAndroidStudioReleasesList
+fun main(args: Array<String>) {
```
DanielSouzaBertoldi, line 37: Yes this port does mirror the changes made in [JEWEL-1404](https://github.com/JetBrains/intellij-community/pull/3624)

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AndroidStudioReleasesGenerator.kt
```diff
@@ -0,0 +1,242 @@
+@file:Suppress("IO_FILE_USAGE")
```
DanielSouzaBertoldi, line 1: The reason we have to suppress this warning:

> Use nio.Path instead of io.File 

Is because if we try to use `nio.Path`, we end up being forced to use IJP's own file utilities which requires us to depend on the IJP as a whole just to run standalone scripts.

in other words, not worth it.

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AnnotateApiDumpChanges.kt
```diff
@@ -95,23 +96,6 @@ internal class AnnotateApiDumpChangesCommand(private val commandRunner: CommandR
             printlnWarn("GITHUB_STEP_SUMMARY environment variable not set")
         }
     }
-
-    /**
-     * Requires the GitHub CLI tool (`gh`) to be present on the system's PATH. Exits the process with an error code if
-     * the tool is not found.
-     */
-    private suspend fun requireGhTool() {
```
DanielSouzaBertoldi, line 103: Moved to `CliktUtils`

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CheckApiDumps.kt
```diff
@@ -1,4 +1,4 @@
-@file:Suppress("RAW_RUN_BLOCKING", "SSBasedInspection")
```
DanielSouzaBertoldi, line 1: Suppressing `SSBasedInspection` is just too broad. I opted to update these suppressions with ones more specific.

Also, the reason we need to suppress `RAW_RUN_BLOCKING` is for the same reason as `IO_FILE_USAGE`. It would force us to use APIs from the IJP which is a big no-no for these scripts.

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/CheckApiDumps.kt
```diff
@@ -59,8 +60,8 @@ internal class CheckUpdatedCommand(private val commandRunner: CommandRunner = De
             val command = buildApiCheckCommand()
             if (verbose) println("Running: $command")
 
-            val outputFile = File.createTempFile("api-check-output", ".log")
-            val scriptFile = File.createTempFile("api-check-script", ".sh")
+            @Suppress("SSBasedInspection") val outputFile = File.createTempFile("api-check-output", ".log")
+            @Suppress("SSBasedInspection") val scriptFile = File.createTempFile("api-check-script", ".sh")
```
DanielSouzaBertoldi, line 64: For cases such as these where the `createTempFile` warning does not have a specific suppression ID I was forced to suppress `SSBasedInspection`.

At least they are scoped to these variables instead of the whole file.

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/RunPerformanceTests.kt
```diff
@@ -0,0 +1,244 @@
...
+
+        println("🚀 Running performance tests...".asSuccess())
+        println("   Test pattern: $testPattern")
+        println("   Heap size: $heapSize")
+        println()
+
+        val classNameFilters = buildClassNameFilters(testPattern)
+
+        // TODO: --per-test needs one JFR file per test method, which requires discovering tests without
+        //  running them. JUnit5BazelRunner already does that discovery internally, it's just not exposed.
```
DanielSouzaBertoldi, line 93: **QUESTION FOR THE BAZEL TEAM!**

This script in the Gradle version uses the option `--test-dry-run` so that it can have a list of all test methods a given module has.

 Checking the `JUnit5BazelRunner` code I can see we fetch the `TestPlan`:

`TestPlan testPlan = getTestPlan();`

So, would it be possible to add an env var (something like "JB_TEST_DRY_RUN") that, when true, could exit before running anything and just printing the testPlan's discovered `TestIdentifier`s?

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/ComparePerformance.kt
```diff
@@ -0,0 +1,1532 @@
...
+
+    internal suspend fun extractTopMethods(file: File): Map<String, Int> {
+        val result =
+            commandRunner(
+                command = "jfr print --events jdk.ExecutionSample ${file.absolutePath}",
+                workingDir = null,
+                exitOnError = false,
+                timeoutAmount = 30.seconds,
+            )
+
```
DanielSouzaBertoldi, line 536: **QUESTION FOR THE BAZEL TEAM!!!!**

The `jfr` command fails, which is weird because it's bundled with the JDK CLI tools. I asked Claude to understand why, and this is what it said:

---

The process is launched with JetBrains' hermetic JBR toolchain
(confirmed via the generated wrapper script: `JAVABIN=${JAVABIN:-${JAVA_RUNFILES}/+jbr_toolchains+remotejbr25_macos_aarch64/bin/java}`), and `jfr` genuinely exists right next to `java` in that exact same directory:

```
.../+jbr_toolchains+remotejbr25_macos_aarch64/bin/jfr   <- exists
.../+jbr_toolchains+remotejbr25_macos_aarch64/bin/java  <- what's actually launched
```

`java.home` on the running JVM correctly reports that directory. It's just not on the launched process's `PATH`, even though it's the directory `java` itself was launched from.

---

So, my question is: is this an oversight or is it intentional?

We could resolve `jfr` ourselves via `System.getProperty("java.home") + "/bin/jfr"` and move on (this works, I tested like this):

```kotlin
private val jfrExecutable: String by lazy {
    val javaHome = System.getProperty("java.home") ?: error("java.home system property is not set")
    val binaryName = if (isWindows) "jfr.exe" else "jfr"
    File(javaHome, "bin/$binaryName").absolutePath
}

// in this line I'm adding the comment, switching `jfr` for:
> command = "$jfrExecutable print --events jdk.ExecutionSample ${file.absolutePath}"
```

But that got me thinking. If there's a supported way to get the toolchain's `bin/` onto the child process's `PATH` (a `run_environment` field, a documented env var, a `--run_under` convention, anything), we'd rather use that than have every Jewel script that needs a JDK CLI tool independently reimplement `java.home`-based path
resolution.

Given that the tool itself is right there, I'm inclined to think there's no harm in exposing it as well. Wdty?

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/JewelBazelScriptsUtils.kt
```diff
@@ -186,6 +199,74 @@ object DefaultCommandRunner : CommandRunner {
...
 }
 
+/**
+ * Splits a command string into argv tokens the way a POSIX shell would for word-splitting and quote removal — no
+ * variable expansion, globbing, or command substitution (Kotlin string templates already handled any interpolation
+ * before this runs). Supports single quotes (`'...'`, no escapes) and double quotes (`"..."`, with backslash escaping
+ * of `\`, `"`, `$`, and `` ` ``), including quoted sections in the middle of a word (e.g. `--flag='some value'` is one
+ * token, `--flag=some value`).
+ */
+internal fun tokenizeCommand(command: String): List<String> {
```
DanielSouzaBertoldi, line 209: The naive way of fetching parameters (`command.split(" ")`) wasn't enough for scripts like `ComparePerformance.kt` and `RunPerformanceTests`, hence why we had to upgrade it to something more robust. Don't worry, I added tests to cover this tokenizer :)

DanielSouzaBertoldi:
@shalupov @mfilippov Hey guys! There are two important questions I added in this PR that I'd love to hear from you. We reached some roadblocks when migrating two scripts:

[One does not have an alternative to `--test-dry-run` from Gradle (though it seems possible to add something akin to that in `JUnit5BazelRunner`)](https://github.com/JetBrains/intellij-community/pull/3636#discussion_r3904681645).

[And the other fails because `bazel run` cannot find `jfr` in the toolchain](https://github.com/JetBrains/intellij-community/pull/3636#discussion_r3904815337).

Any guidance would be very much appreciated!

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/AndroidStudioReleasesGenerator.kt
shalupov: Why you're force to use UIJP own file utilities? IMHO it's better to hide that warning than to use obsolete API

shalupov (changes requested):

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/ComparePerformance.kt
shalupov: I don't think that modifying PATH in any way is the correct way of action. It's better not to assume what can be there. The same with using JAVA_HOME and other variables. `java.home` guarantee the path to actual JVM.

Now, that's not always the case if you want `jbr` to be 100% same as target jvm process to monitor

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/RunPerformanceTests.kt
shalupov: Could you please explain what you want to achieve with `jfr` exactly? Probably it could be solved without getting the list of methods

shalupov: Also, you are not technically tied to JUnit5BazelRunner, you can whatever runner you want

DanielSouzaBertoldi: Hmmmm interesting, didn't think of creating a runner just for this sort of test. Given that the test plan was already being called in the JUnit5BazelRunner class I thought it'd be easier to just modify it instead. Will try a custom runner and let you know :)

shalupov: Well, it's always better to use standard tools if you can, but in case of Bazel you are not limited to. e.g. for junit5 and Bazel it is https://github.com/bazel-contrib/rules_jvm

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/ComparePerformance.kt
DanielSouzaBertoldi: While figuring out a way to stop shelling out to `jfr`, I found out that there's a way to use the JDK's own jdk.jrf APIs in-process. This way we can avoid trying to fetch `jfr` from bin and solve both issues (this and the one I asked about in `RunPerformanceTests`).

Converting this PR to draft for now while I work on this. Thanks for your inputs!

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/RunPerformanceTests.kt
DanielSouzaBertoldi: Updated the code to use JFR's apis directly from the JDK instead of the executable. Thanks!

platform/jewel/build-scripts/bazel/src/main/kotlin/org/jetbrains/jewel/scripts/bazel/ComparePerformance.kt
DanielSouzaBertoldi: Updated the code to use JFR's apis directly from the JDK instead of the executable. Thanks!

platform/jewel/build-scripts/bazel/BUILD.bazel
```diff
@@ -1,5 +1,14 @@
...
 load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_import", "kt_jvm_library", "kt_jvm_test")
+load("//build:compiler-options.bzl", "create_kotlinc_options")
+
+# This module's kt_jvm_library targets otherwise fall back to rules_kotlin's toolchain-wide default of
+# jvm_target 1.8, under which jdk.jfr.consumer's APIs (and java.lang.Record) don't resolve. For jdk.jfr to compile
+# we are forced to bump the jvm version.
+create_kotlinc_options(
+    name = "jvm21-kotlinc-opts",
+    jvm_target = "21",
+)
```
DanielSouzaBertoldi, line 11: **QUESTION FOR THE BAZEL TEAM!**

Is there a shared/sanctioned way to bump `jvm_target` for a `kt_jvm_library`, or should we keep defining our own per-module `create_kotlinc_options` target?

While wiring up the `jdk.jfr.consumer` APIs we hit `Unresolved reference 'jfr'` and `Unresolved reference 'Record'` (i.e. `java.lang.Record`) purely from Kotlin source, even though nothing about our own build config asked for an old target.

The root cause (confirmed via `bazel aquery 'mnemonic("KotlinCompile", <target>)'`) is that `kt_jvm_library` falls back to `rules_kotlin`'s toolchain-wide default of `--kotlin_jvm_target 1.8` whenever a target doesn't explicitly call `create_kotlinc_options`, which hides any JDK API introduced after Java 8 from Kotlin specifically (Java sources in the same target aren't affected).

Is `1.8` intentional as the blanket default, and is there already a shared, discoverable `kotlinc_opts` target elsewhere in the monorepo for "give me a modern JDK" that we should be pointing at instead of rolling our own per module?

DanielSouzaBertoldi: @shalupov @mfilippov hey guys! I created our own runner for both the ComparePerfomance and RunPerformanceTests scripts, and found out this minor quirk along the way. Could you guys provide some guidance please? 😄

shalupov: In common.bazelrc we have:

```
common --@rules_jvm//:default-javac-opts=@community//:default-javac-opts
common --@rules_jvm//:default-kotlinc-opts=@community//:k25
```

are they not working somehow?

shalupov (approved):

mfilippov (approved):
LGTM

rock3r:
Shipit, I trust the Bazel masters :)

platform/jewel/build-scripts/bazel/BUILD.bazel
DanielSouzaBertoldi: Sorry for the late response, but I finally figured out the issue: looks like only in the `jvm_library` wrapper in `jvm-rules/rules/library.bzl:152` the `default-javac-opts` is referenced:

> effective_kotlinc_opts = kotlinc_opts if kotlinc_opts != None else Label("//:default-kotlinc-opts")

I created these files manually, so I thought that, since it's all kotlin, I might as well use the specific kotlin wrappers/macros.

I can change it to `jvm_library` if that's the convention and get rid of creating a brand new `create_kotlinc_options` in the process

shalupov: Thanks! Looks like oops from our side, @eugenezh could you pls take a look?

@DanielSouzaBertoldi Workaround is ok and we'll fix it later

DanielSouzaBertoldi:
Ready to merge

rock3r (approved):
Image

I have nothing useful to contribute but approving to appease the bot

platform/jewel/build-scripts/bazel/BUILD.bazel
eugenezh: @DanielSouzaBertoldi you are using directly "kt_jvm_library" rule, which is inherited from vanilla 'rules_kotlin'.

The monorepo mostly uses "jvm_library" from our in-house bazel rules, which implement a custom compilation worker capable to compile kotlin+java+resources incrementally similar to how the IDEA's built-in JPS compilation works. All default option sets are written and pre-defined for the 'jvm_library' rule.

If you can switch to using 'jvm_library' rule, I would suggest doing this: you'll get at least good incremental compilation and the build behavior will be unified with the one for rest of monorepo packages. 

There is a work in progress to contribute to rules_kotlin all necessary functionality we need to compile our monorepo and eventually we are going to switch to using rules_kotlin directly. For all 'jvm_library' usages the switch will happen automatically.

DanielSouzaBertoldi: No problem, I'll change every kt_jvm_library to the jvm_library then! Thanks for explaining this further :)

DanielSouzaBertoldi: @eugenezh switched all `kt_jvm_library` to `jvm_library`!

Also had to:

1. add `@lib//:kotlin-stdlib` to all `jvm_library` targets. With `kt_jvm_library` that dependency was provided for free
2. had to add `"@lib//:libraries-jetbrains-annotations-jetbrains-annotations"` to `jewel-bazel-scripts-utils` because of one parameter in `JewelBazelScriptsUtils.kt`: `@Language("shell") command: String,`. This dependency was being given for free with `kt_jvm_library` too

Everything is running fine and as expected. Thanks for the pointers!

shalupov: It's actually bad that something like that is "given for free", it's better to reference them explicitly in a large project

DanielSouzaBertoldi: Yeah from the little I know of Bazel seems a bit weird that some rules do offer "transitive dependencies" when the actual value of Bazel comes from being hermetic and having to manually list every single dependency too

DanielSouzaBertoldi:
@nebojsa-vuksic this will be up to you to merge as well :/

> Diff detected, please apply changes from the 'fix.patch' artifact or run 'bazel run format' command locally.

PR #3638, opened 2026-09-03
[JEWEL] IJPL-254570 icons: Add new icons generator

jakub-senohrabek:
Adds generation for the new icons descriptors to the icons generation script.
Not sure about the naming and module placement, but this should be ok.

We should also consider if we want to enable this generation by default, it is enabled for AllIcons only at the moment.

### DIFF
--- a/platform/lang-api/resources/messages/LangBundle.properties
+++ b/platform/lang-api/resources/messages/LangBundle.properties
@@ -628,3 +628,6 @@ notification.content.editor.autoreparse.delay.has.been.restored=Editor autorepar
 action.open.settings.text=Open Settings
 show.a.code.vision.inlay.inlay.when.no.results.are.available=Show a Code Vision inlay when no results are available
 modern.diff.dummy.preview.text=Preview
+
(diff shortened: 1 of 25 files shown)


rock3r:
About to jump in meetings and then end my working week. @wellingtoncosta please review for me?

platform/icons-api/src/com/intellij/platform/icons/design/IconDesigner.kt
```diff
@@ -42,3 +42,7 @@ interface IconAnimationDesigner {
     fun frame(duration: Long, builder: IconDesigner.() -> Unit)
     fun frame(fadeIn: Long, stay: Long, fadeOut: Long, builder: IconDesigner.() -> Unit)
 }
+
+fun IconAnimationDesigner.iconFrame(duration: Long, icon: Icon) = frame(duration) { icon(icon) }
```
wellingtoncosta, line 46: Don't we need to update `api-dump` since these new functions are being added?

wellingtoncosta (approved):

jakub-senohrabek:
New PR: https://github.com/JetBrains/intellij-community/pull/3644

PR #3639, opened 2026-09-03
[JEWEL-750] Implement Markdown scroll sync in standalone sample

wellingtoncosta:
`ScrollingSynchronizer` could only scroll a preview from the editor, through `scrollToLine`, which moves block by
block. The preview stands still while the editor scrolls through a long soft-wrapped line and then jumps, and there was no way to scroll the editor from the preview at all.

This adds continuous, bidirectional scroll sync to `markdown/core`, for both `Markdown` and `LazyMarkdown` previews, and uses it in the standalone sample.

## Changes

* Add `ContinuousScrollingSynchronizer`, a `ScrollingSynchronizer` subclass that can follow the editor continuously rather than block by block. `ScrollingSynchronizer.create` keeps its signature, and the new `createContinuous` factory returns the subclass.
* Support `LazyListState`, so scroll sync works with `LazyMarkdown` as well as `Markdown`. A lazy list has no absolute scroll offset, and the items scrolled past are disposed, so the preview's position is described the way the list itself describes it: which item is at the top, and how far into it. `create` no longer returns `null` for a `LazyListState`.
* Add `syncScrolling`, which keeps the editor and the preview in step in both directions, and `TextLayoutResult.sourceLineOffsets()`, which builds the line-to-offset function for an editor that is a Compose text field.
* Add `acceptContentPosition`, called by `Markdown`, so block positions are relative to the preview instead of the composition root. Also, add `acceptItemBlocks`, called by `LazyMarkdown`, so a source line can be located by item index.
* Guard `ScrollingSynchronizer`'s maps with a lock. Parsing on a background dispatcher while the UI thread read positions could throw `ConcurrentModificationException`.
* The standalone sample's Markdown demo syncs editor and preview through `syncScrolling`, and keeps using `LazyMarkdown`.

## Screen recording

https://github.com/user-attachments/assets/c273f331-b292-4cdb-b676-92a9b1652fdf

### New features

* Markdown scroll sync is now continuous and bidirectional, and works with both `Markdown` and `LazyMarkdown` previews. The new `ContinuousScrollingSynchronizer` follows the editor between block tops instead of jumping from block to block, and `syncScrolling` keeps an editor `ScrollState` and a preview in step in both directions. `TextLayoutResult.sourceLineOffsets()` supplies the line positions for an editor that is a Compose text field.
* `ScrollingSynchronizer.create` now supports `LazyListState`, where it previously returned `null` and logged a warning. Use `LazyListState` with `LazyMarkdown`, and `ScrollState` with `Markdown`.
* The standalone sample's Markdown demo now syncs the editor and the preview.


DanielSouzaBertoldi:
Can we haz screen recordings for before/after plzzzzzzz

platform/jewel/samples/standalone/src/main/kotlin/org/jetbrains/jewel/samples/standalone/markdown/MarkdownPreview.kt
```diff
@@ -125,15 +131,19 @@ internal fun MarkdownPreview(rawMarkdown: CharSequence, modifier: Modifier = Mod
...
 
-    ProvideMarkdownStyling(markdownStyling, blockRenderer, codeHighlighter) {
-        val lazyListState = rememberLazyListState()
-        VerticallyScrollableContainer(lazyListState as ScrollableState, modifier.background(background)) {
-            LazyMarkdown(
-                blocks = markdownBlocks,
-                modifier = Modifier.background(background),
-                contentPadding =
-                    PaddingValues(start = 8.dp, top = 8.dp, end = 8.dp + scrollbarContentSafePadding(), bottom = 8.dp),
-                state = lazyListState,
```
DanielSouzaBertoldi, line 136: I'm really not a big fan of ditching `LazyMarkdown`. We are forcing users to rely on `Markdown` for scroll sync no matter the size of the markdown file itself.

The KDoc for Markdown also explicitly states:

>  * For large amounts of Markdown, such as documents, you can consider using [LazyMarkdown] instead to get better performance.

Which is not necessarily true anymore. Besides, we don't have any other sample that makes use of `LazyMarkdown`.

How hard would it be to achieve scroll sync a `LazyMarkdown` as well?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/SyncScrolling.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+}
+
+private const val NO_OFFSET = -1
+
+/** Scrolls to [target] and returns the offset this state will report as a result, or [NO_OFFSET] if it didn't move. */
+private suspend fun ScrollState.scrollToAndReport(target: Int): Int {
+    val before = value
+    scrollTo(target)
```
DanielSouzaBertoldi, line 71: **Claude finding** (yes i did validate it wasn't hallucinating stuff, but maybe there's something that i'm missing here)

---

`syncScrolling` moves the counterpart pane with `scrollTo`, which takes that pane's scroll mutex at `MutatePriority.Default`. User gestures hold it at `UserInput`, and `MutatorMutex` doesn't queue a lower priority caller, it throws (`MutatorMutex.kt:98`, foundation 1.12.0). The exception escapes collect, cancels that launch, and since it's a `CancellationException` it doesn't reach the parent scope. One direction stops working, nothing is logged, and it never recovers because `LaunchedEffect(synchronizer)` never re-runs.
  
The echo guard is what's supposed to stop us pushing a pane the user is driving. It's built on `snapshotFlow`, which conflates. Compose holds `UserInput` for an entire smooth wheel animation, not one frame, so across those frames the collector can wake on a value that isn't the recorded one, the guard misses, and we push into a pane mid gesture.
  
Repro: wheel-scroll one pane hard for a few seconds, then try the other.

```
try { scrollTo(target) }
catch (e: CancellationException) { currentCoroutineContext().ensureActive(); return NO_OFFSET } 
```

`ensureActive()` keeps real cancellation of syncScrolling working. Compose's own userScroll wraps in supervisorScope for exactly this.
  
Same trap for any host running caret-driven `scrollToLine` alongside `syncScrolling`. Both are `Default`.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/SyncScrolling.kt
```diff
@@ -0,0 +1,88 @@
...
+        // Syncing one pane moves the other, which would sync the first one back.
+        var expectedEditorOffset = NO_OFFSET
+        var expectedPreviewOffset = NO_OFFSET
+        launch {
+            snapshotFlow { editorScrollState.value }
+                .collect { offset ->
+                    if (offset == expectedEditorOffset) {
+                        expectedEditorOffset = NO_OFFSET
+                        return@collect
+                    }
```
DanielSouzaBertoldi, line 41: [snapshotFlow conflates state changes](https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#snapshotFlow(kotlin.Function0)), i.e, it drops intermediary values when the collector is busy.

I would imagine ours is busy a LOT because `scrollTo` suspends, so the preview collector can miss the exact value the editor collector recorded. I think that we can fix this code (and the same code at lines 52-54) by doing:

```kt
val expected = expectedPreviewOffset
expectedPreviewOffset = NO_OFFSET
if (offset == expected) return@collect
```

buuut matching on the offset can't exactly tell an "echo" (syncing one panel based on the other) from the user scrolling back to an exact spot...

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
```diff
@@ -143,6 +151,26 @@ public abstract class ScrollingSynchronizer {
      */
     public abstract fun acceptTextLayout(block: MarkdownBlock, textLayout: TextLayoutResult)
 
+    /**
+     * Accept the [coordinates] of the scrolled content holding all the blocks. Block positions arrive relative to the
+     * composition root; this is what makes them relative to the preview when the preview doesn't start at the root.
+     * Called on first composition and whenever the content moves, including on every scroll.
+     */
+    public open fun acceptContentPosition(coordinates: LayoutCoordinates) {}
```
DanielSouzaBertoldi, line 159: I see we don't have any tests for this little guy. Could we add some? 😄

DanielSouzaBertoldi (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
```diff
@@ -143,6 +151,26 @@ public abstract class ScrollingSynchronizer {
...
+    /**
+     * The preview scroll offset matching [editorOffset], interpolated between the tops of the two blocks around it, so
+     * that the preview keeps moving as long as the editor does.
+     *
+     * [editorOffsetOfLine] returns the editor scroll offset at which a source line starts, or `null` if the editor
+     * doesn't know that line. [editorMaxOffset] is the editor's maximum scroll offset; it is paired with the preview's
+     * so that both reach their ends together.
+     */
+    public abstract fun previewOffsetAt(editorOffset: Int, editorMaxOffset: Int, editorOffsetOfLine: (Int) -> Int?): Int
+
```
rock3r, line 170: Can we move these two abstract methods to a `ContinuousScrollingSynchronizer : ScrollingSynchronizer()` subclass instead?

Adding them here directly breaks existing custom synchronizers to implement a breaking change. We need to preserve binary compatibility even for experimental APIs, and that seems an easy way to do it.

`PerLine` could extend the new subclass, with `syncScrolling` requiring that type. Keep the existing `create()` signature and add a factory returning the new type. That lets implementations opt into continuous sync without breaking existing subclasses or adding artificial defaults.

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/SyncScrolling.kt
```diff
@@ -0,0 +1,88 @@
...
+    editorScrollState: ScrollState,
+    previewScrollState: ScrollState,
+    editorOffsetOfLine: () -> (Int) -> Int?,
+) {
+    coroutineScope {
+        // Syncing one pane moves the other, which would sync the first one back.
+        var expectedEditorOffset = NO_OFFSET
+        var expectedPreviewOffset = NO_OFFSET
+        launch {
+            snapshotFlow { editorScrollState.value }
```
rock3r, line 36: Note, not a change request: these flows only observe scroll offsets. Resizing can change text wrapping without changing either offset, leaving the panes out of sync until the next scroll. The Markdown plugin’s HTML preview has a similar limitation: it skips sync when the top source line is unchanged and has no explicit resize resync. We could handle layout changes in a follow-up, or leave this as-is. No need to address it in this PR. If you want to volunteer for the follow up, file an issue and let's go :)

rock3r (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
```diff
@@ -430,7 +494,152 @@ public abstract class ScrollingSynchronizer {
...
+            // ponytail: linear scan. One entry per top-level block, and the scrolled preview already pays the same per
+            // scroll; make it a binary search if a big document ever shows up in a profile.
+            val index =
+                itemBlocks.indices.lastOrNull { index ->
+                    val offset = editorOffsetOfItem(index, editorOffsetOfLine)
+                    offset != null && offset <= editorOffset
+                } ?: return null
+            val from = editorOffsetOfItem(index, editorOffsetOfLine) ?: return null
+            val to = editorOffsetOfItem(index + 1, editorOffsetOfLine)
+            if (to == null || to <= from) return index to 0f
```
rock3r: [Astra find]

Looks like continuous sync stops within the last lazy item. There is no next item, so this branch always returns a zero fraction. `editorOffsetAtPreview` likewise returns the item's starting editor offset. A document containing just one tall code block therefore maps every scroll position back to the top.

A small regression test for `LazyScrollingSynchronizerTest`, using its existing `doTest` helper:

```kotlin
@Test
public fun `scrolling within the only block advances the editor`() {
    val markdown = "```text\n" + (1..100).joinToString("\n") { "line $it" } + "\n```"
    doTest(markdown) { state, synchronizer ->
        state.scrollToItem(0, 200)
        assertTrue(state.firstVisibleItemScrollOffset > 0)
        assertTrue(
            synchronizer.editorOffsetAtPreview(2000) { line -> line * 20 } > 0
        )
    }
}
```

Astra reproduced the mapping failure with the unchanged `PerLineLazy` implementation in an isolated Kotlin harness with state stubs. Offsets of 400, 1000, and 1700 all mapped to zero in both directions. The project test above and a live UI reproduction have not been run.

Can we map through the final block to the panes' scrollable ends, and cover both directions? `editorMaxOffset` is currently unused in the lazy implementation.

rock3r (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
```diff
@@ -430,7 +495,210 @@ public abstract class ScrollingSynchronizer {
...
+            val bothEndsKnown = editorMaxOffset != Int.MAX_VALUE && previewMaxOffset != Int.MAX_VALUE
+            val end = listOfNotNull(if (bothEndsKnown) Anchor(editorMaxOffset, previewMaxOffset) else null)
+            return blockAnchors
+                // Blocks on either edge would only add a jump: the first block sits below the panes' padding, and
+                // blocks in the last screenful never reach the top of a pane.
+                .filter { it.editorOffset in 1..<editorMaxOffset && it.previewOffset in 1..<previewMaxOffset }
+                .sortedWith(compareBy({ it.editorOffset }, { it.previewOffset }))
+                .let { listOf(Anchor(0, 0)) + it + end }
+        }
+
```
DanielSouzaBertoldi, line 558: This functions is doing a ton of heavy lifting: it runs `mapNotNull` over _all_ of the `blocks2Top`, calls the host lambda once per block, filters, sorts and at the end returns a `listOf(Anchor) + it + end` for two more copies.

`anchors` is called by both `scrollPreviewTo` and `editorOffsetAtPreview`. Both of these functions are called by the `syncScrolling` inside a `snapshotFlow`, so they not only run during scroll but also on snapshot changes too

I would imagine that, if you scroll a really large document, all of these calculations will take a toll on performance since we are sorting/mapping/filtering per frame over every single markdown block the document has.

I'm fine with shipping this as-is, but I really would like if we could file an issue to improve the performance here. We could explore trying to cache the list and invalidating on parse or making `interpolate` do a binary search on a list that's already sorted.

Wdyt??

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/SyncScrolling.kt
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @return A function giving the scroll offset at which a source line starts, or `null` for a line this layout does not
+ *   contain.
+ */
+@ApiStatus.Experimental
+@ExperimentalJewelApi
+public fun TextLayoutResult.sourceLineOffsets(): (Int) -> Int? {
+    val lineStarts = buildList {
+        add(0)
+        layoutInput.text.forEachIndexed { index, char -> if (char == '\n') add(index + 1) }
```
DanielSouzaBertoldi: shouldn't we count `\r` as a line break too?

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
wellingtoncosta: Fixed


https://github.com/user-attachments/assets/ea30131e-ad88-4768-abff-86dafce9faeb

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/Markdown.kt
```diff
@@ -166,11 +169,16 @@ public fun Markdown(
     onUrlClick: (String) -> Unit = {},
     blockRenderer: MarkdownBlockRenderer = JewelTheme.markdownBlockRenderer,
 ) {
+    // Tells the synchronizer where the scrolled content is, so block positions can be made relative to the preview
+    val synchronizer = (JewelTheme.markdownMode as? MarkdownMode.EditorPreview)?.scrollingSynchronizer
+    val contentPositionModifier =
+        if (synchronizer == null) Modifier else Modifier.onGloballyPositioned(synchronizer::acceptContentPosition)
```
DanielSouzaBertoldi: Shouldn't this be remembered using `synchronizer` given that `onGloballyPositioned` is basically capturing it?

rock3r (approved):

DanielSouzaBertoldi (changes requested):

platform/jewel/markdown/core/src/main/kotlin/org/jetbrains/jewel/markdown/scrolling/ScrollingSynchronizer.kt
DanielSouzaBertoldi: i'd really appreciate if you could reply to me

wellingtoncosta: I did try the binary search approach first in `stretchAround`, but it turned out to only optimize the last step. `anchors()` still ran a full `mapNotNull` + `editorOffsetOfLine()` per block + sort + two list concats on every call, and that was the real cost.

So I replaced `anchors()` + `stretchAround()` with a single O(n) pass that finds the bracketing anchors directly. No list rebuild, no sort, no aux lists. Also moved `editorOffsetOfLine()` off the lock.

To achieve O(log n) would require more changes, especially in `blocks2Top` to keep it ordered by source line across calls. I think the current implementation is better than before, both in terms of performance and readability.

DanielSouzaBertoldi: That's fine, thank you for taking the time to improve the performance on this PR! 😄

DanielSouzaBertoldi (approved):

PR #3642, opened 2026-09-07
KTIJ-40160 Kotlin UAST: resolve compiled Kotlin properties to accessor methods in K2 UAST

francescoo22:


francescoo22:
@KirpichenkovPavel

PR #3644, opened 2026-09-09
[jewel] IJPL-254570 icons: Add new icons generator

jakub-senohrabek:
Adds generation for the new icons descriptors to the icons generation script.
Not sure about the naming and module placement, but this should be ok.

We should also consider if we want to enable this generation by default, it is enabled for AllIcons only at the moment.

The branch was changed, check this PR for history:
https://github.com/JetBrains/intellij-community/pull/3638


wellingtoncosta (commented):
The code looks good to me, but some checks are failing and not sure if some of them e.g. spectre tests are related to the issue @rock3r reported in other pull requests.

wellingtoncosta (approved):

platform/icons/generated/com/intellij/platform/icons/AllIconsDescriptors.kt
```diff
@@ -0,0 +1,1628 @@
...
+/**
+ * NOTE THIS FILE IS AUTO-GENERATED
+ * DO NOT EDIT IT BY HAND, run "Generate icon classes" configuration instead
+ */
+@ApiStatus.Experimental
+object AllIconsDescriptors {
+
+  object Actions {
+    /** 16x16 */ val AddDirectory: IconDescriptor = imageIconDescriptor("expui/actions/addDirectory.svg", AllIconsDescriptors::class.java.classLoader)
+    /** 16x16 */ val AddFile: IconDescriptor = imageIconDescriptor("expui/actions/addFile.svg", AllIconsDescriptors::class.java.classLoader)
```
rock3r: Looks like these are eagerly constructed through `IconManager`, so accessing an `AllIconsDescriptors` property before the manager is initialized fails while initializing its holder. This is especially easy in standalone apps, where the manager is set up by `IntUiTheme`.

Could we make descriptor construction independent of the active rendering environment, and only resolve it when it is rendered?

platform/jewel/README.md
```diff
@@ -306,7 +306,7 @@ Please refer to the [release notes](RELEASE%20NOTES.md) to determine the compati
 
 ## Icons
 
-Loading icons is best done with the `Icon` composable, which offers a key-based API that is portable across bridge and
+Loading icons is best done with the `IconDescriptor` composable, which offers a key-based API that is portable across bridge and
```
rock3r: `IconDescriptor` is the descriptor type, but the composable is still named `Icon` (as it should). This looks like an over-eager search and replace. Could we keep `Icon` here? The example immediately below already uses the correct name.

platform/icons-api/src/com/intellij/platform/icons/IconManager.kt
```diff
@@ -81,19 +81,19 @@ interface IconManager {
  * Check the designer interface for layer options. Also check intellij.platform.icons.api.swing module to find out how
  * to convert swing icons and new icons.
  *
- * @see IconManager.toSwingIcon
+ * @see IconManager.createSwingIcon
  */
-fun icon(designer: IconDesigner.() -> Unit): Icon = IconManager.getInstance().icon(designer)
+fun iconDescriptor(designer: IconDesigner.() -> Unit): IconDescriptor = IconManager.getInstance().iconDescriptor(designer)
```
rock3r, line 88: The KDoc above still points callers to `icon {}` and `Icon.createRenderer()`, but these were renamed in this change. Could we update it to the `iconDescriptor` / `IconDescriptor` names?

platform/icons/generated/com/intellij/platform/icons/AllIconsDescriptors.kt
rock3r: Could we generate `@JvmStatic` getters here? Java callers otherwise need `AllIconsDescriptors.Actions.INSTANCE.getAddFile()`, while platform APIs such as `JBLabel` already accept `IconDescriptor`. That would let them use `AllIconsDescriptors.Actions.getAddFile()` instead.

platform/build-scripts/icons/src/org/jetbrains/intellij/build/images/IconsClassGenerator.kt
```diff
@@ -907,6 +944,73 @@ internal open class IconsClassGenerator(
       generator.append(result, "$classAnnotation\npublic static final @NotNull IntelliJIconKey $iconName = ${deprecation.replacementReference};", level)
     }
   }
+
+  private object KotlinNewApiIconClassSpecificsGenerator : IconClassSpecificsGenerator {
+    override val classAnnotation: String = "@ApiStatus.Experimental"
+
+    override fun amendIconClassName(originalName: String): String = "${originalName}Descriptors"
```
rock3r: Could we call this `AllIconDescriptors` instead? It reads more naturally as the descriptors for all icons, while `AllIconsDescriptors` sounds like descriptors belonging to the legacy `AllIcons` holder.

platform/icons/generated/com/intellij/platform/icons/AllIconsDescriptors.kt
```diff
@@ -0,0 +1,1628 @@
...
+package com.intellij.platform.icons
+
+import org.jetbrains.annotations.ApiStatus
+
+/**
+ * NOTE THIS FILE IS AUTO-GENERATED
+ * DO NOT EDIT IT BY HAND, run "Generate icon classes" configuration instead
+ */
+@ApiStatus.Experimental
+object AllIconsDescriptors {
```
rock3r: Could we clarify the lifecycle contract here? These properties eagerly build descriptor graphs through `IconManager`. For remote development, some early normalization may be needed to make the descriptor portable, but I am not sure this should go through an active rendering manager. That seems to defeat the purpose of descriptors being portable and I am not sure if it would actually work across different machines/processes.

And also, where do we guarantee that the manager is initialized before generated descriptors are accessed, both in the IDE and in Jewel standalone? In standalone, `IntUiTheme` activates it only during first composition, after top-level state and view models can already access these properties. It is fine to _provide_ it there, but we should ensure it gets loaded as early as possible at process spawn to avoid issues when the descriptors get classloaded on first access. In the IDE it should be possible to init it in the existing IDE bootstrao path at least...

platform/jewel/README.md
```diff
@@ -359,7 +359,7 @@ In your holders, you can choose which implementation of `IconKey` to use:
 
 ### Painter hints
 
-Jewel has an API to influence the loading and drawing of icons, called `PainterHint`. `Icon` composables have overloads
+Jewel has an API to influence the loading and drawing of icons, called `PainterHint`. `IconDescriptor` composables have overloads
```
rock3r, line 362: Ditto

platform/jewel/RELEASE NOTES.md
```diff
@@ -200,7 +200,7 @@ dependencies {
     `DisabledAppearanceValues`.
 * `ide-laf-bridge`:
   * The `TextColors` and `ThemeDefinition` secondary constructors were removed as in `foundation`.
-  * Both `bridgePainterProvider` functions were removed; use an `IconKey` with the `Icon` composable instead.
+  * Both `bridgePainterProvider` functions were removed; use an `IconKey` with the `IconDescriptor` composable instead.
```
rock3r, line 203: Please undo these spurious changes (again, over-eager search and replace I suppose)

platform/platform-api/src/com/intellij/ui/components/JBLabel.java
```diff
@@ -67,13 +67,13 @@ public JBLabel(@Nullable Icon image) {
   }
 
   @ApiStatus.Internal
-  public JBLabel(@NotNull com.intellij.platform.icons.Icon icon, IconScale scale) {
-    super(SwingIconKt.toSwingIcon(icon, scale));
+  public JBLabel(@NotNull com.intellij.platform.icons.IconDescriptor iconDescriptor, IconScale scale) {
```
rock3r, line 70: These can be accessed through an import now

platform/platform-api/src/com/intellij/ui/dsl/builder/Row.kt
```diff
@@ -307,7 +307,7 @@ interface Row {
   fun icon(icon: Icon): Cell<JLabel>
 
   @ApiStatus.Internal
-  fun icon(icon: com.intellij.platform.icons.Icon, scale: IconScale = IconScale.Default): Cell<JLabel>
+  fun icon(iconDescriptor: com.intellij.platform.icons.IconDescriptor, scale: IconScale = IconScale.Default): Cell<JLabel>
```
rock3r, line 310: Ditto here, import please.

platform/platform-impl/src/com/intellij/ui/dsl/builder/impl/RowImpl.kt
```diff
@@ -320,16 +321,18 @@ internal open class RowImpl(private val dialogPanelConfig: DialogPanelConfig,
     return cell(label)
   }
 
-  override fun icon(icon: com.intellij.platform.icons.Icon, scale: IconScale): CellImpl<JLabel> {
-    val label = JBLabel(icon, scale)
-    label.disabledIcon = com.intellij.platform.icons.icon {
-      icon(icon, IconModifier.colorFilter(ColorFilter.tint(Color.Black, BlendMode.Saturation)))
-    }.toSwingIcon()
+  @ApiStatus.Internal
+  override fun icon(iconDescriptor: com.intellij.platform.icons.IconDescriptor, scale: IconScale): CellImpl<JLabel> {
```
rock3r, line 325: Ditto here, import please.

plugins/devkit/intellij.devkit.compose/src/showcase/ComposePerformanceDemoAction.kt
```diff
@@ -248,12 +248,12 @@ private fun createIconsComponent(): JComponent {
           }
         } else {
           val icons = remember {
-            val icons = mutableListOf<com.intellij.platform.icons.Icon>()
+            val iconDescriptors = mutableListOf<com.intellij.platform.icons.IconDescriptor>()
```
rock3r, line 251: Ditto here, import please.

platform/icons-api/src/com/intellij/platform/icons/swing/SwingIcon.kt
```diff
@@ -20,15 +19,15 @@ fun IconDesigner.swingIcon(legacyIcon: javax.swing.Icon, modifier: IconModifier
  * Converts swing icon to be used by this API.
  * If the icon is just a swing wrapper containing our icon, it will be unpacked and embedded directly.
  */
-fun javax.swing.Icon.toNewIcon(): Icon =
-    IconManager.getInstance().toNewIcon(this)
+fun javax.swing.Icon.toDescriptor(): IconDescriptor =
```
rock3r: No need to fully qualify these anymore, please add imports.

platform/icons-impl/src/com/intellij/platform/icons/impl/DefaultIconManager.kt
```diff
@@ -97,11 +97,11 @@ abstract class DefaultIconManager : IconManager {
         buildCustomSerializers()
     }
 
-    override fun toSwingIcon(icon: Icon, scale: IconScale): ScalableSwingIcon {
+    override fun createSwingIcon(iconDescriptor: IconDescriptor, scale: IconScale): ScalableSwingIcon {
         error("Swing Icons are not supported.")
     }
 
-    override fun toNewIcon(swingIcon: javax.swing.Icon): Icon {
+    override fun toIconDescriptor(swingIcon: javax.swing.Icon): IconDescriptor {
```
rock3r: No need to fully qualify these anymore, please add imports.

platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/IntelliJIconManager.kt
```diff
@@ -52,32 +52,32 @@ class IntelliJIconManager : DefaultIconManager() {
...
     }
     if (designer !is IntelliJIconDesigner) {
       error("Only IntelliJIconDesigner can handle swing icons.")
     }
     designer.addSwingLayer(swingIcon, modifier)
   }
 
-  override fun toNewIcon(swingIcon: javax.swing.Icon): Icon {
-    if (swingIcon is SwingIcon && swingIcon.scale == null) return swingIcon.icon
+  override fun toIconDescriptor(swingIcon: javax.swing.Icon): IconDescriptor {
```
rock3r: No need to fully qualify these anymore, please add imports.

platform/icons-impl/intellij/src/com/intellij/platform/icons/impl/intellij/rendering/SwingIcon.kt
```diff
@@ -22,7 +22,7 @@ import java.util.WeakHashMap
 import kotlin.math.roundToInt
 
 class SwingIcon(
-  val icon: Icon,
+  val iconDescriptor: IconDescriptor,
   val scale: IconScale? = null
 ): javax.swing.Icon, ScalableSwingIcon {
```
rock3r: No need to fully qualify this anymore, please add an import.

plugins/devkit/intellij.devkit.compose/src/showcase/Icons.kt
```diff
@@ -256,7 +233,7 @@ internal fun Icons() {
 }
 
 class ShowcaseIcon(
-  val icon: Icon,
+  val iconDescriptor: IconDescriptor,
   val swingAlternative: javax.swing.Icon?,
```
rock3r, line 237: No need to fully qualify these anymore, please add imports.

rock3r (changes requested):

platform/icons-api/src/com/intellij/platform/icons/swing/SwingIcon.kt
wellingtoncosta: Shouldn't be `toIconDescriptor()`?

platform/icons-api/src/com/intellij/platform/icons/IconManager.kt
```diff
@@ -20,25 +20,25 @@ interface IconManager {
...
      * Converts specific Icon to swing Icon. ! This is an expensive operation and can include image loading, reuse the
      * instance if possible. !
      */
-    fun toSwingIcon(icon: Icon, scale: IconScale = factor(1f)): ScalableSwingIcon
+    fun createSwingIcon(iconDescriptor: IconDescriptor, scale: IconScale = factor(1f)): ScalableSwingIcon
 
     fun addSwingLayer(designer: IconDesigner, swingIcon: javax.swing.Icon, modifier: IconModifier)
 
-    fun toNewIcon(swingIcon: javax.swing.Icon): Icon
+    fun toIconDescriptor(swingIcon: javax.swing.Icon): IconDescriptor
```
wellingtoncosta, line 42: You can also add import here instead

platform/icons-api/src/com/intellij/platform/icons/IconManager.kt
```diff
@@ -20,25 +20,25 @@ interface IconManager {
...
+    suspend fun forceEvaluation(icon: DeferredIconDescriptor): IconDescriptor
 
     /**
      * Converts specific Icon to swing Icon. ! This is an expensive operation and can include image loading, reuse the
      * instance if possible. !
      */
-    fun toSwingIcon(icon: Icon, scale: IconScale = factor(1f)): ScalableSwingIcon
+    fun createSwingIcon(iconDescriptor: IconDescriptor, scale: IconScale = factor(1f)): ScalableSwingIcon
 
     fun addSwingLayer(designer: IconDesigner, swingIcon: javax.swing.Icon, modifier: IconModifier)
```
wellingtoncosta, line 40: Same

platform/platform-api/src/com/intellij/ui/components/JBLabel.java
```diff
@@ -67,13 +67,13 @@ public JBLabel(@Nullable Icon image) {
...
+    super(SwingIconKt.createSwingIcon(iconDescriptor, scale));
   }
 
   @ApiStatus.Internal
-  public JBLabel(@NotNull com.intellij.platform.icons.Icon icon) {
-    this(icon, IconScale.Default);
+  public JBLabel(@NotNull com.intellij.platform.icons.IconDescriptor iconDescriptor) {
```
wellingtoncosta, line 75: Same here

platform/icons-api/src/com/intellij/platform/icons/swing/SwingIcon.kt
rock3r: Yeah good point

platform/icons/generated/com/intellij/platform/icons/AllIconsDescriptors.kt
jakub-senohrabek: IconManager in this case is a glorified factory, where we take care of the biggest differences between standalone and IntelliJ. In IDE it is simple, it is set up at bootstrap. In standalone this is more fragile. 

The rendering is fully decoupled, see IconRendererManager, which is even in different api module. 
In implementation module, they live side-by-side, but if you insist we can also reintroduce split implementation modules. This is tradeoff between setup cost and usability. If you really want to use the icon descriptors render dependency less (even implementation), you still can, just need to have custom IconManager.

The real problem we have, and that stays only in standalone is simple, the IconManager can register both normal and deferredIcons. For normal icons, we don't need coroutine scope and that is fine. However, for deferredIcons, we need coroutine scope, that is why we init the manager on theme, instead of making it lazy load/load at app creation.

I will do more investigation if we can have deferredIcons in standalone without coroutine scope somehow.

PR #3646, opened 2026-09-09
Sort Children Before Printing

alonalbert:
The tests in `//plugins/kotlin/jvm-debugger/test/k2:kotlin-jvm-debugger-test_test`, specifically the test classes

* `org.jetbrains.kotlin.idea.k2.debugger.test.cases.K2IdeK2CodeKotlinEvaluateExpressionTestGenerated`
* `org.jetbrains.kotlin.idea.k2.debugger.test.cases.InlineScopesAndK2IdeK2CodeEvaluateExpressionTestGenerated`
* `org.jetbrains.kotlin.idea.k2.debugger.test.cases.K2IdeK2CodeKotlinVariablePrintingTestGenerated`

Use `FramePrinter` to assert expected vs actual of a stack-frame/variable-list dump.

When running tests on different VM's, for example, Java JVM vs Android ART, the order of objects in a `XValueContainer` can differ causing some tests to fail.

FramePrinter should sort the children before printing them.

https://youtrack.jetbrains.com/issue/IJPL-255207/FramePrinter-Order-Is-Not-Consistent


plugins/kotlin/jvm-debugger/testFramework/testSrc/org/jetbrains/kotlin/idea/debugger/test/util/FramePrinter.kt
```diff
@@ -50,15 +50,15 @@ internal class FramePrinter(private val suspendContext: SuspendContextImpl) {
     fun printTopVariables(frame: XStackFrame): String {
         return buildString {
             append(frame, 0)
-            for (child in frame) {
+            for (child in frame.sorted()) {
```
steelart: Loosing the order is bad: these test also check that, for example, `this` is going first and check other ordering. If Android has another but stable order we can add the corresponding `out` files. Or we can use sorting but only for Android if you don't care about the order there.

alonalbert:
Any updates? The nature of this fix is such that delaying it's commit will result in more and more merge issues since it touches a lot of files

steelart:
I left a comment in the code, but may repeat it here:

Loosing the order is bad: these test also check that, for example, `this` is going first and check other ordering. If Android has another but stable order we can add the corresponding out files. Or we can use sorting but only for Android if you don't care about the order there.

alonalbert:
Can you detail what other ordering is checked?

I will assemble a report of where exactly the Android order is different.

Maybe we come up with a re-ordering that satisfies everything.

steelart:
Yes, we need to look at particular problems with different order. I asked codex for examples of intentional sorting that also checked by these tests:

For a KotlinStackFrame, the existing local order is intentional:
Kotlin locals use the source introduction location.
Variables with the same location use their name.
The Kotlin view places this variables before other locals.

The recursive sort also affects all nested value children:
Object fields become alphabetical.
Array and list indexes become string-sorted.
Map entries become string-sorted by their displayed index.
Renderer-specific child order is lost.

PR #3647, opened 2026-09-09
IJPL-254836 GitHub: back off on GraphQL rate limit and cap request concurrency

alexeyshurygin:
Fixes https://youtrack.jetbrains.com/issue/IJPL-254836

## Problem

`GithubApiRequestExecutor` read the `x-ratelimit-*` headers and the GraphQL `rateLimit` cost only for statistics. Once the primary GraphQL budget (5000 points per hour, per user) was exhausted, the plugin did not back off: every Retry and every reopened pull request view re-sent the whole GraphQL fan-out, each request failed with 403, and the UI offered "Log in again", which cannot help because a new token shares the same per-user bucket.

## Evidence from the ticket

- One large pull request (150+ commits, 80+ files) open in the tool window consumed about 65 points per minute, so the budget lasted about 75 minutes. Several open projects multiply this.
- GitHub reported `used 5042/5000, remaining 0`: the concurrent fan-out is admitted before the counter catches up, so the burst overshoots the budget.
- At the same moment the REST `/rate_limit` endpoint reported the GraphQL budget as untouched. Only the GraphQL `rateLimit` block is a reliable source for that budget.
- Raising the plugin connection timeout does nothing; GitHub rejects immediately.

## Approach (executor only)

- **Rate-limit state** per executor and per resource (`core`, `search`, `graphql`), fed from three sources: the REST headers, the GraphQL `rateLimit` block (the `rates` fragment now also requests `remaining` and `resetAt`), and a 403 rate-limit response (`x-ratelimit-reset` or `Retry-After`; a one-minute cooldown if neither is present).
- **Back-off**: before opening a connection, a request to an exhausted resource fails locally with `GithubRateLimitExceededException` carrying the reset instant. The message reads "GitHub API rate limit exceeded, resets at HH:mm" and surfaces in the existing UI. This is what stops Retry from re-firing the fan-out.
- **Concurrency gate**: a semaphore of 8 around GraphQL request execution so that a burst cannot overshoot the budget.
- Absent headers (some GitHub Enterprise versions) leave the state unknown and never block. Statistics calls are unchanged.

## Tests

`GithubApiRequestExecutorRateLimitTest` wires the executor to a fake connection and counts calls. No test touches the network.

- A 403 rate-limit response followed by 20 more requests results in exactly one network call; every later request throws the typed exception with the reset instant.
- Requests resume after the fake clock passes the reset instant.
- A successful GraphQL response whose `rateLimit` block reports `remaining: 0` puts the executor into back-off without any 403.
- REST headers reporting exhaustion back off that resource only; the GraphQL budget stays usable.
- A rate-limit response without reset headers backs off for the default cooldown.
- Missing rate-limit headers never block.
- 50 concurrent GraphQL requests never exceed the cap in flight.

Disabling the back-off and the gate makes the relevant tests fail. The whole `vcs-github-tests` target passes (1350 tests), including the GraphQL fragment and DTO schema checks.

## Follow-ups (out of scope)

- Reduce the baseline consumption of the pull request view (about 65 points per minute for a large PR): polling and caching redesign.
- The "Log in again" suggestion on a 403 rate-limit error is misleading and could be hidden when the error is a rate limit.

### DIFF
--- a/plugins/github/github-core/resources/graphql/fragment/rates.graphql
+++ b/plugins/github/github-core/resources/graphql/fragment/rates.graphql
@@ -1,3 +1,5 @@
 fragment rates on RateLimit {
   cost
+  remaining
+  resetAt
 }
--- a/plugins/github/github-core/resources/messages/GithubBundle.properties
+++ b/plugins/github/github-core/resources/messages/GithubBundle.properties
@@ -378,6 +378,7 @@ pull.request.details.adjusting.assignees=Adjusting assignees\u2026
 pull.request.details.adjusting.labels=Adjusting labels\u2026
 
 request.response.0=Request response: {0}.
+request.rate.limit.exceeded.resets.at=GitHub API rate limit exceeded, resets at {0}
 group.Github.PullRequest.Diff.Thread.View.Options.text=Show Review Threads
 
 github.actions.file.promo.label = GitHub Actions
--- a/plugins/github/github-core/src/org/jetbrains/plugins/github/api/GithubApiRequestExecutor.kt
+++ b/plugins/github/github-core/src/org/jetbrains/plugins/github/api/GithubApiRequestExecutor.kt
@@ -39,6 +39,14 @@ import java.io.InputStreamReader
 import java.io.Reader
 import java.net.HttpURLConnection
 import java.net.URL
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.Semaphore
+import java.util.concurrent.TimeUnit
 import java.util.zip.GZIPInputStream
 
 /**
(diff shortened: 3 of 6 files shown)


alexeyshurygin:
## Test coverage of new and changed production code files

Measured with the repository's own IntelliJ Coverage agent (`tests.cmd -Dintellij.build.test.coverage.enabled=true ...`) over the whole `intellij.vcs.github.tests` module. **Before** is the parent commit (old code, without `GithubApiRequestExecutorRateLimitTest`). **After** is this PR.

| File | Metric | Before | After |
|---|---|---|---|
| `GithubApiRequestExecutor.kt` | Line | 28/149 (19%) | 128/201 (64%) |
| | Branch | 25/101 (25%) | 88/143 (62%) |
| | Method | 2/42 (5%) | 27/56 (48%) |
| `GHGQLRateLimit.kt` | Line | 0/2 (0%) | 4/4 (100%) |
| `GithubRateLimitExceededException.java` | Line | 0/1 (0%) | 3/4 (75%) |
| | Method | 0/1 (0%) | 2/3 (67%) |

The `rates.graphql` fragment and the bundle key are resources without coverage figures; the fragment is exercised by the schema-validation tests and the bundle key by the back-off test.

Test counts: 1343 passed before, 1352 passed after (the nine new tests).

alexeyshurygin:
Hey whatsup?

PR #3648, opened 2026-09-10
[JEWEL-1442] Add new constructor to AnAction supporting IconDescriptor

wellingtoncosta:
Adds a new constructor to the `AnAction` that receives both `javax.swing.Icon` and `com.intellij.platform.icons.Icon` a.k.a `IconDescriptor`, updating its Presentation API accordlingly. When the flag is off, it returns a dummy icon descriptor, wich means the IDE will look and behave exactly as it does today. When the flag is on, it returns the actual icon descriptor through the new APIs, but keeping the old path unchanged.

The `ActionButton` also listens to the flag so it can decide wether to use the new rendering pipeline in the consumption side, so that the actions in the toolbars are being rendered using the new icons API.


jakub-senohrabek (approved):

platform/platform-impl/src/com/intellij/openapi/actionSystem/impl/ActionButton.java
```diff
@@ -468,7 +475,17 @@ public void setIcon(Icon icon) {
   }
 
   public void updateIcon() {
-    myIcon = myPresentation.getIcon();
+    if (ExperimentalIcons.isEnabled()) {
+      com.intellij.platform.icons.Icon descriptor = myPresentation.getIconDescriptor();
+      if (descriptor != myIconDescriptor || myDescriptorSwingIcon == null) {
+        myIconDescriptor = descriptor;
+        myDescriptorSwingIcon = descriptor == null ? EmptyIcon.ICON_16 : SwingIconKt.toSwingIcon(descriptor, IconScale.Default);
```
rock3r, line 482: Can we make this a visible missing-descriptor marker instead of an empty icon? The point of the flag is to make unmigrated actions obvious, but this makes them look like they just have no icon.

platform/platform-impl/src/com/intellij/openapi/actionSystem/impl/ActionButton.java
```diff
@@ -468,7 +475,17 @@ public void setIcon(Icon icon) {
...
+      }
+      myIcon = myDescriptorSwingIcon;
+    }
+    else {
+      myIcon = myPresentation.getIcon();
+    }
     // set disabled icon if it is specified
     if (myPresentation.getDisabledIcon() != null) {
```
rock3r, line 490: Can we keep the descriptor path for the hover and disabled states too? `getIcon()` can still choose `hoveredIcon`, and an explicit `disabledIcon` is assigned here, so an action can flip back to a legacy icon once interacted with.

platform/editor-ui-api/src/com/intellij/openapi/actionSystem/Presentation.java
```diff
@@ -288,11 +304,52 @@ public void setDescription(@ActionDescription String description) {
...
+    return iconDescriptor;
+  }
+
+  @ApiStatus.Experimental
+  public void setIconDescriptor(@Nullable com.intellij.platform.icons.Icon iconDescriptor) {
+    setIconDescriptorSupplier(iconDescriptor == null ? null : () -> iconDescriptor);
+  }
+
+  @ApiStatus.Experimental
+  public void setIconDescriptorSupplier(@Nullable Supplier<? extends com.intellij.platform.icons.Icon> iconDescriptor) {
```
rock3r, line 333: Should we avoid evaluating descriptor suppliers and emitting this event while experimental icons are off? The descriptor is not consumable in that mode, but `copyFrom()` still reaches this setter for every action update.

platform/editor-ui-api/src/com/intellij/openapi/actionSystem/Presentation.java
```diff
@@ -288,11 +304,52 @@ public void setDescription(@ActionDescription String description) {
...
+    if (myListeners.isEmpty()) {
+      this.iconDescriptor = iconDescriptor;
+      return;
+    }
+
+    com.intellij.platform.icons.Icon oldValue = this.iconDescriptor == null ? null : this.iconDescriptor.get();
+    this.iconDescriptor = iconDescriptor;
+    com.intellij.platform.icons.Icon newValue = iconDescriptor == null ? null : iconDescriptor.get();
```
rock3r, line 341: Can we match `setIconSupplier()` and skip the property event when the old and new descriptors are equal? `copyFrom()` invokes this on each update, so unchanged descriptors currently force `ActionButton` to update and repaint.

platform/editor-ui-api/src/com/intellij/openapi/actionSystem/AnAction.java
```diff
@@ -139,6 +139,34 @@ public AnAction(@NotNull @ActionText Supplier<String> text,
...
+   * @param text           serves as a tooltip when the presentation is a button,
+   *                       and the name of the menu item when the presentation is a menu item (with mnemonic)
+   * @param description    describes the current action,
+   *                       this description will appear on the status bar when the presentation has the focus
+   * @param iconDescriptor the action's icon for the experimental icons pipeline
+   * @param icon           the action's icon for the legacy pipeline
+   */
+  @ApiStatus.Experimental
+  public AnAction(@NotNull @ActionText Supplier<String> text,
+                  @Nullable @ActionDescription Supplier<String> description,
```
rock3r, line 156: Could we add `intellij.platform.icons.api` as a direct `editor-ui-api` dependency and regenerate Bazel? This module now exposes `com.intellij.platform.icons.Icon` in its public API; it currently arrives only incidentally through `core-ui`.

platform/editor-ui-api/src/com/intellij/openapi/actionSystem/Presentation.java
```diff
@@ -288,11 +304,52 @@ public void setDescription(@ActionDescription String description) {
...
+    if (!ExperimentalIcons.isEnabled()) {
+      return NO_ICON_DESCRIPTOR;
+    }
+    Supplier<? extends com.intellij.platform.icons.Icon> iconDescriptor = this.iconDescriptor;
+    return iconDescriptor == null ? null : iconDescriptor.get();
+  }
+
+  /** The raw supplier, not gated by {@link ExperimentalIcons}. */
+  @ApiStatus.Experimental
+  public @Nullable Supplier<? extends com.intellij.platform.icons.Icon> getIconDescriptorSupplier() {
```
rock3r, line 323: Nit: can we keep the nullable result annotation from `getIconSupplier()` / `setIconSupplier()` here? A descriptor supplier can return null at runtime, but this signature exposes it as non-null to Kotlin callers.

rock3r (changes requested):
Quick AI review

wellingtoncosta:
@rock3r could you please review again? I addressed your comments.

rock3r (approved):

wellingtoncosta:
Ready to merge

wellingtoncosta:
Ready to merge

PR #3649, opened 2026-09-10
IJPL-252865 Right-align Copy/Path Reference popup values

bric3:
Super simple fix to align the value in the _Copy Path/Reference_ popup with its right edge.

Also keeps the action label and shortcut in their existing columns.

Original issue https://youtrack.jetbrains.com/issue/IJPL-252865/When-copying-paths-things-are-not-aligned

**Before**
image

**After**
image


bric3:
Thanks @BartvHelvert

PR #3652, opened 2026-09-11
[JEWEL-1265] Fix Speed Search Selection Being Overriden By Mouse Hover

DanielSouzaBertoldi:
## Context

When speed search is active in a `SpeedSearchableComboBox`, the popup list highlighted whatever row the mouse was resting on rather than the row the search matched. The header named the right item and the matched characters got their highlight, but the selection background stayed put under the pointer.

`ListComboBoxImpl` tracks a `hoveredItemIndex`, a preview selection that takes precedence over the real one so that hovering highlights exactly one row but nowhere in the code we were "telling" this variable that it was stale. Speed search moves the selection without the pointer moving, so a leftover hover position kept vetoing the correct highlight until the user nudged the mouse.

## Changes

* `ListComboBoxImpl` now clears the hover preview when the list selection changes or the speed search query changes, using one `snapshotFlow` that reads both.
* Two signals are needed, not one. If the query matches the item that is already selected, `SpeedSearchableLazyColumnScrollEffect` leaves that selection alone and nothing in `listState` changes, so watching the selection would miss it entirely. That was the harder half of this bug: typing the name of the selected item left the highlight stranded on the hovered row.
* `ListComboBoxImpl` takes a new `speedSearchState` parameter that `SpeedSearchableComboBox` fills in. Both the function and the parameter are internal, so the public API is untouched and the Metalava dumps pick up nothing from this change.
* Fixed the showcase "Speed Search API" sample. Its "Selected item:" label indexed `stringItems` while the combo box rendered a different list, so the label showed an unrelated string.
* Added regression tests for both paths: a query that moves the selection, and one that matches the item already selected.

## Screen recordings

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/fd1150ef-ebf9-4ed8-9a88-49b85fcae466  |   https://github.com/user-attachments/assets/edf94ca6-4f54-4fd2-b033-638f4dac4905 |

### Bug fixes

* Speed search in `SpeedSearchableComboBox` now highlights the matched item even when the mouse is resting over another row in the popup.

### DIFF
--- a/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/ComboBoxes.kt
+++ b/platform/jewel/samples/showcase/src/main/kotlin/org/jetbrains/jewel/samples/showcase/components/ComboBoxes.kt
@@ -143,13 +143,14 @@ private fun ListComboBoxes() {
 
         Column(Modifier.weight(1f).widthIn(min = 125.dp), verticalArrangement = Arrangement.spacedBy(8.dp)) {
             Text("Speed Search API")
+            val speedSearchItems = listOf("one", "two", "three")
             var selectedIndex by remember { mutableIntStateOf(2) }
-            val selectedItemText = if (selectedIndex >= 0) stringItems[selectedIndex] else "[none]"
+            val selectedItemText = if (selectedIndex >= 0) speedSearchItems[selectedIndex] else "[none]"
             InfoText(text = "Selected item: $selectedItemText")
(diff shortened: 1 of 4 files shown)


rock3r (approved):

wellingtoncosta (approved):

nebojsa-vuksic (approved):

DanielSouzaBertoldi:
Ready to merge

PR #3654, opened 2026-09-14
[JEWEL-1438] Fix renderer flag leak causing cascading Spectre test failures

DanielSouzaBertoldi:
# Context

`PopupRendererParitySpectreTest` toggles the global `JewelFlags.useCustomPopupRenderer` so it can compare `JDialogRenderer` against Compose's own popup renderer. That assignment sits outside the method's `try` block, so when a measurement fails, the flag keeps whatever the last one gave it. That would be harmless if each test got its own JVM, but that's not the case: Bazel runs the whole Spectre suite through a single `java_test` (`ConsoleLauncher` with `--select-package=org.jetbrains.jewel`), so every later test sees the leaked value and fails its `spectreTest must enable JDialogRenderer` precondition. One flaky window gets reported as six failing tests, five of them pointing nowhere near the real problem.

[This surfaced on the Windows CI lane](https://github.com/JetBrains/intellij-community/actions/runs/34625527480/job/103349576370?pr=3409), where `MatrixApp.awaitWindow()` gave up waiting for the first Compose window. The run died at 10833 ms against a 10s budget and printed zero `PARITY` lines, so it never finished a single measurement. Nothing threw; Skiko's native library had loaded fine. The Linux lane passed on the same commit, and the commit before it on that branch differed only by three lines of KDoc, so this was not a regression. The first window in a JVM is the only one paying one-time renderer initialisation, and that is where it failed.

## Changes
 * Capture `JewelFlags.useCustomPopupRenderer` before `measure()` overwrites it, and restore it in the `finally` block, so a failed measurement stops leaking its renderer choice into unrelated tests. Restoring the captured value instead of hardcoding `true` keeps this correct if the harness default ever changes.
 * Move `app.start()` inside the `try`, so a failure starting the application is covered by the same `finally`.
 * Widen the `awaitWindow()` budget from 10s to 30s (`repeat(100)` to `repeat(300)`). The loop returns as soon as the window appears, so this costs nothing when things go well, and only changes how long a stuck window waits before failing.


rock3r:
Ah had noticed this too but hadn't had time to file/fix it yet! Thanks dude <3

rock3r (approved):

DanielSouzaBertoldi:
No problemo! 😎 giving that this is a rather small fix, I guess it is... *drumroll* Ready to merge

rock3r:
Ludicrous speed

PR #3655, opened 2026-09-15
IJPL-250885 let a container skip IdeKeyEventDispatcher for a focused descendant

rock3r:
`KeyboardAwareFocusOwner` only works when the exact AWT focus owner implements it. Inside a `ComposePanel` the focus owner is an internal skiko canvas the embedder cannot subclass, so Compose content hosted in the IDE cannot claim a key event before the keymap runs an action on it. This is the only platform change Jewel's shortcut dispatch needs; everything else lives in Jewel.

`KeyboardAwareContainer` (platform-impl, experimental) is the ancestor-side counterpart. When `IdeKeyEventDispatcher` starts matching a fresh key event against the keymap, it asks the focus owner's implementing ancestors, innermost first, and stands down when one returns `true`. Nothing in the platform implements it, so dispatch is unchanged until a container opts in.

Two decisions worth reviewing:

- The check sits at the fresh-stroke entry point, not at the top of `dispatchKeyEvent`, so the IDE keeps priority where it already has state: the exact-owner hatch is asked first, a pending chord completes without asking, and the typed event of a shortcut the IDE just performed is still swallowed.
- The ancestor chain is resolved once per focus owner and cached. A hierarchy listener on the owner drops the cache when anything above it is reparented, which Swing reports synchronously inside `add` and `remove`, so reuse only needs the owner's identity.

Returning `true` does not consume the event; it only skips keymap matching. `KeyboardAwareContainerTest` covers the contract, the precedence rules, and every cache invalidation case.

```
./tests.cmd --module intellij.platform.tests --test com.intellij.openapi.keymap.KeyboardAwareContainerTest
```


platform/platform-impl/src/com/intellij/openapi/keymap/impl/IdeKeyEventDispatcher.kt
```diff
@@ -435,9 +439,74 @@ class IdeKeyEventDispatcher(private val queue: IdeEventQueue?) {
...
+   * Asks the [KeyboardAwareContainer]s above the focus owner, innermost first, whether to skip keymap matching for this
+   * event. Called from [inInitState] only, so the IDE keeps priority while it completes a multi-stroke shortcut or swallows
+   * the typed event of a shortcut it just performed, and only after the focus owner itself declined as a
+   * [KeyboardAwareFocusOwner]. It runs before the Windows AltGr check, so a claimed Ctrl+Alt press is not held back as a
+   * possible AltGr stroke. Like the other guards that return before the state machine, a claim is not passed to
+   * [keyGestureProcessor].
+   */
+  private fun skipByKeyboardAwareContainer(focusOwner: Component, e: KeyEvent): Boolean {
+    for (reference in resolveContainersFor(focusOwner)) {
+      val container = reference.get() ?: continue
```
knisht, line 475: can a component even be collected in the middle of the ancestor chain? I feel like if the focus owner is strongly reachable, so are all its ancestors

knisht:
but anyway LGTM

rock3r:
Kostia, was there a set of comments that did not get posted by any chance?

knisht:
nope, that was it :)

platform/platform-impl/src/com/intellij/openapi/keymap/impl/IdeKeyEventDispatcher.kt
knisht: though I guess this is the only way to avoid holding strong references to the whole chain

rock3r:
Ah ok :) The "but anyway" confused me :D Can you approve the PR then? We'll likely need the approval for the merge bot to take it in

knisht (approved):

knisht:
Is merging done through a bot nowadays? I thought we have to do the old-style manual patch application

platform/platform-impl/src/com/intellij/openapi/keymap/impl/IdeKeyEventDispatcher.kt
rock3r: Yeah that's the general goal, my idea is that this would also guard against potential future changes for very little cost

rock3r:
I think the Jewel merge bot can do IJPL as well. Worst case scenario @nebojsa-vuksic can do the MR in Space and Patronus run by hand

rock3r:
@DanielSouzaBertoldi I added you as reviewer just in case, if you can't get to it soon it's fine, just set the bot to merging it

DanielSouzaBertoldi:
I'll take a look at this PR today but I'm pretty sure Nebojsa will have to handle the merge, because AFAIK the bot does not transfer approvals :( but that's fine!

DanielSouzaBertoldi (approved):
LGTM! I'll ask the bot to try and merge it, but like I said i'm pretty sure we'll need a JB developer to merge this PR into the monorepo 🙏🏻

DanielSouzaBertoldi:
Ready to merge

PR #3656, opened 2026-09-15
[JEWEL-991] Match JPopupMenu pointer and focus behavior in submenus

DanielSouzaBertoldi:
## Context

It's finally time to fix an issue that gets on my nerves everytime I'd see it: hovering items in a parent menu should clear the focus of any submenus opened.

The fix for this was exactly one line of code after the [CMP team provided a way on their end to achieve this result](github.com/jetbrains/compose-multiplatform-core/pull/2992), which is cool, but then I thought: "maybe there's room to fix any other issues in menus that I can find" and there were a couple!

So pardon my audacity but please, do add any suggestions/changes you feel might be needed 😄 

## Changes

* `Submenu` popups now set `consumePointerInputOutside = false`, so the submenu layer stops eating pointer events outside itself and the parent menu sees the pointer come back. The flag defaults to the value of `focusable`, so the previous `PopupProperties(focusable = true)` was implicitly opting into the blocking behavior.
* `MenuContent` clears focus when the pointer leaves the menu, unless one of its items is currently holding a submenu open. That exception is Swing's rule: the selection path continues through the item that owns the open submenu, so that item stays armed. This lives on the menu rather than on each item on purpose. Items arm themselves by requesting focus, so a sibling handoff needs no bookkeeping, and putting the disarm on the item instead makes it race that handoff, since `clearFocus` is popup-wide and can wipe the focus a sibling has just taken. That showed up as an item refusing to light up when you moved the pointer back up the list, visible only in the IJP bridge, where `isSwingCompatMode` means hover alone paints nothing.
* Fixed the right-hand space calculation in `AnchorHorizontalMenuPositionProvider`, which decides whether a submenu opens to the right or flips to the left. It used `contentOffsetY` where it meant `contentOffsetX`, and `windowSpaceBounds.width` where it meant `windowSpaceBounds.right`, so the space it measured was off by the vertical offset and by the left margin.
* Reworked the menus row in devkit's `SwingComparisonTabPanel` so the Jewel side can actually show any of this. The compose cell now reserves room for a submenu to open beside its parent, and the button sits in its own box sized to fit it, so the popup keeps anchoring to the button rather than to the sizing box. Both buttons are top aligned, since the space below them is only there to host the popup.
* Added `MenuItemFocusTest` coverage for pointer and keyboard arming: disarm on leaving the menu, focus moving between siblings in both directions, and arrow navigation into a freshly opened submenu, where down picks the first item and up the last, for submenus opened by keyboard or by hover. It also covers that `DirectionDown` never opens a submenu, while `DirectionRight` does and `DirectionLeft` closes it. The backwards sibling case needs the exit and the enter inside a single `performMouseInput` gesture, because separate calls settle between them and never reproduce the handoff.
* Added `MenuItemArmedHighlightTest` for the difference between Swing compat modes. With `swingCompatMode = false` hover outranks focus in `chooseValue`, so a hovered item stays armed while the keyboard arms another one; with it on, only the focused item is armed. Both modes produce identical semantics, so these two tests compare rendered pixels.

## Screenshots/screen recordings

| Platform | Before | After |
|---|--------|--------|
| Standalone | https://github.com/user-attachments/assets/eb39dc00-d958-442a-899c-02be093e1c47 | https://github.com/user-attachments/assets/4ef43fa3-473a-47e2-adfc-447aba2e75c4 |
| Bridge | https://github.com/user-attachments/assets/aac1a76a-016f-4116-b7a9-388ed71ad56e | https://github.com/user-attachments/assets/c79202bd-e4d9-47de-9de8-67ab5ded26d5 |
 
### SwingComparisonTabPanel

| Hovering | Keyboard |
|--------|---|
| https://github.com/user-attachments/assets/5c82636d-c2cc-4964-8329-646587af5783 | https://github.com/user-attachments/assets/b7624f54-420f-4a13-8264-31f97924ecca |

### Clearing focus when reaching popup's border

| Swing | Bridge | Standalone |
|--------|--------|--------|
| https://github.com/user-attachments/assets/6b47092c-06df-432b-90df-7eba11bdc6ff | https://github.com/user-attachments/assets/7d62d42d-4cca-44eb-8010-12cd9c10be4d | https://github.com/user-attachments/assets/aa5fa748-ce2f-467a-9e5a-52f82c95382c |

### Hovering + Keypress (Standalone only)

| Swing Compat On | Swing Compat Off |
|--------|--------|
| https://github.com/user-attachments/assets/de79b2d3-d56c-457a-bbab-47cea0a1542e | https://github.com/user-attachments/assets/73fc3068-0740-4b7e-8cff-09a615a89bbb |

### Bug fixes

* Menus with submenus no longer freeze the parent menu while a submenu is open. Moving the pointer back onto the parent menu arms its items again, the way it does in `JPopupMenu`.
* Menu items no longer stay highlighted after the pointer leaves the menu, so an open submenu and its parent menu can no longer show two armed items at once.
* Fixed the calculation that decides whether a submenu opens to the right of its parent or flips to the left, which could pick the wrong side when the menu had a left margin.


platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
```diff
@@ -74,6 +75,39 @@ class CustomPopupRendererSpectreTest {
...
     }
 
+    @Disabled(
+        "JDialogRenderer puts the popup in its own native window, and Compose inside it never receives a pointer " +
+            "exit when the cursor leaves that window, so onHover(false) never fires and hover-driven state stays " +
+            "stuck. Needs a renderer-level signal. Enable once that exists."
+    )
+    @Test
+    fun `moving the pointer off a menu popup disarms its item`(): Unit = runSpectreTestWithCustomPopupRenderer {
+        val app = SpectreTestApplication()
```
DanielSouzaBertoldi, line 85: I want your opinion on this. Currently, like the description above says so, our popups that use the custom JDialogRenderer does not clear focus when the cursor hits the popups menu edges:

https://github.com/user-attachments/assets/42321c30-f7be-4986-bdb9-cf75db76b039

The submenus close when the parent menu focuses on an item that did not open the submenu, that's ok. The only minor difference is the focus not being cleared when that happens.

So:

Add the fix to this PR, or file a new one? It's gonna be a bit trickier to achieve the same on the custom renderer because we'd have to either:

1. Synthesize a `MOUSE_EXITED` into the scene, which requires messing with reflection into Compose internals
2. Have menu code reach the panel through `LocalComponent`, but that is worse, because `:ui` has to stay renderer-agnostic for the bridge. 
3. ????????????? honestly idk if there's a third option

We'd have to basically create a new `PopupRenderer` signal (pointer-inside, or a synthesized hover exit)...

rock3r: Good catch! We can do it in a follow-up :) Since as you say it is tricky and we can't guarantee it will work properly, no point blocking this fix!

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
```diff
@@ -371,11 +374,17 @@ public fun MenuContent(
...
                 .background(colors.background, menuShape)
                 .width(IntrinsicSize.Max)
-                .onHover { localMenuController.onHoveredChange(it) }
+                .onHover { hovered ->
+                    localMenuController.onHoveredChange(hovered)
+
+                    // Items arm themselves by taking focus, so moving between them needs no bookkeeping here. Only
+                    // leaving the menu has to disarm it, and not while one of its items is holding a submenu open:
+                    // Swing keeps that item armed because the selection path continues through it.
+                    if (!hovered && selectedSubMenu == null) focusManager.clearFocus(force = true)
```
rock3r, line 383: Can we base this guard on whether a submenu is actually open? `selectedSubMenu` tracks hover selection, but keyboard opening and closing only update the item's local `isSelected` state.

I think this gives us two cases:
* Hover a regular item, open a submenu with the arrow keys, then move outside the parent without hovering the submenu owner. `selectedSubMenu` is still null, so we clear the owner's focus despite the open submenu.
* Hover a submenu owner, press Left to close its submenu, then move outside the parent. `selectedSubMenu` is still non-null, so we skip clearing focus even though the submenu is closed.

Unless I am misreading the whole thing, this could be an issue

rock3r (changes requested):

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/popup/CustomPopupRendererSpectreTest.kt
DanielSouzaBertoldi: Filed [JEWEL-1439](https://youtrack.jetbrains.com/issue/JEWEL-1439/Custom-JDialogRenderer-popups-do-not-clear-focus-when-the-cursor-hits-the-edge-of-the-popup)!

DanielSouzaBertoldi: btw i can delete these disabled tests too now that i've filed a ticket just for this

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: You're right, great catch!

I didn't think of these edge cases so I was testing keyboard and pointer navigation separately, but we can definitely do this better :)

Took some refactoring and messing with internal api, but here it is:

Case 1: opening submenu using the keyboard, then leaving the menu using the cursor
Case 2: closing the submenu using the keyboard, then leaving the menu using the cursor

| Scenario | Before | After |
|--------|--------|--------|
| Case 1 | https://github.com/user-attachments/assets/fc9f9d03-13b1-4fd1-aff1-6b25e47020a0 | https://github.com/user-attachments/assets/ccfcdf3c-a092-45c4-a576-e2a1b7b3e2c9 |
| Case 2 | https://github.com/user-attachments/assets/6f0ac3a5-f5f3-4b3c-bd00-ff0bb6111c20 | https://github.com/user-attachments/assets/a06c1599-5677-451e-9ca7-2b8ff01c0975 |

Also added two tests covering these exact scenarios

rock3r (approved):

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
wellingtoncosta, line 912: KDoc should be updated since you added the `onSelectedChange` param.

plugins/devkit/intellij.devkit.compose/src/demo/SwingComparisonTabPanel.kt
```diff
@@ -820,38 +820,38 @@ internal class SwingComparisonTabPanel : BorderLayoutPanel() {
...
-              selectableItem(selected = false, onClick = {}) { Text("Menu Item 3") }
-              separator()
-              submenu(submenu = {
-                selectableItem(selected = false, onClick = {}) { Text("Submenu Item 1") }
-                selectableItem(selected = false, onClick = {}) { Text("Submenu Item 2") }
-                selectableItem(selected = false, onClick = {}) { Text("Submenu Item 3") }
-              }) {
-                Text("Submenu")
+        Box(modifier = Modifier.size(width = 460.dp, height = 200.dp)) {
+          Box(modifier = Modifier.align(Alignment.TopStart)) {
```
wellingtoncosta: Isn't this redundant since `Box` already has `contentAlignment` defaulting to `Alignment.TopStart`?

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/menu/MenuItemFocusTest.kt
```diff
@@ -296,6 +348,290 @@ class MenuItemFocusTest {
...
+
+        composeRule.onNodeWithTag("container").performKeyInput { pressKey(Key.DirectionLeft) }
+        composeRule.waitForIdle()
+
+        composeRule.onNode(hasText("Submenu item 1").and(hasAnyAncestor(isPopup()))).assertDoesNotExist()
+    }
+
+    @Test
+    fun `key press down in a submenu opened by keyboard focuses its first item`() {
+        composeRule.setContent { IntUiTheme { MenuWithSubmenu() } }
```
wellingtoncosta, line 494: I think the 4 tests below can reuse a function to avoid duplicating code. Something like:

```kotlin
@Test
fun `key press down in a submenu opened by keyboard focuses its first item`() {
    assertArrowFocus(
        ::openSubmenuWithKeyboard,
        Key.DirectionDown,
        focused = "Submenu item 1",
        unfocused = "Submenu item 2"
    )
}

private fun assertArrowFocus(open: () -> Unit, key: Key, focused: String, unfocused: String) {
    composeRule.setContent { IntUiTheme { MenuWithSubmenu() } }
    composeRule.waitForIdle()
    open()
    composeRule.waitForIdle()
    composeRule.onNodeWithText("Submenu item 1").assertIsDisplayed()
    composeRule.onNodeWithText("Submenu item 1").assertIsNotFocused()
    composeRule.onNodeWithText("Submenu item 2").assertIsNotFocused()

    composeRule.onNodeWithTag("container").performKeyInput { pressKey(key) }
    composeRule.waitForIdle()
    composeRule.onNodeWithText(focused).assertIsFocused()
    composeRule.onNodeWithText(unfocused).assertIsNotFocused()
}
```

wellingtoncosta (approved):
Left some comments, but LGTM overall

platform/jewel/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/Menu.kt
DanielSouzaBertoldi: Not really, it's not public API 😎 this is not even a proper "KDoc" per se, it's just a comment explaining why do we need to make it internal so that we can test it

wellingtoncosta: Ah yeah right, `@InternalJewelApi`. Missed that.

platform/jewel/ui-tests/src/test/kotlin/org/jetbrains/jewel/ui/component/menu/MenuItemFocusTest.kt
DanielSouzaBertoldi: I agree, but I feel like this adds a heavier cognitive load trying to read it than what we currently have, where you can exactly see the step by step instead of going up and down between functions

DanielSouzaBertoldi:
Ready to merge

DanielSouzaBertoldi:
Since I changed the SwingComparisonPanel code I need your approval too @nebojsa-vuksic 😭

nebojsa-vuksic (approved):

PR #3658, opened 2026-09-18
[JEWEL-1397, JEWEL-1441] Run showcase e2e scenarios in both hosts

rock3r:
`JBPopupRenderer` draws Jewel popups inside the IDE. No test covered it: reverting the whole file left every test passing. This PR adds end-to-end tests that drive the components showcase in a real IDE and in the standalone sample. Both hosts run the same scenarios, so they must behave the same way. The first scenarios check that Escape closes a ComboBox popup. `platform/jewel/docs/e2e-tests.md` explains how to run the tests and how to add scenarios.

## Changes

* `platform/jewel/showcase-e2e` contains the scenarios. They use a small `ShowcaseAutomator` interface and depend on neither Spectre nor the IntelliJ Platform.
* `jewel-bridge-e2e-tests`, defined with `bridge-e2e.bzl`, runs the scenarios in the IDE as a normal `jps_test`. The IDE Starter builds a dev IDE from sources and Spectre attaches to it, so the IDE needs no Spectre dependency. `ShowcaseScenariosSpectreTest` runs the same scenarios in the standalone sample.
* The IDE test presses keys with the IDE Driver's robot. When Spectre is attached to another process, its key events do not go through `IdeEventQueue`, so they never reach the `JBPopup` key handler. With Spectre pressing Escape, the test passed even when `JBPopupRenderer` ignored Escape. With the robot, it fails. The test first checks that the IDE receives OS keyboard input.
* Spectre goes from 0.5.0 to 0.6.0. It fixes the agent socket path on Windows (rock3r/spectre#445) and adds the `spectre-input-coordinator` dependencies.
* The standalone Spectre tests now wait up to 60 seconds for their first window, through one shared helper. They waited 10 seconds, which is too short on a cold CI runner ([JEWEL-1441](https://youtrack.jetbrains.com/issue/JEWEL-1441)).
* GitHub runs the IDE test in two new jobs, `Jewel IDE end-to-end tests` on Linux and on Windows. It has its own jobs because its build compiles the whole platform, and the standalone tests time out when they run during that build. The Linux job frees disk space and adds swap. The Windows job uses `--enable_runfiles` and puts Bazel's output on the D: drive, because the test's runfiles do not fit on C:. All four jobs check out the android repository, which the JPS module graph needs.
* Fixes outside Jewel, needed for a standalone community checkout:
  * The IDE Starter no longer requires an Ultimate checkout.
  * `Git.getRepoRoot()` runs git in the IDE home directory, not in the Bazel runfiles directory.
  * On Windows with `--enable_runfiles`, `JUnit5BazelRunner` now starts, and it sets `idea.home.path` to the checkout.

## Testing

Both tests pass on macOS, on Linux under Xvfb, and on Windows. A few unrelated `jps_test` targets also pass on all three, because the runner change affects every `jps_test`. They also pass on GitHub's Linux and Windows runners. On Windows both tests need an interactive desktop: in a service session, Windows drops the robot's key events.

The Windows `Jewel Spectre tests` job can still fail on `PopupRendererParitySpectreTest`, case `submenu(nested)`. This PR does not touch that test. [JEWEL-1440](https://youtrack.jetbrains.com/issue/JEWEL-1440) tracks it.


platform/jewel/docs/e2e-tests.md
```diff
@@ -0,0 +1,88 @@
...
+Jewel has end-to-end tests that drive the components showcase in both hosts: the standalone sample and the IDE. The
+same scenarios run in both, so a behaviour that works in one host and breaks in the other fails the build.
+
+The tests use [Spectre](https://spectre.sebastiano.dev) to find and click Compose nodes. They need a display.
+
+## What runs where
+
+| | Standalone | IDE |
+|---|---|---|
+| Target | `//platform/jewel/int-ui/int-ui-standalone-tests:jewel-intUi-standalone-spectre-tests` | `//platform/jewel/ide-laf-bridge:jewel-bridge-e2e-tests` |
```
DanielSouzaBertoldi: this got me thinking.... why not rename so that both share `e2e-tests`? It's easier to remember the commands when the only thing that changes is the target where they reside (`int-ui-standalone-tests`/`ide-laf-bridge`)

The same could be applied to other naming conventions down below:

1. `StandaloneShowcaseE2ETest` instead of `ShowcaseScenariosSpectreTest`
2. macro `standalone-e2e.bzl` instead of `spectre.bzl`

platform/jewel/ide-laf-bridge/src/e2eTest/kotlin/org/jetbrains/jewel/bridge/e2e/JewelBridgeShowcaseE2ETest.kt
```diff
@@ -0,0 +1,112 @@
...
+import org.jetbrains.jewel.e2e.PopupScenarios
+import org.junit.jupiter.api.Test
+
+/**
+ * Runs the shared showcase scenarios against Jewel as the IDE hosts it. For the popup scenarios that makes
+ * `JBPopupRenderer` the renderer under test.
+ *
+ * Nothing cheaper reaches it. It builds a real `JBPopup` through `JBPopupFactory`, so it needs a running application
+ * and a display, and before this lane existed reverting the whole file left every test that compiles the bridge module
+ * green.
```
DanielSouzaBertoldi: Not sure if the last part of this comment is relevant 🤔 was this a specific hurdle you encountered while developing this?

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRendererTest.kt
```diff
@@ -0,0 +1,79 @@
...
+        assertTrue(
+            "JBPopupRenderer must declare its own override of every PopupRenderer.Popup overload, or the interface " +
+                "default and the bridge's deprecated overload will delegate to each other forever. Inherited: " +
+                inherited.joinToString { it.toGenericString() },
+            inherited.isEmpty(),
+        )
+    }
+
+    @Test
+    fun `implements the windowShape-aware overload`() {
```
DanielSouzaBertoldi, line 29: so the idea here is that, whenever we remove or add new overloads, we'll need new tests? I get the reasoning behind it, but isn't this a bit overkill tho?

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRendererTest.kt
```diff
@@ -0,0 +1,79 @@
...
+        )
+    }
+
+    @Test
+    fun `keeps both overloads composable`() {
+        val notComposable =
+            JBPopupRenderer::class
+                .java
+                .declaredMethods
+                .filter { it.name == "Popup" && !it.isSynthetic }
```
DanielSouzaBertoldi: Can't you just use `popupOverloads` here instead? It already fetches all popups that aren't synthetic, but it filters on `PopupRenderer` instead of `JBPopupRenderer`. All other tests do use `popupOverloads`

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/showcase/StandaloneShowcaseE2ETest.kt
```diff
@@ -0,0 +1,110 @@
...
+    fun `the showcase behaves in the standalone host`(): Unit = runSpectreTest {
+        assertEquals(true, JewelFlags.useCustomPopupRenderer, "spectreTest must enable JDialogRenderer")
+
+        val app = ShowcaseApplication()
+        app.start()
+        val failures = mutableListOf<String>()
+        try {
+            val automator =
+                SpectreShowcaseAutomator(ComposeAutomator.inProcess(RobotDriver.synthetic(app.awaitWindow())))
+            PopupScenarios.runAll(automator, failures)
```
DanielSouzaBertoldi, line 51: Currently this test scenario only runs scenarios for popups. Let us say that in the future we need to add scenarios for the got it tooltip (which btw i'm still mad its called a 'tooltip'. its ANYTHING BUT that. anywho), then we'd have to add it here too or the idea is to have `@Test` functions for each scenario?

Given the name of this test I'd expect them all to live under here

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/showcase/SpectreShowcaseAutomator.kt
```diff
@@ -0,0 +1,45 @@
...
+package org.jetbrains.jewel.intui.standalone.showcase
+
+import dev.sebastiano.spectre.core.ComposeAutomator
+import org.jetbrains.jewel.e2e.ShowcaseAutomator
+
+/**
+ * Backs the shared popup scenarios with Spectre's in-process automator, against the standalone sample.
+ *
+ * The IDE lane backs the same scenarios with Spectre attached to a running IDE. Keeping both behind one interface is
+ * what stops the two hosts drifting: a popup fix that only lands for one of them fails here.
```
DanielSouzaBertoldi: I understand that popups are the primary case for this PR, but it'll not be the only one in the long run, right? I guess that removing the last part of this sentence is enough to convey the idea behind having both "runners" behind one shared interface

```suggestion
 * what stops the two hosts drifting.
```

platform/jewel/showcase-e2e/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
...
+# Platform class reaches its runtime closure, and Spectre types would tie the scenarios to one host's automator.
+jvm_library(
+    name = "showcase-e2e-scenarios",
+    testonly = True,
+    srcs = glob(
+        ["src/main/kotlin/**/*.kt"],
+        allow_empty = True,
+    ),
+    module_name = "intellij.platform.jewel.showcaseE2eScenarios",
+    visibility = ["//platform/jewel:__subpackages__"],
```
DanielSouzaBertoldi, line 17: Why do we need this subpackages visibility? :o (asking out of curiosity, never encountered this one before)

tools/intellij.tools.ide.starter/src/com/intellij/ide/starter/ide/installer/IdeFromCodeInstaller.kt
```diff
@@ -39,6 +39,12 @@ import kotlin.io.path.pathString
 class IdeFromCodeInstaller(private val useInstallationCache: Boolean = true) : IdeInstaller {
   private val projectRoot by lazy { Path.of(PathManager.getHomePath(false)) }
 
+  // community is a subdirectory only in an Ultimate checkout; in a standalone community checkout the project
+  // root already is the community root. Mirrors IdeaProjectLoaderUtil.guessCommunityHome.
+  private val communityRoot by lazy {
+    projectRoot.resolve("community").takeIf { java.nio.file.Files.isDirectory(it) } ?: projectRoot
```
DanielSouzaBertoldi: nit: Could we please import `java.nio.file.Files` instead of referencing it by its "full legal name"? :P

tools/intellij.tools.ide.starter/src/com/intellij/ide/starter/ide/installer/IdeFromCodeInstaller.kt
```diff
@@ -130,8 +136,10 @@ class IdeFromCodeInstaller(private val useInstallationCache: Boolean = true) : I
         }
         finalVMOptions.writeJavaArgsFile(argsFile)
 
-        val openedPackages = GlobalPaths.instance.checkoutDir
-          .resolve("community/platform/platform-impl/resources/META-INF/OpenedPackages.txt")
+        val openedPackagesRoot = GlobalPaths.instance.checkoutDir
+          .let { it.resolve("community").takeIf { c -> java.nio.file.Files.isDirectory(c) } ?: it }
```
DanielSouzaBertoldi: Same here

tools/intellij.tools.ide.starter/src/com/intellij/ide/starter/ide/JBRResolver.kt
```diff
@@ -153,7 +153,11 @@ object JBRResolver {
 
   fun getRuntimeBuildVersion(): String {
     val jbrVersion = ConfigurationStorage.jbrVersionForDevServer() ?: run {
-      val jbrDependencyFile = GlobalPaths.instance.checkoutDir / "community" / "build" / "dependencies" / "dependencies.properties"
+      // "community" is a subdirectory only in an Ultimate checkout; in a standalone community checkout the
+      // checkout root already is the community root.
+      val checkout = GlobalPaths.instance.checkoutDir
+      val communityRoot = (checkout / "community").takeIf { java.nio.file.Files.isDirectory(it) } ?: checkout
```
DanielSouzaBertoldi: Same here

platform/jewel/docs/e2e-tests.md
rock3r: Yup makes sense!

platform/jewel/ide-laf-bridge/src/e2eTest/kotlin/org/jetbrains/jewel/bridge/e2e/JewelBridgeShowcaseE2ETest.kt
rock3r: Ah stupid Fable, I had asked it to figure out something before dinner and was counting on checking it but it must've pushed 🫩 sorry. Hopefully there isn't much more unchecked slop around

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRendererTest.kt
rock3r: Yeah well for adding, yes. For removing, those tests would be removed. I thought I'd err on the side of caution since we've had a couple cases in the past where one of the overloads was doing weird shit because we passed the wrong default... But we can probably also do without

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/showcase/StandaloneShowcaseE2ETest.kt
rock3r: Probably different files. Good point, this was me over stressing over "yes we start from popups but then it's going to be the whole showcase" and then forgetting shit around

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/showcase/SpectreShowcaseAutomator.kt
rock3r: Fable slop I've missed, sorry

platform/jewel/showcase-e2e/BUILD.bazel
rock3r: I have no idea when it comes to bazel I just cry to the nearest LLM who'll help me. I want to learn it better but never have time 😭

tools/intellij.tools.ide.starter/src/com/intellij/ide/starter/ide/installer/IdeFromCodeInstaller.kt
rock3r: Wtf how did I miss this

DanielSouzaBertoldi (approved):

platform/jewel/showcase-e2e/BUILD.bazel
DanielSouzaBertoldi: yeah i know the feeling LOL

kinda curious tho why it's referencing the subpackages specifically! Will find out why that is 🕵🏻

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/showcase/StandaloneShowcaseE2ETest.kt
DanielSouzaBertoldi: i feel like just changing the function/file name then it's fine :) i'm mostly concerned with having one test with a bunch of `*Scenarios.runAll()` inside hehe

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRendererTest.kt
DanielSouzaBertoldi: Given the historical context then I think it's fine for popups! If we were to do this for every single deprecated overload we have we'd be coooked 😄 though it would make the overall code more resilient.... im conflicted still

DanielSouzaBertoldi:
i forgot to approve the PR yesterday, none of my comments are blockers, just minor clean up across the code so it's fine ❤️

platform/jewel/docs/e2e-tests.md
rock3r: Done: target is `jewel-intUi-standalone-e2e-tests`, test class is `StandaloneShowcaseE2ETest`, and the macro is `standalone_e2e_test` in `standalone-e2e.bzl`, to match `bridge_e2e_test`.

platform/jewel/ide-laf-bridge/src/e2eTest/kotlin/org/jetbrains/jewel/bridge/e2e/JewelBridgeShowcaseE2ETest.kt
rock3r: Removed that paragraph.

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRendererTest.kt
rock3r: Kept the two tests that catch the mutual-delegation bug, and dropped the `@Composable` one.

platform/jewel/ide-laf-bridge/src/test/kotlin/org/jetbrains/jewel/bridge/component/JBPopupRendererTest.kt
rock3r: That test is gone now, see above.

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/showcase/StandaloneShowcaseE2ETest.kt
rock3r: One class per host stays. Each scenario group gets its own object with a `runAll`, and the test class calls the groups one after the other, so a new group is one call here. `docs/e2e-tests.md` describes it.

platform/jewel/int-ui/int-ui-standalone-tests/src/spectreTest/kotlin/org/jetbrains/jewel/intui/standalone/showcase/SpectreShowcaseAutomator.kt
rock3r: Applied.

platform/jewel/showcase-e2e/BUILD.bazel
rock3r: It lets any package under `platform/jewel` depend on the scenarios, and nothing outside Jewel. Added a comment saying so.

tools/intellij.tools.ide.starter/src/com/intellij/ide/starter/ide/installer/IdeFromCodeInstaller.kt
rock3r: Fixed, in all three places.

tools/intellij.tools.ide.starter/src/com/intellij/ide/starter/ide/installer/IdeFromCodeInstaller.kt
rock3r: Fixed.

tools/intellij.tools.ide.starter/src/com/intellij/ide/starter/ide/JBRResolver.kt
rock3r: Fixed.

DanielSouzaBertoldi (approved):
Thanks, all my comments have been addressed :)

rock3r:
Ready to merge

.github/workflows/jewel-checks.yml
```diff
@@ -390,6 +410,155 @@ jobs:
...
+      # The IDE end-to-end test loads the JPS module graph, which includes the android modules.
+      - uses: actions/checkout@v7.0.1
+        name: Checkout JetBrains/android submodule
+        with:
+          repository: JetBrains/android
+          path: android
+          ref: master
+
+      # The IDE end-to-end test builds a dev IDE from every community module. Same preparation as the API check.
+      - name: Free disk space
```
wellingtoncosta, line 433: Not necessarily related to the purpose of this PR, but could we extract this into a reusable composite action since this step duplicates lines 203–245?

platform/jewel/showcase-e2e/src/main/kotlin/org/jetbrains/jewel/e2e/PopupScenarios.kt
```diff
@@ -0,0 +1,102 @@
...
+        automator.pressKey(KeyEvent.VK_ESCAPE)
+
+        return runCatching { automator.waitUntilGone(ShowcaseE2eTags.COMBO_BOX_POPUP) }
+            .fold(
+                onSuccess = { null },
+                onFailure = { "the popup stayed open, so the renderer claimed Escape without acting on it" },
+            )
+    }
+
+    /**
```
wellingtoncosta, line 81: Would it make sense to refactor `comboBoxPopupClosesCleanly` since it shares most of its logic with `escapeClosesHoveredComboBox`?

wellingtoncosta (approved):
Some comments, but LGTM

platform/jewel/showcase-e2e/src/main/kotlin/org/jetbrains/jewel/e2e/PopupScenarios.kt
DanielSouzaBertoldi: Honestly, I don't think so. `escapeClosesHoveredComboBox` hovers the combobox and presse the escape key and folds the result. `comboBoxPopupClosesCleanly` on the other hand just presses the escape key (there's no hover) but it also checks if the popup closed in time or not 

that's too many variables to put under one function, at least imo

.github/workflows/jewel-checks.yml
DanielSouzaBertoldi: Created a `jewel-prepare-runner-storage` composite action which not only frees disk space but also expands the swap pool 🙏🏻

PR #3660, opened 2026-09-20
IDEA-393994 Gradle: Use ResolutionResult API in GradleDependencyResolver

kroune:
GradleDependencyResolver used LenientConfiguration.getAllModuleDependencies() to list the resolved dependencies. This call builds the legacy ResolvedDependency graph. The graph keeps a node with several collections for each variant of each component. On large projects the graph costs hundreds of megabytes of heap during the model build.

The resolver does not use the graph structure. It needs only the flat set of resolved artifacts, the component coordinates, and the variant name of each artifact. The ArtifactView and ResolutionResult APIs give the same data without the legacy graph.

1. The ordering changed, but only from one stable sorting (the old one) to the new stable sorting, all places that rely on this don't pin the old one, they just need any stable sorting. I think that this change is ok.

2. After migration to the new version I had to change multiple other places, had to refactor them a bit too. This is why there are that many changes

**baseline** - https://github.com/kroune/feature-module-3000/actions/runs/35370400110
**after optimization** - https://github.com/kroune/feature-module-3000/actions/runs/35523607297
The change is around -450 mb

the issue this pr partially addresses https://youtrack.jetbrains.com/issue/IDEA-393994/GradleDependencyResolver-retains-legacy-ResolvedDependency-graph-during-sync-leading-to-wasted-RAM; https://github.com/gradle/gradle/issues/38752

note: _I ran with a custom gradle version (with other optimizations included), so that sync can finish successfully on 16 Gb github runner_


kroune:
looking at it again... The GradleDependencyResolver itself is very hard to read/update. May be it is worth to try to refactor it, I am not sure if it is worth to do in this pr or not though

